PR #9, opened 2020-06-11
Update landing page content and style

Nicell:
I've updated the styles a bit and added back the feature section with some things I think are important.


petejohanson:
@Nicell I dig it! One minor nitpick: I'm not sure I love the lack of the border on the "Get Started" call to action on the landing page:

Screenshot from 2020-06-10 23-04-46

What was the thinking on removing it?

Nicell:
Not on purpose. It was on in my dev testing. Must've been a caching issue in the CSS. I'm sending a fixed commit right now!

PR #14, opened 2020-06-12
Initial basic work on nice!nano board support.

petejohanson:
Completely untested, if this breaks you get to keep both halves!


Nicell:
Under [pin configuration of GPIO on the nRF52840 specification](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fps_nrf52840%2Fgpio.html&cp=4_0_0_5_8), it states that the two pin controllers combine the DETECT signal for other peripherals (such as SPI). To me I think this means that we can mix and match P0 and P1 pins within the SPI definition.

petejohanson:
@Nicell added the pin definitions in the DT file for now, with a note it will need testing to verify at some point. Thanks!

app/boards/arm/nice_nano/nice_nano.dts
```diff
@@ -0,0 +1,98 @@
...
+};
+
+/* TODO: Needs testing */
+&spi0 {
+	compatible = "nordic,nrf-spi";
+	/* Cannot be used together with i2c0. */
+	/* status = "okay"; */
+	sck-pin = <13>;
+	mosi-pin = <10>;
+	miso-pin = <11>;
```
Nicell, line 61:
```suggestion
	miso-pin = <43>;
```

app/boards/arm/nice_nano/nice_nano.dts
```diff
@@ -0,0 +1,98 @@
...
+	sda-pin = <17>;
+	scl-pin = <20>;
+};
+
+/* TODO: Needs testing */
+&spi0 {
+	compatible = "nordic,nrf-spi";
+	/* Cannot be used together with i2c0. */
+	/* status = "okay"; */
+	sck-pin = <13>;
```
Nicell, line 59:
```suggestion
	sck-pin = <45>;
```

petejohanson:
Merging. Tested working:

* Firmware update via UF2 image.
* USB HID device appears in Linux host.
* BLE peripheral appears when scanning from a Linux host.
* Using the `petejohanson_proton_handwire` shield which defines a matrix with a few pins, I was able to short 0th column + row, and `hid-recorder` in Linux properly picked up the key press/release, indicating the whole GPIO interrupts/kscan stuff, and USB HID layers are all working as expected.

PR #22, opened 2020-06-13
docs: FAQs (initial version)

### LINKED ISSUE #11: FAQs (initial version)
Initiail version of FAQs.  Required by #7.

Currently a realtime work-in-progress on Google Docs.  Please contact me on Discord if you need the link to contribute.

innovaker:
Closes #11.

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -1,0 +1,103 @@
+---
+id: faq
+title: FAQs
+sidebar_label: FAQs
+---
+
+### Why Zephyr™?
+
+As a best-in-class RTOS, Zephyr™ brings many [benefits](https://www.zephyrproject.org/benefits) to ZMK, such as:
+- A *single* platform [supporting](https://docs.zephyrproject.org/latest/boards) many architectures, processors and boards.
+- Optimization for low-powered, small memory footprint devices.
+- Powerful hardware abstraction and configuration using [DeviceTree](https://docs.zephyrproject.org/latest/guides/dts/index.html) and [Kconfig](https://docs.zephyrproject.org/latest/guides/kconfig/index.html).
+- A BLE stack that periodically obtains [qualification](https://docs.zephyrproject.org/latest/guides/bluetooth/bluetooth-qual.html) listings, making it easier for final products to obtain qualification from the Bluetooth® SIG.
+- Multi-processor support, which is critical for power efficiency in upcoming MCUs.
+- Permissive licencing with its Apache 2.0 open source [license](https://www.apache.org/licenses/LICENSE-2.0).
+- A buzzing developer [community](https://github.com/zephyrproject-rtos/zephyr) including many leading [embedded technology](https://www.zephyrproject.org/project-members) companies.
+- Long term support (LTS) with security updates.
+
+### Why yet another keyboard firmware?
+
+That’s an excellent question!  There are already great keyboard firmwares available, but ZMK has some advantages:
+
+- Zephyr™
+  - See [Why Zephyr™?](#why-zephyr)
+- Licensing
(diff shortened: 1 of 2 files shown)


docs/docs/faq.md
```diff
@@ -0,0 +1,103 @@
+---
+id: faq
+title: FAQs
+sidebar_label: FAQs
+---
+
+### Why Zephyr™?
```
petejohanson, line 7: Making these an `##` or `#` should make a nice right sidebar TOC appear, let folks jump to the FAQ they want

docs/sidebars.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   someSidebar: {
-    'Getting Started': ["intro", "hardware"],
+    'Getting Started': ["intro", "hardware", "faq"],
```
petejohanson, line 3: I would put this before Hardware, especially given that doc isn't as refined. Thoughts?

innovaker: sounds good!

docs/docs/faq.md
innovaker: It appears anyway.  I used ### because # and ## was way too big and I couldn't find an easy grouping.  Happy to change it if we want to change the CSS too.

Nicell: Yeah the ### seems to work on the right sidebar. I think that its size is nice. If we don't want to change the CSS, I think ### is good.

docs/docs/faq.md
```diff
@@ -0,0 +1,103 @@
...
+
+### What is a “shield”?
+
+In ZMK, a *shield* is a *PCB* or *hardwired set of components* that when combined with a MCU only board like the Proton-C or nice!nano, results in a complete usable keyboard. Examples would be keyboard PCBs like the Kyria or Corne.  The shield is usually the big PCB containing all the keys.
+
+### Why *boards* and *shields*?  Why not just “keyboard”?
+
+When a keyboard accepts a small “PCB MCU module” (e.g. *Arduino Pro Micro*) for its “brains”, then it's important to conceptually separate the hardware into a *board* PCB and a *shield* PCB.
+
+The *shield* is a brainless shell containing all the keys, RGB LEDs, encoders etc.  It maps all of these features to a standard pin footprint, such as the Pro Micro pinout.
```
Nicell, line 65: Without giving an example of what a shield is, I think someone could read this as the Pro Micro itself being a shield, which is wrong. Maybe link to "What is a 'Shield'?" for examples (in case someone links this question directly).

docs/docs/faq.md
petejohanson: Odd. I didn't see it from my phone. Let's leave it as is then

docs/docs/faq.md
```diff
@@ -0,0 +1,103 @@
...
+
+To bring this brainless *shield* to life, you attach any MCU *board* matching the footprint.  For instance, the *nice!nano* is *pin-compatible* with the *Arduino Pro Micro*, so you can substitute either *board* onto the *shield*.  But each *board* comes with its own features (MCU, flash, BLE, etc.) which must also be handled.
```
Nicell, line 67: Same thing here. Maybe link to "What is a board?" especially because this doesn't outline full completed keyboards.

docs/docs/faq.md
innovaker: A valid point.  I assumed that they would've already read the two FAQs preceding this question.

innovaker:
If the reader still can't gather what a board and shield is, then I pass the baton onto someone else.

Good luck to anyone editing this page.  Be careful of all the hyperlinks that use auto-generated anchors.

docs/docs/faq.md
innovaker: Added.

docs/docs/faq.md
innovaker: Added.

PR #29, opened 2020-06-15
Initial refactor to have a shield/board have one default keymap.

petejohanson:
* Start of work of focusing on out-of-tree user keymaps.

Add ZMK_CONFIG support.

* General external configuration supported via ZMK_CONFIG,
  which should point to a configuration directory.
* Priority order of overlay files to select.
  * ${SHIELD}_${BOARD}.overlay
  * ${SHIELD}.overlay
  * ${BOARD}.overlay
  * default.overlay
* Equivalent search order for .conf files for initial Kconfig
  values.

You can see a sample repo with a bunch of different configs at https://github.com/petejohanson/zmk-config-examples which also includes a working GH Action. In reality, that repo would normally just have one subdirectory, named `config`.


app/boards/shields/petejohanson_proton_handwire/keymap/keymap.c
```diff
@@ -10,22 +10,22 @@ bool zmk_handle_key_user(struct zmk_key_event *key_event)
 	case CC_LOWR:
 		if (key_event->pressed)
 		{
-			zmk_keymap_layer_activate(1);
+			zmk_keymap_layer_activate(LOWER);
```
Nicell, line 13: How do we plan to abstract layer state/changes when dealing with runtime keymap/layer changes?

innovaker:
I need to look at this again in more detail.

I'm also keen to discuss the impacts of this approach on discoverability of other people's configurations.

app/boards/shields/petejohanson_proton_handwire/keymap/keymap.c
petejohanson: This is probably pure configuration moving forward. In our discussions on Discord instead of this custom code, this would instead be something like an action assigned to this position with:

* Action Type: Momentarily Activate Layer (`MO` macro in QMK)
* Parameter: Layer # To Activate

In current DT defined keymap land, this would be just an entry in the keymap like `MO(LOWER)` and the corresponding "Action Type Implementation" that would handle the press and release events from that logical location.

petejohanson:
@innovaker Yeah, discoverability of other's configurations we discussed on Discord the other day. @Nicell and I discussed some options:

* Short Term: The equivalent of the various "Awesome" repos for other projects, but for ZMK, e.g. `awesome-zmk` that the community manages to publish links to ZMK user config repos.
* Long Term: A forum/marketplace location where users can publish their configs, document them, get them rated, etc.

I'm happy to entertain other ideas though.

petejohanson:
Merging. We can continue brainstorming options to improve discoverability.

PR #31, opened 2020-06-15
Initial work on matrix transforms for non-standard matrixes.

petejohanson:
* Allows "gaps" for keyboards like Kyra.
* Allows keyboards with matrixes that don't map the logical key
  locations to have normal looking keymaps when defined.

This allows creating a Kyria keymap that doesn't have lots of empty/meaningless `ZC_NO` entries.

In the default keymap for each keyboard, you can define an optional "keymap transform" that defines how you map each entry in the keymap to a specific row/column in the physical matrix.


app/boards/shields/kyria/keymaps/default/keymap.overlay
```diff
@@ -0,0 +1,50 @@
...
+		compatible = "zmk,keymap";
+		label ="Default Kyria Keymap";
+		layers = <&default>;
+		transform = <&default_transform>;
+	};
+
+	default_transform: keymap_transform_0 {
+		compatible = "zmk,keymap-transform";
+		map = <
+			IDENTITY_ROW(0, 6)       OFFSET_ROW(6, 0, 6)
```
petejohanson, line 20: This uses some higher level wrappers to make this quick. I will tweak the include file a bit so we can also have per item defines here, e.g. `<POS(1, 2) POS(2,1) ... >` so you can manually define the mapping for each entry to the matrix row/column explicitly.

app/boards/shields/kyria/keymaps/default/keymap.overlay
```diff
@@ -0,0 +1,50 @@
...
+			IDENTITY_ROW(1, 6)       OFFSET_ROW(6, 1, 6)
+			IDENTITY_ROW(2, 8)       OFFSET_ROW(8, 2, 8)
+			OFFSET_ROW(3, 3, 5)      OFFSET_ROW(8, 3, 5)
+		>;
+	};
+
+	five_column_transform: keymap_transform_1 {
```
petejohanson, line 27: This will *nicely* let us define transforms for keyboards that have breakaway PCB sections, so you can have your user config select the different transform, and now you can use the keymap that doesn't list that extra keys that aren't actually there for your physical keyboard.

app/boards/shields/kyria/keymaps/default/keymap.overlay
```diff
@@ -0,0 +1,50 @@
...
+		>;
+	};
+
+	layers {
+		compatible = "zmk,layers";
+
+		default: layer_0 {
+			label = "DEFAULT";
+			keys = <
+	KC_ESC KC_Q KC_W KC_E KC_R KC_T                                   KC_Y KC_U KC_I KC_O KC_P KC_PIPE
```
petejohanson, line 43: You can start to see how we could make this visually easier to understand, like QMK keymaps. I will clean this up even more in this PR to show it more completely.

petejohanson: Ok, pushed an update here that shows the "logical layout" via comments for both the *transform* (shows the MX## pad details), and the keymap layers.

app/boards/shields/kyria/keymaps/default/keymap.overlay
petejohanson: Updated the naming here, and made a `KT_ROW_COL` macro as well, to let you handcode *all* the mappings, if the `KT_ID_ROW` and `KT_OFFSET_ROW` aren't usable when doing a keyboard wit a funky matrix layout.

app/boards/shields/kyria/keymaps/default/keymap.overlay
Nicell: I like the final result of the keys=<>. but the map with offset row helpers seems a bit un-intuitive to me compared to a defined function that takes the keys in as parameters (and can be styled however with white space), and then the function returns the proper matrix with the keys in the right place. I imagine you might be constrained with these files though.

innovaker:
To summarise following an extensive chat on Discord between @innovaker, @petejohanson and @Nicell.
- The principle of mapping the physical matrix to a logical list of positions is sound.  Great idea Pete!  I think the consensus was that we call it the 'matrix transform' rather than a keymap transform.
- The utility functions in the transform don't necessary aid comprehension.  So a simple array of coordinates might be more intuitive.

As for the labelling of those positions with key names (codes) within the DT, I'm still personally unsure and I think it will need revisiting.  We probably need to consider how these ...
- config as JSON (or similar)
- config from nvs
- config API (for real time changes over USB or BLE)
... might look before settling on that.

So, LGTM until we have more progress from a design perspective!

app/boards/shields/kyria/keymaps/default/keymap.overlay
petejohanson: Ok, so completely removed these "helper" macros just now, in favor of smaller, per-entry macros for each individual location, for clarity on the use here. Marking this one resolved.

app/boards/shields/kyria/keymaps/default/keymap.overlay
petejohanson: Ok, so after discussion on Discord, this now looks like this:

Screenshot from 2020-06-16 21-29-22

using just one macro `RC` to map a logical location to a matrix row + column.

app/boards/shields/kyria/keymaps/default/keymap.overlay
petejohanson: Marking this resolved, after discussion on Discord.

petejohanson:
Addressed the macro and naming items from @innovaker, I'm going to merge this as is, and we'll continue our keymap/action discussions as we move forward.

PR #32, opened 2020-06-22
Refactor keymap to behaviors

petejohanson:
This is the draft PR for the refactor to "behaviors" to implement saner keymaps, true single responsibility principle for what used to be called "actions", and set the foundation for split keyboards propagating "position state" to the hub side, for handling.

A high level summary of working behavior:

* Refactored out "matrix transforms" into a separate phase, only leveraged by the core `kscan.c` handler in ZMK, which pushes *everything* above this to work on "key positions" instead of row/column values.
  * Key positions are a numeric value that represent the index into the "ordered list" of logical key positions on the keyboard.
  * Matrix transforms transform from matrix row/column pairs to the correct key position. The "default" matrix transform is the logic of "($row * NUMBER_OF_ROWS) + $column", which effectively "flattens" the rows one after another into one long row, and takes their indexes.
* Behaviors are anything that listens to either *global* events (e.g. "position pressed", or "keycode pressed" or "modifiers pressed"), or listens for "keymap binding pressed/released" events, if they are bound to a specific key position in the keymap.
  * Behaviors can return:
    * Negative number to indicate an error, and stop processing of an event.
    * Zero to indicate success, and stop processing of the event.
    * A positive number to indicate success, and continue processing the event.
  * Keymap bound behaviors are bound with two extra `u32_t` parameters at the time of binding (e.g. `&kp KC_A` or `&mt MOD_LSHFT KC_A`, or `&mo LAYER_RAISE`), which are passed to that behaviors `binding_pressed` and `binding_released` callbacks as parameters, allowing them to take action specific to the binding of that behavior in that specific key position.
  * Behaviors are specified in the DT, there is one "roll up" include file that pulls them all in.
    * Core keymaps would have `#include <behaviors.dtsi>` which includes all behaviors.
    * Eventually, for *complex* keymap binding behaviors, we can configure the *behavior itself* (not the bindings of it) with extra data, e.g. the "tapping term" timeout, or perhaps the "chord combinations".
    * To be added to listen for global events, the DT declaration needs to include "zmk,behavior-global" in the `compatible` list.
* Behaviors are implemented as a driver, with a set of optional callbacks for both global, and keymap binding events. See `include/drivers/behavior.h` for the set of possible callbacks a behavior can implement.
* For split support, the global "hid behavior" that currently listens to mod and keycode pressed/released to send reports, will *also* listen to "position changed" events when in "spoke" mode, and send *those* over to the hub that will be listening for them, and then pass them into the keymap processing flow just as if they had originated locally.
* I've manually renamed a lot of the HID constants as a temporary measure, so they are now just `A`, `B`, etc. to make the bindings look like `&kp A` in the overlay files.
* I haven't fixed consumer key support yet. I envision doing this with *another* behavior, instead of trying to hijack some random bits of the keycode stuff. Will probably turn that into refactoring they "keycode_pressed" and "keycode_released" global behavior callbacks to indicate which usage page the value is from.

I will update the above as I think of more.


app/boards/shields/kyria/keymap/keymap.overlay
```diff
@@ -0,0 +1,34 @@
...
+
+		default: layer_0 {
+			label = "DEFAULT";
+// ---------------------------------------------------------------------------------------------------------------------------------
+// |  ESC  |  Q  |  W  |  E   |  R   |  T   |                                          |  Y   |  U    |  I    |  O   |   P   | "|" |
+// |  <-   |  A  |  S  |  D   |  F   |  G   |                                          |  H   |  J    |  K    |  L   |   ;   |  '  |
+// | SHIFT |  Z  |  X  |  C   |  V   |  B   | L SHIFT | L SHIFT |  | L SHIFT | L SHIFT |  N   |  M    |  ,    |  .   |   /   |  -  |
+//                     | GUI  | DEL  | RET  |  SPACE  |   ESC   |  |   RET   |  SPACE  | TAB  | BSPC  | R-ALT |
+			bindings = <
+	&kp ESC  &kp Q &kp W &kp E &kp R &kp T                                            &kp Y &kp U  &kp I    &kp O   &kp P    &kp PIPE
```
petejohanson, line 27: A more realistic keymap after the refactor.

The behaviors are exposed as DT names you can reference. In this PR are:

* `&kp A` - key press
* `&mo 1` - momentary layer.
* `&mt MOD_LSFT A` - mod-tap behavior.
* `&reset` - device reset (doesn't enter DFU yet)
* `&trans` - Transparent, used on layes above `&mo` keys mostly.

app/boards/shields/kyria/kyria.overlay
```diff
@@ -28,6 +33,34 @@
 		};
 	};
 
+	default_transform: keymap_transform_0 {
```
petejohanson, line 36: Moved these transforms *out* of the keymap, and into the device itself.

In the future, we'd have "variants" for keyboards for breakaway versions, e.g. `kyria_v1_five_column.overlay` that would have the alternate transform.

app/dts/behaviors/hid.dtsi
```diff
@@ -0,0 +1,9 @@
+/ {
+	behaviors {
+		hid_behavior: behavior_hid {
+			compatible = "zmk,behavior-hid", "zmk,behavior-global";
```
petejohanson, line 4: Here is a global behavior. It listens for events and propagates HID updates to hosts (and eventually hubs from a spoke)

app/dts/behaviors/mod_tap.dtsi
```diff
@@ -0,0 +1,9 @@
+/ {
+	behaviors {
+		mt: behavior_mod_tap {
+			compatible = "zmk,behavior-mod-tap", "zmk,behavior-global";
```
petejohanson, line 4: Mod tap is both, a keymap bound and global bound behavior. This is because if there is *some other* keycode pressed while it is held, then it avoids sending the bound keycode, and is used just as a modifier. We can expect other behaviors to act like this too, like layer-tap.

app/dts/behaviors/momentary_layer.dtsi
```diff
@@ -0,0 +1,9 @@
+/ {
+	behaviors {
+		mo: behavior_momentary_layer {
+			compatible = "zmk,behavior-momentary-layer";
```
petejohanson, line 4: Use this instead of those custom keycodes found in QMK keymaps.

app/dts/bindings/zmk,layers.yaml
```diff
@@ -9,6 +9,7 @@ child-binding:
   properties:
     label:
       type: string
-    keys:
-      type: array
-
+      required: true
+    bindings:
```
petejohanson, line 13: Renamed this to "bindings" since there's more than just "keys" in here, there's behaviors to bind to the key positions.

Thoughts on the naming?

app/include/drivers/behavior.h
```diff
@@ -0,0 +1,234 @@
+/*
```
petejohanson, line 1: This is the driver API for behaviors.

app/src/behaviors/behavior_mod_tap.c
```diff
@@ -0,0 +1,89 @@
...
+
+// They keycode is passed by the "keymap" based on the parameter created as part of the assignment.
+static int on_keymap_binding_released(struct device *dev, u32_t position, u32_t mods, u32_t keycode)
+{
+  struct behavior_mod_tap_data *data = dev->driver_data;
+  LOG_DBG("mods: %d, keycode: %d", mods, keycode);
+  
+  zmk_events_modifiers_released(mods);
+  if (data->pending_press_positions & BIT(position)) {
+    zmk_events_keycode_pressed(keycode);
```
petejohanson, line 45: Feeds data back into the event loop.

app/src/hid.c
```diff
@@ -66,19 +66,19 @@ int zmk_hid_unregister_mods(zmk_mod_flags modifiers)
...
 
-    if (ZK_IS_CONSUMER(code))
-    {
-        LOG_DBG("Toggling a consumer key!");
-        TOGGLE_CONSUMER(code, true);
-        return Consumer;
-    }
-    else
-    {
+    // if (ZK_IS_CONSUMER(code))
```
petejohanson, line 74: We'll refactor this a bit, since this won't be needed if we're not shoving a subset of consumer page values into the higher bits of the keycode constants.

app/src/keymap.c
```diff
@@ -1,24 +1,32 @@
...
 
-#if DT_NODE_HAS_PROP(ZMK_KEYMAP_NODE, transform)
-#define ZMK_KEYMAP_TRANSFORM_NODE DT_PHANDLE(ZMK_KEYMAP_NODE, transform)
-#define ZMK_KEYMAP_LEN DT_PROP_LEN(ZMK_KEYMAP_TRANSFORM_NODE, map)
+#define ZMK_KEYMAP_NODE DT_CHOSEN(zmk_keymap)
+#define ZMK_KEYMAP_LAYERS_LEN DT_PROP_LEN(ZMK_KEYMAP_NODE, layers)
+
+#define LAYER_NODE(l) DT_PHANDLE_BY_IDX(ZMK_KEYMAP_NODE, layers, l)
 
-#define _TRANSFORM_ENTRY(i, l) \
```
petejohanson, line 15: This all got simplified when the matrix transforms were moved *lower*, so the keymap *only* operates on key positions now, not on rows/columns at all.

app/src/kscan.c
```diff
@@ -49,11 +49,13 @@ void zmk_kscan_process_msgq(struct k_work *item)
 	while (k_msgq_get(&zmk_kscan_msgq, &ev, K_NO_WAIT) == 0)
 	{
 		bool pressed = (ev.state == ZMK_KSCAN_EVENT_STATE_PRESSED);
-		zmk_key key = zmk_keymap_keycode_from_position(ev.row, ev.column);
-		struct zmk_key_event kev = (struct zmk_key_event){.row = ev.row, .column = ev.column, .key = key, .pressed = pressed};
-
-		LOG_DBG("Row: %d, col: %d, key: %d, pressed: %s\n", ev.row, ev.column, key, (pressed ? "true" : "false"));
-		zmk_handle_key(kev);
+		u32_t position = zmk_matrix_transform_row_column_to_position(ev.row, ev.column);
```
petejohanson, line 52: Matrix transforms here now only.

app/src/matrix_transform.c
```diff
@@ -0,0 +1,30 @@
...
+
+#define _TRANSFORM_ENTRY(i, _) \
+	[(KT_ROW(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i)) * ZMK_MATRIX_COLS) + KT_COL(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i))] = i,
+
+static u32_t transform[] = 
+    { UTIL_LISTIFY(ZMK_KEYMAP_LEN, _TRANSFORM_ENTRY, 0) };
+
+#endif
+
+u32_t zmk_matrix_transform_row_column_to_position(u32_t row, u32_t column)
```
petejohanson, line 21: New separate matrix transform logic. Short and sweet.

app/boards/native_posix.overlay
```diff
@@ -53,27 +54,26 @@
...
 
 		lower: layer_1 {
 			label = "LOWER";
-			keys = <
-				KC_A KC_B KC_C KC_D
-				KC_E KC_F KC_G KC_H
-			>;
+
+			bindings = <
+				&cp M_NEXT &trans &kp O &kp Q
```
petejohanson, line 67: We can come up with a better shortcut for this, but I now added `&cp` to reference the "Consumer Press" when someone pushes a consumer page key, which allows just using the actual consumer usage values directly, instead of the previous keycode hack. We could call this shorthand something else, e.g. `&mp` for "Media Press" since most end users think of these as media keys.

app/dts/behaviors/key_press.dtsi
```diff
@@ -0,0 +1,19 @@
...
+/ {
+	behaviors {
+		kp: behavior_key_press {
+			compatible = "zmk,behavior-key-press";
+			label = "KEY_PRESS";
+			usage_page = <USAGE_KEYPAD>;
+			#binding-cells = <1>;
+		};
+
+		cp: behavior_consumer_press {
```
petejohanson, line 12: I tweaked this a bit to be more generic. This is *reusing* the same "key press behavior" but setting a different usage page in the instance of the driver. This reduces duplicate code, leaves us a consistent internal API we can leverage for the eventual "key position usage page" for the split stuff.

app/include/dt-bindings/zmk/keys.h
```diff
@@ -1,122 +1,120 @@
...
-#define KC_RSFT 0xE5
-#define KC_RALT 0xE6
-#define KC_RGUI 0xE7
-
-#define KC_VOLU 0x80
-#define KC_VOLD 0x81
+#define USAGE_KEYPAD 0x07
+#define USAGE_CONSUMER 0x0C
+
+#define A 0x04
```
petejohanson, line 7: Made these shorter, pending the HID usage tables work from @innovaker. With this refactor we can use both the keypad and the consumer usages *directly*, we don't need any "mash the keycodes together" gymnastics.

app/include/dt-bindings/zmk/keys.h
```diff
@@ -1,122 +1,120 @@
...
-#define KC_MPLY 0x104
-#define KC_MMUT 0x105
-#define KC_MVLU 0x106
-#define KC_MVLD 0x107
-
-#define ZC_TRNS (0xFFFF)
-#define ZC_NO (0xFFFF - 1)
-
-#define ZC_CSTM(n) (0xFFF + n)
+#define M_NEXT 0xB5
```
petejohanson, line 110: Small subset of the "media keys", i.e. consumer page codes.

app/include/zmk/hid.h
```diff
@@ -108,40 +108,21 @@ static const u8_t zmk_hid_report_desc[] = {
...
     /* INPUT (Data,Ary,Abs) */
-    /* REPORT_SIZE (1) */
+    /* REPORT_SIZE (8) */
     HID_GI_REPORT_SIZE,
-    0x01,
+    0x08,
     /* REPORT_COUNT (8) */
     HID_GI_REPORT_COUNT,
-    0x08,
+    0x06,
```
petejohanson, line 123: For consumer page stuff, I implemented "6 key roll over" only, since that's all we really need for this. We could even get away with "4KRO" if we wanted, I don't expect a need for holding a lot of consumer keys at once.

app/src/events.c
```diff
@@ -0,0 +1,83 @@
...
+#include <zmk/behavior.h>
+#include <zmk/events.h>
+#include <sys/util.h>
+
+#define DT_DRV_COMPAT zmk_behavior_global
+#define GLOBAL_BEHAVIOR_LEN DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT)
+
+#define LABEL_ENTRY(i) DT_INST_LABEL(i),
+static const char *global_behaviors[] = {
+    DT_INST_FOREACH_STATUS_OKAY(LABEL_ENTRY)
```
petejohanson, line 13: This is how we pick up all behaviors w/ the extra "zmk,behavior-global" compatible value, and pick those to notify of global events.

petejohanson:
Merging this. We'll continue to iterate on the events/actions piece to "clean up" behaviours and then move forward from there.

PR #64, opened 2020-07-25
Initial RGB Underglow implementation

Nicell:
Example usage can be found here: https://github.com/Nicell/zmk-config

Looking forward to comments on where I can improve my implementation.


app/src/behaviors/behavior_rgb_underglow.c
```diff
@@ -0,0 +1,80 @@
...
+  return 0;
+}
+
+static int on_keymap_binding_pressed(struct device *dev, u32_t position, u32_t action, u32_t _)
+{
+  switch (action)
+  {
+    case RGB_TOG:
+      zmk_rgb_underglow_toggle();
+      break;
```
petejohanson, line 32: Instead of all the breaks, etc. since the action is really the *only* think you want to do, you might consider refactoring all of these to:

```
case RGB_TOG:
  return zmk_rb_underglow_toggle();
case BLAH:
```

And then the very end of the function can just `return -ENOTSUP;` if we didn't match a supported action.

This requires each of those called functions return `int` for success/errno, which is probably good anyways.

app/include/zmk/rgb_underglow.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+void zmk_rgb_underglow_toggle();
+void zmk_rgb_underglow_cycle_effect(int direction);
+void zmk_rgb_underglow_change_hue(int direction);
+void zmk_rgb_underglow_change_sat(int direction);
+void zmk_rgb_underglow_change_brt(int direction);
+void zmk_rgb_underglow_change_spd(int direction);
```
petejohanson, line 8: These should return `int` probably.

app/src/rgb_underglow.c
```diff
@@ -0,0 +1,295 @@
...
+
+#include <logging/log.h>
+
+#include <drivers/led_strip.h>
+#include <device.h>
+#include <drivers/spi.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#define STRIP_LABEL		    DT_LABEL(DT_ALIAS(led_strip))
```
petejohanson, line 22: Instead of an `alias`, we should be using a chosen node, e.g. `zmk,underglow = &led_strip` in the DT.

app/src/rgb_underglow.c
```diff
@@ -0,0 +1,295 @@
...
+#include <kernel.h>
+
+#include <math.h>
+#include <stdlib.h>
+
+#include <logging/log.h>
+
+#include <drivers/led_strip.h>
+#include <device.h>
+#include <drivers/spi.h>
```
petejohanson, line 18: Probably not needed?

app/src/rgb_underglow.c
```diff
@@ -0,0 +1,295 @@
...
+    state.animation_step += state.animation_speed * 2;
+    state.animation_step = state.animation_step % 360;
+}
+
+static void zmk_rgb_underglow_tick(struct k_work *work)
+{
+    switch (state.current_effect)
+    {
+        case UNDERGLOW_EFFECT_SOLID:
+            zmk_rgb_underglow_effect_solid();
```
petejohanson, line 155: Perhaps future thought... but for size sake later, we should probably make each effect be enabled/disabled via Kconfig setting.

app/src/rgb_underglow.c
```diff
@@ -0,0 +1,295 @@
...
+K_TIMER_DEFINE(underglow_tick, zmk_rgb_underglow_tick_handler, NULL);
+
+static int zmk_rgb_underglow_init(struct device *_arg)
+{
+	led_strip = device_get_binding(STRIP_LABEL);
+	if (led_strip) {
+		LOG_INF("Found LED strip device %s", STRIP_LABEL);
+	} else {
+		LOG_ERR("LED strip device %s not found", STRIP_LABEL);
+		return 1;
```
petejohanson, line 187:
```suggestion
		return -EINVAL;
```

app/src/rgb_underglow.c
```diff
@@ -0,0 +1,307 @@
...
+    }
+    
+    state.animation_step = 0;
+
+    return 0;
+}
+
+int zmk_rgb_underglow_toggle()
+{
+    state.on = !state.on;
```
petejohanson, line 224: Something I missed in the first pass. There needs to be some early checks in any of these functions to be sure we actually have an `led_strip` device.

If init failed... the behaviors still can be triggered by the keymap, or other folks might try to call into this, so checks in various "entrypoints" to this should assert the `led_strip` is there and exit early otherwise.

Nicell: Should the behaviors file still be gated by the ZMK_RGB_UNDERGLOW Kconfig value in this case? Also technically the only behavior function that interacts directly with the led strip would be zmk_rgb_underglow_toggle. Should this be the only one that needs to make this check, or should I check in all of the non-static functions so I can make a special return code?

petejohanson: Probably good for all to check. Even though they don't technically interact with the device, the behaviour still failed to perform the requested operation. That should be reported so it can bubble up to higher layers. (E.g. later we might do some reporting to OLEDs, or elsewhere, etc.

petejohanson: And it should still be gated, so that folks not using underglow don't need that code increasing the size of their final image.

PR #70, opened 2020-08-02
Split connection parameter adjustments and 2M PHY added

Nicell:
- Update central split half to only accept 7.5ms
- Update latency from 5 to 30 (maximum officially supported by Apple, good benchmark)
- Add 2M PHY to connection between split halves
- Adds callback function to stop split connection params from resetting to default

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -40,7 +40,7 @@ static void connected(struct bt_conn *conn, u8_t err)
     bt_conn_le_param_update(conn, BT_LE_CONN_PARAM(0x0006, 0x000c, 30, 400));
 
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_ROLE_PERIPHERAL)
-    bt_conn_le_phy_update(default_conn, BT_CONN_LE_PHY_PARAM_2M);
+    bt_conn_le_phy_update(conn, BT_CONN_LE_PHY_PARAM_2M);
 #endif
 
     if (bt_conn_set_security(conn, BT_SECURITY_L2))
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -76,10 +76,39 @@ static void security_changed(struct bt_conn *conn, bt_security_t level,
     }
 }
 
+static bool le_param_req(struct bt_conn *conn, struct bt_le_conn_param *param) {
+    static struct bt_conn_info info;
+
+    bt_conn_get_info(conn, &info);
+
+    if (info.role == BT_CONN_ROLE_MASTER && (param->interval_min != 6 || param->latency != 30)) {
+        return false;
+    }
+
+    return true;
+}
+
+static void le_param_updated(struct bt_conn *conn, u16_t interval,
+				 u16_t latency, u16_t timeout) {
+    static struct bt_conn_info info;
+
+    bt_conn_get_info(conn, &info);
+
+    if (info.role == BT_CONN_ROLE_MASTER && (interval != 6 || latency != 30)) {
+        bt_conn_le_param_update(conn, BT_LE_CONN_PARAM(0x0006, 0x0006, 30, 400));
+    }
+
+    LOG_DBG("Params updated: Interval: %d, Latency: %d", interval, latency);
+};
+
 static struct bt_conn_cb conn_callbacks = {
     .connected = connected,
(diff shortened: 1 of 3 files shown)


app/src/split/bluetooth/central.c
```diff
@@ -67,6 +67,13 @@ static u8_t split_central_notify_func(struct bt_conn *conn,
 		}
 	}
 
+	bt_conn_le_param_update(conn, BT_LE_CONN_PARAM(0x0006, 0x0006, 30, 400));
+
+	struct bt_conn_info info;
+
+	bt_conn_get_info(conn, &info);
+
+	LOG_DBG("Interval: %d, Latency: %d, PHY: %d", info.le.interval, info.le.latency, info.le.phy->rx_phy);
```
Nicell, line 76: This section needs to not happen on each keypress from the peripheral. Maybe we should have a check to see if the interval and latency is incorrect inside here as a fix? It may be more efficient to just update every single time though.

app/src/ble.c
```diff
@@ -72,10 +76,39 @@ static void security_changed(struct bt_conn *conn, bt_security_t level,
...
+};
+
 static struct bt_conn_cb conn_callbacks = {
     .connected = connected,
     .disconnected = disconnected,
     .security_changed = security_changed,
+#if !IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_ROLE_PERIPHERAL)
+    .le_param_req = le_param_req,
+    .le_param_updated = le_param_updated,
+#endif
```
Nicell, line 111: Either of these two callbacks will fix the issue of default params getting set after split halves connect. `le_param_req` stops it from changing, but it also "hijacks" the update system. `le_param_updated` updates the settings again after they've already been negotiated and changed, but it doesn't "hijack" the param update system.

@petejohanson which of these should we keep in your opinion? Or should we still try to track down the root cause of the default params getting set a few seconds into the split connection?

Nicell:
The connection parameters used by the split connection (and also the host device and central half connection) should probably be turned into constants/defines, as they're being used in multiple places now.

app/src/ble.c
petejohanson: Reading through this, I much prefer using the `le_param_req` callback for this. Just based on reviewing the documentation from Zephyr's `include/bluetooth/conn.h`:

```
        /** @brief LE connection parameter update request.
         *
         *  This callback notifies the application that a remote device
         *  is requesting to update the connection parameters. The
         *  application accepts the parameters by returning true, or
         *  rejects them by returning false. Before accepting, the
         *  application may also adjust the parameters to better suit
         *  its needs.
         *
         *  It is recommended for an application to have just one of these
         *  callbacks for simplicity. However, if an application registers
         *  multiple it needs to manage the potentially different
         *  requirements for each callback. Each callback gets the
         *  parameters as returned by previous callbacks, i.e. they are not
         *  necessarily the same ones as the remote originally sent.
         *
         *  @param conn Connection object.
         *  @param param Proposed connection parameters.
         *
         *  @return true to accept the parameters, or false to reject them.
         */
        bool (*le_param_req)(struct bt_conn *conn,
                             struct bt_le_conn_param *param);
```

app/src/ble.c
```diff
@@ -72,10 +76,25 @@ static void security_changed(struct bt_conn *conn, bt_security_t level,
     }
 }
 
+static bool le_param_req(struct bt_conn *conn, struct bt_le_conn_param *param) {
+    static struct bt_conn_info info;
+
+    bt_conn_get_info(conn, &info);
+
+    if (info.role == BT_CONN_ROLE_MASTER && (param->interval_min != 6 || param->latency != 30)) {
```
petejohanson, line 84: Why we are doing this is probably going o be none obvious to later readers of this code... Can we add a short comment here to hint what is being checked here and why?

app/src/ble.c
```diff
@@ -72,10 +76,25 @@ static void security_changed(struct bt_conn *conn, bt_security_t level,
     }
 }
 
+static bool le_param_req(struct bt_conn *conn, struct bt_le_conn_param *param) {
```
petejohanson, line 79: Since this function is unused unless this is a peripheral, the whole function should be wrapped in an `#if`

PR #78, opened 2020-08-09
Add RGB documentation

Nicell:
This adds initial documentation for lighting behaviors as well ad RGB underglow. Step on adding RGB underglow could probably be improved. Suggestions welcome.


petejohanson:
@Nicelll Thanks for working on this!

Thinking about this a bit, I wonder if we might want to have this content split between two audiences:

The current "features" location with pure end user content - what the feature is, demo video, how to turn it on for your board, etc.

And a separate doc, perhaps in a section under the "adding a new board/shield" doc, that is focused on the config and setup needed to add support for a given shield?

The bottom of the end user doc could have a note "Underglow not supported on your board yet? Head over here!"

Thoughts?

PR #79, opened 2020-08-09
core/keymap-release-key-after-layer-deactivated-fix

### LINKED ISSUE #67: Layer deactivation with held key position leads to incorrect key code release
As reported by @Nicell, the following sequence does not work as expected.

1. Activate layer (e.g. by pressing a key position bound to `&mo 1`.
1. Activate a different key position bound with a behavior (e.g. `&kp ESC`) on that newly activated layer.
1. Deactivate the layer (e.g. by releasing the key position bound to `&mo 1`).
1. Release the key position activated in step #2

On the last step, because layer `1` is no longer active, then the behavior that was triggered for the *activation* is *not* triggered for the deactivation. Usually, this leads to sending a "key code release" of the key bound on the *lower* layer instead.

The proposed fix would be something like:

* Whenever a behavior's `behavior_keymap_binding_pressed` returns `0`, that indicates that the behaviour has "handled" the key position activation, and processing should not continue to lower layers.
* When this happens, we should record the layer that "handled" that press. (e.g. `u8_t key_position_pressed_layer_handler[ZMK_KEYMAP_LEN]`
* When we receive a key position released event, the key map should *include* that recorded layer in the "stack" of layers it checks for behaviors to invoke, even if that layer isn't in the current active layer set.

petejohanson:
* If you press a key with a layer active, then
  deactivate the layer (e.g. releasing a `&mo`,
  then release the key, we currently may send
  the wrong key release event.
* Fixes #67.

### DIFF
--- a/app/src/keymap.c
+++ b/app/src/keymap.c
@@ -34,9 +34,6 @@ static u8_t zmk_keymap_layer_default = 0;
 #define TRANSFORMED_LAYER(node) \
   { UTIL_LISTIFY(DT_PROP_LEN(node, bindings), _TRANSFORM_ENTRY, node) },
 
-static struct zmk_behavior_binding zmk_keymap[ZMK_KEYMAP_LAYERS_LEN][ZMK_KEYMAP_LEN] = {
-	DT_INST_FOREACH_CHILD(0, TRANSFORMED_LAYER)
-};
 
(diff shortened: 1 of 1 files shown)


petejohanson:
@Nicell, can you give this branch a test? Seems to work well here.

PR #83, opened 2020-08-11
macOS compatability fixes for setup.sh

petejohanson:
macOS is pinned to bash 3.2, which does not include support for `read -i`, as reported on Discord. Revert the couple lines that switched to using that so we *should* work on macOS, hopefully.


petejohanson:
Tested by @keyboard-magpie and created his properly built repo here: https://github.com/keyboard-magpie/zmk-config/actions

PR #90, opened 2020-08-15
Add DCDC setting for nano.

petejohanson:


Nicell:
Testing this on the nice60 (which has SPI running), idle usage is down from 300uA to 260uA. Not huge gains, but still gains nonetheless!

Nicell:
Testing with Lily58 left connected to peripheral and to a host device:

Idle before: 760uA
Idle after: 630uA 

Peripheral typing (power usage from left still) before: 1000uA
Peripheral typing (power usage from left still) after: 750uA

Central typing before: 1050uA
Central typing after: 920uA 

Central and peripheral typing before: 1300uA
Central and peripheral typing after: 1100uA

PR #91, opened 2020-08-15
Add bluemicro support

Na-Cly:
Added support for BlueMicro52840_V1


app/boards/arm/bluemicro52840_v1/Kconfig.defconfig
```diff
@@ -0,0 +1,30 @@
+# Electronut Labs Papyr board configuration
```
Nicell, line 1: This seems out of place... I see it in the nice!nano def too. Oops?

app/boards/arm/bluemicro52840_v1/bluemicro52840_v1.dts
```diff
@@ -0,0 +1,103 @@
...
+	compatible = "nordic,nrf-uart";
+	current-speed = <115200>;
+	tx-pin = <6>;
+	rx-pin = <8>;
+};
+
+&i2c0 {
+	compatible = "nordic,nrf-twi";
+	sda-pin = <25>;
+	scl-pin = <26>;
```
Nicell, line 52: I think this should be 15 and 17?

Nicell:
#90 should probably be done for this board as well

app/boards/arm/bluemicro52840_v1/arduino_pro_micro_pins.dtsi
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2020 Pete Johanson
```
petejohanson, line 2: This should be you!

app/boards/arm/bluemicro52840_v1/bluemicro52840_v1.dts
```diff
@@ -0,0 +1,103 @@
...
+	rx-pin = <8>;
+};
+
+&i2c0 {
+	compatible = "nordic,nrf-twi";
+	sda-pin = <15>;
+	scl-pin = <17>;
+};
+
+&uart0 {
```
petejohanson, line 55: You're overriding `uart0` *twice* in here, this one, and the one above. Which one is the right one? (if you inherited this from the nano board, we likely need to fix that too)

app/boards/arm/bluemicro52840_v1/board.cmake
```diff
@@ -0,0 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0
```
petejohanson, line 1: This license should be fixed, `MIT`.

docs/static/setup.sh
```diff
@@ -11,7 +11,7 @@ title="ZMK Config Setup:"
 # TODO: Check for user.name and user.email git configs being set
 
 prompt="Pick an MCU board:"
-options=("nice!nano" "QMK Proton-C")
+options=("nice!nano" "QMK Proton-C" "bluemicro52840_v1")
```
petejohanson, line 14: This should be the "friendly name", e.g.:

```suggestion
options=("nice!nano" "QMK Proton-C" "BlueMicro52840 (v1)")
```

Also, you need to add a selection in the `select` block below that will set `board="bluemicro52840_v1"`.

docs/static/setup.sh
```diff
@@ -37,7 +37,7 @@ echo ""
 echo "Keyboard Shield Selection:"
 
 prompt="Pick an keyboard:"
-options=("Kyria" "Lily58" "Corne")
+options=("Kyria" "Lily58" "Corne" "splitreus62")
```
petejohanson, line 40: Please add this in the other PR for this shield, thanks!

Na-Cly: Will do.

app/boards/arm/bluemicro52840_v1/bluemicro52840_v1.dts
Na-Cly: Yeah it's directly from the nice_nano.

app/boards/arm/bluemicro52840_v1/Kconfig.defconfig
petejohanson: Yeah, this should be fixed, and we should fix in the nano board too.

app/boards/arm/bluemicro52840_v1/bluemicro52840_v1.dts
petejohanson: Ok, then removed the first `uart0` node that is above this one please.

docs/static/setup.sh
```diff
@@ -11,7 +11,7 @@ title="ZMK Config Setup:"
 # TODO: Check for user.name and user.email git configs being set
 
 prompt="Pick an MCU board:"
-options=("nice!nano" "QMK Proton-C")
+options=("nice!nano" "QMK Proton-C" "BlueMicro52840 (v1)")
```
petejohanson, line 14: Still missing the extra section in the `select` block right below this.

Na-Cly: Added.

app/boards/arm/bluemicro52840_v1/bluemicro52840_v1.dts
petejohanson: Removed.

Issue #84, opened 2020-08-11
USB HID events sometimes dropped, missing key releases, etc.


petejohanson:
The Zephyr USB stack doesn't implement any queue-ing of endpoint writes, so if ZMK attempts to send any HID updates while there's an active write occurring, it may fail. There is a small "retry" tried, but no real queue that is saved while waiting for the write to complete.

Implementing this will help be sure no USB updates are dropped.

petejohanson:
Fixed by #93, thanks @careyk007 !

PR #93, opened 2020-08-16
Wait for USB HID ready before sending reports

careyk007:
In Zephyr's hid-cdc example, they use a semaphore to signal when the host is ready for the next HID report. I haven't been able to test this with ZMK yet (still waiting for some hardware), but I had noticed similar behavior to #84 on my custom keyboard code and this fixed my issues. 

The downside to this approach is that execution will halt until the host signals that it is ready for the next HID report. Some potential solutions would be to spawn a worker thread that will wait for the host before sending the HID report or to create a queue for the HID reports and a dedicated thread that sends queue'd reports as the host becomes ready for them.


Nicell:
I just tested this on my nice60, and it sure does work! Nothing logs as failed to send over USB, and my WPM on a typing test actually improved almost 20WPM, so real world it's working as well. 

Thanks so much for this! I don't have any personal opinions on the worker queue system... It might be nice in the case that we get a failed USB send for another reason that USB being busy and we want to make sure it still goes through.

Issue #33, opened 2020-06-24
Glossary / Terminology


innovaker:
We're continually trying to define (+remember!) terms for concepts commonly used in ZMK related discussions.  Moreover, reuse of terms from existing firmware or software has led to some conceptual misunderstandings.  This will only compound itself when more contributors and developers join the party.

Let's collate, coalesce and document the language we'd like to use for ZMK.  Using pictures where necessary.

Please use this issue for collation and debate of each term.

PR #95, opened 2020-08-17
Increase comprehension of the docs

BrainWart:
Still working on the changes. The Grammarly score has gone from 57 to 63 if that means anything.

### DIFF
--- a/docs/docs/feature/keymaps.md
+++ b/docs/docs/feature/keymaps.md
@@ -4,10 +4,9 @@ title: Keymaps & Behaviors
 sidebar_label: Keymaps
 ---
 
-ZMK uses a declarative approach to keymaps, instead of using C code for all keymap configuration.
-Right now, ZMK uses the devicetree syntax to declare those keymaps; future work is envisioned for
-supporting dynamic loading of declarative keymaps, e.g., over USB Mass Storage or via a custom BLE
-service.
+ZMK uses a declarative approach to keymaps, instead of using C code for all keymap configuration. Right now,
+ZMK uses the devicetree syntax to declare those keymaps. In the future, we hope to support loading keymaps
+dynamically over USB Mass Storage, or a custom BLE service.
 
 :::note
 For advanced users looking to implement custom behaviors for their keymaps, this will be possible
@@ -23,13 +22,12 @@ their user config directory, and customize the keymap to their liking.
 
 ## Behaviors
 
-ZMK implements the concept of "behaviors", which can be bound to a specific key position, sensor (encoder),
-or layer, to perform particular actions when events occur for that binding (e.g. when a specific key position
-is pressed or released, or an encoder triggers a rotation event).
+ZMK implements the concept of "behaviors". Each can be bound to a specific key position, sensor (encoder), or
+layer. The "behavior" will perform particular actions when events occur for that binding.
 
-For example, the simplest behavior in ZMK is the "key press" behavior, which responds to a key position
-(a certain spot on the keyboard), and when that position is pressed, send a keycode to the host, and
-when the key position is released, updates the host notifying of the keycode being released.
+For example, the simplest behavior in ZMK is the "key press" behavior. It responds to one key on the keyboard.
+When pressed, it will send a keycode to the [host](/docs/terms#host), and when the key is
+released, it updates the host notifying the release of the keycode.
 
 For the full set of possible behaviors, start at the [Key Press](/docs/behavior/key-press) behavior.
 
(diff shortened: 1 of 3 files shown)


docs/docs/feature/keymaps.md
```diff
@@ -46,11 +44,11 @@ in ZMK contains a set of bindings that bind a certain behavior to a certain key
...
-When a key location is pressed/released, the stack of all active layers from "top to bottom" is used,
-and the event is sent to the behavior bound at that position in each layer, for it to perform whatever
-actions it wants to in reaction to the event. Those behaviors can choose to "handle" the event, and stop
-it from being passed to any lower layers, or may choose to "pass it along", and let the next layer
-in the stack _also_ get the event.
+When a key location is pressed or released, the stack of all active [layers](/docs/terms#layer) from "top to
+bottom" is used. The event is sent to the [behavior](/docs/terms#behavior) bound at that position in each
+layer. Those [behaviors](/docs/terms#behavior) can "handle" the event, stop it from being passed to any lower
+[layers](/docs/terms#layer), or choose to "pass it along" and let the [layer](/docs/terms#layer) below handle
+the event.
```
CrossR, line 51:
```suggestion
layer. These [behaviors](/docs/terms#behavior) can "handle" the event, stopping it from being passed to any lower
[layers](/docs/terms#layer), or choose to "pass it along" and let the [layer](/docs/terms#layer) below handle
the event.
```

In its current form, I think it can be read as three options:

 - Handle
 - Stop
 - Pass

Rather than "Handle and stop" or "Pass". Swapping to "stopping" or some other reword here would help that.

docs/docs/feature/keymaps.md
```diff
@@ -65,7 +63,7 @@ be used for that key position.
 
 In this case, the `A` is actually a define for the raw HID keycode, to make keymaps easier to read and write.
 
-For example of a binding that uses two parameters, you can see how "mod-tap" (`mt`) is bound:
+For an example of a binding that uses two parameters, you can see how "mod-tap" (`mt`) is bound:
```
petejohanson, line 66: Can we make this link to the newly added mod tap docs?

docs/docs/terms.md
```diff
@@ -0,0 +1,34 @@
+---
+id: terms
+title: Glossary of Terms
```
petejohanson, line 3: This was so badly needed! Thanks!

docs/docs/terms.md
```diff
@@ -0,0 +1,34 @@
...
+sidebar_label: Terms
+---
+
+### Host
+
+The device that will receive the key strokes from the keyboard
+
+### Central
```
petejohanson, line 11: Should these both be explicit about the context of a split keyboard?

docs/docs/feature/keymaps.md
```diff
@@ -16,20 +15,19 @@ in the future by allowing user configs to add to the CMake application built by
...
-ZMK implements the concept of "behaviors", which can be bound to a certain key position, sensor (encoder),
-or layer, to perform certain actions when events occur for that binding (e.g. when a certain key position
-is pressed or released, or an encoder triggers a rotation event).
+ZMK implements the concept of "behaviors". Each can be bound to a specific key position, sensor (encoder), or
+layer. The "behavior" will perform particular actions when events occur for that binding.
 
-For example, the simplest behavior in ZMK is the "key press" behavior, which responds to key position
-(a certain spot on the keyboard), and when that position is pressed, send a keycode to the host, and
-when the key position is released, updates the host to notify of the keycode being released.
+For example, the simplest behavior in ZMK is the "key press" behavior. It responds to one key on the keyboard.
```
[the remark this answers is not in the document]
BrainWart: I had my little brother read it over. He said that he preferred the 'red one.' I could add them to the terms page for extra measure.

docs/docs/terms.md
BrainWart: I can do that. I was hoping to make it clear that the central is the main part of the keyboard so it could be used throughout the docs as the.. well central.

petejohanson: So.. the problem is, for non splits.... It's primary purpose is to be a peripheral to the host. Hmmm....

BrainWart:
Any ideas on how to define events?

docs/docs/terms.md
BrainWart: How can we make it more clear that the central is the host's peripheral and peripheral is the central's?

petejohanson: Perhaps the term "host peripheral" has meaning here? It's a peripheral to a host, both for the non-split and split scenario. *And* in the split case that same side is also a "split central"?

docs/docs/feature/keymaps.md
BrainWart: Resolved in brainwart/zmk@7c2d2c61e32d4bb0b4896862c57a36214aaa390b

docs/docs/terms.md
BrainWart: something like I've got now?

innovaker:
Sorry I didn't see this earlier @BrainWart.

The following is from a very early draft that Pete, Nicell and I talked about.   Please feel free to cherry pick anything useful.

**Node**
Any keyboard/pad (half), numpad, pedals, etc. connected together (wired or wirelessly) as one device.

**Net**
The internal network connecting all the nodes.

**Hub**
The center node of the network (which does the processing).

**Star**
The shape of the net with the hub at the center.  Every other node connects to the hub.

**Peripheral**
Any node that isn’t the hub, i.e. at the edges.

**Matrix Position**
The position in terms of row and column to which the key is physically wired.

**Key Position or K position**
The logical index for the physical position of the key.  Referred to as K1, K2, etc.

**(K, L) or KL**
The key position within a specific layer.  It has two coordinates: the key position K and the layer number L.

**Binding**
The association between a KL and whatever trigger, action, behavior it calls.

innovaker:
Is this ongoing btw?  Perhaps we could split into multiple PRs to get higher priority bits in faster before changes become stale?

innovaker:
This PR currently includes the scope of #33.

BrainWart:
I haven't had time to mess with this in a while. It's getting pretty stale. If I get the time and gumption to work on this again, I'll start a new PR.

PR #96, opened 2020-08-18
Fix for locking when USB sends fail (e.g. on BLE only)

petejohanson:


app/src/usb_hid.c
```diff
@@ -35,7 +35,13 @@ int zmk_usb_hid_send_report(const u8_t *report, size_t len)
...
 
 	k_sem_take(&hid_sem, K_FOREVER);
-	return hid_int_ep_write(hid_dev, report, len, NULL);
+	int err = hid_int_ep_write(hid_dev, report, len, NULL);
+
+	if (err) {
+		k_sem_give(&hid_sem);
+	}
+
+	return err;
```
careyk007, line 44: I had a bad feeling about this shortly after the PR was merged. I think changing `K_FOREVER` to a sensible value would help prevent any future issues. I'm not too familiar with the USB protocol, so I have no idea what a sensible value would be though, but I would guess ~30ms would be a good starting point.

petejohanson: @careyk007 No worries. It happens.

I took your suggestion and removed the use of `K_FOREVER`, and additionally, added code to *first* check the usb status we're already tracking before even *trying* to send the value to the EP.

PR #98, opened 2020-08-18
Feature: Layer toggle

BrainWart:
Usage:
```
&tog LOWER
```
\
\
Should I include how to use it to raise/lower in the documentation?


app/dts/behaviors/toggle_layer.dtsi
```diff
@@ -0,0 +1,9 @@
+/ {
```
petejohanson, line 1: We've not been consistent here, but please add the copyright + license notice to the header here.

app/src/behaviors/behavior_toggle_layer.c
```diff
@@ -0,0 +1,54 @@
...
+{
+	return 0;
+};
+
+
+static int tog_keymap_binding_pressed(struct device *dev, u32_t position, u32_t layer, u32_t _)
+{
+  LOG_DBG("position %d layer %d current %d", position, layer, zmk_keymap_layer_active(layer));
+
+  if (zmk_keymap_layer_active(layer))
```
petejohanson, line 30: Instead of putting this logic *here*, I would rather we create a nicer API for the keymap, this behavior can call into, e.g. make this code just do:

```
return zmk_keymap_layer_toggle(layer);
```

And then move this business logic into `keymap.c`.

app/src/behaviors/behavior_toggle_layer.c
```diff
@@ -0,0 +1,54 @@
...
+    return zmk_keymap_layer_deactivate(layer);
+
+  return zmk_keymap_layer_activate(layer);
+}
+
+static int tog_keymap_binding_released(struct device *dev, u32_t position, u32_t layer, u32_t _)
```
petejohanson, line 36: This is necessary for now, but I think we likely need to change it so behaviors aren't required to provide both callbacks, and the keymap logic will do the right now.

petejohanson:
> 
> Should I include how to use it to raise/lower in the documentation?

@BrainWart It might be nice to give a "Usage Ideas" or "Normal Usage" section after the example, yeah. That'd be nice.

app/src/keymap.c
```diff
@@ -86,6 +91,14 @@ int zmk_keymap_layer_deactivate(u8_t layer)
 	SET_LAYER_STATE(layer, false);
 };
 
+int zmk_keymap_layer_toggle(u8_t layer)
+{
+	if (zmk_keymap_layer_active(layer))
```
petejohanson, line 96: Please add `{` brackets for this check. We'll use them even for simple checks like this. Thanks!

PR #109, opened 2020-08-20
Add RGB underglow state Kconfig and settings

Nicell:
This adds support for both default RGB state settings for the first time the keyboard is started as well as persisting state changes between resets and flashes of new firmware versions.

This also fixes underglow not working by default due to the split WS2812 driver having SPI and GPIO implementations.


app/src/rgb_underglow.c
```diff
@@ -187,20 +215,35 @@ static int zmk_rgb_underglow_init(struct device *_arg)
...
+        saturation: CONFIG_ZMK_RGB_UNDERGLOW_SAT_START,
+        brightness: CONFIG_ZMK_RGB_UNDERGLOW_BRT_START,
+        animation_speed: CONFIG_ZMK_RGB_UNDERGLOW_SPD_START,
+        current_effect: CONFIG_ZMK_RGB_UNDERGLOW_EFF_START,
         animation_step: 0,
+#ifdef CONFIG_ZMK_RGB_UNDERGLOW_ON_START
         on: true
+#else
+        on: false
+#endif
```
Nicell, line 228: Is there a better way to do this? It feels pretty verbose just to check if something is defined.

app/src/rgb_underglow.c
```diff
@@ -187,20 +215,35 @@ static int zmk_rgb_underglow_init(struct device *_arg)
...
     };
 
-    k_timer_start(&underglow_tick, K_NO_WAIT, K_MSEC(50));
+    settings_subsys_init();
```
Nicell, line 231: I'm thinking this should probably go in a more base system initialized file file rather than this RGB file. Suggestions?

app/src/rgb_underglow.c
```diff
@@ -187,20 +215,35 @@ static int zmk_rgb_underglow_init(struct device *_arg)
...
+    settings_register(&rgb_conf);
+    settings_load();
```
Nicell, line 233: Somewhat the same thing with the `subsys_init`, this should probably be in a more generic file that waits for other pieces to initialize (and register their configs), and then one `settings_load` should be called at once.

app/src/settings.c
```diff
@@ -0,0 +1,13 @@
+#include <device.h>
+#include <init.h>
+#include <kernel.h>
+#include <settings/settings.h>
+
+static int zmk_settings_init(struct device *_arg)
+{
+  return settings_load();
```
petejohanson, line 8: Thoughts on timing on this compared to other init functions?

petejohanson (approved):

PR #126, opened 2020-08-26
Fix keys not releasing when layers change

### LINKED ISSUE #125: Incorrect keycode released when layer is changed.
The incorrect keycode is released when the layer is changed *after* a key is pressed. Having a transparent layer above the key will mask the issue as it falls back to the correct layer.

https://discordapp.com/channels/719497620560543766/719544230497878116/747875909016617002

BrainWart:
This fixes #125 for me. The only shortfall I can see is that it doesn't drop through the layers any more for key release. If a key is transparent, it can only expect to receive a keypress event. I can't think of any uses for transparent keys at the moment.

To test the broken behavior:
key down A -> momentary/toggle a layer -> key up A
A will now be stuck pressed until pressed and released again.

Fixed:
key down A -> momentary/toggle a layer -> key up A
A is correctly released despite the layer change

### DIFF
--- a/app/src/keymap.c
+++ b/app/src/keymap.c
@@ -51,10 +51,10 @@ static u8_t zmk_keymap_layer_default = 0;
 
 // State
 
-// When a behavior handles a key position "down" event, we record that layer
+// When a behavior handles a key position "down" event, we record the layer state
 // here so that even if that layer is deactivated before the "up", event, we
 // still send the release event to the behavior in that layer also.
-static u8_t zmk_keymap_active_behavior_layer[ZMK_KEYMAP_LEN];
+static u32_t zmk_keymap_active_behavior_layer[ZMK_KEYMAP_LEN];
 
 static struct zmk_behavior_binding zmk_keymap[ZMK_KEYMAP_LAYERS_LEN][ZMK_KEYMAP_LEN] = {
 	DT_INST_FOREACH_CHILD(0, TRANSFORMED_LAYER)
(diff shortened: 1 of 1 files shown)


app/src/keymap.c
```diff
@@ -98,50 +97,62 @@ int zmk_keymap_layer_toggle(u8_t layer)
 		return zmk_keymap_layer_deactivate(layer);
 	}
 
-  	return zmk_keymap_layer_activate(layer);
+	return zmk_keymap_layer_activate(layer);
 };
 
-bool is_active_position(u32_t position, u8_t layer)
+bool is_active_position(u32_t position, u8_t current_layer, u32_t layer_state)
```
petejohanson, line 103: Seems like you can remove the `position` parameter here, since it's not used any more.

app/src/keymap.c
```diff
@@ -191,11 +208,14 @@ int zmk_keymap_sensor_triggered(u8_t sensor_number, struct device *sensor)
...
 int keymap_listener(const struct zmk_event_header *eh)
 {
-	if (is_position_state_changed(eh)) {
+	if (is_position_state_changed(eh))
+	{
 		const struct position_state_changed *ev = cast_position_state_changed(eh);
 		return zmk_keymap_position_state_changed(ev->position, ev->state);
 #if ZMK_KEYMAP_HAS_SENSORS
-	} else if (is_sensor_event(eh)) {
+	}
```
petejohanson, line 216: I am strongly against this for conditionals, and mixed feelings about this for function definitions. Please revert these style related changes until we can actually fully settle on something. Thanks!

petejohanson: Reverting these changes will also make it easier to review just the fixes found here.

PR #128, opened 2020-08-27
Rip out zephyr battery example and hack into ZMK

bmcgavin:
You also need CONFIG_ZMK_BATTERY=y
CONFIG_ADC=y 
CONFIG_ADC_NRFX_SAADC=y
 in your .config and you need to turn on adc.status in zephyr's nrf52840.dtsi


bmcgavin:
Nope

```
[00:00:00.006,530] <err> usb_nrfx: Endpoint 0x82 is not enabled
[00:00:00.006,530] <err> usb_transfer: Transfer error -22, ep 0x82
```

PR #130, opened 2020-08-28
[Keyboard] Add QAZ Shield

tominabox1:
Add QAZ shield


app/boards/shields/qaz/qaz.conf
```diff
@@ -0,0 +1,26 @@
...
+CONFIG_SERIAL=y
+CONFIG_CONSOLE=y
+CONFIG_UART_INTERRUPT_DRIVEN=y
+CONFIG_UART_LINE_CTRL=y
+
+# Enable USB UART, and set the console device
+CONFIG_UART_CONSOLE=y
+CONFIG_USB_UART_CONSOLE=y
+CONFIG_UART_CONSOLE_ON_DEV_NAME="CDC_ACM_0"
+CONFIG_USB_UART_DTR_WAIT=n
```
petejohanson, line 26: We have *not* been turning on logging by default in these files, since it is a theoretically security concern, since it logs keystroke information. These should be removed for the version of this for in-tree use.

app/boards/shields/qaz/qaz.keymap
```diff
@@ -0,0 +1,75 @@
...
+	behaviors {
+		rst: behavior_tap_hold1 {
+			compatible = "zmk,behavior-tap-hold";
+			label = "Bootloader Tap - Reset Hold";
+			#binding-cells = <0>;
+			hold_ms = <200>;
+			bindings = <&bootloader>, <&reset>, <&kp BKSP>;
+        };
+
+        num_spc: behavior_tap_hold2 {
```
petejohanson, line 19: Tap hold is not in `main`, so this should not be in here. Once we merge that, this can be restored, but for now this should be in your personal user config keymap for this.

app/boards/shields/qaz/qaz.keymap
```diff
@@ -0,0 +1,75 @@
+#include <behaviors.dtsi>
```
petejohanson, line 1: Please add a correct copyright + license header here. Thanks!

app/boards/shields/qaz/qaz.overlay
```diff
@@ -0,0 +1,48 @@
+#include <dt-bindings/zmk/matrix-transform.h>
+
+/ {
+    chosen {
+        zmk,kscan = &kscan0;
+        zmk,matrix_transform = &default_transform;
+        zmk,underglow = &led_strip;
```
petejohanson, line 7: This should be added in the nice_nano overlay where the led_strip is defined. This will ensure this shield builds properly with other controllers like the BlueMicro840.

app/boards/shields/qaz/qaz.overlay
```diff
@@ -0,0 +1,48 @@
+#include <dt-bindings/zmk/matrix-transform.h>
```
petejohanson, line 1: Please add a license & copyright header here as well. Thanks!

app/boards/shields/qaz/Kconfig.shield
```diff
@@ -0,0 +1,2 @@
+config SHIELD_MY_BOARD
```
petejohanson, line 1: Please add a copyright/license header. Thanks!

app/boards/shields/qaz/Kconfig.defconfig
```diff
@@ -0,0 +1,6 @@
+if SHIELD_QAZ
```
petejohanson, line 1: Please add a copyright/license header. Thanks!

app/boards/shields/qaz/qaz.conf
tominabox1: oops forgot I had that in there

tominabox1:
Updated as requested. I removed RGB support. I have not been able to get it to work on my board as I believe we are using one of the low freq pins (P1.15) so may not be possible.

app/boards/shields/qaz/qaz.keymap
petejohanson: Just this one file still missing a license header.

PR #131, opened 2020-08-28
Native posix integration testing

BrainWart:
Thank you, pete, for getting me started with this. I moved the files to a more _robust_ layout, created tests for a handful of the behaviors, modified the run-test.sh script to allow running all commands and for it to show a brief summary of the successful builds.

Not tested at the moment:
- RGB behaviors
- reset and bootloader behaviors
- mod-tap

I would like to have `run-test.sh` work with a set of tests. I think it would be nice to execute a set of tests by calling run-test.sh with the folder of tests or different selections.


app/tests/Creating Tests.md
```diff
@@ -0,0 +1,23 @@
+# Basics
```
petejohanson, line 1: Any reason this shouldn't be in the "Development" section of the website docs?

app/run-test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
petejohanson, line 1: The little shell script that could! I love the improvements here. Thanks!

Can you please add a copyright header here w/ the both of us?

app/tests/Creating Tests.md
BrainWart: Honest answer? Because I'm not sure if people are reading the docs. I figured it would be right in their face here

BrainWart: Moved and updated to link to other docs

BrainWart:
Yeah. Those tests fail. I'm trying to get some PRs that'll make those pass correctly. None should pass now that it's fixed in main. This branch doesn't have the fix though

PR #133, opened 2020-08-28
feat(bluetooth): Proper basic bond management, new `bt` behavior for resetting bond to host.

petejohanson:
* Add `bt` behavior that can be used to with `BT_CLEAR_BONDS`, `BT_PROF_NEXT`, `BT_PROF_PREV`, and `BT_PROF_SEL 2`.
* Clearing bonds ignores the peripheral bond, for full brute force reset, you can:
  * Use the unpair combo, or
  * Build your firmware with the new `ZMK_BLE_CLEAR_BONDS_ON_START` value (e.g. `CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START=y` in your conf), and then flash with that firmware. Make sure to *revert* this once done.


app/src/behaviors/behavior_bt.c
```diff
@@ -0,0 +1,62 @@
...
+    case BT_IDENT_CLR_CMD:
+        return zmk_ble_identity_clear();
+#if CONFIG_BT_ID_MAX != 1
+    case BT_IDENT_NEXT_CMD:
+        return zmk_ble_identity_next();
+    case BT_IDENT_PREV_CMD:
+        return zmk_ble_identity_prev();
+    case BT_IDENT_SEL_CMD:
+        return zmk_ble_identity_select(arg);
+#endif /* BT_ID_MAX != 1 */
```
megamind4089, line 36: default case with logging cmd might be useful while debugging

app/src/ble.c
```diff
@@ -28,16 +37,249 @@ static struct bt_conn *auth_passkey_entry_conn;
...
+    return 0;
+};
+
+int zmk_ble_identity_select(u8_t index)
+{
+    LOG_DBG("index %d", index);
+    if (index >= IDENTITY_COUNT) {
+        return -EINVAL;
+    }
+
```
megamind4089, line 181: Could add a check for ```index == active_identity``` and return

app/Kconfig
```diff
@@ -40,15 +40,19 @@ menuconfig ZMK_BLE
...
 config ZMK_BLE_INIT_PRIORITY
 	int "Init Priority"
 	default 50
 
+config SETTINGS
+	default y
+
+config BT_SETTINGS
+	default y
+
```
megamind4089, line 55: I did not understand the use of exposing of above two configs clearly and trying to understand.
But enabling the following config results in linker error:
```
diff --git a/app/Kconfig b/app/Kconfig
index b60db05..f12a57d 100644
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -48,7 +48,7 @@ config ZMK_BLE_INIT_PRIORITY
        default 50
 
 config SETTINGS
-       default y
+       default n
 
 config BT_SETTINGS
        default y

```

Also from the zephyr Kconfig, looks like SETTINGS will be enabled if we enable BT_SETTINGS unless we disable it manually as per the above Kconfig

app/src/behaviors/behavior_bt.c
petejohanson: Added as part of the refactor. Good suggestion.

app/Kconfig
petejohanson: Indeed. Removed the extra `SETTINGS` line for now.

app/src/ble.c
```diff
@@ -28,16 +37,162 @@ static struct bt_conn *auth_passkey_entry_conn;
...
+    }
+#endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL) */
+
+    LOG_DBG("Unpairing %s", log_strdup(addr));
+    bt_unpair(BT_ID_DEFAULT, &info->addr);
+}
+
+int zmk_ble_clear_bonds()
+{
+    LOG_DBG("");
```
megamind4089, line 148: :roll_eyes:

petejohanson:
Testing notes:

1. Add `#include <dt-bindings/zmk/bt.h>` to the top of your keymap.
1. Add a few new bindings to your keymap, in particular `&bt BT_CLEAR_BONDS` for clearing the bonds for the currently selected profile, and `&bt BT_PROF_NEXT` to select the next open profile.
1. Create a build for your left + right with `CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START=y` in it, and flash to each half, making sure the other is completely off/unplugged before and after until you're ready.
1. Flash again with that Kconfig setting removed, so bonds persist properly.
1. Test turning on each half in various combinations, and ensuring they pair properly.
1. Test pairing to your host, then cycling to the next profile, and pairing with another host, then cycle back to first profile and confirm you can output to the first host, etc.

Note:

* We do fire events for profile changes, will be used to build out display support, LED flashes, etc.

app/src/ble.c
```diff
@@ -137,7 +359,52 @@ static void auth_cancel(struct bt_conn *conn)
...
 }
 
+#if !IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_ROLE_PERIPHERAL)
+static enum bt_security_err auth_pairing_accept(struct bt_conn *conn, const struct bt_conn_pairing_feat *const feat)
+{
+    struct bt_conn_info info;
+    bt_conn_get_info(conn, &info);
+
+    LOG_DBG("role %d, open? %s", info.role, active_profile_is_open() ? "yes" : "no");
+    if (info.role != BT_CONN_ROLE_SLAVE && !active_profile_is_open()) {
```
Nicell, line 369:
```suggestion
    if (info.role == BT_CONN_ROLE_SLAVE && !active_profile_is_open()) {
```
As discussed in Discord. This should check whether this is a connection to a host. In that case the central side has the role `BT_CONN_ROLE_SLAVE`, so it should be `==`. This is confirmed working on split and non-split on my end.

petejohanson: Fixed.

PR #134, opened 2020-08-29
Add Lily58 Left Encoder Support

chenkevinh:
Also changed the BT unpair keys to better match what's already established for kyria and corne (upper and lower outer keys).


app/boards/shields/lily58/lily58_left.conf
```diff
@@ -1,2 +1,6 @@
+# Uncomment to enable encoder
```
petejohanson, line 1: This is good to have if you're building from a checkout, for users that use the "ZMK Config" user repository approach, they need this in the unified `lily58.conf` file. I'm fine leaving this here, but can you also add it w/ the comment to that file as well?

Thanks!

chenkevinh: Lines have been added to the unified conf file.

PR #136, opened 2020-08-30
feat(test): Add GH Action for tests.

petejohanson:
@BrainWart - Care to give a look?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,62 @@
+on: [push, pull_request]
+
+name: Test
+
+jobs:
+  integration_test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


BrainWart:
Looks awesome! I didn't realize you could set up the docker actions like that. Having the logs as an artifact is a really good idea too.  Would love to see how it looks when a test fails though. `./run-test.sh all` should return 123 when any test fails so I expect it to fail correctly

petejohanson:
@BrainWart Yeah, I had various issues w/ builds failing, etc when working on this, and it definitely fails the job as expected. I did just tweak this a tiny bit, so it will always upload the artifacts, so even on failure the logs are there to inspect further as needed.

Issue #117, opened 2020-08-21
Feature: improve mod-tap behavior


okke-formsma:
I've done some investigations on what good mod-tap behavior could be. 

This is a complex matter, as [other firmwares like QMK often have many configuration options for the behavior of the mod-taps](https://github.com/zmkfirmware/zmk/files/5111046/modtap_v2.pdf).

The configuration options exist to make mod-taps work for multiple scenarios, of which homerow mods are one example and things like 'escape-ctrl' on the capslock position another. ZMK needs options to configure the mod-taps, but we should try to keep this as simple as possible.

[The modtap proposal for ZMK](https://github.com/zmkfirmware/zmk/files/5111053/zmk-modtap-proposal.pdf) contains three possible behaviors.

* The MOD-preferred behavior (left in pdf) works great for thing like escape-ctrl, where the user wants to have the key trigger as 'control' whenever another key is pressed shortly after. This behavior gives the user the quickest response after key presses, which makes it a good candidate as a default behavior.

* In the middle, the balanced option. This behavior is probably good for homerow mods, as it will not active the mods on 'key rolls'. The downside of this behavior is that it needs to wait for keyup or tapping_term, making it a little less responsive than the MOD-preferred behavior. 

* The tap-preferred behavior (right in pdf) shows what the most extreme option is when taps are preferred. This behavior would only pick the 'mod' when the 'tapping_term' timer has expired. This may be a preferred option for some people using homerow or layer mods. The downside of this option is that the user ALWAYS has to wait for the timer to expire before a mod kicks in, which may be too slow for mod-tap shift and friends.

Conclusion;

* Let's implement the mod-preferred behavior as default because it's responsive and a good bet for beginning users.
* Let's implement the balanced option so homerow mod users can use ZMK. This is typically a more advanced crowd who can find configuration settings.
* Let's see if we actually need the tap-preferred option before implementing it.

okke-formsma:
Currently, most mod-tap, layer-tap and other hold-tap behaviors work as expected.

One thing that does not work well yet is nested mod-taps. If you press four mod-taps at approximately the same time, you have to wait for 4x the tapping_term before all 4 mods are applied. This should only be one. To fix this behavior, the "original timestamp" needs to be added to the postition_state_changed event. This is implemented in https://github.com/zmkfirmware/zmk/pull/147.

okke-formsma:
Nested mod-taps are implemented for a while now, this ticket is done.

PR #141, opened 2020-08-30
Hold Taps

okke-formsma:
This merge implements the behavior discussed in https://github.com/zmkfirmware/zmk/issues/117.

The tap-hold behavior will soak new position-events until it can make a decision to 'hold' or 'tap', after which all queued keycodes are bubbled. 

We get three "flavors" of mod-taps: "hold-preferred" (default), "balanced" and "tap-preferred". 

* The hold-preferred is default as it gives the fastest response; as soon as another key is pressed, the hold functionality is enabled. (Or if the timer runs out)
* The balanced behavior decides on hold if another key is pressed and released while the tap-hold was active (Or if the timer runs out)
* The tap-preferred behavior only decides on hold if the timer runs out.

Not all behaviors from the PDF are implemented; specifically 3c and 3d (early bubble of 'simple' keycodes) behave like 3a and 3b instead.

This is a preliminary merge, I would like to invite feedback on C coding style and conventions.

todo;
- [ ] rebase on BrainWarts' branch
- [ ] update docs

later
- [x] fix 3c/3d behavior
- [ ] add timing information to all keypress events, so we can handle nested tap-holds properly.


okke-formsma:
The implementation requires a rather verbose behavior definition for every mod-tap. Instead, we could provide the user with a few predefined behaviors with that take 2 bindings, and a few macros to limit the amount of typing:

`MT(A, LCTL)`  instead of `&th-mt &kp A &kp LCTL`

`LT(A, 1)` instead of `&th-mt &kp A &mo 1`

app/include/zmk/event-manager.h
```diff
@@ -78,6 +78,10 @@ struct zmk_event_subscription {
 #define ZMK_EVENT_RELEASE(ev) \
     zmk_event_manager_release((struct zmk_event_header *)ev);
 
+#define ZMK_EVENT_RELEASE_AGAIN(ev) \
+    zmk_event_manager_release_again((struct zmk_event_header *)ev);
```
okke-formsma, line 82: the naming of this is not great, open to a better name.

app/src/behaviors/behavior_tap_hold.c
```diff
@@ -0,0 +1,432 @@
...
+
+#if DT_NODE_EXISTS(DT_DRV_INST(0))
+
+/************************************************************  DATA SETUP */
+#define ZMK_BHV_TAP_HOLD_MAX_HELD 10
+#define ZMK_BHV_TAP_HOLD_MAX_CAPTURED_KC 40
+
+#define TH_TYPE_MOD_PREFERRED 0
+#define TH_TYPE_BALANCED 1
+#define TH_TYPE_TAP_PREFERRED 2
```
okke-formsma, line 32: these 3  are not used anywhere

BrainWart:
> MT(A, LCTL) instead of &th-mt &kp A &kp LCTL
So, formatting it like would be interesting. I think you would have to find a way to store how many binding cells each behavior would take. You may want to go the route of keeping the bindings property to hold the behaviors and use the cells associated with the behavior to store the data going to the other behaviors. Currently, the keymap supports a max of 2 cells that it will pass to the behavior.

```
mt: behavior-mod-tap {
    compatible = "zmk,behavior-tap-hold";
    label = "keypress keypress";
    #binding-cells = <2>;
    pass_params;
    tapping_term_ms = <300>;
    bindings = <&kp &kp>;
};
```

```
keymap {
    compatible = "zmk,keymap";
    label ="Default keymap";

    default_layer {
        bindings = < &mt LCTL A >;
    };
}
```

Passing `param1` to the first keypress behavior and `param2` to the second. Applying similar logic to a layer tap behavior if you wanted.

app/src/behaviors/behavior_tap_hold.c
```diff
@@ -0,0 +1,432 @@
...
+  bool is_decided;
+  bool is_hold;
+  const struct behavior_tap_hold_config *config;
+  struct k_delayed_work work;
+  bool work_is_cancelled;
+};
+
+struct active_tap_hold* undecided_tap_hold = NULL;
+struct active_tap_hold active_tap_holds[ZMK_BHV_TAP_HOLD_MAX_HELD] = {};
+struct position_state_changed* captured_position_events[ZMK_BHV_TAP_HOLD_MAX_CAPTURED_KC] = {};
```
okke-formsma, line 63: it's a little weird to use global variables here, but it's the only option I saw. The listener to key-position-events does not know for which device it was called. We need the globals to select the correct `active_tap_hold` struct.

app/src/behaviors/behavior_tap_hold.c
```diff
@@ -0,0 +1,432 @@
...
+
+/************************************************************  CAPTURED POSITION HELPER FUNCTIONS */
+int capture_position_event(struct position_state_changed* event) {
```
okke-formsma, line 66: should these be static?

app/src/behaviors/behavior_tap_hold.c
```diff
@@ -0,0 +1,432 @@
...
+void decide_tap_hold(struct active_tap_hold * tap_hold, enum decision_moment event)
+{
+  if (tap_hold->is_decided) {
+    return;
+  }
+
+  if(tap_hold != undecided_tap_hold) {
+    LOG_DBG("ERROR found undecided tap hold that is not the active tap hold");
+    return;
+  }
```
okke-formsma, line 241: this shouldn't happen ever, can we do something with asserts that are only built for debugging or something like that?

app/src/behaviors/behavior_tap_hold.c
```diff
@@ -0,0 +1,432 @@
...
+  LOG_DBG("%d capturing %d %s event", undecided_tap_hold->position, ev->position, ev->state?"down":"up");
+
+  capture_position_event(ev);
+  if (ev->state) { 
+    decide_tap_hold(undecided_tap_hold, TH_OTHER_KEY_DOWN);
+  } else {
+    struct position_state_changed* captured_key_down = find_captured_keydown_event(ev->position); 
+    if(captured_key_down == NULL) {
+      // todo: allow key-up events for non-mod keys pressed before the TH was pressed.
+      // see scenario 3c/3d vs 3a/3b.
```
okke-formsma, line 367: I wasn't sure how to implement this, as we're in a key position listener and don't know yet what kind of keypress will result here.

What is needed is something that finds the current behavior for a specific key position, so we can decide if it's a key press for a 'normal' key or something special like a momentary layer or a mod key.

app/src/event_manager.c
```diff
@@ -24,24 +24,24 @@ int zmk_event_manager_handle_from(struct zmk_event_header *event, u8_t start_ind
...
-                        event->last_listener_index = i;
-                        // Listeners are expected to free events they capture
-                        return 0;
-                }
-            }
+            case ZMK_EV_EVENT_CAPTURED:
+                LOG_DBG("Listener captured the event");
+                event->last_listener_index = i;
+                // Listeners are expected to free events they capture
+                return 0;
```
okke-formsma, line 44: this is a refactor that should not be in here. I'll take it out.

app/tests/tap-hold/.~lock.zmk-modtap-proposal.odg#
```diff
@@ -0,0 +1 @@
+,okke,xps,30.08.2020 20:44,file:///home/okke/.config/libreoffice/4;
```
okke-formsma, line 1: oops, gotta get rid of this file too

okke-formsma:
Closing this in favor of new PR for hold-taps.

PR #146, opened 2020-09-01
Hold-Tap implementation

okke-formsma:
This merge implements the behavior discussed in #117. For some discussion see the previous PR #141 

The tap-hold behavior will soak new position-events until it can make a decision to 'hold' or 'tap', after which all queued keycodes are bubbled.

We get three "flavors" of mod-taps: "hold-preferred" (default), "balanced" and "tap-preferred".

* The hold-preferred is default as it gives the fastest response; as soon as another key is pressed, the hold functionality is enabled. (Or if the timer runs out)
* The balanced behavior decides on hold if another key is pressed and released while the tap-hold was active (Or if the timer runs out)
* The tap-preferred behavior only decides on hold if the timer runs out.


okke-formsma:
One test fails because it depends on #67

app/dts/behaviors/mod_tap.dtsi
```diff
@@ -1,9 +1,12 @@
 / {
 	behaviors {
 		mt: behavior_mod_tap {
```
petejohanson, line 3: So... defining mod-tap this way won't allow for setting multiple modifiers to be used for the "hold" behavior here. We can probably get around this by later adding a new `&mods` that just takes a bit-flag set of modifiers, but this is a "regression" to a certain degree.

I have no *personal* need for multiple modifiers for a mod-tap, but I'm not sure if others have a sense of the need for this?

app/dts/behaviors/homerow_tap.dtsi
```diff
@@ -0,0 +1,12 @@
+/ {
+	behaviors {
+		ht: behavior_homerow_mod {
+			compatible = "zmk,behavior-hold-tap";
+			label = "homerow_mod";
```
petejohanson, line 5: Nitpick: The convention in Zephyr seems to be uppercase for the labels used for registering driver instances, so perhaps:

```suggestion
			label = "HOMEROW_MOD";
```

app/run-test-debug.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
```
petejohanson, line 1: Can you help me understand the point of this? Seems like a duplicate of the existing `run-test.sh` script, with minor tweaks.

Thanks.

app/include/zmk/event-manager.h
```diff
@@ -78,6 +78,10 @@ struct zmk_event_subscription {
 #define ZMK_EVENT_RELEASE(ev) \
     zmk_event_manager_release((struct zmk_event_header *)ev);
 
+#define ZMK_EVENT_RELEASE_AGAIN(ev) \
+    zmk_event_manager_release_again((struct zmk_event_header *)ev);
+
 int zmk_event_manager_raise(struct zmk_event_header *event);
 int zmk_event_manager_raise_after(struct zmk_event_header *event, const struct zmk_listener *listener);
 int zmk_event_manager_release(struct zmk_event_header *event);
+int zmk_event_manager_release_again(struct zmk_event_header *event);
```
petejohanson, line 87: I'm not sure I love the name here... Am I understanding the intent here to be to have this event start getting processed again, but starting at the handler that previously swallowed it? Giving it a chance to handle it again, and otherwise, let it flow to the next handler(s)?

If that's the intent, perhaps some thing like "reprocess_from_capture" or "reraise_at_capture". I don't love those names, but I certainly find them to be slightly clearer than "release again" to describe this.

app/tests/hold-tap/balanced/1/events.patterns
```diff
@@ -0,0 +1,4 @@
+s/.*hid_listener_keycode/kp/p
```
petejohanson, line 1: Can we please give each of these directories better names than `1`, `2`, etc? I am assuming these are the different use cases from your diagramming work (Which is awesome!), but perhaps something like `1-tap-on-press-release-while-held` or somesuch.

Thoughts on adding a `README.md` file in the `app/tests/hold-tap/` directory that sets some context on these tests and the scenarios they cover? Can also link to the diagram files you added to guide people to understand these tests?

More generally: ROCK ON FOR ADDING ALL THESE TESTS.

docs/docs/behavior/hold-tap.md
```diff
@@ -0,0 +1,62 @@
...
+### Configuration
+A code example which configures a mod-tap setting that works with homerow mods:
+
+```
+#include <behaviors.dtsi>
+#include <dt-bindings/zmk/keys.h>
+
+/ {
+	behaviors {
+		&hm: homerow_mods {
```
petejohanson, line 34: Extra `&` here.

Also, since we already *register* this elsewhere by way of the `behaviors.dtsi`, I think it may be more confusing than helpful to have this added like this in the docs?

docs/docs/behavior/mod-tap.md
```diff
@@ -20,11 +16,28 @@ The Mod-Tap behavior either acts as a held modifier, or as a tapped keycode.
...
 
 - Reference: `&mt`
-- Parameter #1: The modifiers to be used when activating as a modifier, e.g. `MOD_LSFT`
+- Parameter #1: The keycode to be sent when activating as a modifier, e.g. `LSFT`
 - Parameter #2: The keycode to sent when used as a tap, e.g. `A`, `B`.
 
 Example:
 
 ```
 &mt MOD_LSFT A
```
petejohanson, line 25: This needs updating too.

```suggestion
&mt LSFT A
```

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -0,0 +1,491 @@
...
+	}
+	return -ENOMEM;
+}
+
+static struct position_state_changed *find_captured_keydown_event(u32_t position)
+{
+	struct position_state_changed *last_match = NULL;
+	for (int i = 0; i < ZMK_BHV_HOLD_TAP_MAX_CAPTURED_EVENTS; i++) {
+		const struct zmk_event_header *eh = captured_events[i];
+		if (eh == NULL) {
```
petejohanson, line 85: So... this makes sense, but *assumes* future refactors don't end up with a "sparse" `captured_events` array. Is this a reasonable assumption? If so, fine, just thought I'd ask the question.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -0,0 +1,491 @@
...
+		} else {
+			struct keycode_state_changed *modifier_event = cast_keycode_state_changed(captured_event);
+			LOG_DBG("Releasing mods changed event 0x%02X %s", modifier_event->keycode, (modifier_event->state ? "pressed" : "released"));
+		}
+		ZMK_EVENT_RELEASE_AGAIN(captured_event);
+	}
+}
+
+
+/************************************************************    ACTIVE TAP HOLD HELPER FUNCTIONS */
```
petejohanson, line 151: I don't love this convention for comments for sections.

Convince me?

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -0,0 +1,491 @@
...
+		return;
+	}
+
+	if (hold_tap != undecided_hold_tap) {
+		LOG_DBG("ERROR found undecided tap hold that is not the active tap hold");
+		return;
+	}
+
+	char *flavor = hold_tap->config->flavor;
+	if (strcmp(flavor, "balanced") == 0) {
```
petejohanson, line 260: We shouldn't be doing string comparisons for this. It's inefficient, and Zephyr offers a better option here. Zephyr will assign numeric values for each enum entry, and those values are what we should be storing, and comparing, when making these "which flavor is it" checks.

app/dts/behaviors/mod_tap.dtsi
okke-formsma: Something like `&mods` would be a great idea. I prefer it over requiring a behavior definition for every tap-hold.

I think we should limit the use of behavior definitions for common use cases. A first time user who wants to set up hold-taps should not have to define a behavior for every key, but just call something like `&ht SFT A`. The next step is to change their settings to their preferences by using `&ht { settings = <value> ; }`, and once they want even more configurability (e.g. per-key tapping term) they'll have to write their own behavior. This way, there is an easy way to use this functionality for new users, and a more complex way for power users to configure the keyboard to their liking.

app/run-test-debug.sh
okke-formsma: It outputs all information necessary to check why a test fails, and debug it without having to refer to lots of different files.

I'll take this out of the PR. A better soluttion is a "debug" or "verbose" parameter in the run test.sh script.

okke-formsma: When running tests regularly, it's very helpful to have all relevant output immediately avaialble without having to examine different output files. I agree that this is not an elegant solution I'll take it out and make a "verbose" flag for the run-test script.

app/include/zmk/event-manager.h
okke-formsma: I struggled with the name too, and don't like it either. Maybe "zmk_event_manager_raise_before" which is used similar to "zmk_event_manager_raise_after" but raises the event just before the current handler?

app/tests/hold-tap/balanced/1/events.patterns
okke-formsma: I committed both the PDF and open document format of the files to the tests/hold-tap folder. I'll add a readme and somewhat descriptive names.

app/src/behaviors/behavior_hold_tap.c
okke-formsma: I think it is reasonable. A captured event should not be released or deleted on it's own, all decision about queued events should be made by the listener.

app/src/behaviors/behavior_hold_tap.c
okke-formsma: Not a big fan either, but I noticed I was getting lost in the file. The file is pushing 500 lines which is about the max I'm comfortable with. I'm not sure if/how to cut it up. 

For now I've removed the sections, I'm open to suggestions.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -0,0 +1,491 @@
...
+		hold_tap->is_decided = true;
+		break;
+	case HT_OTHER_KEY_UP:
+		hold_tap->is_hold = 1;
+		hold_tap->is_decided = true;
+		break;
+	case HT_TIMER_EVENT:
+		hold_tap->is_hold = 1;
+		hold_tap->is_decided = true;
+		break;
```
CrossR, line 209: Not sure on other opinions, but I like having stacked `case` statements for things like this:

```
case HT_OTHER_KEY_UP:
case HT_TIMER_EVENT:
   hold_tap->is_hold = 1;
   ...
```

Since its then obvious that both of those cases give an equivalent result, whereas having multiple cases makes me assume they will have differing effects

docs/docs/behavior/hold-tap.md
okke-formsma: I've removed the "homerow mods" default binding instead of removing it from the docs, I think the homerow mods can be done in too many ways to have it make sense to make a default behavior.

app/src/behaviors/behavior_hold_tap.c
okke-formsma: done. i couldn't figure out how to get the string value at runtime, so there's a rather ugly log_dbg statement now.

app/src/behaviors/behavior_hold_tap.c
petejohanson: Let's ponder. Also, lets not hold this PR up on that. Thanks.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -0,0 +1,489 @@
...
+	}
+
+	if (!hold_tap->is_decided) {
+		return;
+	}
+
+	LOG_DBG("%d decided %s (%s event %d)", 
+		hold_tap->position, 
+		hold_tap->is_hold ? "hold" : "tap", 
+		flavor == 0 ? "hold-preferred" : flavor == 1 ? "balanced": "tap-preferred",
```
petejohanson, line 272: Perhaps move to separate small inline function using a `switch` that'll be equally fast, but not so ugly?

PR #147, opened 2020-09-01
Add timestamps to position and behavior events

okke-formsma:
When events are queued and later replayed, timing information is currently not preserved. This makes timings of things like nested hold-taps difficult. This PR adds timestamp information to position and behavior events, so timings of nested hold-taps can be handled properly. 

image

previously, this would return hold for the first HT and tap for the other three. Now, all three hold-taps are enabled quickly after one another.

This makes homerow mods much easier to work with. Without this implementation, if you press ctrl+alt+shift TH mods at the same time, you have to wait for 3x the tapping term for all three mods to register. With this PR, you only wait for the tapping_term once!

### DIFF
--- a/app/include/drivers/behavior.h
+++ b/app/include/drivers/behavior.h
@@ -20,7 +20,7 @@
  */
 
 typedef int (*behavior_keymap_binding_callback_t)(struct device *dev, u32_t position, u32_t param1,
-                                                  u32_t param2);
+                                                  u32_t param2, s64_t timestamp);
 typedef int (*behavior_sensor_keymap_binding_callback_t)(struct device *dev, struct device *sensor,
                                                          u32_t param1, u32_t param2);
 
@@ -43,17 +43,18 @@ __subsystem struct behavior_driver_api {
  * @retval Negative errno code if failure.
  */
 __syscall int behavior_keymap_binding_pressed(struct device *dev, u32_t position, u32_t param1,
-                                              u32_t param2);
+                                              u32_t param2, s64_t timestamp);
 
 static inline int z_impl_behavior_keymap_binding_pressed(struct device *dev, u32_t position,
-                                                         u32_t param1, u32_t param2) {
+                                                         u32_t param1, u32_t param2,
+                                                         s64_t timestamp) {
     const struct behavior_driver_api *api = (const struct behavior_driver_api *)dev->driver_api;
 
     if (api->binding_pressed == NULL) {
         return -ENOTSUP;
     }
 
-    return api->binding_pressed(dev, position, param1, param2);
+    return api->binding_pressed(dev, position, param1, param2, timestamp);
 }
 
 /**
@@ -65,17 +66,18 @@ static inline int z_impl_behavior_keymap_binding_pressed(struct device *dev, u32
  * @retval Negative errno code if failure.
  */
 __syscall int behavior_keymap_binding_released(struct device *dev, u32_t position, u32_t param1,
-                                               u32_t param2);
+                                               u32_t param2, s64_t timestamp);
 
 static inline int z_impl_behavior_keymap_binding_released(struct device *dev, u32_t position,
-                                                          u32_t param1, u32_t param2) {
+                                                          u32_t param1, u32_t param2,
+                                                          s64_t timestamp) {
     const struct behavior_driver_api *api = (const struct behavior_driver_api *)dev->driver_api;
 
     if (api->binding_released == NULL) {
         return -ENOTSUP;
     }
 
-    return api->binding_released(dev, position, param1, param2);
+    return api->binding_released(dev, position, param1, param2, timestamp);
 }
 
 /**
--- a/app/include/drivers/behavior.h
+++ b/app/include/drivers/behavior.h
@@ -10,6 +10,7 @@
 #include <stddef.h>
 #include <device.h>
 #include <zmk/keys.h>
+#include <zmk/behavior.h>
 
 /**
  * @cond INTERNAL_HIDDEN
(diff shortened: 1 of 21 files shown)


okke-formsma:
I found a bug where the timestamp was not sent when an event was received from a peripheral.

app/include/drivers/behavior.h
```diff
@@ -20,7 +20,7 @@
  */
 
 typedef int (*behavior_keymap_binding_callback_t)(struct device *dev, u32_t position, u32_t param1,
-                                                  u32_t param2);
+                                                  u32_t param2, s64_t timestamp);
```
petejohanson, line 23: It seems like we're starting to reach the point of "too much" in terms of parameters. I wonder if perhaps instead of these proliferation of parameters, if we need a `struct zmk_key_position_event_data` struct that contains all the data? Then the event definition itself would be just the header and then the data? I start this approach for some of the layer state events I have half completed.

Thoughts?

petejohanson (commented):
One API food for thought. Thanks for working on this!

app/include/drivers/behavior.h
innovaker: As in ... this?
```
struct zmk_key_position_event_data {
    u32_t position;
    s64_t timestamp;
};
```
or?

petejohanson: Yes, at a minimum. I do wonder if we should figure out a way to "bundle" the two binding parameters too?

innovaker: Parameters as an inner struct or fixed-size array?  What's the cleanest approach to handling one-param and two-param bindings in this context?

petejohanson: I would still expect a fixed two fields for each possible param, that would be `0` and ignored by behaviors that don't use them.

okke-formsma: yes, I had the same thoughts when developing this. Lets pass by value for now to make memory mgmt simple.

okke-formsma:
```
struct zmk_key_position_event_data {
    32_t param1,
    32_t param2,
    u32_t position;
    s64_t timestamp;
};
```

innovaker: Suggest `position` and `timestamp` at the start of the `struct` as those are the least likely to change in the future.

okke-formsma: The overlap with two existing structs is remarkable. Although the first contains a char* instead of a device*, and the second is allocated on the stack and has the wrong lifetime...

```
struct zmk_behavior_binding {
    char *behavior_dev;
    u32_t param1;
    u32_t param2;
};
```
```
struct position_state_changed {
    struct zmk_event_header header;
    u32_t position;
    bool state;
    s64_t timestamp;
};
```

petejohanson: Yeah, so the other issue here is there's the general "event data", and then there's specific "binding info" that should be passed to the behavior we're activating.

I think perhaps we end up with something like:

```
struct zmk_position_state_changed_data {
  u32_t position;
  bool state;
  s64_t timestamp;
}

struct position_state_changed {
  struct zmk_event_header header;
  struct zmk_position_state_changed_data data;
}
```

And then we can pass two things to the behavior, the position state changed data struct, and the binding struct.

Thoughts?

okke-formsma: I'm now implementing the following struct. Not happy with the naming.

```
struct behavior_keymap_binding_data {
    int layer;
    u32_t position;
    s64_t timestamp;
    u32_t param1;
    u32_t param2;
};
```

petejohanson: @okke-formsma Regarding naming `zmk_behavior_binding_event_data` maybe?

okke-formsma: There's also a behavior binding event for sensors which acts slightly different, I'll see if we can use the same data.

petejohanson: I still am not convinced we don't need two structs, both passed to the behavior callbacks, one for the event data itself, separate from the specific behavior being invoked, and the other for the specific binding data that is associated w/ this specific behavior being invoked.

Thoughts?

okke-formsma: After considering all ideas, I've decided to implement the following:

```
struct zmk_behavior_binding {
    char *behavior_dev;
    u32_t param1;
    u32_t param2;
};

struct zmk_behavior_binding_data {
    int layer;
    u32_t position;
    s64_t timestamp;
};
```

reasoning:
* re-uses existing `struct zmk_behavior_binding`
* "runtime" information stored in `struct zmk_behavior_binding_data`
* not using the position event because this way we can extend the data to include things like the layer information.

app/include/drivers/behavior.h
```diff
@@ -19,8 +20,7 @@
  * (Internal use only.)
  */
 
-typedef int (*behavior_keymap_binding_callback_t)(struct device *dev, u32_t position, u32_t param1,
-                                                  u32_t param2);
+typedef int (*behavior_keymap_binding_callback_t)(struct zmk_behavior_binding *binding, struct zmk_behavior_binding_data data);
 typedef int (*behavior_sensor_keymap_binding_callback_t)(struct device *dev, struct device *sensor,
                                                          u32_t param1, u32_t param2);
```
petejohanson, line 25: If we're going to do this, let's refactor this callback to `(struct zmk_behavior_binding *binding, struct device *sensor)` while we're at it.

app/include/zmk/behavior.h
```diff
@@ -10,4 +10,10 @@ struct zmk_behavior_binding {
     char *behavior_dev;
     u32_t param1;
     u32_t param2;
+};
+
+struct zmk_behavior_binding_data {
```
petejohanson, line 15: The naming I'm still not 100% happy with.... Thoughts on:

```suggestion
struct zmk_behavior_binding_event_data {
```

app/include/zmk/events/keycode-state-changed.h
```diff
@@ -14,16 +14,20 @@ struct keycode_state_changed {
     u8_t usage_page;
     u32_t keycode;
     bool state;
+    u32_t position;
```
petejohanson, line 17: These events might be synthesized from other place, not just from a specific keybinding, per-se. Why was this needed?

petejohanson (commented):
Just one or two minor items, this is really close. Thanks!

app/include/zmk/behavior.h
okke-formsma: let's drop the 'data' part?

okke-formsma:
```

struct zmk_behavior_binding_event {
    struct zmk_behavior_binding* binding;
    int layer;
    u32_t position;
    s64_t timestamp;
};
```
this may be an option too...

app/include/zmk/events/keycode-state-changed.h
okke-formsma: we can do without.

okke-formsma:
I've just force-pushed an updated version that includes fixes for all conversations above.

petejohanson (approved):
Thanks!

PR #151, opened 2020-09-02
Added Iris Shield

kurtis-lew:


app/boards/shields/iris/Kconfig.defconfig
```diff
@@ -0,0 +1,14 @@
+
```
petejohanson, line 1: Please add a copyright header here. (I know we're missing some in `main`, trying to avoid even more getting in without them. Thanks.

app/boards/shields/iris/iris.keymap
```diff
@@ -0,0 +1,56 @@
+#include <behaviors.dtsi>
```
petejohanson, line 1: Please add a copyright header here too. Thanks.

Issue #86, opened 2020-08-12
Feature: "Shifted Keycodes",  e.g. assigning `!` to a key press or mod-tap behavior.


petejohanson:
Potential fix/idea:
*  Use upper bits of the first parameter to `key_press`, etc to have the enum flag of the mods included. UGH, but not sure how to avoid this if we want this to also work w/ things like mod-tap, where the other parameter is already used.
* Add defines to `keys.h` that are <base keycode> + <held mods>, e.g. something like `#define BANG (NUM_1 + HELD_MODS(MOD_LSFT))`
* Updated "key code state changed" event to include *mods* expected with that keycode.
* Handling in the endpoint logic to inspect the mods and handle accordingly. Need to be careful w/ handling mod state, so we don't *release* mods early, etc.

BrainWart:
For people who can't wait for a better solution, brainwart/zmk@e09246a8e664c2e873a21cd8d4903f445156c51d

This does not add new keycodes. It just applies two keys one after another.
Usage: `&mkp LSFT NUM_7`

okke-formsma:
I have an implementation proposal for this in https://github.com/okke-formsma/zmk/tree/modifiers / https://github.com/zmkfirmware/zmk/pull/153

It brings the following changes:

* Modifiers are counted. If you press LSFT twice, only when it's released for the second time the bitflag in the HID is zeroed.
* No more differentiation between `MOD_LSFT` and `LSFT`. It's always `LSFT`. You can use  `&lsft A` or `&kp (A | LSFT)` for the same result. A shift key would be `&kp LSFT` (preferred) or `&lsft 0` (bad form but it works)
* This makes mod-taps with multiple mods possible:  `&mt (LSFT | LCTL) A`

The mods are implemented as flags in the third byte of the `keycode`. This overloads the keycodes a bit, but I feel that's warranted in this case.

There has been some discussion about this on discord: https://discordapp.com/channels/719497620560543766/719544230497878116/751160749203652649

okke-formsma:
We had a chat on discord about not adding shifted keycodes to the default codes, as they are locale-dependent. (e.g. shifted keycodes result in a different character being typed depending on Dutch/Belgian/French/German/..., UK or US setting.)

By forcing users to use `&lsft NUM_3` or `&kp (NUM_3 | LSFT)`, we make them aware of what is happening.

innovaker:
Would this syntax (using MACROs) be more robust in terms of future implementation refactoring?
[...]

PR #153, opened 2020-09-02
initial implementation for modifiers

okke-formsma:
Implements https://github.com/zmkfirmware/zmk/issues/86.

`&sft NUM_1` presses shift, then 1, then releases shift, then releases 1.


app/dts/behaviors/modifiers.dtsi
```diff
@@ -0,0 +1,94 @@
+#include <dt-bindings/zmk/keys.h>
+
+/ {
+	behaviors {
+		ctl: behavior_ctl {
+			compatible = "zmk,behavior-key-press";
+			label = "MOD_CTRL";
+			usage_page = <USAGE_KEYPAD>;
+			#binding-cells = <1>;
+            modifiers = <(MOD_LCTL)>;
```
BrainWart, line 10: Used spaces here and a few other lines in the file while the rest of the file uses tabs for indentation.

app/include/dt-bindings/zmk/keys.h
```diff
@@ -146,4 +146,5 @@
 #define MOD_RCTL (1 << 0x04)
 #define MOD_RSFT (1 << 0x05)
 #define MOD_RALT (1 << 0x06)
-#define MOD_RGUI (1 << 0x07)
\ No newline at end of file
+#define MOD_RGUI (1 << 0x07)
+#define MOD_LAST_MODIFIER MOD_RGUI
```
BrainWart, line 150: What was `MOD_LAST_MODIFIER` for? I don't see it used anywhere.

app/src/hid.c
```diff
@@ -14,32 +14,58 @@ static struct zmk_hid_consumer_report consumer_report = {
...
 
-#define _TOGGLE_MOD(mod, state)                      \
-    if (modifier > MOD_RGUI)                         \
-    {                                                \
-        return -EINVAL;                              \
-    }                                                \
-    WRITE_BIT(kp_report.body.modifiers, mod, state); \
-    return 0;
+// Keep track of how often a modifier was pressed.
+// Only release the modifier iff the count is 0.
```
BrainWart, line 18:
```suggestion
// Only release the modifier if the count is 0.
```

BrainWart (commented):
Thanks, Okke. I saw a couple of things and was wondering what `MOD_LAST_MODIFIER` was for. I haven't had a chance to test it yet personally.

app/include/dt-bindings/zmk/keys.h
okke-formsma: I used it before but didn't need it later. removed.

CrossR:
Just to duplicate the comment I made on discord, I'm seeing some stuck modifiers using this PR merged with main. I see either my LCTL or LSFT stick after pressing them, and I don't see this behaviour on main.

My config is in a branch here, and I don't believe I'm doing anything especially odd, just defining the keys for my layout.

https://github.com/CrossR/zmk_config/compare/CrossR/ShiftedKeyCodes

I'm seeing the stuck modifiers when using the normal LCTL and LSFT on my default layer, which are unchanged from my normal keymap.

okke-formsma:
I've fixed the bug CrossR was running into, and updated the test so it won't happen again.

app/dts/behaviors/modifiers.dtsi
```diff
@@ -0,0 +1,94 @@
...
+			#binding-cells = <1>;
+            modifiers = <(LALT)>;
+		};
+
+		lgui: behavior_gui {
+			compatible = "zmk,behavior-key-press";
+			label = "MOD_GUI";
+			usage_page = <USAGE_KEYPAD>;
+			#binding-cells = <1>;
+            modifiers = <(LALT)>;
```
CrossR, line 34: Copy paste error?
```suggestion
            modifiers = <(LGUI)>;
```

app/dts/behaviors/modifiers.dtsi
```diff
@@ -0,0 +1,94 @@
...
+			#binding-cells = <1>;
+            modifiers = <(RCTL)>;
+		};
+
+		rsft: behavior_rsft {
+			compatible = "zmk,behavior-key-press";
+			label = "RSFT";
+			usage_page = <USAGE_KEYPAD>;
+			#binding-cells = <1>;
+            modifiers = <(RALT)>;
```
CrossR, line 51:
```suggestion
            modifiers = <(RSFT)>;
```

app/src/hid.c
```diff
@@ -75,12 +103,18 @@ int zmk_hid_unregister_mods(zmk_mod_flags modifiers)
 
 int zmk_hid_keypad_press(zmk_key code)
 {
-    if (code >= LCTL && code <= RGUI)
-    {
-        return zmk_hid_register_mod(code - LCTL);
+    zmk_mod_flags mods = SELECT_MODS(code);
+    if (mods) {
+        return zmk_hid_register_mods(mods);
```
CrossR, line 108: I could be mis-understanding this code, but this read as "If there is mods, apply them and stop", and would explain the behaviour I'm seeing. The code following this ("If only mods, stop") makes me assume this isn't needed.
```suggestion
        zmk_hid_register_mods(mods);
```

app/src/hid.c
```diff
@@ -95,9 +129,17 @@ int zmk_hid_keypad_press(zmk_key code)
 
 int zmk_hid_keypad_release(zmk_key code)
 {
-    if (code >= LCTL && code <= RGUI)
-    {
-        return zmk_hid_unregister_mod(code - LCTL);
+    zmk_mod_flags mods = SELECT_MODS(code);
+    if (mods) {
+        return zmk_hid_unregister_mods(mods);
```
CrossR, line 134: Same as press comment.

CrossR (commented):
No stuck mods now! Though I am not seeing any mod applied keypresses, though I believe I see why, assuming I'm following the code correctly.

With the change I've mentioned, "it works on my machine" TM at least.

app/include/zmk/events/keycode-state-changed.h
```diff
@@ -7,22 +7,34 @@
...
 
-inline struct keycode_state_changed *create_keycode_state_changed(u8_t usage_page, u32_t keycode,
-                                                                  bool state) {
+static inline struct keycode_state_changed *create_keycode_state_changed(u8_t usage_page, u32_t keycode, bool state) {
     struct keycode_state_changed *ev = new_keycode_state_changed();
     ev->usage_page = usage_page;
-    ev->keycode = keycode;
+    // this is ugly and after merging https://github.com/zmkfirmware/zmk/pull/300
+    // should move into the keycode_state_changed_from_encoded method
+    // discuss: should we keep LCTL and MOD_LCTL even though it causes this ugly code?
```
okke-formsma, line 28: I don't like this code here, but I'm not sure it's better to in hid.c.

petejohanson: I think one good reason to remove this from here is to help keep clearly distinct the modifiers that are being held/applied because they are being asked for in addition to a different keycode, and a modifier that truly is just a modifier being pressed/released.

That *context* I think is important to preserve, because we'll need to when we work to solve the more nuanced handling of modifiers and when we release them because of other key presses, etc.

Thoughts?

okke-formsma: Agreed, that's a neat way to separate the implicit/explicit modifiers.

app/include/dt-bindings/zmk/keys.h
```diff
@@ -144,11 +143,24 @@
...
-#define MOD_LCTL (1 << 0x00)
-#define MOD_LSFT (1 << 0x01)
-#define MOD_LALT (1 << 0x02)
-#define MOD_LGUI (1 << 0x03)
-#define MOD_RCTL (1 << 0x04)
-#define MOD_RSFT (1 << 0x05)
-#define MOD_RALT (1 << 0x06)
-#define MOD_RGUI (1 << 0x07)
\ No newline at end of file
+#define MOD_LCTL 0x01
```
innovaker, line 146: I'd suggest moving the modifiers codes and macros into their own header - perhaps `modifiers.h`?

For now, `modifiers.h` can be included into `keys.h` until the [single entry point](https://discordapp.com/channels/719497620560543766/719565084208398406/769222960892346438) enhancement for keymaps has been implemented.  The includes to `keys.h` elsewhere in the system (such as in `keycode-state-changed.h`) can then be replaced by `modifiers.h` which is a tighter design.

innovaker:
Thanks @okke-formsma!   Just a quick suggestion for now (above), will look closer tomorrow with fresh eyes.

app/include/zmk/hid.h
```diff
@@ -164,8 +164,8 @@ struct zmk_hid_consumer_report {
 
 int zmk_hid_register_mod(zmk_mod modifier);
 int zmk_hid_unregister_mod(zmk_mod modifier);
-int zmk_hid_register_mods(zmk_mod_flags modifiers);
-int zmk_hid_unregister_mods(zmk_mod_flags modifiers);
+int zmk_hid_implicit_modifiers_press(zmk_mod_flags implicit_modifiers);
+int zmk_hid_implicit_modifiers_release();
```
innovaker, line 168: I'm not sure if the modifier implicit/explicit/counting state and related functions should live within `hid.h/c`.

Rationale:
 - it reduces the cohesion of the `hid` module.
 - there's weak coupling to the rest of the `hid` module.
 - it feels like a conceptual layer above `hid`.
 - the `modifiers` system will likely be part of or linked to the *internal state* subsystem(s) that have been discussed recently.
 - `hid.c` itself can probably be refactored into `hid_keyboard.c` and `hid_consumer.c`

Perhaps it's better suited to a dedicated module, say `modifiers.h/c` or `hid_modifiers.h/c`?

petejohanson: I think this is fine for now, and we can refactor if/when we have a plan/need for the state changes.

app/src/hid.c
```diff
@@ -8,29 +8,44 @@
...
+        LOG_DBG("Modifiers set to 0x%02X", kp_report.body.modifiers);                              \
+    }
 
-int zmk_hid_register_mods(zmk_mod_flags modifiers) {
-    kp_report.body.modifiers |= modifiers;
+int zmk_hid_register_mod(zmk_mod modifier) {
+    explicit_modifier_counts[modifier]++;
+    LOG_DBG("Modifier %d count %d", modifier, explicit_modifier_counts[modifier]);
+    WRITE_BIT(explicit_modifiers, modifier, true);
+    SET_MODIFIERS(explicit_modifiers);
```
innovaker, line 33: If we're going for `explicit` and `implicit`, let's be consistent throughout?

app/include/dt-bindings/zmk/keys.h
petejohanson: Fully agreed. I'm going to merge this from @okke-formsma , and then follow up with that small refactor.

petejohanson (approved):
Thanks for this! I've been playing with it the past couple hours, and it seems really solid. Definitely good enough we should get this in, and pound on it a bit, get the HID refactor from @innovaker in with the nice defines for common shifted codes, and call it a day. Thanks again!

Issue #140, opened 2020-08-30
Document encoders usage (`sensor-bindings` in keymaps).


petejohanson:
We need to document the steps necessary to:

* Add encoder(s) to shield.
* Enable encoders in a user keymap.
* Bind encoder sensors to behaviors in a keymap.

innovaker:
@petejohanson, should this be closed?

PR #154, opened 2020-09-03
Encoder documentation

chenkevinh:
Added documentation on configuring encoders and adding encoders to new boards/shields. Issue #140


docs/docs/dev-guide-new-shield.md
```diff
@@ -195,6 +199,81 @@ Further documentation on behaviors and bindings is forthcoming, but a summary of
 - `trans` is the "transparent" behavior, useful to be place in higher layers above `mo` bindings to be sure the key release is handled by the lower layer. No binding arguments are required.
 - `mt` is the "mod-tap" behavior, and takes two binding arguments, the modifier to use if held, and the keycode to send if tapped.
 
+## Adding Features
+
+### Encoders
+
+EC11 encoder support can be added to your board or shield by adding the appropriate lines to your board/shield's configuration (.conf), device tree (.dtsi), and overlay (.overlay) files.
```
petejohanson, line 206: It seems like we should also have a note/section that links to the encoders feature doc from here, so users adding the shield can add the default sensor bindings as well.

Thoughts?

petejohanson (commented):
Just one thought, otherwise this looks good!

docs/docs/dev-guide-new-shield.md
chenkevinh: Yup good call, I'll add it as another tab to update the keymap as well

petejohanson (approved):
Thanks!

PR #156, opened 2020-09-03
Fix Sofle build.

CrossR:
Got this error: https://github.com/CrossR/zmk_config/runs/1068763312 when enabling encoders in my config, and my assumption is that its because they default to `"okay"` and I set them to that in the l/r overlay, it was causing a conflict.

~~Whilst I was fixing this, updated the intro page to move bits around and add macros since I've seen a few people question if ZMK had them yet.~~ Backed out in favour of #159.


petejohanson (approved):
LGTM.

PR #157, opened 2020-09-03
Update iris_right. overlay - fixed column gpios

kurtis-lew:


petejohanson (changes requested):
@kurtis-lew These change for the Iris look great, but it seems you also mixed in some troubleshooting page work. Can we split these out?

Issue #21, opened 2020-06-13
Standardize Keys


innovaker:
- Revisit/refactor current `dt-bindings/zmk/keys.h`
- Include any useful values for USB or HOGP from HID Usage Tables v1.12:
https://www.usb.org/sites/default/files/documents/hut1_12v2.pdf
- Consider values and naming conventions found in existing KB firmware, Linux, Windows, Android, etc.
- Decide upon any conventions for ZMK.
- Open up consultation with the ZMK contributor community before finalizing.

petejohanson:
Updated usage tables version: https://usb.org/sites/default/files/hut1_2.pdf

tominabox1:
Add scope of #158 to this

innovaker:
Let's make a concerted effort to nail this down, including @tominabox1's efforts on #161 and taking into account the effects of #86.

innovaker:
It's RFC time!  Feedback, suggestions and discussions welcomed!  Everything is still on the table at this point.

To facilitate this first pass, we're using Google Sheets as a _living document_.  It contains the proposed names (tags) and descriptions that are used for the `key-press` behavior (`&kp`, `&cp`):

RFC: https://docs.google.com/spreadsheets/d/1s6URDIL7pqcMiUvY8s0esWvLIA1dKNmHqp6jEUtrqTw/

- Feel free to make _focused comments_ (per cell) in the sheet itself, but please use this GitHub issue for more general or _cross-cutting_ discussions.
- Please **_let people know who you are_**, *either* by:
  - signing into Google with an account that has the same visible name as your Discord/GitHub name, OR ...
  - adding your Discord/GitHub name onto each comment (i.e. `-innovaker`).
- The _white cells_ contain the stuff that will be put into ZMK.  This is the focus of the RFC.
- The _grey cells_ contain useful background information from other sources.  Please don't comment on these cells.
- I'll be updating the document as each thread of discussion progresses, so it will change over time.
- Be sure to login to the sheet if you want to receive ongoing notifications to comments.
- Care was taken to ensure consistency, but there's bound to be some oversights.  So ... what can you spot!? :)

**Design Considerations (thus far)**
- There are many ways to slice the "keys cake" and everyone tends to have their own opinions/biases.  Taking this into consideration, I'm suggesting we support a variety of logical aliases for each code - both long and short forms.
- Due to C's namespace limitations and overlaps, the prefix naming conventions _at time of writing this post_ are: 
   - Keyboard: _none_
[...]

PR #161, opened 2020-09-04
Keys.h human readable definitions

### LINKED ISSUE #158: Document scancode preprocessor definitions
Currently the preprocessor defs for use in keymaps is a bit opaque at times, some abbreviations aren't obvious as to what they are supposed to be. Even end-of-line comments in `keys.h` would be of use.

tominabox1:
Added human readable definitions to the key codes in keys.h. Also added numpad defines. Resolves #158 and perhaps #21?

### DIFF
--- a/app/include/dt-bindings/zmk/keys.h
+++ b/app/include/dt-bindings/zmk/keys.h
@@ -147,7 +147,7 @@
 #define M_PREV 0xB6 // Media Previous
 #define M_STOP 0xB7 // Media Stop
 #define M_EJCT 0xB8 // Media Eject
-#define M_PLAY 0xCD // Media Play
+#define M_PLAY 0xCD // Media Play/Pause
 #define M_MUTE 0xE2 // Media Mute
 #define M_VOLU 0xE9 // Media Volume Up
 #define M_VOLD 0xEA // Media Volume Down
--- a/app/include/dt-bindings/zmk/keys.h
+++ b/app/include/dt-bindings/zmk/keys.h
@@ -116,6 +116,21 @@
 #define ATSN 0xCE   // Keypad @ (At Sign)
 #define BANG 0xCF   // Keypad ! (Exclamation)
 
+#define PENT 0x58   // Keypad Enter
+#define P1  0x59    // Keypad 1 and End
+#define P2  0x5A    // Keypad 2 and Down Arrow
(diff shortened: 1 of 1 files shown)


app/include/dt-bindings/zmk/keys.h
```diff
@@ -4,146 +4,161 @@
...
+#define PRCT 0xC4   // Keypad % (Percent)
+#define LABT 0xC5   // Keypad > (Greater Than)
+#define RABT 0xC6   // Keypad < (Less Than)
+#define AMPS 0xC7   // Keypad & (Ampersand)
+#define PIPE 0xC9   // Keypad | (Pipe)
+#define COLN 0xCB   // Keypad : (Colon)
+#define HASH 0xCC   // Keypad # (Pound/Hash/Number)
+#define KSPC 0xCD   // Keypad Space
+#define ATSN 0xCE   // Keypad @ (At Sign)
+#define BANG 0xCF   // Keypad ! (Exclamation)
```
CrossR, line 117: For quite a few of these, do we want them there if they don't do anything (or at least that is my current understanding).

Perhaps there should be a block comment in this file with "These symbols can be achieved with X syntax instead, and are ommited here due to changing  for different OS keyboard layouts, see this doc: ...".

Then we can list all these symbols there so if a user does a search in this file for "Exclamation" they get a link to the upcoming shifted docs instead of being confused why its missing.

innovaker:
@tominabox1, this is great!  Thanks for doing it 👍 

#21 has a larger scope, but your efforts will be folded into it for sure.

Please keep this issue open until #21 is closed.

innovaker:
@tominabox1, I'd like to integrate your descriptions into an upcoming RFC document for #21 (for community feedback).  The RFC document will likely be a Google sheet (read-only).  Please may I have your permission to use it for this purpose?

tominabox1:
> @tominabox1, I'd like to integrate your descriptions into an upcoming RFC document for #21 (for community feedback). The RFC document will likely be a Google sheet (read-only). Please may I have your permission to use it for this purpose?

Yes of course!

tominabox1:
I'm OK if we just close this since @innovaker has really superseded this work with his work on #236

innovaker:
Thanks @tominabox1!  I was waiting until #236 gets merged, but you're right - we might as well close this now.

PR #163, opened 2020-09-06
Add RoMac / non-split template support

bmcgavin:
You will definitely want to clone this repo : https://github.com/bmcgavin/zmk-config-template and then remove the .git folder and push up to the zmkfirmware account - when this is done I can change this PR to point to the 'official' template repo.


petejohanson:
@bmcgavin I've just gone ahead and forked that repo of yours into https://github.com/zmkfirmware/zmk-config-template so you can update w/ that repo URL.

petejohanson (approved):
LGTM! Thanks.

PR #167, opened 2020-09-08
Blank OLED after one second

bmcgavin:
Quick hack to blank the OLED after it displays the ZMK message for one second


bmcgavin:
This is silly, closing.

PR #170, opened 2020-09-10
Create a simple macro behavior

BrainWart:
I haven't done any testing after bringing it up to main. It was designed very similarly to the original hold-tap design. I have a sample [keymap](https://github.com/BrainWart/zmk-kyria-keymap/blob/simple-macro/config/kyria.keymap) from when I was testing.

Currently, there are 2 macro modes: tap, roll-on?. 'tap' will press and release each behavior in order. The non-'tap' mode will press each key in order then release them in reverse order. Rolled on, rolled off. Haven't worked on any actual documentation yet because I don't expect this to be too close to what the final macro code will look like.


okke-formsma:
A few thoughts

* I think most use cases for the 'roll-on' behavior is covered by the modifiers branch, so I'm not sure if that's still needed. 
* Shall we execute the tap macros on key down instead of key up?

BrainWart:
I think the roll on could still be used for machine side combos using Auto hotkey or something similar. I don't think it should be default anymore though. Change the tap parameter to hold maybe and swap the default around. Is there any reasons you can think of to execute macros on key up rather than key down?

okke-formsma:
Maybe it would be cool to be able to do both - execute one combo on key down, and one on key up.

Or have three slots; key down taps, key up taps and holds. No "mode selection" required.

BrainWart:
From Discord:

> [3:16 AM] BrainWart: @Okke, what did you mean by slots? Did we ever figure out a way to use behaviors without using bindings?
> [3:17 AM] Okke: I was hoping we could put multiple bindings definitions in one behavior... But no such luck?(edited)
> [3:19 AM] BrainWart: Like multiple bindings? May be possible to do something similar to the keymap, but I think that would mean each one would have to be defined for every macro. Or at least the order would be significant. So the first is always tap-down second tap-up, and third hold/roll on & roll off
> [3:21 AM] Okke: Hmm  that's not great either
> [3:24 AM] BrainWart: Maybe the mode toggle just changes how the keys are released? Does it matter?
> [3:24 AM] BrainWart: Yeah I guess it does for combos   Or not. Release order won't matter either if all the keys are pressed when the key is hit and released when the key is released. Maybe the toggle is hold and it just defines whether or not the macro can be held. otherwise it executes on key down.(edited)
> [3:28 AM] BrainWart: if the user wants to change it to be on key up or something fancy, they could combine it with the tap dance stuff in the future. I don't think the macro stuff should be too complex
> [3:28 AM] Okke: Or toggle between up/down/hold
> [3:28 AM] BrainWart: could do that
> [3:29 AM] Okke: And if someone really wants all three, you could specify the up and down in the bindings for the hold (edited)
> [3:30 AM] BrainWart: I don't think I quite understand
> [3:30 AM] Okke: On my phone now, will make example later(edited)
> [3:31 AM] Okke: "nest them"
> [3:32 AM] Okke: But that use case is theoretical at best(edited)
> [3:33 AM] BrainWart: Gotcha so something like:
> ```
> macro1 {
>   mode: key-up;
>   bindings: <blah>;
> }
> 
> macro2 {
>   mode: key-down;
>   bindings: <blah>;
> }
> 
> macro3 {
>   mode: hold;
>   bindings: < &macro1 &macro2 >;
> }
> ```
> [3:34 AM] BrainWart: On down, macro 3 will execute both but, macro 1 only fires on key up
> [3:34 AM] BrainWart: and the opposite for macro 2
> [3:44 AM] Okke: Exactly! The ability to combine behaviors is really powerful
> [3:44 AM] BrainWart: Can be for sure
> [3:44 AM] BrainWart: Not a newb feature though xD
> [3:49 AM] Okke: Definitely not. I think we should provide just enough default behaviors that newbs don't need to touch behaviors
> [3:49 AM] Okke: Then they can start configuring them by overriding a few params
> [3:50 AM] Okke: Next level adding a new behavior themselves to have more variety in settings (eg mod taps with different timings)
> [3:50 AM] Okke: Next level writing combos/tap dance/macros
> [3:52 AM] Okke: I kinda want to get started on the combos because I feel I need those on a 36 key board
> [3:52 AM] BrainWart: They'll be a chore to get right
> [3:53 AM] Okke: Yeah there's some tricky stuff for sure
> [3:53 AM] Okke: I think I have a plan though
> [3:55 AM] Okke: I'll have to start diagramming again, without that I'm lost
> [4:02 AM] BrainWart: That's fair. Your diagrams are good too
> [4:03 AM] BrainWart: I don't have my kyria handy. I'll have to try the changes later today

okke-formsma:
This PR seems pretty good, but I think we need some tests to verify each of it's options works.

okke-formsma:
closed in favor of #685

PR #180, opened 2020-09-13
Shell script tidy up.

CrossR:
I've tidied up the existing `setup.sh` script as well as implementing a few of the TODOs.
I also added an equivalent script for PowerShell, since (IMO) asking for WSL just to run a single script seems overkill.

Assuming it looks good, I can add it to the docs as well, should just be a case of running

```
iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/CrossR/zmk/CrossR/ShellScriptTidyUp/docs/static/setup.ps1'))
```

The main question I had, which I expect someone has already thought about, is how to deal with the growing list of boards/shields and multiple sources of truth? Its already a pain when there is the source code and the bash script, and now there is a third source. Having a single location and deriving these two scripts from it seems best, though probably a pain in bash/powershell. That question doesn't need to be answered now though.

Potentially easier in something like Python, but I also like that the setup so far only really needs git so adding a dependency on Python could be a pain for Windows.

### DIFF
--- a/docs/static/setup.ps1
+++ b/docs/static/setup.ps1
@@ -16,7 +16,7 @@ function Get-Choice-From-Options {
 
         if ($selection -eq $Options.length + 1) {
             Write-Host "Goodbye!"
-            exit
+            exit 1
         }
         elseif ($selection -le $Options.length) {
             $choice = $($selection - 1)
@@ -40,7 +40,7 @@ function Test-Git-Config {
 
     if ($lastExitCode -ne 0) {
         Write-Host $ErrMsg
-        exit
+        exit 1
     }
 }
 
@@ -49,7 +49,7 @@ try {
 }
 catch [System.Management.Automation.CommandNotFoundException] {
     Write-Host "Git is not installed, and is required for this script!"
-    exit
+    exit 1
 }
 
 Test-Git-Config -Option "user.name" -ErrMsg "Git username not set!`nRun: git config --global user.name 'My Name'"
@@ -134,7 +134,7 @@ $do_it = Read-Host "Continue? [Yn]"
 
 if ($do_it -ne "" -and $do_it -ne "Y" -and $do_it -ne "y") {
     Write-Host "Aborting..."
-    exit
+    exit 1
 }
 
 git clone --single-branch "$repo_path" "$repo_name"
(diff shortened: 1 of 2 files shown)


petejohanson:
> The main question I had, which I expect someone has already thought about, is how to deal with the growing list of boards/shields and multiple sources of truth? Its already a pain when there is the source code and the bash script, and now there is a third source. Having a single location and deriving these two scripts from it seems best, though probably a pain in bash/powershell. That question doesn't need to be answered now though.
> 
> Potentially easier in something like Python, but I also like that the setup so far only really needs git so adding a dependency on Python could be a pain for Windows.

@CrossR I had originally envisioned, once this got big enough, or we added a PS script, would be change the `setup.sh` to actually be `setup.sh.in` and have it be templated, so that we could process it at "build time" for the docs site, and inject the current set of shields, their settings, etc. there.

If you're up for trying to take a stab at that, that would be great. If not, we can create an issue to capture that plan.

Either way, I'm not sure we should *block* getting the PS script in on that work. I will try to do a detailed review on this PR as time permits.

Thanks!

petejohanson (commented):
Two minor nitpicks. Can you give me a screenshot/log of using the PS script, since I can't easily test myself?

Thanks for working on this!

CrossR:
Comments should be fixed, and I've stuck a quick screen recording up here : https://drive.google.com/file/d/1LO6qa4gElXLTmvc-pRoput84V-pyciRY/view?usp=sharing and the repo it made is here : 
https://github.com/CrossR/temp_zmk_config.

Could be worth someone else on Windows giving it a quick run as well?

docs/static/setup.ps1
```diff
@@ -0,0 +1,189 @@
+$ErrorActionPreference = "Stop"
```
petejohanson, line 1: Can you please add a copyright and license header here?

docs/static/setup.sh
```diff
@@ -1,14 +1,25 @@
-#!/bin/sh
+#!/bin/bash
```
petejohanson, line 1: While you are at it, can you please add a header here too after the magic start line?

petejohanson (commented):
A couple minor header needs, then this is good to go. Thanks.

docs/static/setup.ps1
```diff
@@ -0,0 +1,193 @@
+# Copyright (c) 2020 Ryan Cross
```
petejohanson, line 1: Sorry I wasn't clear before, can we stick with the single author file approach please, e.g. 


```suggestion
# Copyright (c) 2020 The ZMK Contributors
```

docs/static/setup.sh
```diff
@@ -1,14 +1,29 @@
...
+
+# Copyright (c) 2020 Pete Johanson, Ryan Cross
```
petejohanson, line 3: Ditto.

petejohanson (changes requested):
Sorry, one minor tweak, then this is good.

petejohanson (approved):
Thanks!

PR #186, opened 2020-09-14
Integration Tests Should Fail on Build Errors

BrainWart:
The issue was the script would return an exit code of 0 because the echo succeeds. Just added an exit statement and all's well.


petejohanson (approved):
:+1:

PR #190, opened 2020-09-18
docs: "%USERPROFILE%\AppData\Roaming\Python\Python38\Scripts"

### LINKED ISSUE #189: docs: Windows Setup: 'west' not found
The current setup guide for Windows development has an oversight which I think was introduced by #135.

`pip3 install --user -U west` installs west to the local user
but the suggested `C:\Python38\Scripts` is global

This can trip up unsuspecting users.

Further chat about this can be found here:
https://discordapp.com/channels/719497620560543766/719544230497878116/756592929560199228

innovaker:
Fixes #189.

### DIFF
--- a/docs/docs/dev-setup.md
+++ b/docs/docs/dev-setup.md
@@ -213,7 +213,7 @@ source ~/.bashrc
 <TabItem value = 'win'>
 
 1. See the [Environment Variables](#environment-variables) section on how to get to the Environment Variables page.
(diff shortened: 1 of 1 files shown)


innovaker:
Please could someone double check this on a fresh(ish) install before it's accepted.

jmding8:
Thanks for this, the new directory looks correct. As a newbie, it would also be helpful to be explicit that this should be added to the ```Path``` environmental variable.

innovaker:
I think this PR has been superseded, see #189.  Closing for now.

PR #192, opened 2020-09-19
VS Code improvements

joelspadin:
Added documentation to help set up VS Code. Also added some extension recommendations.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -8,6 +8,7 @@
 		"esbenp.prettier-vscode",
 		"ms-vscode.cpptools",
 		"plorefice.devicetree",
+		"spadin.zmk-tools",
 		"trond-snekvik.kconfig-lang",
 		"twxs.cmake",
 	],
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,16 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"ms-vscode.cpptools",
+		"plorefice.devicetree",
+		"trond-snekvik.kconfig-lang",
+		"twxs.cmake",
+	],
+	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+	"unwantedRecommendations": [
+
+	]
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
 
 	// List of extensions which should be recommended for users of this workspace.
 	"recommendations": [
+		"esbenp.prettier-vscode",
 		"ms-vscode.cpptools",
 		"plorefice.devicetree",
 		"trond-snekvik.kconfig-lang",
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
 
 	// List of extensions which should be recommended for users of this workspace.
 	"recommendations": [
+		"dbaeumer.vscode-eslint",
 		"esbenp.prettier-vscode",
 		"ms-vscode.cpptools",
 		"plorefice.devicetree",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,9 @@
 {
+  "C_Cpp.autoAddFileAssociations": false,
   "files.associations": {
     "*.overlay": "dts",
-    "*.keymap": "dts"
+    "*.keymap": "dts",
+    "*.h": "c"
   },
   "[c]": {
     "editor.formatOnSave": true
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,8 @@
   "files.associations": {
     "*.overlay": "dts",
     "*.keymap": "dts"
+  },
+  "[c]": {
+    "editor.formatOnSave": true
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,6 @@
 {
   "C_Cpp.autoAddFileAssociations": false,
   "files.associations": {
-    "*.overlay": "dts",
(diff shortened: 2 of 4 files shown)


innovaker:
These are some great ideas, thanks @ChaosinaCan!  I've been running with the compile_commands on since you suggested it on Discord and its very useful.  I only took a quick look (for now) ...
- Perhaps touch on what the Compilation Database facilitates?
- Have the compiler paths been verified empirically on each OS?
- Is this better as one or two PRs, Pete?

ChaosinaCan:
I added a new patch which I hope better explains what a compilation database is and what it enables in VS Code.

I have fully verified the compiler paths on Ubuntu and Windows. Fedora is the same as Ubuntu in http://localhost:3000/docs/dev-setup#toolchain-installation so I didn't test that. 

I have only partially verified it on a Rasbperry Pi. I installed gcc-arm-none-eabi and found where it installed GCC, but I haven't tested whether running vscode on a RPi and using that configuration works. (I can do this if you like, I just didn't spend the time to set up ZMK on a RPi yet.)

I do not have access to anything that runs macOS so I was not able to verify that. I am not sure whether VS Code will run in an environment where GNUARMEMB_TOOLCHAIN_PATH is set, so that may need to be changed to say you always need to enter the full path.

innovaker:
Thanks!!!

I checked the RPi path too and can confirm.  Unfortunately my pi is headless.  Perhaps testing a working setup via VS Code Remote (SSH) would suffice?

We need a mac user!

idan:
🍎 user here, will see how this plays with #209 since I used the docker container in order to avoid having to set up the toolchain on my actual machine.

idan:
#209 is working, now getting into the documentation there. I'll try the stuff here tonight, maybe the right approach is to rebase/merge this on top of #209?

ChaosinaCan:
Rebased onto the head of main and added a few more config improvements.

ChaosinaCan:
@idan I don't think anything in our PRs directly conflicts, but if #209 gets merged before this, I'll update dev-ide-integration.md to include instructions for Docker. If this gets merged first, I can just make a new PR.

idan:
With #209 getting merged first, I'm happy to help get this one rebased since I caused it some mess 😊

joelspadin:
It all rebased cleanly. I just needed to add Docker info to the new docs file I added.

Also added another setting tweak since I noticed the C/C++ extension likes to edit .vscode/settings.json.

docs/docs/dev-ide-integration.md
```diff
@@ -100,11 +104,20 @@ If not, you will need to list the full path instead of using the `${env}` placeh
...
 This assumes you export `GNUARMEMB_TOOLCHAIN_PATH` in your `~/.bash_profile`.
 If not, you will need to list the full path instead of using the `${env}` placeholder.
 
+</TabItem>
+<TabItem value="docker">
+
+Your compiler path is
+
+```sh
+/opt/toolchains/zephyr-toolchain-arm-0.11.2/arm-zephyr-eabi/bin/arm-none-eabi-gcc
```
joelspadin, line 119: Should be arm-zephyr-eabi-gcc

joelspadin: Fixed.

.vscode/extensions.json
```diff
@@ -0,0 +1,16 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"ms-vscode.cpptools",
```
petejohanson, line 7: So, if I leave this extension enabled, it tries to parse *thousands* of files, and bombs. I had to tweak my local config to only include `app/` subdirectory, and then that worked. Perhaps I have way too many west/zephyr modules in my crufty local checkout?

Anyone else encountered issues w/ this?

joelspadin: The C/C++ extension? I haven't had that problem, possibly because I have a c_cpp_properties.json set up. That said, I also use it for a project at work which is far larger than ZMK without issues.

innovaker:
Thanks @joelspadin, it looks good.  I had a cursory look through and it seems fine.  I haven't tested it yet though.

My only concern is around having `tasks.json` in git.  But that's a more general issue than this PR so I kicked off a conversation about it in [Discord](https://discordapp.com/channels/719497620560543766/719565084208398406/767834957962805308).

innovaker:
I sense there is still value in this PR but some bits are blocking others.

If there's no dependency between the doc addition and the vscode workspace settings, please may I suggest plucking out `dev-ide-integration` into a separate PR (assuming it's still relevant)?

joelspadin:
Yeah. I'll rebase this and split it up into smaller chunks this weekend.

joelspadin:
Docs changes split out to https://github.com/zmkfirmware/zmk/pull/387

joelspadin:
Rebased and removed docs changes. Also added eslint and Prettier extension recommendations and format on save for docs files.

mcrosson:
I've been running with the editor settings while working on zmk using vs code + devcontainer setup and it's been quite a positive experience. There are mild delays sometimes when it comes to the save but on the whole I prefer the editor settings being active than not. Saves many possible hiccups with clang-format when building a PR.

.vscode/settings.json
```diff
@@ -1,6 +1,25 @@
 {
+  "C_Cpp.autoAddFileAssociations": false,
   "files.associations": {
-    "*.overlay": "dts",
```
joelspadin, line 3: The ZMK Tools extension now does this automatically, so a setting is not necessary.

.vscode/settings.json
```diff
@@ -1,6 +1,25 @@
...
-    "*.keymap": "dts"
```
joelspadin, line 4: ZMK Tools now requires that .keymap files *not* be overridden so it can handle them as keymap files and not DT files.

joelspadin:
Updated to add format on save for TypeScript files and tweaked tasks.json to allow setting a default build task in a code workspace.

joelspadin:
This is super out of date. A new PR would be better for any config improvements now.

PR #207, opened 2020-09-24
Remove bt_unpair_combo from romac_plus.overlay

reizero00:
I'm not sure but I found the bt_unpair_combo in the overlay file. This PR seeks to clean this up.


petejohanson (approved):
👍

Issue #208, opened 2020-09-24
Discuss: proposed docker instructions for development setup


idan:
After chatting a bit with @petejohanson yesterday in discord, I tinkered with getting image builds working for local development on my mac using docker. It works, it's way simpler to set up, and I think there might be some wisdom in making it the recommended development setup.

The series of commands to go from nothing to build on MacOS amounts to this:

```bash
# first, install docker
# Either https://docs.docker.com/docker-for-mac/install/
# or
$ brew install docker
# then...
$ docker pull zmkfirmware/zephyr-west-action-arm
$ git clone https://github.com/zmkfirmware/zmk.git
$ cd zmk
$ docker run -w /zmk -v "$PWD:/zmk" zmkfirmware/zephyr-west-action-arm init "-l app"
$ docker run -w /zmk -v "$PWD:/zmk" zmkfirmware/zephyr-west-action-arm update
$ docker run -w /zmk -v "$PWD:/zmk" zmkfirmware/zephyr-west-action-arm update zephyr-export
$ docker run -w /zmk/app -v "$PWD:/zmk" zmkfirmware/zephyr-west-action-arm build "-b nice_nano -- -DSHIELD=sofle_left"
```

I have a .uf2 file sittin' pretty in `zmk/app/build/zephyr` as a result.

### Pros

1. Standardization — the docker image which powers the GH action is the same one used for development. [Dev/prod parity!](https://12factor.net/dev-prod-parity)
2. The development setup instructions ought to be almost identical across Windows/Linux/MacOS.

### Cons / Annoyances discovered so far

- The `west update` step took a hot minute without appearing to have output.
- `west`: four chars `docker run ...`: one billion chars. Maybe a good little shell script to do the invocations is called for.
- I have some nice!nanos, which flash with UF2s. If direct flashing with `west` is called for, I don't know whether that's possible with docker, and whether that's going to be a giant PITA (giving docker the right USB device, etc etc). Even if it works, it might be very different across OSes.
- … Other stuff I haven't encountered yet!

I haven't started actually developing yet. By not installing any of the development dependencies on my machine, it might also hamper my editor's ability to do nice things like symbol autocompletion. Need to kick the tires further to see how that plays out.

WDYT? Happy to roll a docs PR but wanted a short discussion first.

petejohanson:
[...]

PR #209, opened 2020-09-24
Add VSCode devcontainer configuration

idan:
This creates the a basic harness that permits users to develop with their local checkout inside the same docker container which powers the GitHub action. Does not yet contain docs and tests, but I did try it out with a clean checkout and was able to build images. 🎉 

See #208 for discussion about where this can go.

## Steps to use:

1. Install docker
2. Install the [Remote - Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) either via the extensions panel, or in your favorite terminal:
`code --install-extension ms-vscode-remote.remote-containers`
3. Opening the `zmk` folder in VS Code now displays a prompt to re-open inside a container:
image
4. Click `Reopen in Container` and the VS Code window will reload and do the necessary initial docker setup. This can take a little while.
5. Open a terminal in VS Code and issue the following commands to initialize your environment:

```bash
$ west init -l app
$ west update # and go make lunch, this is going to take a while
$ cd app
$ west build -b YourBoardName -- -DSHIELD=YourShieldName
```

That's it!


CrossR:
Some potential to combine this with #192, and include the Docker ext as a recommended extension, though also raises the question of it the instructions provided there work for a Docker based workflow (making the compilation database, using it etc).

ChaosinaCan:
I suspect that the compilation database part is still correct, but you'd need to change the compiler path to match wherever it ends up being in the container. Maybe we just add Docker to the list of OSes in all the docs pages?

petejohanson:
> I suspect that the compilation database part is still correct, but you'd need to change the compiler path to match wherever it ends up being in the container. Maybe we just add Docker to the list of OSes in all the docs pages?

Yeah, if we can make this seemless, I definitely think it should be an "OS tab/option" in the dev setup docs.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "ZMK Development",
+  "dockerFile": "Dockerfile",
```
petejohanson, line 3: It isn't easily found in the docs, but you can just replace this with:

```suggestion
  "image": "zmkfirmware/zephyr-west-action-arm",
```

And remove the extraneous `Dockerfile`.

The other thing I had to do locally here, to make this work on my Fedora machine w/ SELinux enforcement enabled, was to *also* add a line:

```
"runArgs": ["--security-opt", "label=disable"],
```

Does it work for you locally if you also have that in this file?

petejohanson (changes requested):
One tweak, and a question around SELinux settings.

.devcontainer/devcontainer.json
idan: Yup, fixed in 164cd97

idan:
So, the tweak above worked. Onto new issues!

Git pull/push with SSH remotes doesn't work inside the container because the docker image lacks SSH. Fixable with `apt update; apt install ssh`. We should probably bake that into the image? Once that's there, it Just Works™, and the devcontainers extension properly forwards things to the host's ssh-agent.

The more annoying (but less critical) thing I've run into is that GPG commit signing doesn't work. Like SSH, GPG is not installed in the image (easily fixable like above). Once it's there, it still doesn't work — the `~/.gnupg` directory doesn't even exist:

image


Supposedly, this [should work as of v118 of the remote-containers extension](https://github.com/microsoft/vscode-remote-release/issues/72), but there's no documentation. However, [another issue](https://github.com/microsoft/vscode-remote-release/issues/3221#issuecomment-654724269) identifies a control to test against. When I try to gpg from inside the "try a sample" node container, I can see my keyring and a socket, and signing works fine:

image

My current hunch is that maybe the image needs to contain GPG already when it boots or the VSCode remote-containers extension doesn't try to set up forwarding.

idan:
To be clear:

* SSH is critical, you can't git without it in many cases
* Commit signing is nice to have
* You can always commit from outside the container; this is all about quality of life at this point

idan:
OK, it works!

.devcontainer/devcontainer.json
idan: Ended up returning to the dockerfile approach so we could use it to install SSH and GPG.

petejohanson (commented):
One comment on the dockerfile, but certainly minor. I would love to get this in, so we can build on it, even if we aren't 100% happy with the documentation piece.

Thoughts @idan ?

idan:
@petejohanson yes! I'll hack on the docs tonight and try to get it into a happy place, or move them to a separate branch so we can commit the docker bits and do the docs in a follow-onw

petejohanson (approved):
This is fabulous! Thanks sticking with this, and the docs updates.

Issue #202, opened 2020-09-23
Implement polling based press detection for matrix KSCAN driver.


petejohanson:
Currently, the matrix KSCAN driver in ZMK (`app/drivers/zephyr/kscan_gpio_matrix.c`) *only* implements key press detection via interrupts, which is great for low power.

However, for STM32 wired keyboards, this can be a problem as STM32 doesn't allow enabling interrupts for the same pin on different ports, e.g. `&gpioa 15` and `&gpiob 15`, at the same time.

To support those, we should implement polling based detection for the matrix driver. We *already* have this for the direct wire KSCAN driver (`app/drivers/zephyr/kscan_gpio_direct.c`), so we can use that as a reference for how to do this.

It should be behind a new Kconfig flag, so it can be enabled only for wired builds that aren't going for low power usage.

shurikai:
It appears that this behavior is already controlled in the gpio_direct driver using the KConfig flag of `CONFIG_ZMK_KSCAN_GPIO_POLLING=y`. Is there any objection to me reusing this flag for the other driver in this instance?

petejohanson:
@shurikai So... I think it may be worthwhile to have a separate config for each. In the case of a funky composite device we may want to turn on polling for one but not the other.

shurikai:
Will do.

shurikai:
Is there a standard for naming these config options?

petejohanson:
@shurikai Somewhat. They should definitely be `ZMK_` prefixed, ideally then followed by the feature/area that will use it, then followed by the specific name of the config. So for this, perhaps something like `ZMK_KSCAN_MATRIX_POLLING` and we rename the existing one to `ZMK_KSCAN_DIRECT_POLLING`.

Thoughts?

shurikai:
That seems like a reasonable approach to me. I kept coming up with clumsier names. I’ll make that the name going forward.

PR #212, opened 2020-09-25
Add polling (instead of interrupt) based matrix kscan detection, to avoid interrupt limits on stm32

shurikai:
This branch is intended to address #202. I added code to do timer-based polling of the matrix instead of the original interrupt-only model.  This will hopefully pave the way for a functional Planck or Preonic board.

Summary of changes:

- A new Kconfig flag was added to enable matrix scanning (ZMK_KSCAN_MATRIX_POLLING). This is defaulted to n overall, but should be set to y for boards like Planck. 
- The polling flag used by the direct wire driver was renamed to better match the new one (ZMK_KSCAN_DIRECT_POLLING).
- Code changes were made to the matrix driver (kscan_gpio_matrix.c) to add the appropriate matrix poll logic when the flag is enabled. 

All changes were modeled after the code written in kscan_gpio_direct.c, which does something similar.

### DIFF
--- a/app/drivers/zephyr/kscan_gpio_direct.c
+++ b/app/drivers/zephyr/kscan_gpio_direct.c
@@ -33,9 +33,9 @@ struct kscan_gpio_config {
 };
 
 struct kscan_gpio_data {
-#if defined(ZMK_KSCAN_DIRECT_POLLING)
+#if defined(CONFIG_ZMK_KSCAN_DIRECT_POLLING)
     struct k_timer poll_timer;
-#endif /* defined(ZMK_KSCAN_DIRECT_POLLING) */
+#endif /* defined(CONFIG_ZMK_KSCAN_DIRECT_POLLING) */
     kscan_callback_t callback;
     union work_reference work;
     struct device *dev;
@@ -53,7 +53,7 @@ static const struct kscan_gpio_item_config *kscan_gpio_input_configs(struct devi
     return cfg->inputs;
 }
 
-#if !defined(ZMK_KSCAN_DIRECT_POLLING)
+#if !defined(CONFIG_ZMK_KSCAN_DIRECT_POLLING)
 
 struct kscan_gpio_irq_callback {
     union work_reference *work;
@@ -101,7 +101,7 @@ static void kscan_gpio_irq_callback_handler(struct device *dev, struct gpio_call
     }
 }
 
-#else /* !defined(ZMK_KSCAN_DIRECT_POLLING) */
+#else /* !defined(CONFIG_ZMK_KSCAN_DIRECT_POLLING) */
 
 static void kscan_gpio_timer_handler(struct k_timer *timer) {
     struct kscan_gpio_data *data = CONTAINER_OF(timer, struct kscan_gpio_data, poll_timer);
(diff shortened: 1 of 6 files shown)


app/drivers/zephyr/kscan_gpio_direct.c
```diff
@@ -33,9 +33,9 @@ struct kscan_gpio_config {
 };
 
 struct kscan_gpio_data {
-#if defined(CONFIG_ZMK_KSCAN_GPIO_POLLING)
+#if defined(ZMK_KSCAN_DIRECT_POLLING)
```
megamind4089, line 36: I think the macros are all defined with CONFIG_ prefix.
Please test this with and without polling and make sure this works

shurikai: I think this might have been a merge issue before submitting. Will test and correct.

app/drivers/zephyr/kscan_gpio_matrix.c
```diff
@@ -31,7 +31,8 @@ struct kscan_gpio_item_config {
 #define _KSCAN_GPIO_ROW_CFG_INIT(idx, n) _KSCAN_GPIO_ITEM_CFG_INIT(n, row_gpios, idx)
 #define _KSCAN_GPIO_COL_CFG_INIT(idx, n) _KSCAN_GPIO_ITEM_CFG_INIT(n, col_gpios, idx)
 
-static int kscan_gpio_config_interrupts(struct device **devices,
+COND_CODE_0(ZMK_KSCAN_MATRIX_POLLING,
```
petejohanson, line 34: Ditto, this needs to have the `CONFIG_` prefix.

Also, for entire function declarations, just use a normal `#if` or `#ifdef` preprocessor check here.

app/drivers/zephyr/kscan_gpio_matrix.c
```diff
@@ -69,6 +71,7 @@ static int kscan_gpio_config_interrupts(struct device **devices,
     };                                                                                             \
     struct kscan_gpio_data_##n {                                                                   \
         kscan_callback_t callback;                                                                 \
+        COND_CODE_0(ZMK_KSCAN_MATRIX_POLLING, (), (struct k_timer poll_timer;))                    \
```
petejohanson, line 74: Ditto, needs prefix.

petejohanson (changes requested):
Few big ticket items. Thanks!

shurikai:
I've made some changes to fix the comments above, but haven't had a chance to fully test them.

petejohanson (commented):
Just one comment/question.

app/drivers/zephyr/kscan_gpio_matrix.c
```diff
@@ -190,15 +197,33 @@ static int kscan_gpio_config_interrupts(struct device **devices,
...
+            (return kscan_gpio_disable_interrupts_##n(dev);),                                      \
+            (struct kscan_gpio_data_##n *data = dev->driver_data;                                  \
+             k_timer_stop(&data->poll_timer);                                                      \
+             return 0;))                                                                           \
     };                                                                                             \
+    COND_CODE_0(CONFIG_ZMK_KSCAN_MATRIX_POLLING, (),                                                      \
+        (static void kscan_gpio_timer_handler(struct k_timer *timer) {                             \
+             struct kscan_gpio_data_##n *data =                                                    \
+                         CONTAINER_OF(timer, struct kscan_gpio_data_##n, poll_timer);              \
+             k_work_submit(&data->work.work);                                                      \
```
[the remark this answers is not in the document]
petejohanson: I'm going to merge this as is, but possibly some follow up work needed here.

petejohanson (approved):
Fabulous! Thanks for this.

PR #233, opened 2020-10-02
Shield and board definitions for M60

megamind4089:
Added board and shield definition for M60 keyboard and makerdiary nrf52840 m.2 module

Tested all keys and working fine

https://imgur.com/a/Ry9HgvA


app/boards/arm/nrf52840_m2/board.cmake
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: MIT
```
petejohanson, line 1: Please add the copyright header here too.

app/boards/arm/nrf52840_m2/board.cmake
```diff
@@ -0,0 +1,6 @@
...
+
+board_runner_args(nrfjprog "--nrf-family=NRF52" "--softreset")
+
+include(${ZEPHYR_BASE}/boards/common/blackmagicprobe.board.cmake)
```
petejohanson, line 5: Does this come w/ the adafruit nrf52 bootloader by default? Seems to based on the UF2 build stuff. Likely, we should ditch the blackmagicprobe, and just leave nrfjprog as the option here, since eventually we'll likely want to move to adafruit-nrfjprog as our other flash utility option.

app/boards/shields/m60/Kconfig.defconfig
```diff
@@ -0,0 +1,10 @@
+
+if SHIELD_M60
+
+config ZMK_KEYBOARD_NAME
+	default "m60"
+
+config ZMK_KSCAN_MATRIX_POLLING
+	default y
```
petejohanson, line 8: Is this in fact needed? Does interrupt based not work? If not, that would be a bummer, since it'll be much less power efficient.

If so, this should hopefully be temporary, the work I'm doing to enable using PORT events should remove the 8-max IN interrupt problem I *suspect* you might have hit, if you have this here.

app/boards/shields/m60/m60.keymap
```diff
@@ -0,0 +1,26 @@
+#include <behaviors.dtsi>
```
petejohanson, line 1: Missing header, please.

app/boards/shields/m60/m60.keymap
```diff
@@ -0,0 +1,26 @@
...
+#include <dt-bindings/zmk/keys.h>
+#include <dt-bindings/zmk/bt.h>
+
+/ {
+	keymap0: keymap {
+		compatible = "zmk,keymap";
+
+		default_layer {
```
petejohanson, line 9: We've been trying to include at least a layer with the `&bt` bindings *somewhere* for most default keymaps.

app/boards/shields/m60/m60.keymap
```diff
@@ -0,0 +1,26 @@
...
+// | CAPS  |  A  |  S  |  D  |  F  |  G  |  H  |  J  |  K  |  L  |  ;  |  '  |     ENTER    |
+// |  SHIFT  |  Z  |  X  |  C  |  V  |  B  |  N  |  M  |  ,  |  .  |  /  |      SHIFT       |
+// |  CTL  |  WIN  |  ALT  |            SPACE               | ALT | MO(1) |  WIN  | CTRL    |
+// ------------------------------------------------------------------------------------------
+			bindings = <
+	&kp ESC &kp NUM_1 &kp NUM_2 &kp NUM_3 &kp NUM_4 &kp NUM_5 &kp NUM_6 &kp NUM_7 &kp NUM_8 &kp NUM_9 &kp NUM_0 &kp MINUS &kp EQL  &kp BKSP
+	&kp TAB  &kp   Q   &kp   W   &kp   E   &kp   R   &kp   T   &kp   Y   &kp   U   &kp   I   &kp   O   &kp   P   &kp LBKT &kp RBKT &kp BSLH
+	&kp CLCK  &kp   A   &kp   S   &kp   D   &kp   F   &kp   G   &kp   H   &kp   J   &kp   K   &kp   L   &kp  SCLN  &kp  QUOT        &kp RET
+	&kp LSFT   &kp   Z   &kp   X   &kp   C   &kp   V   &kp   B   &kp   N   &kp   M   &kp  CMMA &kp  DOT  &kp FSLH      &kp RSFT
+	&kp LCTL &kp   LGUI   &kp   LALT                     &kp SPC                         &kp  RALT  &mo  1  &kp  RGUI  &kp  RCTL
```
petejohanson, line 22: The `&mo 1` here won't work, since there's no other layer for the stock keymap.

Thoughts on at least putting the `&bt` stuff in a layer there?

petejohanson (changes requested):
Thanks for the PR! A few thoughts/comments, and a few fixes needed.

app/boards/arm/nrf52840_m2/board.cmake
megamind4089: No, but adding blackmagic probe will not affect anything else. 
I feel like we can add any number of supported runners here(also in other boards), so that the end user need not edit the board file, in case he use the different supported runner.

As a firmware, we should not ditch any hardware support, thereby not forcing users to use same hardware.
But as for default, we can make the nrfjproj (and later adafruit-nrfjprog) as the first runner so that it will be used as default.
Does this make sense ?

app/boards/shields/m60/Kconfig.defconfig
megamind4089: Oops, this is an oversight. I was trying ZMK with black pill f411 boards and used this config before i receive the M60.
When porting the code, copied all without thinking much.
Will try once whether all keys working without this polling stuff and remove if not needed.

app/boards/shields/m60/m60.keymap
megamind4089: I tried to keep it simple for this initial PR and add that later. 
But its not much and i think i can add it in this PR itself

app/boards/shields/m60/Kconfig.defconfig
megamind4089: Yes its not needed. Removed it

app/boards/shields/m60/m60.keymap
```diff
@@ -0,0 +1,52 @@
...
+&trans         &bt BT_CLR &none      &none      &none      &none      &none      &none      &none      &none      &none      &none      &none      &reset
+&trans           &bt BT_SEL 0 &bt BT_SEL 1 &bt BT_SEL 2 &bt BT_SEL 3 &bt BT_SEL 4 &kp LARW   &kp DARW   &kp UARW   &kp RARW   &none      &none      &trans
+&trans                &none      &none      &none      &none      &none      &none      &none      &none      &none      &none      &trans
+&trans       &trans       &trans       &trans                                                                     &trans     &trans     &trans     &trans
+			>;
+		};
+
+		unknown_layer {
+			bindings = <
+&none      &none      &none      &none      &none      &none      &none      &none      &none      &none      &none      &none      &none      &kp NUM_1
```
megamind4089, line 44: Added this layer to use for alignment. I was not able to define a layer with all none keys.
I think it is a bug ?? So as  a workaround, changed a single key to NUM_1

app/boards/arm/nrf52840_m2/nrf52840_m2.dts
```diff
@@ -0,0 +1,97 @@
...
+	cts-pin = <13>;
+};
+
+&usbd {
+	compatible = "nordic,nrf-usbd";
+	status = "okay";
+};
+
+
+&flash0 {
```
petejohanson, line 64: I think I saw mention of additional flash on this? Not needed for the initial PR, but if it does have extra flash, we should get it defined at some point.

app/boards/shields/m60/m60.keymap
petejohanson: Yeah, that seems to be a current bug w/ upstream Zephyr.

Can we just remove this layer altogether? You end up using a bunch of extra space in the final build uf2, I'm not really sure what for?

app/boards/shields/splitreus62/splitreus62.keymap
```diff
@@ -1,9 +1,10 @@
 /*
- * Copyright (c) 2020 Derek Schmell
```
petejohanson, line 2: @Na-Cly Any issue w/ this update?

app/boards/shields/romac/romac.keymap
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Pete Johanson, Richard Jones
```
petejohanson, line 2: @bmcgavin you good here? We've been checking with contributors before replacing these.

petejohanson (changes requested):
One question/likely change for the keymap, and a few confirmations needed for the header tweaks. If those authors aren't readily available, I would suggest reverting just those couple changes, so we don't block merging this.

app/boards/shields/romac/romac.keymap
megamind4089: Oops, Will revert if there is a concern. Just changed when i tried to change mine.

bmcgavin: No problem here.

app/boards/shields/splitreus62/splitreus62.keymap
Na-Cly: No issues here. Go for it.

app/boards/arm/nrf52840_m2/board.cmake
petejohanson: Yeah, this makes sense to me. Thanks!

megamind4089:
Thanks @bmcgavin and @Na-Cly 

@petejohanson Addressed the comments.

petejohanson (approved):
Looks great! Thanks for adding these.

PR #235, opened 2020-10-03
fix(ec11): allow more than one encoder

ChaosinaCan:
Fixes "device.h:101:11: error: redefinition of '__device_ec11'" in firmware that has more than one encoder.


petejohanson (approved):
Perfect one liner!

PR #237, opened 2020-10-04
Fix direct GPIO when using interrupts

ChaosinaCan:
The changes to kscan_composite.c are just a code improvement I noticed while debugging this code. I can split that out into a separate PR if you'd like.


petejohanson (approved):
If either of these changes was larger, I would definitely want them split, but given the scope of this, this is fine.

Thanks for the fixes!

ChaosinaCan:
You're welcome!

PR #238, opened 2020-10-04
Expose USB status without having USB HID output.

petejohanson:


petejohanson:
This is needed to fully support the power/sleep changes from #211.

PR #242, opened 2020-10-07
[RDY] Driver to control the external power output

megamind4089:
This PR adds support to control the external power output from controllers like nice!nano, nRFMicro etc

I have implemented based on my understanding of Pete suggestion on this feature.
I may also have overdone some parts.

**Testing done:**
* Tested by enabling and disabling the ext_power from application and verified
* Verified the application does not crash with boards that does not have ext_power support
_Note:_ 
I did not test this in nice!nano since I don't have the boards. Will get help from others once the behavior PR is up

**Next Steps:**
* Create a behavior PR to control enable/disable ext_power


app/dts/bindings/zmk,ext-power-nicenano.yaml
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright (c) 2020, The ZMK Contributors
+# SPDX-License-Identifier: MIT
+#
+
+description: |
+  Allows controlling the external power output of nice!nano
+
+compatible: "zmk,ext-power-nicenano"
```
petejohanson, line 9: We don't need a different *binding* for this at all. We'll just need it in the hardware description of the `nice_nano.dts` file where it would configure the generic driver.

app/dts/bindings/zmk,ext-power-generic.yaml
```diff
@@ -0,0 +1,16 @@
...
+# SPDX-License-Identifier: MIT
+#
+
+description: |
+  Generic driver for controlling the external power output
+  by toggling the control-gpio pin status
+
+compatible: "zmk,ext-power-generic"
+
+properties:
```
petejohanson, line 12: You'll likely need a `label` property here for registering the driver instance.

app/include/drivers/ext_power.h
```diff
@@ -0,0 +1,110 @@
...
+ * @brief Get the current status of the external power output
+ * @param dev Pointer to the device structure for the driver instance.
+ * @param status Status sent to the user
+ *
+ * @retval 0 If successful.
+ * @retval Negative errno code if failure.
+ */
+__syscall int ext_power_get(struct device *dev);
+
+static inline int z_impl_ext_power_get(struct device *dev, bool *status) {
```
petejohanson, line 86: Hmmm.... Thoughts on instead using negative return integers for error, `0` for off, `1` for on?

petejohanson (commented):
A few early thoughts. Thanks!

app/include/drivers/ext_power.h
megamind4089: Yeah I considered that, but should we return '0' if some driver does not implement get (which is most unlikely).
Also I can definitely see the benefits of returning status in the return code, which will simplify application code by calling the api without the need for extra `status` variable.

I observed some zephyr driver api's returning status directly.

Its basically boils down to project preference I think, and its better if it is consistent across the entire project. So its your decision

petejohanson: If the specific driver doesn't support get, it should be returning `-ENOTSUP` anyways, so I don't think that's a problem.

I think we should stick to a single int return here, yes. I find it much preferable to bool pointer + return code.

app/src/ext_power_generic.c
```diff
@@ -0,0 +1,109 @@
...
+    struct device *gpio;
+    bool status;
+};
+
+static int ext_power_generic_enable(struct device *dev)
+{
+    struct ext_power_generic_data *data = dev->driver_data;
+    const struct ext_power_generic_config *config = dev->config_info;
+
+    if (gpio_pin_set(data->gpio, config->pin, GPIO_OUTPUT_HIGH)) {
```
petejohanson, line 35: Since you're not using the "raw" version of this, you should be using just `1` for the value, AFAICT, e.g.:

```suggestion
    if (gpio_pin_set(data->gpio, config->pin, 1)) {
```

app/src/ext_power_generic.c
```diff
@@ -0,0 +1,109 @@
...
+    data->status = true;
+    return 0;
+}
+
+static int ext_power_generic_disable(struct device *dev)
+{
+    struct ext_power_generic_data *data = dev->driver_data;
+    const struct ext_power_generic_config *config = dev->config_info;
+
+    if (gpio_pin_set(data->gpio, config->pin, GPIO_OUTPUT_LOW)) {
```
petejohanson, line 48: Ditto:

```suggestion
    if (gpio_pin_set(data->gpio, config->pin, 0)) {
```

app/src/ext_power_generic.c
```diff
@@ -0,0 +1,109 @@
...
+        return -EINVAL;
+    }
+
+    if (gpio_pin_configure(data->gpio, config->pin, config->flags | GPIO_OUTPUT)) {
+        LOG_DBG("Failed to configure ext-power control pin");
+        return -EIO;
+    }
+
+    // Enable the VCC Output on init
+    if (gpio_pin_set(data->gpio, config->pin, GPIO_OUTPUT_HIGH)) {
```
petejohanson, line 81: I have mixed feelings about just doing this on init. I'd rather move this to the *application* init. Thoughts?

petejohanson (commented):
The driver part of this looks really good. Just a few comments. Thanks!

app/include/drivers/ext_power.h
megamind4089: > If the specific driver doesn't support get, it should be returning -ENOTSUP anyways, so I don't think that's a problem.

Once the application got used with the boolean return status, like
```
if (get()) {
    do_something();
}
```
returning -ENOTSUP for unsupported apis will result in the above been succeed.

But I am all in for single int return since it application code a lot

app/dts/bindings/zmk,ext-power.yaml
```diff
@@ -0,0 +1,12 @@
+#
```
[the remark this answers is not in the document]
megamind4089: Ideally the most generic parameters/nodes of a driver will go in there. 
But in our case, there is no such thing. I will remove it.

app/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ target_sources(app PRIVATE src/sensors.c)
 target_sources_ifdef(CONFIG_ZMK_DISPLAY app PRIVATE src/display.c)
 target_sources(app PRIVATE src/event_manager.c)
 target_sources_ifdef(CONFIG_ZMK_BLE app PRIVATE src/ble_unpair_combo.c)
+target_sources_ifdef(CONFIG_ZMK_BLE app PRIVATE src/ext_power_generic.c)
```
megamind4089, line 34: Can I add a Kconfig to control this driver ?

petejohanson: Yes, I would say so. Definitely shouldn't be tied to the `ZMK_BLE` Kconfig setting.

app/dts/bindings/zmk,ext-power-generic.yaml
```diff
@@ -0,0 +1,20 @@
...
+  control-gpios:
+    type: phandle-array
+    required: true
+  label:
+    type: string
+    required: false
```
petejohanson, line 19: This should be required, since it's used to register the single driver instance in the driver itself:

```suggestion
    required: true
```

app/src/ext_power_generic.c
```diff
@@ -0,0 +1,91 @@
...
+    struct device *gpio;
+    bool status;
+};
+
+static int ext_power_generic_enable(struct device *dev) {
+    struct ext_power_generic_data *data = dev->driver_data;
+    const struct ext_power_generic_config *config = dev->config_info;
+
+    if (gpio_pin_set(data->gpio, config->pin, 1)) {
+        LOG_DBG("Failed to set ext-power control pin");
```
petejohanson, line 35: This should be a warning, IMHO, not a debug:

```suggestion
        LOG_WRN("Failed to set ext-power control pin");
```

app/src/ext_power_generic.c
```diff
@@ -0,0 +1,91 @@
...
+    data->status = true;
+    return 0;
+}
+
+static int ext_power_generic_disable(struct device *dev) {
+    struct ext_power_generic_data *data = dev->driver_data;
+    const struct ext_power_generic_config *config = dev->config_info;
+
+    if (gpio_pin_set(data->gpio, config->pin, 0)) {
+        LOG_DBG("Failed to clear ext-power control pin");
```
petejohanson, line 47: Ditto:

```suggestion
        LOG_WRN("Failed to clear ext-power control pin");
```

app/src/ext_power_generic.c
```diff
@@ -0,0 +1,91 @@
...
+    const struct ext_power_generic_config *config = dev->config_info;
+
+    data->gpio = device_get_binding(config->label);
+    if (data->gpio == NULL) {
+        LOG_ERR("Failed to get ext-power control device");
+        return -EINVAL;
+    }
+
+    if (gpio_pin_configure(data->gpio, config->pin, config->flags | GPIO_OUTPUT)) {
+        LOG_DBG("Failed to configure ext-power control pin");
```
petejohanson, line 70: I would consider this an error too, and use `LOG_ERR` here as well

petejohanson (changes requested):
Couple minor comments. This is *really* close to merging. Thanks.

megamind4089:
Thanks @petejohanson for reviewing. Addressed the comments
Also rebased and squashed

app/include/drivers/ext_power.h
petejohanson: I think the key here is to have the code do:

```
if (get() > 0) {
  do_something();
}
```

which is still reasonable.

app/Kconfig
```diff
@@ -95,6 +95,10 @@ config ZMK_IDLE_SLEEP_TIMEOUT
 
 endif
 
+config ZMK_EXT_POWER
+	bool "Enable support to control external power output"
+	default n
```
petejohanson, line 100: I'm not sure defaulting `n` makes sense here. Without this, I think the nrfMicro won't automatically have VCC on, will it?

megamind4089: Yep, it won't. Fixed it.

petejohanson (approved):
Awesome! Looking forward to the next step on this.

PR #248, opened 2020-10-08
docs: add benign uf2 transfer error for macOS to troubleshooting

onebigdoor:
docs: add benign uf2 transfer error for macOS to troubleshooting


petejohanson (approved):
LGTM!

Issue #181, opened 2020-09-13
Update new shield documentation to include split shield info


petejohanson:
The current dev docs for adding a new shield does not cover the details on how to add a new split shield. It should be updated to include that info:

- split .conf files
- core `.dtsi` file and then left/right overlays, including gpio settings.
- Kconfig differences

kurtis-lew:
@petejohanson I was thinking that for t his issue, the "New Shield" page is getting pretty long. Would it be viable to make a tabbed option for users to select between a split new shield guide or a full-size new shield guide?

kurtis-lew:
> @petejohanson I was thinking that for this issue, the "New Shield" page is getting pretty long. Would it be viable to make a tabbed option for users to select between a split new shield guide or a full-size new shield guide?

Actually, scratch this, I have an even better idea of how to format this part of the docs. I'll submit another PR within the next few hours.

petejohanson:
> @petejohanson I was thinking that for this issue, the "New Shield" page is getting pretty long. Would it be viable to make a tabbed option for users to select between a split new shield guide or a full-size new shield guide?

Seems like there is enough overlap, you could have one doc, with tabs, and then the common sections don't need to be repeated in two different places, get out of sync, etc.

caksoylar:
Looks like this was already addressed, closing.

Issue #138, opened 2020-08-30
Missing documentation on the user of `ZMK_CONFIG` define for user keymaps, etc.


petejohanson:
We only have "implicit" documentation on the use of the `ZMK_CONFIG` variable and corresponding directory.

Docs should include:

* Describe the motivation behind the config directories.
  * Keep in-tree keymaps as "stock", and allow users to customize without adding to the core repo.
  * Allow users to version their custom handwires or niche boards out of tree, without forking ZMK and requiring constant merging of upstream ZMK.
* What it is, and what it can contain.
  * Toplevel shield/board config and keymap overrides.
  * `boards/` directory structure for out of tree boards and shields.
* How to build locally using `ZMK_CONFIG`.

petejohanson:
@kurtis-lew is working on this for us. Thanks!

kurtis-lew:
@petejohanson
I don't remember who it was, I believe it was @chenkevinh, but I think that [this page]( https://zmkfirmware.dev/docs/customization) in the docs serves as a good starting point for further documentation for this issue. I'd think it would be appropriate to add the extra details (motivation behind config directories, what they are/what they contain, how to build locally) here instead of adding an entirely new page. What do you say?

caksoylar:
Looks like this was already addressed, closing.

PR #253, opened 2020-10-09
Updating Docs - Issue #138/Issue #181

kurtis-lew:


petejohanson:
@kurtis-lew Are you planning to address all the items from #138 with this, or just the "how to build with it" piece?

kurtis-lew:
> @kurtis-lew Are you planning to address all the items from #138 with this, or just the "how to build with it" piece?

@petejohanson Just the how to build with it piece. I'll submit a separate PR for the other items.

petejohanson:
@kurtis-lew Ok, sounds good. I'll assign that whole issue to you then so it's clear you're working on the rest. Thanks!

petejohanson:
@kurtis-lew Is this ready for review, btw?

kurtis-lew:
@petejohanson yes, it is.

docs/docs/dev-guide-new-shield.md
```diff
@@ -98,6 +132,74 @@ this might look something like:
 };
 ```
 
+:::note
+For split keyboards, it is preferred to define only the `col-gpios` or `row-gpios` in the common shield .dtsi, depending on the diode-direction.
```
petejohanson, line 136:
```suggestion
For split keyboards, it is preferred to define only the `col-gpios` or `row-gpios` in the common shield .dtsi, depending on the `diode-direction` value.
```

docs/docs/dev-guide-new-shield.md
```diff
@@ -98,6 +132,74 @@ this might look something like:
...
+For `col2row` directed boards like the iris, a sample kscan in the .dtsi folder would look like this:
```
petejohanson, line 137:
```suggestion
For `col2row` directed boards like the iris, a sample kscan in the shared .dtsi file would look like this:
```

docs/docs/dev-guide-new-shield.md
```diff
@@ -98,6 +132,74 @@ this might look something like:
...
+};
+```
+
+```
+// iris_right.overlay
+
+#include "iris.dtsi" 
+
+&default_transform { // The matrix transform for this board is 6 columns over because the matrix on the left side is 6 columns wide.
+	col-offset = <6>;
```
petejohanson, line 183: We should call out the use of the offset stuff explicitly too.

docs/docs/dev-guide-new-shield.md
petejohanson: Also, before diving into the specifics of the matrix stuff, can we describe the `.dtsi` file and why we use it, how it is shared common DT settings that gets included into the specific left/right overlay files as a starting point?

petejohanson (changes requested):
This is great! A few comments/suggestions

PR #256, opened 2020-10-10
Exit setup if no write permissions in directory

### LINKED ISSUE #178: setup.sh: Add check to ensure directory is writable before continuing
Currently, running the `setup.sh` script from a directory doesn't have access to write to causes a semi-cryptic error when it goes to create the local repo directory.

We should add a check earlier in the `docs/static/setup.sh` shell script to verify we can write to the current directory, and if not exit with a helpful error message.

careyk007:
Added a check in `setup.sh` that will print a message and exit if the user does not have write permissions in the current directory. Should close #178.

I haven't ever used powershell, but I would be happy to try something similar if a similar issue exists in Windows.

### DIFF
--- a/docs/static/setup.ps1
+++ b/docs/static/setup.ps1
@@ -59,6 +59,19 @@ catch [System.Management.Automation.CommandNotFoundException] {
 Test-Git-Config -Option "user.name" -ErrMsg "Git username not set!`nRun: git config --global user.name 'My Name'"
 Test-Git-Config -Option "user.email" -ErrMsg "Git email not set!`nRun: git config --global user.name 'example@myemail.com'"
 
+$permission = (Get-Acl $PSScriptRoot).Access | 
+?{$_.IdentityReference -match $env:UserName `
+	-and $_.FileSystemRights -match "Read" `
+		-or $_.FileSystemRights -match "Write"	} | 
+			
+		Select IdentityReference,FileSystemRights
+
+If (-Not $permission){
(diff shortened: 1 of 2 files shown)


petejohanson:
> Added a check in `setup.sh` that will print a message and exit if the user does not have write permissions in the current directory. Should close #178.
> 
> I haven't ever used powershell, but I would be happy to try something similar if a similar issue exists in Windows.  

We should add the check there too, just to be safe. I actually have powershell on my Linux laptop I can use to test.

careyk007:
I took a stab at implementing this for powershell based on an example I found [here](https://www.experts-exchange.com/questions/28071040/PowerShell-how-to-verify-and-set-permissions-on-folder-and-files.html).

petejohanson:
The bash version changes are good, tested and worked as expected here.

@CrossR can you test the PowerShell tweaks? I tried to test using `pwsh` on Linux, but the `Get-Acl` cmdlet is Windows only I have learned.

careyk007:
@petejohanson I was able to successfully test the changes on a Windows machine. Seemed to be working as expected for me, but might still be a good idea to have someone else verify.

petejohanson:
> @petejohanson I was able to successfully test the changes on a Windows machine. Seemed to be working as expected for me, but might still be a good idea to have someone else verify.

You testing works well enough for me. Thanks!

petejohanson (approved):
Thanks!

PR #261, opened 2020-10-10
fix: check out Bash scripts with LF

joelspadin:
Always check out Bashs scripts with Unix-style line endings so they work if you run them inside Docker on Windows.


joelspadin:
Moving this so it's not on my repo's main.

PR #263, opened 2020-10-10
fix: allow nested zmk-config for dev container

joelspadin:
VS Code can't share two folders with a Docker container at the same time, so to build with `-DZMK_CONFIG` the zmk-config repo needs to be a subfolder of the zmk repo. Added zmk-config to .gitignore to allow this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@
 /modules
 /tools
 /zephyr
+/zmk-config
 /build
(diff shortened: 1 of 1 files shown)


petejohanson (approved):
I'd like to explore using additional mounts for the container in some standard way, but at least for now, this is a compromise solution that works. Thanks!

PR #264, opened 2020-10-11
Add support for wget to setup.sh

### LINKED ISSUE #200: setup.sh: curl always required
https://zmkfirmware.dev/docs/user-setup provides the option to use `wget`, presumably as an alternative to `curl`.

But it fails on line 112 and line 115 of `setup.sh` which expects `curl` to be on the system.

careyk007:
I have added the ability to use either `wget` or `curl` to `setup.sh`. If the user only has one installed, then the script will use that one. I give the user the choice between the two if both are present, but realized that many users may not care which one gets used. I can remove that option and default to `curl` if both are present if that would make things simpler for the user.

Should close #200.

### DIFF
--- a/docs/static/setup.sh
+++ b/docs/static/setup.sh
@@ -39,21 +39,21 @@ if [ ! -w `pwd` ]; then
     exit 1
 fi
 
+# Parse all commandline options
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        -w|--wget) force_wget="true"; break;;
+        *) echo "Unknown parameter: $1"; exit 1;;
+    esac
+    shift
+done
+
 if [[ $curl_exists == "true" && $wget_exists == "true" ]]; then
-    prompt="Detected both curl and wget, which one would you like to use? [default is curl]"
-    options=("curl" "wget")
-    PS3="$prompt "
-    select opt in "${options[@]}" "Quit"; do
-        case "$REPLY" in
-        
-        1 ) download_command="curl -O "; break;;
-        2 ) download_command="wget "; break;;
-
-        $(( ${#options[@]}+1 )) ) echo "Goodbye!"; exit 1;;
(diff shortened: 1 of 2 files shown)


docs/static/setup.sh
```diff
@@ -29,6 +39,30 @@ if [ ! -w `pwd` ]; then
     exit 1
 fi
 
+if [[ $curl_exists == "true" && $wget_exists == "true" ]]; then
+    prompt="Detected both curl and wget, which one would you like to use? [default is curl]"
+    options=("curl" "wget")
+    PS3="$prompt "
```
megamind4089, line 45: I do not think prompting user will add any value in selecting one over the other, since most (regular) users does not care about it. Can proceed with whatever is available. 
But it might be worth to add an option to force using one over the other in some specific scenario.
Thoughts ?

careyk007: I like the force option, that will allow the users who do care the ability to choose.

careyk007:
When running `setup.sh`, the user can specify either `-w` or `--wget` to force using `wget` over `curl` if both are available. It will still use `curl` if `wget` is not available, which may break the user's expectations. I would imagine that users who choose to force  `wget` will probably know for certain that it is on their system though, so I don't think it's too big of an issue.

Is there somewhere in the documentation that this option should be added?

petejohanson:
> When running `setup.sh`, the user can specify either `-w` or `--wget` to force using `wget` over `curl` if both are available. It will still use `curl` if `wget` is not available, which may break the user's expectations. I would imagine that users who choose to force `wget` will probably know for certain that it is on their system though, so I don't think it's too big of an issue.
> 
> Is there somewhere in the documentation that this option should be added?

This would be added to the `user-setup.md` doc that gives the command to download and run the script.

petejohanson (approved):
Thanks!

PR #268, opened 2020-10-11
fix: don't leak bt_conn refs

joelspadin:
bt_conn_lookup_addr_le() gives us a new reference that must be released with bt_conn_unref()


petejohanson (approved):
Thanks!

PR #269, opened 2020-10-11
feat: only send HID reports to one endpoint

### LINKED ISSUE #206: Ability to switch between USB output and other BLE profiles
Currently, we don't have any runtime support for selecting between outputting HID events to USB or BLE, in particular if both are actively connected.

We should have the ability to:

* When both USB and BLE are *connected*, a behavior you can use to pick one or the other as the active output
* When only one or the other is connected, then that should be defaulted to the output destination.
* When both become connected, we should remember previously selected active output. (stored using Zephyr settings system)

joelspadin:
Fixes #206.

By default reports are sent to USB if both USB and BLE are connected. This also adds a new endpoints behavior to switch between sending to USB or BLE in that case.


app/src/ble.c
```diff
@@ -163,9 +169,9 @@ int update_advertising() {
     struct bt_conn *conn;
     enum advertising_type desired_adv = ZMK_ADV_NONE;
 
-    if (active_profile_is_open() || !active_profile_is_connected()) {
+    if (active_profile_is_open() || !zmk_ble_active_profile_is_connected()) {
```
joelspadin, line 172: Was the `!active_profile_is_connected()` intended here. It looks like that makes the `else if` case below dead code.

app/src/endpoints.c
```diff
@@ -4,58 +4,238 @@
...
+
+    return 0;
+}
+
+struct settings_handler endpoints_handler = {.name = "endpoints", .h_set = endpoints_handle_set};
+#endif /* IS_ENABLED(CONFIG_SETTINGS) */
+
+static int zmk_endpoints_init(struct device *_arg) {
+#if IS_ENABLED(CONFIG_SETTINGS)
+    settings_subsys_init();
```
joelspadin, line 158: This seems a bit redundant with the code in settings.c and ble.c. Is there a better way to do this?

app/src/ble.c
petejohanson: Yeah, this was here due to me playing w/ the directed advertising. I think this can be removed and allow the below `else if` block handle the situation properly.

petejohanson (commented):
A few comments. Are the changes to `docs/package-lock.json` really required?

joelspadin:
> A few comments. Are the changes to `docs/package-lock.json` really required?

Nope. I wasn't intending to change that and probably just accidentally included it.

app/include/dt-bindings/zmk/endpoints.h
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#define ENDPOINT_TOGGLE_CMD 0
+#define ENDPOINT_USB_CMD 1
+#define ENDPOINT_BLE_CMD 2
+
+#define END_TOG ENDPOINT_TOGGLE_CMD
```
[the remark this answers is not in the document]
joelspadin: Done.

joelspadin:
Tested using a nice!nano, and Windows (USB and BLE), and an Android phone (BLE only).

(I don't really see any good way to unit test this unless you have a way to simulate different endpoints.)

adamcstephens:
This works as I'd expect. I'm able to send to USB and BLE separately as toggled while remaining plugged in.

app/src/endpoints.c
petejohanson: This *is* redundant, but the method properly handles being called multiple times and exits early. We can leave for now, until we have a more nuanced settings load order handling

petejohanson (approved):
Love it! Thanks

joelspadin:
You're welcome!

PR #270, opened 2020-10-12
Refactor Kconfig

### LINKED ISSUE #204: Refactor/Cleanup ZMK's KConfig
ZMK's KConfig is currently a little messy.  This screenshot of `west build -t menuconfig` sums it up ...

image

... which menu items belong to ZMK?

### Task Guidelines:
- Redo the menus so that it's easier to spot what's ZMK specific.
- Align with Zephyr's Best Practices: https://docs.zephyrproject.org/2.3.0/guides/kconfig/tips.html
- Introduce some logical formatting to the KConfig file to improve readability - are indents possible?
- Identify and remove unused settings (such as `ZMK_ACTION_MOD_TAP`).
- Anything else which seems wise.
- Don't forget to run the tests before and after :)

careyk007:
menuconfig should look a little nicer now. Screenshots below show the various menu screens. I went with a buried menu approach, but could change it to show/hide options on the main window if that would be better.

In addition to visual layout changes, I removed `ZMK_ACTION_MOD_TAP` since it was no longer used. I also added some code to enable the same config options used for BLE HID over GATT when just enabling BLE for Split communications. (i.e. `select BT`, `select BT_SMP`, etc. are selected for both `ZMK_BLE` and `ZMK_SPLIT_BLE`). I'm not sure if this was strictly necessary, but it doesn't seem to have broken anything.

zmk_overview
zmk_menu
zmk_basic_setup
zmk_split_support
zmk_display_options
zmk_power_management
zmk_kscan_settings

Should close #204.


careyk007:
Ah shoot, checks are failing. Consider this WIP until further notice.

Mariappan:
That's so cool man :+1:

innovaker:
Thanks @careyk007! 👍   Looks good!  Maybe switch it from PR to Draft until it's ready for review?

careyk007:
Checks are passing now, taking this out of draft.

app/Kconfig
```diff
@@ -118,9 +94,17 @@ if ZMK_SPLIT
 
 config ZMK_SPLIT_BLE
 	bool "Split keyboard support via BLE transport"
-	depends on ZMK_BLE
```
petejohanson, line 121: I don't really agree with this tweak. Can you explain why this was necessary?

petejohanson (commented):
In general, loving it! A few questions though on the specifics.

Thanks.

app/Kconfig
```diff
@@ -9,34 +13,36 @@ config USB_DEVICE_PRODUCT
...
 
-menuconfig ZMK_USB
+config ZMK_USB
 	bool "USB"
 	select USB
 	select USB_DEVICE_STACK
 	select USB_DEVICE_HID
 
-if USB
+if ZMK_USB
```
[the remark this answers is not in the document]
careyk007: That makes sense. Switching to `ZMK_USB` allows it to be shown/hidden based on the selection of `USB` in menuconfig. I see the value of keeping it as `USB` though. Thinking through it, the Init Priorities (both USB and BLE) should probably be moved outside of basic keyboard setup since most users will not need to modify those defaults.

app/Kconfig
careyk007: I think this came from wanting a split keyboard that only interfaced with a host via USB. Each half would need to enable BLE for connecting to the other half, but not for connecting to a host. I may have misinterpreted what I needed to do to set that up though.

careyk007:
@petejohanson Thanks for the feedback! I reverted my changes to `USB` and moved the initialization priorities for USB and BLE to the advanced page. 

I would like your thoughts on how to enable BLE for split communication while only allowing USB interaction with a host. I'm happy to revert my changes to `ZMK_SPLIT_BLE` if depending on `ZMK_BLE` is the best approach.

app/Kconfig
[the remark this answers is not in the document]
petejohanson: Yeah, those init priorities really should be moved somewhere "advanced"

app/Kconfig
petejohanson: We can explore that, but I'd rather not make that also part of the refactor here. My general philosophy is:

* Merge a pure refactor, with no additions/features.
* Build your work on top of the refactor as a follow up PR.

It allows keeping focus, and also helping identify the true source of regressions. It also helps reviewers understand scope of changes much more clearly.

petejohanson (changes requested):
In general, this is great! I would ask that we revert the BLE/Split stuff for now, and keep this purely refactor focused, then worth through alternate USB/BLE topology support in a follow up.

Thanks!

careyk007:
Thanks for the reviews! I have reverted my changes to `ZMK_SPLIT_BLE` as well as `build.yaml` to keep this a pure refactor.

petejohanson (approved):
Thanks!

PR #274, opened 2020-10-12
Tweak devcontainer setups for GitHub Codespaces

idan:
Properly detect if we're running inside a codespace, and set up correct defaults so it works.

The reason for this change is that [`${containerWorkspaceFolder}` variable is not supported by Codespaces](https://code.visualstudio.com/docs/remote/devcontainerjson-reference#_variables-in-devcontainerjson).

This will probably almost certainly break if someone forks into a repository that is _not_ called `zmk`, but I think this is an acceptable tradeoff for right now. Hopefully we can find a more robust solution in the future.

cf. #209 for the original work on this.

### DIFF
--- a/.devcontainer/.bashrc
+++ b/.devcontainer/.bashrc
@@ -1,6 +1,8 @@
 export LS_OPTIONS='-F --color=auto'
-eval "`dircolors`"
 alias ls='ls $LS_OPTIONS'
+if [ "${CODESPACES}" = "true" ]; then
+  export WORKSPACE_DIR="$HOME/workspace/zmk"
+fi
 if [ -f "$WORKSPACE_DIR/zephyr/zephyr-env.sh" ]; then
   source "$WORKSPACE_DIR/zephyr/zephyr-env.sh"
 fi
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,6 +3,9 @@
   "dockerFile": "Dockerfile",
   "extensions": ["ms-vscode.cpptools"],
   "runArgs": ["--security-opt", "label=disable"],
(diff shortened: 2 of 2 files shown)


.devcontainer/.bashrc
```diff
@@ -1,6 +1,8 @@
 export LS_OPTIONS='-F --color=auto'
-eval "`dircolors`"
```
idan, line 2: Removed because it works without it

.devcontainer/.bashrc
```diff
@@ -1,6 +1,8 @@
...
 alias ls='ls $LS_OPTIONS'
+if [ "${CODESPACES}" = "true" ]; then
+  export WORKSPACE_DIR="$HOME/workspace/zmk"
```
idan, line 4: On Codespaces, the workspace folder is always in `$HOME/workspace/REPONAME`

.devcontainer/devcontainer.json
```diff
@@ -3,6 +3,9 @@
   "dockerFile": "Dockerfile",
   "extensions": ["ms-vscode.cpptools"],
   "runArgs": ["--security-opt", "label=disable"],
-  "containerEnv": {"WORKSPACE_DIR": "${containerWorkspaceFolder}"}
+  "containerEnv": {"WORKSPACE_DIR": "${containerWorkspaceFolder}"},
+  "settings": { 
+		"terminal.integrated.shell.linux": "/bin/bash"
```
idan, line 8: If you happen to have VS Code settings sync turned on, and you normally use a shell other than `bash` (I use `fish`), then codespaces will complain about that shell not being present and fail to open a terminal. Somehow this works fine when doing containerized local development, but not in codespaces.

petejohanson (approved):
Thanks!

PR #280, opened 2020-10-14
DocumentedTX Power Config in Troubleshooting

Tsquash:
Adds documentation for changing the config values of TX power in the `shield.conf` file. Being able to change this value helps some users with poor connection.

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -80,3 +80,13 @@ to avoid accidental bonding between the halves.
 
 After completing these steps, pair the halves of the split keyboard together by resetting them at the same time. Most commonly, this is done by grounding the reset pins
 for each of your keyboard's microcontrollers or pressing the reset buttons at the same time.
+
+### Connectivity Issues ###
+
+Some users may experience poor connection from their keyboard to their host device, whether it be from a bad Bluetooth card, a metal enclosure on the keyboard or host, or distance to the host device. Increasing the transmit power of your board may solve this problem. To do this, simply enable the `CONFIG_BT_CTLR_TX_PWR_PLUS_8`  configuration value in the `.conf` file of your user config directory as such:
+
+```
+CONFIG_BT_CTLR_TX_PWR_PLUS_8=y
+``` 
+
+On the `NRF52840`, the value `PLUS_8` can be set to any multiple of four between `MINUS_20` and `PLUS_8`. The default value for this config is `0`, but if you are having connection issues it is recommended to set it to `PLUS_8` because the power consumption difference is negligible.
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -83,10 +83,10 @@ for each of your keyboard's microcontrollers or pressing the reset buttons at th
 
 ### Connectivity Issues ###
 
-Some users may experience poor connection from their keyboard to their host device, whether it be from a bad Bluetooth card, a metal enclosure on the keyboard or host, or distance to the host device. Increasing the transmit power of your board may solve this problem. To do this, simply enable the `CONFIG_BT_CTLR_TX_PWR_PLUS_8`  configuration value in the `.conf` file of your user config directory as such:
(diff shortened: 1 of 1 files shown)


docs/docs/troubleshooting.md
```diff
@@ -79,4 +79,14 @@ to avoid accidental bonding between the halves.
 1. Flash the actual image for each half of the split keyboard (e.g `my_board_left.uf2` to the left half, `my_board_right.uf2` to the right half)
 
 After completing these steps, pair the halves of the split keyboard together by resetting them at the same time. Most commonly, this is done by grounding the reset pins
-for each of your keyboard's microcontrollers or pressing the reset buttons at the same time.
\ No newline at end of file
+for each of your keyboard's microcontrollers or pressing the reset buttons at the same time.
+
+### Connectivity Issues ###
+
+Some users may experience poor connection from their keyboard to their host device, whether it be from a bad Bluetooth card, a metal enclosure on the keyboard or host, or distance to the host device. Increasing the transmit power of your board may solve this problem. To do this, simply enable the `CONFIG_BT_CTLR_TX_PWR_PLUS_8`  configuration value in the `.conf` file of your user config directory as such:
```
innovaker, line 86: > Some users may experience poor connection from their keyboard to their host device, whether it be from a bad Bluetooth card, a metal enclosure on the keyboard or host, or distance to the host device.

Some users may experience a poor connection between the keyboard and the host.  This might be due to poor quality BLE hardware, a metal enclosure on the keyboard or host, or the distance between them.

innovaker, line 86: > Increasing the transmit power of your board may solve this problem.

Increasing the transmit power of the keyboard's BLE radio may alleviate this problem.

innovaker, line 86: `CONFIG_BT_CTLR_TX_PWR_PLUS_8` has a default value already, no?  So perhaps

>To do this, simply enable the `CONFIG_BT_CTLR_TX_PWR_PLUS_8`

To do this, set the `CONFIG_BT_CTLR_TX_PWR_PLUS_8`

docs/docs/troubleshooting.md
```diff
@@ -79,4 +79,14 @@ to avoid accidental bonding between the halves.
...
+
+```
+CONFIG_BT_CTLR_TX_PWR_PLUS_8=y
+``` 
+
+On the `NRF52840`, the value `PLUS_8` can be set to any multiple of four between `MINUS_20` and `PLUS_8`. The default value for this config is `0`, but if you are having connection issues it is recommended to set it to `PLUS_8` because the power consumption difference is negligible.
```
innovaker, line 92: > On the `NRF52840`, 
On the `nRF52840`,

innovaker (changes requested):
Good work @Tsquash!  Thanks for adding this.

My suggestions are mainly phrasing.

Perhaps also add an implicit link (hidden behind `CONFIG_BT_CTLR_TX_PWR_PLUS_8`) to the Zephyr documentation:
https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_BT_CTLR_TX_PWR_PLUS_8.html

Tsquash:
Thanks @innovaker for the review! I have changed the wording of the parts you recommended, but I was not able to add the implicit link behind the markdown inline code, so instead I added a sentence at the very bottom which includes the link you gave. "For more information on changing the transmit power, please refer to the Zephyr docs."

innovaker:
The changes look fine, thanks @Tsquash.

Looking at that [documentation](https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_BT_CTLR_TX_PWR_PLUS_8.html) page, it lists `SOC_NRF52840` as a dependency.  Shouldn't that mean it's enabled by default for the nRF52840 chip?  Am I missing something here?  What do you think @petejohanson?

innovaker (approved):

petejohanson:
> The changes look fine, thanks @Tsquash.
> 
> Looking at that [documentation](https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_BT_CTLR_TX_PWR_PLUS_8.html) page, it lists `SOC_NRF52840` as a dependency.  Doesn't that mean it's enabled by default for the nRF52840 chip?  Am I missing something here?  What do you think @petejohanson?

No, that means that config option is only available to even use if that config is set. So, only that nrf chip supports that setting.

petejohanson (approved):
Thanks!

Issue #160, opened 2020-09-04
Feature: sticky keys  (one shot keys)


okke-formsma:
Feature similar in behavior to QMK's https://beta.docs.qmk.fm/using-qmk/software-features/one_shot_keys.

`&osk LSFT` or `&osk LCTL` or `&osk A` (the last one would print 'aaaaaaaa' until the timer runs out or another key is pressed. not very useful but still)

`&osl 1` to toggle to layer 1 until next keycode is pressed or timer runs out.

okke-formsma:
An overview of the behavior such as I think should work.

* The one-shot key behaves as a normal modifier when another key is pressed while it is down
* If the one-shot key is released before another key is pressed, its key-up event will be delayed until after release_after_ms from the one-shot key up event.
* If another key is pressed while waiting for the release_after_ms timer, the one-shot key will be released just after the other key.

[zmk-oneshot-proposal.pdf](https://github.com/zmkfirmware/zmk/files/5184629/zmk-oneshot-proposal.pdf)

okke-formsma:
https://github.com/okke-formsma/zmk/tree/one-shot-key contains a working implementation for one-shot-keys, as described in this rfc.

okke-formsma:
* I'm renaming to sticky keys (&sk for sticky key,  &sl for sticky layer).
* This needs some changes after the behavior bindings signature changed

okke-formsma:
[zmk-sticky-key-proposal.pdf](https://github.com/zmkfirmware/zmk/files/5390075/zmk-sticky-key-proposal.pdf)

For now, sticky keys only respond to keycode events, so a sticky layer key that accesses bluetooth behaviors will not disable the layer after the bluetooth key is released.

okke-formsma:
Merged in  #284

PR #284, opened 2020-10-16
Sticky keys behavior (one-shot mods and one-shot layers)

okke-formsma:
implements sticky keys for https://github.com/zmkfirmware/zmk/issues/160


okke-formsma:
@petejohanson do you know of a way to test the sensors in our test suite? It's not even compiling on a board with sensors like kyria even though the normal test suite succeed.

okke-formsma:
rebased onto latest main, please review.

app/dts/behaviors/sticky_key.dtsi
```diff
@@ -0,0 +1,20 @@
+/ {
```
petejohanson, line 1: Missing copyright/license header.

app/include/zmk/events/keycode-state-changed.h
```diff
@@ -18,14 +18,16 @@ struct keycode_state_changed {
     u32_t keycode;
     u8_t implicit_modifiers;
     bool state;
+    s64_t timestamp;
```
petejohanson, line 21: So, this refactor is mixed in with the feature itself. I'm not opposed to that being in this PR, but can we at least break that into a sequence of two commits, the refactor, and then the feature, so we can review this a bit more easily?

innovaker:
Can we please not bundle vaguely related commits (the mass license headers) into PRs?  A decent git tool will let you fire off a second PR with ease.

petejohanson:
I'd second @innovaker's request. Please do the header add (which is welcome!) as a fresh PR. Thanks! In the meantime, I'll review the code from the other two commits.

okke-formsma:
Can we focus on the code instead of bikeshedding which commits are in this PR?

Frankly I'm a bit bummed out about this thing being open for six weeks and only getting comments about license headers and trivial commits that could be split out to another PR. I don't think that's the thing we should focus on, especially considering the limited time each of us has available. 

I'll split the header commit out to another PR after the reviews are done, so I can batch all changes in one go instead of working on this N times.

innovaker:
@okke-formsma, your contributions are always most appreciated and we're grateful for your patience.  As someone who rebases daily, I can appreciate how frustrating it can be when PRs linger.  With greatest respect though, please keep in mind that as maintainers and reviewers we also have competing priorities and other concerns that might not be immediately obvious.  Also remember that doing a thorough review of an entirely new feature requires a significant *block* of time.  Tidying up a PR takes far less time in comparison, yet makes it significantly easier (faster) for reviewers, which increases the chances of it being reviewed sooner.

innovaker (commented):
I've reviewed the first commit, will do the second later.

6348afb7b39d0b591f9382f1b4f231eb5239fb1d:
- Looks fine, only minor suggestions.
- Commit title: `feat(events): add timestamp to keycode_state_changed and sensor_event`
  - Please be specific.  I had to dig around a little to gather it was only two of the event types.
- Commit message: `dueto` > `due to`
- Can confirm this commit is formatted and passes existing tests.

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,275 @@
...
+#include <zmk/events/position-state-changed.h>
+#include <zmk/events/keycode-state-changed.h>
+#include <zmk/events/modifiers-state-changed.h>
+#include <zmk/hid.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#if DT_NODE_EXISTS(DT_DRV_INST(0))
+
+#define ZMK_BHV_STICKY_KEY_MAX_HELD 10
```
innovaker, line 26: I've noticed the the `ZMK_BHV_` prefix in hold tap also.

Suggest we simplify it to `ZMK_` as there's unlikely to be namespace conflicts?

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,275 @@
...
+
+// increase this if you need more keys in the board
+#define ZMK_BHV_STICKY_KEY_POSITION_NOT_USED 9999
```
innovaker, line 29: Given `s32_t position`, would a negative magic number be better?

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,275 @@
...
+
+struct behavior_sticky_key_config {
+    s32_t release_after_ms;
+    struct zmk_behavior_binding behavior;
+};
+
+// this data is specific for each hold-tap
```
innovaker, line 36: why each hold-tap?

innovaker (commented):
Did a quick pass of eca8a2ccc99f8fb974435f048f1f6798f3850e9f, I'll let @petejohanson do the deep dive.

app/src/behaviors/behavior_sticky_key.c
okke-formsma: I continued this convention from pete's original hold-tap implementation. I don't have a strong opinion on this naming scheme.

app/src/behaviors/behavior_sticky_key.c
okke-formsma: good call

app/src/behaviors/behavior_sticky_key.c
okke-formsma: old copy-paste error. I've removed the entire line as it's rather unnecessary.

okke-formsma:
Thanks for the review @innovaker, I've made most fixes you proposed except for the rename of the constants.

app/src/behaviors/behavior_sticky_key.c
petejohanson: Other note: Do we really think 10 is a reasonable default for this?

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,274 @@
...
+
+struct behavior_sticky_key_config {
+    s32_t release_after_ms;
+    struct zmk_behavior_binding behavior;
+};
+
+struct active_sticky_key {
+    s32_t position;
+    s32_t param1;
+    s32_t param2;
```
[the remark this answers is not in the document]
petejohanson: Ok, I see your use of -1 for unused... Perhaps just use MAX value instead, to avoid typing this as signed?

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,274 @@
...
+        active_sticky_key[i].timer_started = false;
+        active_sticky_key[i].modified_key_usage_page = 0;
+        active_sticky_key[i].modified_key_keycode = 0;
+        return &active_sticky_key[i];
+    }
+    return NULL;
+}
+
+static void clear_sticky_key(struct active_sticky_key *sticky_key) {
+    sticky_key->position = ZMK_BHV_STICKY_KEY_POSITION_NOT_USED;
```
petejohanson, line 74: Any reason not to just use `memset` to clear this out entirely, and then set the position value?

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,274 @@
...
+    }
+    if (!sticky_key->timer_is_cancelled) {
+        release_sticky_key_behavior(sticky_key, k_uptime_get());
+    }
+    clear_sticky_key(sticky_key);
+}
+
+static int behavior_sticky_key_init(struct device *dev) {
+    static bool init_first_run = true;
+    if (init_first_run) {
```
petejohanson, line 241: This seems to be to avoid issues w/ multiple inits because you have both one shot keys and one shot layers as different behavior driver instances. Is there a reason they need to "share" the sticky key state between both?

petejohanson (commented):
A few review thoughts/questions.

app/src/behaviors/behavior_sticky_key.c
innovaker: ah, I didn't spot that they're local.  In which case they probably don't even need a prefix.  I'm not fussed.

10 seemed like an absolute maximum to me.  We could probably get away with less without any issues though?

petejohanson: As written, this shares the sticky state between sticky keys and layers, which may explain why it is larger. Will wait for @okke-formsma to weigh in.

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,274 @@
...
+};
+
+struct active_sticky_key {
+    s32_t position;
+    s32_t param1;
+    s32_t param2;
+    const struct behavior_sticky_key_config *config;
+    // timer data.
+    bool timer_started;
+    s32_t release_at;
```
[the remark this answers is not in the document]
okke-formsma: actually it should be a s64_t because it's a timestamp.

app/src/behaviors/behavior_sticky_key.c
okke-formsma: `    struct k_delayed_work release_timer;` needs to be setup and I think it's not supposed to be cleared.

app/src/behaviors/behavior_sticky_key.c
okke-formsma: * this is to circument issues in `sticky_key_keycode_state_changed_listener`, where there is no access to the behaviors' data. (At least, I haven't found a way).
* Without the global `active_sticky_keys`, the 'data' struct needs to be pased to many methods.
* It also saves a little memory by having a single `active_sticky_keys` array that is shared by all instances. (it's btter to have a single big pool than multiple small pools)

okke-formsma:
Repushed with updated type signatures. Also renamed the array with active sticky keys to `active_sticky_keys` from `active_sticky_key`

okke-formsma:
@petejohanson I've addressed all cancerns, can you have a look some time?

app/src/behaviors/behavior_sticky_key.c
petejohanson: This works, and explanation makes total sense. Thanks!

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,272 @@
...
+        return 0;
+    }
+    struct keycode_state_changed *ev = cast_keycode_state_changed(eh);
+    for (int i = 0; i < ZMK_BHV_STICKY_KEY_MAX_HELD; i++) {
+        struct active_sticky_key *sticky_key = &active_sticky_keys[i];
+        if (sticky_key->position == ZMK_BHV_STICKY_KEY_POSITION_NOT_USED) {
+            continue;
+        }
+
+        if (strcmp(sticky_key->config->behavior.behavior_dev, "KEY_PRESS") == 0 &&
```
petejohanson, line 182: Hmm... I don't love keying off this label, but this might be the pragmatic way to do this. Any other options you'd considered?

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+    init_first_run = false;
+    return 0;
+}
+
+struct behavior_sticky_key_data {};
+static struct behavior_sticky_key_data behavior_sticky_key_data;
+
+#define _TRANSFORM_ENTRY(idx, node)                                                                \
```
petejohanson, line 253: Do we need to extra this somewhere? Seems like this is duplicated from the keymap code.

petejohanson (commented):
Few more minor follow up.s

app/src/behaviors/behavior_sticky_key.c
okke-formsma: Previously I had this keyed on key_position, but that requires `keycode_state_changed` to gain a `position` field too.

okke-formsma: I think that's fine, because the `zmk_behavior_binding_event` already has the position, so it can just be passed on and used as a unique identifier.

app/src/behaviors/behavior_sticky_key.c
okke-formsma: yeah it's copied in hold taps too. Mind if we commit this as-is and refactor the three common usages out in a separate PR?

petejohanson: Fine by me.

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -0,0 +1,272 @@
...
+#include <zmk/endpoints.h>
+#include <zmk/event-manager.h>
+#include <zmk/events/position-state-changed.h>
+#include <zmk/events/keycode-state-changed.h>
+#include <zmk/events/modifiers-state-changed.h>
+#include <zmk/hid.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#if DT_NODE_EXISTS(DT_DRV_INST(0))
```
petejohanson, line 24: This should really be:

`#if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)`

app/src/behaviors/behavior_sticky_key.c
petejohanson: Let's leave it then, and consider circling back on this later.

petejohanson (approved):
Just one minor comment, not worth holding this up.

Issue #38, opened 2020-06-24
Community Funded USB VID


innovaker:
From: https://discord.com/channels/719497620560543766/719565084208398406/722950144177209395

jpconstantineau:
Perhaps consider getting one here: https://github.com/openmoko/openmoko-usb-oui

petejohanson:
@jpconstantineau Nice! I hadn't seen this project. I will check it out.

petejohanson:
Request submitted there: https://github.com/openmoko/openmoko-usb-oui/pull/15

peterjc:
```
Vendor ID | Product ID | Description
...
0x1d50 | 0x615e | [https://zmkfirmware.dev ZMK Firmware]
```

Nice 🚀

PR #285, opened 2020-10-18
Use official USB product ID

petejohanson:
* Use openmoko product ID from:
  https://github.com/openmoko/openmoko-usb-oui/pull/15


petejohanson:
Closes #38

PR #292, opened 2020-10-21
Add zmk_kscan_gpio_demux kscan driver & NIBBLE keyboard

jaygreco:
Added NIBBLE keyboard folder and keymap. The biggest change is the addition of a new kscan driver, zmk_kscan_gpio_demux, along with some driver cleanup while porting from the standard matrix driver. I also added some comments to the driver.

The demux chip is active-low ~~and the driver allows for both row2col and col2row configurations. Currently, polling is the only method that has been tested as working, but hooks for interrupt-based scanning were left in as well for the future.~~

Tested locally on a NIBBLE keyboard with nice!nano MCU.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ jobs:
           - settings_reset
           - quefrency_left
           - quefrency_right
+          - nibble
         include:
           - board: proton_c
             shield: clueboard_california
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,7 +2,5 @@
   "files.associations": {
     "*.overlay": "dts",
     "*.keymap": "dts"
-  },
-  "C_Cpp.dimInactiveRegions": false,
-  "C_Cpp.errorSquiggles": "Disabled"
+  }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,7 @@
   "files.associations": {
     "*.overlay": "dts",
     "*.keymap": "dts"
-  }
+  },
+  "C_Cpp.dimInactiveRegions": false,
+  "C_Cpp.errorSquiggles": "Disabled"
 }
--- a/app/boards/shields/nibble/Kconfig.defconfig
+++ b/app/boards/shields/nibble/Kconfig.defconfig
@@ -1,0 +1,13 @@
+# Copyright (c) 2020 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+if SHIELD_NIBBLE
+
+config ZMK_KEYBOARD_NAME
+	default "NIBBLE_BLE"
+
+config ZMK_USB
+	default y
+
+endif
+
--- a/app/boards/shields/nibble/Kconfig.defconfig
+++ b/app/boards/shields/nibble/Kconfig.defconfig
@@ -4,7 +4,7 @@
 if SHIELD_NIBBLE
 
 config ZMK_KEYBOARD_NAME
-	default "NIBBLE_BLE"
+	default "NIBBLE"
 
 config ZMK_USB
 	default y
--- a/app/boards/shields/nibble/nibble.overlay
+++ b/app/boards/shields/nibble/nibble.overlay
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <dt-bindings/zmk/matrix-transform.h>
+
+/ {
+	chosen {
+		zmk,kscan = &kscan0;
+		zmk,matrix_transform = &default_transform;
+	};
+
+	kscan0: kscan {
+		compatible = "zmk,kscan-gpio-demux";
+		label = "KSCAN";
+		polling-interval-msec = <25>;
+		input-gpios
+			= <&pro_micro_d 15 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+			, <&pro_micro_d 14 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+			, <&pro_micro_d 16 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+			, <&pro_micro_d 10 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+			, <&pro_micro_d 4 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+			;
+		output-gpios
+			= <&pro_micro_a 3 GPIO_ACTIVE_HIGH>
+			, <&pro_micro_a 2 GPIO_ACTIVE_HIGH>
+			, <&pro_micro_a 1 GPIO_ACTIVE_HIGH>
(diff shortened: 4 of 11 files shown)


app/boards/shields/nibble/nibble.overlay
```diff
@@ -0,0 +1,51 @@
...
+/ {
+	chosen {
+		zmk,kscan = &kscan0;
+		zmk,matrix_transform = &default_transform;
+	};
+
+	kscan0: kscan {
+		compatible = "zmk,kscan-gpio-demux";
+		label = "KSCAN";
+		diode-direction = "col2row";
```
petejohanson, line 18: Is there really any other option for this other than `col2row`, as this is implemented? Seems like having the other direction would require tweaks to the read code to account for tweaking the demux stuff as well.

If that's the case, perhaps we shouldn't stick w/ the previous naming from the regular matrix KSCAN driver, and just use `input-gpios` and `output-gpios`?

app/drivers/zephyr/Kconfig
```diff
@@ -13,6 +13,10 @@ config ZMK_KSCAN_DIRECT_POLLING
 	bool "Poll for key event triggers instead of using interrupts on direct wired boards."
     default n
 
+config ZMK_KSCAN_DEMUX_POLLING
```
petejohanson, line 16: It certainly seems to me, that polling is required given an IO expander like this. From some quick digging, it doesn't seem like there's a way to get this to set *all* the columns to active, in order for the input rows to trigger interrupts for any key press, right?

If that's the case, we way want to look at making the polling rate configurable on this, to be sure users can tweak their power usage accordingly, since it won't be as efficient as using interrupts.

.github/workflows/build.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           - settings_reset
           - quefrency_left
           - quefrency_right
+          - nibble
```
petejohanson, line 26: In addition to adding there here, you should add to `docs/static/setup.sh` and `docs/static/setup.ps1` please.

petejohanson (commented):
Fabulous start! One or two minor notes, and them some specific questions on the IO stuff.

Thanks!

app/boards/shields/nibble/nibble.overlay
jaygreco: Great point. I'd left it generic, but now I'm leaning toward taking the diode direction logic out and only supporting the one direction. As you pointed out, the implementation wouldn't actually work in the reverse direction.

.github/workflows/build.yml
jaygreco: Will do.

app/drivers/zephyr/Kconfig
jaygreco: That's exactly right. Only one col can be on at once. A psuedo-interrupt configuration would work, whereby the columns are set at a periodic rate and the result isn't read back, but rather pins are configured as interrupts. I'm not sure that it would buy much in the way of power savings so probably wouldn't be worth the effort. I like the idea of configurable poll rate in msec as part of the overlay. I'm going to refactor to remove the interrupt code and implement the configurable polling rate.

jaygreco:
Fantastic feedback, thanks! I'm going to work on a rewrite over the next few days focusing on removing the unneeded features discussed above and adding the configurable polling rate. I'll push and update the PR when it's ready to go.

jaygreco:
Alright! Made the changes discussed above. In summary:
- removed kscan interrupt mode
- removed diode direction
+ added configurable polling time
* removed unused functions, optimized driver.c with general cleanup
+ added the shield to `docs/static/setup.sh` and `docs/static/setup.ps1`

.vscode/settings.json
```diff
@@ -2,5 +2,7 @@
   "files.associations": {
     "*.overlay": "dts",
     "*.keymap": "dts"
-  }
+  },
+  "C_Cpp.dimInactiveRegions": false,
```
petejohanson, line 6: We need to generally solve the "how to manage settings.json", but for now, can you please revert these changes? Thanks.

app/boards/shields/nibble/Kconfig.defconfig
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2020 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+if SHIELD_NIBBLE
+
+config ZMK_KEYBOARD_NAME
+	default "NIBBLE_BLE"
```
petejohanson, line 7: This can just be:

```suggestion
	default "NIBBLE"
```
for the keyboard name.

app/drivers/zephyr/kscan_gpio_demux.c
```diff
@@ -0,0 +1,270 @@
...
+                return -EINVAL;                                                                    \
+            }                                                                                      \
+            err = gpio_pin_configure(input_devices[i], in_cfg->pin, GPIO_INPUT | in_cfg->flags);   \
+            if (err) {                                                                             \
+                LOG_ERR("Unable to configure pin %d on %s for input", in_cfg->pin, in_cfg->label); \
+                return err;                                                                        \
+            } else {                                                                               \
+                LOG_DBG("Configured pin %d on %s for input", in_cfg->pin, in_cfg->label);          \
+            }                                                                                      \
+            irq_callbacks_##n[i].work = &data->work;                                               \
```
petejohanson, line 215: If we're using timers and polling because of the IO expander, it seems like we can ditch all of the interrupt setup here, and remove the callback handlers being stored, etc.

app/drivers/zephyr/kscan_gpio_demux.c
```diff
@@ -0,0 +1,270 @@
...
+        data->callback = callback;                                                                 \
+        LOG_DBG("Configured GPIO %d", n);                                                          \
+        return 0;                                                                                  \
+    };                                                                                             \
+                                                                                                   \
+    /* KSCAN API enable function */                                                                \
+    static int kscan_gpio_enable_##n(struct device *dev) {                                         \
+        LOG_DBG("KSCAN API enable");                                                               \
+        struct kscan_gpio_data_##n *data = dev->driver_data;                                       \
+        k_timer_start(&data->poll_timer, K_MSEC(POLL_INTERVAL(n)), K_MSEC(POLL_INTERVAL(n)));      \
```
petejohanson, line 182: I believe any running timers can possibly wake Zephyr from deep sleep on nrf52. We don't need to solve in this PR, but we might want a follow up to hook into the sleep state hooks in Zephyr, and disable this timer when we enter a sleep state.

petejohanson (changes requested):
This is *really* close. A few remaining comments.

petejohanson (approved):
Thanks!

petejohanson:
@jaygreco Some merge conflicts since someone else added a board recently too. Can you quickly rebase w/ fixes for the setup scripts, then we'll merge? Thanks!

petejohanson:
@jaygreco Looks like this rebase didn't quite work as intended. Happy to help you on Discord if you need it. Thanks!

petejohanson (approved):
Looks great!

myst729:
Is this possible to support analog multiplexers (e.g. CD74HC4067) that outputs analog values (work with hall effect sensors and magnet switches)?

PR #293, opened 2020-10-23
Add Battery Voltage Divider Driver

Nicell:
This is a big step towards having battery reporting over BLE. This driver implements a voltage divider meant for batteries using the Sensors API. 

Feedback is welcome on the style/quality of my code!


app/boards/arm/bluemicro840/bluemicro840_v1_defconfig
```diff
@@ -10,6 +10,10 @@ CONFIG_ARM_MPU=y
 # enable GPIO
 CONFIG_GPIO=y
 
+CONFIG_ADC=y
+
+CONFIG_NEWLIB_LIBC=y
```
petejohanson, line 15: Can you remind me why this is needed?

app/drivers/zephyr/battery_voltage_divider.c
```diff
@@ -0,0 +1,209 @@
...
+};
+
+static uint8_t lithium_ion_mv_to_pct(int16_t bat_mv) {
+    // Magic function that maps mV to this discharge graph from adafruit:
+    // https://learn.adafruit.com/li-ion-and-lipoly-batteries/voltages
+    return round(106.818 +
+                 (-0.032685 - 106.818) / pow(1 + pow(bat_mv / 3679.35, 58.979), 0.347386));
+}
+
+static int bvd_sample_fetch(struct device *dev, enum sensor_channel chan) {
```
petejohanson, line 56: You should be checking the specific requested channel, and return `-ENOTSUP` if not the battery channel.

petejohanson (commented):
A couple minor comments.

app/boards/arm/bluemicro840/bluemicro840_v1_defconfig
Nicell: To use the math functions [here](https://github.com/zmkfirmware/zmk/pull/293/files#diff-a1f4d756abd82b30b45f6d1d469c6df584597f5b23cb1f922b8247cae4c5b86fR52), newlib is required. See: https://docs.zephyrproject.org/latest/guides/c_library.html

app/drivers/zephyr/battery_voltage_divider.c
```diff
@@ -0,0 +1,214 @@
...
+    struct adc_sequence as;
+    uint16_t adc_raw;
+    uint16_t voltage;
+    uint8_t state_of_charge;
+};
+
+static uint8_t lithium_ion_mv_to_pct(int16_t bat_mv) {
+    // Magic function that maps mV to this discharge graph from adafruit:
+    // https://learn.adafruit.com/li-ion-and-lipoly-batteries/voltages
+    return round(106.818 +
```
aakkds, line 52: I am wondering if it is really worth to have something like this.
Converting a read voltage to a percent is quite complicated as it depends on a lot of factors (the type of battery, the temperature, the humidity, the current drawn, the number of cycles the battery went through, ...).
In order to have something more or less accurate, an actual battery monitor chip would be required.

Introducing a dependency on `CONFIG_NEWLIB_LIBC` to be able to perform some exponential calculation in order to make a mapping to something that will not be accurate seems overkill.

IMHO just reading the voltage (using a device tree defined voltage divider as you did) and considering the discharge to be linear would not be a much worst approximation. For example considering 4.2V to be 100% and 3.4V to be 0%.

app/drivers/zephyr/battery_voltage_divider.c
```diff
@@ -0,0 +1,214 @@
...
+
+static int bvd_init(struct device *dev) {
+    struct bvd_data *drv_data = dev->driver_data;
+    const struct bvd_config *drv_cfg = dev->config_info;
+
+    drv_data->adc = device_get_binding(drv_cfg->io_channel.label);
+
+    if (drv_data->adc == NULL) {
+        LOG_ERR("ADC %s failed to retrieve", drv_cfg->io_channel.label);
+        return -ENOENT;
```
aakkds, line 145: Minor: the convention is usually to return `-ENODEV` when a `device_get_binding` returns `NULL`.
(feel free to ignore :) )

app/drivers/zephyr/battery_voltage_divider.c
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#define DT_DRV_COMPAT zmk_battery_voltage_divider
+
+#include <device.h>
+#include <drivers/kscan.h>
```
aakkds, line 10: This looks not necessary no?

app/drivers/zephyr/battery_voltage_divider.c
```diff
@@ -0,0 +1,214 @@
...
+    drv_data->acc = (struct adc_channel_cfg){
+        .gain = ADC_GAIN_1_5,
+        .reference = ADC_REF_INTERNAL,
+        .acquisition_time = ADC_ACQ_TIME(ADC_ACQ_TIME_MICROSECONDS, 40),
+        .input_positive = SAADC_CH_PSELP_PSELP_AnalogInput0 + drv_cfg->io_channel.channel,
+    };
+
+    drv_data->as.resolution = 12;
+#else
+#error Unsupported ADC
```
aakkds, line 183: So the whole `battery_voltage_divider` is dependend on `ADC_NRFX_SAADC`?
I don't understand why, it sounds like just having an ADC is enough, is there a specific functionality you rely on that I didn't get?

aakkds:
Hey @Nicell !
I took the liberty to join, hope you don't mind :)
That will be cool to have the battery reporting, I just have one "concern" related to the convertion `mV -> %`.
Basically, I don't think doing a complicated mapping to a discharge curve that does not correspond to the physical battery used is worth it.
I also added some minor stuff that I saw, which you can probably ignore ;)

app/drivers/zephyr/battery_voltage_divider.c
Nicell: While I do think that this function is more accurate than a simple linear approximation, I agree that the amount it's more accurate probably isn't worth the newlib dependency. Good point on that. I'll readjust (Probably 4.2 to 3.45 to err on the side of caution for people playing with low %s).

app/drivers/zephyr/battery_voltage_divider.c
Nicell: Good catch. Thanks!

app/drivers/zephyr/battery_voltage_divider.c
Nicell: So the reason for this is that the ADC channel config set up really depends on the chip we're using as well as the voltage divider configuration from my understanding. For now almost everyone is using nrf52 with a very specific voltage divider setup recommended from Nordic (hence the specific gain, reference, acquisition time, etc). The reason it's set up this way is so we can add more else ifs in the future to handle other chips. Possibly making the gain/ref/acquisition options in the DT, but I'm not certain of this, and at this point we're all using the same set up.

Nicell:
@aakkds Thanks for the review! Probably not worth the newlib dependency for the more accurate function. I also addressed the NRFX ADC comment. Let me know if you think there's a better way to generalize this, but from what I've seen it's not that simple, and making a fully configurable ADC channel configuration system may overcomplicate defining the voltage divider for our use case.

app/drivers/zephyr/battery_voltage_divider.c
aakkds: Honestly I think that it should be quite easy to have conditions where the linear approx is actually better. But it is true than during the mostly non-linear part of the discharge, i.e the very beginning/very end, the exponential approx is most likely more accurate.
Sounds good to have 3.45 as a low threshold.

app/drivers/zephyr/battery_voltage_divider.c
aakkds: Thanks for the explanation but I still do not get it:
The voltage divider comes from the device tree and is not dependent on the ADC itself. As for the ADC configuration, this should be generic. If another MCU is used, I would expect that the ADC reading would work out of the box with the configuration you set up. So I would just remove this `#ifdef` pattern.
But I may be missing something, so if you find it simpler/better to keep as is, all good :)

app/drivers/zephyr/battery_voltage_divider.c
```diff
@@ -0,0 +1,218 @@
...
+
+        uint16_t millivolts = val * (uint64_t)drv_cfg->full_ohm / drv_cfg->output_ohm;
+        LOG_DBG("ADC raw %d ~ %d mV => %d mV\n", drv_data->adc_raw, val, millivolts);
+        uint8_t percent = lithium_ion_mv_to_pct(millivolts);
+        LOG_DBG("Percent: %d", percent);
+
+        drv_data->voltage = millivolts;
+        drv_data->state_of_charge = percent;
+    } else {
+        LOG_DBG("Failed to read ADC: %d", rc);
```
aakkds, line 100: While first reading this I thought that it made sense to not return an error so that the power GPIO is still disabled.
This can have a bad side effect for the application calling the `sensor_sample_fetch`.
The call to `sensor_sample_fetch` will most likely succeed due to `rc = gpio_pin_set(drv_data->gpio, drv_cfg->power_gpios.pin, 0);` below, even if `adc_read` failed.

I think it is acceptable as the `LOG_DBG` at least will indicate something is wrong, but be aware that even if the application is doing error checking, this can get undetected.

If you want to take this into account, you could introduce a second variable `rc_2` to do `rc_2 = adc_read(...);`
and in ligne 112 `return (rc_2 != 0 ? rc_2 : rc);`

app/drivers/zephyr/battery_voltage_divider.c
Nicell: Removing the `ifdef `pattern would result in errors for non-nrf ADCs because I reference `SAADC_CH_PSELP_PSELP_AnalogInput0`, which is only for nrfx. This error is more explicit to someone who is trying to use a new MCU (with different ADC reqs).

Along with this, from my understanding other ADCs need different parts of the adc_channel_cfg defined compared to what nrf ADCs need. I'm personally not interested in researching every other MCU Zephyr supports to make a bunch of logic to handle each type. At the moment we only use nrf52, and every single one uses these specific gains, references, and acquisition times. A generic solution would be a nice enhancement to this in the future, but I don't think it's necessary to have this driver added to ZMK.

I personally couldn't find any generic samples that supported any ADC in Zephyr's repo. I believe the reasons for that are the same reasons I haven't created a generic solution. I could definitely be wrong about some of this though. I'm basing it off of what samples I could find and the source code.

aakkds: Oh yeah sorry I missed the `SAADC_CH_PSELP_PSELP_AnalogInput0` and indeed I see that the examples in zephyr repo use MCU specific gain/ref/acqu_time config.

app/boards/arm/bluemicro840/bluemicro840_v1_defconfig
```diff
@@ -10,6 +10,8 @@ CONFIG_ARM_MPU=y
 # enable GPIO
 CONFIG_GPIO=y
 
+CONFIG_ADC=y
```
petejohanson, line 13: Instead of just forcing this this way, you should have a new Kconfig setting for this new driver, and that that driver `depends on ADC`

app/boards/arm/nice_nano/nice_nano_defconfig
```diff
@@ -10,6 +10,8 @@ CONFIG_ARM_MPU=y
 # enable GPIO
 CONFIG_GPIO=y
 
+CONFIG_ADC=y
```
petejohanson, line 13: Ditto.

app/boards/arm/nrfmicro/nrfmicro_13_defconfig
```diff
@@ -10,6 +10,8 @@ CONFIG_ARM_MPU=y
 # enable GPIO
 CONFIG_GPIO=y
 
+CONFIG_ADC=y
```
petejohanson, line 13: Same

app/drivers/zephyr/battery_voltage_divider.c
petejohanson: I agree we should check for both. Probably easier to just have the later code in the disable block set `int rc2 = gpio_pin_set(blah);` and check that `rc2` internally there and bail if it fails.

petejohanson (commented):
Couple minor things.

petejohanson (approved):
Thanks!

PR #297, opened 2020-10-25
Add Reviung41 Shield Definitions

Nicell:
Add definitions for Reviung41 shield from Japan, and it's available in the US from a few vendors including Little Keyboards and Boardsource.xyz.

The board is open source and can be found here: https://github.com/gtips/reviung/tree/master/reviung41


petejohanson (approved):
LGTM! Super clean PR.

Thanks.

PR #236, opened 2020-10-04
feat: HID Usages + Keys Standardisation

innovaker:
- Adds HID Usages for ZMK (#217).
- Standardizes key aliases/codes (#21 - still open for RFC).
- Enhances HID reports to take advantage of the new codes (#234).

### Issues
- Closes #217.
- Closes #21.
- Closes #234.
- Supersedes and closes #246.
- Supersedes and closes #158.
- Supersedes and closes #161.

### Testing
- End-to-end tests covering:
  - all key aliases
  - both the keyboard (keypad) and consumer HID reports
- ZMK on Nordic's nRF52840 Dongle.
  - `kscan-mock` running on dongle systematically pressed and released each key.
- Node scripts on Raspberry Pi (Raspbian Buster)
   - Scripts running on the Pi parsed events from `/dev/hidrawX` and `/dev/input/...` devices and compared incoming data against expected key usage codes.
   - Scripts not included in PR because end-to-end framework isn't generic enough and some of the code is my own.
- Test logs:
  - [e2e-2020-10-04T12-14-37-usb.log](https://github.com/zmkfirmware/zmk/files/5323138/e2e-2020-10-04T12-14-37-usb.log)
  - [e2e-2020-10-04T12-14-37-ble.log](https://github.com/zmkfirmware/zmk/files/5323139/e2e-2020-10-04T12-14-37-ble.log)
- Once HID reports were enhanced, usage codes themselves never caused any errors.
- Untested usage codes should also be viable for the future as everything was developed systematically via scripting.

Issue #218, opened 2020-09-27
docs: Codes (Keys & Controls)


innovaker:
Compliments #21.

PR #301, opened 2020-10-26
docs: Add eslint, prettier and eslint-mdx support

innovaker:
The docs are in need of standard linting and formatting.  Especially before larger changes start coming through which may otherwise cause conflicts.  Research plus some [discussions](https://discord.com/channels/719497620560543766/719544230497878116/767821099835981864) on Discord suggested the best solution is currently `eslint` combined with `prettier`.

- I'm not an expert in this area, so this is just my best (naive) attempt based on official guidelines.
- I've not set any explicit settings or overrides so it's using the defaults which seemed to match our needs, but I welcome feedback or suggestions.
- Prettier changed a fair amount, so I'd be grateful if this PR could go through a critical review (hopefully by more than one person) before the changes are merged.  We may want to change some settings and reformat before we accept the PR.
- I wasn't sure if `eslint-mdx` needed any other settings to work alongside prettier?
- I tested it locally using `npm run start` and `npm run build` - pages seemed fine.
- I'd prefer to get this through soon - hopefully whilst I finish checking/testing #21 - because it's a prerequisite of me pushing out #218 which would be very useful for the reviewers of #236.
- It might be better if we had this in the project root, but that would require creating a separate `package.json` (I didn't have time) so it's a follow-up task for the future.
- Another follow-up task is to integrate `eslint` and `prettier` into the GitHub build checks.  Until that's done, reviewers (and contributors) will need to conscientiously run `eslint` and `prettier` on docs locally before accepting a docs-related PR.  This of course can be done easily by `CLI` or by having the VS Code extensions installed and *activated*.
- I plan to introduce TypeScript support *after* #236 is through because it's not needed yet and would slow down the current critical path.
- `eslint` currently produces the following errors, but I wasn't sure if it's an issue with our code or one of the `eslint` plugins.  Anyone want to pick that up please?
```
.../zmk/docs/src/pages/index.js
  34:20  error  'imageUrl' is missing in props validation     react/prop-types
  34:30  error  'title' is missing in props validation        react/prop-types
  34:37  error  'description' is missing in props validation  react/prop-types

✖ 3 problems (3 errors, 0 warnings)
```


.vscode/settings.json
```diff
@@ -2,5 +2,8 @@
   "files.associations": {
     "*.overlay": "dts",
     "*.keymap": "dts"
+  },
+  "editor.codeActionsOnSave": {
```
petejohanson, line 6: I'm a bit on the fence about turning this on for everyone using VSCode... But this would likely help us avoid issues in PRs from folks.

petejohanson (approved):
I think as a whole, this is great. We should *not* hold this out for long, due to all the formatting changes. I'd like to get it in ASAP. @Nicell any thoughts on the eslint/prettier stuff?

.vscode/settings.json
innovaker: Happy to remove/change it.  But yes, I turned it on for that reason.

Note, I turned it on for *everything*, whereas it can be more gradual.  Perhaps just `eslint` or `prettier` etc.

I should probably have it as a separate commit.

innovaker: From reading around, I'm also getting the impression it could introduce unexpected corrections in its current form.

innovaker: I've removed that setting for now.  It can go in workspace settings until we've experimented and found the optimum approach.

innovaker:
Force pushed to remove the `source.fixAll` setting from `.vscode/settings.json` which I suspect snuck in.
Re-tested cherry picked changes with:
```
npm install
npx prettier --write .
```
(prettier made no additional changes)

Nicell:
@innovaker Thanks for doing this! In terms of the eslint errors, these are valid, but they won't be once TypeScript is involved. Nonetheless, I've fixed them in this commit: https://github.com/Nicell/zmk/commit/303dfc8a610bc0ab855eb51af84b81db446806d5

Also, I looked closer at the CRLF/LF issues I was having, and it seems to be related to the fact that my Git converts these line endings to CRLF, but back to LF when committing. Prettier doesn't understand this and converts them to LF, but when I stage them, the changes disappear as Git handles the conversion and sees there's no change. See: https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings

We could remedy this visual glitch (in VSCode at least) by turning `endOfLine` to `auto` on Prettier. This should play nicer with Windows Git configurations.

innovaker:
Thanks @Nicell, I've included your commit in this PR.

Regarding `CRLF`/`LF`, the original prettier issue is probably worth considering as well as their blog post:
- https://github.com/prettier/prettier/issues/6932
- https://prettier.io/blog/2020/03/21/2.0.0.html

If I'm interpreting it correctly, it's suggesting that the default `LF` is generally the best option and leads to less EOL issues overall.  I'm not sure how that gels with what you experienced though @Nicell.

Any thoughts?

innovaker:
I *suspect* it's because you're using `autoCRLF` for git on Windows.  What's the rationale there?  Are you working with tools that don't support `LF`?

Nicell:
It was probably a setting that got set long ago on this machine where I never ran into problems with it. I'll probably turn it off, but I think it's fair to consider the fact that this issue didn't crop up with the main source code of the firmware.

It would seem to me that EOL on auto is more in line with how the firmware is currently formatted.

innovaker:
Should we not be mandating `LF` as a project though, much as those links described?

petejohanson:
I am strongly inclined to stick w/ auto for line endings for prettier, instead of forcing `LF` for all platforms. I have *attempted* that on previous projects, and it tends to bite back in odd unexpected ways.

I get that the prettier folks have encountered projects where it was a problem, and so changed the default.

@Nicell Do you have bandwidth to test this PR, but with updated prettier config to change the EOL settings, and see how it works there?

Nicell:
@petejohanson @innovaker I just tested with `endOfLine: 'auto'`, and it worked as expected with no LF changes using autoCRLF on Windows.

petejohanson (approved):
LGTM!

PR #303, opened 2020-10-26
feat(kscan_mock): Increase max number of events

innovaker:
This is necessary for testing a large number of events (e.g. every key code) within a single build/pass.

The `u8_t` limitation became apparent during end-to-end testing of #21.

Tested on the Nordic nRF52840 Dongle.


petejohanson (approved):
One liner love!

Issue #223, opened 2020-09-30
Upgrade to Zephyr 2.4.0


petejohanson:
Zephyr 2.4.0 was just released, which includes lots of fixes for the BLE stack, in addition to other changes.

We should explore updating, perhaps once that release has had time to "settle".

- we currently have two patches on top of 2.3.0 that need evaluation:
  - USB send fix, may be unneeded with the semaphore fixes we added.
  - some LD fixes for our native posix tests. I attempted to upstream these, but I think they were reverted. Will need further investigation.

We also have some interesting CMake usage, we'll need to test if change to Zephyr's build impact ours.

benjizhai:
https://github.com/benjizhai/zmk/tree/zephyr-2.4.0

USB mode works but somehow bluetooth does not (connection dropping 0x08). Need to explore further.

HW: nrfmicro_13 on lily58

petejohanson:
@benjizhai Thanks for doing some legwork on this! I would like to get a few more things stable and rolled out before we disrupt things w/ the Zephyr upgrade. I will *definitely* be looking at your branch when we get there though.

PR #306, opened 2020-10-27
Add BAS GATT reporting

### LINKED ISSUE #47: Feature: Proper battery reporting via BLE battery service
ZMK already exposes the standard BLE battery service, but doesn't have any code to actually send values to the connected host. Needs a generic implementation that can be plugged for different keyboards as needed.

Nicell:
This adds BAS GATT reporting using any device binding with the label `BATTERY`. Closes #47.

Notes:

- `bt_gatt_bas_set_battery_level` has been renamed in Zephyr 2.4.0 -> #223 
- `zmk_battery_update` should eventually raise a battery event instead of directly posting to the BAS GATT
- If the board doesn't have a battery reporting device, init will fail and none of the timers will run. Should this be behind a Kconfig value instead?
- I'm thinking that the update function should only update if the battery value has changed since last time. This would for example save from an unnecessary ePaper display refresh in the future.


app/src/battery.c
```diff
@@ -0,0 +1,77 @@
...
+#include <drivers/sensor.h>
+#include <bluetooth/services/bas.h>
+
+#include <logging/log.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct device *battery;
+
+static int zmk_battery_update(struct device *battery) {
```
Nicell, line 19: Should I make this function available to the rest of the application? I guess that's the reason I have the `battery` device passed in to here.

petejohanson: I don't see why we would need to, at least not to begin with. This should be able to be pretty self contained.

app/src/battery.c
```diff
@@ -0,0 +1,77 @@
...
+
+static int zmk_battery_init(struct device *_arg) {
+    battery = device_get_binding("BATTERY");
+
+    if (battery) {
+        LOG_DBG("Found battery reporting device.");
+    } else {
+        LOG_DBG("No battery device labelled BATTERY found.");
+        return -ENODEV;
+    }
```
petejohanson, line 63: We typically don't bother logging or branching for the success case. Usually we'd just do:

```suggestion
    if (battery == NULL) {
        LOG_DBG("No battery device labelled BATTERY found.");
        return -ENODEV;
    }
```

petejohanson (commented):
Two super minor comments.

petejohanson (approved):
Thanks!

Katona:
Hi! Would it be possible to report the battery level of both half in case of split keyboards? Either through the "master" (maybe with pressing some combo) or directly to the host via Bluetooth (if it's possible at all)?

PR #307, opened 2020-10-28
Mount zmk-config to /workspaces/zmk-config in VSCode remote container.

petejohanson:


petejohanson:
@idan @joelspadin Any thoughts?

joelspadin:
Can't say I've ever used that feature before, but it looks reasonable to me. What happens if you don't have that variable defined?

petejohanson:
Tested this last night, and it massively broke if you *don't* have the extra env. variable defined, since docker ends up trying to mount an entry w/ something like `source=,` and it doesn't like that.

Need to consider what fallback path to put there perhaps?

pluskid:
I tested locally and it works when I have the environment variable defined -- I'm able to build my custom shield in my user config directory.

The problem is to deal with the situation where the environment variables is not defined. Since the devcontainer.json does not allow conditioning, maybe we can use something like

```
"source=${localWorkspaceFolder}/../zmk-config,target=/workspaces/zmk-config,type=bind"
```

instead. And ask the user to place their config dir along side the zmk folder and name it as `zmk-config`. So that even if this folder does not exist, docker won't end up seeing an empty string as with `source=,`.

adamcstephens:
I couldn't find a way to set a default for vscode's environment interpolation. @pluskid's suggestion of a standard location may make the most sense then.

petejohanson:
Does it work if there's a non-empty string, but no directory that exists at that path? If so, this could be a workable compromise.

adamcstephens:
Likely docker will just create that directory then. Would need to test, but that's typical docker behavior.

pluskid:
Yeah, I read from [here (2017)](https://stackoverflow.com/questions/44574078/mount-non-existing-host-directory-into-non-root-container) that non-existing mounting point will be initialized by docker (owned by root). Haven't tested either -- I can try it probably tonight.

Another option is to use `source=${localEnv:ZMK_CONFIG_DIR}/config`, and ask the user to specify `ZMK_CONFIG_DIR` as the path to their user config repo, and we manually append `/config` sub-directory. Now if the environment variable is not defined, then docker will try to mount `/config`. But I think this is less preferable if docker creates the directory in the host. Having an arbitrary called `/config` created at the root directory is clearly less preferable than having something called `zmk-config` created besides the existing zmk source dir. And I do not know what would `/config` mean under Windows.

pluskid:
A third option is to use named volume. I'm new to docker, but based on what I read so far, it seems that if you mount a named volume that does not exist, it will be created [in proper location with proper permission (instead of root)](https://docs.docker.com/storage/volumes/#start-a-container-with-a-volume). Then, when the user need to use their own zmk-config, they can use the `docker` command to delete the default volume, and create a new volume with the given name by [binding it to a local folder](https://forums.docker.com/t/how-to-access-docker-volume-data-from-host-machine/88063).

petejohanson:
Closing this in favor of the solution in #340 which we'll merge once it's rebased.

PR #308, opened 2020-10-28
docs(codes): Add (key) codes documentation

innovaker:
- Adds documentation (#218) for standardized HID codes (#21).
- Compliments PR #236.
- Can be previewed here: https://deploy-preview-308--zmk.netlify.app/docs/codes/

### Features / Notes
- Mostly data-driven (React components) to ease maintenance, minimize boilerplating and maximize future potential.
- Data components embedded into mdx pages to ease reorganisation/refactoring and add contextual notes.
- Includes a bespoke table footnotes system that leverages mdx.
- Is statically prerendered as part of the docusaurus build, but runs as React components in the browser.
- This is the 3rd iteration, although it's the first to be publicly released 👍 
  - The 1st was mostly HTML = maintenance nightmare! 👎 
  - The 2nd was a hybrid with some components = uncomfortable / messy 👎 
  - I also experimented with markdown tables but didn't like it, and it was a maintenance nightmare. 👎 👎 

### Issues
- Closes #218.

### Testing
- `eslint . --ext js,jsx,md,mdx`
- `npx prettier --write .`
- `npm run build` and `npm run serve`
- visually checked pages for mistakes and oversights
- checked for console errors
- tested multiple footnotes

### Future
- Upgrade to TypeScript.
- Add realtime filtering features (specification in progress).
- Add expandable detail for each item (specification in progress).
- Test each code on each OS and update documentation data (_community effort_!!!).
- Explore generic/adaptive names for common uses such as `VOL_UP` (after OS testing).
- Explore expanding Codes documentation beyond HID (other behaviors).
- Explore potential benefits of moving data outside of docs (perhaps a different format) for use throughout ZMK systems as a single source of truth.

### Feedback / Reviews
- I welcome ideas or feedback!
- This was a big piece of work with lots of data involved, so I expect there to be bugs/oversights!
  - Please report what you spot so that it can be corrected before it affects everyone else!
- We're still open to discussions around key code names, but there has already been a 5+ week RFC process so please check that first (see #21).
- I've never flexed my React skills outside of my personal projects, so I welcome constructive criticism and technical reviews!

PR #309, opened 2020-10-28
docs: Set netlify node version to lts/*

innovaker:
Prerequisite of #218 and PR #308.

WARNING: UNTESTED.


innovaker:
The `netlify/zmk/deploy-preview` logs suggest this change switches netlify's node from v10.23.0 to v14.15.0.

This should fix the deployment issue experienced by PR #308.

innovaker:
Switching this into Draft mode because we found this:

https://docs.netlify.com/configure-builds/environment-variables/#declare-variables

which might be the more appropriate way of approaching the issue.

What do you think @petejohanson?

petejohanson:
https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript seems to say either way is totally acceptable. The advantage of the `.nvmrc` is it will be honored for local users of `nvm` as well, which seems to be a win.

innovaker:
If you're happy with it, please merge when ready.

PR #312, opened 2020-10-30
feat(docs): Add power profiler

Nicell:
This adds a battery life estimator to the docs website. This is currently working quite well I'd say, but I'd like to get some feedback before I go into the details of cleaning up the code. Try it here: https://deploy-preview-312--zmk.netlify.app/power-profiler

Todo:

- [ ] Document how the calculations are done and some of the methodology
- [ ] Document how to add new boards or features
- [x] Add MIT license headers (as opposed to the rest of the docs)

### DIFF
--- a/docs/src/components/power-estimate.js
+++ b/docs/src/components/power-estimate.js
@@ -1,0 +1,266 @@
+/*
+ * Copyright (c) 2021 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+import React from "react";
+import PropTypes from "prop-types";
+import { displayPower, underglowPower, zmkBase } from "../data/power";
+import "../css/power-estimate.css";
+
+// Average monthly discharge percent
+const lithiumIonMonthlyDischargePercent = 5;
+// Average voltage of a lithium ion battery based of discharge graphs
+const lithiumIonAverageVoltage = 3.8;
+// Average discharge efficiency of li-ion https://en.wikipedia.org/wiki/Lithium-ion_battery
+const lithiumIonDischargeEfficiency = 0.85;
+// Range of the discharge efficiency
+const lithiumIonDischargeEfficiencyRange = 0.05;
+
+// Proportion of time spent typing (keys being pressed down and scanning). Estimated to 2%.
+const timeSpentTyping = 0.02;
+
+// Nordic power profiler kit accuracy
+const measurementAccuracy = 0.2;
+
+const batVolt = lithiumIonAverageVoltage;
+
+const palette = [
+  "#bbdefb",
+  "#90caf9",
+  "#64b5f6",
+  "#42a5f5",
+  "#2196f3",
+  "#1e88e5",
+  "#1976d2",
+];
+
+function formatUsage(microWatts) {
+  if (microWatts > 1000) {
+    return (microWatts / 1000).toFixed(1) + "mW";
+  }
+
+  return Math.round(microWatts) + "µW";
+}
+
+function voltageEquivalentCalc(powerSupply) {
+  if (powerSupply.type === "LDO") {
+    return batVolt;
+  } else if (powerSupply.type === "SWITCHING") {
+    return powerSupply.outputVoltage / powerSupply.efficiency;
+  }
+}
+
+function formatMinutes(minutes, precision, floor) {
+  let message = "";
+  let count = 0;
+
+  let units = ["year", "month", "week", "day", "hour", "minute"];
+  let multiples = [60 * 24 * 365, 60 * 24 * 30, 60 * 24 * 7, 60 * 24, 60, 1];
+
+  for (let i = 0; i < units.length; i++) {
+    if (minutes >= multiples[i]) {
+      const timeCount = floor
+        ? Math.floor(minutes / multiples[i])
+        : Math.ceil(minutes / multiples[i]);
+      minutes -= timeCount * multiples[i];
+      count++;
+      message +=
+        timeCount + (timeCount > 1 ? ` ${units[i]}s ` : ` ${units[i]} `);
+    }
+
+    if (count == precision) return message;
+  }
+
+  return message || "0 minutes";
+}
+
+function PowerEstimate({
+  board,
+  splitType,
+  batteryMilliAh,
+  usage,
+  underglow,
+  display,
+}) {
+  if (!board || !board.powerSupply.type || !batteryMilliAh) {
+    return (
+      <div className="powerEstimate">
+        <h3>
+          <span>{splitType !== "standalone" ? splitType + ": " : " "}...</span>
+        </h3>
+        <div className="powerEstimateBar">
+          <div
+            className="powerEstimateBarSection"
+            style={{
+              width: "100%",
+              background: "#e0e0e0",
+              mixBlendMode: "overlay",
+            }}
+          ></div>
+        </div>
+      </div>
+    );
+  }
+
+  const powerUsage = [];
+  let totalUsage = 0;
+
+  const voltageEquivalent = voltageEquivalentCalc(board.powerSupply);
+
+  // Lithium ion self discharge
+  const lithiumMonthlyDischargemAh =
+    parseInt(batteryMilliAh) * (lithiumIonMonthlyDischargePercent / 100);
+  const lithiumDischargeMicroA = (lithiumMonthlyDischargemAh * 1000) / 30 / 24;
+  const lithiumDischargeMicroW = lithiumDischargeMicroA * batVolt;
+
+  totalUsage += lithiumDischargeMicroW;
+  powerUsage.push({
+    title: "Battery Self Discharge",
+    usage: lithiumDischargeMicroW,
+  });
+
+  // Quiescent current
+  const quiescentMicroATotal =
+    parseInt(board.powerSupply.quiescentMicroA) +
+    parseInt(board.otherQuiescentMicroA);
+  const quiescentMicroW = quiescentMicroATotal * voltageEquivalent;
+
+  totalUsage += quiescentMicroW;
+  powerUsage.push({
+    title: "Board Quiescent Usage",
+    usage: quiescentMicroW,
+  });
+
+  // ZMK overall usage
+  const zmkMicroA =
+    zmkBase[splitType].idle +
+    (splitType !== "peripheral" ? zmkBase.hostConnection * usage.bondedQty : 0);
+
+  const zmkMicroW = zmkMicroA * voltageEquivalent;
+  const zmkUsage = zmkMicroW * (1 - usage.percentAsleep);
+
+  totalUsage += zmkUsage;
+  powerUsage.push({
+    title: "ZMK Base Usage",
+    usage: zmkUsage,
+  });
+
+  // ZMK typing usage
+  const zmkTypingMicroA = zmkBase[splitType].typing * timeSpentTyping;
+
+  const zmkTypingMicroW = zmkTypingMicroA * voltageEquivalent;
+  const zmkTypingUsage = zmkTypingMicroW * (1 - usage.percentAsleep);
+
+  totalUsage += zmkTypingUsage;
+  powerUsage.push({
+    title: "ZMK Typing Usage",
+    usage: zmkTypingUsage,
+  });
+
+  if (underglow.glowEnabled) {
+    const underglowAverageLedMicroA =
+      underglow.glowBrightness *
+        (underglowPower.ledOn - underglowPower.ledOff) +
+      underglowPower.ledOff;
+
+    const underglowMicroA =
+      underglowPower.firmware +
+      underglow.glowQuantity * underglowAverageLedMicroA;
+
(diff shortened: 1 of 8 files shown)


innovaker:
@Nicell, it looks like 2871dca9bfce2588e23da0905d7142ef5ff1c1a1 has affected this PR and GitHub isn't handling it gracefully.  Please could you rebase onto `HEAD` before your next push.  Thanks!

joric:
I need uW/uA switch, I'm way too used to uA rather than to uW.

innovaker:
It's looking good @Nicell!

Technical reviews aside, my primary concern with a "ZMK-branded" power profiler will be its accuracy.  What margin of error are we talking here?  What data is it based on?  Can the estimated margins of error be represented visually?

Even with the disclaimer, having large margins of error will unfortunately lead to bad mouthing about ZMK.

Nicell:
> Technical reviews aside, my primary concern with a "ZMK-branded" power profiler will be its accuracy. What margin of error are we talking here? What data is it based on? Can the estimated margins of error be represented visually?
> 
> Even with the disclaimer, having large margins of error will unfortunately lead to bad mouthing about ZMK.
@innovaker 

Margin of error of the calculation should be quite small when thinking about it. Most error will come from how accurate the data is that the user inputs. Maybe we can quantify that? Especially the battery mAh.

Nicell:
@innovaker @joelspadin I've finished up this PR finally and would appreciate some feedback on it! For some reason netlify is failing to deploy it (says there's no changes??), so you'll have to test it locally. Thanks 😄

docs/src/data/power.js
```diff
@@ -0,0 +1,55 @@
+export const zmkBase = {
+  hostConnection: 10,
+  standalone: {
+    idle: 0,
+    typing: 320,
+  },
+  central: {
+    idle: 500, // need to test
```
joelspadin, line 8: Do these still need testing?

docs/src/pages/power-profiler.js
```diff
@@ -0,0 +1,357 @@
...
+                    />
+                    <span>mAh</span>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+
+          {board === "custom" && (
+            <div className="profilerSection">
```
joelspadin, line 108: Consider pulling this massive block out into a separate component and using something like `useContext()` to share its state back to this component.

docs/src/pages/power-profiler.js
```diff
@@ -0,0 +1,387 @@
...
+  const { value: glowEnabled, bind: bindGlowEnabled } = useInput(false);
+  const { value: glowQuantity, bind: bindGlowQuantity } = useInput(10);
+  const { value: glowBrightness, bind: bindGlowBrightness } = useInput(1);
+
+  const { value: displayEnabled, bind: bindDisplayEnabled } = useInput(false);
+  const { value: displayType, bind: bindDisplayType } = useInput("");
+
+  const [disclaimerAcknowledged, setDisclaimerAcknowledged] = useState(
+    typeof window !== "undefined"
+      ? localStorage.getItem("zmkPowerProfilerDisclaimer") === "true"
```
innovaker, line 33: Nice touch 👍

docs/src/pages/power-profiler.js
```diff
@@ -0,0 +1,387 @@
...
+      : false
+  );
+
+  const currentBoard =
+    board === "custom"
+      ? {
+          powerSupply: {
+            type: psuType,
+            outputVoltage: outputV,
```
innovaker, line 42: Based on other the naming conventions, presumably this is in volts rather than milli-volts or micro-volts?

docs/src/pages/power-profiler.js
```diff
@@ -0,0 +1,387 @@
...
+  return (
+    <Layout
+      title={`ZMK Power Profiler`}
+      description="Estimate your keyboard's power usage and battery life on ZMK."
+    >
+      <header className={classnames("hero hero--primary", styles.heroBanner)}>
+        <div className="container">
+          <h1 className="hero__title">ZMK Power Profiler</h1>
+          <p className="hero__subtitle">
+            {"Estimate your keyboard's power usage and battery life on ZMK."}
```
innovaker, line 59: See above.

docs/src/pages/power-profiler.js
```diff
@@ -0,0 +1,387 @@
...
+
+  return (
+    <Layout
+      title={`ZMK Power Profiler`}
+      description="Estimate your keyboard's power usage and battery life on ZMK."
+    >
+      <header className={classnames("hero hero--primary", styles.heroBanner)}>
+        <div className="container">
+          <h1 className="hero__title">ZMK Power Profiler</h1>
+          <p className="hero__subtitle">
```
innovaker, line 58: what does `hero__` refer tohere?

docs/src/pages/power-profiler.js
innovaker: Agreed, breaking this into subcomponents might help readability - my eyes glossed over on this file.  That said, I appreciate that'll probably increase the boilerplate code somewhat.

docs/src/pages/power-profiler.js
```diff
@@ -0,0 +1,387 @@
...
+        <div className="disclaimerHolder">
+          <div className="disclaimer">
+            <h3>Disclaimer</h3>
+            <p>
+              This profiler makes many assumptions about typing activity,
+              battery characteristics, hardware behavior, and doesn&apos;t
+              account for error of user inputs. While it tries to estimate power
+              usage using real power readings of ZMK, every person will have
+              different results that may be worse or even better than the
+              estimation given here.
```
innovaker, line 370: Am I seeing double?  Refactor?

innovaker (commented):
Nice work!  There's some really neat ideas in here and you've clearly put a lot of effort into it!

Below are notes that came to mind as I read through the code.  I appreciate it's still early days for this feature though.

- "Power profiler" implies it takes readings from your keyboard (to "profile" your power use), much as I believe the Nordic Power Profiler does?  Suggestions:
a) Battery Life Calculator
b) Power Usage Estimator/Guesstimator
c) Juice Projector/Predictor 😃 
or some combination/variation of the above words.

- Headers/licenses please!

- Are you intending to maintain this going forward?  I suspect ZMK is going to evolve somewhat this year which will probably invalidate the profiler's assumptions and data.  We can't expect firmware contributors to keep it in sync.

- It's safe to assume that the concept of "split" will be generalized to more than two nodes later this year.  It's probably worth keeping that in the back of your mind for design choices.

- Even with the big fat disclaimer, I still have concerns about introducing this feature without implementing a clear feedback loop for data from actual keyboards to verify and improve the model.  But I won't stand in the way of it going live.

- Does the page adapt to mobile devices well?  I'd imagine this particular page will be access from phones a lot.

docs/src/components/power-estimate.js
```diff
@@ -0,0 +1,242 @@
...
+
+const palette = [
+  "#bbdefb",
+  "#90caf9",
+  "#64b5f6",
+  "#42a5f5",
+  "#2196f3",
+  "#1e88e5",
+  "#1976d2",
+];
```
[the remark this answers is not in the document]
Nicell: Due to how they're used, I feel they're better this way. Using them as classes would be pretty cumbersome

docs/src/components/power-estimate.js
```diff
@@ -0,0 +1,242 @@
...
+  underglow,
+  display,
+}) {
+  if (!board || !board.powerSupply.type || !batterymAh) {
+    return (
+      <div className="powerEstimate">
+        <h3>... {splitType !== "standalone" ? "on " + splitType : ""}</h3>
+        <div className="powerEstimateBar">
+          <div
+            className="powerEstimateBarEl"
```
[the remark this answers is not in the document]
Nicell: Yes, It means element. I can rename some that are a bit to vague.

docs/src/components/power-estimate.js
```diff
@@ -0,0 +1,242 @@
...
+              width: (p.usage / totalUsage) * 100 + "%",
+              background: palette[i],
+            }}
+          >
+            <div className="powerEstimateTooltipWrap">
+              <div className="powerEstimateTooltip">
+                <div>
+                  {p.title} - {Math.round((p.usage / totalUsage) * 100) + "%"}
+                </div>
+                <div style={{ fontSize: "14px" }}>
```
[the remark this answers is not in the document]
Nicell: Depends. I was a bit lazy here and should be using `rem`s instead.

docs/src/data/power.js
```diff
@@ -0,0 +1,55 @@
...
+    idle: 490,
+    typing: 380,
+  },
+  peripheral: {
+    idle: 20,
+    typing: 365,
+  },
+};
+
+export const zmkBoards = {
```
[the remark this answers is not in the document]
Nicell: Ideally we can have metadata files for each board. For now this will be held here.

docs/src/pages/power-profiler.js
Nicell: Docusaurus's class name. Reused from the home page.

Nicell:
> * "Power profiler" implies it takes readings from your keyboard (to "profile" your power use), much as I believe the Nordic Power Profiler does?  Suggestions:
>   EDIT: having now used Nordic's Online Power Profiler - I now appreciate where you got the name from.

Does this mean we can stick to this name? Your other suggestions just feel too long IMO. Thoughts?

> * Are you intending to maintain this going forward?  I suspect ZMK is going to evolve somewhat this year which will probably invalidate the assumptions and data.  We can't expect firmware contributors to keep it in sync.

Yes, I'll definitely plan to keep this maintained. Ideally I'll be updating the numbers on a schedule by retesting all of the values.

> * It's safe to assume that the concept of "split" will be generalized to more than two nodes later this year.  It's probably worth keeping that in the back of your mind for design choices.

This is a fair point. It should be pretty easy to extend this to handle that. 

> * Even with the big fat disclaimer, I still have concerns about introducing this feature without implementing a clear feedback loop for data from actual keyboards to verify and improve the model.  But I won't stand in the way of it going live.

I'm not sure if there's much more we can do. If it means anything, of the people I've asked, their battery life has fallen within the range besides a few exceptions (whom I think had some other power stuff going on they weren't accounting for. LEDs, OLEDs, etc).

> * Does the page adapt to mobile devices well?  I'd imagine this particular page will be accessed from phones a lot.
It does! Try it out.

docs/src/components/power-estimate.js
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  const powerUsage = [];
+  let totalUsage = 0;
+
+  const voltageEquivalent = voltageEquivalentCalc(board.powerSupply);
+
+  // Lithium ion self discharge
+  const lithiumDischarge =
+    ((parseInt(batterymAh) * 1000 * lithiumIonMonthlyDischargePercent) /
```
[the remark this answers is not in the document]
Nicell: I've redone how these are all calculated so that the variables are much more direct and found the only real calculation I can pull out is the conversion from micro amps to milliwatt usage, which feels superfluous. Hopefully what I've updated to is much more readable (should only need to know ohms law and metric conversion now 😅).

Nicell:
I've addressed nearly everything from the reviews. I should note that the headers I've added are under the MIT license rather than the rest of the docs that has a creative commons license. I chose this, as I feel this work should be under MIT since it's not really directly documentation. If anyone has some opposition, I'm open to persuasion to use the other license instead.

joelspadin (approved):
There are still a few places in here with massive functions that could be split into smaller components, but this seems like a good start to me.

Nicell:
@innovaker @petejohanson, this PR is holding me up on a few projects. Any chance this could get a review soon? I have a feeling we could iterate on this forever, so hopefully we can focus on most iterations happening in future PRs.

markstos:
Cool project. As a USB keyboard user, I'm curious about the power draw of QMK vs ZMK.  I noticed there are USB power meters that could be added in-line between the laptop and the keyboard to measure the power usage over say, 24 hours:

https://www.amazon.com/Klein-Tools-ET920-Capacity-Resistance/dp/B07FMHHS43/

If someone had a keyboard with hot-swap sockets with both QMK and ZMK-programmed chips, it would be possible to check power draw for idling and "typical use" with otherwise identical PCBs and switches.  I'm curious if ZMKs low-power features translate to USB as well. (I suppose I should open a new thread for this...)

Nicell:
@markstos This power profiler is mostly targeted to handle power readings for batteries, but the results should be pretty close to what you'd get off USB. Of course, USB power pull isn't nearly as important as battery pull when wireless. Sure ZMK may idle at only 20uA higher than quiescent, but even 10mA will barely make a difference when connected over USB. A lot of this also comes down to the MCU of choice. The same hardware setup on QMK and ZMK *probably* won't have very different power usage.

petejohanson (approved):
I think this is a totally acceptable starting point for this. It's definitely used by folks already, and the numbers have been reported to be "on target", which combined w/ the disclaimer, is enough for me. Thanks @Nicell !

PR #316, opened 2020-10-31
fix(boards): Correct adafruit boot partition size

Nicell:


Nicell:
Closing in favor of #322

PR #318, opened 2020-10-31
Fix connection to incorrect peripheral

xudongzheng:
I believe the check should be `if (bt_uuid_cmp(...))` since `bt_uuid_cmp()` returns "true" when the UUID is different. Allocating `struct bt_uuid_128` ensures that memory is correctly allocated for the custom UUID.

### DIFF
--- a/app/src/split/bluetooth/central.c
+++ b/app/src/split/bluetooth/central.c
@@ -187,19 +187,19 @@ static bool split_central_eir_found(struct bt_data *data, void *user_data) {
 
         for (i = 0; i < data->data_len; i += 16) {
             struct bt_le_conn_param *param;
-            struct bt_uuid uuid;
+            struct bt_uuid_128 uuid;
             int err;
 
-            if (!bt_uuid_create(&uuid, &data->data[i], 16)) {
+            if (!bt_uuid_create(&uuid.uuid, &data->data[i], 16)) {
                 LOG_ERR("Unable to load UUID");
                 continue;
             }
 
-            if (!bt_uuid_cmp(&uuid, BT_UUID_DECLARE_128(ZMK_SPLIT_BT_SERVICE_UUID))) {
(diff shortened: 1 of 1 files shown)


app/src/split/bluetooth/central.c
```diff
@@ -187,19 +187,19 @@ static bool split_central_eir_found(struct bt_data *data, void *user_data) {
...
             int err;
 
-            if (!bt_uuid_create(&uuid, &data->data[i], 16)) {
+            if (!bt_uuid_create(&uuid.uuid, &data->data[i], 16)) {
                 LOG_ERR("Unable to load UUID");
                 continue;
             }
 
-            if (!bt_uuid_cmp(&uuid, BT_UUID_DECLARE_128(ZMK_SPLIT_BT_SERVICE_UUID))) {
+            if (bt_uuid_cmp(&uuid.uuid, BT_UUID_DECLARE_128(ZMK_SPLIT_BT_SERVICE_UUID))) {
```
petejohanson, line 198: I'm really not sure I agree with this. We want to be connecting to this discovered peripheral *only* if we found a UUID advertised the matches the ZMK split service UUID. That means these UUIDs need to be equal, which is why we're negating the result of `bt_uuid_cmp`. Help me see what I'm missing here?

Nicell: See [here](https://docs.zephyrproject.org/apidoc/latest/group__bt__uuid_gafe17513c1f91cb3f3e61648b71620c6b.html). When the two UUIDs match, it returns `0` or false. When you negate this, we now run this code whenever the two UUIDs match, which from what I'm reading is the opposite of what we want. Maybe an explicit `!= 0` (is not the same) would be easier to read?

petejohanson: oh! I'm sorry, I should have read this more carefully. This fix does make sense.

petejohanson (approved):
@xudongzheng Tested here locally, working great. Thanks for the fix!

PR #320, opened 2020-11-01
update doc for new shield in user repo

pluskid:
Update the doc to add some pointers for users to create their own shield in a user repository. Mostly based on my own experience (with a lot of help from ZMK developers in the discord server).

One extra suggestion: Maybe we could have links to two official example repos:

1. One that only contains config files for customizing keymaps based on built-in shields.
2. One that defines new shield in user repo.

Although one could probably find a lot of examples just by github search, it is not clear which one is most up-to-date or working properly as references. So it might be good to have official examples as references.

### DIFF
--- a/docs/docs/dev-guide-new-shield.md
+++ b/docs/docs/dev-guide-new-shield.md
@@ -32,7 +32,7 @@ This guide describes how to add shield to the ZMK main repository. If you are bu
 own prototype or handwired keyboard, it is recommended to use your own user config repository. Follow the
 [user setup guide](./user-setup.md) to create your user config repository first. When following the rest
 of this guide, replace the `app/` directory in the ZMK main repository with the `config/` directory in your
-user config repository. Form example, `app/boards/shields/<keyboard_name>` should now be 
+user config repository. For example, `app/boards/shields/<keyboard_name>` should now be 
 `config/boards/shields/<keyboard_name>`.
 :::
 
--- a/docs/docs/dev-guide-new-shield.md
+++ b/docs/docs/dev-guide-new-shield.md
@@ -75,7 +75,11 @@ that make sense to have different defaults when this shield is used. One main it
 that usually has a new default value set here is the `ZMK_KEYBOARD_NAME` value,
 which controls the display name of the device over USB and BLE.
 
-The updated new default values should always be wrapped inside a conditional on the shield config name defined in the `Kconfig.shield` file. Here's the simplest example file. Note: do not make the keyboard name too long, otherwise the bluetooth advertising might fail and you will not be able to find your keyboard from your laptop / tablet.
+The updated new default values should always be wrapped inside a conditional on the shield config name defined in the `Kconfig.shield` file. Here's the simplest example file. 
+
+:::warning
+Do not make the keyboard name too long, otherwise the bluetooth advertising might fail and you will not be able to find your keyboard from your laptop / tablet.
+:::
 
 ```
 if SHIELD_MY_BOARD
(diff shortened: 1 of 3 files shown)


docs/docs/dev-guide-new-shield.md
```diff
@@ -27,6 +27,15 @@ ZMK support for split keyboards requires a few more files than single boards to
 
 ## New Shield Directory
 
+:::note
+This guide describes how to add shield to the ZMK main repository. If you are building firmware for your 
+own prototype or handwired keyboard, it is recommended to use your own user config repository. Follow the
+[user setup guide](./user-setup.md) to create your user config repository first. When following the rest
+of this guide, replace the `app/` directory in the ZMK main repository with the `config/` directory in your
+user config repository. Form example, `app/boards/shields/<keyboard_name>` should now be
```
Nicell, line 35:
```suggestion
user config repository. For example, `app/boards/shields/<keyboard_name>` should now be 
```

docs/docs/user-setup.md
```diff
@@ -33,6 +33,12 @@ Following the steps in this guide, you will:
 1. Update the configuration flags for your user config, to enable/disable any ZMK features you would like to include.
 1. Commit and push the updates to trigger a new GitHub Action run.
 
+:::note
+If you are building firmware for a new keyboard shield that is not included in the built-in
+list of shields, you can choose any shield from the list that is similar to yours to generate the repository, 
+and edit / add necessary files according to the [guide for adding new keyboard shield](./dev-guide-new-shield.md).
+:::
+
```
Nicell, line 41: Would this be better placed right next to the step where a user selects their shield? https://zmkfirmware.dev/docs/user-setup/#keyboard-shield-selection

docs/docs/dev-guide-usb-logging.md
```diff
@@ -50,6 +50,13 @@ CONFIG_UART_CONSOLE_ON_DEV_NAME="CDC_ACM_0"
 CONFIG_USB_UART_DTR_WAIT=n
 ```
 
+:::note
+If you are debugging your own keyboard in your [user config repository](./user-setup.md), use
+`config/boards/shields/<your_keyboard>/<your_keyboard>.conf` instead of `app/prj.conf`. In Github
+Actions, you can search the `Kconfig file` build log to verify the options above have been enabled 
+for you successfully.
+:::
+
```
Nicell, line 59: I think this would be best placed between the code snippet and the description of where to put it. What do you think?

image

Nicell (commented):
These are great clarifications you've added @pluskid! Thanks for this. I've added a couple thoughts. Let me know what you think.

pluskid:
@Nicell Thanks a lot for the feedback! I updated the doc.

docs/docs/dev-guide-new-shield.md
```diff
@@ -66,7 +75,7 @@ that make sense to have different defaults when this shield is used. One main it
 that usually has a new default value set here is the `ZMK_KEYBOARD_NAME` value,
 which controls the display name of the device over USB and BLE.
 
-The updated new default values should always be wrapped inside a conditional on the shield config name defined in the `Kconfig.shield` file. Here's the simplest example file:
+The updated new default values should always be wrapped inside a conditional on the shield config name defined in the `Kconfig.shield` file. Here's the simplest example file. Note: do not make the keyboard name too long, otherwise the bluetooth advertising might fail and you will not be able to find your keyboard from your laptop / tablet.
```
petejohanson, line 78: The note about the name should be in a separate `:::warning` section to call it out a bit more explicitly.

petejohanson (changes requested):
Just one comment. Thanks!

docs/docs/dev-guide-new-shield.md
pluskid: Done, more explicit now! Though it would be better if we have an explicit length limit here, but I do not know what that is...

petejohanson (approved):
Awesome! Thanks for taking the time to improve the docs based on your experiences.

PR #321, opened 2020-11-02
Add Github.com personal token to prerequisites.

evanheckert:
Per @petejohanson 's request, after finding this to be my blocker to a successful push to my github repo. I'm on MacOS, and was getting the error:

` ! [remote rejected] main -> main (refusing to allow an OAuth App to create or update workflow '.github/workflows/build.yml' without 'workflow' scope)`

I already had a personal access token saved in the MacOS Keychain, but since the "workflow" scope hadn't been selected, I needed to delete that one and create a new one with that option. Worked like a charm after that.


petejohanson (approved):
Thanks!

PR #322, opened 2020-11-02
fix(boards): Remap flash to account for bootloader

### LINKED ISSUE #315: Nice nano will periodically get stuck in bootloader, requiring bootloader reflash
Since we introduced BLE profiles, we've had many user experience problems w/ their nice nano getting "stuck" in the bootloader, and after reflashing ZMK, the next power cycle it goes directly to bootloader again. Reflashing the bootloader w/ adafruit-nrfutil seems to be the only fix for this.

There seems to be some correlation w/ this issue and heavy BLE profile switching, reported by myself and several other users.

Some guesses:

* We're exhausting flash space for the NVS settings backend. (Explore flash circular buffer?)
* We have timing issues w/ the settings calls, causing settings to write to the wrong location.
* Something else entirely.

Nicell:
Closes #315 🎉 

We lose around 50KB of code space with this change, but I think that's a fine tradeoff!

### DIFF
--- a/app/boards/arm/bluemicro840/bluemicro840_v1.dts
+++ b/app/boards/arm/bluemicro840/bluemicro840_v1.dts
@@ -82,27 +82,32 @@
 		#address-cells = <1>;
 		#size-cells = <1>;
 
-		boot_partition: partition@0 {
-			label = "adafruit_boot";
-			reg = <0x000000000 0x0000C000>;
+		sd_partition: partition@0 {
+			label = "softdevice";
+			reg = <0x00000000 0x00026000>;
 		};
 		code_partition: partition@26000 {
 			label = "code_partition";
-			reg = <0x00026000 0x000d2000>;
+			reg = <0x00026000 0x000c6000>;
 		};
 
 		/*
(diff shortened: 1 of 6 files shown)


petejohanson:
Tested working fine on my Kyria. If we can get one or two more successful tests in, we should get this in ASAP.

adamcstephens:
I got this working on my board eventually and now everything is :+1: Paired and switching profiles with linux, mac and ipad.

It took me a few attempts, with some trips back to main, to verify this though. My mac would not pair the first two times, but on the third attempt it took.

PippinW:
I am experiencing this issue (#315) on BlueMicro840 w/ Sofle v1.  I am also a frequent BLE profile switcher (3 profiles).
The issue is limited to LHS, with default left-handedness. Power-cycling goes straight to bootloader. 
RHS behaves normally.

I suspect this fix will also work for other arms given the underlying cause and this fix.  Just reporting for the record that this issue occurs beyond nice!nano arms. 
I wIll report back to confirm if this works on BM840.

PR #323, opened 2020-11-02
feat(underglow): Debounce state settings save to reduce flash wear

Nicell:
This adds debounce functionality of one minute to the settings save function. This will greatly reduce how often settings are saved when adjusting hue, saturation, brightness, and effects, which will ultimately reduce flash wear.


petejohanson (approved):
LGTM! Thanks!

PR #340, opened 2020-11-07
zmk-config mounting based on docker volume

pluskid:
The `zmk-config` mounting mechanism implemented in #307 has bad default behavior that makes docker unhappy when the user did not specify an environment variable for custom `zmk-config`.

This implementation uses docker volume. When user did not create a custom zmk-config volume, docker will create a default one and mount correctly. The customization setup is more complicated (involves deleting the default volume), but the default behavior is much well defined.

I tested it locally on MacOS.


petejohanson (approved):
Perfect! Much nicer solution.

petejohanson:
@pluskid Can you please rebase onto latest `main`, then we can merge this? Thanks.

pluskid:
rebased

PR #341, opened 2020-11-07
docs(codes): add modifier functions

### LINKED ISSUE #330: docs: Document modifiers
We don't have docs for shifted keycodes right now. I'd appreciate it if someone can write it down in a nice way. 

Some pointers;

* LS(A) outputs a shifted a (A)
* LC(A) outputs left ctrl+a
* LG(A) outputs gui+a
* LA(A) outputs alt+a
* RS, RC, RG, RA are the right-side equivalents of the L* modifiers.
* you can combine modifiers like this LA(LC(A)) outputs alt+ctrl+a.
* you can roll-over these modifiers without problems, for example if you press `LS(A)` and then `B`, the shift will be released when the B is pressed, and it will output Ab.
* you can use the modified keycodes anywhere you would use a normal keycode.
* maybe an advanced section can explain how shifted keycodes for keys like AT_SIGN are implemented (e.g. LS(NUMBER_1))

innovaker:
Refactor and expand codes documentation to include modifier functions.

Closes #330.

### DIFF
--- a/docs/docs/codes/_keyboard-keypad.mdx
+++ b/docs/docs/codes/_keyboard-keypad.mdx
@@ -24,6 +24,8 @@ import Table from "@site/src/components/codes/Table";
 
 ### Modifiers
 
+The [Modifiers](./modifiers) page includes further information.
+
 <Table group="keyboard-modifiers" />
 
 ### Locks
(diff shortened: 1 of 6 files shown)


innovaker:
@okke-formsma, please review.

PR #344, opened 2020-11-08
feat(docs): add keymap upgrader

### LINKED ISSUE #299: feat: Develop keymap migration script for standardized keys
One of the consequences of #21 is that existing keymaps - both user and default - will subsequently contain deprecated key names.  Although we're planning to gradually phase them out over time, it would be helpful if there was a script to do the grunt work for each developer/user.  The *initial* use case being: user runs script, checks the replacements, adjusts any cosmetics (as necessary), and tests it.  Updates to any contributed keymaps can then be PR-ed with relative confidence.

### Specification
- Must be multi-platform (i.e. on the web or a multi-platform script).
- Must replace deprecated key names in a keymap.
- Must not replace or damage anything else!

### Substitutions
This is a list of the current default replacements.  The list's specifics may be updated as #21 goes through its final review cycles, but the general form should stay the same.
```js
[
  { from: "NUM_1", to: "N1" },
  { from: "NUM_2", to: "N2" },
  { from: "NUM_3", to: "N3" },
  { from: "NUM_4", to: "N4" },
  { from: "NUM_5", to: "N5" },
  { from: "NUM_6", to: "N6" },
  { from: "NUM_7", to: "N7" },
  { from: "NUM_8", to: "N8" },
  { from: "NUM_9", to: "N9" },
  { from: "NUM_0", to: "N0" },
  { from: "BKSP", to: "BSPC" },
  { from: "SPC", to: "SPACE" },
  { from: "EQL", to: "EQUAL" },
  { from: "TILD", to: "TILDE" },
  { from: "SCLN", to: "SEMI" },
  { from: "QUOT", to: "SQT" },
  { from: "GRAV", to: "GRAVE" },
  { from: "CMMA", to: "COMMA" },
  { from: "PRSC", to: "PSCRN" },
  { from: "SCLK", to: "SLCK" },
  { from: "PAUS", to: "PAUSE_BREAK" },
  { from: "PGUP", to: "PG_UP" },
  { from: "PGDN", to: "PG_DN" },
  { from: "RARW", to: "RIGHT" },
  { from: "LARW", to: "LEFT" },
  { from: "DARW", to: "DOWN" },
  { from: "UARW", to: "UP" },
  { from: "KDIV", to: "KP_DIVIDE" },
  { from: "KMLT", to: "KP_MULTIPLY" },
  { from: "KMIN", to: "KP_MINUS" },
  { from: "KPLS", to: "KP_PLUS" },
  { from: "UNDO", to: "K_UNDO" },
  { from: "CUT", to: "K_CUT" },
  { from: "COPY", to: "K_COPY" },
  { from: "PSTE", to: "K_PASTE" },
  { from: "VOLU", to: "K_VOL_UP" },
  { from: "VOLD", to: "K_VOL_DN" },
  { from: "CURU", to: null },
  { from: "LPRN", to: "LPAR" },
  { from: "RPRN", to: "RPAR" },
  { from: "LCUR", to: "LBRC" },
  { from: "RCUR", to: "RBRC" },
  { from: "CRRT", to: "CARET" },
  { from: "PRCT", to: "PRCNT" },
  { from: "LABT", to: "LT" },
  { from: "RABT", to: "GT" },
  { from: "COLN", to: "COLON" },
  { from: "KSPC", to: null },
  { from: "ATSN", to: "AT" },
  { from: "BANG", to: "EXCL" },
  { from: "LCTL", to: "LCTRL" },
  { from: "LSFT", to: "LSHFT" },
  { from: "RCTL", to: "RCTRL" },
  { from: "RSFT", to: "RSHFT" },
  { from: "M_NEXT", to: "C_NEXT" },
  { from: "M_PREV", to: "C_PREV" },
  { from: "M_STOP", to: "C_STOP" },
  { from: "M_EJCT", to: "C_EJECT" },
  { from: "M_PLAY", to: "C_PP" },
  { from: "M_MUTE", to: "C_MUTE" },
  { from: "M_VOLU", to: "C_VOL_UP" },
  { from: "M_VOLD", to: "C_VOL_DN" },
  { from: "GUI", to: "K_CMENU" },
];
```

joelspadin:
Added a documentation page with a script that upgrades deprecated key codes and behaviors to their replacements.

It does not attempt to make sure that it's only updating things in the bindings arrays, as properly parsing DT files is hard. It just uses regex word boundaries to make sure it's only replacing entire words.

Fixes #299


docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,55 @@
...
+
+function upgradeKeycodes(text) {
+  for (const [deprecated, replacement] of Object.entries(Codes)) {
+    if (replacement) {
+      text = replaceWord(text, deprecated, replacement);
+    } else {
+      text = replaceWord(
+        text,
+        `&kp\\s+${deprecated}`,
+        `&none /* Cannot upgrade ${deprecated} */`
```
innovaker, line 31: Perhaps something like:
`/* ${deprecated} no longer exists */`

docs/src/data/keymap-upgrade.js
```diff
@@ -0,0 +1,75 @@
...
+  M_PLAY: "C_PP",
+  M_MUTE: "C_MUTE",
+  M_VOLU: "C_VOL_UP",
+  M_VOLD: "C_VOL_DN",
+  GUI: "K_CMENU",
+};
+
+export const Behaviors = {
+  "&cp": "&kp",
+};
```
innovaker, line 75: Can we add this please?

`"&inc_dec_cp": "&inc_dec_kp"`

Thanks!

docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,55 @@
+import { Codes, Behaviors } from "./data/keymap-upgrade";
+
+// The start of a behvioar like "&kp" isn't a word boundary,
+// so also match following the start of a line or whitespace.
+const WordBegin = "(?:\\b|^|(?<=\\s))";
+const WordEnd = "\\b";
+const UnimportantWhitespace = "( +)(?!\\S)";
```
innovaker, line 7: Should we be using UpperCase?  Any thoughts @Nicell?

docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,55 @@
...
+function replaceWord(text, word, replacement) {
+  // Match the word and all unimportant whitespace following it
+  const pattern = new RegExp(
+    `${WordBegin}(${word})${WordEnd}(${UnimportantWhitespace})?`,
+    "gm"
+  );
+  return text.replace(pattern, (match, word, spaceAfter) => {
+    return getAlignedReplacement(word, replacement, spaceAfter?.length ?? 0);
+  });
+}
```
innovaker, line 48: I noticed it screwed up some of the ascii layouts that people use.  We could probably minimize that by adding rudimentary context wrapping to the RegExp.  Something along the lines of:
```
`(?:bindings\\s?=\\s?<[^>]+)${WordBegin}(${word})${WordEnd}(${UnimportantWhitespace})?(?:[^>]+>)`
```
It should handle `sensor-bindings` also.  What do you think please?

innovaker (changes requested):
Looks great @joelspadin!  Thanks for doing this 🥇 

Much of it is as I'd personally approach it.  Just some minor suggestions.

innovaker:
p.s. I didn't get a chance to actually *test* the builds of any converted keymaps, but hope to give some a go towards the end of the day and will compare before/after with git diff.

docs/src/keymap-upgrade.js
joelspadin: I'd considered doing something like that, but if someone had this in their keymap, it'd break.
```
bindings = <
    /* < = > (whoops, the regex stops here)  */
    &kp LT &kp EQL &kp GT
>;
```
You'd need a smarter parser that's able to recognize comments, and while you can use regexes to build such a parser, regexes aren't suited for that on their own.

docs/src/data/keymap-upgrade.js
joelspadin: Done.

docs/src/keymap-upgrade.js
joelspadin: Done.

docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,55 @@
+import { Codes, Behaviors } from "./data/keymap-upgrade";
+
+// The start of a behvioar like "&kp" isn't a word boundary,
```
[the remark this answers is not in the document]
joelspadin: Whoops. Fixed.

docs/docs/codes/keymap-upgrader.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Keymap Upgrader
+sidebar_label: Keymap Upgrader
+hide_title: true
+---
+
+# Keymap Upgrader
+
+Many codes have been renamed to be more consistent with each other.
+Paste the contents of a .keymap file below to upgrade all deprecated codes to their replacements.
```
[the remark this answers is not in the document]
joelspadin: Done.

joelspadin:
Fixed the review comments and another issue I found where it wouldn't replace `<&cp` with `<&kp` because it didn't see `<&` as a word boundary.

docs/src/keymap-upgrade.js
innovaker: That's true, however:
- most (all?) shields have their ascii comments *above* the `bindings`, not inside them
- comments inside `bindings` containing `>` are likely to be rare
- it's better *not* to convert, than to convert unrelated parts of the file that the user doesn't anticipate or notice changing

Any thoughts?

innovaker:
Thanks for following-up @joelspadin.  My only concern is regarding modifying parts of the file that shouldn't be.  Please see my reasoning.

docs/src/keymap-upgrade.js
innovaker: The other approach to improve context is to include the behavior tags into the regexp, but it's probably more involved due to the numerous behaviors.

docs/src/keymap-upgrade.js
Nicell: I personally don't use all upper case on `const`s in my JS code, but it definitely comes down to preference. I generally don't see many JS projects that use upper case for `const`.

innovaker: Fair enough 👍  I'd not seen PascalCase used for this type of variable before which is why it caught my eye.

docs/src/keymap-upgrade.js
joelspadin-garmin: I see three options here:
1. Do a simple word search and replace, which may convert unrelated parts of the file.
2. Use some naïve regex parsing to try to limit searches to specific parts of the file, which may fail to upgrade things that were supposed to be upgraded.
3. Use a proper parser.

In option 1, I believe the only things it might unexpectedly convert would be comments. Anything else would be seen by the preprocessor as an instance of one of the macros we're trying to upgrade, and thus should be upgraded too.

I personally see option 2 as worse because instead of unexpectedly converting some extra things that shouldn't have any effect on the function of the keymap, it might unexpectedly _not_ convert some things that do affect the function of the keymap.

Option 3 would obviously be the most reliable, but it is a lot harder. I tried that first using https://tree-sitter.github.io/tree-sitter/ but couldn't figure out how to resolve some ambiguities between the syntax for child nodes and properties. I could try porting Zephyr's DT parser to JavaScript instead of writing a parser from scratch.

joelspadin: Whoops. I forgot I was still logged into my work account. If it wasn't already obvious, the above is also me.

innovaker: Yeah, that was my impression from what you've said before.  But I tried it on many files and on balance I found that only stuff inside the bindings usually needed changing, and any comments that were also changed really threw me off.  I'd imagine the vast majority of users will be the same as most maps are rudimentary.

joelspadin: For my own reference later, https://github.com/zephyrproject-rtos/zephyr/blob/master/scripts/dts/dtlib.py is Zephyr's parser. It is not trivial, but a port of that to JS would also be useful for making a VS Code (or similar IDE) extension for autocomplete in keymaps.

joelspadin:
I really hope this version works for other people, because I spent way too long fighting with tree-sitter to get it to load properly.

docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/**
+ * Returns the node for the named capture.
+ * @param {string} name
+ * @param {any[]} captures
+ * @returns {Parser.SyntaxNode | null}
+ */
+function findCapture(name, captures) {
+  for (const c of captures) {
```
joelspadin, line 136: Query API returns something that looks like

```ts
type Result = { 
  captures: { 
    name: string;
    node: Parser.SyntaxNode; 
  }[]; 
}[];
```
All the queries I'm running should just return results with a single capture, so this just finds and returns it.

docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,234 @@
...
+  return null;
+}
+
+/**
+ * Given a parameter to a keymap behavior, returns a list of nodes beginning
+ * with the behavior and including all parameters.
+ * Returns an empty array if no behavior was found.
+ * @param {Parser.SyntaxNode} paramNode
+ */
+function findBehaviorNodes(paramNode) {
```
joelspadin, line 151: The syntax tree for `<&kp A &kp B>` looks like

```C
(integer_cells // <
  (reference (identifier)) // (& (kp))
  (identifier) // (A)
  (reference (identifier)) // (& (kp))
  (identifier) // (B)
) // >
```

Given node "A" or "B", this walks back through siblings to find the previous `&kp` node, then walks forwards to find either the next `&kp` node or the end of the list. The list it returns includes everything from the previous behavior node up to but not including the next behavior node, which is the behavior and all its parameters.

docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,234 @@
...
+  edits = sortEdits(edits);
+
+  console.log(edits);
+
+  /** @type string[] */
+  const chunks = [];
+  let currentIndex = 0;
+
+  for (const edit of edits) {
+    if (edit.startIndex < currentIndex) {
```
joelspadin, line 221: Given something silly like `&cp CURU`, we can both try to change `&cp` to `&kp` and `&cp CURU` to `&none /* "&cp CURU" no longer exists. */`. Those edits overlap so they can't both be applied.

That shouldn't happen in any reasonable keymap so I just made it throw out an arbitrary one of the edits instead of trying to do something smarter to resolve that.

joelspadin:
Added a bunch of comments with context since understanding what the new parser is doing isn't trivial.

Also see https://github.com/joelspadin/tree-sitter-devicetree/tree/main/test/corpus for some examples of how the parser sees things.

joelspadin:
Added a really hacky workaround for the problem where Firefox treats the path to tree-sitter.wasm as relative and Chrome et al. treat it as absolute. I just set up webpack to replace `"tree-sitter.wasm"` with `"/tree-sitter.wasm"` in tree-sitter.js.

docs/src/docusaurus-tree-sitter-plugin/index.js
```diff
@@ -0,0 +1,39 @@
...
+          test: /web-tree-sitter/,
+          loader: "null-loader",
+        });
+      } else {
+        // web-tree-sitter has a hard-coded path to tree-sitter.wasm,
+        // (see https://github.com/tree-sitter/tree-sitter/issues/559)
+        // which some browsers treat as absolute and others as relative.
+        // This breaks everything. Rewrite it to always use an absolute path.
+        rules.push({
+          test: /tree-sitter\.js$/,
```
joelspadin, line 19: I couldn't figure out how to get webpack to match exactly node_modules/web-tree-sitter/tree-sitter.js. Anything more specific I used just made it not do anything.

innovaker (changes requested):
Wow @joelspadin, you took this way further than I expected.  And shockingly quickly!

It LGTM, although I must admit I didn't completely understand the core system works.  I only had one suggestion.

Have you done much A/B testing?  i.e. diff-ing results as well as building them after?  If not, then I'll help you do some before we merge.

When it comes to merging, would you prefer the commits kept as they are, or squashed?

docs/src/keymap-upgrade.js
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Returns a string with text replacements applied.
+ * @param {string} text
+ * @param {TextEdit[]} edits
+ */
+function applyEdits(text, edits) {
+  edits = sortEdits(edits);
+
+  console.log(edits);
+
```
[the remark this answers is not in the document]
joelspadin: Whoops. Yeah.

docs/docs/codes/keymap-upgrader.mdx
```diff
@@ -0,0 +1,18 @@
+---
+title: Keymap Upgrader
+sidebar_label: Keymap Upgrader
+hide_title: true
+---
```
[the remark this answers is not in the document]
joelspadin: Done.

joelspadin:
For some reason I can't reply directly to a couple comments, so

> When it comes to merging, would you prefer the commits kept as they are, or squashed?

Squashing is fine. I really only left it as two commits in case we found major issues with using WASM or something and I needed to revert to the regex method.

> I must admit I don't really understand this bit, but it seems to work and I trust your judgement.

If for example you put `CMMA` outside of the key bindings but still somewhere that the C preprocessor would see it as an identifier, it would have replaced it with the number for that key, which would probably make your keymap either fail to compile or at the very least not do what you intended. After the upgrade, it will be equally broken as before.

> Have you done much A/B testing? i.e. diff-ing results as well as building them after? If not, then I'll help you do some before we merge.

I have not done very extensive testing on this yet. So far I've just run a couple existing keymaps through it and then invented ones to test edge cases like that `&inc_dec_kp A CURU` gets turned into `&none`.

This also seems like it would be a good candidate for unit testing. Maybe we could pull in something like https://mochajs.org/ and add unit tests for the upgrader?

joelspadin:
After running this on a few more keymaps, should `CURU` be upgraded to `DOLLAR` now? Some of the existing keymaps used it and the comments above it showed it was used for `$`.

joelspadin:
Added minor fixes:
* Changed tab size to 4 in the input and output
* Updated parser to allow `&references` as node names (unlikely to affect keymaps)

innovaker:
> For some reason I can't reply directly to a couple comments, so

I've found that also.  Try doing it through the `Files changed` tab.

innovaker:
> After the upgrade, it will be equally broken as before.

Oh I think I understand now.  Well, you know the saying ... if it's ... broken, don't fix it 👍

innovaker:
> I have not done very extensive testing on this yet.

Will endeavor to do some later today.  It might also be worth calling in air support from our loyal testers.

innovaker:
> This also seems like it would be a good candidate for unit testing. Maybe we could pull in something like https://mochajs.org/ and add unit tests for the upgrader?

Can do!  Please don't feel obliged though given that the keymap upgrader will effectively be redundant in a few months (unless we make further code changes in the future).

innovaker:
> After running this on a few more keymaps, should CURU be upgraded to DOLLAR now?  Some of the existing keymaps used it and the comments above it showed it was used for $.

That's an excellent question.  I think we removed it simply because we weren't sure of its value.  But given that most of our initial users probably treated it as `DOLLAR` anyway, that's probably a good idea.  Might be best to use the `DLLR` abbreviation in this context?

innovaker (approved):

innovaker:
@joelspadin, will pass it over to Pete for merging once you're happy and give me the green light.  Feel free to squash it yourself as it's one less thing for him to worry about.

joelspadin:
> Can do! Please don't feel obliged though given that the keymap upgrader will effectively be redundant in a few months (unless we make further code changes in the future).

Fair enough. I'll remap `CURU` to `DLLR` and then this should be ready (once it's been given enough testing). I've run about half the keymaps in this repo through it so far and the `CURU` thing is the only issue I've noticed so far, but I also don't have any of these keyboards to test with.

docs/src/data/keymap-upgrade.js
```diff
@@ -0,0 +1,76 @@
...
+  KMLT: "KP_MULTIPLY",
+  KMIN: "KP_MINUS",
+  KPLS: "KP_PLUS",
+  UNDO: "K_UNDO",
+  CUT: "K_CUT",
+  COPY: "K_COPY",
+  PSTE: "K_PASTE",
+  VOLU: "K_VOL_UP",
+  VOLD: "K_VOL_DN",
+  CURU: null,
```
joelspadin, line 45:
```suggestion
  CURU: DLLR,
```

docs/src/data/keymap-upgrade.js
```diff
@@ -0,0 +1,76 @@
...
+  KMLT: "KP_MULTIPLY",
+  KMIN: "KP_MINUS",
+  KPLS: "KP_PLUS",
+  UNDO: "K_UNDO",
+  CUT: "K_CUT",
+  COPY: "K_COPY",
+  PSTE: "K_PASTE",
+  VOLU: "K_VOL_UP",
+  VOLD: "K_VOL_DN",
+  CURU: DLLR,
```
joelspadin, line 45:
```suggestion
  CURU: "DLLR",
```

joelspadin:
Squashed and updated the page to indicate that it doesn't try to upgrade `#define`s.

innovaker:
@joelspadin, did we get many bites on the [testing](https://discordapp.com/channels/719497620560543766/719544230497878116/778667572945485834) (_sigh_, nobody posted on the issue)?  I lost track, what was the feedback please?

I've been busy myself, but will review #401.

joelspadin:
I think we only had one person who reported testing it and said it worked. Didn't see any other responses.

tominabox1:
Tested on both of my boards and happy to report success on both. Cool feature!

joelspadin:
Rebased and added replacements for things like `MOD_RGUI` -> `RGUI` since there were apparently some keymaps still using that in `&mt`.

innovaker:
Tested thoroughly via #401.

@petejohanson, LGTM.

innovaker:
Awesome job @joelspadin! 👍   Thanks!

PR #346, opened 2020-11-08
feat(underglow): RGB toggle controls ext_power

Nicell:
This adds the ability for RGB underglow toggling to control ext_power to save immensely on battery when RGB is off. It also can be turned off if people want to have more fine tuned control if they have more than RGB on their board for example.

This PR also fixes the error being thrown on release of any underglow behavior and the quick blip of red LED on start when using settings.

### DIFF
--- a/app/src/rgb_underglow.c
+++ b/app/src/rgb_underglow.c
@@ -15,6 +15,7 @@
 #include <logging/log.h>
 
 #include <drivers/led_strip.h>
+#include <drivers/ext_power.h>
 
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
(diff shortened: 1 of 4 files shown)


tominabox1:
Tested on my QAZ, zero voltage when using the RGB_TOG behavior. Nice work. 

Also testing the power on by default config and that seems to be working as expected also.

app/src/rgb_underglow.c
```diff
@@ -299,6 +299,25 @@ int zmk_rgb_underglow_toggle() {
 
     state.on = !state.on;
 
+#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_EXT_POWER)
+    struct device *ext_power = device_get_binding("EXT_POWER");
```
petejohanson, line 303: I would just initialize a static variable in the init once, instead of fetching each time.

petejohanson (changes requested):
Just one minor comment. Thanks!

app/src/rgb_underglow.c
```diff
@@ -244,6 +246,11 @@ static int zmk_rgb_underglow_init(struct device *_arg) {
         return -EINVAL;
     }
 
+    ext_power = device_get_binding("EXT_POWER");
```
petejohanson, line 249: You should change this to be wrapped in an `#if`, same with the actual declaration of the static variable.

petejohanson (changes requested):
One more item. Thanks!

petejohanson (approved):
Perfect! Thanks.

PR #348, opened 2020-11-08
feat: Add external power settings

Nicell:
This PR adds settings support to external power.

### DIFF
--- a/app/src/ext_power_generic.c
+++ b/app/src/ext_power_generic.c
@@ -6,8 +6,11 @@
 
 #define DT_DRV_COMPAT zmk_ext_power_generic
 
+#include <stdio.h>
 #include <device.h>
 #include <init.h>
+#include <kernel.h>
+#include <settings/settings.h>
 #include <drivers/gpio.h>
 #include <drivers/ext_power.h>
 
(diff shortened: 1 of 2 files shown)


app/src/power.c
```diff
@@ -35,6 +36,20 @@ enum power_states sys_pm_policy_next_state(s32_t ticks) {
     s32_t current = k_uptime_get();
     if (power_last_uptime > 0 && !is_usb_power_present() &&
         current - power_last_uptime > MAX_IDLE_MS) {
+
+#if IS_ENABLED(CONFIG_ZMK_EXT_POWER)
```
petejohanson, line 40: Instead of adding this here, look to switch the ext power driver to use `DEVICE_DEFINE` so you can pass a PM control function: https://docs.zephyrproject.org/2.3.0/reference/drivers/index.html#c.DEVICE_DEFINE

petejohanson (changes requested):
Can we split the settings work from the PM work? The PM stuff needs tweaks, but settings looks reasonable.

app/src/ext_power_generic.c
```diff
@@ -27,6 +29,26 @@ struct ext_power_generic_data {
     bool status;
 };
 
+static bool ext_power_generic_state;
+static bool ext_power_generic_state_initalized = false;
+
+#if IS_ENABLED(CONFIG_SETTINGS)
+static void ext_power_save_state_work() {
+    settings_save_one("ext_power/state", &ext_power_generic_state, sizeof(ext_power_generic_state));
```
petejohanson, line 37: Why not `sprintf(setting_name, "ext_power/state/%s", DT_INST_LABEL(0));` to future proof this for multiple?

app/src/ext_power_generic.c
```diff
@@ -27,6 +29,26 @@ struct ext_power_generic_data {
     bool status;
 };
 
+static bool ext_power_generic_state;
```
petejohanson, line 32: Why is this needed? Can't you just use the driver `data->state` for this?

petejohanson (commented):
Couple ideas/questions.

app/src/ext_power_generic.c
Nicell: Thanks for this. It was the missing piece in how I would be able to generalize the solution more. I've updated to use this as well as cut the global state for the driver in favor of using status variables on the driver.

app/src/ext_power_generic.c
```diff
@@ -25,8 +28,33 @@ struct ext_power_generic_config {
...
     bool status;
+#if IS_ENABLED(CONFIG_SETTINGS)
+    bool settings_init;
+#endif
 };
 
+#if IS_ENABLED(CONFIG_SETTINGS)
+static void ext_power_save_state_work(struct k_work *work) {
+    char setting_path[40];
+    struct device *ext_power = device_get_binding(DT_INST_LABEL(0));
```
petejohanson, line 39: This works. In the future, we'll need to tweak this to put the work inside the driver data, then do something like `strict ext_data data* = CONTAINER_OF(work, struct ext_data, work);` (paraphrasing, I'm on my phone), to get the specific instance this work is for.

petejohanson (approved):

PR #356, opened 2020-11-10
docs: update untested iOS keyboard codes

onebigdoor:
update untested iOS keycodes
i'm unsure my method correctly tests consumer codes, so they are excluded
this gist includes the method i used in iOS/xcode to generate my test output, as well as my test output
https://gist.github.com/onebigdoor/5ee09fd5148a925b3b272a1f7435c7b0

attached the keymap i used `keycode.keymap` as `keycode.txt`
[keycode.txt](https://github.com/zmkfirmware/zmk/files/5519529/keycode.txt)


innovaker:
Thanks @onebigdoor!

I've verified your analysis of the log you provide using a script.  Our results (commits) only differed on two codes:
 - `CAPSLOCK` ([`0x07 0x39`])
   - commit: `true`
   - log: `false`
- `K_POWER` ([`0x07 0x66`])
   - commit: `null`
   - log: `false`

Please could you independently verify the above and update your PR if you agree?

Also, please could you answer the following Qs (it's the PR template I've been working on) ...

```
### iOS version/build?

### What codes did you test?

### How did you test?

```

Muchas gracias.

innovaker:
I just noticed `K_POWER` wasn't in the keymap.  My bad.  Feel free to test it anyway.

onebigdoor:
oh, got it.  i'll test `K_POWER` and update

onebigdoor:
i manually tested `K_POWER` and confirmed that it worked, so i've updated that.

innovaker:
Thanks for doing this!  For prosperity, please may I check:
- What was the iOS version?
- Was this run on a device, i.e. an iPhone?  Or emulated?  Please could you give a few more details for those of us unfamiliar with iOS development.

innovaker:
For future observers, this was the contents of the original gist:

### Code
```objective-c
// used this method to print all keypresses to console output

override func pressesBegan(_ presses: Set<UIPress>, with event: UIPressesEvent?) {
        super.pressesBegan(presses, with: event)
        guard let key = presses.first?.key else { return }
        var modifiers = ""
        if (key.modifierFlags.contains(.alphaShift)) {
            modifiers += "caps lock "
        }
        if (key.modifierFlags.contains(.shift)) {
            modifiers += "shift "
        }
        if (key.modifierFlags.contains(.command)) {
            modifiers += "command "
        }
        if (key.modifierFlags.contains(.alternate)) {
            modifiers += "alt "
        }
        if (key.modifierFlags.contains(.control)) {
            modifiers += "control "
        }
        if (key.modifierFlags.contains(.numericPad)) {
            modifiers += "numPad "
        }
        

        print(String(format:"%02X", key.keyCode.rawValue), " ", modifiers)
    }
```

### Log
[log.txt](https://github.com/zmkfirmware/zmk/files/5531249/log.txt)

onebigdoor:
tested on physical device, iPhone 7, running iOS 13.6.1 connected to xcode in debug mode.  the xcode project is essentially blank except for the added `pressesBegan` method logging keypresses to console.

innovaker (approved):
I can confirm that the resulting `diff` matches what I'd expect based on confirmation analysis of the log + comments.

@petejohanson, how would you prefer to play this please?  @onebigdoor's code has effectively run through the Keyboard codes to check for iOS support.  My only concern is whether we could be missing any useful background information/footnotes (i.e. app support), but the same could be said for the existing Linux and Android codes.  Are you happy to merge or should we wait for verification by other iOS testers first?  Thanks!

@petejohanson, whenever we merge, please squash these commits into one and rename appropriately.

innovaker:
Converted to draft as [discussed](https://discordapp.com/channels/719497620560543766/719544230497878116/776473031123795968).  Is on hold for further verification by @onebigdoor.

innovaker:
@petejohanson, changes accepted, though please check https://github.com/zmkfirmware/zmk/pull/356#pullrequestreview-529176430 before merging (if you haven't already seen it).

innovaker:
I'd like to suggest further investigating the `true` options such as `K_MUTE` and `K_PASTE` etc. before merging this PR.  Given what we saw on #377 with `K_POWER`, I'm concerned these could be false positives - i.e. the OS recognises the codes but doesn't act on them.

@onebigdoor, any thoughts on that please?  I wouldn't expect you to do it, we can draw up a checklist for others to verify manually.  We could also consider splitting the PR into those which that are a safe bet vs. those that aren't.

onebigdoor:
i manually tested `CAPSLOCK` and confirmed it works as expected, sending code `39` through the xcode console using same method.  i reran the same test script, and it's still skipping `CAPSLOCK`.  i combed through it, but haven't found the smoking gun yet.

innovaker:
My thoughts on this are the same as my [comment](https://github.com/zmkfirmware/zmk/pull/377#issuecomment-726696654) on macOS.

innovaker:
Please see my [comment](https://github.com/zmkfirmware/zmk/pull/377#issuecomment-727014384) on macOS.

onebigdoor:
updated reverting codes we want to manually test to `null`.  i left were F13-F24 which i'm confident came through correctly, and can be additionally documented further if necessary.  i also kept `CAPSLOCK` and `K_POWER`, and added `C_POWER` because i had tested them manually.

docs/src/data/hid.js
```diff
@@ -2265,7 +2265,7 @@ export default [
       linux: true,
       android: true,
       macos: null,
-      ios: null,
+      ios: true,
```
innovaker, line 2268: Are you sure this doesn't need manual testing?

docs/src/data/hid.js
```diff
@@ -2240,7 +2240,7 @@ export default [
       linux: true,
       android: true,
       macos: true,
-      ios: null,
+      ios: true,
```
innovaker, line 2243: Are you sure this doesn't need manual testing?

docs/src/data/hid.js
```diff
@@ -2983,7 +2985,7 @@ export default [
       linux: true,
       android: false,
       macos: null,
-      ios: null,
+      ios: true,
```
innovaker, line 2988: Are you sure this doesn't need manual testing?

innovaker (commented):
Thanks @onebigdoor!  I compared it against the sheet we were using and your feedback.  Only a few follow-up queries.

docs/src/data/hid.js
onebigdoor: nope.  those three escaped my revision.  i appreciate your thoroughness.  i've updated those three keys to `null`

innovaker:
When doing a final visual check on the rendered page this morning, `K_APPLICATION` caught me eye.  What does it actually do on iOS?  On Windows I believe it's the equivalent of the right-click menu.

innovaker:
Paging @onebigdoor, did you see my last message please?

onebigdoor:
`K_APPLICATION` has no effect on the apps i tried: Gmail, Safari, and Notes.

innovaker:
@onebigdoor, I've rebased this PR as [discussed](https://discord.com/channels/719497620560543766/784523018050469939/786244439563042816), however I couldn't force-push into your repository.

Instead, I've pushed it to: https://github.com/innovaker/zmk/tree/docs/ios-keyboard-codes

Please could you pull it and force-push it to your repository so that this PR updates?

I'd also be grateful if you could please eyeball it for mistakes because every line conflicted (as anticipated).

innovaker:
> `K_APPLICATION` has no effect on the apps i tried: Gmail, Safari, and Notes.

Should this be reflected in the PR?  The deploy preview has a star for `K_APPLICATION` on iOS.

onebigdoor:
@innovaker no.  i don't know why it closed.  i'm working on the push / merge.

onebigdoor:
i think i've got it.  just need to add a footnote about `K_APPLICATION` because the keycode comes through even though it apparently does nothing.  you can reopen, or i will in a bit.

innovaker:
> @innovaker no. i don't know why it closed. i'm working on the push / merge.

ah ok!  I think you might be the only person able to reopen it!  It should be the `Reopen and comment` button next to the `Comment` button :)

innovaker:
> i think i've got it. just need to add a footnote about `K_APPLICATION` because the keycode comes through even though it apparently does nothing. you can reopen, or i will in a bit.

Yeah, I feel like this is one of those "might not fit the current documentation" situations.

Are you intending to make it:
- ⭐ with footnote 
- ❌ with footnote
- ❔ with footnote

... or do we need a third state?

docs/docs/codes/_footnotes/ios-application.mdx
```diff
@@ -0,0 +1,2 @@
+Keycode is registered, but has no visible effect in tested apps (Gmail, Safari, Notes)
```
innovaker, line 1:
```suggestion
Key press is registered, but has no visible effect in tested apps (Gmail, Safari, Notes)
```

docs/src/data/footnotes.js
```diff
@@ -5,9 +5,13 @@
  */
 
 import example from "@site/docs/codes/_footnotes/example.mdx";
+import iosPower from "@site/docs/codes/_footnotes/ios-power.mdx";
 import macosPower from "@site/docs/codes/_footnotes/macos-power.mdx";
+import iosApplication from "@site/docs/codes/_footnotes/ios-application.mdx";
```
innovaker, line 10:
```suggestion
import iosApplication from "@site/docs/codes/_footnotes/ios-application.mdx";
import iosPower from "@site/docs/codes/_footnotes/ios-power.mdx";
import macosPower from "@site/docs/codes/_footnotes/macos-power.mdx";
```

docs/src/data/footnotes.js
```diff
@@ -5,9 +5,13 @@
...
 
 export default {
   example,
+  iosPower,
   macosPower,
+  iosApplication,
```
innovaker, line 16:
```suggestion
  iosApplication,
  iosPower,
  macosPower,
```

innovaker (commented):
Thanks @onebigdoor!

Very minor suggestions, mostly around keeping the footnotes file alphabetical as it's likely to grow organically otherwise.

PR #358, opened 2020-11-10
refactor(docs): Extract keymap examples into their own files

innovaker:
Eases maintenance and promotes reuse.


petejohanson (approved):
Thanks!

PR #360, opened 2020-11-10
fix(underglow): Switch from legacy SPI to SPIM

### LINKED ISSUE #305: RGB strips flickering
It appears there have been several reports of RGB strips and LEDs flickering in the current configurations. I am using the following strip and experienced this issue on my QAZ

`https://www.aliexpress.com/item/4000169932545.html`

I've tried another led strip (don't recall the source) and also had the same results. Tried different numbers of LEDs, behavior was the same.

Steps to reproduce: enable leds in your board.config. LEDs flicker in static and animated modes.

Nicell:
This replaces the legacy Nordic SPI master implementation in favor of SPIM. It has been confirmed to fix #305 for those who have tested it. It also reduces power consumption of the underglow part of the firmware from 55µA to 30µA. #312 should note this.

The reason for this fix working is speculated to be because SPIM does not require the CPU to transfer data packets to and from RAM. Because of this it's speculated that the CPU was busy for a few µs, which caused the strip update to get reset in the middle of transfer.

### DIFF
--- a/app/boards/shields/corne/boards/nice_nano.overlay
+++ b/app/boards/shields/corne/boards/nice_nano.overlay
@@ -1,5 +1,5 @@
 &spi1 {
-	compatible = "nordic,nrf-spi";
+	compatible = "nordic,nrf-spim";
 	/* Cannot be used together with i2c0. */
 	status = "okay";
 	mosi-pin = <6>;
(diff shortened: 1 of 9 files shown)


petejohanson (approved):
Thanks for spending so much time on this, @tominabox1 too!

PR #362, opened 2020-11-11
fix &mo bug when 'top layer' is not &trans #289

### LINKED ISSUE #289: &mo inside another &mo issue
When a momentary layer `(&mo)` exists inside another momentary layer on the same split, the opposite split gets stuck in that layer until both sides are reset.

This is my keymap for reference https://github.com/simonwjackson/zmk-config/blob/406e252b7440bcef113c8793a82a8b5b5a4cadc6/config/corne.keymap

okke-formsma:
this fixes #289


j1n6:
I ran into this similar issue, after merging this to the main branch locally & built the firmware, nested layer modifier seem to work for me.
For reference, this is the Keymap configured for testing: 
https://github.com/j1n6/zmk-keymap-custom-crkbd/blob/master/config/corne.keymap

app/src/keymap.c
```diff
@@ -130,14 +130,12 @@ int zmk_keymap_apply_position_state(int layer, u32_t position, bool pressed, s64
...
+        zmk_keymap_active_behavior_layer[position] = zmk_keymap_layer_state;
+    }
     for (int layer = ZMK_KEYMAP_LAYERS_LEN - 1; layer >= zmk_keymap_layer_default; layer--) {
-        u32_t layer_state =
-            pressed ? zmk_keymap_layer_state : zmk_keymap_active_behavior_layer[position];
-        if (is_active_layer(layer, layer_state)) {
+        if (is_active_layer(layer, zmk_keymap_active_behavior_layer[position])) {
             int ret = zmk_keymap_apply_position_state(layer, position, pressed, timestamp);
-
-            zmk_keymap_active_behavior_layer[position] = zmk_keymap_layer_state;
```
okke-formsma, line 139: this line inside of the loop was the cause of the bug. The `zmk_keymap_active_behavior_layer[position]` is updated  after the &mo is processed, causing the wrong layers to be activated on key up.

petejohanson (approved):
Sweet! Thanks!

PR #363, opened 2020-11-11
feat: add retro tapping (disable hold-tap timer)

### LINKED ISSUE #339: hold tap improvement: Retro tapping
new flavor for hold-taps which only triggers the hold behavior on interrupt. On release, if no other key is pressed, tap the 'tapping' key.

okke-formsma:
closes #339.

If you set the tapping_term_ms to `0`, the timer is disabled for this hold-tap:

- The 'hold-preferred' flavor with tapping_term_ms `0` triggers the hold behavior when another key is pressed.
- The 'balanced' flavor with tapping_term_ms `0` triggers the hold behavior when another key is pressed and released.
- The 'tap-preferred' flavor with tapping_term_ms `0` never triggers the hold behavior (so don't set the tapping_term for these to 0)!

### DIFF
--- a/app/src/behaviors/behavior_hold_tap.c
+++ b/app/src/behaviors/behavior_hold_tap.c
@@ -45,6 +45,7 @@ struct behavior_hold_tap_config {
     int tapping_term_ms;
     struct behavior_hold_tap_behaviors *behaviors;
     enum flavor flavor;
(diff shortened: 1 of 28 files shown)


okke-formsma:
closing this, because it does not properly do retro tapping.

PR #364, opened 2020-11-11
refactor(docs): Remove explicit document ids

innovaker:
Aligns older documents with newer file-based ids.  Lays the groundwork for upcoming organizational refactors.

@petejohanson, the newer documents haven't been setting explicit ids and have instead relied on implicit ids from the filenames and directory layouts.  We should probably align everything to one convention.  Please reject the PR if you feel explicit ids are the better path.


petejohanson (approved):
Nah, this works for me. Thanks!

PR #366, opened 2020-11-11
refactor(docs): Switch from singular noun to plural noun directories

innovaker:
This refactor makes for better (human-readable) URLs.  It also aligns better with the newer documents.


innovaker:
I appreciate that this is an issue that depends on the perspective you view it from.  I went for `codes` over `code` because it makes index pages (e.g. `/docs/codes`, `/docs/features`, `/docs/behaviors`) seem more natural.

PR #377, opened 2020-11-12
docs: add macos undocumented keycodes

onebigdoor:
adds undocumented keycodes for macOS
tested by logging output with Karabiner Event-Viewer, a part of Karabiner Elements v12.10.0
log file `macos_keycode_log.txt` is attached
keymap `keycode.txt` is attached
[keycode.txt](https://github.com/zmkfirmware/zmk/files/5532341/keycode.txt)
[macos_keycode_log.txt](https://github.com/zmkfirmware/zmk/files/5532718/macos_keycode_log.txt)
macbook pro 10.15.7


innovaker:
Thanks @onebigdoor!  Am I correct in assuming that the log you've uploaded was cutdown from the raw (original) to only the undocumented keys?

onebigdoor:
no, that's the firehose of all the keypresses from the keycode file

innovaker:
At first glance it seems like a weird subset.  It doesn't even begin with the alphabet.  Am I missing something?

innovaker:
Thanks for the fixed log @onebigdoor.  I'm pleased to confirm that your commits match my own analysis of the log.

May I suggest you also try `C_POWER` as it was deliberately missing from the keymap.

onebigdoor:
confirmed `C_POWER` manually with Karabiner.  it had no effect on the system (did not prompt shutdown or sleep)

innovaker:
Well, that's just disappointing isn't it.  I presume `K_POWER` was the same?

innovaker:
Rightyo, just `K_POWER` then! 👍

onebigdoor:
`K_POWER` was in my test script.  it similarly had no effect

innovaker:
That's weird because `K_POWER`'s usage code is in the log detected as `power`.  Perhaps it merits manual investigation before we move forward with this PR?

onebigdoor:
sorry, i'm being confusing.  both codes were recognized, and came through as `power`. neither had an effect on the system like pressing a hardware power button would.

onebigdoor:
accordingly, i marked each as `true` in `hid.js`

innovaker:
That's a critical distinction.  The OS interprets the codes, but doesn't seem to act on them.  Huh, useful.

It makes me wonder how many others codes are like that (#356 also), because at best they'll need a footnote per code and at worst they're effectively not supported.

I suspect some of the more exotic codes we've listed as `true` in this commit are going to need further manual investigation to gather specifics, ugh.  Don't worry, others can help with that.

onebigdoor:
yeah, that's a good point.  i was thinking of supported as "the os sees that you pressed it" but if we list a command as supported on a platform, and it literally has no effect, then that's very confusing.   i think it's worth listing what codes are recognized even if they don't perform the expected function.  they could be potentially leveraged in creative ways.  but i'm unsure how to clearly document that, and it probably makes sense to be a separate PR

onebigdoor:
strengthening the case for verifying with manual testing: i went back to try `C_POWER` and `K_POWER` again.  each has no effect on the system with a quick press (like the speed of normal typing or testing), but each sleeps on a deliberate press, and prompts a shut down if long-pressed.   it may be asking to much that we attempt to document the expected system functions of the recognized key presses, and this is a good example.

innovaker:
It sounds like the automated approach can only be trusted for the standard keys then.  Non-standard keys need closer inspection and extra information recorded.  I suggest using #375 for facilitating that.

I suggest:
- strip this PR down to only the safe bets, so that it can move forward
- add the outstanding keys (those that merit a closer look) as a checklist onto #375
- start collating information about those keys on #375, so that others can chip in
- start discussing ideas (on Discord/#development) for the best approaches to capturing each type of information on the docs
  - i.e. extra icons, footnotes, etc.  I've already implemented much of what we'll need and I can add anything missing.

What do you think?  I'm happy to pick up the slack for the splitting/checklist if you'd rather not.

innovaker:
As [discussed](https://discordapp.com/channels/719497620560543766/719544230497878116/776841539795943465), please see my best attempt at shortlisting here:
https://docs.google.com/spreadsheets/d/1fDFhbnFYIdSEQqp8br0MhKdxTeptO71XPvRysMOKuEc/
 - `N` (green): codes I think can continue in the PR without further verification
 - `Y` (yellow): codes I think merit further investigation

I used numerous resources to come to these conclusions but it was difficult to give the green light on many of them because Apple's own Developer Documentation has them listed.  So I can't help but wonder if Karabiner could also be a weak link?

The ones I green lit were mainly judged by combining the test results with Apple's documentation and my own experiences.

My gut says that much of the verification will come up negative, but I also suspect we'll only know for sure by digging deeper.

onebigdoor:
i took a look at the doc, and it's not immediately clear how your notation works.  can you add a legend?  i'm not particularly suspicious of Karabiner since its purpose is to intercept HID commands and allow you to manipulate them. the EventViewer is packaged as a debugging tool to assist.  it's plausible that it, or something else about my setup is causing an anomaly, of course, so i'm still very much in favor of independent verification.  my method on iOS is coming through UIKit, so there's a level of abstraction above the HID commands.  i've verified it's not showing consumer codes.  `K_POWER` and `C_POWER` both lock the home screen and don't show in the debugger.

innovaker:
Sorry, my bad.  I thought the legend I included in my [comment](https://github.com/zmkfirmware/zmk/pull/377#issuecomment-727037941) was sufficient.  I've amended it.  TDLR; strip the yellow codes `Y` from the PR.

innovaker (changes requested):
Thanks @onebigdoor!  Sorry it took longer than I expected to review.  Only one minor suggestion.

This work has also highlighted a bug in the footnotes which I'm patching: #395

innovaker (approved):

PR #380, opened 2020-11-13
refactor(hid): Replace modifier usage values with their preprocessor definitions

innovaker:
Improves readability of the HID reports.


petejohanson (approved):
Good call, thanks!

PR #381, opened 2020-11-13
refactor: change `keypad` to `keyboard`

innovaker:
The application usage used by report 1 is `HID_USAGE_GD_KEYBOARD`.  Moreover, the `keys` usage page (0x07) that primarily feeds into this report is predominantly keyboard codes.  The rest of the system should align with this naming convention.

Note: an alternative is to use `key` or `keys`, instead of `keyboard` / `keypad`.  I'm still not sure on that.  This PR is a step in the right direction however.

end-2-end test logs:
- [e2e-2020-11-13T14-04-25-ble.log](https://github.com/zmkfirmware/zmk/files/5537275/e2e-2020-11-13T14-04-25-ble.log)
- [e2e-2020-11-13T14-04-25-usb.log](https://github.com/zmkfirmware/zmk/files/5537276/e2e-2020-11-13T14-04-25-usb.log)

### DIFF
--- a/app/src/hid.c
+++ b/app/src/hid.c
@@ -10,7 +10,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/hid.h>
 #include <dt-bindings/zmk/modifiers.h>
 
-static struct zmk_hid_keypad_report kp_report = {
+static struct zmk_hid_keyboard_report kp_report = {
     .report_id = 1, .body = {.modifiers = 0, ._reserved = 0, .keys = {0}}};
 
 static struct zmk_hid_consumer_report consumer_report = {.report_id = 2, .body = {.keys = {0}}};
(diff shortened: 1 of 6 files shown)


app/src/hid.c
```diff
@@ -10,7 +10,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/hid.h>
 #include <dt-bindings/zmk/modifiers.h>
 
-static struct zmk_hid_keypad_report kp_report = {
+static struct zmk_hid_keyboard_report kp_report = {
```
petejohanson, line 13: Thoughts on renaming the variable too, e.g. `kb_report`?

docs/docs/behaviors/key-press.md
```diff
@@ -40,12 +40,12 @@ For instructions to resolve such an error, click [here](../troubleshooting###Imp
 
 ## Keypad Key Press
```
petejohanson, line 41: This header should be updated too. Now that we don't have consumer stuff separate, I'd argue for just "Key Press" here.

petejohanson (commented):
Generally looks good! Couple minor items. I'm fine with the naming, seems clearer anyways.

docs/docs/behaviors/key-press.md
innovaker: Good call, removed from this PR.  See #413.

app/src/hid.c
innovaker: Good idea.  It's slightly outside the scope of this PR so I'll follow up with another after this is merged.

innovaker:
Rebased, stripped out docs changes into #413, retested.

petejohanson (approved):
LGTM!

PR #382, opened 2020-11-13
refactor: 32-bit hid usage definitions

innovaker:


joelspadin:
Copying my comments from Discord:

Representing key codes with something like this could be useful for all the code outside DT handling (adjust field names as necessary):

```C
union hid_keycode {
  u32_t id;
  struct {
    u16_t page;
    u16_t code;
  };
};
```

That way you can just assign the u32 value from DT to `keycode.id` and get the conversions to usage page and code within that page for free (without any macros) with `keycode.page` and `keycode.code`.

(Note that, IIRC, anonymous structs inside unions are technically a C11 thing, but to my knowledge they're supported in every major compiler.)

Also, if you don't want to have to write `union` everywhere:

```C
typedef union hid_keycode { ... } keycode_t;
```

Some examples of how you could use this: https://godbolt.org/z/MM1Yar

innovaker:
Thanks @joelspadin!  That'll likely be the next stage (different PR), but I'll carry the idea forward to it!

PR #384, opened 2020-11-14
Updated function calls with missing reference. GPIO missing.

4pplet:
Updated function calls with missing reference. GPIO missing.

To my understanding:
DT_INST_GPIO_PIN() should be called instead of DT_INST_PIN()
DT_INST_GPIO_FLAGS() should be called instead of DT_INST_FLAGS()

Current implementation does not compile when using control pin.


Nicell (approved):
LGTM. Thanks 👍

petejohanson (approved):

PR #385, opened 2020-11-14
west test command to run testsuite

okke-formsma:
This PR adds  West command 'test' that runs our testsuite.

For now it adds little value, but this opens the way to run the testsuite from python instead of a bash script. This may enable us to run the tests on Windows and add nice things like pretty printing.

The command is invoked as `west test` and runs the current test runner.

### DIFF
--- a/app/scripts/west-commands.yml
+++ b/app/scripts/west-commands.yml
@@ -1,0 +1,9 @@
+# Copyright (c) 2020, ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+west-commands:
+  - file: scripts/west_commands/test.py
+    commands:
+      - name: test
+        class: Test
+        help: run zmk testsuite
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
         id: west-build
         with:
           entrypoint: /bin/bash
-          args: '-c "cd app && ./run-test.sh all"'
+          args: '-c "west test"'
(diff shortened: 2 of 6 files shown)


petejohanson (changes requested):
Can we update our GitHub workflow to invoke the tests this way, to be sure this continues working and doesn't regress?

okke-formsma:
The `west test` command can now be run in any directory (not necessarily `app`) and exits with the correct exit code.

app/scripts/west-commands.yml
```diff
@@ -0,0 +1,6 @@
+west-commands:
```
petejohanson, line 1: Please add a copyright/license header.

petejohanson (changes requested):
Looks generally good. One missing header, and this needs a rebase, then we can merge.

okke-formsma:
@petejohanson fixed, ready for merge

app/scripts/west-commands.yml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020, ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+west-commands:
+  - file: scripts/west_commands/test.py
+    commands:
+      - name: test
+        class: Test
+        help: run zmk testsuite
```
innovaker, line 9:
```suggestion
        help: run ZMK testsuite
```

petejohanson (approved):
Perfect! Thanks.

PR #386, opened 2020-11-14
Add pre-commit hooks for clang-format and prettier

okke-formsma:
clang-format and prettier need to be run on all files committed to ZMK.  Pre-commit hooks makes it easier to not forget to run these tools.

Installation is easy;

```
pip install pre-commit
pre-commit install
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+fail_fast: false
+repos:
+  - repo: https://github.com/pocc/pre-commit-hooks
+    rev: master
+    hooks:
+      - id: clang-format
+        args:
+          - -i
+  - repo: https://github.com/prettier/prettier
+    rev: master
+    hooks:
+      - id: prettier
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,11 @@ documentation to areas not currently covered are greatly appreciated.
 - To get started, from the `docs/` directory, run `npm ci` and then `npm start`.
 - Enhancements should be submitted as pull requests to the `main` branch of ZMK.
 
+### Formatting
+
+ZMK uses `prettier` to format documentation files. You can run prettier with `npm run prettier:format`.
+You can setup git to run prettier automatically when you commit by installing the pre-commit hooks: `pip3 install pre-commit`, `pre-commit install`.
+
 ## Code Contributions
 
 ### Development Setup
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+fail_fast: false
+repos:
+  - repo: https://github.com/pocc/pre-commit-hooks
+    rev: master
+    hooks:
+      - id: clang-format
```
joelspadin, line 6: Do these hooks only run if you're changing files relevant to the hook? It would be nice to not require you to install clang-format if you're only changing docs files and not require node if you're only changing app files.

okke-formsma: The pre-commit hooks are only run on the files that are changed. However, an 'environment' is created for all hooks that are specified, whether they are needed for the changed files or not. For the clang-format hook, clang-format needs to be installed in the system beforehand. 

pre-commit hooks are optional, so anyone who does not want to use them can just decide not to run `pre-commit install` and not be bothered with them.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+        args:
+          - -i
```
petejohanson, line 8: Does this pre-commit format and stage updates and include them in your commit? Or does this generate a new set of changes and bail, letting you verify first?

okke-formsma: the latter.

The commit fails (it's pre-commit hook after all) and then you need to add the changed files and commit again. So it's impossible to commit badly formatted files, except by using the `-n` flag with `git commit`.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - repo: https://github.com/prettier/prettier
+    rev: master
+    hooks:
+      - id: prettier
```
innovaker, line 12: Presumably `prettier` runs on any file it supports - not just docs?

okke-formsma: yup: https://github.com/pre-commit/mirrors-prettier

> By default, all files are passed to prettier, if you want to limit the file list, adjust types / files:
```
    -   id: prettier
        types: [javascript]
```

okke-formsma:
@innovaker @petejohanson any blockers for merging this?

petejohanson (approved):
No reason not to, since this is opt in. Thanks!

PR #387, opened 2020-11-15
docs: Add docs for setting up vscode

joelspadin:
Added a docs page with tips for setting up VS Code's code completion to work in ZMK's source files. Info for other IDEs can be added here later as needed.

### DIFF
--- a/docs/docs/development/ide-integration.md
+++ b/docs/docs/development/ide-integration.md
@@ -65,6 +65,8 @@ Change these options:
 | IntelliSense mode                     | gcc-arm                                              |
 | Advanced Settings > Compiler commands | `${workspaceFolder}/app/build/compile_commands.json` |
 
+#### Compiler Path
+
 <OsTabs>
 <TabItem value="debian">
 
@@ -137,3 +139,21 @@ ${env:ZEPHYR_SDK_INSTALL_DIR}/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc
 
 </TabItem>
 </OsTabs>
+
+#### Compiler Commands Path
+
+When building with all default options, the path to the compilation database file
+is `${workspaceFolder}/app/build/compile_commands.json` as shown in the table above,
+however some arguments to `west build` can change this path.
+
+The `-d` or `--build-dir` option lets you change the build directory to something
+other than `build`. Replace `build` in the above path with what you set this to.
+For example, if you build with `-d build/my_shield`, the path is
+`${workspaceFolder}/app/build/my_shield/compile_commands.json`. If you use this
+to keep builds for multiple keyboards separate, you may want to create a separate
+C/C++ configuration for each one in VS Code.
+
+You can also build from the root folder of the project instead of the `app`
+folder by adding `-S app` to your CMake arguments. In this case, simply remove
+`app` from the path to `compile_commands.json`, for example,
+`${workspaceFolder}/build/compile_commands.json`.
--- a/docs/docs/development/ide-integration.md
+++ b/docs/docs/development/ide-integration.md
@@ -1,0 +1,139 @@
+---
+title: IDE Integration
+sidebar_label: IDE Integration
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+export const OsTabs = (props) => (<Tabs
+groupId="operating-systems"
+defaultValue="debian"
+values={[
+{label: 'Debian/Ubuntu', value: 'debian'},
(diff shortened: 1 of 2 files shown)


petejohanson:
@joelspadin This looks great! Can we get this rebased for the sidebar conflicts, so we can get this in?

joelspadin:
Rebased.

petejohanson:
@joelspadin Build isn't happy:

```
9:09:01 PM: Error: Docusaurus found broken links!
9:09:01 PM: 
9:09:01 PM: Please check the pages of your site in the list bellow, and make sure you don't reference any path that does not exist.
Note: it's possible to ignore broken links with the 'onBrokenLinks' Docusaurus configuration, and let the build pass.
9:09:01 PM: Exhaustive list of all broken links found:
9:09:01 PM: 
9:09:01 PM: - On source page path = /docs/development/ide-integration:
9:09:01 PM:    -> linking to dev-build-flash#building (resolved as: /docs/development/dev-build-flash)
9:09:01 PM: 
9:09:01 PM:     at Object.reportMessage (/opt/build/repo/docs/node_modules/@docusaurus/core/lib/server/utils.js:37:19)
9:09:01 PM:     at Object.handleBrokenLinks (/opt/build/repo/docs/node_modules/@docusaurus/core/lib/server/brokenLinks.js:142:17)
9:09:01 PM:     at async build (/opt/build/repo/docs/node_modules/@docusaurus/core/lib/commands/build.js:104:5)
```

joelspadin:
Fixed.

docs/docs/development/ide-integration.md
```diff
@@ -0,0 +1,139 @@
...
+You will need to update this path any time you switch to a new version of the Zephyr SDK.
+:::
+
+</TabItem>
+<TabItem value="win">
+
+Your compiler path is
+
+```
+${env:GNUARMEMB_TOOLCHAIN_PATH}/bin/arm-none-eabi-gcc.exe
```
Nicell, line 89: I didn't realize that some of these tabs show an echo command that you copy the result of and some you just copy and paste (no need to run in cmd/bash). Not sure if the could be made consistent maybe? Or maybe a bit more obvious for people who are quickly skimming 😅

docs/docs/development/ide-integration.md
```diff
@@ -0,0 +1,139 @@
...
+a new configuration for you, but if the text box under **Configuration name** is empty,
+click **Add Configuration**, enter a name, and click **OK**.
+
+Change these options:
+
+| Option                                | Value                                                |
+| ------------------------------------- | ---------------------------------------------------- |
+| Compiler path                         | Path to your toolchain's GCC binary (see below)      |
+| IntelliSense mode                     | gcc-arm                                              |
+| Advanced Settings > Compiler commands | `${workspaceFolder}/app/build/compile_commands.json` |
```
Nicell, line 66: I sometimes build using the `-s app` flag, so the build folder will be in the root directory instead. Can we make it so both are added somehow? If not, this is enough.

docs/docs/development/ide-integration.md
joelspadin: I don't think we can make it be consistent because for some configurations (Docker, Windows), VS Code is running with the variables pointing to the compiler path already in its environment, but for others you have to source Zephyr first so only the integrated terminal will see those variables in its environment.

docs/docs/development/ide-integration.md
joelspadin: Yeah, this also breaks down if you use `-d` to give builds their own directories. I could maybe add a section explaining how to find compiler_commands.json if you don't build with defaults?

docs/docs/development/ide-integration.md
Nicell: I suppose we could convert them all to echo commands, but then we're not taking advantage of the environment variables for platforms that support it. I'm not sure why I didn't realize this at first, so not sure what wording would make it easier to understand. This is probably fine then.

docs/docs/development/ide-integration.md
Nicell: Yeah that seems like a good addition.

joelspadin: Added some more details for this.

petejohanson: The additions for the multiple build directories looks good to me. Thanks!

petejohanson (approved):
Looks great to me. Thanks @joelspadin

joelspadin:
You're welcome!

PR #389, opened 2020-11-15
refactor: make constant #binding-cells definitions optional

okke-formsma:
Each behavior is required to specify a #binding-cells value, which seems redundant. This PR makes it unnecessary to repeat this  constant value in each behavior.


okke-formsma:
```
devicetree error: <Node /behaviors/behavior_hold_tap_balanced in 'native_posix.dts.pre.tmp'> lacks #binding-cells
```
I'm not quite sure why the binding-cells property is still required, any thoughts?

okke-formsma:
it seems I forgot to commit all files. I'll try again later.

PR #390, opened 2020-11-16
docs: Explain ZMK stays connected to inactive host 

bcat:
This behavior was not obvious to me, and I actually considered reporting it as a bug until @Nicell explained to me on Discord that it's WAI. :)

### DIFF
--- a/docs/docs/behaviors/bluetooth.md
+++ b/docs/docs/behaviors/bluetooth.md
@@ -13,6 +13,10 @@ computer/laptop/keyboard should receive the keyboard input; many of the commands
 Please note there are only five available Bluetooth profiles by default. If you need to increase the number of available profiles you can set `CONFIG_BT_MAX_CONN` in your `zmk-config` `.conf` file.
 :::
 
+:::note Connection Management
+As an implementation detail, a ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
+:::
+
 ## Bluetooth Command Defines
 
 Bluetooth command defines are provided through the [`dt-bindings/zmk/bt.h`](https://github.com/zmkfirmware/zmk/blob/main/app/include/dt-bindings/zmk/bt.h) header,
--- a/docs/docs/behaviors/bluetooth.md
+++ b/docs/docs/behaviors/bluetooth.md
@@ -14,7 +14,7 @@ Please note there are only five available Bluetooth profiles by default. If you 
 :::
 
 :::note Connection Management
-As an implementation detail, a ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
+A ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
 :::
 
 ## Bluetooth Command Defines


innovaker:
@bcat, it looks like 2871dca9bfce2588e23da0905d7142ef5ff1c1a1 has affected this PR and GitHub isn't handling it gracefully.  Please could you rebase onto `HEAD` and force push.  Thanks!

bcat:
> @bcat, it looks like [2871dca](https://github.com/zmkfirmware/zmk/commit/2871dca9bfce2588e23da0905d7142ef5ff1c1a1) has affected this PR and GitHub isn't handling it gracefully. Please could you rebase onto `HEAD` and force push. Thanks!

Done.

docs/docs/behaviors/bluetooth.md
```diff
@@ -9,6 +9,9 @@ The bluetooth behavior allows management of various settings and states related
 between the keyboard and the host. By default, ZMK supports five "profiles" for selecting which bonded host
 computer/laptop/keyboard should receive the keyboard input; many of the commands here operation on those profiles.
 
+To enable fast switching between hosts, the keyboard will still show as "connected" on all host systems that are
+in range, though only the host bonded in the active profile will actually receive keystrokes.
```
Nicell, line 13: This isn't exactly right because it will only stay connected to any devices that have connected during the advertising period. In general the connection system works something like this:

- If the device is on an BT profile that isn't connected, it will advertise until it connects to the device on that profile
- Any other profile's host can connect while the device is advertising
- Previously connected profiles will stay connected after switching to another profile

It's a bit hard to convey what this all means in one or two sentences, so maybe this explanation is close enough? The only real incorrect piece here is that if the keyboard's current profile is connected, it won't advertise and therefore other host systems won't connect during this time even if they're in range. Thoughts?

bcat: Sorry for the delay, I was unavailable for some days due to a family thing.

What about wording like this?

> As an implementation detail, a ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.

I think that addresses the point I was actually confused about (e.g., whether or not this behavior was a bug) without adding too many low-level details that end users aren't likely to care about.

Nicell: Yeah this seems like a great way of putting it, thanks!

Nicell: @bcat Could you submit this change, so we can get this change merged? Thanks 😄

bcat: Whoops, lost track of this PR. Yes, done.

Nicell (approved):

docs/docs/behaviors/bluetooth.md
```diff
@@ -13,6 +13,10 @@ computer/laptop/keyboard should receive the keyboard input; many of the commands
 Please note there are only five available Bluetooth profiles by default. If you need to increase the number of available profiles you can set `CONFIG_BT_MAX_CONN` in your `zmk-config` `.conf` file.
 :::
 
+:::note Connection Management
+As an implementation detail, a ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
```
innovaker, line 17:
```suggestion
A ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
```
What do you think?  If you agree, please accept the change and I'll squash the commits during merge.

innovaker (commented):
Thanks @bcat!  Minor suggestion that came to mind when proof-reading.

docs/docs/behaviors/bluetooth.md
okke-formsma: I agree, let's get this merged.

PR #391, opened 2020-11-16
QAZ shield keymap formatting tweaks.

petejohanson:
Looking to use the QAZ shield for an upcoming meetup dem, and think some whitespace formatting tweaks helps it helps it be more readable for the demo.

### DIFF
--- a/app/boards/shields/qaz/qaz.keymap
+++ b/app/boards/shields/qaz/qaz.keymap
@@ -31,27 +31,27 @@
 
         default_layer {
             bindings = <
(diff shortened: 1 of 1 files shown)


petejohanson:
@tominabox1 Can I get your thoughts on these whitespace changes? I tried to capture the physical layout as best I could, while still having some vertical alignment.

tominabox1:
Its fine by me. It's going to be hard to do with the homerow mods and stuff. Thanks for updating!

PR #396, opened 2020-11-17
docs(docs/codes): Add comments of primary code names to hid's os data

innovaker:
This should make GitHub PR reviews easier because GitHub's change summaries do not currently include sufficient context without lots of digging.

Changes made with a RegExp Replace:
Find: `(names: \[")([\w\d_]+?)("[^	]+?os: \{)`
Replacement: `$1$2$3 // $2`


innovaker:
GitHub only seems to show 3 lines above and below by default, and it doesn't seem to be configurable.  I'm therefore closing this PR as I don't considered the extra potential maintenance worth the partial benefit.

PR #397, opened 2020-11-18
refactor(ble): silence warning `auth_passkey_display` not used

innovaker:
Silences the following build warning ...

```
../src/ble.c:407:13: warning: 'auth_passkey_display' defined but not used [-Wunused-function]
  407 | static void auth_passkey_display(struct bt_conn *conn, unsigned int passkey) {
```

@petejohanson, is this obsolete code?

Zephyr v2.4.0 (#223) seems to have stricter build rules by default, so this will soon cause an error.

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -404,6 +404,7 @@ static struct bt_conn_cb conn_callbacks = {
     .security_changed = security_changed,
 };
 
+/*
 static void auth_passkey_display(struct bt_conn *conn, unsigned int passkey) {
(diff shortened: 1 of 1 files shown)


petejohanson:
@innovaker Not obsolete, per se. It'll need to be refactored once we get more display code in, to actually do something useful. Definitely can be, at a minimum, commented out for now.

innovaker:
Thanks @petejohanson.  Please feel free to rebase and merge this PR.

PR #398, opened 2020-11-18
fix(battery): add missing `struct device *` in `zmk_battery_int`

innovaker:
This became apparent whilst laying the groundwork for #223.


Nicell (commented):
I'm a bit confused why this change was made. Line 17 already defines `battery`, and making this scoped variable will mean `zmk_battery_update` can no longer access the proper battery device binding.

innovaker (commented):
Good spot @Nicell!  This is why reviews are so critical.

#223 introduces a lot of refactoring and this was part of changes to 33 other files.  I simply didn't spot the battery variable at the top.

Closing.

PR #408, opened 2020-11-19
build(west): set `self: path:` to nothing

innovaker:
The self path should be `app`.  Setting it to nothing achieves the same result but with more flexibility if it becomes a module.

At time of writing, this change has also been made _implicitly_ in PR #385.


innovaker:
The purpose of this PR has been deprecated by the merge of PR #385.

PR #409, opened 2020-11-19
fix(setup.ps1): add jorne and jian to $splits

innovaker:
This was missing from:
- 3d030ec2abed9c8d48387e0002066c70bd5f085c
- f60439802ddd44ffec9d5081e2c1ea43e5ef5358


innovaker:
Scratch that, they were added *together* in f604398 and I missed that.

PR #414, opened 2020-11-21
refactor(hid): replace `kp_report` with `keyboard_report`

innovaker:
Aligns with other existing conventions.

Ref https://github.com/zmkfirmware/zmk/pull/381#discussion_r528212807, I looked around and noticed there was already precedent for `keyboard_report` rather than `kb_report`.  It also aligns with how `consumer_report` is treated and how the reports types are named.


petejohanson (approved):
Thanks!

PR #415, opened 2020-11-22
Initial display widget approach

petejohanson:


app/src/battery.c
```diff
@@ -35,7 +38,16 @@ static int zmk_battery_update(struct device *battery) {
 
     LOG_DBG("Setting BAS GATT battery level to %d.", state_of_charge.val1);
 
-    return bt_gatt_bas_set_battery_level(state_of_charge.val1);
+    rc = bt_gatt_bas_set_battery_level(state_of_charge.val1);
```
Nicell, line 41: Should there be an event listener in the bluetooth files that handles this change?

app/src/display/Kconfig
```diff
@@ -0,0 +1,33 @@
...
+endchoice
+
+choice ZMK_DISPLAY_STATUS_SCREEN
+    prompt "Default status screen for displays"
+    default ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
+
+config ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
+    bool "Built in status screen"
+    select ZMK_WIDGET_BATTERY_STATUS
+    select ZMK_WIDGET_OUTPUT_STATUS
```
Nicell, line 24: Can any of these `select`s be switched to `depends on`? Although by the looks of it none of these should have issues with being selects.

app/src/display/widgets/battery_status.c
```diff
@@ -0,0 +1,81 @@
...
+        strcat(text, LV_SYMBOL_BATTERY_FULL);
+    } else if (level > 65) {
+        strcat(text, LV_SYMBOL_BATTERY_3);
+    } else if (level > 35) {
+        strcat(text, LV_SYMBOL_BATTERY_2);
+    } else if (level > 5) {
+        strcat(text, LV_SYMBOL_BATTERY_1);
+    } else {
+        strcat(text, LV_SYMBOL_BATTERY_EMPTY);
+    }
```
Nicell, line 51: I like this a lot as the battery percent isn't super accurate. Will hopefully reduce people being confused by the percent going up and down somewhat randomly.

app/src/battery.c
petejohanson: That's debatable. Until we have some other true "source of truth for the battery level", other than the BAS service, I think this makes sense, since this is recording the updated state.

I see a future refactor for `app/src/battery.c` to keep the state locally, then we pull the updating of the BAS GATT service into an event handler.

app/src/display/Kconfig
petejohanson: Reading the caveats here, I think there's not much risk in going w/ `select` here. If you pick this, it should be selecting those other symbols as well

app/src/display/widgets/battery_status.c
petejohanson: Yeah, there's a reason I didn't just print the battery percentage as a number for the first pass on this.

app/src/display/Kconfig
petejohanson: I actually refactored this to make them `default y if BT`, then made the default built in status screen handle them being enabled or not.

PR #417, opened 2020-11-23
feat(ext-power): Cut power when PM is sleeping

Nicell:
This adds support for cutting ext-power when the power management policy is set to low power using the [central Zephyr device power management system](https://docs.zephyrproject.org/2.3.0/reference/power_management/index.html#central-method).

### DIFF
--- a/app/src/ext_power_generic.c
+++ b/app/src/ext_power_generic.c
@@ -31,6 +31,9 @@ struct ext_power_generic_data {
 #if IS_ENABLED(CONFIG_SETTINGS)
     bool settings_init;
 #endif
+#ifdef CONFIG_DEVICE_POWER_MANAGEMENT
+    uint32_t pm_state;
+#endif
 };
 
 #if IS_ENABLED(CONFIG_SETTINGS)
@@ -139,6 +142,10 @@ static int ext_power_generic_init(const struct device *dev) {
         return -EIO;
     }
 
+#ifdef CONFIG_DEVICE_POWER_MANAGEMENT
+    data->pm_state = DEVICE_PM_ACTIVE_STATE;
+#endif
+
 #if IS_ENABLED(CONFIG_SETTINGS)
     settings_subsys_init();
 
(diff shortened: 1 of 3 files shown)


innovaker (commented):
This looks OK on the face of it, but I don't know this part of the system yet (and I can't test it), so I'm going to assign it to @petejohanson.

app/src/ext_power_generic.c
```diff
@@ -167,6 +174,39 @@ static int ext_power_generic_init(const struct device *dev) {
...
+#ifdef CONFIG_DEVICE_POWER_MANAGEMENT
+static int ext_power_generic_pm_control(const struct device *dev, uint32_t ctrl_command,
+                                        void *context, device_pm_cb cb, void *arg) {
+    int rc;
+    struct ext_power_generic_data *data = dev->driver_data;
+
+    switch (ctrl_command) {
+    case DEVICE_PM_SET_POWER_STATE:
+        if (*((uint32_t *)context) == DEVICE_PM_ACTIVE_STATE) {
+            data->pm_state = DEVICE_PM_ACTIVE_STATE;
```
petejohanson, line 186: Shouldn't this also be calling `ext_power_generic_enable`?

petejohanson (changes requested):
Just one comment. Probably not needed *yet*, but when we get to turning devices to low power state due to user idle, we'll want the ability to switch power back on again.

app/src/ext_power_generic.c
Nicell: Not in this case. When the device turns back on from sleep, the ext_power device is restarted and will automatically turn ext power back to its last saved state from settings. When we disable before going to sleep, that off state isn't saved in settings, so we'll properly retrieve the last state.

If we explicitly enabled here, people's power cut offs would turn back on whenever they wake up and get saved that way.

petejohanson: My point is, when we add per-device power state management, we will want to cut/restore power state based on idle/active.

So we'll need an API for "on/off" that isn't state persisting. So we can call *that* from here.

Doesn't need to be this PR, so I'm going to approve for now.

We also need to consider that other devices may say "no, don't sleep now!" and abort the sleep state change, I believe.

petejohanson (approved):
Approved, with the knowledge that this is a first step.

Thanks @Nicell!

app/src/ext_power_generic.c
```diff
@@ -167,6 +174,39 @@ static int ext_power_generic_init(const struct device *dev) {
     return 0;
 }
 
+#ifdef CONFIG_DEVICE_POWER_MANAGEMENT
+static int ext_power_generic_pm_control(const struct device *dev, uint32_t ctrl_command,
+                                        void *context, device_pm_cb cb, void *arg) {
+    int rc;
+    struct ext_power_generic_data *data = dev->driver_data;
```
innovaker, line 181: This looks like the Zephyr 2.3.0 API.  Was this tested with 2.4.0?

PR #427, opened 2020-11-26
Update devcontainer config to facilitate working with documentation

mcrosson:
Updates the devcontainer config to include npm and expose port 3000 via the docker image so local updates to documentation are easier to put together.

This is especially handy for anyone using the docker+vscode dev environment


mcrosson:
The port update was merged via another PR and the upcoming devcontainer enhancements will address NPM being missing.

Closing PR as it's no longer necessary given other change sin the zmk pipeline.

Issue #422, opened 2020-11-25
Windows Dev Setup (Suggestions)


mcrosson:
I recently went to setup the native Windows development environment on Windows 10 version 20H2 and ran into troubles with the ```choco install dfu-util``` step. From what I was able to figure out with help from the Discord channel the ```dfu-util``` choco package is no longer maintained or install able via the tool.

The docker instructions, on the other hand, worked great and I've had no issues using the docker + vscode approach on an Intel i7 CPU *or* an Intel i5 CPU. The i5 is a little slower but not prohibitively slow. As long as a user has 4Gb of RAM they shouldn't run into any hassles using this approach on Windows.

Would it be prudent/advisable to update the documentation to use this approach by default for Windows users? If so, where in the docs should be updated? I may have some time in the coming weeks to submit the necessary documentation updates to default the docker + vscode approach for Windows users.

KingCoinless:
I'm inclined to agree with this approach. I had far too many issues associated with CMake on Windows which were completely eliminated when running Docker with VSCode. I've heard others having issues with Choco or DFU-Util as well. Unaware users on Windows will likely select the 'Windows' option over 'VSCode + Docker', and have to deal with a fidgety and arguably less stable process as a result. If developers don't want to dual-boot or use a Linux system, they should utilize VSCode + Docker to avoid this conundrum.

mcrosson:
Based on feedback in Discord and a previous PR, closing this issue as native builds on Windows should still be maintained.

If anything some better organization of the setup docs may be a better approach.

PR #428, opened 2020-11-26
Overhaul windows documentation

mcrosson:
Per discord and issue #422 this is an update to the docs that makes it clear the docker+vscode environment should be used for development on windows

### DIFF
--- a/docs/docs/development/setup.md
+++ b/docs/docs/development/setup.md
@@ -153,25 +153,9 @@ sudo dnf install -y \
 </TabItem>
 <TabItem value="win">
 
-:::note
-Use `cmd.exe` with these instructions rather than PowerShell.
+:::danger
+Windows Users: Please use the VS Code & Docker development environment for developing ZMK on Windows.
 :::
-
-Chocolatey is recommended and used for the following instructions. You can manually install each of these applications and add them to your `PATH` if you don't want to use Chocolatey.
-
-1. [Install Chocolatey](https://chocolatey.org/install)
-2. Open `cmd.exe` as **Administrator**
-3. Run the following `choco` commands:
-   ```shell
-   choco install cmake --installargs 'ADD_CMAKE_TO_PATH=System'
-   choco install ninja gperf python git
-   ```
-
-It is recommended to install `dfu-util` to avoid any later confusion while flashing devices. You can do this by running this command with chocolatey:
-
-```shell
-choco install dfu-util
-```
 
 </TabItem>
 <TabItem value="mac">
@@ -197,6 +181,10 @@ This setup leverages the same [image which is used by the GitHub action](https:/
 The docker container includes `west` and the compilation toolchain. If you're using docker and VS Code, you can skip right to [Source Code](#source-code).
 :::
 
+:::info
+If you're using the Docker container on Windows you can open a WSL2 environment and clone the zmk sources within WSL2 for a faster build process. The files will be available outside of the WSL2 enviornment at `\\wsl$` in explorer.
+:::
+
 </TabItem>
 </OsTabs>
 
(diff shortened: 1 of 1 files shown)


Nicell:
I personally disagree with removing all of the documentation for Windows. We can always offer Docker as an alternative in the opening of the documentation for all platforms, but all of this Windows related information is still very useful. I personally can't get Docker to work on my computer properly.

mcrosson:
> I personally disagree with removing all of the documentation for Windows. [...] but all of this Windows related information is still very useful. 

How did you setup the dfu-utils packages? There are a number of choco commands in the docs that just do not work for fresh installations at this point in time.

> I personally can't get Docker to work on my computer properly.

At all? Or is there a specific piece that's not working for you? Which version of windows, cpu and how much ram do you have on your machine?

Nicell:
> How did you setup the dfu-utils packages? There are a number of choco commands in the docs that just do not work for fresh installations at this point in time.

If this is the case I suggest we fix the instructions to make them work rather than remove them. I haven't tested these instructions in a while, and I wasn't aware that they were broken. Docker is a great solution, however, having native instructions is important as well.

> At all? Or is there a specific piece that's not working for you? Which version of windows, cpu and how much ram do you have on your machine?

Correct, Docker doesn't work on my machine. I've tried a various amount of things as many of my projects use Docker, but my WSL2 backend doesn't work. I'm not looking to run through a bunch of more fixes for it. The only fix seems to be to re-install Windows at this point, and I'm not interested in that at the moment.

mcrosson:
> If this is the case I suggest we fix the instructions to make them work rather than remove them. I haven't tested these instructions in a while, and I wasn't aware that they were broken. Docker is a great solution, however, having native instructions is important as well.

Would you be amenable to suggesting docker + vscode as a better approach on windows as it's usually (clearly there are edge cases where wsl2 gets broken) self-contained and easy to setup while also retaining instructions for setting up a native toolchain?

> Correct, Docker doesn't work on my machine. I've tried a various amount of things as many of my projects use Docker, but my WSL2 backend doesn't work. I'm not looking to run through a bunch of more fixes for it. The only fix seems to be to re-install Windows at this point, and I'm not interested in that at the moment.

Understood, I wanted to be sure it wasn't a specific limitation and specific to something like WSL2 being broken as you point out.

Nicell:
> Would you be amenable to suggesting docker + vscode as a better approach on windows as it's usually (clearly there are edge cases where wsl2 gets broken) self-contained and easy to setup while also retaining instructions for setting up a native toolchain?

I think it makes more sense to offer it as a simple solution for all platforms. I don't see any reason why we would need to single out Windows to recommend it.

At the same time I'm not a fan of WSL development for ZMK in general. I did it for a while early on, and moving files back to the Windows file system to flash is cumbersome in my opinion. It's a good solution, but it's not a perfect solution. For my workflow, native Windows works better, so I'd like for us to be able to offer both to developers.

The `dfu-utils` package is not necessary for most of our developers, so we could probably remove that. I hope we can fix any other issues there may be. I can hopefully run through the instructions again soon.

mcrosson:
> > Would you be amenable to suggesting docker + vscode as a better approach on windows as it's usually (clearly there are edge cases where wsl2 gets broken) self-contained and easy to setup while also retaining instructions for setting up a native toolchain?
> 
> I think it makes more sense to offer it as a simple solution for all platforms. I don't see any reason why we would need to single out Windows to recommend it.

I can confirm it's not an option for Raspberry Pi builds but it should be workable on Linux/OSX/WIndows as a solid default. I'll close the PR and work on a different documentation update that notes Docker+VSCode is more self-contained and pretty quick to get online while leaving the native build options and documentation alone.

Would it be good to put the Docker+VSCode option first as the default one that's shown with a note to the effect of "this is probably the easiest, start here"?

> At the same time I'm not a fan of WSL development for ZMK in general. I did it for a while early on, and moving files back to the Windows file system to flash is cumbersome in my opinion. It's a good solution, but it's not a perfect solution. For my workflow, native Windows works better, so I'd like for us to be able to offer both to developers.

I don't think I'd go native WSL2 only dev either for ZMK. The Docker + VSCode stuff leverages the local Windows file system and makes it pretty seamless to work between the dev environ in VSCode and explorer for the output.

> The `dfu-utils` package is not necessary for most of our developers, so we could probably remove that. I hope we can fix any other issues there may be. I can hopefully run through the instructions again soon.

Understood, and thank you.

PR #441, opened 2020-11-28
build(west): add clone-depth = 1 to zephyr and uf2

innovaker:
This reduces zephyr's footprint to 33% and dramatically reduces the cloning times (`west update`).

What do you think @petejohanson?  Please check it locally just in case there's any environmental factors.

If this PR is merged, then I'll reset the CI build cache with a follow-up PR.  I suspect we'll see improvements.

### DIFF
--- a/app/west.yml
+++ b/app/west.yml
@@ -10,6 +10,7 @@ manifest:
     - name: zephyr
       remote: petejohanson
       revision: zmk-v2.3.0-with-fixes
+      clone-depth: 1
       import:
         # TODO: Rename once upstream offers option like `exclude` or `denylist`
         name-blacklist:
@@ -36,5 +37,6 @@ manifest:
     - name: uf2
       remote: microsoft
       path: tools/uf2
+      clone-depth: 1
   self:
     west-commands: scripts/west-commands.yml


petejohanson:
Does this force submodules of zephyr to also be shallow clones?

innovaker2:
Sadly not.  I tried adding `clone-depth` to them also but it requires you to specify their paths if you do so.  A flaw in the overlay approach imo.

innovaker:
Sadly not, I tried adding `clone-depth` to them, but then it requires you the `paths` to be specified also.  A flaw in the overlay system imo.

petejohanson:
Yeah, that's what I meant, west modules. Ah well, even this is a win.

We should try to consider how to do this for user repo GH Actions to drop their build times too.

innovaker:
> We should try to consider how to do this for user repo GH Actions to drop their build times too.

Agreed.  I've already drafted a version for the equivalent in the `modular` PoC.  I'll look to backport it when I get a chance.

innovaker:
> We should try to consider how to do this for user repo GH Actions to drop their build times too.

It will be picked up automatically I believe?  The issue with the user repos will be invalidating the cache.  Most of them will be fine as they don't build that often so it will expire naturally.

petejohanson:
> > We should try to consider how to do this for user repo GH Actions to drop their build times too.
> 
> It will be picked up automatically I believe?  The issue with the user repos will be invalidating the cache.  Most of them will be fine as they don't build that often so it will expire naturally.

I mean we should have them shallow clone ZMK itself, and then inherit this shallow cloning.

innovaker:
Oh right, I see.  So that's the templates then.  I don't think we really have the tools in place to do a change like that yet, I don't think?  Across everyone's repositories?

I also doubt the performance gains for ZMK itself will be that significant at this stage.

For the above reasons, and given that the `zmk-config` user repositories might potentially be superseded - depending on the outcomes of the modular PoC - I'd suggest putting that idea on the backburner for now.

PR #455, opened 2020-12-02
Update cmake note in docs to be 3.15 which is the current minimum version

mcrosson:
This is a re-implementation of GitHub PR 403 that bumps the minimum cmake version note in the docs


petejohanson (approved):
Thanks!

PR #457, opened 2020-12-02
ci(build): replace board matrix array with list

innovaker:
A multi-line list can be more merge friendly than a single-line array.


petejohanson (approved):
Fully agree!

PR #459, opened 2020-12-02
ci(build): add a .uf2 extension to artifacts

innovaker:
@petejohanson, any objections?


innovaker:
My mistake, withdrawn.

PR #461, opened 2020-12-02
ci(build): improve uf2 guard condition

innovaker:
Checking for the existence of the uf2 file is a more scalable solution.


innovaker:
@petejohanson, please choose between this PR _or_ #463.

innovaker:
This option will lead to having at least two different archive files (zips) - `hex` in one, `uf2` in the other.

petejohanson:
Closed in favor of #463

PR #463, opened 2020-12-02
ci(build): remove proton_c guard from uf2 archive step

innovaker:
Guards conditions aren't strictly required because the build only logs a warning if the file doesn't exist.  It still completes successfully.


innovaker:
@petejohanson, please choose between this PR _or_ #461.

innovaker:
I personally prefer this approach.  It's tidier in the code and the artifacts.

petejohanson (approved):
Yeah, works for me, if it just warns instead of erroring on failing to find a file.

PR #464, opened 2020-12-02
ci(build): add support for board-only builds

innovaker:
Lays the groundwork for CI building boards that don't require shields.

Tested using `dz60rgb_rev1` board.  See https://github.com/innovaker/zmk/actions/runs/397136358 for results.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,13 +86,28 @@ jobs:
         id: west-zephyr-export
         with:
           args: "zephyr-export"
+      - name: Prepare variables
+        id: variables
+        run: |
+          SHIELD_ARG=
+          ARTIFACT_NAME="${{ matrix.board }}"
+
+          if [ -n "${{ matrix.shield }}" ]; then
+            SHIELD_ARG="-DSHIELD=${{ matrix.shield }}"
+            ARTIFACT_NAME="${ARTIFACT_NAME}-${{ matrix.shield }}"
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -86,13 +86,30 @@ jobs:
...
         with:
           args: "zephyr-export"
+      - name: Prepare build parameters
+        id: build-params
+        run: |
+          SHIELD=
+          if [ -n "${{ matrix.shield }}" ]; then
+            SHIELD="-DSHIELD=${{ matrix.shield }}"
+          fi
+          echo ::set-output name=shield::${SHIELD}
```
petejohanson, line 96: Thoughts on using something more generic, to allow other build steps to optionally add other defines?

petejohanson (commented):
Just one question.

.github/workflows/build.yml
innovaker: Yeah, that's a good question and that's actually what I did for `zmk-docker`'s workflow (which is more refined).

I suspect in this case I was just powering through and didn't take a step back before pushing.

I can see the merits so I'll refactor it and re-push.

innovaker: That seems to have done the trick.  This build (a PR which will build off this one) demonstrates the result:
https://github.com/zmkfirmware/zmk/runs/1501689317

petejohanson (approved):
Yeah, dig it!

PR #466, opened 2020-12-03
core/ble-split-central-notify-main-thread

### LINKED ISSUE #221: ZMK crashes on key spam.
ZMK crashes when keys are pressed too rapidly. 

At minimum 1 reset is required to fix once it's occurred. 

Last time I watched it on the debugger the the failed to send message count hit around 40 and then the debugger stopped outputting lines. 

This is a very repeatable issue, so no issue testing it.

petejohanson:
* Avoid corruption by using work to process
  peripheral key position events on the main
  work thread, like local kscan events are.
* FIxes #221


innovaker:
Interesting!  Presumably it stealing too much time on the thread that calls `split_central_notify_func`?  The BT stack equivalent of an interrupt?

petejohanson:
> Interesting! Presumably it was stealing too much time on the thread that calls `split_central_notify_func`? The BT stack equivalent of an interrupt?

@innovaker 

I *suspect* that we had multi-threaded non-locked access to all the other parts of the HID, processing, etc. and that causes issues. The kscan stuff gets ISR triggers, then passes to the main work thread to process the events, but the BT stuff is handling from the BT host thread. So if both occur at the same time, e.g. fast typing, simultaneous presses, etc. it *may* be causing issues.

I just pushed some fixes, and now am running this on my main corne daily driver, I'm going to see if it fixes things or not.

petejohanson:
Of course... I was just able to reproduce this by mashing *only* on my central side, so this may be an important fix, but isn't the whole story.

tominabox1:
This seems to disable sleep on my zmk-config build based test on my QAZ.  https://github.com/tominabox1/zmk-config/tree/spamfix

petejohanson:
Closing this, in favor of the Zephyr 2.4.0 which seems to have addressed this, from my testing.

PR #470, opened 2020-12-04
Tweak/cleanup bluetooth documentation so codes are more clear

mcrosson:
This update mainly removes the `BT_*_CMD` codes which shouldn't be used in practice and better calls attention to `BT_SEL` needing a parameter passed to function correctly


petejohanson (approved):
Thanks!

PR #477, opened 2020-12-04
feat(devcontainer): add forwardPorts 3000

innovaker:
Required for testing the documentation with the host's browser.


mcrosson:
This works really well for anyone working on the documentation. VS Code will also pop up a little notice when the docusaurus server starts so you know the port is active.

petejohanson (approved):

PR #479, opened 2020-12-04
feat(devcontainer): add volume mount for root user

innovaker:
Caches VS Code extensions and other profile settings between zmk containers.

Based on advice from KemoNine.


mcrosson:
I can confirm this works well and I haven't run into any hassles or hiccups with the updated approach to `/root` being a volume.

petejohanson (approved):

PR #480, opened 2020-12-04
feat(devcontainer): persist bash history

innovaker:
When combined with the root user volume, this commit instructs bash to save each command to the bash history after execution, thereby sharing the bash history between any containers that use the volume.

Based on the advice of KemoNine.


innovaker:
#479 is the prerequisite to this PR.  I will rebase after it's closed.

mcrosson:
Any chance this can be flipped from a draft? The root volume mount just landed in main.

petejohanson (approved):

PR #481, opened 2020-12-04
adopt zmk-docker

innovaker:


mcrosson:
I can confirm the docker image in this PR works well. I've been using it as my default container for dev since the call for testing went out.

petejohanson:
I can confirm that the zmk-docker stuff is solid, both for current 2.3.0 Zephyr use in `main`, and for the upcoming 2.4 branch. We should definitely move forward with getting that repo brought under the `zmkfirmware` organization, and publishing those Docker images under our `zmkfirmware` Docker hub org.

innovaker:
Great!  Thanks @petejohanson.

I've created the [zmk-docker](https://github.com/zmkfirmware/zmk-docker) repository, but before I push to it we will need to setup the following secrets:
- `DOCKERHUB_TOKEN`
- `DOCKERHUB_USERNAME`
- `GHCR_TOKEN`

Please could you assist?

For the `GHCR_TOKEN`, my current understanding is that a GitHub Personal Access Token is required because [GITHUB_TOKEN doesn't currently work for the GitHub Container Registry](https://docs.github.com/en/free-pro-team@latest/packages/guides/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry).  It needs to be a collaborator account (so you, me or Nicell), or perhaps a new [machine user](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users) - whichever you think is most appropriate.

petejohanson:
Secrets available

I already had two similarly named secrets, `DOCKER_HUB_USERNAME` and `DOCKER_HUB_PASSWORD` (which is just a token value), and I just added the `GHCR_TOKEN` secret with a personal access token that includes registry write/read access.

innovaker:
I'll adapt `innovaker/zmk-docker` to match, test it and then migrate it.

innovaker:
As discussed on [Discord](https://discord.com/channels/719497620560543766/719544230497878116/786245767291076628), I'm building/testing to this:
image

petejohanson:
> As discussed on [Discord](https://discord.com/channels/719497620560543766/719544230497878116/786245767291076628), I'm building/testing to this:
> image

@innovaker What permissions did you set for the token? Want to be sure we're aligned on that, and I don't need some extra permission granted that you happened to have added.

innovaker:
image

innovaker:
I'm not even sure if the repo is required.

innovaker:
Also, if we switch to using `ghcr.io` for the candidates then we might have to add `read:packages`.

petejohanson:
> Also, if we switch to using `ghcr.io` for the candidates (in the future) then we might have to add _read:packages_.
> Currently candidate images only get pushed to dockerhub.

At least when I created my token, it seemed that `write:packages` was implying/forcing `read:packages`, so I think we'll be ok.

innovaker:
Test-run with https://github.com/innovaker/zmk-docker came out fine.

I'm going to start pushing to https://github.com/zmkfirmware/zmk-docker in 4 stages:
- push up to 2.3 release
- tag release
- push up to 2.4 release
- tag release

I'll do the first 3 stages and will leave instructions on the last one for you @petejohanson so that you're familiar with the process going forward.

petejohanson:
> Test-run with https://github.com/innovaker/zmk-docker came out fine.
> 
> I'm going to start pushing to https://github.com/zmkfirmware/zmk-docker in 4 stages:
> 
> * push up to 2.3 release
> * tag release
> * push up to 2.4 release
> * tag release
> 
> I'll do the first 3 stages and will leave instructions on the last one for you @petejohanson so that you're familiar with the process going forward.

Can we add the instructions to that repo's README.md, for future reference? Even just basic ones to get something in writing would be handy.

innovaker:
> Can we add the instructions to that repo's README.md, for future reference? Even just basic ones to get something in writing would be handy.

OK to PR it after?

petejohanson:
> > Can we add the instructions to that repo's README.md, for future reference? Even just basic ones to get something in writing would be handy.
> 
> OK to PR it after?

Most definitely.

innovaker:
Error during tagging of the 2.3.0 release.

From https://github.com/zmkfirmware/zmk-docker/runs/1525553976?check_suite_focus=true

> denied: failed_precondition: Improved container support has not been enabled for '***'. Learn more: https://docs.github.com/packages/getting-started-with-github-container-registry/enabling-improved-container-support
> Error: Process completed with exit code 1.

petejohanson:
> Error during tagging of the 2.3.0 release.
> 
> From https://github.com/zmkfirmware/zmk-docker/runs/1525553976?check_suite_focus=true
> 
> > denied: failed_precondition: Improved container support has not been enabled for '***'. Learn more: https://docs.github.com/packages/getting-started-with-github-container-registry/enabling-improved-container-support
> > Error: Process completed with exit code 1.
> 
> I must've done that before I started development and since forgotten.

Just updated the `zmkfirmware` org with those settings. Re-run the tasks?

innovaker:
Running again.

innovaker:
Same error.

innovaker:
After several [Discord discussions](https://discord.com/channels/719497620560543766/719544230497878116/786283838723981372), I'm building to this:

image

innovaker:
Test run on my account passed.  Beginning the trench run ...

innovaker:
image

innovaker:
Thanks for [fixing](https://discord.com/channels/719497620560543766/719544230497878116/786416427548672010) that @petejohanson.

The first 3 stages have completed successfully.  The last stage (for you) is as follows please:
- Fetch https://github.com/zmkfirmware/zmk-docker
- Tag the latest commit with `2.4.0-0.11.4`
- Push the tag

About 7.5 minutes later it should create a `2.4` git tag which indicates it was successful.

If you do it via the CLI, please make a note of the commands as it'll save me double checking for the documentation.

We can then redirect our attention to this PR.

innovaker:
Great!  4th stage looks good.

Now that `zmk-docker` has been adopted by zmkfirmware, we can focus on this PR.

The build/test workflows will read the image from zmkfirmware on the GitHub Container Registry.  This avoids any bandwidth limitations from DockerHub.

Although we'll probably make the `zmk-build-arm` and `zmk-dev-arm` packages public on ghcr.io eventually, we should still provide login credentials from the build/test workflows.  This will make testing on forks (of `zmk` and `zmk-docker`) easier.  Along the same line of thinking, we should probably add a `GHCR_OWNER` secret as well.

i.e.
```
    container:
      image: ghcr.io/${{ secrets.GHCR_OWNER }}/zmk-build-arm:2.3
      credentials:
        username: ${{ github.repository_owner }}
        password: ${{ secrets.GHCR_TOKEN }}
```

So we'll need these secrets added to https://github.com/zmkfirmware/zmk:
- `GHCR_OWNER` = `zmkfirmware`
- `GHCR_TOKEN` which matches up with the username.

We may also need `GHCR_USERNAME` instead of `github.repository_owner`?  I suggest digging around in the GitHub package permissions system before deciding that.  I can't see the packages on https://github.com/orgs/zmkfirmware/packages but I'm guessing you need to look here: https://github.com/orgs/zmkfirmware/packages/container/zmk-build-arm/settings

WARNING: as noted on [Discord](https://discord.com/channels/719497620560543766/719544230497878116/786596884105527296), let's keep the ghcr.io packages private until making them public becomes an absolute necessity (probably after the beta).  It just gives us some flexibility as public DockerHub images can be deleted, whereas public ghcr.io images cannot.

Please could you sanity check my thinking above, explore the permissions options and add the appropriate secrets.  I'll update the PR to match.

petejohanson:
Thinking about this more... We need to update the user config templates to leverages these images as well, which means we really will need these to be public. The docker stuff is used extensively there, and should be migrated/updated as well. We don't need to do that at the same time as this work, but that would be the final "need them to be public" step.

innovaker:
> Thinking about this more... We need to update the user config templates to leverages these images as well, which means we really will need these to be public. The docker stuff is used extensively there, and should be migrated/updated as well. We don't need to do that at the same time as this work, but that would be the final "need them to be public" step.

Noted.  We could also leverage DockerHub for that without needing ghcr.io public I believe?  It's along the same lines as the devcontainer = much less traffic so the 6 hour slots shouldn't be an issue.

mcrosson:
> 
> 
> > Thinking about this more... We need to update the user config templates to leverages these images as well, which means we really will need these to be public. The docker stuff is used extensively there, and should be migrated/updated as well. We don't need to do that at the same time as this work, but that would be the final "need them to be public" step.
> 
> Noted. We could also leverage DockerHub for that without needing ghcr.io public I believe? It's along the same lines as the devcontainer = much less traffic so the DockerHub 6 hour limits shouldn't be an issue.

Rarely should folks run into the DockerHub rate limits and since you have more flexibility with public/non-public images and deletes on DockerHub I'd recommend leaning that direction.

petejohanson:
> > > Thinking about this more... We need to update the user config templates to leverages these images as well, which means we really will need these to be public. The docker stuff is used extensively there, and should be migrated/updated as well. We don't need to do that at the same time as this work, but that would be the final "need them to be public" step.
> > 
> > 
> > Noted. We could also leverage DockerHub for that without needing ghcr.io public I believe? It's along the same lines as the devcontainer = much less traffic so the DockerHub 6 hour limits shouldn't be an issue.
> 
> Rarely should folks run into the DockerHub rate limits and since you have more flexibility with public/non-public images and deletes on DockerHub I'd recommend leaning that direction.

Good point. Works for me.

petejohanson:
Thinking: If we use the ghcr.io packages for our workflows and keep them, then won't forks have major issues running the Actions in their forks?

innovaker:
> Thinking: If we use the ghcr.io packages for our workflows and keep them, then won't forks have major issues running the Actions in their forks?

They would be required to provide the secrets to leverage the Actions, that's a good point.

mcrosson:
> 
> 
> > Thinking: If we use the ghcr.io packages for our workflows and keep them, then won't forks have major issues running the Actions in their forks?
> 
> They would be required to provide the secrets to leverage the Actions, that's a good point.

At that point wouldn't a fork just be updating their actions to use the DockerHub image as that's what they are deving against and is mirrored from the zmk CI side anyhow?

If you're going to bother to fork you can bother to update GH actions IMO

innovaker:
> If you're going to bother to fork you can bother to update GH actions IMO

That doesn't make PRs easy.

petejohanson:
> > If you're going to bother to fork you can bother to update GH actions IMO
> 
> That doesn't make PRs easy.

Perhaps we dig for a way to select a different base container depending on environment?

mcrosson:
> 
> 
> > If you're going to bother to fork you can bother to update GH actions IMO
> 
> That doesn't make PRs easy.

Even in my fork for PRing core features the CI fails then works fine when I open the PR. The documentation I plan on updating to include most/all of the commands you should be running ahead of CI checks so it should be a moot point longer term.

innovaker:
Following further Discord [discussion](https://discord.com/channels/719497620560543766/719544230497878116/786645303877042226), we're going with DockerHub as a first pass and will monitor it for rate problems as we scale.  Public ghcr.io packages will be our future fallback option if rates become an issue.

I'll update the PR accordingly.

innovaker:
CI build logs LGTM.

petejohanson (approved):
Spot checked builds, tests, zips. Everything looks good!

PR #484, opened 2020-12-07
Add layer change event

mcrosson:
This PR implements an event for when layers change on the keyboard. This will be used in the future for OLED status widget showing the highest, active layer.


app/include/zmk/events/layer-state-changed.h
```diff
@@ -0,0 +1,19 @@
...
+#include <zmk/event-manager.h>
+
+struct layer_state_changed {
+    struct zmk_event_header header;
+    u32_t layer;
+    bool state;
+    s64_t timestamp;
+};
+
+ZMK_EVENT_DECLARE(layer_state_changed);
```
petejohanson, line 19: Check out `app/include/zmk/events/modifiers-state-changed.h`, for an example of a more complete "event factory function", that I think we should be using for new event types in the future, instead of requiring folks wishing to fire the event to set the fields themselves.

Better encapsulation, and easier API, so you can then do:

```
ZMK_EVENT_RAISE(create_layer_state_changed(layer, state));
```

And the factory would set the `timestamp` to current time, etc.

app/include/zmk/keymap.h
```diff
@@ -6,7 +6,10 @@
 
 #pragma once
 
+u8_t zmk_layer_default();
+u32_t zmk_layer_state();
```
petejohanson, line 10: When exposing bitflag values like this, I usually like to create a type alias to make them easier to work with remember what they are, etc, e..g

```
typedef u32_t zmk_layers_state;
```

or somesuch, and then use that pervasively.

petejohanson (changes requested):
A few comments. Thanks for the PR!

mcrosson:
A side note : will squash if desired once the PR is ready for merge.

app/src/keymap.c
```diff
@@ -16,9 +16,11 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #include <zmk/event-manager.h>
 #include <zmk/events/position-state-changed.h>
+#include <zmk/events/layer-state-changed.h>
 #include <zmk/events/sensor-event.h>
 
-static u32_t zmk_keymap_layer_state = 0;
+typedef u32_t zmk_layers_state;
```
petejohanson, line 22: This should be defined in `app/include/zmk/keymap.h` so it can be used in the definition of the `zmk_layer_state()` function.

app/include/zmk/keymap.h
petejohanson, line 10: These should both be `zmk_keymap_layer_*` to have the proper prefix.

petejohanson (changes requested):
Couple minor follow ups. Thanks!

app/include/zmk/keymap.h
```diff
@@ -6,7 +6,12 @@
 
 #pragma once
 
+typedef u32_t zmk_keymap_layers_state;
+
+u8_t zmk_keymap_layer_default();
+zmk_keymap_layers_state zmk_keymap_layer_state();
 bool zmk_keymap_layer_active(u8_t layer);
+u8_t zmk_highest_layer_active();
```
petejohanson, line 14: This should be:

```suggestion
u8_t zmk_keymap_highest_layer_active();
```

app/src/keymap.c
```diff
@@ -76,20 +77,35 @@ static struct zmk_behavior_binding zmk_sensor_keymap[ZMK_KEYMAP_LAYERS_LEN]
...
+
+zmk_keymap_layers_state zmk_keymap_layer_state() { return _zmk_keymap_layer_state; }
 
 bool zmk_keymap_layer_active(u8_t layer) {
-    return (zmk_keymap_layer_state & (BIT(layer))) == (BIT(layer));
+    return (_zmk_keymap_layer_state & (BIT(layer))) == (BIT(layer));
 };
 
-int zmk_keymap_layer_activate(u8_t layer) { SET_LAYER_STATE(layer, true); };
+u8_t zmk_highest_layer_active() {
```
petejohanson, line 97: Ditto:

```suggestion
u8_t zmk_keymap_highest_layer_active() {
```

petejohanson (changes requested):
One more naming fix. Then this looks good to go!

petejohanson (approved):
LGTM!

Issue #487, opened 2020-12-07
CMake issue when using 3.19+


mrzealot:
The build docs specifically mention that our cmake installs shouldn't be "old", but it's only a lower limit at 3.15. But after I installed a fresh cmake v3.19.1, it failed the build saying something along the lines of:

`in function 'exit': exit.c:(.text.exit+0x60): undefined reference to '_exit'`

Tracked this to a [related bug report](https://github.com/zephyrproject-rtos/zephyr/issues/22060) where the consensus seemed to be that this is caused by cmake 3.19 (either 3.19.0 or 3.19.1). Downgrading to 3.18.5 did indeed help. This might be worth noting in the docs @mcrosson...

PR #488, opened 2020-12-07
Add note about cmake 3.19 not working with zephr ; addresses issue #487

mcrosson:
Add a note about cmake 3.19 with the current zmk version having a bug : https://github.com/zephyrproject-rtos/zephyr/issues/22060

Addresses issue #487


petejohanson (approved):
Thanks!

PR #489, opened 2020-12-07
(feature) Add &to keycode/behavior

mcrosson:
Add `&to` behavior/key code to allow enabling a layer while also disabling all other layers. Similar to toggle but includes the disable part that's not included in the toggle behavior.

Code was modeled after the `&tog` behavior.


app/src/keymap.c
```diff
@@ -99,6 +99,14 @@ int zmk_keymap_layer_toggle(u8_t layer) {
     return zmk_keymap_layer_activate(layer);
 };
 
+int zmk_keymap_layer_to(u8_t layer) {
+    for (int i = 0; i < 32; i++) {
+        zmk_keymap_layer_deactivate(i);
```
okke-formsma, line 104: `zmk_keymap_layer_state = 0` is an easier way of writing this loop.

okke-formsma (commented):
looks good! I think it makes sense to get rid of the loop in zmk_keymap_layer_to, otherwise perfect.

mcrosson:
Per Discord conversation( https://discord.com/channels/719497620560543766/719565084208398406/787032810523066409) updated PR with appropriate is layer active logic as well as tweaked the way layers are deactivated in the &to key code handling

okke-formsma:
to make sure the implementation is correct and doesn't break in the future, please add some tests. Scenarios to consider

* &to from layer 0 to layer 1, verify no layer release event is emitted for default layer, verify a layer active event is sent for the new layer
* &to from layer 1 to the default layer, verify a layer release for layer 1
* &to from layer 1 to layer 1, verify either no layer events for layer 1 are emitted (it was already active so the &to should not fire any events imo)

mcrosson:
@okke-formsma I added a set of test cases similar to what you suggested. I'm not sure I got them 100% but I think they are close if nothing else, can you review and let me know?

mcrosson:
Looks like the new test failed but I'm unsure how/why...

app/src/keymap.c
```diff
@@ -115,8 +121,20 @@ int zmk_keymap_layer_toggle(uint8_t layer) {
     return zmk_keymap_layer_activate(layer);
 };
 
-bool is_active_layer(uint8_t layer, zmk_keymap_layers_state layer_state) {
-    return (layer_state & BIT(layer)) == BIT(layer) || layer == _zmk_keymap_layer_default;
+int zmk_keymap_layer_to(uint8_t layer) {
```
okke-formsma, line 124: this method causes 32 layer_state_changed events (31 for disabling all layers except the default layer + 1 for enabling the to-layer).

Layer-changed-events should only be triggered if the layer state actually changes.

Not sure if the layer that is being moved to should be disabled & enabled again (or just kept enabled if it was already enabled). I don't have a strong opinion on this.

app/tests/to-layer/normal/events.patterns
```diff
@@ -0,0 +1,2 @@
+s/.*hid_listener_keycode/kp/p
+s/.*to_keymap_binding/to/p
```
okke-formsma, line 2: Please add some pattern to record all layer-state-changed events, so you can verify which layer state changed events are seen.

app/src/keymap.c
```diff
@@ -115,8 +121,20 @@ int zmk_keymap_layer_toggle(uint8_t layer) {
...
+    if (zmk_keymap_layer_active(layer)) {
+        return -1;
+    }
+
+    for (int i = 0; i < 32; i++) {
```
okke-formsma, line 129: as per joels suggestion, maybe add a define for mx layers

```c
#define MAX_LAYERS (sizeof(the_layer_variable) * 8)
```

okke-formsma (changes requested):
overall good! I think we need to find a way to limit the number of triggered events somewhat.

mcrosson:
Number of events is greatly reduced -- reduced to only the affected layers ; also added debug info/test snapshot to match expectations

app/src/keymap.c
```diff
@@ -115,8 +127,23 @@ int zmk_keymap_layer_toggle(uint8_t layer) {
     return zmk_keymap_layer_activate(layer);
 };
 
-bool is_active_layer(uint8_t layer, zmk_keymap_layers_state layer_state) {
-    return (layer_state & BIT(layer)) == BIT(layer) || layer == _zmk_keymap_layer_default;
+int zmk_keymap_layer_to(uint8_t layer) {
+    // Default layer is *always* active ; if going TO the default layer, need to short circuit this
+    // condition to ensure proper operation
+    if (layer != zmk_keymap_layer_default() && zmk_keymap_layer_active(layer)) {
```
okke-formsma, line 133: There's a bug in here; if layers 0, 1, 2, 3 are active, and going TO layer 1, layers 2 and 3 should be disabled.

docs/docs/behaviors/layers.md
```diff
@@ -57,6 +57,21 @@ Example:
 &lt LOWER SPACE
 ```
 
+## To Layer
+
+The "to layer" behavior enables a layer and disables _all_ other layers.
```
okke-formsma, line 62: maybe add a note about excluding the default layer.

okke-formsma (changes requested):

okke-formsma:
lgtm @mcrosson I think it's time to rebase and get this onto Pete's plate.

okke-formsma (approved):
lgtm. Let's rebase and get it onto Pete's plat.

mcrosson:
Rebased to latest main branch, squashed and ready for @petejohanson

app/src/behaviors/behavior_to_layer.c
```diff
@@ -0,0 +1,46 @@
...
+    LOG_DBG("position %d layer %d", event.position, binding->param1);
+    return 0;
+}
+
+static const struct behavior_driver_api behavior_to_driver_api = {
+    .binding_pressed = to_keymap_binding_pressed,
+    .binding_released = to_keymap_binding_released,
+};
+
+static const struct behavior_to_config behavior_to_config = {};
```
petejohanson, line 40: There's some cruft inherited from other behaviors in our code, but we don't need to carry it forward... If you don't use any config, or data, you can skip defining these types, declaring instances, and just pass `NULL` for these two fields to the device init macro.

petejohanson (changes requested):
Just one nitpick, otherwise this looks good. Thanks!

app/src/keymap.c
```diff
@@ -20,6 +20,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/events/sensor_event.h>
 
 static zmk_keymap_layers_state_t _zmk_keymap_layer_state = 0;
+#define MAX_LAYERS (sizeof(_zmk_keymap_layer_state) * 8)
```
petejohanson, line 23: We have a define for this, check maybe keymap.h?

petejohanson (changes requested):
Missed one thing. Sorry.

app/include/zmk/keymap.h
```diff
@@ -11,9 +11,11 @@ typedef uint32_t zmk_keymap_layers_state_t;
 uint8_t zmk_keymap_layer_default();
 zmk_keymap_layers_state_t zmk_keymap_layer_state();
 bool zmk_keymap_layer_active(uint8_t layer);
+bool zmk_keymap_layer_active_with_state(uint8_t layer, zmk_keymap_layers_state_t state_to_test);
```
petejohanson, line 14: Sorry, one last thing I missed. This is only used internally in `keymap.c`, we shouldn't publish this method yet, unless it's actually needed.

petejohanson (changes requested):
Sorry, I keep doing the "one more thing" review approach, which sucks. This really is close, I promise.

petejohanson (approved):
Sweet!

PR #494, opened 2020-12-09
fix(west) test command should not swallow errors.

okke-formsma:
the os.system command does not return status codes but something else. The more modern `subprocess.run` has a friendlier interface.


innovaker:
I'm happy to confirm this fails as expected:
https://github.com/innovaker/zmk/runs/1526287222

petejohanson (approved):

PR #495, opened 2020-12-10
Add generic idle/sleep activity code, display blanking/resume on activity state change.

petejohanson:
Currently, the new "idle" timeout is 30 seconds by default, but is controllable via Kconfig. 

We'll want to consider future enhancement to possible also do "ext power cutoff on idle" as a strategy too, for keyboards w/ RGB.

This *just* focuses on display blanking on idle, as a start.


mcrosson:
A quick test with my tidbit on both battery and usb shows the screen blanking after the 30s default

PR #498, opened 2020-12-11
ci(workflows): remove zephyr.base-prefer configfile

innovaker:
CI builds successfully without this step.


petejohanson (approved):
Indeed. Thanks!

Issue #361, opened 2020-11-11
Modifiers improvement: LS(LC(LGUI)) should create only explicit modifiers


okke-formsma:
the LS, LC, LG and LA modifiers are always creating implicit modifiers that are disabled on the next keydown.

This is not expected behavior when the "main key" pressed is a modifier like `LS(LC(LGUI))`, in which case all three modifiers should be explicit modifiers.

Possible solution:
* add explicit_modifiers to keycode-state-changed events
* update keypress behavior to send explicit modifiers when the 'main' key is a modifier key
* update hid_listener.c to set and unset the explicit modifiers too

okke-formsma:
created PR #499 which addresses this issue.

PR #499, opened 2020-12-11
Explicit modifiers ("hyper key")

okke-formsma:
Also see #361. 

This PR makes LS(LEFT_CONTROL) work as if shift and control were both
pressed explicitly. Previously, the left shift would have been released
as soon as another key was pressed. The implicit behavior is useful in
case of LS(NUMBER_1) when rolling over to other keys.

There are also a few refactor commits included, which clean up related code.


mrzealot:
Did a quick test with my `&mt LC(LS(LALT)) RET` key, works like a charm now. Would shop here again.

okke-formsma:
rebased onto 2.4.0

app/include/dt-bindings/zmk/modifiers.h
```diff
@@ -25,4 +25,8 @@
 #define RC(keycode) APPLY_MODS(MOD_RCTL, keycode)
 #define RS(keycode) APPLY_MODS(MOD_RSFT, keycode)
 #define RA(keycode) APPLY_MODS(MOD_RALT, keycode)
-#define RG(keycode) APPLY_MODS(MOD_RGUI, keycode)
\ No newline at end of file
+#define RG(keycode) APPLY_MODS(MOD_RGUI, keycode)
+
+#define IS_MOD(usage_page, keycode)                                                                \
+    (usage_page == HID_USAGE_KEY && keycode >= HID_USAGE_KEY_KEYBOARD_LEFTCONTROL &&               \
+     keycode <= HID_USAGE_KEY_KEYBOARD_RIGHT_GUI)
```
innovaker, line 32: Should this be a macro or a function?  @joelspadin, any thoughts please?

app/src/hid.c
```diff
@@ -49,6 +49,32 @@ int zmk_hid_unregister_mod(zmk_mod modifier) {
     return 0;
 }
 
+int zmk_hid_register_mods(zmk_mod_flags modifiers) {
+    int err;
+    for (zmk_mod i = 0; i < 8; i++) {
```
petejohanson, line 54: `sizeof(zmk_mod_flags_t)`?

app/src/hid.c
```diff
@@ -49,6 +49,32 @@ int zmk_hid_unregister_mod(zmk_mod modifier) {
...
+    }
+    return 0;
+}
+
+int zmk_hid_unregister_mods(zmk_mod_flags modifiers) {
+    int err;
+    for (zmk_mod i = 0; i < 8; i++) {
+        if (modifiers & (1 << i)) {
+            err = zmk_hid_unregister_mod(i);
+            if (err) {
```
petejohanson, line 70: Do we really want to bail on the loop if some entry errors?

petejohanson (commented):
A couple thoughts.

app/include/dt-bindings/zmk/modifiers.h
joelspadin: I don't see anything in here that would require it to be a macro, so my vote is function. I can almost guarantee you that the compiler will inline this if you make it a function, and the resulting binary will be very close to if not identical to with a macro.

app/src/hid.c
okke-formsma: Nah. There's no way to recover from an error either, so I'll remove the error handling.

app/include/dt-bindings/zmk/modifiers.h
okke-formsma: would you define that function in the header file, and if not, where should we put it? I would like to keep this close together with the `SELECT_MODS` macro and friends.

joelspadin: Two ways you can do this:

1. Define it in the header as an inline function. (Needs to be inline so the linker won't complain when it sees that every C file that includes modifiers.h has its own copy of the function.)
2. Declare it in the header and define it in some other C file.

petejohanson: So, the challenge here, is this header here *really* should only be macros, and those for keymaps. Either way this code shouldn't really be here. Perhaps inline function in `app/include/zmk/keys.h`?

petejohanson (commented):
On a whole, this looks great. Just one comment on the macro vs. inline function discussion.

petejohanson (approved):

petejohanson:
@okke-formsma can you rebase to fix merge conflicts?

okke-formsma:
@petejohanson done

PR #500, opened 2020-12-11
feature(ci): Build split w/ display enabled for testing.

petejohanson:
Recent failures w/ the Zephyr 2.4 stuff pointed out a deficiency in our current automated build, which is we only build board w/ their default configs.

This includes an additional two entries in our build matrix w/ the display stuff enabled, but doesn't archive them (yet), just builds to be sure we didn't break display code.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,7 @@ jobs:
           - splitreus62_right
           - tg4x
           - tidbit
+        cmake-args: [""]
         include:
           - board: dz60rgb_rev1
           - board: nrf52840_m2
@@ -62,6 +63,14 @@ jobs:
           - board: planck_rev6
           - board: proton_c
             shield: clueboard_california
+          - board: nice_nano
+            shield: kyria_left
+            cmake-args: -DCONFIG_ZMK_DISPLAY=y
+            skip-archive: true
+          - board: nice_nano
+            shield: kyria_right
+            cmake-args: -DCONFIG_ZMK_DISPLAY=y
+            skip-archive: true
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -104,8 +113,9 @@ jobs:
           echo ::set-output name=shield-arg::${SHIELD_ARG}
           echo ::set-output name=artifact-name::${ARTIFACT_NAME}
       - name: Build (west build)
-        run: west build -s app -b ${{ matrix.board }} -- ${{ steps.variables.outputs.shield-arg }}
+        run: west build -s app -b ${{ matrix.board }} -- ${{ steps.variables.outputs.shield-arg }} ${{ matrix.cmake-args }}
       - name: Archive artifacts
+        if: ${{ !matrix.skip-archive }}
         uses: actions/upload-artifact@v2
         with:
           name: "${{ steps.variables.outputs.artifact-name }}"


innovaker (commented):
Sounds like a good idea.  I see this as the first stepping stone towards [distilling the build coverage](https://discord.com/channels/719497620560543766/719544230497878116/787026173104357396) which we should probably implement soon given the increasing error rates.

Also, we should probably agree on a convention for the variables used in workflows.  I used hyphens in my recent refactors and you've used underscores in this PR.  Let's settle on one convention.  A cursory look through the [start-workflows](https://github.com/actions/starter-workflows/tree/main/ci) suggest it's not clear-cut though.

petejohanson:
Since we're deferring #502 w/ the archive upload fix, @innovaker any opinion on this for now?

innovaker:
> Since we're deferring #502 w/ the archive upload fix, @innovaker any opinion on this for now?

Sorry, this one got past me.  Sure, if you feel it's needed before #502 is completed.  I'll do a review.

petejohanson:
> > Since we're deferring #502 w/ the archive upload fix, @innovaker any opinion on this for now?
> 
> Sorry, this one got past me. Sure, if you feel it's needed before #502 is completed. I'll do a review.

Yes please. I just pushed fixes for the underscore/hyphen discrepancy, this is a stop gap to be sure display changes don't break things.

.github/workflows/build.yml
```diff
@@ -55,13 +55,22 @@ jobs:
           - splitreus62_right
           - tg4x
           - tidbit
+        build-arg: [""]
```
innovaker, line 58:
```suggestion
        build-args: [""]
```
I can see this argument turning into a catch-all argument, so perhaps we should make it plural?

Also, I guess it's technically `cmake-args` (perhaps?) as it's after the `--`?  That said, I'm not sure what terminology is used to describe the group of arguments before the `--`!?

.github/workflows/build.yml
```diff
@@ -104,8 +113,9 @@ jobs:
           echo ::set-output name=shield-arg::${SHIELD_ARG}
           echo ::set-output name=artifact-name::${ARTIFACT_NAME}
       - name: Build (west build)
-        run: west build -s app -b ${{ matrix.board }} -- ${{ steps.variables.outputs.shield-arg }}
+        run: west build -s app -b ${{ matrix.board }} -- ${{ steps.variables.outputs.shield-arg }} ${{ matrix.build-arg }}
       - name: Archive artifacts
+        if: ${{ !matrix.skip-archive }}
```
innovaker, line 118: What's the benefit of adding archive skipping?  Is this strictly necessary?

innovaker (commented):
For your consideration ...

.github/workflows/build.yml
innovaker: `cmake-args` also seems to match well with: https://docs.zephyrproject.org/2.4.0/guides/west/build-flash-debug.html#permanent-cmake-arguments

.github/workflows/build.yml
petejohanson: Without this, the archive and uf2 images will be the same, and so we'd replace the previously archived version with default settings w/ this overridden one w/ various optional things turned on.

.github/workflows/build.yml
petejohanson: I think `cmake-args` works for me. Those other things would be `west-args`.

.github/workflows/build.yml
innovaker: Oh right, I didn't realise it overwrote artifacts with the same name:
https://github.com/actions/upload-artifact#uploading-to-the-same-artifact

Yeah, then I guess this is the simplest approach.

innovaker: Alternatively we could provide a unique suffix for the artifact name.  That might also prove more flexible going forward.

petejohanson: We can look at doing that, I don't see the value (yet) in archiving these though, these are for test coverage, not "use this image"

.github/workflows/build.yml
petejohanson: @innovaker Updated to `cmake-args`

innovaker (approved):
LGTM.

In future versions we should consider the merits of a default `archive: true/false` parameter over `skip-archive`.

PR #514, opened 2020-12-15
(feat) Allow setting underglow color by key press

mcrosson:
This PR implements a behavior to allow setting the HSV of the underglow via a keymap binding.

### DIFF
--- a/app/include/dt-bindings/zmk/rgb.h
+++ b/app/include/dt-bindings/zmk/rgb.h
@@ -4,14 +4,29 @@
  * SPDX-License-Identifier: MIT
  */
 
-#define RGB_TOG 0
-#define RGB_HUI 1
-#define RGB_HUD 2
-#define RGB_SAI 3
-#define RGB_SAD 4
-#define RGB_BRI 5
-#define RGB_BRD 6
-#define RGB_SPI 7
-#define RGB_SPD 8
-#define RGB_EFF 9
-#define RGB_EFR 10
+#define RGB_TOG_CMD 0
+#define RGB_HUI_CMD 1
+#define RGB_HUD_CMD 2
+#define RGB_SAI_CMD 3
+#define RGB_SAD_CMD 4
+#define RGB_BRI_CMD 5
+#define RGB_BRD_CMD 6
+#define RGB_SPI_CMD 7
+#define RGB_SPD_CMD 8
+#define RGB_EFF_CMD 9
+#define RGB_EFR_CMD 10
+#define RGB_COLOR_HSB_CMD 11
+
+#define RGB_TOG RGB_TOG_CMD 0
+#define RGB_HUI RGB_HUI_CMD 0
+#define RGB_HUD RGB_HUD_CMD 0
+#define RGB_SAI RGB_SAI_CMD 0
+#define RGB_SAD RGB_SAD_CMD 0
+#define RGB_BRI RGB_BRI_CMD 0
+#define RGB_BRD RGB_BRD_CMD 0
+#define RGB_SPI RGB_SPI_CMD 0
+#define RGB_SPD RGB_SPD_CMD 0
+#define RGB_EFF RGB_EFF_CMD 0
+#define RGB_EFR RGB_EFR_CMD 0
+#define RGB_COLOR_HSB(h, s, v) RGB_COLOR_HSB_CMD(((h) << 16) + ((s) << 8) + (v))
+#define RGB_COLOR_HSV RGB_COLOR_HSB
--- a/app/src/behaviors/behavior_rgb_underglow.c
+++ b/app/src/behaviors/behavior_rgb_underglow.c
@@ -21,28 +21,31 @@ static int behavior_rgb_underglow_init(const struct device *dev) { return 0; }
 static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
                                      struct zmk_behavior_binding_event event) {
     switch (binding->param1) {
-    case RGB_TOG:
+    case RGB_TOG_CMD:
         return zmk_rgb_underglow_toggle();
-    case RGB_HUI:
+    case RGB_HUI_CMD:
         return zmk_rgb_underglow_change_hue(1);
-    case RGB_HUD:
+    case RGB_HUD_CMD:
         return zmk_rgb_underglow_change_hue(-1);
-    case RGB_SAI:
+    case RGB_SAI_CMD:
         return zmk_rgb_underglow_change_sat(1);
-    case RGB_SAD:
+    case RGB_SAD_CMD:
         return zmk_rgb_underglow_change_sat(-1);
-    case RGB_BRI:
+    case RGB_BRI_CMD:
         return zmk_rgb_underglow_change_brt(1);
-    case RGB_BRD:
+    case RGB_BRD_CMD:
         return zmk_rgb_underglow_change_brt(-1);
-    case RGB_SPI:
+    case RGB_SPI_CMD:
         return zmk_rgb_underglow_change_spd(1);
-    case RGB_SPD:
+    case RGB_SPD_CMD:
         return zmk_rgb_underglow_change_spd(-1);
-    case RGB_EFF:
+    case RGB_EFF_CMD:
         return zmk_rgb_underglow_cycle_effect(1);
-    case RGB_EFR:
+    case RGB_EFR_CMD:
         return zmk_rgb_underglow_cycle_effect(-1);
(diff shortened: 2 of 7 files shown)


app/src/rgb_underglow.c
```diff
@@ -339,6 +339,20 @@ int zmk_rgb_underglow_toggle() {
...
+int zmk_rgb_underglow_set_hsb(uint16_t hue, uint8_t saturation, uint8_t brightness) {
+    if (hue > 360 || saturation > 255 || brightness > 255) {
+        return -ENOTSUP;
+    }
+
+    state.hue = hue;
+    state.saturation = saturation;
+    state.brightness = brightness;
+
+    zmk_rgb_underglow_effect_solid();
```
Nicell, line 351:
```suggestion
    state.current_effect = UNDERGLOW_EFFECT_SOLID;
```

app/src/rgb_underglow.c
```diff
@@ -339,6 +339,20 @@ int zmk_rgb_underglow_toggle() {
     return zmk_rgb_underglow_save_state();
 }
 
+int zmk_rgb_underglow_set_hsb(uint16_t hue, uint8_t saturation, uint8_t brightness) {
+    if (hue > 360 || saturation > 255 || brightness > 255) {
```
Nicell, line 343:
```suggestion
    if (hue > 360 || saturation > 100 || brightness > 100) {
```

app/src/behaviors/behavior_rgb_underglow.c
```diff
@@ -21,28 +21,31 @@ static int behavior_rgb_underglow_init(const struct device *dev) { return 0; }
...
-    case RGB_SPD:
+    case RGB_SPD_CMD:
         return zmk_rgb_underglow_change_spd(-1);
-    case RGB_EFF:
+    case RGB_EFF_CMD:
         return zmk_rgb_underglow_cycle_effect(1);
-    case RGB_EFR:
+    case RGB_EFR_CMD:
         return zmk_rgb_underglow_cycle_effect(-1);
+    case RGB_HSV_COLOR_CMD:
```
Nicell, line 46:
```suggestion
    case RGB_HSB_COLOR_CMD:
```
Let's keep everything consistent with the current HSB. HSV define points to HSB already, so this should be a fine change 👍

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -103,16 +103,15 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
...
-                    }))                                                                            \
+                (static int kscan_gpio_enable_interrupts_##n(const struct device *dev) {           \
+                    return kscan_gpio_config_interrupts(kscan_gpio_input_devices_##n(dev),         \
+                                                        kscan_gpio_input_configs_##n(dev),         \
+                                                        INST_INPUT_LEN(n), GPIO_INT_LEVEL_ACTIVE); \
+                } static int kscan_gpio_disable_interrupts_##n(const struct device *dev) {         \
+                    return kscan_gpio_config_interrupts(kscan_gpio_input_devices_##n(dev),         \
+                                                        kscan_gpio_input_configs_##n(dev),         \
+                                                        INST_INPUT_LEN(n), GPIO_INT_DISABLE);      \
+                }))                                                                                \
```
Nicell, line 114: I think this formatting change snuck in 😄

Nicell (changes requested):
Just a couple small changes. Looks pretty good otherwise.

app/include/dt-bindings/zmk/rgb.h
```diff
@@ -4,14 +4,29 @@
...
+#define RGB_HUD_CMD 2
+#define RGB_SAI_CMD 3
+#define RGB_SAD_CMD 4
+#define RGB_BRI_CMD 5
+#define RGB_BRD_CMD 6
+#define RGB_SPI_CMD 7
+#define RGB_SPD_CMD 8
+#define RGB_EFF_CMD 9
+#define RGB_EFR_CMD 10
+#define RGB_HSB_COLOR_CMD 11
```
Nicell, line 18: Minor nitpick, this is RGB_HSB_COLOR... where the defines below are RGB_COLOR_HSB. Could you consolidate that to one order of names?

app/src/rgb_underglow.c
```diff
@@ -346,6 +346,18 @@ int zmk_rgb_underglow_toggle() {
...
 }
 
+int zmk_rgb_underglow_set_hsb(uint16_t hue, uint8_t saturation, uint8_t brightness) {
+    if (hue > 360 || saturation > 100 || brightness > 100) {
+        return -ENOTSUP;
+    }
+
+    state.hue = hue;
+    state.saturation = saturation;
+    state.brightness = brightness;
```
Nicell, line 356: Are we sure we don't want to set the animation to solid color? I guess this makes sense if you want to keep breathing. Staying on one of the spectrum animations seems weird though. Maybe it could be an option somehow. Kconfig doesn't feel like the place for it though.

docs/docs/behaviors/lighting.md
```diff
@@ -42,10 +43,31 @@ The "RGB underglow" behavior completes an RGB action given on press.
 ### Behavior Binding
 
 - Reference: `&rgb_ug`
-- Parameter: The RGB action define, e.g. `RGB_TOG` or `RGB_BRI`
+- Parameter #1: The RGB action define, e.g. `RGB_TOG` or `RGB_BRI`
+- Parameter #1: Only applies to `RGB_COLOR_HSB` and is the HSB values of the color to set within parenthesis and separated by a common (see below for an example)
```
Nicell, line 47: The way `RGB_COLOR_HSB` is used doesn't really seem like a two parameter system to the user due to the macro. I'm not sure how this can be worded better, but the example definitely shows it well.

app/src/rgb_underglow.c
mcrosson: I waffled on this point for a bit and decided to leave it be ; if anything if you set the color to solid while a spectrum animation is running I'm inclined to say 'user error' or 'keymap bug'.

I don't know if this is tenable longer term but a config option would need to be through KConfig or similar `.conf` file and that seems like a kludge.

docs/docs/behaviors/lighting.md
mcrosson: It may not 'feel right' but it's in line with the rest of the documentation and how the underlying behavior system sees things. With the example I think this will be OK.

Nicell: Fair enough. I'd correct it to say Parameter #2 at least 👍

app/src/rgb_underglow.c
Nicell: Yeah this is fine as a start. Something for us to consider in the future.

Nicell (approved):
Looks good!

app/include/dt-bindings/zmk/rgb.h
```diff
@@ -4,14 +4,29 @@
...
+#define RGB_HUD RGB_HUD_CMD 0
+#define RGB_SAI RGB_SAI_CMD 0
+#define RGB_SAD RGB_SAD_CMD 0
+#define RGB_BRI RGB_BRI_CMD 0
+#define RGB_BRD RGB_BRD_CMD 0
+#define RGB_SPI RGB_SPI_CMD 0
+#define RGB_SPD RGB_SPD_CMD 0
+#define RGB_EFF RGB_EFF_CMD 0
+#define RGB_EFR RGB_EFR_CMD 0
+#define RGB_COLOR_HSB(h, s, v) RGB_COLOR_HSB_CMD(((h) << 16) + ((s) << 8) + (v))
```
petejohanson, line 31: Isn't this missing a space?

```suggestion
#define RGB_COLOR_HSB(h, s, v) RGB_COLOR_HSB_CMD (((h) << 16) + ((s) << 8) + (v))
```

petejohanson (commented):
In general, looks good. One comment on the defines for the keymap.

app/include/dt-bindings/zmk/rgb.h
mcrosson: Not according to clang-format -i....

If I add a space then re-run clang-format it nukes the space.

mcrosson:
bump?

app/include/dt-bindings/zmk/rgb.h
petejohanson: Ok, surprising, but I'll take it!

petejohanson (approved):

PR #518, opened 2020-12-15
Add general info on tests to run when updating the docker container

mcrosson:
Add some general information on tests to run when updating the Docker container to the documentation testing page.

### DIFF
--- a/docs/docs/development/tests.md
+++ b/docs/docs/development/tests.md
@@ -1,14 +1,16 @@
 ---
-title: Tests
-sidebar_label: Tests
+title: Testing
+sidebar_label: Testing
 ---
+
+## General Tests
 
 Running tests requires [native posix support](posix-board). Any folder under `/app/tests`
 containing `native_posix.keymap` will be selected when running `west test`.
 
 Run a single test with `west test <testname>`, like `west test tests/toggle-layer/normal`.
 
-## Creating a New Test Set
+### Creating a New Test Set
 
 1. Copy the test set that most closely resembles the tests you will be creating.
 2. Rename the newly created test set to the behavior you're testing e.g, toggle-layer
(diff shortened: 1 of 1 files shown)


innovaker:
Thanks for this @mcrosson.  It's important we have it recorded.

Perhaps we should document this in the zmk-docker project itself?

I feel like we might be able to automate a lot of it into:
https://github.com/zmkfirmware/zmk-docker/blob/main/.github/workflows/containers.yml
Much the same was as I'm hoping to integrate ZMK build/tests into that workflow in the future as it'll save us a lot of manual re-testing whenever the base image is bumped.

What do you think?

mcrosson:
> Perhaps we should document this in the zmk-docker project itself?

I feel like this should be part of the overall docs. There is only one spot for ZMK documentation and I feel like all the documentation should be in one place to facilitate discovery.


> 
> I feel like we might be able to automate a lot of it in:
> https://github.com/zmkfirmware/zmk-docker/blob/main/.github/workflows/containers.yml
> Much the same was as I'm hoping to integrate ZMK build/tests into that workflow in the future as it'll save us a lot of manual re-testing whenever the base image is bumped.
> 
> What do you think?

If it can check run status output against the build information, perhaps. My concern is that running a command just to trap the output for a check may increase container size / screw with caching some.

It may be smarter to do something like `docker run -it --rm /foo/bar` and comparing the output (to whatever degree is least fiddly for verification) IMO.

innovaker:
> If it can check run status output against the build information, perhaps. My concern is that running a command just to trap the output for a check may increase container size / screw with caching some.

I think you misunderstood - nothing in my previous message suggested adding it to the Dockerfile.  I was talking about the GitHub workflow (see link) which could effectively download the container and test each command after the image candidate has been published.

mcrosson:
> I think you misunderstood - nothing in my previous message suggested adding it to the Dockerfile. I was talking about the GitHub workflow (see link) which could effectively download the container and test each command after the image candidate has been published.

I definitely misunderstood. If the commands can be added to the final vetting of the container build that'd be ideal.

petejohanson:
@mcrosson @innovaker With the recent enhancements to the GH actions for `zmkfirmware/zmk-docker`, can this be closed?

mcrosson:
Possibly, I need to defer to @innovaker on this one

petejohanson:
We've automated most of this in the ZMK docker repo at this point. Closing, but feel free to reopen if I missed something. Thanks again!

PR #521, opened 2020-12-17
refactor(core): define usage page as uint16_t

innovaker:
Aligns with the HID specification.

Usage page values were sometimes declared as `uint8_t` and sometimes `uint16_t`.  This commit aligns all instances with the HID specification for consistency.

### DIFF
--- a/app/include/zmk/endpoints.h
+++ b/app/include/zmk/endpoints.h
@@ -18,4 +18,4 @@ int zmk_endpoints_select(enum zmk_endpoint endpoint);
 int zmk_endpoints_toggle();
 enum zmk_endpoint zmk_endpoints_selected();
 
-int zmk_endpoints_send_report(uint8_t usage_page);
+int zmk_endpoints_send_report(uint16_t usage_page);
--- a/app/include/zmk/events/keycode_state_changed.h
+++ b/app/include/zmk/events/keycode_state_changed.h
@@ -14,7 +14,7 @@
 
 struct keycode_state_changed {
     struct zmk_event_header header;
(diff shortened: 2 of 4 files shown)


petejohanson:
Do we actually have any usage page values that are this large? I'm not sure this is needed if it's not ever actually used.

innovaker:
> Do we actually have any usage page values that are this large? I'm not sure this is needed if it's not ever actually used.

Heh, I thought you'd ask that.  Not currently!  But not catering for bits 9 to 16 closes the door on any potential user enhancements such as vendor-defined usage pages.  We're serving the majority, but I'd suggest we also facilitate the minority.

Also, we should align it to one or the other, and given it's 32-bit architecture it might as well match the specification as it's unlikely to cost anything meaningful - I believe?

Moreover, this is just the groundwork to a more tangible refactor I'm working on which will let us pass usages around as one identity.  At time of writing it'll likely be one of these branches ...

---

https://github.com/innovaker/zmk/commits/refactor-unify-usage-struct
i.e.
```
struct hid_usage {
    uint16_t page;
    uint16_t id;
};

struct hid_usage usage;
```
---

https://github.com/innovaker/zmk/tree/refactor-unify-usage-u32
i.e.
```
uint32_t usage;
```

innovaker:
I've several branches waiting in the wings that use this as groundwork so please let me know if I haven't convinced you yet 😄

petejohanson (approved):
Thanks for your patience. Reviewed the HID stuff. Yeah, this makes sense to me.

Thanks!

PR #523, opened 2020-12-18
refactor: replace filename hyphens with underscores

innovaker:
Aligns *.h and *.c to underscore naming convention.

Note: this would be a minor breaking change for users with custom boards or shields.


innovaker:
I hit this whilst refactoring a related part of the system.  I wasn't sure if there was a rationale for having two conventions implementation.

petejohanson (changes requested):
This is a good call, but to avoid breaking external boards, we should preserve the old filename, but with a single line content that imports the proper named file, along with a warning/deprecation notice. Keep local build clean, help keep external boards/shields from breaking.

innovaker:
> This is a good call, but to avoid breaking external boards, we should preserve the old filename, but with a single line content that imports the proper named file, along with a warning/deprecation notice. Keep local build clean, help keep external boards/shields from breaking.

Good idea.  I've added these files:
- `kscan-mock.h`
- `matrix-transform.h`

And tested them by stripping out the commits for the boards, shields, and tests here:
- https://github.com/innovaker/zmk/actions/runs/437888743
- https://github.com/innovaker/zmk/actions/runs/437888742

app/include/dt-bindings/zmk/kscan-mock.h
```diff
@@ -4,11 +4,4 @@
...
 
-#pragma once
-
-#define ZMK_MOCK_IS_PRESS(v) ((v & (0x01 << 31)) != 0)
-#define ZMK_MOCK_PRESS(row, col, msec) (row + (col << 8) + (msec << 16) + (0x01 << 31))
-#define ZMK_MOCK_RELEASE(row, col, msec) (row + (col << 8) + (msec << 16))
-#define ZMK_MOCK_ROW(v) (v & 0xFF)
-#define ZMK_MOCK_COL(v) ((v >> 8) & 0xFF)
-#define ZMK_MOCK_MSEC(v) ((v & ~(0x01 << 31)) >> 16)
+#include "kscan_mock.h"
```
petejohanson, line 7: Can we add a warning above this: https://stackoverflow.com/a/3826868

app/include/dt-bindings/zmk/matrix-transform.h
```diff
@@ -4,7 +4,4 @@
  * SPDX-License-Identifier: MIT
  */
 
-#define KT_ROW(item) (item >> 8)
-#define KT_COL(item) (item & 0xFF)
-
-#define RC(row, col) (((row) << 8) + (col))
\ No newline at end of file
+#include "matrix_transform.h"
```
petejohanson, line 7: Ditto

petejohanson (changes requested):

app/include/dt-bindings/zmk/kscan-mock.h
innovaker: Something like this?
`#warning "'kscan-mock.h' has been deprecated and replaced by 'kscan_mock.h'"`

app/include/dt-bindings/zmk/matrix-transform.h
innovaker: `#warning "'matrix-transform.h' has been deprecated and replaced by 'matrix_transform.h'"`

app/include/dt-bindings/zmk/kscan-mock.h
innovaker: >  along with a warning/deprecation notice
apologies for missing that btw

innovaker: Done, tested here:
- https://github.com/innovaker/zmk/actions/runs/438462011
- https://github.com/innovaker/zmk/actions/runs/438462009

app/include/dt-bindings/zmk/matrix-transform.h
innovaker: Ditto above.

petejohanson (approved):
Thanks!

PR #527, opened 2020-12-19
Eek

MangoIV:
- added the board shield for the eek! keyboard
- added a basic keymap for the eek! keyboard

### DIFF
--- a/docs/docs/development/setup.md
+++ b/docs/docs/development/setup.md
@@ -280,7 +280,7 @@ platform.
 To build firmwares for the ARM architecture (all supported MCUs/keyboards at this point), you'll need to install the Zephyr™ ARM SDK to your system:
 
 ```
-export ZSDK_VERSION=0.11.2
+export ZSDK_VERSION=0.11.4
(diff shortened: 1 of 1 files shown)


app/boards/shields/eek/Kconfig.defconfig
```diff
@@ -0,0 +1,6 @@
+if SHIELD_EEK
+
+config ZMK_KEYBOARD_NAME
+    default "zmk~eek!"
```
joelspadin, line 4: To be consistent with the other shields in the repo, the keyboard name shouldn't contain "zmk".

Maybe also capitalize it to match this? https://www.cbkbd.com/product/eek-keyboard

joelspadin (changes requested):
I'm not sure exactly what happened here, but it looks like you've somehow merged two copies of the entire repo together. There should be about 14 commits in this pull request, not over 1000.

I and/or others can help untangle this over Discord.

MangoIV:
Ouff and i was wondering why it showed so many commits. It might be because I accidentally commuted a large file and removed it from the commit history 
I will clean that up. (Including your other remark. Thanks for the review

Issue #525, opened 2020-12-18
Add "Overview" page as the first link under "Behaviors" in docs


petejohanson:
Currently, to get sense of all the behaviors you can use in ZMK, you need to read *every single* behavior details page.

We should add a new "Behavior Overview" page that has a table showing each possible behavior, with description and one or more examples, and a link to the detailed page to learn more.

PR #528, opened 2020-12-19
fix(docs): Tweak overview of behaviors.

petejohanson:
* New shield behavior docs are slightly out of date.

### DIFF
--- a/docs/docs/development/new-shield.md
+++ b/docs/docs/development/new-shield.md
@@ -364,11 +364,12 @@ The two `#include` lines at the top of the keymap are required in order to bring
 
 ### Keymap Behaviors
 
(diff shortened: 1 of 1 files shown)


innovaker:
May I suggest we take this approach instead?

https://github.com/innovaker/zmk/commits/docs-behaviors-overview-example
( It's just a PoC so I haven't completed the list.  Moreover, a succinct table might be more readable )

It would minimize on future maintenance and it's something we can farm out.

innovaker:
I just spotted #525 👍

petejohanson:
@innovaker I'm happy w/ your approach/POC, definitely. I will close this once you get that other PR open.

Issue #509, opened 2020-12-12
Add an event for behavior triggers


okke-formsma:
We currently have the following events

* position_state_changed
    * keycode_state_changed
    * layer_state_changed

What is missing from this event flow are behavior bindings. This means they can not be observed like other events.

* position_state_changed
  * behavior_state_changed
    * keycode_state_changed
    * layer_state_changed

This would simplify the signature of `static int on_keymap_binding_released(struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event)`. 

```
struct zmk_behavior_binding_event {
    char *behavior_dev;
    u32_t param1;
    u32_t param2;
    bool state;
    int layer;
    u32_t position;
    s64_t timestamp;
};
```

The return value of on_keymap_binding_released is currently used for deciding if the behavior is transparent (e.g. drop down to the next active layer).  The event return values have to change a little;  a 0 return value would mean the behavior is trnasparent, and `ZMK_EV_EVENT_HANDLED` and `ZMK_EV_EVENT_CAPTURED` would mean the behavior is not transparent.

PR #532, opened 2020-12-19
Introduce behavior-state-changed event

okke-formsma:
This is a prototype for a behavior-state-changed event, as suggested in https://github.com/zmkfirmware/zmk/issues/509

* other events can listen to behavior-state-changed events
* the behavior-binding functions now can capture the behavior-state-changed event, whereas the current lifetime of *bindings is not guaranteed


app/src/behavior.c
```diff
@@ -0,0 +1,53 @@
...
+    if (ret < 0) {
+        return ret;
+    }
+    switch (ret) {
+    case ZMK_BEHAVIOR_TRANSPARENT:
+        return ZMK_EV_EVENT_BUBBLE;
+    case ZMK_BEHAVIOR_OPAQUE:
+        return ZMK_EV_EVENT_HANDLED;
+    case ZMK_BEHAVIOR_CAPTURED:
+        return ZMK_EV_EVENT_CAPTURED;
```
okke-formsma, line 38: This mapping is a little ugly, but the zero (0) return value for events is 'bubble' and the zero return value for behaviors is 'opaque'. The code of behaviors and events assumes these return values often when returning output from other functions.

innovaker:
See #523.

okke-formsma:

- position_state_changed
  - keycode_state_changed
  - layer_state_changed

- position_state_changed
  - behavior_state_changed
    - keycode_state_changed
    - layer_state_changed

This PR adds the beahvior_state_changed event which is invoked just before an event is invoked.

Issue #451, opened 2020-11-30
Sticky keys: quick-release and slow-release.


okke-formsma:
In some cases, a sticky key should release just after the modified key is pressed. In other cases, it should stay active until the other key is released. 

An example of the first use case, where someone uses a sticky-keys layer and wants to drop down to the base layer to type a letter
* `tap &sl STICK, <access layer>, tap &sk LEFT_SHIFT, <release layer>, tap &kp A`

An example of the second use case, where multiple sticky keys add up:
* `tap &sk LEFT_SHIFT, tap &sk LEFT_CONTROL, tap &kp A`

So we need a setting on a sticky key whether it should release quickly or slowly. Default for `&sl` should be quick, default for `&sk` should be slow.

okke-formsma:
fixed in my branch `feature/sticky-key-quick-release`, waiting for #450 to be merged before creating a PR.

PR #537, opened 2020-12-21
feature(sticky_keys): add quick-release

okke-formsma:
when quick-release is enabled, the sticky key is released immediately after another key is pressed, instead of being released when the 'modified' key is released.

This will be the default behavior for sticky layers, so the following sequence works as expected:
* sticky layer tapped on base layer
* sticky shift tapped on first layer
* A key tapped on base layer

Without quick-release, the sticky layer would stay active until the sticky key was released, making it impossible to press any keys on the base layer:
* sticky layer tapped on base layer
* sticky shift tapped on first layer
* (unable to press anything on base layer because sticky layer is still active)

Also see https://github.com/zmkfirmware/zmk/issues/451


petejohanson (approved):

petejohanson:
@okke-formsma needs rebase.

okke-formsma:
@petejohanson Rebased!

PR #538, opened 2020-12-21
ci(build): ignore failing upload-artifacts step

okke-formsma:
Ignore failed "archive artifacts" steps which are making lots of builds fail. This is needed because currently many "archive artifact" steps failing with an "ECONNRESET" error.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,3 +112,4 @@ jobs:
           path: |
             build/zephyr/zmk.hex
             build/zephyr/zmk.uf2
+        continue-on-error: true


innovaker (changes requested):
Looks fine, but please add the reasoning to the commit body rather than just the PR.  The PR is more likely to get lost over time (or hard to find), whereas the commit is eternal and easily accessible by anyone who has a checkout.

innovaker (approved):
Thanks!

I don't think you mean this though ...
> There are many build steps failing due to

perhaps ...
> The upload-artifacts step is frequently failing due to

innovaker:
I'd also suggest adding the bug tracking issue as a footer so we don't have to dig for it in the future, i.e.
`See: actions/upload-artifact#116`

petejohanson (approved):

PR #539, opened 2020-12-22
Handle events from peripherals on the main work thread, avoid concurrent GATT notifications to avoid locking up central side..

petejohanson:
Resurrecting this change, since we're still having occasional failures w/ crashes on splits, which I believe could be related (but not fully fixed) to this.

Previously on #466 it was noted that this caused failures to sleep, but I cannot reproduce that here. Setting a low sleep timeout, and my Kyria properly sleeps and awakes from deep sleep.

This now has fixes for splits, and for non-splits, that have encountered spam/crash issues.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -102,7 +102,7 @@ config ZMK_SPLIT
 
 if ZMK_SPLIT
 
-config ZMK_SPLIT_BLE
+menuconfig ZMK_SPLIT_BLE
 	bool "Split keyboard support via BLE transport"
 	depends on ZMK_BLE
 	default y
@@ -124,6 +124,18 @@ config ZMK_SPLIT_BLE_CENTRAL_POSITION_QUEUE_SIZE
 endif
 
 if !ZMK_SPLIT_BLE_ROLE_CENTRAL
+
+config ZMK_SPLIT_BLE_PERIPHERAL_STACK_SIZE
+	int "BLE split peripheral notify thread stack size"
+	default 512
+
+config ZMK_SPLIT_BLE_PERIPHERAL_PRIORITY
+	int "BLE split peripheral notify thread priority"
+	default 5
+
+config ZMK_SPLIT_BLE_PERIPHERAL_POSITION_QUEUE_SIZE
+	int "Max number of key position state events to queue to send to the central"
+	default 10
 
 config ZMK_USB
 	default n
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -41,7 +41,7 @@ config USB_NUMOF_EP_WRITE_RETRIES
 #ZMK_USB
 endif
 
-config ZMK_BLE
+menuconfig ZMK_BLE
 	bool "BLE (HID over GATT)"
 	select BT
 	select BT_SMP
@@ -57,6 +57,22 @@ if ZMK_BLE
 
 config SYSTEM_WORKQUEUE_STACK_SIZE
 	default 2048
+
+config ZMK_BLE_THREAD_STACK_SIZE
+	int "BLE notify thread stack size"
+	default 512
+
+config ZMK_BLE_THREAD_PRIORITY
+	int "BLE notify thread priority"
+	default 5
+
+config ZMK_BLE_KEYBOARD_REPORT_QUEUE_SIZE
+	int "Max number of keyboard HID reports to queue for sending over BLE"
+	default 20
+
+config ZMK_BLE_CONSUMER_REPORT_QUEUE_SIZE
+	int "Max number of consumer HID reports to queue for sending over BLE"
+	default 5
 
 config ZMK_BLE_CLEAR_BONDS_ON_START
 	bool "Configuration that clears all bond information from the keyboard on startup."
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -94,10 +94,18 @@ config ZMK_SPLIT_BLE
 
 if ZMK_SPLIT_BLE
 
-config ZMK_SPLIT_BLE_ROLE_CENTRAL
+menuconfig ZMK_SPLIT_BLE_ROLE_CENTRAL
 	bool "Central"
 	select BT_CENTRAL
 	select BT_GATT_CLIENT
+
+if ZMK_SPLIT_BLE_ROLE_CENTRAL
+
+config ZMK_SPLIT_BLE_CENTRAL_POSITION_QUEUE_SIZE
+	int "Max number of key position state events to queue when received from peripherals"
+	default 5
+
+endif
 
 if !ZMK_SPLIT_BLE_ROLE_CENTRAL
 
--- a/app/src/hid.c
+++ b/app/src/hid.c
@@ -55,7 +55,9 @@ int zmk_hid_unregister_mod(zmk_mod_t modifier) {
             continue;                                                                              \
         }                                                                                          \
         keyboard_report.body.keys[idx] = val;                                                      \
-        break;                                                                                     \
+        if (val) {                                                                                 \
+            break;                                                                                 \
+        }                                                                                          \
     }
 
 #define TOGGLE_CONSUMER(match, val)                                                                \
@@ -64,7 +66,9 @@ int zmk_hid_unregister_mod(zmk_mod_t modifier) {
             continue;                                                                              \
         }                                                                                          \
         consumer_report.body.keys[idx] = val;                                                      \
-        break;                                                                                     \
+        if (val) {                                                                                 \
+            break;                                                                                 \
+        }                                                                                          \
     }
 
 int zmk_hid_implicit_modifiers_press(zmk_mod_flags_t implicit_modifiers) {
--- a/app/src/hog.c
+++ b/app/src/hog.c
@@ -5,6 +5,7 @@
  */
 
 #include <settings/settings.h>
+#include <init.h>
 
 #include <logging/log.h>
 
@@ -156,28 +157,115 @@ struct bt_conn *destination_connection() {
     return conn;
 }
 
+K_THREAD_STACK_DEFINE(hog_q_stack, CONFIG_ZMK_BLE_THREAD_STACK_SIZE);
+
+struct k_work_q hog_work_q;
+
+K_MSGQ_DEFINE(zmk_hog_keyboard_msgq, sizeof(struct zmk_hid_keyboard_report_body),
+              CONFIG_ZMK_BLE_KEYBOARD_REPORT_QUEUE_SIZE, 4);
+
+void send_keyboard_report_callback(struct k_work *work) {
+    struct zmk_hid_keyboard_report_body report;
+
+    while (k_msgq_get(&zmk_hog_keyboard_msgq, &report, K_NO_WAIT) == 0) {
+        struct bt_conn *conn = destination_connection();
+        if (conn == NULL) {
+            return;
+        }
+
+        struct bt_gatt_notify_params notify_params = {
+            .attr = &hog_svc.attrs[5],
+            .data = &report,
+            .len = sizeof(report),
+        };
+
+        int err = bt_gatt_notify_cb(conn, &notify_params);
+        if (err) {
+            LOG_ERR("Error notifying %d", err);
+        }
+
+        bt_conn_unref(conn);
+    }
+}
+
+K_WORK_DEFINE(hog_keyboard_work, send_keyboard_report_callback);
+
 int zmk_hog_send_keyboard_report(struct zmk_hid_keyboard_report_body *report) {
-    struct bt_conn *conn = destination_connection();
-    if (conn == NULL) {
-        return -ENOTCONN;
+    int err = k_msgq_put(&zmk_hog_keyboard_msgq, report, K_MSEC(100));
+    if (err) {
+        switch (err) {
+        case -EAGAIN: {
+            LOG_WRN("Keyboard message queue full, popping first message and queueing again");
+            struct zmk_hid_keyboard_report_body discarded_report;
+            k_msgq_get(&zmk_hog_keyboard_msgq, &discarded_report, K_NO_WAIT);
+            return zmk_hog_send_keyboard_report(report);
+        }
+        default:
+            LOG_WRN("Failed to queue keyboard report to send (%d)", err);
+            return err;
+        }
     }
 
-    LOG_DBG("Sending to NULL? %s", conn == NULL ? "yes" : "no");
+    k_work_submit_to_queue(&hog_work_q, &hog_keyboard_work);
 
-    int err = bt_gatt_notify(conn, &hog_svc.attrs[5], report,
-                             sizeof(struct zmk_hid_keyboard_report_body));
-    bt_conn_unref(conn);
-    return err;
+    return 0;
 };
+
+K_MSGQ_DEFINE(zmk_hog_consumer_msgq, sizeof(struct zmk_hid_consumer_report_body),
+              CONFIG_ZMK_BLE_CONSUMER_REPORT_QUEUE_SIZE, 4);
+
+void send_consumer_report_callback(struct k_work *work) {
+    struct zmk_hid_consumer_report_body report;
+
+    while (k_msgq_get(&zmk_hog_consumer_msgq, &report, K_NO_WAIT) == 0) {
+        struct bt_conn *conn = destination_connection();
+        if (conn == NULL) {
+            return;
+        }
+
+        struct bt_gatt_notify_params notify_params = {
+            .attr = &hog_svc.attrs[10],
+            .data = &report,
+            .len = sizeof(report),
+        };
+
+        int err = bt_gatt_notify_cb(conn, &notify_params);
+        if (err) {
+            LOG_DBG("Error notifying %d", err);
+        }
+
+        bt_conn_unref(conn);
+    }
+};
+
+K_WORK_DEFINE(hog_consumer_work, send_consumer_report_callback);
 
 int zmk_hog_send_consumer_report(struct zmk_hid_consumer_report_body *report) {
-    struct bt_conn *conn = destination_connection();
-    if (conn == NULL) {
-        return -ENOTCONN;
+    int err = k_msgq_put(&zmk_hog_consumer_msgq, report, K_MSEC(100));
+    if (err) {
+        switch (err) {
+        case -EAGAIN: {
+            LOG_WRN("Consumer message queue full, popping first message and queueing again");
+            struct zmk_hid_consumer_report_body discarded_report;
+            k_msgq_get(&zmk_hog_consumer_msgq, &discarded_report, K_NO_WAIT);
+            return zmk_hog_send_consumer_report(report);
+        }
+        default:
+            LOG_WRN("Failed to queue consumer report to send (%d)", err);
+            return err;
+        }
     }
 
-    int err = bt_gatt_notify(conn, &hog_svc.attrs[10], report,
-                             sizeof(struct zmk_hid_consumer_report_body));
-    bt_conn_unref(conn);
-    return err;
+    k_work_submit_to_queue(&hog_work_q, &hog_consumer_work);
+
+    return 0;
 };
+
+int zmk_hog_init(const struct device *_arg) {
+    k_work_q_start(&hog_work_q, hog_q_stack, K_THREAD_STACK_SIZEOF(hog_q_stack),
+                   CONFIG_ZMK_BLE_THREAD_PRIORITY);
+
+    return 0;
+}
+
+SYS_INIT(zmk_hog_init, APPLICATION, CONFIG_ZMK_BLE_INIT_PRIORITY);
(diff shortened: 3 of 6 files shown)


app/src/split/bluetooth/central.c
```diff
@@ -33,6 +33,29 @@ static struct bt_uuid_128 uuid = BT_UUID_INIT_128(ZMK_SPLIT_BT_SERVICE_UUID);
...
+
+K_MSGQ_DEFINE(peripheral_event_msgq, sizeof(struct zmk_split_peripheral_event), 4, 4);
+
+void peripheral_event_work_callback(struct k_work *work) {
+    struct zmk_split_peripheral_event ev;
+    while (k_msgq_get(&peripheral_event_msgq, &ev, K_NO_WAIT) == 0) {
+        struct position_state_changed *pos_ev = new_position_state_changed();
+        pos_ev->position = ev.position;
+        pos_ev->state = ev.state;
+        pos_ev->timestamp = ev.timestamp;
```
okke-formsma, line 50: slightly unrelated (not something we necessarily have to pick up in this ticket) but maybe we should always use `create_position_state_changed(position, state, timestamp)` functions to create events so the compiler complains about all call sites when a new field is added?

Nicell:
Tested this on my nice!60. It does in fact stop the keypresses getting stuck and the board freezing up, but I also can't type faster than about 50WPM before a ton of my keypresses are dropped and it becomes annoying to type.

petejohanson:
> Tested this on my nice!60. It does in fact stop the keypresses getting stuck and the board freezing up, but I also can't type faster than about 50WPM before a ton of my keypresses are dropped and it becomes annoying to type.

Ugh. Interesting. Are you able to get me logs?

innovaker:
Current branch passed my end-2-end test suite.
- [e2e-2021-01-13T21-00-28-ble.log](https://github.com/zmkfirmware/zmk/files/5811071/e2e-2021-01-13T21-00-28-ble.log)
- [e2e-2021-01-13T21-00-28-usb.log](https://github.com/zmkfirmware/zmk/files/5811072/e2e-2021-01-13T21-00-28-usb.log)

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
+
+    int err = k_msgq_put(&zmk_hog_keyboard_msgq, report, K_MSEC(100));
+    if (err) {
+        switch (err) {
+        case -EAGAIN:
+            LOG_WRN("keyboard message queue full, popping first message and queueing again");
+            k_msgq_get(&zmk_hog_keyboard_msgq, &discarded_report, K_NO_WAIT);
+            return zmk_hog_send_keyboard_report(report);
+        default:
+            LOG_WRN("Failed to queue consumer report to send (%d)", err);
```
innovaker, line 210:
```suggestion
            LOG_WRN("Failed to queue keyboard report to send (%d)", err);
```
?

innovaker:
I've completed extensive testing.

`main`
7ms: Appears to crash: HID reports stop arriving, logger stops logging, Linux input device left with stuck key(s).
8ms: Completes successfully.

`core/ble-split-central-notify-main-thread`
7ms: Doesn't crash, but drops reports.
8ms: Completes successfully.  This is also true with `HOG_REPORT_QUEUE_SIZE 1`.

These results confirm that the current fix works.

Also, having narrowed it down to the 7ms / 8ms boundary with repeatabable results, it suggests a queue might not be strictly necessary because we merely need to throttle and consolidate (= discard everything except latest in the case of keyboard and consumer reports) when a new event occurs below the connection interval (usually 7.5ms but negotiated during connect).  That said, a tiny workqueue still makes sense because it simplifies the thread implementation and is a catch-all safety-net for any delay issues.

Given the above, I guess the question is ... how lightweight can we make the thread(s)?

innovaker:
In the future, mouse reports will need a split consolidation strategy:
 - buttons will need a *latest* algorithm, as per keys.
 - `x`, `y` will need an accumulative algorithm.

So ultimately, the appropriate consolidation algorithm is per data item and it depends on the data item's control type and flags.  The latter are delcared in its report descriptor and defined by the [HID specification](https://www.usb.org/sites/default/files/hut1_21_0.pdf?#page=21).  This lends itself to a generalized consolidation algorithm.  I was antipating we'd hit this which is why those flags are currently noted in `hid_usage.h` as inline comments.

With that in mind, and given that mouse keys will likely need an internal consolidation state/algorithm (to handle multiple simultaneous inputs), I can see the consolidation (and maybe the throttling by extension) potentially moving into a layer above `hog`.  That'll make more sense as we move forward.  USB may theoretically need consolidation too depending on the rate setting.

petejohanson:
> I've completed extensive testing.
> 
> `main`
> 7ms: Appears to crash: HID reports stop arriving, logger stops logging, Linux input device left with stuck key(s).
> 8ms: Completes successfully.
> 
> `core/ble-split-central-notify-main-thread`
> 7ms: Doesn't crash, but drops reports.
> 8ms: Completes successfully. This is also true with `HOG_REPORT_QUEUE_SIZE 1`.
> 
> These results confirm that the current fix works.
> 
> Also, having narrowed it down to the 7ms / 8ms boundary with repeatabable results, it suggests a queue might not be strictly necessary because we merely need to throttle and consolidate (= discard everything except latest in the case of keyboard and consumer reports) when a new event occurs below the connection interval (usually 7.5ms but negotiated during connect). That said, a tiny workqueue still makes sense because it simplifies the thread implementation and is a catch-all safety-net for any delay issues.
> 
> Given the above, I guess the question is ... how lightweight can we make the thread(s)?

Thanks for the testing! As designed, there is only one new thread, for the dedicated hog workqueue. They are fairly light weight, and we can tune the stack size as needed later. The challenge for doing this *without* a queue, is that we are unlikely to know *when* we will be *blocked* by the call to GATT notify, and if we do that from the *system* workqueue, that will deadlock.

I think we stick with this fix for now, and there's some upstream Zephyr work that may alleviate this for the next Zephyr release.

petejohanson:
> In the future, mouse reports will need a split consolidation strategy:
> 
> * buttons will need a _latest_ algorithm, as per keys.
> * `x`, `y` will need an accumulative algorithm.
> 
> So ultimately, the appropriate consolidation algorithm is per data item and it depends on the data item's control type and flags. The latter are declared in its report descriptor and defined by the [HID specification](https://www.usb.org/sites/default/files/hut1_21_0.pdf?#page=21). This lends itself to a generalized consolidation algorithm. I was anticipating we'd hit this which is why those flags are currently noted in `hid_usage.h` as inline comments.
> 
> With that in mind, and given that mouse keys will likely need an internal consolidation state/algorithm (to handle multiple simultaneous inputs), I can see the consolidation (and maybe the throttling by extension) potentially moving into a layer above `hog`. That'll make more sense as we move forward. USB may theoretically need consolidation too depending on the rate setting.

Yes! I was considering this for a first pass of this for these reports as well, A follow up could identify subsequent HID reports that are all "key releases", and combine and re-insert into the queue, to help bring this down.

We'll need to figure out how to generalize this somewhere *before* or at least abstracted within the HOG and USB versions, I could see that being helpful.

app/src/hog.c
petejohanson: Good catch. Fixed as part of a rebase onto latest `main`.

innovaker:
There has been further discussion around how this might affect behaviors which spawn multiple events/reports "instantaneously":
https://discord.com/channels/719497620560543766/784523018050469939/799316212075987045

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
     return conn;
 }
 
+#define HOG_STACK_SIZE 512
+#define HOG_PRIORITY 5
```
innovaker, line 161: Does priority 5 have any particular meaning or is it arbitrary?

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
-    struct bt_conn *conn = destination_connection();
-    if (conn == NULL) {
-        return -ENOTCONN;
+    struct zmk_hid_keyboard_report_body discarded_report;
+
+    int err = k_msgq_put(&zmk_hog_keyboard_msgq, report, K_MSEC(100));
+    if (err) {
+        switch (err) {
+        case -EAGAIN:
+            LOG_WRN("keyboard message queue full, popping first message and queueing again");
```
innovaker, line 206:
```suggestion
            LOG_WRN("Keyboard message queue full, popping first message and queueing again");
```
aligns with consumer equivalent

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
 }
 
+#define HOG_STACK_SIZE 512
+#define HOG_PRIORITY 5
+
+K_THREAD_STACK_DEFINE(hog_q_stack, HOG_STACK_SIZE);
+
+struct k_work_q hog_work_q;
+
+#define HOG_REPORT_QUEUE_SIZE 5
```
innovaker, line 167: As [discussed](https://discord.com/channels/719497620560543766/784523018050469939/799337259840045167), we should probably increase the queue size.

app/src/hid.c
```diff
@@ -55,7 +55,9 @@ int zmk_hid_unregister_mod(zmk_mod_t modifier) {
             continue;                                                                              \
         }                                                                                          \
         keyboard_report.body.keys[idx] = val;                                                      \
-        break;                                                                                     \
+        if (val) {                                                                                 \
+            break;                                                                                 \
+        }                                                                                          \
```
innovaker, line 60: I think I appreciate what this does, but what precipitated it?  Please could you add the reasoning to the commit body.

app/src/split/bluetooth/service.c
```diff
@@ -45,12 +48,68 @@ BT_GATT_SERVICE_DEFINE(
     BT_GATT_DESCRIPTOR(BT_UUID_NUM_OF_DIGITALS, BT_GATT_PERM_READ, split_svc_num_of_positions, NULL,
                        &num_of_positions), );
 
+#define SERVICE_STACK_SIZE 512
+#define SERVICE_PRIORITY 5
```
innovaker, line 52: Should these be defined inline here or at the top or in `service.h`?

Should they be Kconfig variables?

app/src/split/bluetooth/service.c
```diff
@@ -45,12 +48,68 @@ BT_GATT_SERVICE_DEFINE(
...
+
+K_THREAD_STACK_DEFINE(service_q_stack, SERVICE_STACK_SIZE);
+
+struct k_work_q service_work_q;
+
+#define POSITION_STATE_QUEUE_SIZE 10
```
innovaker, line 58: As above.

app/src/split/bluetooth/service.c
```diff
@@ -45,12 +48,68 @@ BT_GATT_SERVICE_DEFINE(
...
+    k_work_submit_to_queue(&service_work_q, &service_position_notify_work);
+
+    return 0;
+}
+
 int zmk_split_bt_position_pressed(uint8_t position) {
     WRITE_BIT(position_state[position / 8], position % 8, true);
-    return bt_gatt_notify(NULL, &split_svc.attrs[1], &position_state, sizeof(position_state));
+    return send_position_state();
+    // return bt_gatt_notify(NULL, &split_svc.attrs[1], &position_state, sizeof(position_state));
```
innovaker, line 99:
```suggestion
```
leftover?

app/src/split/bluetooth/service.c
```diff
@@ -45,12 +48,68 @@ BT_GATT_SERVICE_DEFINE(
...
 }
 
 int zmk_split_bt_position_released(uint8_t position) {
     WRITE_BIT(position_state[position / 8], position % 8, false);
-    return bt_gatt_notify(NULL, &split_svc.attrs[1], &position_state, sizeof(position_state));
-}
\ No newline at end of file
+    return send_position_state();
+    // return bt_gatt_notify(NULL, &split_svc.attrs[1], &position_state, sizeof(position_state));
```
innovaker, line 105:
```suggestion
```
leftover?

innovaker (commented):
Looks great overall!

Just some nickpicking and clarifications ...
- c574fa7c2cf16bf5177727ecd2e3dd378fc84f9c
  - commit message: `FIxes` → `Fixes`
- 8193a01b3725c07d6ddee3e3316150701a0f761f
  - commit message "LE conn param update tweaks" → "improve LE params logging"
- One might argue the minor `POS_STATE_LEN` refactor is technically a separate concern/pre-commit but we're all friends here 🏳️

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
+            return;
+        }
+
+        struct bt_gatt_notify_params notify_params = {
+            .attr = &hog_svc.attrs[5],
+            .data = &report,
+            .len = sizeof(struct zmk_hid_keyboard_report_body),
+        };
+
+        err = bt_gatt_notify_cb(conn, &notify_params);
```
joelspadin, line 188:
```suggestion
        int err = bt_gatt_notify_cb(conn, &notify_params);
```
Combining declaration and initialization prevents accidentally using the variable before it's initialized.

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
+    while (k_msgq_get(&zmk_hog_consumer_msgq, &report, K_NO_WAIT) == 0) {
+        struct bt_conn *conn = destination_connection();
+        if (conn == NULL) {
+            return;
+        }
+
+        struct bt_gatt_notify_params notify_params = {
+            .attr = &hog_svc.attrs[10],
+            .data = &report,
+            .len = sizeof(struct zmk_hid_consumer_report_body),
```
joelspadin, line 235:
```suggestion
            .len = sizeof(report),
```
Same as above.

app/src/hog.c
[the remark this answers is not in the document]
petejohanson: Moved to Kconfig.

app/src/hog.c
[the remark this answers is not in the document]
petejohanson: Moved to Kconfig.

app/src/hog.c
petejohanson: See https://docs.zephyrproject.org/2.4.0/reference/kernel/scheduling/index.html#scheduling-algorithm

Thread priorities impact the scheduling algorithm. Moved to Kconfig, so we can easily tweak as we see fit.

app/src/split/bluetooth/central.c
```diff
@@ -33,6 +33,29 @@ static struct bt_uuid_128 uuid = BT_UUID_INIT_128(ZMK_SPLIT_BT_SERVICE_UUID);
 static struct bt_gatt_discover_params discover_params;
 static struct bt_gatt_subscribe_params subscribe_params;
 
+struct zmk_split_peripheral_event {
+    uint32_t position;
+    uint32_t state;
+    int32_t timestamp;
+};
+
+K_MSGQ_DEFINE(peripheral_event_msgq, sizeof(struct zmk_split_peripheral_event), 4, 4);
```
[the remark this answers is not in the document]
petejohanson: Done. Moved to Kconfig.

app/src/hog.c
petejohanson: Indeed. Fixed.

app/src/hog.c
petejohanson: Bumped the default in the Kconfig. Made two values, one for keyboard, one for consumer, and left the consumer one smaller. That should have less churn, but can be easily bumped later.

app/src/split/bluetooth/service.c
petejohanson: Moved to Kconfig.

app/src/split/bluetooth/service.c
petejohanson: Ditto. Now in Kconfig.

petejohanson: Moved to Kconfig.

app/src/split/bluetooth/service.c
petejohanson: Indeed. Moved to Kconfig.

app/src/hid.c
```diff
@@ -55,7 +55,9 @@ int zmk_hid_unregister_mod(zmk_mod_t modifier) {
             continue;                                                                              \
         }                                                                                          \
         keyboard_report.body.keys[idx] = val;                                                      \
-        break;                                                                                     \
+        if (val) {                                                                                 \
```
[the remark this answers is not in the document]
petejohanson: Arbitrary. I'm happy to switch to inline functions.

app/src/hid.c
petejohanson: Added a bit more detail to the commit message.

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
+    k_work_submit_to_queue(&hog_work_q, &hog_keyboard_work);
 
-    int err = bt_gatt_notify(conn, &hog_svc.attrs[5], report,
-                             sizeof(struct zmk_hid_keyboard_report_body));
-    bt_conn_unref(conn);
-    return err;
+    return 0;
 };
 
+K_MSGQ_DEFINE(zmk_hog_consumer_msgq, sizeof(struct zmk_hid_consumer_report_body),
```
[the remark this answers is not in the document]
petejohanson: Yeah, I pondered just queueing the notify params, but those take a *pointer* to the notify data, so doing that won't snapshot the hid report structs like this implementation does. So we need a queue for each one, since queue's need to define a size for each entry.

That or some union crazy. I'd favor leaving as is for now, and possibly refactor later.

app/src/split/bluetooth/service.c
```diff
@@ -45,12 +48,68 @@ BT_GATT_SERVICE_DEFINE(
...
+        if (err) {
+            LOG_DBG("Error notifying %d", err);
+        }
+    }
+};
+
+K_WORK_DEFINE(service_position_notify_work, send_position_state_callback);
+
+int send_position_state() {
+    uint8_t discarded_state[POS_STATE_LEN];
```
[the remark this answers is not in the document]
petejohanson: Thanks. I never remember that.

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
+    }
+};
+
+K_WORK_DEFINE(hog_consumer_work, send_consumer_report_callback);
+
 int zmk_hog_send_consumer_report(struct zmk_hid_consumer_report_body *report) {
-    struct bt_conn *conn = destination_connection();
-    if (conn == NULL) {
-        return -ENOTCONN;
+    struct zmk_hid_consumer_report_body discarded_report;
```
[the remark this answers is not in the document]
petejohanson: Thanks. I never remember that.

app/src/hog.c
petejohanson: Indeed. Fixed.

app/src/hog.c
```diff
@@ -156,28 +157,120 @@ struct bt_conn *destination_connection() {
...
+        int err;
+        struct bt_conn *conn = destination_connection();
+        if (conn == NULL) {
+            return;
+        }
+
+        struct bt_gatt_notify_params notify_params = {
+            .attr = &hog_svc.attrs[5],
+            .data = &report,
+            .len = sizeof(struct zmk_hid_keyboard_report_body),
```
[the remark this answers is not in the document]
petejohanson: Indeed. Fixed.

app/src/hid.c
innovaker: ah, so it's the discards that leads to this scenario - yep, makes sense.  Thanks!

app/src/hid.c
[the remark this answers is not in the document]
innovaker: > Probably not something to change in this PR, but is there a reason these are macros instead of functions?

I was also wondering that for a future refactor PR.

innovaker:
e2e still passes after refactors 👍
- [e2e-2021-01-15T10-37-20-usb.log](https://github.com/zmkfirmware/zmk/files/5819971/e2e-2021-01-15T10-37-20-usb.log)
- [e2e-2021-01-15T10-37-20-ble.log](https://github.com/zmkfirmware/zmk/files/5819972/e2e-2021-01-15T10-37-20-ble.log)

innovaker (commented):
Thanks for making the changes!

Only noticed one minor point.

@joelspadin, please approve this PR when you're happy and I'll then rubber stamp it for merging.

innovaker (approved):
Approved on the basis that @petejohanson will merge it once suggestions are accepted/rejected and @joelspadin is happy.

app/Kconfig
```diff
@@ -58,6 +58,22 @@ if ZMK_BLE
...
+	default 5
+
+config ZMK_BLE_KB_REPORT_QUEUE_SIZE
+	int "Max number of keyboard HID reports to queue for sending over BLE"
+	default 20
+
+config ZMK_BLE_CC_REPORT_QUEUE_SIZE
+	int "Max number of consumer HID reports to queue for sending over BLE"
+	default 5
+
```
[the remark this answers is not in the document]
petejohanson: Indeed. Accepted this suggestion for now, to be aligned w/ existing code.

joelspadin (approved):

PR #541, opened 2020-12-23
docs(codes): macOS Support: Brightness and media keycode compatability for MacOS Catalina 10.15.7

thedeagler:
### macOS version/build?
Catalina 10.15.7

### What codes did you test?
Does work
* `C_BRI_UP` = ⭐
* `C_BRI_DN` = ⭐
* `C_PREV` = ⭐
* `C_PP` = ⭐
* `C_NEXT` = ⭐
* `K_MUTE` = ⭐
* `C_VOL_DN` = ⭐
* `C_VOL_UP` = ⭐

Do not work
* `K_VOL_DN` = ❌
* `K_VOL_UP` = ❌

### How did you test?
Added the keycodes to my keymap, and flashed the board. Connected via bt to the laptop, and pressed the keys.

Can't think of anything, but happy to answer any questions ¯\\\_(ツ)\_/¯

Tried on two separate mbps running osx catalina 10.15.7. I can test out on a couple more computers later.


innovaker:
Awesome!  Thanks @thedeagler!

I'll put out a request to macOS users to verify your findings.

I'm not familiar with macOS ... does it have different media player applications?  If so, it's worth testing on each and adding footnotes to the documentation where applicable.  On Windows for instance, media key support varies between applications quite a lot.

adamcstephens:
Media keys on Mac are intercepted by the OS and applications just respond to the events from the OS. Applications can't even see the key presses themselves, so they should be consistent between apps.

I can confirm the volume and play/pause report is correct, but would need to map the others to test them.

tominabox1:
My findings on Big Sur 11.2 Beta agree with OP. 👍🏼

innovaker (approved):

PR #542, opened 2020-12-23
(feat/enhance) Add ws2812 to lilly58 + nice!nano

mcrosson:
Small update to facilitate using ws2812 strips with the lily58 pro and nice!nano

### DIFF
--- a/app/boards/shields/lily58/boards/nice_nano.overlay
+++ b/app/boards/shields/lily58/boards/nice_nano.overlay
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+&spi1 {
+	compatible = "nordic,nrf-spim";
+	status = "okay";
+	mosi-pin = <6>;
+	// Unused pins, needed for SPI definition, but not used by the ws2812 driver itself.
+	sck-pin = <5>;
+	miso-pin = <7>;
+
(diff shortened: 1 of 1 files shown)


app/boards/shields/lily58/boards/nice_nano.overlay
```diff
@@ -0,0 +1,34 @@
...
+	led_strip: ws2812@0 {
+		compatible = "worldsemi,ws2812-spi";
+		label = "WS2812";
+
+		/* SPI */
+		reg = <0>; /* ignored, but necessary for SPI bindings */
+		spi-max-frequency = <4000000>;
+
+		/* WS2812 */
+		chain-length = <1>; /* number of LEDs */
```
Nicell, line 24: Maybe set this to a reasonable/common value? I think I've seen kits sold with 5 for each side?

Nicell (commented):
Just one small note!

app/boards/shields/lily58/boards/nice_nano.overlay
mcrosson: The chain length is variable from what I've seen. I can make this 5 as a default just to be 'safe' as that seems like it may be a more common situation.

Nicell (approved):
Thanks!

innovaker (changes requested):
Code looks fine but please improve the readability of the commit message.

I've DM'ed you further advice with a link to the conventional commit specification.

innovaker (approved):

PR #547, opened 2020-12-28
Invoke behaviors with different "locality"

petejohanson:
With this change, behaviors can now have different "locality":

* Central (e.g. key press)
* Global (e.g. RGB to toggle on/off on both sides)
* Reset (e.g. to reset whatever device the "reset" binding was on, so you can have a reset on both sides, and it does what you expect)

This should makes split power and RGB management work, at least for the basics.


app/src/split/bluetooth/central.c
```diff
@@ -126,14 +130,25 @@ static uint8_t split_central_discovery_func(struct bt_conn *conn, const struct b
...
+        run_behavior_handle = bt_gatt_attr_value_handle(attr);
     } else {
         subscribe_params.notify = split_central_notify_func;
         subscribe_params.value = BT_GATT_CCC_NOTIFY;
         subscribe_params.ccc_handle = attr->handle;
 
         split_central_subscribe(conn);
 
-        return BT_GATT_ITER_STOP;
+        memcpy(&uuid, BT_UUID_DECLARE_128(ZMK_SPLIT_BT_CHAR_RUN_BEHAVIOR_UUID), sizeof(uuid));
```
joelspadin, line 143: Is it intended that split_central_process_connection and this function share the static uuid variable? It looks like everything here overwrites the UUID passed to bt_gatt_discover while split_central_process_connection does not. If that's not intended, maybe put separate static variables into each function to ensure they can't both access the same variable.

I'm not sure I fully understand what this code is doing yet, but if split_central_process_connection is always supposed to be discovering ZMK_SPLIT_BT_SERVICE_UUID, it looks like it actually is re-discovering the last thing you tried to discover.

app/src/keymap.c
```diff
@@ -134,24 +145,50 @@ int zmk_keymap_apply_position_state(int layer, uint32_t position, bool pressed,
...
+    if (err) {
+        LOG_ERR("Failed to get behavior locality %d", err);
+        return err;
+    }
+
+    switch (locality) {
+    case BEHAVIOR_LOCALITY_CENTRAL:
+        return invoke_locally(binding, event, pressed);
+    case BEHAVIOR_LOCALITY_EVENT_SOURCE:
+#if IS_ENABLED(CONFIG_ZMK_SPLIT) && IS_ENABLED(CONFIG_ZMK_BLE) &&                                  \
```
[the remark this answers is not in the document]
joelspadin: Or just `IS_BLE_CENTRAL`? Having both "split" and "central" is maybe a bit redundant since you can't have a central without a split.

app/west.yml
```diff
@@ -15,7 +15,6 @@ manifest:
         # TODO: Rename once upstream offers option like `exclude` or `denylist`
         name-blacklist:
           - ci-tools
-          - hal_atmel
```
Nicell, line 18: Was this change purposeful?

Nicell (commented):
I'm pretty excited for this change! Particularly I'd like to get someone who has RGB on a split that could test this out, as I just have plain split boards. Could we ask for testers in the Discord?

app/dts/behaviors/reset.dtsi
```diff
@@ -16,7 +16,7 @@
 
 		bootloader: behavior_reset_dfu {
 			compatible = "zmk,behavior-reset";
-			label = "BOOTLOADER_RESET";
+			label = "BOOTLOAD";
```
innovaker, line 19: Why does this need to be changed?  I'm missing something.

app/include/drivers/behavior.h
```diff
@@ -86,7 +127,7 @@ static inline int z_impl_behavior_keymap_binding_released(struct zmk_behavior_bi
  * @brief Handle the a sensor keymap binding being triggered
  * @param dev Pointer to the device structure for the driver instance.
  * @param sensor Pointer to the sensor device structure for the sensor driver instance.
- * @param param1 User parameter specified at time of behavior binding.
+ * @param param1 User parameter specified at time of behavior binding. diligence
```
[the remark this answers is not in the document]
innovaker: comic value?

app/src/keymap.c
```diff
@@ -134,24 +145,50 @@ int zmk_keymap_apply_position_state(int layer, uint32_t position, bool pressed,
     behavior = device_get_binding(binding->behavior_dev);
 
     if (!behavior) {
-        LOG_DBG("No behavior assigned to %d on layer %d", position, layer);
+        LOG_WRN("No behavior assigned to %d on layer %d", position, layer);
```
innovaker, line 148: Did this fall out of testing?

app/dts/behaviors/reset.dtsi
joelspadin: I think these strings are sent as part of the message to tell the peripheral to invoke a behavior?

petejohanson: Correct, the call to the peripheral passes behavior label (the unambiguous identifier we have for these things), the two params, and the state.

These had to be shortened to fit in a normal MTU for BLE.

app/src/keymap.c
petejohanson: Crept in, yeah. Was easier to spot in log output, and felt like a better level.

app/west.yml
petejohanson: No, this snuck in accidentally. We do need to restore this for the wired SAMD21 stuff I'm working on, but shouldn't be this branch.

app/src/keymap.c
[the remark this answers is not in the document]
petejohanson: Yeah, indeed. Fixed.

okke-formsma:
The code looks fine to me. There are some comments by @joelspadin that haven't been addressed yet @petejohanson.

app/src/split/bluetooth/central.c
petejohanson: This code is cribbed from a discovery example from Zephyr. As I read it, they're trying to reduce memory usage by re-using several variables as they discover/iterate the attributes. Let me review these changes again before we move forward, be sure we're still happy with how this is shaping up.

app/include/drivers/behavior.h
```diff
@@ -35,6 +44,28 @@ __subsystem struct behavior_driver_api {
...
+                                    enum behavior_locality *locality);
+
+static inline int z_impl_behavior_get_locality(const struct device *behavior,
+                                               enum behavior_locality *locality) {
+    if (behavior == NULL) {
+        return -EINVAL;
+    }
+
+    const struct behavior_driver_api *api = (const struct behavior_driver_api *)behavior->api;
+    memcpy(locality, &api->locality, sizeof(enum behavior_locality));
```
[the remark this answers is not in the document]
petejohanson: Fixed. Thanks!

app/include/drivers/behavior.h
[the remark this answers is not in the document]
petejohanson: Thanks. Removed.

app/west.yml
petejohanson: I don't see this here any more. This was from a different commit to `main`.

app/src/split/bluetooth/central.c
```diff
@@ -315,10 +330,30 @@ static struct bt_conn_cb conn_callbacks = {
...
 
+int zmk_split_bt_invoke_behavior(const bt_addr_le_t *source, struct zmk_behavior_binding *binding,
+                                 struct zmk_behavior_binding_event event, bool state) {
+    struct zmk_split_run_behavior_payload payload = {.data = {
+                                                         .param1 = binding->param1,
+                                                         .param2 = binding->param2,
+                                                         .position = event.position,
+                                                         .state = state,
+                                                     }};
+    strncpy(payload.behavior_dev, binding->behavior_dev, ZMK_SPLIT_RUN_BEHAVIOR_DEV_LEN);
```
joelspadin, line 341: Note that strncpy() does *not* guarantee the copied string is null terminated. As long as the source and destination are the same size (and the source is null terminated) it should be safe, but strlcpy() would be better if that's available.

It's not a standard C library function, but it is a pretty common extension. If it's not available, I found some public domain implementations at https://lwn.net/Articles/33812/

Another quick fix is to strncpy() length - 1 characters and then manually set the last byte to '\0'.

app/src/split/bluetooth/central.c
petejohanson: @joelspadin Ok, I've cleaned this up a *lot*, and in the process, I believe found and fixed a split bug that would sometimes cause peripherals not to get subscribed to properly if the central is kept on and the peripheral is reset and reconnected to.

The changes depend on the fix in zmkfirmware/zephyr#3 being merged first, then this can be tested.

app/src/split/bluetooth/central.c
petejohanson: Good catch. Fixed with manual last byte setting approach, last I looked `strlcpy` wasn't there.

joric:
Need global &sleep shortcut as well.

mcrosson:
> 
> 
> Need global &sleep shortcut as well.

Sleep is over at https://github.com/zmkfirmware/zmk/pull/649 and will get some attention after this is merged.

app/src/split/bluetooth/service.c
```diff
@@ -24,12 +26,59 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 static uint8_t num_of_positions = ZMK_KEYMAP_LEN;
 static uint8_t position_state[POS_STATE_LEN];
 
+#define BEHAVIOR_LABEL_LEN 12
```
joelspadin, line 29: Unused?

app/src/split/bluetooth/service.c
```diff
@@ -24,12 +26,59 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+
+    // We run if:
+    // 1: We've gotten all the position/state/param data.
+    // 2: We have a null terminated string for the behavior device label.
+    if ((end_addr > sizeof(struct zmk_split_run_behavior_data)) &&
+        payload->behavior_dev[end_addr - sizeof(struct zmk_split_run_behavior_data)] == '\0') {
+        struct zmk_behavior_binding binding = {
+            .param1 = payload->data.param1,
+            .param2 = payload->data.param2,
+            .behavior_dev = payload->behavior_dev,
```
joelspadin, line 61: It would be good to document somewhere as obvious as possible that if you're writing a behavior, you cannot count on binding->behavior_dev being valid after a handler returns, and if you need to save that data you should use `device_get_binding()` and save the device pointer instead (I think that's safe?) or `strdup` the string and free it when you're done with it.

app/src/split/bluetooth/central.c
```diff
@@ -326,6 +351,27 @@ static struct bt_conn_cb conn_callbacks = {
...
 
+int zmk_split_bt_invoke_behavior(const bt_addr_le_t *source, struct zmk_behavior_binding *binding,
+                                 struct zmk_behavior_binding_event event, bool state) {
+    struct zmk_split_run_behavior_payload payload = {.data = {
+                                                         .param1 = binding->param1,
+                                                         .param2 = binding->param2,
+                                                         .position = event.position,
+                                                         .state = state,
+                                                     }};
+    strncpy(payload.behavior_dev, binding->behavior_dev, ZMK_SPLIT_RUN_BEHAVIOR_DEV_LEN - 1);
```
joelspadin, line 362: It would be nice to verify that this didn't result in truncation. Maybe log an error with the name of the offending behavior?

Also consider using `sizeof(payload.behavior_dev)` instead of `ZMK_SPLIT_RUN_BEHAVIOR_DEV_LEN` so you don't have to change this if you change how the length of `behavior_dev` is determined.

Using strlcpy():
```c
// http://cantrip.org/strlcpy.c
size_t strlcpy(char *dst, const char *src, size_t size)
{
	const size_t len = strlen(src);
	if (size != 0) {
		memcpy(dst, src, (len > size - 1) ? size - 1 : len);
		dst[size - 1] = 0;
	}
	return len;
}

if (strlcpy(payload.behavior_dev, binding->behavior_dev, sizeof(payload.behavior_dev)) >= 
    sizeof(payload.behavior_dev)) {
    DBG_ERR("Truncated!");
}
```
As an addition to this code:
```c
if (strlen(binding->behavior_dev) >= sizeof(payload.behavior_dev)) {
    DBG_ERR("Truncated!");
}
```

app/src/split/bluetooth/service.c
```diff
@@ -24,12 +26,59 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+        return BT_GATT_ERR(BT_ATT_ERR_INVALID_OFFSET);
+    }
+
+    memcpy(payload + offset, buf, len);
+
+    // We run if:
+    // 1: We've gotten all the position/state/param data.
+    // 2: We have a null terminated string for the behavior device label.
+    if ((end_addr > sizeof(struct zmk_split_run_behavior_data)) &&
+        payload->behavior_dev[end_addr - sizeof(struct zmk_split_run_behavior_data)] == '\0') {
```
joelspadin, line 57: I think using `offsetof` would make it more obvious what this is doing, and it would handle changes to the layout of `zmk_split_run_behavior_payload` without needing to update this.

Might look something like this:
```c
const size_t behavior_dev_offset = offsetof(struct zmk_split_run_behavior_payload, behavior_dev);
if (end_addr > behavior_dev_offset && payload->behavior_dev[end_addr - behavior_dev_offset] == '\0') {
    ...
```

joelspadin (commented):
Looks good to me aside from some minor things (see comments).

karnadii:
if the RGB toggle on/off keycode is on the slave/peripheral, it only trigger the master. if the keycode is on the master/central. it will trigger both central and peripheral.

app/src/split/bluetooth/central.c
```diff
@@ -83,70 +89,85 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
...
+        return BT_GATT_ITER_CONTINUE;
+    }
+
+    LOG_DBG("Found split service");
+    discover_params.uuid = NULL;
+    discover_params.func = split_central_chrc_discovery_func;
+    discover_params.start_handle = attr->handle + 1;
+    discover_params.type = BT_GATT_DISCOVER_CHARACTERISTIC;
+
+    err = bt_gatt_discover(conn, &discover_params);
```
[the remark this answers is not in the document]
petejohanson: Indeed. Fixed.

app/src/split/bluetooth/central.c
```diff
@@ -83,70 +89,85 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
...
         return BT_GATT_ITER_STOP;
     }
 
     LOG_DBG("[ATTRIBUTE] handle %u", attr->handle);
 
-    if (!bt_uuid_cmp(discover_params.uuid, BT_UUID_DECLARE_128(ZMK_SPLIT_BT_SERVICE_UUID))) {
-        memcpy(&uuid, BT_UUID_DECLARE_128(ZMK_SPLIT_BT_CHAR_POSITION_STATE_UUID), sizeof(uuid));
-        discover_params.uuid = &uuid.uuid;
-        discover_params.start_handle = attr->handle + 1;
+    if (attr->user_data &&
```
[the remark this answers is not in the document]
petejohanson: Much nicer to move that check to an early guard clause instead. Good catch, thanks!

petejohanson:
> if the RGB toggle on/off keycode is on the slave/peripheral, it only trigger the master. if the keycode is on the master/central. it will trigger both central and peripheral. Demo https://imgur.com/a/yKuqD5v (the master is the right half)

Can you please capture the logs from both sides when doing this? Would help me try to diagnose what's up w/ that. Thanks.

karnadii:
@petejohanson here is the log. (the code is not updated to your latest commit)

## Log on central

pressing RGB toggle on central 
```
[00:04:29.903,076] <dbg> zmk.kscan_gpio_read_0: Sending event at 2,7 state on
[00:04:29.903,137] <dbg> zmk.zmk_kscan_process_msgq: Row: 2, col: 7, position: 43, pressed: true

[00:04:29.903,167] <dbg> zmk.zmk_keymap_apply_position_state: layer: 3 position: 43, binding name: RGB_UG
[00:04:29.903,198] <dbg> zmk.on_keymap_binding_convert_central_state_dependent_params: RGB relative convert to absolute (1/0)
[00:04:30.026,977] <dbg> zmk.kscan_gpio_read_0: Sending event at 2,7 state off
[00:04:30.027,038] <dbg> zmk.zmk_kscan_process_msgq: Row: 2, col: 7, position: 43, pressed: false

[00:04:30.027,069] <dbg> zmk.zmk_keymap_apply_position_state: layer: 3 position: 43, binding name: RGB_UG
[00:04:30.027,099] <dbg> zmk.on_keymap_binding_convert_central_state_dependent_params: RGB relative convert to absolute (2/0)
[00:04:30.440,979] <dbg> zmk.split_central_notify_func: [NOTIFICATION] data 0x2000ef7f length 16
[00:04:30.441,040] <dbg> zmk.peripheral_event_work_callback: Trigger key position state change for 62
[00:04:30.441,101] <dbg> zmk.zmk_keymap_apply_position_state: layer: 0 position: 62, binding name: MO
[00:04:30.441,131] <dbg> zmk.mo_keymap_binding_released: position 62 layer 3
[00:04:30.441,131] <dbg> zmk.set_layer_state: layer_changed: layer 3 state 0
[00:04:30.441,162] <dbg> zmk.set_layer_symbol: Layer changed to 0
```
pressing RGB toggle on peripheral (only trigger central)

```
[00:09:57.208,465] <dbg> zmk.split_central_notify_func: [NOTIFICATION] data 0x2000ef7f length 16
[00:09:57.208,557] <dbg> zmk.peripheral_event_work_callback: Trigger key position state change for 62
[00:09:57.208,587] <dbg> zmk.zmk_keymap_apply_position_state: layer: 0 position: 62, binding name: MO
[00:09:57.208,618] <dbg> zmk.mo_keymap_binding_pressed: position 62 layer 3
[00:09:57.208,618] <dbg> zmk.set_layer_state: layer_changed: layer 3 state 1
[00:09:57.208,648] <dbg> zmk.set_layer_symbol: Layer changed to 3
[00:09:58.573,486] <dbg> zmk.split_central_notify_func: [NOTIFICATION] data 0x2000ef7f length 16
[00:09:58.573,547] <dbg> zmk.peripheral_event_work_callback: Trigger key position state change for 0
[00:09:58.573,608] <dbg> zmk.zmk_keymap_apply_position_state: layer: 3 position: 0, binding name: RGB_UG
[00:09:58.573,638] <dbg> zmk.on_keymap_binding_convert_central_state_dependent_params: RGB relative convert to absolute (2/0)
[00:09:58.760,986] <dbg> zmk.split_central_notify_func: [NOTIFICATION] data 0x2000ef7f length 16
[00:09:58.761,047] <dbg> zmk.peripheral_event_work_callback: Trigger key position state change for 0
[00:09:58.761,077] <dbg> zmk.zmk_keymap_apply_position_state: layer: 3 position: 0, binding name: RGB_UG
[00:09:58.761,108] <dbg> zmk.on_keymap_binding_convert_central_state_dependent_params: RGB relative convert to absolute (1/0)
[00:09:58.925,964] <dbg> zmk.split_central_notify_func: [NOTIFICATION] data 0x2000ef7f length 16
[00:09:58.926,025] <dbg> zmk.peripheral_event_work_callback: Trigger key position state change for 62
[00:09:58.926,086] <dbg> zmk.zmk_keymap_apply_position_state: layer: 0 position: 62, binding name: MO
[00:09:58.926,116] <dbg> zmk.mo_keymap_binding_released: position 62 layer 3
[00:09:58.926,116] <dbg> zmk.set_layer_state: layer_changed: layer 3 state 0
[00:09:58.926,147] <dbg> zmk.set_layer_symbol: Layer changed to 0
[00:10:00.071,411] <dbg> zmk.bvd_sample_fetch: ADC raw 3819 ~ 2797 mV => 3943 mV
```

## Log on peripheral

pressing rgb toggle on central
```
[00:18:24.643,707] <dbg> zmk.split_listener:
[00:18:33.879,638] <dbg> zmk.kscan_gpio_read_0: Sending event at 4,6 state on
[00:18:33.879,699] <dbg> zmk.zmk_kscan_process_msgq: Row: 4, col: 6, position: 62, pressed: true

[00:18:33.879,730] <dbg> zmk.split_listener:
[00:18:34.589,599] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:18:34.589,599] <dbg> zmk.split_svc_run_behavior: INVOKE THE BEHAVIOR: RGB_UG with params 1 0
[00:18:34.727,081] <dbg> zmk.kscan_gpio_read_0: Sending event at 4,6 state off
[00:18:34.727,142] <dbg> zmk.zmk_kscan_process_msgq: Row: 4, col: 6, position: 62, pressed: false

[00:18:34.727,172] <dbg> zmk.split_listener:
[00:18:34.732,238] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:18:34.732,330] <dbg> zmk.split_svc_run_behavior: INVOKE THE BEHAVIOR: RGB_UG with params 2 0
[00:18:37.974,548] <dbg> zmk.kscan_gpio_read_0: Sending event at 4,6 state on
[00:18:37.974,578] <dbg> zmk.zmk_kscan_process_msgq: Row: 4, col: 6, position: 62, pressed: true

[00:18:37.974,609] <dbg> zmk.split_listener:
[00:18:38.579,895] <dbg> zmk.kscan_gpio_read_0: Sending event at 4,6 state off
[00:18:38.579,956] <dbg> zmk.zmk_kscan_process_msgq: Row: 4, col: 6, position: 62, pressed: false


```

pressing RGB toggle on peripheral (only trigger central)
```
[00:20:31.373,748] <dbg> zmk.split_listener:
[00:20:31.385,772] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:20:31.559,631] <dbg> zmk.kscan_gpio_read_0: Sending event at 0,0 state off
[00:20:31.559,692] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 0, pressed: false

[00:20:31.559,722] <dbg> zmk.split_listener:
[00:20:31.573,242] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:20:31.573,364] <dbg> zmk.split_svc_run_behavior: INVOKE THE BEHAVIOR: RGB_UG with params 2 0
[00:20:31.802,825] <dbg> zmk.kscan_gpio_read_0: Sending event at 4,6 state off
[00:20:31.802,886] <dbg> zmk.zmk_kscan_process_msgq: Row: 4, col: 6, position: 62, pressed: false

[00:20:31.802,917] <dbg> zmk.split_listener:
[00:20:34.039,581] <dbg> zmk.kscan_gpio_read_0: Sending event at 4,6 state on
[00:20:34.039,611] <dbg> zmk.zmk_kscan_process_msgq: Row: 4, col: 6, position: 62, pressed: true

```

Lenbok:
What's the current status of this PR?

petejohanson:
> What's the current status of this PR?

@Lenbok I need to revisit this after finding some deficiencies; In particular, triggering any behavior that invokes the peripheral side too quickly results in deadlocks on the left side.

I'm *hoping* that the Zephyr 2.5.0 upgrade in #736 has some BT stack improvements that will fix this, so I wanted to prep and test that before coming back to this one.

elpablete:
Is there any way I can help with this? I have a corne with underglow and nice!nanos

truonglong88:
Waiting for this too long... Appreciate your works.

joric:
@elpablete I'll test it gladly. I loove RGB.

krikun98:
Just to log this here: from my testing, this PR doesn't work when merged to `main` since the Zephyr 2.5 update.

edmondhuang:
Antoher common issue logged:
About left<->right data synchronise, can use exact value to align left and right RGB effect and brightness.
Take this as sample, if I open the left side and save the light effect #3, then open right side. However right side default is #1. Then they will never be synchronised after that, while I press MOD+ or MOD- for cycle switch light effect.
Sugguess don't use the +1/-1, and use exact value instead.

0nOff:
@petejohanson ur fix work for me (sofle rgb (v2.1) and nice!nano), good job! 
but if fast change hue, saturation or brightness, then keyboard freezing and need reset both splits

petejohanson:
> @petejohanson ur fix work for me (sofle rgb (v2.1) and nice!nano), good job! 
> but if fast change hue, saturation or brightness, then keyboard freezing and need reset both splits

This is the main remaining known issue with this PR. Fix for that is in the works.

petejohanson:
Ok, I've pushed fixes for the deadlock. Can everyone please test?

0nOff:
> Ok, I've pushed fixes for the deadlock. Can everyone please test?

Work for Sofle 2.1 rgb and nice!nano v2, no deadlock.

mayday9905:
Still deadlocks using nrfmicro_13. Setup from zero and compile in docker vscode

mayday9905:
It works correctly, good work

pfn:
I checked out and rebased this PR onto latest `main`, my experience:

1. reboot:
  * when rebooting the central side is able to restore last-rgb state from eeprom, the peripheral side keeps default behavior (static red)
  * cycling the rgb modes gets the peripheral side to sync and have the correct behavior, it stays in sync until reboot again
2. Using `&rgb_ug RGB_TOG` turns off OLED displays
  * OLED screens remain turned off until reset, toggling again will not turn them back on.

Outside of these behaviors, RGB, at least, works as desired for me.

uuhan:
It works for me to bootloader each part of my splited corne keyboard.

lewurm:
Just as another data point, RGB mode is in sync now for me too.

Are there any known blockers for this to be merged?

petejohanson:
> Just as another data point, RGB mode is in sync now for me too.
> 
> Are there any known blockers for this to be merged?

Yes, there were some reports of additional latency of inputs on the peripheral side, which I believe is tied to calls into the bluetooth controller (via HCI) to fetch the connection remote address. I'm almost done w/ some additional refactors to avoid this, and will let everyone know when this is ready for re-test.

ddudek:
> * OLED screens remain turned off until reset, toggling again will not turn them back on.
It was the same in the main branch when I tested, so I think not related to this PR.

Tested this and I can confirm, rgb works for kyria + nice!nano v2, although there's a noticable random delay on a slave side.

petejohanson:
I've just pushed a tweak to avoid calls to look up peripheral addresses constantly for events, which should hopefully address the reported added latency from peripheral events/keypresses.

Please test again!

Thanks all.

pfn:
Testing the new updates on this branch seems to work for me

lewurm:
> Testing the new updates on this branch seems to work for me

Another +1 from my side.  FWIW I never noticed the mentioned delay by others (nice!nano v2 on a custom dactyl build).

jlmkerr:
Also working for me. Many thanks.

filterpaper:
While `&bootloader` triggers consistently on the peripheral side, that keycode and `&reset` no longer triggers for central.

Percentnineteen:
&ext_power EP_TOG (and EP_OFF) only seem to affect central - peripheral remains powered for me.

pfn:
> &ext_power EP_TOG (and EP_OFF) only seem to affect central - peripheral remains powered for me.

This works for me as I use it to disable my oled and rgb, it does so on both sides as expected.

Percentnineteen:
> > &ext_power EP_TOG (and EP_OFF) only seem to affect central - peripheral remains powered for me.
> 
> This works for me as I use it to disable my oled and rgb, it does so on both sides as expected.

I don't have rgb - perhaps that is related? (Maybe it only works if rgb is enabled?)

Percentnineteen:
I've done a little more testing and still unable to get this to work properly. I've found the following:
1. if I checkout main and re-build, my &reset and &bootloader (no matter which side of the split has the codes) will trigger the central side only (as expected)
2. if I checkout this pr, &reset and &bootloader work correctly on the peripheral side but no longer do anything on the central side
3. also, with this pr, &ext_power EP_TOG, EP_ON and EP_OFF only cause the desired effect on the central side - the peripheral is left unaffected

I wonder if maybe there is another pr I need in concert with this one?

PedroDiogo:
I have just checked out this branch and things look good!

Tested:
- External power on and off. Both halves External power turn on and off at the same time.
- RBG lighting effect cycle. Both halves started off with different patterns but synchronised to the same pattern after I cycled it once.

Will daily drive this build for the next days and report back if I find anything.
(My build rebased `main` on this branch, no conflicts)

rbitanga:
It looks like I independently developed the same functionality this week and only discovered this PR when I was going through open issues to reference in a PR. The one thing I see missing is state synchronization once a peripheral connects to the central half (for instances where the peripheral was disconnected while a behavior was invoked.) I have an additional write only characteristic for the RGB state and export/import state methods for the rgb underglow subsystem. Synchronizing the state as opposed to running a behavior would allow the animation steps to remain in sync. I had minimal safety checks in the write callback opting to only check if the length matches the expected size of the state struct. For more safety, I was thinking of including checks for the firmware version from the DIS but I don't have that implemented.

I don't have anything implemented for explicitly synchronizing the external power state since I was relying on `#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_EXT_POWER)` and updated `rgb_settings_set` to have
```
#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_EXT_POWER)
    if (ext_power) {
        if (state.on && !ext_power_state) {
            int rc = ext_power_enable(ext_power);
            if (rc != 0) {
                LOG_ERR("Unable to enable EXT_POWER: %d", rc);
            }
        } else if (!state.on && ext_power_state) {
            int rc = ext_power_disable(ext_power);
            if (rc != 0) {
                LOG_ERR("Unable to disable EXT_POWER: %d", rc);
            }
        }
    }
#endif
    if (!state.on) {
        k_timer_stop(&underglow_tick);
    }
```
after previously fetching the external power state with `ext_power_state = ext_power_get(ext_power);` in `zmk_rgb_underglow_init`. Importing the rgb underglow state would enable/disable external power as needed.

RGB underglow/external power state might not be coupled in some configurations though so perhaps the central should just fetch the state and write to the run behavior characteristic

Percentnineteen:
I turned on USB debugging to see if I could add anything to my previous report. I couldn't glean anything interesting out of the reports when I tried to toggle external power but I did see something weird when trying to use &reset or &bootloader on the central side. Doing that, the logging receives an error:
<err> zmk: Source not connected

I added to the log message and saw the value of peripherals[payload_wrapper.source].state is 0 (PERIPHERAL_SLOT_STATE_OPEN). At first I thought this was maybe due to the fact that I was only testing central but even when I connected peripheral, I had no luck.

petejohanson:
> It looks like I independently developed the same functionality this week and only discovered this PR when I was going through open issues to reference in a PR. The one thing I see missing is state synchronization once a peripheral connects to the central half (for instances where the peripheral was disconnected while a behavior was invoked.) I have an additional write only characteristic for the RGB state and export/import state methods for the rgb underglow subsystem. Synchronizing the state as opposed to running a behavior would allow the animation steps to remain in sync. I had minimal safety checks in the write callback opting to only check if the length matches the expected size of the state struct. For more safety, I was thinking of including checks for the firmware version from the DIS but I don't have that implemented.
> 
> I don't have anything implemented for explicitly synchronizing the external power state since I was relying on `#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_EXT_POWER)` and updated `rgb_settings_set` to have
> 
> ```
> #if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_EXT_POWER)
>     if (ext_power) {
>         if (state.on && !ext_power_state) {
>             int rc = ext_power_enable(ext_power);
>             if (rc != 0) {
>                 LOG_ERR("Unable to enable EXT_POWER: %d", rc);
>             }
>         } else if (!state.on && ext_power_state) {
>             int rc = ext_power_disable(ext_power);
>             if (rc != 0) {
>                 LOG_ERR("Unable to disable EXT_POWER: %d", rc);
>             }
>         }
>     }
> #endif
>     if (!state.on) {
>         k_timer_stop(&underglow_tick);
>     }
> ```
> 
> after previously fetching the external power state with `ext_power_state = ext_power_get(ext_power);` in `zmk_rgb_underglow_init`. Importing the rgb underglow state would enable/disable external power as needed.
> 
> RGB underglow/external power state might not be coupled in some configurations though so perhaps the central should just fetch the state and write to the run behavior characteristic

Sorry you missed this!

Related, but not exactly the same, is state synchronization between halves. I would like to get this in as is, and the continue exploring additional work here while also getting some other work moving forward.

Looking forward to seeing what you came up w/!

petejohanson:
> 

I've just pushed a fix for invoking `&reset`, `&bootloader`, etc on the central side. Testing would be appreciated.

Regarding external power.... That seems to be working here on my Kyria, but that also has RGB. I will test on my Corne as well. Can you get logs for the peripheral side if you're still having issues w/ the latest on this branch?

app/src/split/bluetooth/service.c
```diff
@@ -24,12 +26,57 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+    if (end_addr > sizeof(struct zmk_split_run_behavior_payload)) {
+        return BT_GATT_ERR(BT_ATT_ERR_INVALID_OFFSET);
+    }
+
+    memcpy(payload + offset, buf, len);
+
+    // We run if:
+    // 1: We've gotten all the position/state/param data.
+    // 2: We have a null terminated string for the behavior device label.
+    if ((end_addr > sizeof(struct zmk_split_run_behavior_data)) &&
```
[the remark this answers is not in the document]
petejohanson: Can you offer some more details? Seems to be working just fine here w/ this as is.

petejohanson:
>     2. if I checkout this pr, &reset and &bootloader work correctly on the peripheral side but no longer do anything on the central side

Please retest, I pushed a fix for this.

> 
>     3. also, with this pr, &ext_power EP_TOG, EP_ON and EP_OFF only cause the desired effect on the central side - the peripheral is left unaffected
> 

@Percentnineteen  oI've tested this again on my Corne, no underglow enabled, and external power properly enables/disables just fine. Can you please pull and be sure you've got the latest code from this PR? There was a small fix for this I included while ago along w/ a rebase, so be sure the latest sha1/commit after you pull is:

```
commit d3b47391c9d210466a2e542d91c3cf558d89e4b3 (HEAD -> core/peripheral-behavior-invocation, origin/core/peripheral-behavior-invocation)
Author: Peter Johanson <xxx@xxx.com>
Date:   Sun Jan 30 03:38:10 2022 +0000

    fix(split): Add define for local source.
    
    * Add `ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL` and use
      it consinstently to fix bug w/ local `&reset`, `&bootloader`, etc.
```

Percentnineteen:
I tested the update and the &reset and &bootloader work now on the central side. However, I tested &ext_power EP_TOG/ON/OFF and they still don't seem to work. I logged the peripheral and got this:
```
[00:00:42.379,058] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state on
[00:00:42.379,119] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5
, pressed: true
[00:00:42.379,150] <dbg> zmk.split_listener:
[00:00:42.397,796] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:42.519,042] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state off
[00:00:42.519,134] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5
, pressed: false
[00:00:42.519,165] <dbg> zmk.split_listener:
[00:00:42.532,806] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:42.532,897] <dbg> zmk.split_svc_run_behavior: INVOKE THE BEHAVIOR: EXTPOW
ER with params 1 0
[00:00:44.166,046] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state on
[00:00:44.166,137] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5
, pressed: true
[00:00:44.166,137] <dbg> zmk.split_listener:
[00:00:44.182,800] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:44.262,054] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state off
[00:00:44.262,145] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5
, pressed: false
[00:00:44.262,176] <dbg> zmk.split_listener:
[00:00:44.272,796] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:44.272,888] <dbg> zmk.split_svc_run_behavior: INVOKE THE BEHAVIOR: EXTPOW
ER with params 0 0
```

I issued EXT_TOG twice. The central side turned off immediately after the first command executed while peripheral stayed on. I assume peripheral should shut off immediately also. Central does not turn on when turning the power back on but I believe that is due to a different issue with reinitializing the OLED (VCC goes high, the display just doesn't come back).

I believe I have the latest from this PR. The top commit when I do "git log" is:
```commit d3b47391c9d210466a2e542d91c3cf558d89e4b3 (HEAD -> locality-pr)
Author: Peter Johanson <peter@peterjohanson.com>
Date:   Sun Jan 30 03:38:10 2022 +0000

    fix(split): Add define for local source.

    * Add `ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL` and use
      it consinstently to fix bug w/ local `&reset`, `&bootloader`, etc.
```

petejohanson:
@Percentnineteen Hmm.... I suspect maybe only the release, not the press, is getting sent. Can you pull again ans flash the right side and get the logs one more time?

Thanks for all the testing!

Percentnineteen:
> 

No problem. I pulled and got this in the logs:
```[00:00:28.903,045] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state on
[00:00:28.903,137] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: true
[00:00:28.903,137] <dbg> zmk.split_listener:
[00:00:28.914,764] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:28.986,053] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state off
[00:00:28.986,145] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: false
[00:00:28.986,175] <dbg> zmk.split_listener:
[00:00:28.997,253] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:28.997,344] <dbg> zmk.split_svc_run_behavior: EXTPOWER with params 1 0: pressed? 0
[00:00:38.375,061] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state on
[00:00:38.375,122] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: true
[00:00:38.375,152] <dbg> zmk.split_listener:
[00:00:38.387,268] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:38.490,081] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state off
[00:00:38.490,142] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: false
[00:00:38.490,173] <dbg> zmk.split_listener:
[00:00:38.507,232] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
[00:00:38.507,263] <dbg> zmk.split_svc_run_behavior: EXTPOWER with params 0 0: pressed? 0
```

petejohanson:
> > 
> 
> No problem. I pulled and got this in the logs:
> ```[00:00:28.903,045] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state on
> [00:00:28.903,137] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: true
> [00:00:28.903,137] <dbg> zmk.split_listener:
> [00:00:28.914,764] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
> [00:00:28.986,053] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state off
> [00:00:28.986,145] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: false
> [00:00:28.986,175] <dbg> zmk.split_listener:
> [00:00:28.997,253] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
> [00:00:28.997,344] <dbg> zmk.split_svc_run_behavior: EXTPOWER with params 1 0: pressed? 0
> [00:00:38.375,061] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state on
> [00:00:38.375,122] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: true
> [00:00:38.375,152] <dbg> zmk.split_listener:
> [00:00:38.387,268] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
> [00:00:38.490,081] <dbg> zmk.kscan_matrix_read: Sending event at 0,0 state off
> [00:00:38.490,142] <dbg> zmk.zmk_kscan_process_msgq: Row: 0, col: 0, position: 5, pressed: false
> [00:00:38.490,173] <dbg> zmk.split_listener:
> [00:00:38.507,232] <dbg> zmk.split_svc_run_behavior: offset 0 len 19
> [00:00:38.507,263] <dbg> zmk.split_svc_run_behavior: EXTPOWER with params 0 0: pressed? 0
> ```
> 

Yeah, only seeing the releases, not the presses, which is the issue. 

Can you link to your keymap? Is this invoked from a combo, or hold-tap maybe?

petejohanson:
@Percentnineteen Pushed one more small fix. Can you try one more tie? Should only require flashing the peripheral side.

app/src/split/bluetooth/service.c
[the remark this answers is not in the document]
petejohanson: @ReFil Found an off-by-one error in the second part of this conditional, that may have been the root of the problem. Can you try again?

Percentnineteen:
Trying to link to my zmk-config repo but it's a fork of the miryoku repo and GitHub seemingly won't let me share it...

No combos - I've just got a layer-tap to activate a layer and have a key mapped to &ext_power EP_TOG.

After converting from miryoku's format to default dtsi format looks like this:
```    BASE_layer {
      label = "Base";
      bindings = <
      &kp Q             &kp W             &kp E             &kp R             &kp T             &kp Y             &kp U             &kp I             &kp O             &kp P
      &hm LGUI A        &hm LALT S        &hm LCTRL D       &hm LSHFT F       &kp G             &kp H             &hm LSHFT J       &hm LCTRL K       &hm LALT L        &hm LGUI SQT
      &lt BUTTON Z      &hm RALT X        &kp C             &kp V             &kp B             &kp N             &kp M             &kp COMMA         &hm RALT DOT      &lt BUTTON SLASH
                                          &lt MEDIA ESC     &lt NAV SPC       &lt MOUSE TAB     &lt SYM RET       &lt NUM BSPC      &lt FUN DEL
      >;
    };
    MEDIA_layer {
      label = "Media";
      bindings = < 
      &bootloader       &reset            U_NA              U_NA              U_NA          &ext_power EP_TOG   &ext_power EP_OFF   &ext_power EP_ON    &rgb_ug RGB_TOG   &rgb_ug RGB_BRI
      &kp LGUI          &kp LALT          &kp LCTRL         &kp LSHFT         U_NA          &kp C_PREV          &kp C_VOL_DN        &kp C_VOL_UP        &kp C_NEXT        &out OUT_TOG
      U_NA              &kp RALT          U_NA              U_NA              U_NA          &bt BT_SEL 0        &bt BT_SEL 1        &bt BT_SEL 2        &bt BT_SEL 3      &bt BT_CLR
                                          U_NA              U_NA              U_NA          &kp C_STOP          &kp C_PP            &kp C_MUTE
      >;
    };
```

Percentnineteen:
Crap! I didn't see the last fix before my last message. I pulled that last fix:
```
commit 90816173d06300a6dcd1d1d018389bef2f07836f (HEAD -> locality-pr, petejohanson/core/peripheral-behavior-invocation)
Author: Peter Johanson <peter@peterjohanson.com>
Date:   Sun Jan 30 05:32:12 2022 +0000

    fix(split): Fix an off-by-one error in split svc.

    * Properly check end of behavior device string for null terminator.

```

After pulling this, it works. Awesome.

app/src/split/bluetooth/service.c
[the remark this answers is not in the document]
ReFil: yep will do, thanks

app/include/dt-bindings/zmk/rgb.h
```diff
@@ -17,7 +17,8 @@
 #define RGB_SPD_CMD 10
 #define RGB_EFF_CMD 11
 #define RGB_EFR_CMD 12
-#define RGB_COLOR_HSB_CMD 13
+#define RGB_EFS_CMD 13
```
Nicell, line 20: Is this required to get each side synced?

app/src/keymap.c
```diff
@@ -179,20 +195,46 @@ int zmk_keymap_apply_position_state(int layer, uint32_t position, bool pressed,
...
+    enum behavior_locality locality = BEHAVIOR_LOCALITY_CENTRAL;
+    err = behavior_get_locality(behavior, &locality);
+    if (err) {
+        LOG_ERR("Failed to get behavior locality %d", err);
+        return err;
+    }
+
+    switch (locality) {
+    case BEHAVIOR_LOCALITY_CENTRAL:
+        return invoke_locally(&binding, event, pressed);
```
Nicell, line 207: Is there a reason this doesn't need to be wrapped in `#if ZMK_BLE_IS_CENTRAL`? Is it because if it's pressed on the peripheral, the position gets sent to the central and never gets sent back to be invoked?

app/src/rgb_underglow.c
```diff
@@ -332,18 +332,28 @@ int zmk_rgb_underglow_off() {
...
+    state.current_effect = effect;
     state.animation_step = 0;
 
     return zmk_rgb_underglow_save_state();
 }
 
+int zmk_rgb_underglow_cycle_effect(int direction) {
+    return zmk_rgb_underglow_select_effect(zmk_rgb_underglow_calc_effect(direction));
+}
+
```
Nicell, line 356: This answered my question on why the selection behavior was needed. Looks good to me!

Nicell (commented):
I mostly reviewed this so that I could familiarize myself with the changes going forward. All the changes make sense to me, and the idea of central, global, and source looks good. Nice touch that global sends to _all_ peripherals, which I presume will make dongle-ing easier in the future. I left a few comments, but I think I answered one myself. Great work!

app/src/keymap.c
petejohanson: That's because technically, for a split or non-split, you'd always want to run this locally, and keymaps aren't ever processed anywhere else, so it doesn't need to be conditional in either case.

app/src/rgb_underglow.c
petejohanson: Yup! This helps be sure we can invoke the behavior on the peripheral side w/ absolute (not relative "next effect") parameters to get it to have the exact effect the central side has set as well.

app/include/dt-bindings/zmk/rgb.h
petejohanson: You saw why elsewhere, marking resolved! Thanks for the review!

app/src/split/bluetooth/service.c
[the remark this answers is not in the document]
petejohanson: For my own notes: verified on Discord the off-by-one-error fix did resolve this for @ReFil. Marking resolved.

app/src/keymap.c
Nicell: Just clicked for me that the `#if ZMK_BLE_IS_CENTRAL` is being used to identify if the device is a split keyboard, not if it's the central half. Makes a lot more sense. Would `#if IS_ENABLED(CONFIG_ZMK_SPLIT)` be a better check since only central will ever be running it anyways?

app/src/split/bluetooth/central.c
petejohanson: Fixed! (nearly a year later, oof).

app/src/split/bluetooth/service.c
petejohanson: Indeed. Refactored.

Issue #460, opened 2020-12-02
settings_reset shield only resets Bluetooth settings


Nicell:
Currently the settings_reset shield only resets Bluetooth settings, we should probably expand this to clear out the entire NVS storage, as many more parts are starting to incorporate settings.

jmding8:
Would it be possible to trigger this reset behavior from a keycode, instead of having to flash it?

My MBP occasionally disconnects from my keyboard, where the only way to restore the connection is to go through the process of flashing the settings_reset shield, which is quite a hassle

Lenbok:
I get this too on linux where sometimes the only way to get a connection re-established is to flash the settings reset. It would be immensely helpful if there were a way to bind this to a key instead of having to reflash twice.

joelspadin:
Done in 1dfcfc7d3f074e538900a6df684cc99e92c60a21

PR #548, opened 2020-12-29
(docs) Add note about settings reset to troubleshooting documentation

mcrosson:
The troubleshooting docs were missing information on the settings reset images that can be flashed to fully reset a boards settings. This PR adds information on where the pre-built images live as well as how to build one locally.


Nicell:
@mcrosson These settings_reset images don't actually clear all settings as their name suggests. See #460. This image is only used to clear BLE bonds, which is outlined in the part where halves won't pair with each other a bit above this in the same file.

mcrosson:
> 
> 
> @mcrosson These settings_reset images don't actually clear all settings as their name suggests. See #460. This image is only used to clear BLE bonds, which is outlined in the part where halves won't pair with each other a bit above this in the same file.

Is there an image that actually zaps the whole of the settings somewhere?

Nicell:
@mcrosson No there isn't. #460 is tracking that.

caksoylar:
This exists now, closing.

PR #550, opened 2020-12-29
(docs) Add additional information to FAQ about split keyboards

mcrosson:
Added additional information on how to pair splits as well as TRRS wires providing power when using batteries to the FAQ


tominabox1:
The zmk terminology is "central" and "peripheral" when referring to the "primary" and "non-primary" halves. I think you want to use those terms here.  It might also be worthwhile to note that split primaries will show up as `$keyboard Left` by default.

mcrosson:
> 
> 
> The zmk terminology is "central" and "peripheral" when referring to the "primary" and "non-primary" halves. I think you want to use those terms here. It might also be worthwhile to note that split primaries will show up as `$keyboard Left` by default.

Nice catch. Updated the PR accordingly.

tominabox1:
Cool! lgtm!

docs/docs/faq.md
```diff
@@ -80,6 +80,20 @@ Please note, many keyboards only have a single PCB which includes the “brains
...
+
+### Powering Split Keyboards
+
+Keyboard halves should be powered _independently_ of each other.
+
+:::warning
+
+Do _NOT_ use TRRS cables with keyboards that provide power via TRRS when the peripheral keyboard half is running on battery power. The power provided by TRRS cables can cause batteries on the peripheral keyboard half to fail and in the worst case catch fire.
+
+:::
```
Nicell, line 95: I'm not quite sure where this has come from? As most devices use a similar schematic to the nice!nano, I'll use this to describe what would happen. The VCC pins connected through the TRRS cable are both outputs of the regulated 3.3V LDO. This means there's no direct connection between the batteries or the charger. Rather it's now a connection between two LDOs outputting 3.3V (and a MOSFET to cut power, but that's besides the point).

It's true that when plugging in you may short ground and power for a split second, which is obviously not good. If you have no battery plugged in on the right half, there's absolutely nothing to worry about, and if you do, the issue is quite minimal IMO. I wouldn't *recommend* people do this, but it also shouldn't cause anything to fail or explode.

docs/docs/faq.md
```diff
@@ -80,6 +80,20 @@ Please note, many keyboards only have a single PCB which includes the “brains
 
 Currently, ZMK only supports wireless split, but wired split is possible and we welcome contributions!
 
+### Pairing Split Keyboards
+
+Only the central half of a keyboard should be paired to a computer or device. The peripheral half will auto-connect to the central half via BLE and should _not_ be paired directly to a computer or device.
```
Nicell, line 85: This feels like something that should be outlined in the getting started with ZMK rather than in the FAQ. I doubt many would come here to learn how to pair their device for the first time.

docs/docs/faq.md
mcrosson: This came from a conversation with @petejohanson ; If it's a non nice!nano there is no guarantee that there won't be an issue with batteries on the peripheral half, no?

Do you have a way of wording the 'not recommended' in a way that makes it clear that folks shouldn't use the TRRS cable(s) when a battery is setup on the peripheral half of the board?

docs/docs/faq.md
petejohanson: Agreed.

docs/docs/faq.md
petejohanson: I misunderstood the scope.of.thw concern here, definitely. I'll defer to @Nicell on this.

petejohanson: @mcrosson @Nicell So.... not needed at this point?

mcrosson: Is it still unsafe to have a TRRS cable plugged in for splits?

petejohanson: @Nicell Is the concern here still negligible w/ the nano v2?

Nicell: @petejohanson I'd still say it's generally negligible. It's undefined behavior though so it's not recommended. It may be worth stating it's not a good idea when the peripheral side has a battery, but I think this warning is a bit extreme.

PR #551, opened 2020-12-29
Show layer label in display widget.

petejohanson:


Nicell:
Tested on the left half of my lily58, and it's working well.

Might want some documentation on labels for keymaps. Also we should add label props to existing keymaps.

petejohanson:
> Tested on the left half of my lily58, and it's working well.
> 
> Might want some documentation on labels for keymaps. Also we should add label props to existing keymaps. 

Yeah, both were going to be follow up PRs. Any concerns @Nicell ?

Nicell:
> Yeah, both were going to be follow up PRs. Any concerns @Nicell ?

As long as they're planned for future PRs, I have no concerns. Code looks fine as far as I can tell 👍

petejohanson:
Created #578 and #579 to document those two future items. Merging this one.

PR #553, opened 2020-12-29
Use bluetooth symbol instead of wifi symbol

allymparker:
Low hanging fruit!  I don't know if there was a reason to use the wifi symbol instead of Bluetooth?

### DIFF
--- a/app/src/display/widgets/output_status.c
+++ b/app/src/display/widgets/output_status.c
@@ -49,12 +49,12 @@ void set_status_symbol(lv_obj_t *label) {
     case ZMK_ENDPOINT_BLE:
         if (active_profie_bonded) {
             if (active_profile_connected) {
-                sprintf(text, LV_SYMBOL_WIFI "%i " LV_SYMBOL_OK, active_profile_index);
+                sprintf(text, LV_SYMBOL_BLUETOOTH "%i " LV_SYMBOL_OK, active_profile_index);
             } else {
-                sprintf(text, LV_SYMBOL_WIFI "%i " LV_SYMBOL_CLOSE, active_profile_index);
+                sprintf(text, LV_SYMBOL_BLUETOOTH "%i " LV_SYMBOL_CLOSE, active_profile_index);
             }
         } else {
-            sprintf(text, LV_SYMBOL_WIFI "%i " LV_SYMBOL_SETTINGS, active_profile_index);
+            sprintf(text, LV_SYMBOL_BLUETOOTH "%i " LV_SYMBOL_SETTINGS, active_profile_index);
         }
         break;
     }


petejohanson:
Thanks for the quick PR!

@allymparker There are a few reasons, sadly...

1. We're not certified by bluetooth SIG, and they can be selective in enforcement.
2. It looks *really* bad at lower resolutions, especially on e-ink.

I'm open to other ideas though...

allymparker:
Re point 1 - I've submitted a support request to the Bluetooth sig after reading through their licencing docs and not being able to come to a conclusion. I DIDN'T make any explicit references to the ZMK project itself.

As for 2. What res are the eink displays?  I've been wondering about creating a [new font for LVGL ](https://docs.lvgl.io/v7/en/html/overview/font.html?highlight=symbol#add-new-symbols) to pull in a couple more symbols as I couldn't find any built in ones for the WPM counter display.  I'm wondering if the [bluetooth-b](https://fontawesome.com/icons/bluetooth-b?style=brands) symbol might work better at lower resolutions?

Lenbok:
Yeah, I tried this to see what it was like on my corne, and it's basically unrecognizable: 

image

Can you make a version of this PR that uses the bluetooth-b symbol instead?

allymparker:
Sadly I got a response from the Bluetooth SIG and the answer is no.

> Thank you for your mail to the Bluetooth SIG.  The Bluetooth word mark and logos are registered trademarks and as such their use can only be done under license and in relation to qualified goods and services.  Even though you are a using a qualified module, such as the nRF52840, you would not be able to use the Bluetooth word mark or logos in relation to the end product as the Bluetooth product would not be considered qualified in its current state.
> 
> There are some cases where you could use the Bluetooth word mark from a fair use perspective (IE - you could state that the module is Bluetooth enabled, etc.) - but you would not be able to use the Bluetooth figure mark in the software display - unless the entire end product was properly qualifeid.
> 
> If you have any further questions, please feel free to reach out to brand.manager@bluetooth.com.

innovaker:
That's a shame.

> There are some cases where you could use the Bluetooth word mark from a fair use perspective (IE - you could state that the module is Bluetooth enabled, etc.) - but you would not be able to use the Bluetooth figure mark in the software display - unless the entire end product was properly qualifeid.

This however is useful to know as we've tended to avoid doing so.  Thanks! :)

petejohanson:
> Sadly I got a response from the Bluetooth SIG and the answer is no.
> 
> > Thank you for your mail to the Bluetooth SIG.  The Bluetooth word mark and logos are registered trademarks and as such their use can only be done under license and in relation to qualified goods and services.  Even though you are a using a qualified module, such as the nRF52840, you would not be able to use the Bluetooth word mark or logos in relation to the end product as the Bluetooth product would not be considered qualified in its current state.
> > There are some cases where you could use the Bluetooth word mark from a fair use perspective (IE - you could state that the module is Bluetooth enabled, etc.) - but you would not be able to use the Bluetooth figure mark in the software display - unless the entire end product was properly qualifeid.
> > If you have any further questions, please feel free to reach out to [brand.manager@bluetooth.com](mailto:brand.manager@bluetooth.com).

Thanks for trying! Never know for sure until you do.

Lenbok:
How about instead of using any of the bluetooth logos, we just use the vanilla un-trademarked runic B (unicode 16D2) instead?

tominabox1:
I agree, I'd like a symbol that is less wifi and more bluetooth but if we can't use the actual cool B, some creative replacement would be better.... Maybe just "Bt"?

Lenbok:
The Bluetooth figure mark is a runic B / runic h composite inside an oval, so there would definitely be no impediment to using just a runic B on it's own, because that is not trademarked. I'm less sure whether the B / h composite (bluetooth-b linked above) without the oval would be OK. 

Personally I think that since this is about a status indicator rather than use of symbols in branding / marketing (which is what the trademarks are for), the question shouldn't really even be applicable.

Nicell:
It should be noted that the default items available in LVGL are pretty limited. You can see everything here: https://docs.lvgl.io/v7/en/html/overview/font.html

Looks like the runic B would not be supported. Nor is there really an alternative wireless icon outside of the WiFi one.

Lenbok:
It wouldn't be a bad thing if ZMK includes an example of how to add a custom symbol :-)

PR #555, opened 2020-12-29
is31fl3733 Driver

Retic1337:


app/drivers/led/is31fl3733/is31fl3733.c
```diff
@@ -0,0 +1,202 @@
+/*
+ * Copyright (c) 2020 Cameron Banna.
```
petejohanson, line 2: We generally use an "AUTHOR files" approach for copyright, and would do something like:

```suggestion
 * Copyright (c) 2020 The ZMK Contributors
```

And then you can open a separate PR to add yourself to the toplevel `AUTHORS` file

Also, this would also need a license header. Is this at all based on an existing Zephyr driver you modified, or from scratch?

app/drivers/led/is31fl3733/Kconfig
```diff
@@ -0,0 +1,20 @@
...
+	help
+	  Enable LED driver for IS31FL3733.
+
+		The IS31FL3733 is a general purpose 12 by 16 LED
+		matrix driver with 1/12 cycle rate. The device can be
+		programmed via an I2C compatible interface. Each
+		LED can be dimmed individually with 8-bit PWM data
+		which allowing 256 steps of linear dimming
+
+config LED_INIT_PRIORITY
```
petejohanson, line 15: This seems odd to be here. This should ideally leverage an existing init priority used by other LED drivers, or have a name that indicates it is specific to this driver.

app/drivers/zephyr/dts/bindings/led/is31fl3733/issi,is31fl3733.yaml
```diff
@@ -0,0 +1,9 @@
+description: |
+  Driver for the is31fl3733 LED driver
+
+compatible: "issi,is31fl3733"
+
+properties:
+  label:
+    required: true
+    type: string
```
petejohanson, line 9: This should be replaced with:

```suggestion
include: i2c-device.yaml
```

Which adds the `label` and `reg` property as expected, and sets `on-bus: i2c` as well.

petejohanson (changes requested):
Thanks for the PR! A few comments/issues.

app/drivers/led/is31fl3733/is31fl3733.c
Retic1337: Okay thanks. This is from scratch, only referenced existing drivers for how to use the api, and the datasheet for offsets.

app/drivers/led/is31fl3733/Kconfig
Retic1337: ill change this to match other led drivers.

app/drivers/zephyr/dts/bindings/led/is31fl3733/issi,is31fl3733.yaml
Retic1337: prefect!

petejohanson:
Is there any reason not to use the led_strip API for this? Does the i2c API allow setting lots of values with a single reg update?

yiancar:
Hello :)
Just a few suggestions here:
1) calculating the col and row at every single led update to figure out which reg to update seems a bit superfluous. This might become a problem as the extra multiply/divides will start piling up.

2) usually in effects a whole frame is calculated and then updated. For this reason I suggest giving the capability of updating a pwm buffer where its index are the leds, and then on a separate function sending that whole buffer to the led driver.

Both points are to optimize compute as well as the use of the i2c bus.

PR #556, opened 2020-12-30
Create a WPM counter and display widget

allymparker:
A pretty basic algorithm.

For some reason, this doesn't work on the right half of a split (always shows 0) and I'm having trouble getting the right alignment to work when WPM hits 2 digits (it overlaps with the edge of the screen) so I've centered it for now


app/src/wpm.c
```diff
@@ -0,0 +1,83 @@
...
+
+static uint8_t wpm_state;
+static uint8_t wpm_update_counter;
+static uint32_t key_pressed_count;
+
+int zmk_wpm_get_state() { return wpm_state; }
+
+int wpm_event_listener(const struct zmk_event_header *eh) {
+
+    LOG_DBG("Got an event");
```
okke-formsma, line 38: you may wanna clean up the debugs a bit, and maybe make them a bit less 'chatty', see other debug statements throughout the project.

okke-formsma (commented):
in general this looks pretty good. A few remarks. 

* I think it would be great if you added a few tests which check if the "WPM changed to xxx" show the correct WPM.
* There are a few debug statements that can be removed or made more concise

Also an idea, but you can also leave that for 'version 2':
* instead of resetting the WPM after `WPM_RESET_INTERVAL_SECONDS`, it'd be cool to keep a running average over an interval by keeping a array with data. (e.g., keep track of the counts for the last 5 seconds in an array of counters, calculate the average over those 5 seconds to display and clear one of the counters every second).

app/src/wpm.c
```diff
@@ -0,0 +1,83 @@
...
+    }
+    return 0;
+}
+
+void wpm_work_handler(struct k_work *work) {
+    wpm_update_counter++;
+    wpm_state = (key_pressed_count / CHARS_PER_WORD) /
+                (wpm_update_counter * WPM_UPDATE_INTERVAL_SECONDS / 60.0);
+
+    if (wpm_update_counter == WPM_RESET_INTERVAL_SECONDS) {
```
okke-formsma, line 56: Maybe change this into >= to make sure there's no chance of missing the interval

app/src/wpm.c
allymparker: Good call - I was planning to drop the log message commit and put in some better logs 👍

app/src/wpm.c
allymparker: Will do

allymparker:
I've addresssed the comments and added a couple of tests.  I think this is ready to go for v1 👍 

> Also an idea, but you can also leave that for 'version 2':
> 
>     instead of resetting the WPM after WPM_RESET_INTERVAL_SECONDS, it'd be cool to keep a running average over an interval by keeping a array with data. (e.g., keep track of the counts for the last 5 seconds in an array of counters, calculate the average over those 5 seconds to display and clear one of the counters every second).

I'll start taking a look at this

okke-formsma:
lgtm

okke-formsma (approved):
lgtm

app/src/display/widgets/Kconfig
```diff
@@ -22,5 +22,12 @@ config ZMK_WIDGET_OUTPUT_STATUS
     default y if BT
     select LVGL_USE_LABEL
     select LVGL_FONT_MONTSERRAT_16
+    
+config ZMK_WIDGET_WPM_STATUS
+    bool "Widget for displaying typed words per minute"
+    default y
```
mcrosson, line 28: Should this be a default widget for OLEDs? Seems like this would be more of an optional feature.

mcrosson (commented):
Is there a reason the two files below are included in this PR? The changes to them are outside the scope of the PR and change unrelated ZMK behavior.

* `.vscode/c_cpp_properties.json`
* `app/boards/shields/sofle/sofle.conf`

app/src/display/widgets/Kconfig
petejohanson: Agreed this should be opt-in.

allymparker:
> Is there a reason the two files below are included in this PR? The changes to them are outside the scope of the PR and change unrelated ZMK behavior.
> 
>     .vscode/c_cpp_properties.json
>     app/boards/shields/sofle/sofle.conf

Inadvertent mistake! Apologies!

I'll make the changes now

app/src/display/widgets/Kconfig
allymparker: Done

okke-formsma:
@allymparker I think it's time to squash all commits to a single one, with a proper commit message (see https://www.conventionalcommits.org/en/v1.0.0/ and other commits in this repo).

allymparker:
Done

app/src/display/widgets/Kconfig
petejohanson: To make this not default, this `default y` line should be removed, which then will use the implicit `default n` from Kconfig.

app/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ target_sources(app PRIVATE src/kscan.c)
 target_sources(app PRIVATE src/matrix_transform.c)
 target_sources(app PRIVATE src/hid.c)
 target_sources(app PRIVATE src/sensors.c)
+target_sources(app PRIVATE src/wpm.c)
```
petejohanson, line 31: For users that don't care about WPM, we should be able to not have this built at all, including an extra listener, etc.

Can we move this *also* behind a new Kconfig setting?

petejohanson (changes requested):
Thanks for this! On a whole, this looks great! Just a couple minor build/Kconfig items.

app/CMakeLists.txt
allymparker: I've put it behind a config session but I'm not sure how I can get the tests to add this for compilation.  Could you point me in the right direction?

petejohanson: You'd want to enable it in a file like `app/tests/wpm/1-single_keypress/native_posix.conf` and add the `CONFIG_ZMK_WPM=y` there for each test that needs to enable it.

On that note, I'm fine w/ using just one Kconfig flag for both the widget, and the supporting code for now, but longer term we may want two flags one for the core WPM code, another for the widget, so folks can use the core WPM stuff, but leverage it elsehwere (GATT notification? Keycode that sends some hid keycodes to send the current WPM to the host?) for folks that don't have displays.

allymparker: Right, finally got the tests working! I ended up adding another kconfig option to enable just the wpm calc too. If you're happy with the new commits I'll squash them ready for merge

app/src/display/widgets/Kconfig
```diff
@@ -22,5 +22,11 @@ config ZMK_WIDGET_OUTPUT_STATUS
     default y if BT
     select LVGL_USE_LABEL
     select LVGL_FONT_MONTSERRAT_16
+    
+config ZMK_WIDGET_WPM_STATUS
+    bool "Widget for displaying typed words per minute"
+    depends on !ZMK_SPLIT || ZMK_SPLIT_BLE_ROLE_CENTRAL
+    select LVGL_USE_LABEL
+    select LVGL_FONT_MONTSERRAT_16
```
mcrosson, line 30: Shouldn't this also set `ZMK_WPM=y` to avoid build errors? I imagine if a user turns on the WPM widget they are going to want `ZMK_WPM` set as well

mcrosson:
Can you also flip the alignment to `LV_ALIGN_IN_BOTTOM_RIGHT` so it doesn't collide with the layer status widget?

app/src/display/widgets/Kconfig
allymparker: Good Catch, I'll get that added

allymparker:
Thanks @mcrosson I was struggling with stopping the number overflowing on the right! That's why I shifted to centre.  I'll give that a go too

allymparker:
Made the changes.  If you're happy let me know and I can squash ready for mege

app/src/display/widgets/Kconfig
```diff
@@ -25,6 +25,7 @@ config ZMK_WIDGET_OUTPUT_STATUS
     
 config ZMK_WIDGET_WPM_STATUS
     bool "Widget for displaying typed words per minute"
+    depends on ZMK_WPM
```
mcrosson, line 28: I was actually suggesting if `CONFIG_ZMK_WIDGET_WPM_STATUS` it set it should auto-select `CONFIG_ZMK_WPM` as well instead of depending on `CONFIG_ZMK_WPM`

mcrosson (changes requested):
A couple notes as well as a code comment

- Overall look and position of widget look good
- I see a LOT of compile warnings when `CONFIG_ZMK_USB_LOGGING=n` is set

app/src/display/widgets/Kconfig
petejohanson: Right, so instead of `depends on` you would use `select` here.

allymparker: Thanks folks - everyday's a learning day!

I think I've solved the logging warnings and changed it to select.

mcrosson: Log warnings and select changes are good to go

mcrosson (approved):
Just reviewed the latest code and it looks good to go. @petejohanson / @okke-formsma  this is ready for merge IMO

allymparker:
Hmm so now I've made the change to `select` instead of `depends on` I need to switch both new config switches on for the tests to pass. Is that expected?

mcrosson:
> 
> 
> Hmm so now I've made the change to `select` instead of `depends on` I need to switch both new config switches on for the tests to pass. Is that expected?

You should only need to specify `CONFIG_ZMK_WIDGET_WPM_STATUS=y` in the test case config now.

allymparker:
Ah it's because the widget config flag depends on the DISPLAY flag. If I switch the display flag on though the build is totally broken(and much longer). Probably cheaper to just switch on the 2 WPM flags?

mcrosson:
> 
> 
> Ah it's because the widget config flag depends on the DISPLAY flag. If I switch the display flag on though the build is totally broken(and much longer). Probably cheaper to just switch on the 2 WPM flags?

If all you're doing is testing the core WPM feature I'd only turn on `CONFIG_ZMK_WPM=y` for the tests since the OLED isn't covered by test code at this time.

allymparker:
Right! Got it all working using just the one flag.  I've also rebased to pick up the breaking changes in the event API so should be good to merge!

mcrosson (approved):

mcrosson:
@allymparker I'm getting some build errors with the latest zmk sources ; can you update to the latest main and re-run a build. It looks like some additional changes may have been added to main since you last updated the sources

app/src/display/widgets/wpm_status.c
```diff
@@ -0,0 +1,67 @@
...
+
+int wpm_status_listener(const struct zmk_event_header *eh) {
+    struct wpm_state_changed *ev = cast_wpm_state_changed(eh);
+    struct zmk_widget_wpm_status *widget;
+    SYS_SLIST_FOR_EACH_CONTAINER(&widgets, widget, node) { set_wpm_symbol(widget->obj, ev->state); }
+    return 0;
+}
+
+ZMK_LISTENER(widget_wpm_status, wpm_status_listener)
+ZMK_SUBSCRIPTION(widget_wpm_status, wpm_state_changed);
```
mcrosson, line 67: To fixup / reconcile this block with the current main sources you'll need to change it to the following:

```
int wpm_status_listener(const zmk_event_t *eh) {
    struct zmk_wpm_state_changed *ev = as_zmk_wpm_state_changed(eh);
    struct zmk_widget_wpm_status *widget;
    SYS_SLIST_FOR_EACH_CONTAINER(&widgets, widget, node) { set_wpm_symbol(widget->obj, ev->state); }
    return 0;
}

ZMK_LISTENER(widget_wpm_status, wpm_status_listener)
ZMK_SUBSCRIPTION(widget_wpm_status, zmk_wpm_state_changed);
```

mcrosson (commented):
Added some comments on how to fix the build with the latest main branch sources.

app/src/display/widgets/wpm_status.c
allymparker: Thanks.  Didn't see that when I was building locally but did when I built my config.  I'll look into that tomoz.  I think that's it :)

mcrosson (approved):

petejohanson (approved):
Thanks!

PR #559, opened 2020-12-31
feat(core): Add USB logging Kconfig option

Nicell:
This adds a singular Kconfig option that will enable all the USB logging related config options. It also bumps the buffer size a bit to reduce the dropped message count.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -293,6 +293,50 @@ config ZMK_KSCAN_COMPOSITE_DRIVER
 #KSCAN Settings
 endmenu
 
+menu "USB Logging"
+
+config ZMK_USB_LOGGING
+	bool "Enable USB CDC ACM logging to help debug"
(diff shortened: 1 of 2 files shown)


app/Kconfig
```diff
@@ -293,6 +293,50 @@ config ZMK_KSCAN_COMPOSITE_DRIVER
 #KSCAN Settings
 endmenu
 
+menu "USB Logging"
+config ZMK_USB_LOGGING
+	bool "Enable USB CDC ACM logging to help debug"
+	default n
```
petejohanson, line 299: Zephyr docs suggest just omitting this line, since that's the default default.

```suggestion
```

petejohanson (changes requested):
Minor nitpick, otherwise this looks good.

petejohanson (approved):

PR #568, opened 2021-01-02
chore: move prettier to root

innovaker:
Please check this in detail as there may be subtle nuances I've missed.


innovaker:
This PR doesn't (yet?) include changes to:
- setup documentation
- `.devcontainer`

I welcome PRs to this branch.

innovaker:
Rather than maintain the prettier script commands in both the root's and docs's `package.json`, I favoured moving it solely into the root.  A consequence of this is that `npm run prettier:write` must be run from root whereas `npm run lint` is run from docs.  Thoughts?  We may also wish to consider moving `lint` to the root.

.prettierignore
```diff
@@ -0,0 +1,6 @@
+build
+node_modules
+docs/.docusaurus
+/modules
+/tools
+/zephyr
```
Nicell, line 6:
```suggestion
/zephyr
app/build

```

Nicell (changes requested):
I'm not a huge fan of having the two commands being split in which directory they should be run. I'm wondering if we should add a lint command to root that runs within docs, so you don't have to move around to do your checks.

joelspadin:
> I'm not a huge fan of having the two commands being split in which directory they should be run. I'm wondering if we should add a lint command to root that runs within docs, so you don't have to move around to do your checks.

Maybe there's something clever we could do with [NPM workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to more gracefully handle one NPM package nested inside another? Then you could just have Prettier installed in both places (hopefully done automatically by running `npm install` from the root), configure the root Prettier to ignore docs, and set up the root Prettier script to run Prettier in the root followed by running it in docs.

innovaker:
> Maybe there's something clever we could do with NPM workspaces 

Worth trying for sure 👍  Good spot.

innovaker:
> Maybe there's something clever we could do with [NPM workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to more gracefully handle one NPM package nested inside another?

I've tried this with by adding:
```
  "workspaces": [
    "docs"
  ],
```
to `/package.json`.

Running `npm ci` in `/` installs prettier, but it also installs the docs.  I'm not sure who it was, but I believe someone expressed an interest in avoiding that behavior?  I'm personally fine with it though.

Also, inheritance seems to be limited (???) so we'd have to:
- re-add `prettier:check` and `prettier:format` to `/docs/package.json`'s scripts and keep those lines synced with root
- copy `/.prettierrc.js` into `/docs/prettierrc.js` and keep it synced with root
- re-add `/docs/.prettierignore`

We'd be abusing the purpose of workspaces slightly, but I think the pros probably outweigh the cons.

Thoughts?

PR #552, opened 2020-12-29
Add Events page to development section. 

allymparker:
Add an events page to the developer docs section.

Note:  I'm a beginner to C++ so if I've named anything wrong e.g. module/method/macro, please let me know!

PR #571, opened 2021-01-02
 Add Events page to development section

allymparker:
Add an events page to the developer docs section.

Note: I'm a beginner to C++ so if I've named anything wrong e.g. module/method/macro, please let me know!
Closes: #552 which was on my main branch

### DIFF
--- a/docs/docs/development/events.md
+++ b/docs/docs/development/events.md
@@ -1,7 +1,9 @@
 ---
 title: Events
 ---
+
 ## Overview
+
 ZMK uses events to decouple behaviours and to allow new functionality to be added to existing events.
 
 ## Subscribing to events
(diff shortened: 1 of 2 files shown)


allymparker:
I've done my best with this - I can't seem to find any usages of modifier state changed so wasn't able to document it.  I also wasn't sure of the difference between `sensor` and `sensor_number` in the sensor_changed event.

okke-formsma:
maybe it's helpful to show how the 'common' flow of events works for normal keypresses

* position state changed
  * (behavior is called, this will probably also be a behavior-state-changed event some time in the future)
    * keycode state changed *or*
    * layer state changed *or*
    * ble_active_profile_changed

PR #573, opened 2021-01-03
feature(events): add behavior-state-changed event

okke-formsma:
This commit adds a behavior-state-changed event which is called just before a behavior is activated. The last listener (in `behavior.c`) actually calls the behavior.

This set of commits leads up to a cleanup of the entire behavior handling, making complex behaviors like hold-tap and sticky keys a little simpler. The full set of planned commits can be found at https://github.com/zmkfirmware/zmk/pull/532.

TLDR;

* position_state_changed
    * keycode_state_changed
    * layer_state_changed

* position_state_changed
  * behavior_state_changed
    * keycode_state_changed
    * layer_state_changed


okke-formsma:
Why is this change needed? It's needed so we can create behaviors based on behavior events, such as combos that trigger on <&kp A, &kp B> instead of positions <1, 2>.

okke-formsma:
Rebased.

okke-formsma:
@petejohanson could you have a look at this?

app/include/zmk/events/behavior_state_changed.h
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+#pragma once
+
+#include <zephyr.h>
+#include <zmk/event_manager.h>
+#include <zmk/behavior.h>
+
+struct zmk_behavior_state_changed {
```
petejohanson, line 13: This struct and the existing binding event struct are duplicative and confusing after this addition. It seems like we should be unifying this as part of this work.

app/src/behavior.c
```diff
@@ -0,0 +1,59 @@
...
+
+#include <zmk/event_manager.h>
+#include <zmk/events/position_state_changed.h>
+#include <zmk/events/behavior_state_changed.h>
+#include <zmk/events/sensor_event.h>
+
+int zmk_behavior_state_changed_listener(const zmk_event_t *zmk_ev) {
+    struct zmk_behavior_state_changed *ev = as_zmk_behavior_state_changed(zmk_ev);
+    if (ev == NULL) {
+        return 0;
```
petejohanson, line 25: If we're getting any other event type, something is really amiss, and I feel like we should be returning an error (we might not be doing this elsewhere, but we should), I would suggest `-ENOTSUP` or somesuch.

app/src/behavior.c
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    int err = behavior_keymap_binding_convert_central_state_dependent_params(&ev->binding, event);
+    if (err) {
+        LOG_ERR("Failed to convert relative to absolute behavior binding (err %d)", err);
+        return err;
+    }
+
+    int ret;
+    if (ev->pressed) {
```
petejohanson, line 40: Perhaps the behavior driver API needs to evolve too.... I'm on the fence about explicit pressed/released functions, versus generic "state changed" function, which gives us more flexibility down the road.

petejohanson (commented):
A few thoughts, thanks for the ping.

app/src/behavior.c
okke-formsma: In https://github.com/zmkfirmware/zmk/issues/794 I argue that we should have explicit pressed/released event handler functions for all events for which the press/release duality exists, as most users of events dispatch to a different handler function based on the `state` now anyway.

app/include/zmk/events/behavior_state_changed.h
okke-formsma: I feel `zmk_behavior_binding` is a valid type on it's own. It's a nice succinct description of the things that exist in a keymap:

```
struct zmk_behavior_binding {
    char *behavior_dev;
    uint32_t param1;
    uint32_t param2;
};
```

okke-formsma: On second thought, were you talking about this?

```
struct zmk_behavior_binding_event {
    int layer;
    uint32_t position;
    int64_t timestamp;
};
```

Yeah, that one and `zmk_behavior_state_changed` should definitely be unified.

petejohanson: Yes exactly.

app/src/behavior.c
petejohanson: Bump.

petejohanson (commented):
@okke-formsma Any idea on if/when you might be able to get back to this

This and #547 will certainly conflict, so considering ordering/timing.

PR #577, opened 2021-01-03
refactor(docs): move contents of CONTRIBUTING.md into docs

innovaker:
Improves discoverability.


innovaker:
I prepared this in November 2019 but it got lost in the wash.

innovaker:
We'd probably benefit from an entirely new `Contributing` section to make it really prominent and to provide space for additional pages.  Thoughts?  Anyone want to take that on?

mcrosson:
> 
> 
> We'd probably benefit from an entirely new `Contributing` section to make it really prominent and to provide space for additional pages. Thoughts? Anyone want to take that on?

What do you envision going into the contributing section beyond what's already in this PR? Seems like it'd be smart to keep it as is from this PR and enhance/add as things pop up.

Nicell:
> We'd probably benefit from an entirely new `Contributing` section to make it really prominent and to provide space for additional pages. Thoughts? Anyone want to take that on?

I would agree with this. Having separate docs that are more direct on specific parts of the codebase would be good. For now this seems like a good start.

PR #510, opened 2020-12-13
Refactor central/peripheral config to allow user overrides of role

petejohanson:
* Move central/peripheral setting to Kconfig.defconfig
  files to allow left/right .conf files to override
  central/peripheral role for splits.

PR #580, opened 2021-01-04
docs(shield): Add docs based on #510 changes

Nicell:
This updates the docs for split boards to show how split configurations should be done based on the changes in #510.


petejohanson (approved):
Good catch! Thanks.

PR #581, opened 2021-01-04
Initial simple CODEOWNERS file

petejohanson:
I've created two new GH teams for this, instead of listing individuals in `CODEOWNERS`. Once this is in, we can tweak the @zmkfirmware/boards-shields-team  and @zmkfirmware/docs-team teams as needed, etc.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,5 @@
+* @zmkfirmware/core
+
+/app/boards @zmkfirmware/boards-shields
+
+/docs @zmkfirmware/docs


Nicell:
Wondering if another team called @zmkfirmware/admin or @zmkfirmware/core should be added in place of the direct @petejohanson references. Thoughts?

Also is it required to add @petejohanson to the docs/boards locations if full global ownership is already added?

petejohanson:
> Wondering if another team called @zmkfirmware/admin or @zmkfirmware/core should be added in place of the direct @petejohanson references. Thoughts?
> 
> Also is it required to add @petejohanson to the docs/boards locations if full global ownership is already added?

Core team works for me.


I believe it is, since that is a catch all used.if nothing else matches.

innovaker:
or `@zmkfirmware/maintainers`, just throwing it on the pile of suggestions.

Nicell:
> I believe it is, since that is a catch all used.if nothing else matches.

I see that now, makes sense. Last nitpick, @org/team-name is the format, so I think it's a bit redundant to have "team" in the team name.

innovaker:
We may wish to align that with GitHub's terminology so it's easy for anyone to pick up ...
https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization
i.e. `maintainers` or `admin` depending on the nuance

innovaker:
> I believe it is, since that is a catch all used.if nothing else matches.

https://github.com/zephyrproject-rtos/zephyr/blob/master/CODEOWNERS
I'm not seeing that on Zephyr's CODEOWNERS?

petejohanson:
Ok, so I've changed from my name, to just @zmkfirmware/core, and removed the `-teams` suffix. I also removed the "multiple entries" for now, instead just having me on each of the teams for visibility. I can choose to remove myself from either team later depending on how I'm feeling.

@innovaker They chose *not* to have a catch all, in which case no one gets auto assigned. I'd rather we not do that.

Thoughts on this?

Nicell (approved):
Looks good now! Thanks for setting this up.

Issue #575, opened 2021-01-03
document commit guidelines


innovaker:
Let's develop a ZMK orientated simplified summary of [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

It should *concisely* explain:
- why use conventions?
- the commit message format
- an example commit
- title/subject
  - a list of types, and when to use each
  - a list of scopes, and when to use each
- body/description
  - what should be included
    - relevant background
    - reasoning where relevant
- footers
  - formats
  - list of possible footers
 - FAQs?

Use bullet point wherever possible.  The purpose of this page is to be scanned and referenced.

Whoever takes on this task, please contact me before starting development.

Until then, let's use this issue to collate together possible types, scopes and footers.

innovaker:
#### Footers
- `Closes: #xxx` - closes GitHub issue.
- `See: #xxx` - see related GitHub issue or PR.
- `See: b82391a408edd8b933eab595993dd4d39320a647` - see related issue.

mcrosson:
Should it be documented that multiple commits are good for larger change's that are all interwoven/inter dependent? ie. For something like a shield overhaul there should be multiple commits with the different changes made.

Any thoughts about squashes ahead of a merge being needed/desired.

innovaker:
I guess we could include a very concise guide on commit best practices, but keep it short and link to one of the many guides that are available.

mcrosson:
looks like the following are most heavily used in git already for conventional commits

List pulled from `git log --format=oneline | grep -o "[a-z]\+([^)]\+)" | sort | uniq -c`

Types:
  - fix
  - feat
  - docs
  - refactor

Scopes:
  - Docs
  - [Shield]
  - [Feature]
  - [Zone]

mcrosson:
Below is a code snippet that will setup commitlint+husky as a local git hook for `commit-msg` and verification commit messages follow standard conventional commit semantics. This is meant to be run _locally_ and tested with the current dev container.

``` sh
# cd /path/to/zmk
npm install --save-dev @commitlint/{cli,config-conventional}
echo "module.exports = {extends: ['@commitlint/config-conventional']};" > commitlint.config.js
npm install --save-dev husky@next
npm install pinst --save-dev
npx husky install
npx husky add .husky/commit-msg "npx --no-install commitlint --edit $1"
npx commitlint --from HEAD~1 --to HEAD --verbose
touch foo
git commit -m "foo: this will fail"
[...]

PR #586, opened 2021-01-06
chore: add commitlint + config-conventional

innovaker:
Linting for commit messages.

Currently leverages the conventional commits specification configuration which we can fork and tailor as #575 progresses.


innovaker:
My feeling from recent reviews is that the project has reached to a point where it would benefit from linting the commit messages.  I encourage to adopt this as a starting point.  Thoughts?

innovaker:
@petejohanson, any general thoughts on this please?

PR #596, opened 2021-01-08
MA730 Contactless encoder driver

ReFil:
This is an attempt at a driver for the Monolithic Power [MA730](https://www.monolithicpower.com/en/documentview/productdocument/index/version/2/document_type/Datasheet/lang/en/sku/MA730/document_id/3563/), a 14 bit contactless magnetic encoder. It communicates over SPI.
I used @petejohanson s ec11 driver and a the driver for the LSM6DSL (included with zephyr) as a starting point and worked from there
Currently I've got the initialisation and SPI reading functions done, as well as the data structs i need.
Still to do: Timer based polling as i cannot use interrupts with this chip
as an extension to the timer based polling estimate rpm from change in angle and time
handle underflowing gracefully, 
Change resolution based on current layer (if this is possible, using a sensor attr update?)
connect it to the zmk keymap-sensor device same as the ec11

my instantiation code can be seen in https://github.com/ReFil/zmk/tree/rotr/app/boards/arm/rotr/rotr.dts


ReFil:
Closing as has been deprecated by a better driver

myst729:
I use MA730's ABZ output, reusing EC11 driver. It looks fine to me.

PR #597, opened 2021-01-08
docs(codes): Windows Support: Language/International Inputs

KingCoinless:
### Windows version/build?
Windows 10 Version 1909
### What codes did you test?
LANG1
LANG2
LANG3
LANG4
LANG5
INT1
INT2
INT3
INT4
INT5
INT6
### How did you test?
Manual use with my Helix split keyboard using the associated ZMK shield, within both standard QWERTY input and Japanese Microsoft IME.
I was able to get `LANG1` and `LANG2` to both work without issue on Windows. They function to switch between English and Japanese input. However, the latter three codes (`LANG3-5`) and all `INT_` codes did not produce any results for me. Either they are incompatible, or are used exclusively in other languages, and do not serve dual purposes in this case.
Not as far as I am aware, no.

### DIFF
--- a/docs/src/data/hid.js
+++ b/docs/src/data/hid.js
@@ -3378,7 +3378,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=86",
     os: {
-      windows: null,
+      windows: false,
       linux: true,
       android: false,
       macos: null,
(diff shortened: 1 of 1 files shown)


caksoylar (approved):

dxmh (approved):
Thanks for taking the time to do these!

Although we’ve not yet had a second verification on these codes, we’d like to get these merged to keep things moving forward ([as per discussion on Discord][1]). 👍 

[1]: https://discord.com/channels/719497620560543766/883452966114324550/925566006594785280

PR #599, opened 2021-01-08
docs(codes): Windows Support: Media Controls

KingCoinless:
### Windows version/build?
Windows 10 Version 1909
### What codes did you test?
C_REC
C_PLAY
C_PP
C_PAUSE
C_STOP
C_NEXT
C_PREV
C_FF
C_RW
C_SLOW
C_SLOW2
C_REPEAT
C_SHUFFLE
C_CAPTIONS
C_SNAPSHOT
K_PP
K_STOP2
K_STOP3
K_NEXT
K_PREV
K_APP
### How did you test?
Manual use with my Helix split keyboard using the associated ZMK shield. Tested within Foobar2k, MPC-HC, Audacity, and YouTube.
Nearly every Keyboard command (prefix `K_`) did not work whatsoever. `C_PP` and `C_PAUSE` are seemingly synonymous in effect, acting as dual start/stop functions. Interestingly, `C_PLAY` restarts tracks from the beginning. In my testing, `K_APP` has an effect of opening context menus within videos/streamed content in Windows. I believe it may have alternative uses in other specific applications. `C_NEXT` works to play the next track in a playlist within Foobar2k, but I could not get `C_PREV` to revert to a previous track.
Not as far as I am aware, no.

### DIFF
--- a/docs/src/data/hid.js
+++ b/docs/src/data/hid.js
@@ -5896,7 +5896,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=137",
     os: {
-      windows: false,
+      windows: true,
       linux: true,
       android: true,
       macos: null,
--- a/docs/src/data/hid.js
+++ b/docs/src/data/hid.js
@@ -2664,7 +2664,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=86",
     os: {
-      windows: null,
+      windows: true,
       linux: true,
       android: true,
       macos: null,
@@ -4260,7 +4260,7 @@ export default [
     documentation:
       "https://source.android.com/devices/input/keyboard-devices#hid-keyboard-and-keypad-page-0x07",
     os: {
-      windows: null,
+      windows: false,
       linux: true,
       android: true,
       macos: null,
@@ -4282,7 +4282,7 @@ export default [
     documentation:
       "https://source.android.com/devices/input/keyboard-devices#hid-keyboard-and-keypad-page-0x07",
     os: {
-      windows: null,
+      windows: false,
       linux: true,
       android: true,
       macos: null,
@@ -4304,7 +4304,7 @@ export default [
     documentation:
       "https://source.android.com/devices/input/keyboard-devices#hid-keyboard-and-keypad-page-0x07",
     os: {
-      windows: null,
+      windows: false,
       linux: true,
       android: true,
       macos: null,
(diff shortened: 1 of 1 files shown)


docs/src/data/hid.js
```diff
@@ -5896,7 +5896,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=137",
     os: {
-      windows: null,
+      windows: false,
```
caksoylar, line 5899: `C_PREV` in general works for me, specifically for Spotify. Is it possible that it is a foobar-specific issue?

caksoylar (approved):
Looks good to me based on the code changes and the deploy preview. The documented behavior is also consistent for the subset of common keys I tried (play/pause and next), except `C_PREV` also works fine in my experience as I noted in the comment.

If we don't get feedback from the author we might prefer to merge this as is, then I can open a separate PR for `C_PREV`, if folks prefer.

KingCoinless:
> Looks good to me based on the code changes and the deploy preview. The documented behavior is also consistent for the subset of common keys I tried (play/pause and next), except `C_PREV` also works fine in my experience as I noted in the comment.
> 
> If we don't get feedback from the author we might prefer to merge this as is, then I can open a separate PR for `C_PREV`, if folks prefer.

I appreciate you for reviewing many of these old docs PRs. I would attempt to fix this in a secondary PR, but unfortunately, I'm without a proper machine to work on and do so at this time. I will comment and provide feedback as need be otherwise. I don't wish to inconvenience you, but your suggested solution would be the best option at this time.

docs/src/data/hid.js
caksoylar, line 5899:
```suggestion
      windows: true,
```
Thanks for responding @KingCoinless, maybe we can try just accepting this suggestion to fix this PR?

KingCoinless: It may be software-specific. Did you use the web client or the application for Spotify? It may be good to test it on both fronts.

KingCoinless: Sure, I'll commit your suggestion. I haven't worked with Github in a while, so let me know if anything is amiss.

caksoylar: Sorry, I just noticed this. I also tested with Spotify web player and Youtube and they both work well with `C_PREV` as well.

KingCoinless: If it works for you now, I think it's fine to list `C_PREV` as functional in the docs. If you don't mind submitting a commit suggestion, I'll swiftly apply it.

dxmh (approved):

PR #600, opened 2021-01-08
docs(codes): Windows Support: Display Adjustment Commands

KingCoinless:
### Windows version/build?
Windows 10 Version 1909
### What codes did you test?
C_BRI_UP
C_BRI_DN
C_BRI_MIN
C_BRI_MAX
C_BRI_AUTO
C_BKLT_TOG
C_ASPECT
C_PIP
### How did you test?
Manual use with my Helix split keyboard using the associated ZMK shield on the desktop and in video windows.
Nothing but `C_BRI_UP` worked whatsoever to change the brightness or perspective settings of my display.
Not as far as I am aware, no.

### DIFF
--- a/docs/src/data/hid.js
+++ b/docs/src/data/hid.js
@@ -5035,7 +5035,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=134",
     os: {
-      windows: null,
+      windows: false,
       linux: false,
       android: true,
       macos: null,
@@ -5140,7 +5140,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=134",
     os: {
-      windows: null,
+      windows: false,
       linux: true,
       android: null,
       macos: null,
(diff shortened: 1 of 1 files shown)


docs/src/data/hid.js
```diff
@@ -5182,7 +5182,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=134",
     os: {
-      windows: null,
+      windows: false,
```
caksoylar, line 5185:
```suggestion
      windows: true,
```
This also seems to work for me, tested on Windows 11 desktop (shows the brightness slider but doesn't affect the external display) and Windows 10 Surface Book (actually sets screen brightness down as expected.

caksoylar (approved):

I tested all eight on my Surface Book and brightness up/down work well as I noted in the comment. Others don't seem to do anything for me however I haven't tried with a video player application, just on the desktop.

docs/src/data/hid.js
KingCoinless: Looks good, committing changes. An aside, what version of windows is the Surface Book on? My testing was done on Win 10 1909.

caksoylar: Thanks! This is Surface Book Gen 1, on Windows 10 21H2 19044.1415.

dxmh (approved):

PR #605, opened 2021-01-11
fix(boards): Use unused pins for m2 board UART.

petejohanson:
* Current pins conflict w/ m60 shield matrix pins, causing
  odd matrix events when logging/UART is enabled.

### DIFF
--- a/app/boards/arm/nrf52840_m2/nrf52840_m2.dts
+++ b/app/boards/arm/nrf52840_m2/nrf52840_m2.dts
@@ -55,10 +55,10 @@
 	compatible = "nordic,nrf-uarte";
 	status = "okay";
 	current-speed = <115200>;
-	tx-pin = <19>;
-	rx-pin = <21>;
-	rts-pin = <23>;
-	cts-pin = <25>;
+	tx-pin = <42>;
+	rx-pin = <43>;
+	rts-pin = <44>;
+	cts-pin = <45>;
 };
 
 &usbd {


petejohanson:
@megamind4089 or @Nicell have this hardware I believe, to test?

app/boards/arm/nrf52840_m2/nrf52840_m2.dts
```diff
@@ -55,10 +55,10 @@
 	compatible = "nordic,nrf-uarte";
 	status = "okay";
 	current-speed = <115200>;
-	tx-pin = <19>;
-	rx-pin = <21>;
-	rts-pin = <23>;
-	cts-pin = <25>;
+	tx-pin = <1>;
```
Nicell, line 58: This pin is used for the external 32.768kHz crystal. Wondering if that could affect things.

app/boards/arm/nrf52840_m2/nrf52840_m2.dts
```diff
@@ -55,10 +55,10 @@
...
+	rx-pin = <2>;
+	rts-pin = <3>;
+	cts-pin = <4>;
```
Nicell, line 61: All three of these are exposed by the m.2 board.

Nicell (commented):
Taking a look at the m.2, it seems like the pins the only unused pin is P0.17. Every other one seems to be in use. Not sure what we can do about this besides choose some pins that are unlikely to be used by the shield.

app/boards/arm/nrf52840_m2/nrf52840_m2.dts
petejohanson: Ah, certainly could. Good catch.

petejohanson:
@Nicell Yeah, this is certainly a stop-gap. I would like to cherry pick https://github.com/zephyrproject-rtos/zephyr/pull/31068 into ZMK's fork, so we can avoid this for normal usage.

mcrosson:
According to the main MakerDiary nrf52840 docs [here (Apache 2.0 license)](https://github.com/makerdiary/nrf52840-m2/blob/master/boards/arm/nrf52840_m2/nrf52840_m2.dts) the pins on the zmk dts are wrong.

I've got an experimental branch that reconciles the upstream dts with the zmk dts at https://github.com/makerdiary/nrf52840-m2/blob/master/boards/arm/nrf52840_m2/nrf52840_m2.dts

I have one of these boards showing shortly that I plan on making my primary keyboard. If preferred I can pick this up along with any other cleanup/updates once I have a board in-hand.

PR #607, opened 2021-01-14
Clean up uart0 devicetree, don't enable by default

petejohanson:
* Set pro-micro TX/RX pins, but don't enable the UART
  by default.

### DIFF
--- a/app/boards/arm/nice_nano/nice_nano.dts
+++ b/app/boards/arm/nice_nano/nice_nano.dts
@@ -65,12 +65,8 @@
 
 &uart0 {
 	compatible = "nordic,nrf-uarte";
-	status = "okay";
-	current-speed = <115200>;
-	tx-pin = <19>;
-	rx-pin = <21>;
-	rts-pin = <23>;
-	cts-pin = <25>;
+	tx-pin = <6>;
+	rx-pin = <8>;
 };
 
 &usbd {
--- a/app/boards/arm/nrf52840_m2/nrf52840_m2.dts
+++ b/app/boards/arm/nrf52840_m2/nrf52840_m2.dts
@@ -51,16 +51,6 @@
 	status = "okay";
 };
 
-&uart0 {
-	compatible = "nordic,nrf-uarte";
-	status = "okay";
-	current-speed = <115200>;
-	tx-pin = <19>;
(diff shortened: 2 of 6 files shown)


app/boards/arm/nrf52840_m2/nrf52840_m2.dts
```diff
@@ -51,16 +51,6 @@
...
 
-&uart0 {
-	compatible = "nordic,nrf-uarte";
-	status = "okay";
-	current-speed = <115200>;
-	tx-pin = <19>;
-	rx-pin = <21>;
-	rts-pin = <23>;
-	cts-pin = <25>;
-};
```
mcrosson, line 62: According to the makerdiary docs / stock dts file (apache 2.0 licensed) the nrf52840 has a uart and has the following definition per the rest of this PR.

```
&uart0 {
	compatible = "nordic,nrf-uart";
	current-speed = <115200>;
	tx-pin = <16>;
	rx-pin = <15>;
	//rts-pin = <14>;
	//cts-pin = <13>;
};
```

petejohanson: Oh. I see that when plugged into https://wiki.makerdiary.com/nrf52840-m2-devkit/ dev kit, those are the pins they use. That seems to me to be something you'd add to the devkit definition, not to this. one.

mcrosson: Agreed, closing review / comment.

app/boards/arm/nice_nano/nice_nano.dts
```diff
@@ -65,12 +65,9 @@
 
 &uart0 {
 	compatible = "nordic,nrf-uarte";
-	status = "okay";
 	current-speed = <115200>;
```
Nicell, line 68:
```suggestion
```

Nicell (changes requested):

app/boards/arm/nice_nano/nice_nano.dts
petejohanson: Fixed. Good catch.

Nicell (approved):

PR #610, opened 2021-01-15
chore(core): replace `Pete Johanson` with `The ZMK Contributors`

innovaker:
Does not include boards and shields.

Should these also be done?


innovaker:
Forgive any unintentional impertinence.  I came across them when tidying up the headers last month.

petejohanson (approved):

PR #611, opened 2021-01-16
Improved battery SOC measurement by means of LUT and linear interpolation

elagil:
Hi!

I implemented an improved battery state of charge (SOC) estimation algorithm by means of voltage measurement.

The SOC is estimated with a look up table (LUT) that contains values from the discharge curve of the cell. My cell has a curve that [looks like this](https://www.eremit.de/storage/images/Charge%20Diagramm%20LiPo.png?hash=cc32156d2c028d6a1dd13f027ac967296ca13725&shop=84214678&width=720&height=2560).

These values are used as anchors for a linear interpolation between adjacent LUT fields.

While not yet implemented, it would be ideal if one could add the discharge curve of the cell that is used to their own user project, alongside other personal settings (like keymap).

### DIFF
--- a/app/drivers/sensor/battery_voltage_divider/battery_voltage_divider.c
+++ b/app/drivers/sensor/battery_voltage_divider/battery_voltage_divider.c
@@ -42,17 +42,57 @@ struct bvd_data {
     uint8_t state_of_charge;
 };
 
-static uint8_t lithium_ion_mv_to_pct(int16_t bat_mv) {
-    // Simple linear approximation of a battery based off adafruit's discharge graph:
-    // https://learn.adafruit.com/li-ion-and-lipoly-batteries/voltages
+struct charge_level {
+    int16_t mv;
+    uint8_t pct;
+};
 
-    if (bat_mv >= 4200) {
-        return 100;
-    } else if (bat_mv <= 3450) {
-        return 0;
+#define BATTERY_CHARGE_LEVEL_LUT_COUNT 27
+struct charge_level charge_level_lut[BATTERY_CHARGE_LEVEL_LUT_COUNT] = {
+    {3434, 0},  {3457, 4},  {3487, 8},  {3520, 12}, {3545, 15}, {3577, 19}, {3595, 23},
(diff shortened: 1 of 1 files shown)


Nicell (commented):
Hi @elagil!

I see you've modeled this LUT based off a test of your battery's performance. I'm wondering how much more accurate this will be considering the many factors that can affect a battery's voltage like temperature, current draw, age, and more. What are your thoughts on this?

The current linear approximation was chosen due to the presence of all of those factors making voltage a pretty inaccurate way to measure SOC. Dedicated fuel gauges should probably be used for those looking for great accuracy with battery life.

elagil:
Hi @Nicell,

thanks for your comments. Certainly, this method of battery level estimation is not very accurate in the first place, but it is very usable at the edges of the charge curve. Where the slope of the curve is large enough, the algorithm can estimate the level of charge rather well. I read a few articles about this topic and it is generally recommended to estimate SOC with a combined algorithm: Use voltage at the edges of the discharge curve and use a battery gauge (coulomb counter) in between these extremes.

Anyway, in the edge regions, the previous linear function had - by design - the largest error, which I found very unfortunate. These are the areas where the algorithm actually has a chance to deliver accurate predictions.

For example, with the original estimation method, my fully charged battery was reported to be 92 % charged. My particular model only reaches roughly 4.18 V when full, so it was necessary to add this LUT approach for a better representation of charge.

elagil:
@Nicell Another idea, since I don't see much interest in this feature: We could make the current battery estimation a weak definition and let the user overwrite it with a custom function.

Nicell:
@elagil Yeah I'd be more in favor of allowing users to override this existing curve with their own rather than replacing this with one that fits a specific battery's characteristics. Maybe we could make a property on the driver to enter a LUT. That would be easier to edit at the user level IMO.

elagil:
@Nicell Ok, I guess I need to learn how Zephyr works at this point.

petejohanson:
@elagil Are you still working on this?

elagil:
@petejohanson I didn't have time yet, unfortunately, but I still plan to do it.

petejohanson:
> @petejohanson I didn't have time yet, unfortunately, but I still plan to do it.

No worries, just wanted to check in with you. Thanks!

PR #616, opened 2021-01-18
refactor(core): Use /omit-if-no-ref/ for behaviors.

petejohanson:
Found this when digging for some unrelated devicetree stuff. This nicely allows us to avoid compiling in behaviors not in use, *without* requiring new Kconfig values for each behavior. Only those behaviors actually referenced from somewhere in the DT will be kept, and have the generated entries triggering the driver instances to be created.

With this change, buliding my Corne:

```
❯ west build -d build/corne_left
[3/7] Linking C executable zephyr/zephyr_prebuilt.elf
           FLASH:      184436 B       792 KB     22.74%
            SRAM:       56215 B       256 KB     21.44%
        IDT_LIST:         152 B         2 KB      7.42%
[7/7] Linking C executable zephyr/zmk.elf
Converting to uf2, output size: 369152, start address: 0x26000
Wrote 369152 bytes to /home/peter/git/zmk/app/build/corne_left/zephyr/zmk.uf2
```

Current `main`:

```
[216/220] Linking C executable zephyr/zephyr_prebuilt.elf
           FLASH:      188600 B       792 KB     23.26%
            SRAM:       59351 B       256 KB     22.64%
        IDT_LIST:         152 B         2 KB      7.42%
[220/220] Linking C executable zephyr/zmk.elf
Converting to uf2, output size: 377344, start address: 0x26000
Wrote 377344 bytes to /home/peter/git/zmk/app/build/corne_left/zephyr/zmk.uf2
```

Not drastic savings, but one we get "for free", which can help as we try to expand our possible set of target devices.


okke-formsma:
clang is complaining, otherwise lgtm.

okke-formsma (approved):
lgtm, although I believe clang is unhappy about something.

PR #620, opened 2021-01-19
Refactor nibble / update to match build docs

mcrosson:
This update makes the following changes related to the nibble

- Adds a README
- Enables underglow by default that's standard in the assembly docs
- Enables encoder that's part of the assembly docs by default
- Sets up the core OLED config in case users opt to build with the OLED
- Adds labels to the layers
- Adjusts keymaps slightly to better support the default build with encoder present


mcrosson:
@petejohanson Would you be willing to review this once the CI finishes given you just merged a related Nibble update?

app/boards/shields/nibble/nibble.overlay
```diff
@@ -47,3 +54,21 @@ RC(4,0) RC(4,1) RC(4,2) RC(4,3)                 RC(4,6)                 RC(4,9)
 		>;
 	};
 };
+
+&pro_micro_i2c {
+	status = "disabled";
```
Nicell, line 59: I believe you'll want this enabled to have `ZMK_DISPLAY=y` alone turn on the display for you

app/boards/shields/nibble/nibble.keymap
```diff
@@ -8,27 +8,40 @@
 #include <dt-bindings/zmk/keys.h>
 #include <dt-bindings/zmk/bt.h>
 
-#define DEFAULT 0
-#define FUNC 1
+#define LAYER_ID_DEFAULT 0
+#define LAYER_ID_FUNC 1
```
Nicell, line 12: I've never seen `LAYER_ID` before. Is this something we are doing now? I assume it's to avoid conflicts with the layer name.

mcrosson: This is a hold over from the original dev's approach. I'm not seeing it anywhere else and can remove it if preferred.

app/boards/shields/nibble/nibble.overlay
mcrosson: The example in the nibble assembly docs doesn't include an OLED so it's disabled by default. I updated the readme to include a note about changing the status of the pro_micro_i2c to the readme in case folks are looking to enable the OLED instead of enabling it by default where it wouldn't be used by anything.

mcrosson:
@Nicell Can I also bug you about this PR too?

app/boards/shields/nibble/nibble.overlay
Nicell: I'm not sure I understand. That seems to be the point of ZMK_DISPLAY being off by default. Having this disabled as well just means you need to enable two separate things.

app/boards/shields/nibble/nibble.keymap
Nicell: I suppose it's up to you. It seems a bit odd to me, but there's nothing wrong about it.

app/boards/shields/nibble/nibble.overlay
mcrosson: This is a similar approach I took with the tidbit. I can enable the i2c side but i feel like it mildly complicates the builds w/o the need for i2c (and display).

Mind if I drag @petejohanson in for final verdict?

app/boards/shields/nibble/nibble.keymap
```diff
@@ -8,27 +8,40 @@
...
 
 / {
+    sensors {
+		compatible = "zmk,keymap-sensors";
+		sensors = <&encoder_1>;
+	};
```
Nicell, line 18: Seem to be some rogue tabs in here. Spaces are used in the rest of the file.

app/boards/shields/nibble/nibble.keymap
mcrosson: Agreed, do you prefer I go with the ZMK standard or the ported version that's pre-existing?

The only gain to keeping the pre-existing is it won't break keymaps based on the default that's present in main right now.

I'm good with either option.

Nicell (changes requested):

app/boards/shields/nibble/nibble.keymap
mcrosson: Converted to tabs, sorry about that

mcrosson:
@Nicell I think everything is addressed now

app/boards/shields/nibble/nibble.keymap
```diff
@@ -8,30 +8,43 @@
...
+			label = "Function";
+			
+			sensor-bindings = <&inc_dec_kp C_VOLUME_UP C_VOLUME_DOWN>;
+
+			bindings = <
+			&kp TILDE   &kp F1    &kp F2    &kp F3    &kp F4    &kp F5    &kp F6    &kp F7    &kp F8    &kp F9    &kp F10   &kp F11   &kp F12 &trans      &kp END
+&kp C_MUTE  &trans      &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans  &trans      &bootloader
+&bt BT_CLR  &trans      &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans            &trans      &trans
 &bt BT_PRV  &trans      &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans    &trans  &trans      &trans
 &bt BT_NXT  &trans      &trans    &trans                        &trans                        &trans    &trans    &trans    &kp C_PREV        &kp C_PP    &kp C_NEXT
```
Nicell, line 46: These layers  seem to be misaligned compared to how they were before. Columns aren't lining up.

app/boards/shields/nibble/nibble.keymap
Nicell: I believe the ones that are in main don't have LAYER_ID seeing as GitHub is marking this as a change, no?

Nicell (changes requested):

app/boards/shields/nibble/nibble.keymap
mcrosson: Taken care of now

app/boards/shields/nibble/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## OLED Builds
+
+If using an OLED screen, you'll need to change the following in your local nibble config or add them to the end of the file.
+
+```
+CONFIG_ZMK_DISPLAY=y
+```
+
+```
```
Nicell, line 31:
```suggestion
```

Nicell (changes requested):

app/boards/shields/nibble/README.md
mcrosson: This should be removed in the latest code

Nicell (approved):
Thanks for the update!

PR #624, opened 2021-01-19
Fix offset in NIBBLE transformation and keymap

jaygreco:
The NIBBLE code was written before the unified ANSI + ISO board was released, which added an additional key to support ISO left shift. After the key was added (row 0, column 0), the matrix transformation was missing RC(0,0), which caused some strange behavior in the kscan driver: pressing the 7 key sent an esc/alt.

Adding RC(0,0) and updating the default keymap to match this resolves the issue. &trans is used in the new slot in the keymap and can be replaced with the pipe/backslash key if a user is building with an ISO layout.

Tested locally on my dev NIBBLE and also by @mcrosson.


mcrosson:
@petejohanson / @innovaker I can confirm this works and without it a LOT of bash/ssh/putty users are going to be very unhappy.

mcrosson (approved):

petejohanson (approved):

PR #625, opened 2021-01-21
run-test.sh: use the POSIX '-L 1' for xargs

jeffrizzo:
This allows tests to be run on platforms where '-l'
isn't implemented, such as MacOS.


petejohanson (approved):
Thanks!

PR #626, opened 2021-01-22
Add optional `binding_convert_central_state_dependent_params` callback to the behavior driver API.

petejohanson:
Based on [discussions](https://discord.com/channels/719497620560543766/719565084208398406/801454819340189726) in Discord, this adds a new (optional) behavior driver callback, that allows behaviors to convert a relative binding invocation (e.g. "Toggle Power" or "Next Profile", into an *absolute* one.

This is *mostly* important for behaviors that, after #547, won't only be executed locally, but maybe "globally" on central and peripheral, we can ensure all devices converge to the same final state, not be out of sync.

I've only implemented this so far for ext power and RGB, since the "profile select" *only* matters for central side anyways.

### DIFF
--- a/app/src/behaviors/behavior_rgb_underglow.c
+++ b/app/src/behaviors/behavior_rgb_underglow.c
@@ -25,6 +25,10 @@ static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
     switch (binding->param1) {
     case RGB_TOG_CMD:
         return zmk_rgb_underglow_toggle();
+    case RGB_ON_CMD:
+        return zmk_rgb_underglow_on();
+    case RGB_OFF_CMD:
+        return zmk_rgb_underglow_off();
     case RGB_HUI_CMD:
         return zmk_rgb_underglow_change_hue(1);
     case RGB_HUD_CMD:
--- a/app/src/behaviors/behavior_rgb_underglow.c
+++ b/app/src/behaviors/behavior_rgb_underglow.c
@@ -33,12 +33,57 @@ on_keymap_binding_convert_central_state_dependent_params(struct zmk_behavior_bin
         }
 
         binding->param1 = state ? RGB_OFF_CMD : RGB_ON_CMD;
-        LOG_DBG("RGB relative toggle convert to absolute %s", state ? "OFF" : "ON");
-        return 0;
+        break;
+    }
+    case RGB_BRI_CMD: {
+        struct zmk_led_hsb color = zmk_rgb_underglow_calc_brt(1);
+
+        binding->param1 = RGB_COLOR_HSB_CMD;
+        binding->param2 = RGB_COLOR_HSB_VAL(color.h, color.s, color.b);
+        break;
+    }
+    case RGB_BRD_CMD: {
+        struct zmk_led_hsb color = zmk_rgb_underglow_calc_brt(-1);
+
+        binding->param1 = RGB_COLOR_HSB_CMD;
+        binding->param2 = RGB_COLOR_HSB_VAL(color.h, color.s, color.b);
+        break;
+    }
+    case RGB_HUI_CMD: {
+        struct zmk_led_hsb color = zmk_rgb_underglow_calc_hue(1);
+
+        binding->param1 = RGB_COLOR_HSB_CMD;
+        binding->param2 = RGB_COLOR_HSB_VAL(color.h, color.s, color.b);
+        break;
+    }
+    case RGB_HUD_CMD: {
+        struct zmk_led_hsb color = zmk_rgb_underglow_calc_hue(-1);
+
+        binding->param1 = RGB_COLOR_HSB_CMD;
+        binding->param2 = RGB_COLOR_HSB_VAL(color.h, color.s, color.b);
+        break;
+    }
+    case RGB_SAI_CMD: {
+        struct zmk_led_hsb color = zmk_rgb_underglow_calc_sat(1);
+
+        binding->param1 = RGB_COLOR_HSB_CMD;
+        binding->param2 = RGB_COLOR_HSB_VAL(color.h, color.s, color.b);
+        break;
+    }
+    case RGB_SAD_CMD: {
+        struct zmk_led_hsb color = zmk_rgb_underglow_calc_sat(-1);
+
+        binding->param1 = RGB_COLOR_HSB_CMD;
+        binding->param2 = RGB_COLOR_HSB_VAL(color.h, color.s, color.b);
+        break;
     }
     default:
         return 0;
     }
+
+    LOG_DBG("RGB relative convert to absolute (%d/%d)", binding->param1, binding->param2);
+
+    return 0;
 };
 
 static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
@@ -71,8 +116,9 @@ static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
     case RGB_EFR_CMD:
         return zmk_rgb_underglow_cycle_effect(-1);
     case RGB_COLOR_HSB_CMD:
-        return zmk_rgb_underglow_set_hsb((binding->param2 >> 16) & 0xFFFF,
-                                         (binding->param2 >> 8) & 0xFF, binding->param2 & 0xFF);
+        return zmk_rgb_underglow_set_hsb((struct zmk_led_hsb){.h = (binding->param2 >> 16) & 0xFFFF,
+                                                              .s = (binding->param2 >> 8) & 0xFF,
+                                                              .b = binding->param2 & 0xFF});
     }
 
     return -ENOTSUP;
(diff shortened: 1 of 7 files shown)


okke-formsma:
Instead of doing the implementation this way, why not do the secondary behavior call from the 'relative' behavior? 

I think a simpler solution is to create a new event struct and call ```behavior_keymap_binding_pressed(&binding, new_event);``` like hold_tap and other methods are already doing when they need to invoke another behavior.

app/src/behaviors/behavior_rgb_underglow.c
```diff
@@ -18,11 +18,35 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 static int behavior_rgb_underglow_init(const struct device *dev) { return 0; }
 
+static int on_keymap_binding_to_absolute(struct zmk_behavior_binding *binding,
```
okke-formsma, line 21: I'm not sure about this name, maybe something like "preprocess on central"?

petejohanson: That feels like naming it after the current one consumer of the API, versus declaring the generic intent of the API.

okke-formsma: I feel like the "'on_keymap_binding_to_absolute" actually names it after the current one consumer? "preprocess" is much more general then "to absolute", and this callback is only executed on the central so I think it's worth mentioning.

petejohanson: Hmm..... `binding_convert_central_state_dependent_params`?

petejohanson:
@Nicell are you able to test the RGB changes here?

Nicell (commented):
I see this only adds absolute calls for on/off of RGB. Are you planning to have absolute calls for hue, saturation, brightness, speed, and effect?

Confirmed that toggling still works on my nice60.

petejohanson:
> I see this only adds absolute calls for on/off of RGB. Are you planning to have absolute calls for hue, saturation, brightness, speed, and effect?
> 
> Confirmed that toggling still works on my nice60.

Thank you! Totally missed this, and we can use the set color stuff from @mcrosson to do that, can't we?

Nicell:
Yeah, we can use it for hue, saturation, and brightness, but we'll need something else for speed and effect.

okke-formsma (approved):
lgtm

app/src/rgb_underglow.c
```diff
@@ -313,63 +335,70 @@ int zmk_rgb_underglow_cycle_effect(int direction) {
...
-            rc = ext_power_disable(ext_power);
-        }
+struct zmk_led_hsb zmk_rgb_underglow_calc_hue(int direction) {
+    struct zmk_led_hsb color = state.color;
 
-        if (rc != 0) {
-            LOG_ERR("Unable to toggle EXT_POWER: %d", rc);
-        }
+    if (color.h == 0 && direction < 0) {
+        return color;
```
Nicell, line 355: I don't think this is sufficient, and it also doesn't support wrapping around. Instead of this, after updating the hue with `+= direction * CONFIG_ZMK_RGB_UNDERGLOW_HUE_STEP`, if the result is less than 0, add 360.

app/src/rgb_underglow.c
```diff
@@ -313,63 +335,70 @@ int zmk_rgb_underglow_cycle_effect(int direction) {
...
+    color.h = color.h % 360;
 
-        k_timer_stop(&underglow_tick);
+    return color;
+}
+
+struct zmk_led_hsb zmk_rgb_underglow_calc_sat(int direction) {
+    struct zmk_led_hsb color = state.color;
+
+    if (color.s == 0 && direction < 0) {
```
Nicell, line 368:
```suggestion
    if (color.s - CONFIG_ZMK_RGB_UNDERGLOW_SAT_STEP < 0 && direction < 0) {
```
Something like this is better. Or even setting to zero if it's less than 0 after the fact. If the step is more than 1, this won't catch all cases (which I doubt many people would use 1).

app/src/rgb_underglow.c
```diff
@@ -313,63 +335,70 @@ int zmk_rgb_underglow_cycle_effect(int direction) {
...
+    return color;
 }
 
-int zmk_rgb_underglow_set_hsb(uint16_t hue, uint8_t saturation, uint8_t brightness) {
-    if (hue > 360 || saturation > 100 || brightness > 100) {
-        return -ENOTSUP;
+struct zmk_led_hsb zmk_rgb_underglow_calc_brt(int direction) {
+    struct zmk_led_hsb color = state.color;
+
+    if (color.b == 0 && direction < 0) {
```
Nicell, line 384: Same idea here as above.

Nicell (changes requested):
Noticed some bad logic that you've inherited from me. Other than that looks pretty good. HSB refactor seems nice.

petejohanson:
@Nicell Ok, fixed up the calc functions, proper looping for hue, proper clamping for saturation and brightness. Thoughts? Tested fine on my left half of my Kyria.

app/src/rgb_underglow.c
```diff
@@ -292,101 +265,144 @@ int zmk_rgb_underglow_save_state() {
...
+        return 0;
     }
 
+    state.current_effect += direction;
+
+    if (state.current_effect >= UNDERGLOW_EFFECT_NUMBER) {
+        state.current_effect = 0;
+    }
+
+    state.animation_step = 0;
```
Nicell, line 336:
```suggestion
    state.current_effect += direction % UNDERGLOW_EFFECT_NUMBER;

    if (state.current_effect < 0) {
        state.current_effect += UNDERGLOW_EFFECT_NUMBER;
    }

    state.animation_step = 0;
```
I think this simplifies the logic and would support the possibility of a different direction that -1 or 1 (if someone wants that for some reason). Also the return statement on line 327 probably should be `zmk_rgb_underglow_save_state()` if we don't switch to this.

app/src/rgb_underglow.c
```diff
@@ -292,101 +265,144 @@ int zmk_rgb_underglow_save_state() {
...
 }
 
-int zmk_rgb_underglow_set_hsb(uint16_t hue, uint8_t saturation, uint8_t brightness) {
-    if (hue > 360 || saturation > 100 || brightness > 100) {
+int zmk_rgb_underglow_toggle() {
+    return state.on ? zmk_rgb_underglow_off() : zmk_rgb_underglow_on();
+}
+
+int zmk_rgb_underglow_set_hsb(struct zmk_led_hsb color) {
+    if (color.h > HUE_MAX || color.s > SAT_MAX || color.b > BRT_MAX) {
```
Nicell, line 346: Should we also check for negatives?

app/src/rgb_underglow.c
```diff
@@ -292,101 +265,144 @@ int zmk_rgb_underglow_save_state() {
...
-    if (!led_strip)
-        return -ENODEV;
+struct zmk_led_hsb zmk_rgb_underglow_calc_hue(int direction) {
+    struct zmk_led_hsb color = state.color;
 
-    if (state.hue == 0 && direction < 0) {
-        state.hue = 360 - CONFIG_ZMK_RGB_UNDERGLOW_HUE_STEP;
-        return 0;
+    color.h += HUE_MAX + (direction * CONFIG_ZMK_RGB_UNDERGLOW_HUE_STEP);
+    color.h %= HUE_MAX;
```
Nicell, line 359:
```suggestion
    color.h %= HUE_MAX;
    
    if (color.h < 0) {
        color.h += HUE_MAX;
    }
```
I think this is necessary

Nicell (changes requested):

app/src/rgb_underglow.c
petejohanson: @Nicell  Double check the line above this one. It adds an *extra* `HUE_MAX` before performing the modulo.

Nicell: Ah! Missed that. Clever, makes sense.

app/src/rgb_underglow.c
petejohanson: Used a similar modulo trick as in the other areas instead. Just pushed the updated version.

app/src/rgb_underglow.c
```diff
@@ -145,32 +126,26 @@ static void zmk_rgb_underglow_effect_breathe() {
 
 static void zmk_rgb_underglow_effect_spectrum() {
     for (int i = 0; i < STRIP_NUM_PIXELS; i++) {
-        int hue = state.animation_step;
-        int sat = state.saturation;
-        int brt = state.brightness;
-
-        struct led_hsb hsb = {hue, sat, brt};
+        struct zmk_led_hsb hsb = state.color;
+        hsb.h = state.animation_speed;
```
Nicell, line 130:
```suggestion
        hsb.h = state.animation_step;
```

Nicell (changes requested):

app/src/rgb_underglow.c
petejohanson: Good catch! Fixed.

Nicell (approved):

PR #635, opened 2021-01-25
fix(behaviors): Updated grsec snapshots for new log format.

petejohanson:


okke-formsma:
lgtm I'm not sure why the tests are still 'failing' here, even though you had a good run too:

image

petejohanson:
@okke-formsma That's because they're different workflows. The "Build" for all the boards is fine, the integration tests is the "Tests" workflow, which is still failing for the implicit mods test.

PR #636, opened 2021-01-25
Add a missing log_strdup()

jeffrizzo:
Quiets a log error when debug logging in `ble.c` is enabled.


petejohanson (approved):
Good catch, thanks!

petejohanson:
Tests are failing, because this was forked when we had the small regression in tests in `main`. Merging.

PR #640, opened 2021-01-28
GPIO driver refactor - polling, multiple instance fixes, remove default Kconfigs

petejohanson:
This refactor is needed for my upcoming Ferris board work. In particular, the Ferris:

* Uses the composite kscan driver, to compose the left and right half matrixes.
* Uses matrix polling, which seems to have been broken by the interrupt work, done for sleep recently.

While there, I took the opportunity to add some more useful macros to make the conditional code a little bit easier to read.

There's also a small commit in here as well that removes some `default n` settings from some Kconfig settings to allow them to be defaulted yes elsewhere as needed.

### DIFF
--- a/app/drivers/kscan/kscan_gpio_matrix.c
+++ b/app/drivers/kscan/kscan_gpio_matrix.c
@@ -51,6 +51,11 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
 }
 #endif
 
+#define COND_POLLING(code) COND_CODE_1(CONFIG_ZMK_KSCAN_MATRIX_POLLING, (code), ())
+#define COND_INTERRUPTS(code) COND_CODE_1(CONFIG_ZMK_KSCAN_MATRIX_POLLING, (), (code))
+#define COND_POLL_OR_INTERRUPTS(pollcode, intcode)                                                 \
+    COND_CODE_1(CONFIG_ZMK_KSCAN_MATRIX_POLLING, pollcode, intcode)
+
 #define INST_MATRIX_ROWS(n) DT_INST_PROP_LEN(n, row_gpios)
 #define INST_MATRIX_COLS(n) DT_INST_PROP_LEN(n, col_gpios)
 #define INST_OUTPUT_LEN(n)                                                                         \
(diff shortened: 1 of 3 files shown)


app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -207,25 +220,22 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
         return 0;                                                                                  \
     };                                                                                             \
     static int kscan_gpio_enable_##n(const struct device *dev) {                                   \
-        COND_CODE_1(CONFIG_ZMK_KSCAN_MATRIX_POLLING,                                               \
-                    (struct kscan_gpio_data_##n *data = dev->data;                                 \
-                     k_timer_start(&data->poll_timer, K_MSEC(10), K_MSEC(10)); return 0;),         \
-                    (int err = kscan_gpio_enable_interrupts_##n(dev);                              \
-                     if (err) { return err; } return kscan_gpio_read_##n(dev);))                   \
+        COND_POLL_OR_INTERRUPTS((struct kscan_gpio_data_##n *data = dev->data;                     \
+                                 k_timer_start(&data->poll_timer, K_MSEC(10), K_MSEC(10));         \
```
joelspadin, line 224: Am I correctly interpreting this that when polling is enabled, we poll at a constant 100 Hz? It might be useful to make that configurable, for example if someone wanted to make a wired keyboard that polls as fast as possible.

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -262,19 +272,21 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
...
         }                                                                                          \
         data->dev = dev;                                                                           \
-        COND_CODE_1(CONFIG_ZMK_KSCAN_MATRIX_POLLING,                                               \
-                    (k_timer_init(&data->poll_timer, kscan_gpio_timer_handler, NULL);), ())        \
         (COND_CODE_0(DT_INST_PROP(n, debounce_period), (k_work_init), (k_delayed_work_init)))(     \
             &data->work, kscan_gpio_work_handler_##n);                                             \
+        COND_POLL_OR_INTERRUPTS(                                                                   \
+            (k_timer_init(&data->poll_timer, kscan_gpio_timer_handler_##n, NULL);                  \
+             kscan_gpio_set_output_state_##n(dev, 0);),                                            \
+            (kscan_gpio_set_output_state_##n(dev, 1);))                                            \
```
joelspadin, line 289: Would I be correct in guessing that the interrupt behavior is different here because it sets all the outputs active, waits for an interrupt to happen on any input, and then does a regular scan?

joelspadin (approved):
I'm not sure I fully understand how the kscan driver works, but everything I see changing here looks reasonable to me.

app/drivers/kscan/kscan_gpio_matrix.c
petejohanson: Yeah, I hadn't remembered hardcoding.this, but it does seem so, yeah. Good item to move into Kconfig soon, yeah.

app/drivers/kscan/kscan_gpio_matrix.c
petejohanson: Correct. We want to drive all outputs to active when "idle", so that any key press will trigger an interrupt. Then we keep polling/scanning until all keys are released again.

petejohanson (commented):
Self approving, been using this extensively for several days on Ferris and on other interrupt based boards like Corne.

PR #642, opened 2021-01-29
New board: Ferris 0.2

petejohanson:
New non-wireless stm32 boards, the Ferris 0.2. No underglow yet, but basic matrix/typing is working great.

Based on the initial work in #640 and then two more commits on top of it.

The first adds a mcp23017 driver, based on the upstream mcp23s17 driver. Upstream is working on a driver, so instead adding to our fork of Zephyr, I'd prefer we add privately, and then remove once we update Zephyr versions to a new release.


pierrechevalier83:
I can confirm that the firmware works as intended for me. I tested it with a keymap that's pretty close to what I use with qmk (https://gist.github.com/pierrechevalier83/4b7648f07ac8e3f899a8701d273a573b) and mod tap behaviour felt good. Everything functioned as expected.

Couple of notes for things that can be improved later:
* For me, the keyboard wasn't showing in lsusb if a working right hand wasn't plugged.
* I'm aware the backlight isn't yet supported.

dzervas:
I confirm that the MCP23017 part works for me too: https://github.com/dzervas/zmk-config

PR #643, opened 2021-01-29
Include sensor value with sensor_event

allymparker:
Add the sensor value to the sensor event and remove the device/sensor struct.

I've tested this works on my sofle as I didn't see any tests for encoders yet.

One thing I did notice while testing is the there are a TON of false-positive events for the sensor where it's just sending zero... I added a check before raising the initial event to reduce them, but I'm not sure if that's desired or not.

### DIFF
--- a/app/include/drivers/behavior.h
+++ b/app/include/drivers/behavior.h
@@ -23,8 +23,7 @@
 typedef int (*behavior_keymap_binding_callback_t)(struct zmk_behavior_binding *binding,
                                                   struct zmk_behavior_binding_event event);
 typedef int (*behavior_sensor_keymap_binding_callback_t)(struct zmk_behavior_binding *binding,
-                                                         const struct device *sensor,
-                                                         int64_t timestamp);
+                                                         int32_t value, int64_t timestamp);
 
 __subsystem struct behavior_driver_api {
     behavior_keymap_binding_callback_t binding_convert_central_state_dependent_params;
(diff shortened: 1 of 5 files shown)


petejohanson:
> Add the sensor value to the sensor event and remove the device/sensor struct.
> 
> I've tested this works on my sofle as I didn't see any tests for encoders yet.
> 
> One thing I did notice while testing is the there are a TON of false-positive events for the sensor where it's just sending zero... I added a check before raising the initial event to reduce them, but I'm not sure if that's desired or not.

The zeros are due to the resolution setting. We send the raw delta in the second sensor channel data value, and the value divided by the resolution in the first one.

This allows us to have folks react to low level deltas, or high level/resolution dependent deltas.

app/include/drivers/behavior.h
```diff
@@ -23,8 +23,7 @@
 typedef int (*behavior_keymap_binding_callback_t)(struct zmk_behavior_binding *binding,
                                                   struct zmk_behavior_binding_event event);
 typedef int (*behavior_sensor_keymap_binding_callback_t)(struct zmk_behavior_binding *binding,
-                                                         const struct device *sensor,
-                                                         int64_t timestamp);
+                                                         int32_t value, int64_t timestamp);
```
petejohanson, line 26: So, the main reason for the initial design, was the fact that when we have more complex sensors, like trackpoints, we should ideally be able to use the same sensor event, w/ more channels of data, for this.

Unfortunately, that then leads to structs w/ dynamic length arrays, requiring more complicated allocation/free strategies, among other complications.

This is one of the largest challenges with any work here, scope. If we want to get a first pass that just supports single hardcoded sensor channel to get encoders going, that puts us one step back on more complete pointer support.

Need to brainstorm/sleep on this.

allymparker: So I've been having more of a think about this.  None of this is criticism, just observations based on previous experience: 

* Part of the reason this is hard is that we're trying to abstract a concept early (trying to create a generic sensor event). We don't have a pointer implementation so there's ambiguity about how it's implemented. 
* I've previously used what I've dubbed "rule of three" to allow teams to move past abstraction ambiguity previously.  It goes something like "Until we have 3 examples of something, the chances of us creating the right abstraction are low (YAGNI), therefore until we are in that position, why don't we create the most naive/simple version for each instance to make usage specific.  When we have the next implementation we can refactor into a likely better abstraction"
* The naive implementation in this PR is not great.  I've essentially violated the Open Closed principal by using a basic value type as the argument in all the APIs.  This can be resolved using a struct.  This allows the api to be open to extension (we can add new sensor values to the struct) but closed to modification (the api doesn't need to change to add new functionality)

Here's what I'm thinking, at the moment, there's no data channel being past in the zephyr event.  Adding a new struct with the current 2 values moves us in the right direction, without closing any doors for later more complex sensor events.

I've got an hour now, so I'll give that a go to see what you think.

PS 
As a side note for pointing devices, I believe we only need a handful of well known values: X, Y, V/Scroll up, H/Scrolldown and buttons potentially represented as a bit mask.  These could just be represented as 5 integers in the event and avoid the dynamic array lengths?

PR #644, opened 2021-01-30
fix(docs): Set description to our tagline

Nicell:


petejohanson (approved):
Yes! Has regularly bugged me when posting links to social media. Thanks!

PR #649, opened 2021-01-30
feat(behavior) Add &sleep behavior

mcrosson:
Add a new behavior that'll allow users to put their keyboard to sleep ahead of timeouts. This is meant to be used as a binding on a keymap.


app/include/zmk/activity.h
```diff
@@ -8,4 +8,5 @@
 
 enum zmk_activity_state { ZMK_ACTIVITY_ACTIVE, ZMK_ACTIVITY_IDLE, ZMK_ACTIVITY_SLEEP };
 
-enum zmk_activity_state zmk_activity_get_state();
\ No newline at end of file
+enum zmk_activity_state zmk_activity_get_state();
+int activity_set_state(enum zmk_activity_state state);
```
Nicell, line 12:
```suggestion
int zmk_activity_set_state(enum zmk_activity_state state);
```
To keep in line with the rest of our shared functions, I'm thinking this should be prefixed with `zmk_`.

app/src/activity.c
```diff
@@ -47,19 +48,21 @@ enum zmk_activity_state zmk_activity_get_state() { return activity_state; }
...
 
 void activity_work_handler(struct k_work *work) {
     int32_t current = k_uptime_get();
     int32_t inactive_time = current - activity_last_uptime;
 #if IS_ENABLED(CONFIG_ZMK_SLEEP)
-    if (inactive_time > MAX_SLEEP_MS) {
-        set_state(ZMK_ACTIVITY_SLEEP);
+    // Second half of || statement is to ensure a user using &sleep doesn't lose sleep state when
+    // the idle or sleep timers expire
+    if (inactive_time > MAX_SLEEP_MS || activity_state == ZMK_ACTIVITY_SLEEP) {
```
Nicell, line 60: I'm wondering if this handler is even called when ZMK is in deep sleep. I'd imagine it shouldn't, but if it does, why don't we just clear the timer? Seems unnecessary to run this while we're in sleep.

app/src/behaviors/behavior_sleep.c
```diff
@@ -0,0 +1,30 @@
...
+static int behavior_sleep_init(const struct device *dev) { return 0; };
+
+static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
+                                      struct zmk_behavior_binding_event event) {
+    return activity_set_state(ZMK_ACTIVITY_SLEEP);
+}
+
+static const struct behavior_driver_api behavior_sleep_driver_api = {
+    .binding_released = on_keymap_binding_released,
+};
```
Nicell, line 27: `.binding_pressed` should be defined so you don't get an error in logs when pressing this key. You can have a function that just returns an opaque response like this: https://github.com/zmkfirmware/zmk/blob/main/app/src/behaviors/behavior_rgb_underglow.c#L54

app/src/behaviors/behavior_sleep.c
```diff
@@ -0,0 +1,30 @@
...
+#include <zmk/behavior.h>
+#include <zmk/activity.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+static int behavior_sleep_init(const struct device *dev) { return 0; };
+
+static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
+                                      struct zmk_behavior_binding_event event) {
+    return activity_set_state(ZMK_ACTIVITY_SLEEP);
```
Nicell, line 22:
```suggestion
#if IS_ENABLED(CONFIG_ZMK_SLEEP)
    return activity_set_state(ZMK_ACTIVITY_SLEEP);
#endif
    return ZMK_BEHAVIOR_OPAQUE;
```
This is so we don't sleep when sleep isn't turned on and we press this key.

Nicell (changes requested):

app/dts/behaviors/sleep.dtsi
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/ {
+	behaviors {
+		sleep: behavior_sleep {
```
petejohanson, line 9: Please add the `/omit-if-no-ref/` here like other behaviors were updated to include.

app/src/behaviors/behavior_sleep.c
```diff
@@ -0,0 +1,30 @@
...
+#include <device.h>
+#include <drivers/behavior.h>
+#include <logging/log.h>
+
+#include <zmk/behavior.h>
+#include <zmk/activity.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+static int behavior_sleep_init(const struct device *dev) { return 0; };
```
petejohanson, line 18: Add a wrapper `#if HAS_DEVICE_STATUS_OKAY(DT_DRV_COMPAT)` like other drivers do now, so this doesn't compile if the DT is omitted.

petejohanson (changes requested):
Few additional tweaks needed.

petejohanson:
@mcrosson did you see the review items on this?

mcrosson:
> 
> 
> @mcrosson did you see the review items on this?

Yes, I thought it was agreed this was 'on hold' until after locality landed? I was planning on working on any necessary changes once locality landed.

app/include/zmk/activity.h
mcrosson: Changed

app/src/behaviors/behavior_sleep.c
mcrosson: Added

app/src/behaviors/behavior_sleep.c
mcrosson: Added

app/dts/behaviors/sleep.dtsi
mcrosson: Added

app/src/behaviors/behavior_sleep.c
mcrosson: Added

app/src/activity.c
mcrosson: I didn't see a great spot to reset the timer and if the timer code does run at all in sleep mode... we're still going to need the 2nd half of the condition to ensure it's never going to wake up when asleep.

mcrosson:
Had some spare brain cycles and time ; updated according to the above feedback

app/src/activity.c
```diff
@@ -34,7 +34,8 @@ int raise_event() {
         (struct zmk_activity_state_changed){.state = activity_state}));
 }
 
-int set_state(enum zmk_activity_state state) {
+int activity_set_state(enum zmk_activity_state state) {
```
Nicell, line 37: I think you have to rename all of these `activity_set_state`s to `zmk_activity_set_state`.

app/src/activity.c
Nicell: What I'm thinking is this handler shouldn't even be called once you're in deep sleep. Have you tested it without this additional conditional? From my understanding when we go into deep sleep, timer-based handlers are no longer run since we don't set up any timer wakeups and therefore don't have any timers running.

Nicell (changes requested):

app/src/behaviors/behavior_sleep.c
```diff
@@ -0,0 +1,43 @@
...
+}
+
+static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
+                                     struct zmk_behavior_binding_event event) {
+    return ZMK_BEHAVIOR_OPAQUE;
+}
+
+static const struct behavior_driver_api behavior_sleep_driver_api = {
+    .binding_released = on_keymap_binding_released,
+    .binding_pressed = on_keymap_binding_pressed,
```
petejohanson, line 37: This should have a `.locality` set:

```suggestion
    .binding_pressed = on_keymap_binding_pressed,
    .locality = BEHAVIOR_LOCALITY_GLOBAL,
```

docs/docs/behaviors/misc.md
```diff
@@ -40,3 +40,22 @@ Example:
...
+### Behavior Binding
+
+- Reference: `&sleep`
+- Parameters: None
+
+Example:
+
+```
+&sleep
+```
```
petejohanson, line 61: Please document this as being a global behavior, that will affect all components of a split. See other global behaviors, e.g. RGB, for an example docs.

petejohanson (changes requested):
A few follow ups.

petejohanson:
@mcrosson Any plans to wrap this up, fix current conflicts, etc?

hyx0329:
I was trying to implement soft power off/lockout in the application code, happened to almost reimplement this.

However, regardless of whether I called `set_state`, I can not wake up my keyboard through typing. This works great for my "power off" purpose, but not for the "sleep" here. When the keyboard is put to sleep by `activity_work_handler` in `activity.c`, I can wake the keyboard up by typing. So I wonder which part of code make it possible to wake up the keyboard by typing.

In my code, the behavior is triggered upon pressing the key. My code is based on current latest main branch (https://github.com/zmkfirmware/zmk/commit/94789a092c6d8ddf5ab5ad88eb011a4ece6b5cee).

Nick-Munnich:
I recall someone requesting a &sleep behaviour in addition to &soft_off, to use &sleep for power conserving and &soft_off for transportation. For that reason I would say that it's not obsolete yet.

PR #652, opened 2021-01-30
fix(core): Assert BLE device name is correct length

### LINKED ISSUE #613: Enforce maximum keyboard name length
Setting too long a value for `ZMK_KEYBOARD_NAME` causes BLE advertising to fail mysteriously for users.

For new users first writing a new shield/board this can be confusing.and frustrating.

We should use a preprocessor check in `src/ble.c` that uses `#error "name too long"` to fail the build in this case.

Exact maximum length needs validation/testing.

Nicell:
Closes #613. I found the max length by incrementally adding a character until it broke. 16 seems to be the maximum. While researching I found that the max should be 29 for BLE (31 bytes - 2 bytes for header), so I'm not sure where the limitation is.

Preprocessor `#if` and `#error` cannot be used because you can't compare a `sizeof` output.

### DIFF
--- a/app/boards/shields/splitreus62/Kconfig.defconfig
+++ b/app/boards/shields/splitreus62/Kconfig.defconfig
@@ -16,7 +16,7 @@ endif
 if SHIELD_SPLITREUS62_RIGHT
 
 config ZMK_KEYBOARD_NAME
-	default "Splitreus62 Right"
+	default "Splitreus62 Rt"
 
 endif
 
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -63,6 +63,8 @@ static uint8_t active_profile;
 #define DEVICE_NAME CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
 
+BUILD_ASSERT(DEVICE_NAME_LEN <= 16, "ERROR: BLE device name is too long. Max length: 16");
+
 #define IS_HOST_PERIPHERAL                                                                         \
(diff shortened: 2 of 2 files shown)


petejohanson:
> Closes #613. I found the max length by incrementally adding a character until it broke. 16 seems to be the maximum. While researching I found that the max should be 29 for BLE (31 bytes - 2 bytes for header), so I'm not sure where the limitation is.
> 
> Preprocessor `#if` and `#error` cannot be used because you can't compare a `sizeof` output.

I believe that's the max for the whole advertising data, we also have other data in there:

```
static const struct bt_data zmk_ble_ad[] = {
#if IS_HOST_PERIPHERAL
    BT_DATA(BT_DATA_NAME_COMPLETE, DEVICE_NAME, DEVICE_NAME_LEN),
    BT_DATA_BYTES(BT_DATA_GAP_APPEARANCE, 0xC1, 0x03),
#endif
    BT_DATA_BYTES(BT_DATA_FLAGS, (BT_LE_AD_GENERAL | BT_LE_AD_NO_BREDR)),
    BT_DATA_BYTES(BT_DATA_UUID16_SOME,
#if IS_HOST_PERIPHERAL
                  0x12, 0x18, /* HID Service */
#endif
                  0x0f, 0x18 /* Battery Service */
                  ),
#if IS_SPLIT_PERIPHERAL
    BT_DATA_BYTES(BT_DATA_UUID128_ALL, ZMK_SPLIT_BT_SERVICE_UUID)
#endif
};
```
```

petejohanson:
@Nicell check seems reasonable. Are you planning to add a second commit that fixes up the failing boards?

petejohanson (approved):

PR #653, opened 2021-01-31
feat(docs): Generate hardware support page from board metadata

Nicell:
Fields I've added to the Zephyr yaml:

```
manufacturer [required]
    creator or producer of the hardware

compatible [optional]
    the supported, if any, interconnect standard between boards and shields i.e. Pro Micro, M.2, etc
    shields can also use a board's identifier in place of an interconnect if it's made for only one type of hardware
    this setting can also optionally be an array

components [optional]
    an array of components for a given board or shield. useful for splits where [left, right] can be used

hidden [optional]
    whether this piece of hardware should be hidden from directories

url [required]
    URL to the main page about the hardware
```

I've also added yamls for shields, which isn't standard to Zephyr. `shield` has been added to the `type` enum.


app/boards/arm/bluemicro840/bluemicro840_v1.yaml
```diff
@@ -13,3 +13,6 @@ supported:
   - ieee802154
   - pwm
   - watchdog
+manufacturer: BlueMicro
+compatible: Pro Micro
```
petejohanson, line 17: Using a "friendly name" for this feels really odd. I had expected something like:

```suggestion
compatible: pro_micro
```

or

```suggestion
compatible: promicro
```

app/boards/arm/nice_nano/nice_nano.yaml
```diff
@@ -13,3 +13,6 @@ supported:
   - ieee802154
   - pwm
   - watchdog
+manufacturer: Nice Keyboards
+compatible: Pro Micro
```
petejohanson, line 17: Is the idea that boards and shields each include a `compatible` value, and if they match, the can work together?

For a few edge case boards, we may need to make the `compatible` an array. I'm thinking of the Pancake v1, which supports both pro micro and feather footprints.

app/boards/shields/bfo9000/bfo9000.yaml
```diff
@@ -0,0 +1,7 @@
+identifier: bfo9000
+name: BFO-9000
+type: shield
+manufacturer: keeb.io
+compatible: Pro Micro
+split: true
```
petejohanson, line 6: Instead of a simple split boolean, to support future combinations, e.g. multiple peripherals for a split w/ a numpad that's separate, thoughts on a `components` or somesuch array, e.g.:

```suggestion
components: [left, right]
```

More work for every split board, but makes it easier to unambiguously setup which shields to build for if we automate setting up the template repos off this.

docs/src/components/hardware-list.js
```diff
@@ -0,0 +1,51 @@
...
+          </li>
+        ))}
+      </ul>
+      <span />
+      {hardware.compatible.map((h) => (
+        <React.Fragment key={h.name}>
+          <span />
+          <h3>{h.name}</h3>
+
+          <h4>Boards</h4>
```
petejohanson, line 21: Thoughts on how to handle split wired not working yet?

petejohanson (commented):
Thanks @Nicell for kicking this off! Looking good, a few questions/thoughts...

app/boards/arm/bluemicro840/bluemicro840_v1.yaml
Nicell: Totally agree with you on this. I originally had it as you have here, but then I ran into the issue of where do I get the friendly name from? I figured this was good enough for now, but it needs to be addressed.

app/boards/arm/nice_nano/nice_nano.yaml
Nicell: This is the idea, yeah. I think an array makes sense, although I wonder if I can support both, as having the array syntax in every yaml seems a bit annoying when only a few boards out there support multiple interconnects.

app/boards/shields/bfo9000/bfo9000.yaml
Nicell: Yeah, this is a good idea. Will probably need to be fleshed out more once dongles are a thing (building with/without a dongle in the mix). Maybe even an array of arrays that contain possible set ups?

docs/src/components/hardware-list.js
Nicell: Could check if there's a components list and split them apart? Although The formatting might get weird. Maybe instead we just make a big disclaimer and get working on wired support? 😅

app/boards/shields/bfo9000/bfo9000.yaml
innovaker: Or "nodes"?

An array is a good start.  We may need more flexible topologies in the future but we should probably figure out the firmware approach to that first.

innovaker:
Thanks @Nicell!  It's looking good!  I'll proof/check the fine detail of the yaml when I get more time.

- We may wish to consider a maintainer field (GitHub handle or repository).

- We'll need a module/repository URI in the coming future.

- We may wish to consider namespaces in the future as I can envisage different people submitting independent takes on a board/shield using the same identifier.  In the modular paradigm, the repository URI may suffice (although it doesn't work so well for community modules).

- Given we're reusing Zephyr's yaml, have you any thoughts on the scenario where someone wants to add an existing Zephyr board or shield into ZMK?  Should they simply duplicate the yml file from Zephyr?  What about updates/improvements to the original?

app/boards/shields/bfo9000/bfo9000.yaml
```diff
@@ -0,0 +1,7 @@
+identifier: bfo9000
+name: BFO-9000
+type: shield
+manufacturer: keeb.io
+compatible: pro_micro
+components: [left, right]
```
innovaker, line 6: Minor: this syntax doesn't read or merge as easily as the list syntax.
e.g.
```
components:
   - left
   - right
```
I appreciate it'll probably make no difference as we're unlikely to produce merge conflicts on this line, but consistency throughout is usually a good thing.  A single syntax for arrays also gives beginners one less thing to worry about or get wrong for toolchain or compatible.

app/boards/arm/bdn9/bdn9_rev2.yaml
```diff
@@ -1,5 +1,5 @@
-identifier: bdn9
-name: keeb.io BDN9
+identifier: bdn9_rev2
```
innovaker, line 1: It's likely some numbnut will submit a PR where the identifier doesn't match the filenames.  I suggest we pre-empt that with a GitHub workflow/job that protects against it.  Or at least create an issue to do so or we'll forget.

app/boards/arm/bdn9/bdn9_rev2.yaml
```diff
@@ -12,3 +12,6 @@ supported:
   - switches
   - underglow
   - per_key
+manufacturer: keeb.io
```
innovaker, line 15: If the url is mandatory then the manufacturer field seems superfluous (and arguably ambiguous), especially if it isn't displayed anywhere.

app/boards/arm/dz60rgb/dz60rgb_rev1.yaml
```diff
@@ -17,3 +17,6 @@ supported:
   - nvs
   - can
   - kscan
+manufacturer: dztech
+standalone: true
```
innovaker, line 21: The meaning of `standalone` didn't jump out at me which means it probably won't for others either.  It seems to be `true` when `compatible` is blank?  So, `shieldless` I guess?  But the type of `bdn9_rev2` is `mcu` not `keyboard`, which confuses my limited understanding.

app/boards/arm/dz60rgb/dz60rgb_rev1.yaml
```diff
@@ -1,5 +1,5 @@
-identifier: DZ60RGB_rev1
-name: DZ60RGBREV1
+identifier: dz60rgb_rev1
+name: DZ60RGB Rev1
 type: keyboard
```
innovaker, line 3: The `type` property here is defined here: https://docs.zephyrproject.org/latest/guides/test/twister.html#twister-script
`mcu` or `qemu` - for testing purposes.
I'd suggest we don't improvise on this field.

The directory structure indicates whether it's a board or shield.

app/boards/arm/bluemicro840/bluemicro840_v1.yaml
```diff
@@ -13,3 +13,6 @@ supported:
   - ieee802154
   - pwm
   - watchdog
+manufacturer: BlueMicro
+compatible: pro_micro
```
innovaker, line 17: I appreciate that having `compatible` as a string *or* array is prettier, it:
- doesn't immediately standout to users that it *can* be an array
- makes it slightly more complicated to process (most developers ignore the documentation, so the first array to appear may break stuff)

The list syntax avoids that complication.

app/boards/arm/dz60rgb/dz60rgb_rev1.yaml
Nicell: This type should be `mcu`, yes. My mistake from when I made this board long ago. I'm not so sure what the issue is with extending this to have `shield` on it. It makes this yaml much more consumable without needing path context.

innovaker: Technically we're breaking the inherited intention (schema) of that field by doing so.  A different, dedicated field which indicates whether it's a shield would be more appropriate.  None of the shields in Zephyr have such a field though - or the `type` field for that matter.

innovaker: I appreciate that none of the shields in Zephyr have a yaml file at all.

innovaker: The easy way to approach this is to have two globs.  One for boards, one for shields.

Nicell: We should consider whether every application that ever consumes these yaml files will have the full path context when working with them. I'm not sure we can guarantee that, so I think we should still have something indicating whether a piece of hardware is a shield.

We could remove the `type` entirely and check if it's undefined to tell if it's a shield, but I don't really get how that's breaking the twister script any less than using a type that doesn't exist for hardware that shouldn't even be run with twister.

app/boards/arm/dz60rgb/dz60rgb_rev1.yaml
Nicell: Yeah if there's no compatible you *could* call it standalone with the current set. Although there's a chance support is added for a board that has no compatible shield interconnect but isn't a standalone board. That would be a board used for handwires.

app/boards/arm/bdn9/bdn9_rev2.yaml
Nicell: It seems like a rather important detail to leave out for a hypothetical future database full of these hardware pieces. It offers a common property between hardware pieces created by the same entity.

app/boards/arm/dz60rgb/dz60rgb_rev1.yaml
innovaker: As discussed privately, a lot of the semantic overlap issues can be solved with our own `.zmk.yaml` files.  I'd suggest going in that directly with a bespoke (unwritten) data schema that meets our needs.

app/boards/arm/bdn9/bdn9_rev2.yaml
innovaker: If so, a KV lookup (manufacturer id to display name) would be more appropriate, less error prone and more extensible.  So if we're not displaying it yet, we might as well leave it out until later (I've already extracted a copy to sheets if you want to save it for said future efforts).

app/boards/shields/tidbit/tidbit.yaml
```diff
@@ -0,0 +1,7 @@
+identifier: tidbit
```
innovaker, line 1: This doesn't seem to account for the `19key` variant?

Technically variants can (should?) be extrapolated from the filesystem as they're `${SHIELD}-${VARIANT}.overlay`.

How do variants differ from "components" described elsewhere in this issue?  They follow the same file system naming conventions.  Should we treat them as one or differently in the metadata?

innovaker:
@Nicell, my ongoing analysis of the metadata is here: https://docs.google.com/spreadsheets/d/12L8ueXwdkhbAvoAKSipLkbrqlooo_wFvjo4-LYvuVlM/
I've added you as an editor, anyone else can comment on the sheet.

There were only a few minor issues.

I also checked the names of each keyboard against what's published by the maker, seller or community.  I tentatively suggest we match their individual naming styles/approaches as it simplifies shield reviews (the rule would be: compare it against its website or group buy page).

I should be able to flip the sheet into a code generator if needed without too much trouble so let me know if so.

docs/src/components/hardware-list.js
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+import hardware from "../data/hardware.json";
+
+const compatibleLookUp = {
+  pro_micro: "Pro Micro",
+  proton_c: "Proton-C",
+  makerdiary_m2: "makerdiary M.2",
```
innovaker, line 7: From @Nicell's original post:
> shields can also use a board's identifier in place of an interconnect if it's made for only one type of hardware
    this setting can also optionally be an array

I find the duality of `compatible` intriuging.

Interface specifications are either:
 a) defined from the outset (e.g. Adafruit Feather and nRF52840 M.2 Module)
 b) become "standard" based on the original board (e.g. [SparkFun] Pro Micro)

For (b), at what point does it switch from shields that fits a board (whose display name we presumably look up in the board's metadata), to shields that fits an interface "standard" defined in this lookup list.

Also, for the sake of argument, let's assume the Pro Micro ran ARM and thus ZMK.  Would the board and its namesake interface not compete for the same identifier: `pro_micro`.  We're lucky not to be in that situation now but it highlights the potential for conflicts in the duality.

It might be simpler to have something along the lines of:
```yaml
- compatibility
  - boards:
    -
  - interfaces:
    - 
```  
or just
```yaml
- boards:
  -
- interfaces:
  - 
```  

Thoughts anyone?

petejohanson: For the pro micro case, it seems like them being the same is a *good* thing, since you don't need to retroactively change anything, when someone creates a second board that "mimics" the existing one.

innovaker: I can roll with that.  So the code would need to treat interfaces and boards on a priority basis ...
i.e. (psuedo code, without guards) `name = interfaces[id] ? interfaces[id].name : boards[id].name;`

Are there any fringe cases that would screw that approach up?

(I'm using the word interface rather than compatible for this discussion because it's less ambiguous)

innovaker:
To help speed this along, I've knocked together this WIP branch (which I accidentally pushed to @Nicell's repository first, sorry about that!) which is generated from the sheet I previously linked:
https://github.com/innovaker/zmk/tree/docs/generate-hardware-page
Once the dicsussions reach their conclusions I'll update it accordingly, taking the tedium away to allow @Nicell to focus on the code.

app/boards/arm/nrfmicro/nrfmicro_11_flipped.yaml
```diff
@@ -1,5 +1,5 @@
 identifier: nrfmicro_11_flipped
-name: nrfmicro_11_flipped
+name: nRFMicro v1.1 Flipped
```
innovaker, line 2: The nRFMicro is a prime example of the potential for *families* of versions/revisions and variants, which I'm guessing we'll want to group together at some point.  If we don't account for how we group/link them, it may be error prone.

Thoughts on making them part of the `yaml` structure?
e.g. `nrfmicro.yml`:
```yaml
id: nrfmicro
name: nRFMicro
url: https://github.com/joric/nrfmicro
versions: 
  - id: nrfmicro_11
    name: nRFMicro v1.1
    variants:
    - id: flipped
  - id: nrfmicro_13
    name: nrfMicro v1.3
    version: 1.3
```
I'm just spitballing here.  I'm not even sure if I like the idea myself.

Nicell: I kind of like having the separate yamls for each version. What if we kept the individual one and had a `nrfmicro.yaml` that connects the three existing yamls together as versions. Or I guess the path could be utilized to figure that out too, but I'd rather not rely on the path.

petejohanson: The only downside here is complicating the simple version, always requiring two files...

Thoughts on single file, but using an `import` or `include` approach to make it effectively one file, bit allowing individual management?

Nicell: I do like the idea of importing/including. Makes sense for projects that are just called "board" but then have a revision later. In that case, the original can be imported with just changes done to id/anything else.

app/boards/arm/dz60rgb/dz60rgb_rev1.yaml
innovaker: I think I understand where you're coming from.  So "standalone" is mostly indicating that the board is a functioning keyboard/device out-of-the-box.

We need to ensure boards which aren't standalone and don't fit into named scheme are handled by the hardware page, i.e. miscellaneous/other for lack of a better term.

docs/src/components/hardware-list.js
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+import hardware from "../data/hardware.json";
+
+const compatibleLookUp = {
```
innovaker, line 4: Although `compatible` makes sense in the context of these `yml` files - as in "compatbility" or "compatible with" -  for discussions and documentation we should probably adopt more articulate terminology.

For instance, `interconnect` (as @Nicell used in his [original post](https://github.com/zmkfirmware/zmk/pull/653#issue-564559826)) or `interface` (as I used in another [thread](https://github.com/zmkfirmware/zmk/pull/653/commits/4aa3fa663716e63ba1a0dd3b9f4618afead96fa3#r574804767)) are two possible examples.  They're both similar to `pinout` but generalised to any form of shield connection.

Moreover, we should adopt terminology to differentiate between "established"/"named" schemes (e.g. the [sparkfun] `pro_micro`, `makerdiary_m2` or `adafruit_feather`) and any other interface.  Keeping in mind the points I made in https://github.com/zmkfirmware/zmk/pull/653/commits/4aa3fa663716e63ba1a0dd3b9f4618afead96fa3#r574804767.

The code should also reflect this terminology.

Thoughts?

app/boards/arm/nrfmicro/nrfmicro_11_flipped.yaml
petejohanson: Ok, so digging into upstream Zephyr here a bit, they definitely use a "YAML file per version" for boards, at least. Take a look at `zephyr/boards/arm/nucleo_l452re` for example, that contains both `nucleo_l452re.yaml` and `nucleo_l452re_p.yaml` with the details of each. You can see other examples in `stm32_min_dev`, `efr32_radio`, etc.

I think we should stick w/ that convention for boards. Each revision of a board might include significant changes, different chip onboard, etc. so having them be clearly defined per board version makes sense to me.

Now, since Zephyr doesn't include YAML files for the shields, we don't have much to use as a reference there, but I'd say useful to consider the current and future use cases.

* Non-split, single shield with compatibility with some module/board footprint.
* Keyboard with 2+ shields
* Keyboard with 2+ shields to be built in "dongle mode"
* Multi-shield keyboards where each component includes different additional hardware (e.g. left side only has encoder, right side only has OLED)
* Multiple revisions of each of the above.

To be honest... It really makes me wonder if we should be going the other way with this, and do a file per each uniquely named shield. Accept some duplication, but be *really* explicit.

The main detractor to that approach seems to be: "How do I know which pieces usually come together and I want to build as "one thing", if we go that route.

I wonder, if we go that route, w/ explicit per-shield file (e.g.. `lily58_left.yaml`/`lily58_right.yaml`, `lily58_pro_left.yaml`/`lily58_pro_right.yaml`), we then have some top level aggregation YAML file to create the connections, e.g. `lily58.yaml` or maybe `configs.yaml`.

That would have something like we talked about:

```yaml
configs:
  - identifier: lily58
    shields:
      - lily58_pro_left
      - lily58_pro_right
  - identifier: lily58_pro
    shields:
      - lily58_pro_left
      - lily58_pro_right
```

I believe the "per half" explicit docs are needed, in order to allow for each half including different hardware (encoders, OLEDs, etc) and our metadata should ideally capture that for future uses.

The `configs.yaml` file should make it easier to have tools like @joelspadin "user config tool" automate setup of a "full configuration"  from that as a starting point.

Non-split shields I think could *skip* the `config.yml` file, and the config tool could assume each shield should be built on their own.

@Nicell @innovaker Thoughts?

joelspadin: > I wonder, if we go that route, w/ explicit per-shield file (e.g.. lily58_left.yaml/lily58_right.yaml, lily58_pro_left.yaml/lily58_pro_right.yaml), we then have some top level aggregation YAML file to create the connections, e.g. lily58.yaml or maybe configs.yaml.

+1 to this idea. I recommend going with `configs.yaml`. Having a consistently named file would make it much easier for a tool to identify the correct file. If a folder contains `lily58.yaml`, `lily58_pro.yaml`, `lily58_left.yaml`, `lily58_pro_left.yaml`, etc... then the only way I can think to pick out that `lily58.yaml` and `lily58_pro.yaml` are the top-level files would be to read each file.

> Non-split shields I think could skip the config.yml file, and the config tool could assume each shield should be built on their own.

This should work. While having a `configs.yaml` file *everywhere* would make it easier for people writing tools, it would also be pretty easy to check for the absence of such a file and then scan the directory for all .yaml files in that case.

petejohanson: > 
> > Non-split shields I think could skip the config.yml file, and the config tool could assume each shield should be built on their own.
> 
> This should work. While having a `configs.yaml` file _everywhere_ would make it easier for people writing tools, it would also be pretty easy to check for the absence of such a file and then scan the directory for all .yaml files in that case.

Yeah, we'd probably need some rudimentary fallback, but I don't think it needs to be too fancy. We should be pushing for the `configs.yaml` file as the standard.

Nicell: I'm curious how this `configs.yaml` file would look for dongle mode. I know we don't really have a dongle mode yet to reference, but would we have a different shield variant on each shield for dongle mode?

As such:
```yaml
configs:
  - identifier: lily58
    shields:
      - lily58_pro_left
      - lily58_pro_right
  - identifier: lily58_pro
    shields:
      - lily58_pro_left
      - lily58_pro_right
  - identifier: lily58_dongle
    shields:
      - some_dongle
      - lily58_pro_left_dongle
      - lily58_pro_right_dongle
  - identifier: lily58_pro_dongle
    shields:
      - some_dongle
      - lily58_pro_left_dongle
      - lily58_pro_right_dongle
```

And then how do we handle the multiple possible dongles? Obviously, we could have more than one board that could work as a dongle.

Outside of this dongle question, this makes sense to me.

petejohanson: I envisioned dongle mode being not part of this. You'd build all the shields for a version with overriding role to be peripheral, then build a standard dongle board + the keymap + # of dongles. I don't think the left/right sides change other than the config override.

So, long story short... I don't think it actually needs to be in the configs.

joelspadin: Maybe obvious, but if we go with the configs.yaml file, we should probably give each config a `name` property to select the display name for splits. Tools could grab the name from the first listed board/shield (and probably should continue to use that as a fallback in case a name isn't given in configs.yaml), but that might pick out a name that has "left" or "right" in it.

docs/src/components/hardware-list.js
innovaker: Anyone?  I'll elaborate a bit more if it helps ...

I'm not keen on `compatible` because it's *primarily* used as an adjective.

I personally prefer `interface` because:
a) it's a noun, which makes it easy to use in both code and sentences.
b) its dictionary definition fits the concept, making it fairly self-explanatory.
c) it has a natural/easy plural.
**but**, I suspect it *may* inadvertantly have namespace issues with some programming languages that use it as a keyword 😞

`interconnect` is a verb or adjective, so using it as a noun feels weird.

app/boards/arm/nrfmicro/nrfmicro_11_flipped.yaml
innovaker: There's been some discussion related to this thread here:
https://discord.com/channels/719497620560543766/719544230497878116/821115603720077334

docs/src/components/hardware-list.js
innovaker: I just noticed @petejohanson also used the word `footprint` in another comment.  Effectively a synonym of `pinout`.

app/boards/arm/nrfmicro/nrfmicro_11_flipped.yaml
innovaker: Regarding Zephyr's conventions, I'm happy breaking with them for this issue.  As far as I can tell, the primary (only?) purpose of Zephyr's `${BOARD}.yaml` files is to feed the Twister test framework and there's hints that it's only a transitory solution:
- https://docs.zephyrproject.org/latest/guides/test/twister.html?highlight=twister#board-configuration
- https://docs.zephyrproject.org/latest/guides/dts/design.html?highlight=yaml#example-remaining-work

Whereas we're developing the metadata for these (immediate) purposes:
- web
  - hardware page
  - setup scripts and/or its replacement
- CI builds
- and maybe everything else here too: https://discord.com/channels/719497620560543766/719544230497878116/821417305081905163

So quite different purposes.  That said, there's always merit in cherry picking the best aspects of Zephyr's approach where beneficial.

Nicell: > The examples given above only seem to explore one type of grouping: left and right (halves)

I'd disagree with this. You can have however many parts in each shield array that you'd like. We just currently only have left and right halve examples.

Family, version, and variant seem like good classifiers. So a family would have version and versions would have variants? I think parts can probably be put in arrays as we have above.

docs/src/components/hardware-list.js
petejohanson: I'd be down w/ `interface`, `pinout` or `footprint`.

app/boards/arm/nrfmicro/nrfmicro_11_flipped.yaml
petejohanson: Are you talking about for use in the aggregate `configs.yaml` file? The individual files? Something else?

petejohanson: If we're talking about the aggregate file, weld need to be clear on how we take that discrete taxonomy,.and still have an easy mapping to specific shield me for the build. If the convention is `family_rev_part_layout` with layout optional, I could see this working well.

petejohanson: As I look at this more, and play w/ a JSON Schema to allow us to validate these files, I'm less and less convinced about the need to have the `parts` of each keyboard have their own file. It ends up with even *more* duplication, which I'm not sure is really necessary.

As I ponder the "each half has different hardware" part, I am seeing ways we can still do this without needing anything too crazy.

In particular, Looking at the JSON schema for what I'm playing with, you can easily end up w/ something like:

```
   "shield": {
      "title": "Shield",
      "type": "object",
      "additionalProperties": false,
      "required": [ "id", "name", "url", "type", "interconnects" ],
      "properties": {
        "id": { "$ref": "#/$defs/id" },
        "name": { "type": "string" },
        "url": { "type": "string", "format": "uri" },
        "manufacturer": { "type": "string" },
	"version": { "type": "string" },
        "type": {
          "type": "string",
	  "const": "shield"
        },
	"parts": { "$ref": "#/$defs/keyboard_parts" },
	"features": { "$ref": "#/$defs/features" },
        "interconnects": { "$ref": "#/$defs/interconnects" }
      }
```

Where you have a couple defs like:

```
    "keyboard_parts": {
      "type": "array",
      "items": {
        "oneOf": [
          { "type": "string" },
          { "$ref": "#/$defs/part_details" }
        ]
      }
    },
    "features": {
      "type": "array",
      "items": {
        "type": "string",
        "enum": [ "display", "encoder" ]
      }
    },
```

Which lets you have a basic set of features listed at the shield level, if those features apply to all parts, or you can specify the full details of the features present on each part of the shield/keyboard.

If we do this, we can get away with skipping the extra `configs.yml` file, and simplify this as "search for all *.zmk.yml files, inspect the content" to get each "build target" (which might include building multiple parts).

I'll post a full gist once I'm ready w/ sample files + schema to validate them.

innovaker: @petejohanson posted a gist here: https://discord.com/channels/719497620560543766/719544230497878116/825361233074454529

innovaker: Thanks @petejohanson!

---

As we work through this, let's keep in mind that each purpose I listed has quite different needs.  Zephyr only cares about a unique id, whereas everything else we're building cares about the attributes and relationships, the more meaningful the better.  @petejohanson's exploration of a schema - or more broadly speaking the underlying data model - will help us define these aspects.

---

I'd like to highlight that `version` and `variant` (and `part` to some extent) are sometimes defined by the maker.   Each of them are optional and independent, and they don't always follow the hierarchy/relationships we might expect.

---

> Are you talking about for use in the aggregate configs.yaml file? The individual files? Something else?

I was talking conceptually - data modelling - rather than files.

---

I think listing the features is a good idea.  I'd been considering it a future effort, but I can see the logic in defining them as part of the initial design.

---

`part` currently conveys a few meanings such as:
- which bit?
- what part of the body (hand) uses it?
  - and coupled with that: probable physical position relative to the user and maybe the other parts too.
- what role?
  - This is debatable because it's still a largely undeveloped grey area.
  - It's an important concept but I think we should fully explore other topologies before defining it.

So perhaps we want to consider capturing these meanings individually?  They may not be critical now, but they probably will be in the future.

Nicell: I'm putting some parts of the example schema @petejohanson has made in here to offer suggestions/thoughts.

> ```
>     "keyboard_parts": {
>       "type": "array",
>       "items": {
>         "oneOf": [
>           { "type": "string" },
>           { "$ref": "#/$defs/part_details" }
>         ]
>       }
>     },
> ```

I think we need a better property name than this since, although we are a keyboard firmware, there's the opportunity to have lots more input devices added in the future. Parts is too generic. As is components or nodes. Sub-assemblies makes more sense to me, but it's a bit long-winded and still doesn't feel right.

Along with this, what does "type" mean? I'm assuming it's the `id` associated with this part of the network, but we're still lacking a lot of other information about what this part actually is/does as @innovaker pointed out. Although is it truly important for us to deliberately state what this part does? Maybe it could be optional descriptors.

> ```
>     "variant": {
>       "oneOf": [
> 	{ "type": "string" },
> 	{
>           "type": "object",
> 	  "required": [ "id", "features" ],
>           "properties": {
>             "id": { "$ref": "#/$defs/id" },
>             "features": { "$ref": "#/$defs/features" }
>           }
> 	}
>       ]
>     },
> ```

I think it makes sense to have the `id`, which is associated with the variant overlay, but I'm not sure how features would work here. Are they adding on to existing features or overwriting the "parent" features. This seems like a tricky piece, as variants could have the exact same features, added features, or a whole different set of features. How do we make this easy to manage for people writing a variant?

> ```
>     "features": {
>       "type": "array",
>       "items": {
>         "type": "string",
>         "enum": [ "display", "encoder" ]
>       }
>     },
> ```

I'm glad this section exists, but I'm wondering how we keep this up to date and allow for new features to be added. We definitely don't want this to be opened up because then it becomes a nightmare for filtering and any data processing of features. We need a simple way to define and add features without having to adjust the schema directly everytime IMO.

> ```
>     "interconnects": {
>       "type": "array",
>       "minItems": 1,
>       "items": {
>         "type": "string",
>         "minLength": 1
>       }
>     },
> ```

After thinking about it for a while, I really do think a simple interconnect array is sufficient for our needs of `shield <-> board` connection because of how Zephyr works. I'm wondering though how we'll do the translation from `id` to a readable name as well as how we'll moderate the `id`s of generic interconnects that aren't specific to boards i.e. `pro_micro` (generic) vs. `proton_c` (board specific).

Outside of these parts, the board/shield properties look relatively straightforward outside of the larger discussion about variants and versions. Thanks for putting this initial schema together @petejohanson !

petejohanson:
I've got a branch up at https://github.com/petejohanson/zmk/tree/boards-shields/hardware-metadata-schema where I'm iterating on this stuff.

petejohanson:
I co-opted this from @Nicell

Closing

Issue #288, opened 2020-10-20
Disable "Tapping force Hold": make tap-hold on &mt keys output a held tap action.


krikun98:
This feature is vital on 40% boards, where essentially all modifiers are on &mt keys. Currently, ZMK outputs a long hold action, and with, for example, my backspace on (&mt RSFT BKSP) there's no way to hold the key - it'll just be a long shift.

okke-formsma:
@krikun98 I'm not sure if you are looking for retro tapping or tapping force hold.

From QMK docs: 
https://beta.docs.qmk.fm/using-qmk/software-features/tap_hold#retro-tapping
Retro Tapping: Holding and releasing a dual function key without pressing another key will result in nothing happening. With retro tapping enabled, releasing the key without pressing another will send the original keycode even if it is outside the tapping term.

https://beta.docs.qmk.fm/using-qmk/software-features/tap_hold#tapping-force-hold
Tapping force Hold: When the user holds a key after tapping it, the tapping function is repeated by default, rather than activating the hold function. This allows keeping the ability to auto-repeat the tapping function of a dual-role key. TAPPING_FORCE_HOLD removes that ability to let the user activate the hold function instead, in the case of holding the dual-role key after having tapped it.

krikun98:
"Tapping force Hold" is the default mode for ZMK. I didn't know "Retro Tapping" was a pre-existing term, I thought it was something you came up with. I want a possibility to disable "Tapping force Hold", i.e. to mimic the default behavior.
The description for "Retro Tapping" doesn't sound like what I need. I'll rename the issue.

joric:
Need this asap for jorne backspace. How to enable autorepeat maybe there's a quick patch I could do? Need that urgently.
My keymap: https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/jorne/jorne.keymap

okke-formsma:
Sorry, no work on this has been done yet. I can't think of a quick fix right now other than disabling the hold tap

okke-formsma:
@joric it's your lucky day, if you care to test my branch?

joric:
@okke-formsma I didn't get autorepeat with your code. Am I supposed to do a specific tap dance? I have a key `&mt RSHFT BSPC` and I want BSPC to autorepeat on hold.

andreiboyanov:
[...]

PR #655, opened 2021-01-31
behaviors(hold-tap): Implement quick_tap_ms (TAPPING_FORCE_HOLD)

okke-formsma:
Tap-and-hold a hold-tap to hold the tap behavior so it can repeat.
After a tap, if the same key is pressed within `quick_tap_ms`, the
tap behavior is always picked.

This is useful for things like `&ht LSHFT BACKSPACE` where holding
the backspace is required.

Implements #288

### DIFF
--- a/app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
@@ -13,6 +13,9 @@ properties:
     required: true
   tapping_term_ms:
     type: int
+  quick_tap_ms:
+    type: int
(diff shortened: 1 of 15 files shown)


petejohanson (approved):
This is awesome!

Working like a champ for my layer taps on my Ferris, and tested by a few users, includes great tests as usual.

Thanks @okke-formsma

petejohanson:
@okke-formsma Argh! The refactor for the transform stuff caused conflicts. Can you please rebase, then we'll merge? Thanks!

petejohanson (approved):

PR #656, opened 2021-02-01
Update new keyboard shield testing documentation.

aaron-nunley:
Clarifying the need for `west flash` in a dockerized environment and providing a high-level overview of testing a board with UF2 support.  As a first-time developer in this environment, it wasn't entirely clear to me that `west flash` shouldn't work from within a Docker container and it seemed like a good opportunity to clarify the documentation.

### DIFF
--- a/docs/docs/development/build-flash.md
+++ b/docs/docs/development/build-flash.md
@@ -118,11 +118,17 @@ Now start VSCode and rebuild the container after being prompted. You should be a
 
 ## Flashing
 
-Once built, the previously supplied parameters will be remembered so you can run the following to flash your
-board with it in bootloader mode:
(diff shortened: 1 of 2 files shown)


Nicell (commented):
Thanks for clarifying this section @aaron-nunley. I'm thinking that the building and flashing page also needs to be updated to make the instructions on `.uf2` as clear as this. Could you update that as well with similar verbiage? 

Building Flasing section to update: https://zmkfirmware.dev/docs/development/build-flash/#flashing

Nicell (approved):
LGTM. Thanks!

petejohanson (approved):
Thanks!

PR #658, opened 2021-02-01
refactor(boards): Remove split right keyboard names

Nicell:
This PR's purpose is to reduce people's chances of accidentally pairing with the wrong half of the board, as they shouldn't be pairing with the peripheral half.


app/boards/shields/sofle/Kconfig.defconfig
```diff
@@ -11,13 +11,6 @@ config ZMK_SPLIT_BLE_ROLE_CENTRAL
 	
 endif
 
-if SHIELD_SOFLE_RIGHT
-
-config ZMK_KEYBOARD_NAME
-	default "Sofle Right"
-
-endif
-
```
mbarrben, line 20: Forgot removing " Left" from Sofle left shield name? 🙂

fsargent:
What’s the purpose of this change?

petejohanson (approved):
Finally had a chance to test this here, and works fine, and I fully agree this will help reduce support issues.

Issue #659, opened 2021-02-02
Can't select "makerdiary nRF52840 M.2" when running setup.sh


y4my4m123:
You can't select the 4th board option (makerdiary nRF52840 M.2) when running setup.sh.
The board selection part of the script looks like this at the moment:

    prompt="Pick an MCU board:"
    options=("nice!nano" "QMK Proton-C" "BlueMicro840 (v1)" "makerdiary nRF52840 M.2")

    echo "$title"
    echo ""
    echo "MCU Board Selection:"
    PS3="$prompt "
    select opt in "${options[@]}" "Quit"; do

        case "$REPLY" in

        1 ) board="nice_nano"; break;;
        2 ) board="proton_c"; break;;
        3 ) board="bluemicro840_v1"; break;;
        3 ) board="nrf52840_m2"; break;;

        $(( ${#options[@]}+1 )) ) echo "Goodbye!"; exit 1;;
        *) echo "Invalid option. Try another one."; continue;;

        esac
    done

I think its simply fixed by changing the last 3 to a 4.

PR #660, opened 2021-02-02
Change the board selection part of setup.sh, so that the 4th option ("makerdiary nRF52840 M.2") can be selected

y4my4m123:
This should fix Issue [#659](https://github.com/zmkfirmware/zmk/issues/659).


petejohanson (approved):
Good catch, thanks!

PR #661, opened 2021-02-03
Feat combo layers

mcrosson:
Add layer filters to combos. This allows a combo to be active for one or more layers as well as global in scope.

The default is for a combo to be globally applicable.


mcrosson:
@okke-formsma Would you be willing to review this?

app/src/combo.c
```diff
@@ -105,16 +107,37 @@ static int initialize_combo(struct combo_cfg *new_combo) {
 }
 
 static int setup_candidates_for_first_keypress(int32_t position, int64_t timestamp) {
+    // Due to layer filtering we need to track the candidates[index] stuff by hand
+    int number_of_combo_candidates = -1; // First increment moves this to 0 as expected
+    uint8_t highest_active_layer = zmk_keymap_highest_layer_active();
     for (int i = 0; i < CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY; i++) {
+        number_of_combo_candidates++; // Increment on each pass of the loop
```
okke-formsma, line 114: I think we can clean this up a bit; 

initialize number_of_combo_candidates to 0, then remove this line and only increment it in the `if (combo_layer_valid)` clause.

app/src/combo.c
```diff
@@ -105,16 +107,37 @@ static int initialize_combo(struct combo_cfg *new_combo) {
...
+        if (combo->layers[0] == -1) {
+            combo_layer_valid = true;
+        } else {
+            for (int j = 0; j < sizeof(combo->layers) / sizeof(combo->layers[0]); j++) {
+                if (combo->layers[j] == highest_active_layer) {
+                    combo_layer_valid = true;
+                    break;
+                }
+            }
+        }
```
okke-formsma, line 131: I think we should extract lines 121-131 into a separate function like `combo_active_on_layer(combo, layer)`

docs/docs/features/combos.md
```diff
@@ -29,6 +30,7 @@ Combos configured in your `.keymap` file, but are separate from the `keymap` nod
 - `key-positions` is an array of key positions. See the info section below about how to figure out the positions on your board.
 - `bindings` is the behavior that is activated when the behavior is pressed.
 - (advanced) you can specify `slow-release` if you want the combo binding to be released when all key-positions are released. The default is to release the combo as soon as any of the keys in the combo is released.
+- (advanced) you can specity a list of layers (`layers = <0 1...>`) to limit a combo to specific layers. this is an _optional_ parameter and defaults to `-1` which is global scope.
```
okke-formsma, line 33: I think this should not be in the 'advanced' section but in the default.

okke-formsma (changes requested):
In general, looks good. I see room for a few improvements.

Also please add at least one test to cover this functionality.

app/src/combo.c
mcrosson: Re-worked the incrementing logic to simplify/cleanup

app/src/combo.c
mcrosson: Extracted this as a method per suggestion

docs/docs/features/combos.md
mcrosson: Move up in docs to be the last item before advanced and removed advanced modifier from text

mcrosson:
> Also please add at least one test to cover this functionality.

Test case added.


This should be good for re-review

app/dts/bindings/zmk,combos.yaml
```diff
@@ -20,3 +20,5 @@ child-binding:
       default: 50
     slow-release:
       type: boolean
+    layers:
+      type: array
```
okke-formsma, line 24: I think setting a default of -1 would be nice.

app/src/combo.c
```diff
@@ -104,17 +106,40 @@ static int initialize_combo(struct combo_cfg *new_combo) {
     return 0;
 }
 
+static bool combo_active_on_layer(struct combo_cfg *combo, uint8_t layer) {
```
okke-formsma, line 109: maybe this is simpler?
```
static bool combo_active_on_layer(struct combo_cfg *combo, uint8_t layer) {
    if (combo->layers[0] == -1) {
        // -1 in the first layer position is global layer scope
        return true;
    }
    for (int j = 0; j < CONFIG_ZMK_COMBO_MAX_LAYER_FILTERS_PER_COMBO; j++) {
        if (combo->layers[j] == layer) {
            return true;
        }
    }
    return false;
}
```

app/src/combo.c
```diff
@@ -104,17 +106,40 @@ static int initialize_combo(struct combo_cfg *new_combo) {
...
+                combo_layer_valid = true;
+                break;
+            }
+        }
+    }
+    return combo_layer_valid;
+}
+
 static int setup_candidates_for_first_keypress(int32_t position, int64_t timestamp) {
+    // Due to layer filtering we need to track the candidates[index] stuff by hand
```
okke-formsma, line 127: I think we can drop the comment.

app/dts/bindings/zmk,combos.yaml
okke-formsma: Hmmm the default is managed by `.layers = DT_PROP_OR(n, layers, {-1}),`. That works for me.

app/src/combo.c
mcrosson: Nice catch, this is far simpler

app/src/combo.c
mcrosson: Agreed

okke-formsma:
I've found a bug

```
#include <dt-bindings/zmk/keys.h>
#include <behaviors.dtsi>
#include <dt-bindings/zmk/kscan-mock.h>

/* it is useful to set timeout to a large value when attaching a debugger. */
#define TIMEOUT (60*60*1000)

/ {
	combos {
		compatible = "zmk,combos";
		combo_one {
			timeout-ms = <TIMEOUT>;
			key-positions = <0 1>;
			bindings = <&kp X>;
			layers = <1>;
		};
	};

	keymap {
		compatible = "zmk,keymap";
		label ="Default keymap";

		default_layer {
			bindings = <
				&kp A &kp B
				&kp C &tog 1
			>;
		};
	};
};

&kscan {
	events = <
		/* Combo One */
		ZMK_MOCK_PRESS(0,0,10) 
		ZMK_MOCK_PRESS(0,1,10) 
		ZMK_MOCK_RELEASE(0,0,10) 
		ZMK_MOCK_RELEASE(0,1,10)
	>;
};
```

This presses the combo even though we're still on layer 0. Thishappens because the default value for the layers that are not specified is 0:

image

The way to fix this is to store the number of layers with the layers array. See `key_position_len` for an example of how to do this.

mcrosson:
@okke-formsma Pushed a fix for the bug you found as well as other cleanup previously mentioned

app/src/combo.c
```diff
@@ -451,6 +472,8 @@ ZMK_SUBSCRIPTION(combo, zmk_position_state_changed);
         .behavior = KEY_BINDING_TO_STRUCT(0, n),                                                   \
         .virtual_key_position = ZMK_KEYMAP_LEN + __COUNTER__,                                      \
         .slow_release = DT_PROP(n, slow_release),                                                  \
+        .layers = DT_PROP_OR(n, layers, {-1}),                                                     \
```
okke-formsma, line 475: let's use DT_PROP() here without the default

okke-formsma (approved):
Thanks!

app/src/combo.c
```diff
@@ -33,6 +34,8 @@ struct combo_cfg {
     // the virtual key position is a key position outside the range used by the keyboard.
     // it is necessary so hold-taps can uniquely identify a behavior.
     int32_t virtual_key_position;
+    int8_t layers[CONFIG_ZMK_COMBO_MAX_LAYER_FILTERS_PER_COMBO];
+    int32_t layers_len;
```
petejohanson, line 38: Why use a fixed length array here, if we're also going to need to store a length parameter? We should be able to do something like we do in `app/drivers/kscan/kscan_gpio_direct.c` and make this:

```suggestion
    int32_t layers_len;
    int8_t layers[];
```

If we go this route, we can skip the fixed limit in the Kconfig you added.

app/Kconfig
```diff
@@ -301,6 +301,10 @@ config ZMK_COMBO_MAX_KEYS_PER_COMBO
 	int "Maximum number of keys per combo"
 	default 4
 
+config ZMK_COMBO_MAX_LAYER_FILTERS_PER_COMBO
```
petejohanson, line 304: See my other comment, I think we can skip this.

petejohanson (changes requested):
Just one minor comment. Otherwise looks good.

docs/docs/features/combos.md
```diff
@@ -49,3 +51,4 @@ There are three global combo parameters which are set through KConfig. You can s
 - `CONFIG_ZMK_COMBO_MAX_PRESSED_COMBOS` is the number of combos that can be active at the same time. Default 4.
 - `CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY` is the maximum number of combos that can be active on a key position. Defaults to 5. (So you can have 5 separate combos that use position `3` for example)
 - `CONFIG_ZMK_COMBO_MAX_KEYS_PER_COMBO` is the maximum number of keys that need to be pressed to activate a combo. Default 4. If you want a combo that triggers when pressing 5 keys, you'd set this to 5 for example.
+- `ZMK_COMBO_MAX_LAYER_FILTERS_PER_COMBO` is the maximum number of layer filters that can be specified. Default 4. If you want use layer filters with a combo that spans more than 4 layers, you'll need to change this value.
```
petejohanson, line 54: This entry can be removed now that you 'nixed the Kconfig setting.

petejohanson (approved):

PR #664, opened 2021-02-04
fix(core): Enable sensor by default

Nicell:


mcrosson:
Are there any spots in the code specifically looking for CONFIG_SENSOR?

Nicell:
> Are there any spots in the code specifically looking for CONFIG_SENSOR?

No, not in our codebase. Looking through Zephyr didn't turn up anything obvious.

petejohanson (approved):

PR #667, opened 2021-02-06
Retro tapping

okke-formsma:
Implements #339 : Retro-tapping.

The first commits don't change anything in the logic or execution of hold-taps, they're pure refactors to make the final change sweet & short.


app/src/behaviors/behavior_hold_tap.c
```diff
@@ -280,12 +281,85 @@ static inline char *flavor_str(enum flavor flavor) {
...
+    struct zmk_behavior_binding_event event = {
+        .position = hold_tap->position,
+        .timestamp = hold_tap->timestamp,
+    };
+
+    struct zmk_behavior_binding binding;
+    if (hold_tap->status & STATUS_HOLD) {
+        binding.behavior_dev = hold_tap->config->hold_behavior_dev;
+        binding.param1 = hold_tap->param_hold;
+        binding.param2 = 0;
```
petejohanson, line 331: The assignments of `0` here shouldn't be necessary, that should be the default value in the struct.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -280,12 +281,85 @@ static inline char *flavor_str(enum flavor flavor) {
...
+    } else {
+        binding.behavior_dev = hold_tap->config->tap_behavior_dev;
+        binding.param1 = hold_tap->param_tap;
+        binding.param2 = 0;
```
petejohanson, line 335: Ditto.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -280,12 +281,85 @@ static inline char *flavor_str(enum flavor flavor) {
...
+    struct zmk_behavior_binding_event sub_behavior_data = {
+        .position = hold_tap->position,
+        .timestamp = hold_tap->timestamp,
+    };
+
+    struct zmk_behavior_binding sub_behavior_binding;
+    if (hold_tap->status & STATUS_HOLD) {
+        sub_behavior_binding.behavior_dev = hold_tap->config->hold_behavior_dev;
+        sub_behavior_binding.param1 = hold_tap->param_hold;
+        sub_behavior_binding.param2 = 0;
```
petejohanson, line 351: Ditto

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -280,12 +281,85 @@ static inline char *flavor_str(enum flavor flavor) {
...
+    } else {
+        sub_behavior_binding.behavior_dev = hold_tap->config->tap_behavior_dev;
+        sub_behavior_binding.param1 = hold_tap->param_tap;
+        sub_behavior_binding.param2 = 0;
```
petejohanson, line 355: Ditto.

petejohanson (changes requested):
A few thoughts. Thanks!

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -280,12 +281,85 @@ static inline char *flavor_str(enum flavor flavor) {
...
+    case STATUS_HOLD_INTERRUPT:
+        return "hold-interrupt";
+    case STATUS_TAP:
+        return "tap";
+    default:
+        return "UNKNOWN STATUS";
+    }
+}
+
+static inline char *decision_moment_str(enum decision_moment decision_moment) {
```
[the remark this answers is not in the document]
joelspadin: Since you are returning constant strings, you *should* return a `const*`. Otherwise you return a pointer to a buffer that someone can try to modify even though it's really const, and then you run into a wonderful thing called undefined behavior.

app/src/behaviors/behavior_hold_tap.c
joelspadin: As it is currently declared, `binding` is not initialized. It has whatever value happens to be on the stack there. Either this `= 0` needs to stay, or you should change line 327 to something like `binding = {0};` (My preference would be for the latter, because then any new members you add to the struct get initialized too without changing the code.)

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -36,12 +36,30 @@ enum flavor {
...
 };
 
+enum status {
+    STATUS_UNDECIDED = 0,
+    STATUS_TAP = 1,
+    STATUS_HOLD_INTERRUPT = 2,
+    STATUS_HOLD_TIMER = 4,
+};
+
+#define STATUS_HOLD (STATUS_HOLD_INTERRUPT | STATUS_HOLD_TIMER)
```
joelspadin, line 46: Since this appears to only be used as a bit mask and isn't a valid value, maybe rename to `STATUS_HOLD_MASK`?

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -36,12 +36,30 @@ enum flavor {
     ZMK_BHV_HOLD_TAP_FLAVOR_TAP_PREFERRED = 2,
 };
 
+enum status {
```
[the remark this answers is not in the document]
joelspadin: Is it even possible for multiple flags to be active at the same time? In my skimming the code I only saw direct assignments to single values.

If the only reason this is using bit flags is for the `if (hold_tap->status & STATUS_HOLD)` checks, I suggest changing this back to a regular enum with auto assigned values and adding:

```c
inline boolean status_is_hold(enum status status) {
  return status == STATUS_HOLD_INTERRUPT || status == STATUS_HOLD_TIMER;
}

...
if (status_is_hold(hold_tap->status)) {
```

okke-formsma:
@petejohanson @joelspadin thanks for your comments, I've taken them all aboard.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -200,34 +215,23 @@ static struct active_hold_tap *store_hold_tap(uint32_t position, uint32_t param_
...
-        hold_tap->is_hold = 1;
-        hold_tap->is_decided = true;
-        break;
+        hold_tap->status = STATUS_HOLD_TIMER;
+        return;
     case HT_QUICK_TAP:
-        hold_tap->is_hold = 0;
-        hold_tap->is_decided = true;
-        break;
+        hold_tap->status = STATUS_TAP;
```
petejohanson, line 234: Minor nitpick, can you add a `return;` to this case, in case `default` is changed later.

petejohanson (changes requested):
One minor comment. Thanks!

okke-formsma:
I've made `no-hold-flash` always enabled for `retro-tap`, not separately configurable. I think this is what 99% of the users want. If any user really wants it configurable I'm happy to change it back, but let's start with fewer config options.

okke-formsma:
@petejohanson I think we're close to the finish with this one

petejohanson (approved):
Thanks!

PR #668, opened 2021-02-06
style(hold-tap)!: write hold-tap params with dashes

okke-formsma:
tapping_term_ms -> tapping-term-ms
quick_tap_ms -> quick-tap-ms

Unfortunately I have not found a way to make this backwards compatible, so I don't know if you want to make this change. Thoughts?

### DIFF
--- a/app/boards/shields/qaz/qaz.keymap
+++ b/app/boards/shields/qaz/qaz.keymap
@@ -18,7 +18,7 @@
             compatible = "zmk,behavior-hold-tap";
             label = "homerow mods";
             #binding-cells = <2>;
-            tapping_term_ms = <225>;
+            tapping-term-ms = <225>;
             flavor = "tap-preferred";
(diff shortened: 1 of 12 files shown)


petejohanson:
@okke-formsma I prefer the hyphen version as well. Can we keep the underscore version, mark both optional, and add a build assert if neither are specified?

Then we can add to a SOTF the changes needed, and give folks time to migrate?

okke-formsma:
I haven't found a way to make both versions work at the same time. Just specifying both doesn't work.

petejohanson:
> I haven't found a way to make both versions work at the same time. Just specifying both doesn't work.

Doesn't work at what level? DT processing? The macros to extract the values in the code?

okke-formsma:
Gheh, I must have done something wrong before because now it suddenly works like a charm with fallback. I'll update the PR.

petejohanson (approved):

PR #670, opened 2021-02-07
Add event trace_id to fix bug in sticky keys and simplify hold-taps

okke-formsma:
The 'event trace_id' is a unique identifier that identifies the keypress-keyrelease combination. It can be used to identify which position-state-changed events, behavior calls, and keycode-state-changed events were the result of which keypress (and which release).

The main benefactor of this functionality is sticky keys, where just a position & layer ID combination is not enough to correctly identify a behavior invocation.

Additionally, this is useful for hold-taps, where keycode events generated by the behavior itself must be ignored. So far this was solved with a clunky workaround.


app/src/events/position_state_changed.c
```diff
@@ -7,4 +7,15 @@
 #include <kernel.h>
 #include <zmk/events/position_state_changed.h>
 
-ZMK_EVENT_IMPL(zmk_position_state_changed);
\ No newline at end of file
+ZMK_EVENT_IMPL(zmk_position_state_changed);
+
+uint32_t zmk_last_event_trace_id = 0;
```
joelspadin, line 12: These should be static unless you mean for them to be globals accessible from other files (without going through zmk_event_trace_id).

app/src/events/position_state_changed.c
```diff
@@ -7,4 +7,15 @@
...
+uint32_t zmk_event_trace_ids[ZMK_KEYMAP_LEN] = {0};
+
+uint32_t zmk_event_trace_id(uint32_t position, bool generate) {
```
joelspadin, line 15: Maybe add "get" somewhere in the name? Having a verb somewhere in the name of a function usually makes it easier to understand what it's doing (unless you mean for "trace" to be the verb, but I'm assuming not given that it's used in variable names too).

Also, I'm generally not a fan of boolean parameters since in code like
```c
zmk_event_trace_id(position, TRUE);
```
you have no idea what the 2nd parameter does without looking at the definition of the function. That said, it is convenient for the couple places this is being called, though if it starts getting used elsewhere you may run into the same readability problem.

Consider something like this, which is more verbose but tells you exactly what it's doing:
```c
uint32_t zmk_event_trace_id_get(uint32_t position) {
  return zmk_event_trace_ids[position];
}
uint32_t zmk_event_trace_id_create(uint32_t position) {
  zmk_last_event_trace_id++;
  zmk_event_trace_ids[position] = zmk_last_event_trace_id;
  return zmk_event_trace_id_get(position);
}

...
event = {
  .trace_id = pressed ? zmk_event_trace_id_create(position) : zmk_event_trace_id_get(position),
};
```

Alternatively:
```c
uint32_t zmk_event_trace_id_get(uint32_t position) {
  return zmk_event_trace_ids[position];
}
void zmk_event_trace_id_update(uint32_t position) {
  zmk_last_event_trace_id++;
  zmk_event_trace_ids[position] = zmk_last_event_trace_id;
}

...
if (pressed) {
  zmk_event_trace_id_update(position);
}
event = {
  .trace_id = zmk_event_trace_id_get(position),
};
```

joelspadin (commented):
Going to bed now. Will take a look at the rest of this later.

app/src/events/position_state_changed.c
okke-formsma: Thanks for your extensive reply! I've added a commit that implements the create/get version.  I like how the intent is clearer when calling with a hardcoded value. 

On the other hand this solution exposes an implementation detail of trace_ids to the caller, by making the caller responsible for asking for a new trace_id or retrieving an existing trace_id. Having this decision within the trace_id method puts the responsibility in a better place in my opinion.

The update/get variety may be the best middle ground between the two. 

I see arguments for each option, I'm not quite sure what the best one is yet.

joelspadin: > On the other hand this solution exposes an implementation detail of trace_ids to the caller, by making the caller responsible for deciding if a new trace_id should be created or an existing one retrieved.

I'd argue that the caller is still responsible for deciding that in the version that takes a boolean parameter. You just one have one function that does one of two things based on a parameter instead of two functions that each do one thing. 

If you want to hide away the fact that new trace IDs get generated on key press, then renaming the parameter to something like `is_press` (like you mentioned) would be better, though it would still be difficult to understand if any later code ends up needing to pass just `TRUE` or `FALSE` for that parameter. You could also rename the "update" function to something like `zmk_event_trace_id_notify_pressed(position)` (though then it's not super clear if you're notifying the trace ID system of a press, as intended, or if you want to send a notification). Naming things is hard.

okke-formsma: I've decided to go with `zmk_get_event_trace_id(uint32_t position, bool pressed)`, as I think it's the clearest code for now. I don't expect other call sites that will use literal booleans, and if that will be the case we can always refactor this method.

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -184,44 +186,41 @@ static int sticky_key_keycode_state_changed_listener(const zmk_event_t *eh) {
...
+        if (sticky_key->trace_id == ev->trace_id) {
+            // don't catch events generated by the sticky key behavior itself
             continue;
         }
 
         // If events were queued, the timer event may be queued late or not at all.
         // Release the sticky key if the timer should've run out in the meantime.
         if (sticky_key->release_at != 0 && ev->timestamp > sticky_key->release_at) {
             stop_timer(sticky_key);
+            LOG_DBG("releasing 1");
```
joelspadin, line 202: Should these messages be left in? I'm not sure what the significance of release 1, 2, and 3 are.

joelspadin (commented):
I'm not super familiar with the problem being solved here, so I'll let Pete give the final approval, but this looks reasonable to me.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -446,9 +431,9 @@ static int position_state_changed_listener(const zmk_event_t *eh) {
         decide_hold_tap(undecided_hold_tap, HT_TIMER_EVENT);
     }
 
-    if (!ev->state && find_captured_keydown_event(ev->position) == NULL) {
-        // no keydown event has been captured, let it bubble.
-        // we'll catch modifiers later in modifier_state_changed_listener
+    if (ev->trace_id < undecided_hold_tap->trace_id) {
```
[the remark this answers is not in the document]
okke-formsma: Yeah I made a similar calculation and decided that if someone hits this bug they're due for some undefined behavior.

app/src/behaviors/behavior_sticky_key.c
okke-formsma: whoops

okke-formsma:
@petejohanson  could you have a look at this?

petejohanson (commented):
I need to consider this more, but I think the trace ID isn't really part of every payload, but part of some "event context" instead, and should be pervasive, and surfaced in the general event manager API, not hidden in some event payloads.

Thoughts?

okke-formsma:
> I need to consider this more, but I think the trace ID isn't really part of every payload, but part of some "event context" instead, and should be pervasive, and surfaced in the general event manager API, not hidden in some event payloads.

I'm not sure if it's applicable to many events outside the `position`, (future) `behavior` and `keypress` events. If we'd put the trace ID on the `zmk_event_t`, that would complicate construction for all events without clear benefit. Also, pointers to `zmk_event_t` would need to be kept around instead of pointers to the events themselves (e.g. `zmk_position_state_changed`), adding a bit of complexity too.  So, at least for now, I'd opt to keep the scope limited.

app/include/zmk/keymap.h
```diff
@@ -18,7 +20,7 @@ int zmk_keymap_layer_toggle(uint8_t layer);
 int zmk_keymap_layer_to(uint8_t layer);
 const char *zmk_keymap_layer_label(uint8_t layer);
 
-int zmk_keymap_position_state_changed(uint32_t position, bool pressed, int64_t timestamp);
+int zmk_keymap_position_state_changed(const struct zmk_position_state_changed *pos_ev);
```
okke-formsma, line 23: example of a function signature that would become more complex if we'd need to keep a `zmk_event_t` around instead of `zmk_position_state_changed`.

PR #675, opened 2021-02-09
fix(ext-power): Initialize as soon as settings are available

Nicell:
This adjusts the initialization of the `ext_power` driver to be right after the flash systems have been initialized (All flash systems initialize by `POST_KERNEL` priority `80`). This way we know `ext_power` is properly initialized for the rest of the devices that need external power on start.

Currently, this fixes displays not turning on after deep sleep, but this should prevent similar issues in the future.

This is a step towards recovering devices when `ext_power` is in use. The next step is tying more devices to the PM policy so they can re-initialize without the entire system needing to be reset when `ext_power` is used while we keep context (idle, behaviors, etc).

I'm wondering if this init priority should be a Kconfig value.

Some boards' power switch takes some substantial time to actually switch to powered on, so the `init-delay-ms` property has been added, which holds the system for `n` milliseconds after enabling (or disabling) the power switch on initialization. This makes sure that when devices like the OLED are initialized, the power is ready to be used.

~~This is also not fully tested but works in theory. We need to reconfirm this change works and doesn't break the settings integration with `ext_power`~~ Re-init tested by @mcrosson, @tominabox1, and @karnadii. Settings confirmed to hold on start by @mcrosson.

### DIFF
--- a/app/src/ext_power_generic.c
+++ b/app/src/ext_power_generic.c
@@ -23,6 +23,7 @@ struct ext_power_generic_config {
     const char *label;
     const uint8_t pin;
     const uint8_t flags;
+    const uint16_t init_delay_ms;
 };
 
 struct ext_power_generic_data {
@@ -171,6 +172,10 @@ static int ext_power_generic_init(const struct device *dev) {
     ext_power_enable(dev);
 #endif
 
+    if (config->init_delay_ms) {
+        k_msleep(config->init_delay_ms);
+    }
+
     return 0;
 }
 
@@ -210,7 +215,8 @@ static int ext_power_generic_pm_control(const struct device *dev, uint32_t ctrl_
 static const struct ext_power_generic_config config = {
     .label = DT_INST_GPIO_LABEL(0, control_gpios),
     .pin = DT_INST_GPIO_PIN(0, control_gpios),
-    .flags = DT_INST_GPIO_FLAGS(0, control_gpios)};
+    .flags = DT_INST_GPIO_FLAGS(0, control_gpios),
+    .init_delay_ms = DT_INST_PROP_OR(0, init_delay_ms, 0)};
 
 static struct ext_power_generic_data data = {
     .status = false,
--- a/app/src/ext_power_generic.c
+++ b/app/src/ext_power_generic.c
@@ -223,13 +223,15 @@ static const struct ext_power_api api = {.enable = ext_power_generic_enable,
                                          .disable = ext_power_generic_disable,
                                          .get = ext_power_generic_get};
 
+#define ZMK_EXT_POWER_INIT_PRIORITY 81
+
 #ifdef CONFIG_DEVICE_POWER_MANAGEMENT
 DEVICE_DEFINE(ext_power_generic, DT_INST_LABEL(0), ext_power_generic_init,
-              &ext_power_generic_pm_control, &data, &config, APPLICATION,
-              CONFIG_APPLICATION_INIT_PRIORITY, &api);
+              &ext_power_generic_pm_control, &data, &config, POST_KERNEL,
+              ZMK_EXT_POWER_INIT_PRIORITY, &api);
 #else
 DEVICE_AND_API_INIT(ext_power_generic, DT_INST_LABEL(0), ext_power_generic_init, &data, &config,
-                    APPLICATION, CONFIG_APPLICATION_INIT_PRIORITY, &api);
+                    POST_KERNEL, ZMK_EXT_POWER_INIT_PRIORITY, &api);
 #endif /* CONFIG_DEVICE_POWER_MANAGEMENT */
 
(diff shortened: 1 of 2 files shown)


mcrosson:
Initial testing looks good.

I have *not* tried checking the settings but the oled *is* re-initializing when waking from deep sleep.

tominabox1:
The additional delay fixed the issue I was having with my BLe Chiffre 👍🏼

mcrosson:
bump? sounds like this is working all around for folks

karnadii:
> The additional delay fixed the issue I was having with my BLe Chiffre 👍🏼

does this fix the wake up after deep sleep issue?

tominabox1:
Yes!

karnadii:
it is definitely fixes problem wake up after sleep. but when ext vcc is cut off the oled still not reinit. ext power toogle and rgb underglow still won't turn oled back on.

Nicell:
> it is definitely fixes problem wake up after sleep (the oled reinit and no freeze after wake up). but when ext vcc is cut off the oled still not reinit. ext power toggle and rgb underglow toggle still won't turn oled back on.

This is known and is a separate issue than the one this fixes. Power management policy related items need to be added to drivers that make use of ext power, but for now this fixes ext power not being on early enough during full initialization.

karnadii:
> > it is definitely fixes problem wake up after sleep (the oled reinit and no freeze after wake up). but when ext vcc is cut off the oled still not reinit. ext power toggle and rgb underglow toggle still won't turn oled back on.
> 
> This is known and is a separate issue than the one this fixes. Power management policy related items need to be added to drivers that make use of ext power, but for now this fixes ext power not being on early enough during full initialization.

ok, seems I misunderstand the PR. at least we have deep sleep now. I won't worry about my battery now. thanks for the PR.

karnadii:
I have tried on my marvelous65-split using nrfmicro 1.3. but it did not work, the oled did not re-init and after a while it freeze. while my marvelous65 (nrfmicro 1.4, non-split) worked. Please someone with split keyboard confirm this.

tominabox1:
in your .dts file you have to add the appropriate init delay

`ext-power {
		compatible = "zmk,ext-power-generic";
		label = "EXT_POWER";
		control-gpios = <&gpio0 7 GPIO_ACTIVE_LOW>;
		init-delay-ms = <250>;
	};`

karnadii:
> in your .dts file you have to add the appropriate init delay
> 
> `ext-power { compatible = "zmk,ext-power-generic"; label = "EXT_POWER"; control-gpios = <&gpio0 7 GPIO_ACTIVE_LOW>; init-delay-ms = <250>; };`

what is this `control-gpios = <&gpio0 7 GPIO_ACTIVE_LOW>`? is it like the underglow will be ignored?

tominabox1:
> > in your .dts file you have to add the appropriate init delay
> > `ext-power { compatible = "zmk,ext-power-generic"; label = "EXT_POWER"; control-gpios = <&gpio0 7 GPIO_ACTIVE_LOW>; init-delay-ms = <250>; };`
> 
> what is this `control-gpios = <&gpio0 7 GPIO_ACTIVE_LOW>`? is it like the underglow will be ignored? unused pin?

in the .dts file for your keyboard, if your board supports ext power toggle, there will be this section. With this PR, you have to add the init-delay-ms line to that block of code.

app/src/ext_power_generic.c
```diff
@@ -210,7 +215,8 @@ static int ext_power_generic_pm_control(const struct device *dev, uint32_t ctrl_
 static const struct ext_power_generic_config config = {
     .label = DT_INST_GPIO_LABEL(0, control_gpios),
     .pin = DT_INST_GPIO_PIN(0, control_gpios),
-    .flags = DT_INST_GPIO_FLAGS(0, control_gpios)};
+    .flags = DT_INST_GPIO_FLAGS(0, control_gpios),
+    COND_CODE_0(DT_INST_NODE_HAS_PROP(0, init_delay_ms), (0), (DT_INST_PROP(0, init_delay_ms)))};
```
petejohanson, line 219:
```suggestion
    .init_delay_ms = DT_INST_PROP_OR(0, init_delay_ms, 0)};
```

Use `DT_INST_PROP_OR`. I know there's other code that uses this approach, but it was before I found that helper.

app/src/ext_power_generic.c
```diff
@@ -223,13 +229,15 @@ static const struct ext_power_api api = {.enable = ext_power_generic_enable,
                                          .disable = ext_power_generic_disable,
                                          .get = ext_power_generic_get};
 
+#define ZMK_EXT_POWER_INIT_PRIORITY 81
```
petejohanson, line 232: This is fine for now, but something to consider moving to Kconfig as needed.

petejohanson (approved):
Thanks for digging into this!

PR #682, opened 2021-02-18
feat(split): Increase split interval during idle

Nicell:
This PR makes the split connection interval increase during idle. There's a brief period (about 0.6-1s) where the chances of having misordered keypresses are higher when coming out of idle, but this overall works well from my testing. This makes wireless battery consumption of the central half 1/3rd what it currently is when idle, and the peripheral's battery life is mostly unaffected.

For some reason, the `bt_conn_foreach` finds two connections rather than one that matches the role of master, and the first one always fails to update. Not sure what's up with that, but the second one updates the split connection properly.

Would appreciate people testing this branch to see how it performs for them. Specifically, I'd be interested in how this affects your typing experience (if at all).


scottg88:
I've tried this on my nice!nano + sofle split briefly. Haven't seen any issues when coming out of idle yet. I'll daily drive this for a day and see if any issues come up. I'm very interested in all improvements that help with battery life so I'll test wherever I can.

scottg88:
After trying this out for about 8 hours, I haven't practically run into any years. I can reproduce the out of order key presses you mentioned but I have to try hard, I didn't run into it in practice.

Nicell:
Thanks for testing this @scottg88. I don't currently daily drive a split, so I'd like to hear a few others' experiences before moving this forward. Sounds like this is pretty promising.

app/src/split/bluetooth/central_listener.c
```diff
@@ -0,0 +1,73 @@
...
+#include <zmk/event_manager.h>
+#include <zmk/events/activity_state_changed.h>
+
+static void set_sleep_params(struct bt_conn *conn, void *data) {
+    struct bt_conn_info info;
+
+    bt_conn_get_info(conn, &info);
+
+    if (info.role == BT_CONN_ROLE_MASTER) {
+        int err = bt_conn_le_param_update(conn, BT_LE_CONN_PARAM(0x0012, 0x0012, 10, 400));
```
petejohanson, line 16: Can we create `#define`s for this and the wake versions to make this a bit clearer?

petejohanson (changes requested):
Just one comment, otherwise, the *code* and approach looks good. Haven't had a chance to test myself yet.

caksoylar:
I have been testing this for three days on my split and haven't run into any issues. However I also couldn't intentionally reproduce the out of order keys (after waiting 30 seconds for idle).

Nicell:
An issue found from testing this PR: https://discord.com/channels/719497620560543766/784523018050469939/827426904470192159

> I've been using #682 for about a week, Iv not had any key order issues, but I have had the peripheral half drop connection with the host, a few times it may or may not be related,but it feels like it coincides with using #682

Nicell:
If anyone would like to pick this up, it would need some updates to use the Kconfig values that control connection parameters. Maybe new Kconfig options for idle multipliers could be defined. More testing around the two halves losing their connection when the connection settings update needs to be investigated.

PR #691, opened 2021-02-23
Add some clarity about enabling quick tap

cdarwin:
Simple addition to the docs

### DIFF
--- a/docs/docs/behaviors/hold-tap.md
+++ b/docs/docs/behaviors/hold-tap.md
@@ -45,6 +45,27 @@ Defines how long a key must be pressed to trigger Hold behavior.
 
 If you press a tapped hold-tap again within `quick_tap_ms` milliseconds, it will always trigger the tap behavior. This is useful for things like a backspace, where a quick tap+hold holds backspace pressed. Set this to a negative value to disable. The default is -1 (disabled).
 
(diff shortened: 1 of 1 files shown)


okke-formsma:
Thanks for the PR.

I'm not sure if I like this big swath of code in the docs there. Maybe just limit it to the one line? If we need further explanation about how to set settings, we should probably put that in a separate docs section and link to it.

urob:
I agree that this is not the right place to clarify on the structure of settings. But I also share the sentiment of the original post that it would be useful to explain somewhere in the docs the general structure of `.keymap` and other configuration files. Maybe a medium-sized example configuration going beyond the one given here https://zmk.dev/docs/features/keymaps could be useful. E.g., the medium-sized example could define:
-  two layers
- one or two combos
- a simple macro
- one or two custom behaviors
- change the default settings for some build-in behaviors such as `&lt` (this one can be a bit confusing, as it happens outside the main node)

caksoylar:
I think we can close this now since we have the tapping term example in both [mod-tap](https://zmk.dev/docs/behaviors/mod-tap#configuration) and [layer-tap](https://zmk.dev/docs/behaviors/layers#configuration) with pointers to hold-tap docs for all available properties, in addition to more generic docs at https://zmk.dev/docs/config#changing-devicetree-properties. Thanks for the PR!

PR #697, opened 2021-02-27
fix(combos): Fix stuck keys when pressing long combos.

### LINKED ISSUE #686: Keys in combos and mod-tap getting stuck
Board: nice!nano
Shield: QAZ
User repo: https://github.com/jreiter/zmk-config/tree/qaz

The firmware I was using at the time that I most recently reproduced this was built using Github actions here: https://github.com/jreiter/zmk-config/actions/runs/578609025

Sometimes when I'm typing on my QAZ with nice!nano using ZMK, I've noticed some keys get stuck.  So far, I've only seen this happen with keys that are involved in a combo or mod-tap - primarily shift (mod-tap in my keymap) and tab (combo in my keymap).  It's pretty easy for me to reproduce by doing typing tests, so it seems to primarily happen while typing as fast as possible (about 90-100 wpm for me).  Once the key is stuck, the only way I've been able to figure out to get my board working again is to get to bootloader and reflash.

tio logs attached from the last time I made this happen, when tab got stuck.
[tio.log](https://github.com/zmkfirmware/zmk/files/6005385/tio.log)

### LINKED ISSUE #696: Bug: sticky key quick-release releases too quickly
My &sk LSFT key configured with `quick-release` doesn't do anything if I try to use the sticky behavior.  Pressing it triggers shift, but on key release the sticky behavior is stopped immediately.

okke-formsma:
To properly retrigger hold-taps when a combo is _not_ activated, some position down events need to be reraised instead of released. The corresponding position up events were never reraised, causing a potential stuck key. The hold-tap would catch the reraised position down event, while the key up is processed normally. Once the hold-tap releases the position down event, the position up event has already occurred and we've got a stuck key.

tl;dr this fix makes sure position down and position up events are not reordered, preventing a stuck key.

fixes #686 and #713


app/src/combo.c
```diff
@@ -224,7 +224,7 @@ static int clear_candidates() {
 }
 
 static int capture_pressed_key(const zmk_event_t *ev) {
-    for (int i = 0; i < CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY; i++) {
+    for (int i = 0; i < CONFIG_ZMK_COMBO_MAX_KEYS_PER_COMBO; i++) {
```
okke-formsma, line 227: this is a bonus fix that I've sneakily included in this PR. I noticed the wrong constant was used here, which can lead to an array overflow.

BrainWart:
For those having this issue but not sure how to check it out, please change the `west.yml` file in your repo to the following:

```
manifest:
  remotes:
    - name: zmkfirmware
    - name: okke-formsma
      url-base: https://github.com/okke-formsma
  projects:
    - name: zmk
      remote: okke-formsma
      revision: bugfix-stuck-keys
      import: app/west.yml
  self:
    path: config
```

okke-formsma:
@petejohanson can we get this in soon? It's a real bug people have reported multiple times. I have run this code on my keyboard for almost two weeks and it's solid, other people also reported success with it

petejohanson (approved):

PR #700, opened 2021-02-28
fix(logging): remove unnecessary newline

okke-formsma:
Newlines are automatically added to logs. These newlines are very distracting while reading through big logfiles.


Nicell (approved):

petejohanson (approved):
Indeed! Thanks.

PR #701, opened 2021-02-28
refactor(core): make the event manager a bit easier to read

okke-formsma:
Gets rid of two levels of nesting by some trivial reordering.


petejohanson (approved):
Thanks!

Issue #699, opened 2021-02-28
Layer change event zmk_layer_state_changed for layer 0 inconsistent


ergodone:
In  looking at Layer change event I noticed there is a zmk_layer_state_changed event for layer 0 the first time a transition is made _back_ to it but no reports when it is initially activated or when it is subsequently returned to. 

I noticed this when I was trying to set layer leds based on the zmk_layer_state_changed event.

My key board has the mapping between layers set to: &to 1  -> &to 2  -> &to 3  -> &to 0

In the logs below you can see the first transition from layer 3->0 is reported but the subsequent one is not.  Note I added the _zmk_layer_state_changed_listener_ log message just before the event is sent.

Here's the patch I'd propose, see any issues with this?

```
diff --git a/app/src/keymap.c b/app/src/keymap.c
index 75ec6bf..7567d7f 100644
--- a/app/src/keymap.c
+++ b/app/src/keymap.c
@@ -19,8 +19,9 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/events/layer_state_changed.h>
 #include <zmk/events/sensor_event.h>
 
-static zmk_keymap_layers_state_t _zmk_keymap_layer_state = 0;
-static uint8_t _zmk_keymap_layer_default = 0;
+static const uint8_t _zmk_keymap_layer_default = 0;
+static zmk_keymap_layers_state_t _zmk_keymap_layer_state = BIT(_zmk_keymap_layer_default);
+
 
 #define DT_DRV_COMPAT zmk_keymap
 
@@ -87,7 +88,7 @@ static inline int set_layer_state(uint8_t layer, bool state) {
     zmk_keymap_layers_state_t old_state = _zmk_keymap_layer_state;
     WRITE_BIT(_zmk_keymap_layer_state, layer, state);
     // Don't send state changes unless there was an actual change
-    if (old_state != _zmk_keymap_layer_state) {
+    if (old_state != _zmk_keymap_layer_state || layer == _zmk_keymap_layer_default) {
         LOG_DBG("layer_changed: layer %d state %d", layer, state);
         ZMK_EVENT_RAISE(create_layer_state_changed(layer, state));
     }
```

```
[00:00:14.607,574] <dbg> zmk.position_state_changed_listener: 17 bubble (no undecided hold_tap active)
[00:00:14.607,604] <dbg> zmk.zmk_event_manager_handle_from: Listener captured the event
[...]

PR #702, opened 2021-02-28
fix(core): Initialize layer state with base layer activated

okke-formsma:
The first time layer 0 is activated, a 'layer activated' trigger is sent, but the layer can never be deactiated because it's the base layer. This is a bit unexpected.

This PR initializes the layer state to 1, so the base layer is always activated, and no "ghost triggers" for the base layer are sent.


okke-formsma:
a better solution was discussed in #699

PR #719, opened 2021-03-04
Update hold-tap.md documentation

elagil:
The tapping-term field had an incorrect name (- instead of _), which is now corrected. The example didn't compile previously.

### DIFF
--- a/docs/docs/behaviors/hold-tap.md
+++ b/docs/docs/behaviors/hold-tap.md
@@ -73,7 +73,7 @@ This example configures a hold-tap that works well for homerow mods:
 			compatible = "zmk,behavior-hold-tap";
 			label = "HOMEROW_MODS";
 			#binding-cells = <2>;
-			tapping-term-ms = <150>;
+			tapping_term_ms = <150>;
 			quick_tap_ms = <0>;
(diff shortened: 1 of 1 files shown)


petejohanson:
Odd. Both should work fine... See https://www.github.com/zmkfirmware/zmk/tree/main/app%2Fdts%2Fbindings%2Fbehaviors%2Fzmk%2Cbehavior-hold-tap.yaml

What error did you get?

caksoylar:
I indeed tried changing `tapping_term_ms` to `tapping-term-ms` and got a successful build with no difference in the produced binaries: https://github.com/caksoylar/zmk-config/actions/runs/628895980

elagil:
Interesting! It said:

`devicetree error: 'tapping-term-ms' appears in /behaviors/homerow_mods in polilla_w.dts.pre.tmp, but is not declared in 'properties:' in /workspaces/zmk/app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml`

But then I noticed that my branch is behind the master. So after updating, it builds. Sorry!

petejohanson:
@elagil no worries! Glad you got it sorted.

PR #720, opened 2021-03-05
only send zmk_battery_state_changed on change

ergodone:
Currently the battery polling sends a zmk_battery_state_changed every minute whether or not the charge level changes.  This change will only send zmk_battery_state_changed if the battery level changes.

### DIFF
--- a/app/include/zmk/events/battery_state_changed.h
+++ b/app/include/zmk/events/battery_state_changed.h
@@ -14,4 +14,6 @@ struct zmk_battery_state_changed {
     uint8_t state_of_charge;
 };
 
+int32_t zmk_battery_state_of_charge();
+
 ZMK_EVENT_DECLARE(zmk_battery_state_changed);
--- a/app/include/zmk/events/battery_state_changed.h
+++ b/app/include/zmk/events/battery_state_changed.h
@@ -14,6 +14,4 @@ struct zmk_battery_state_changed {
     uint8_t state_of_charge;
 };
 
-int32_t zmk_battery_state_of_charge();
-
 ZMK_EVENT_DECLARE(zmk_battery_state_changed);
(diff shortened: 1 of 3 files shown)


app/include/zmk/events/battery_state_changed.h
```diff
@@ -14,4 +14,6 @@ struct zmk_battery_state_changed {
     uint8_t state_of_charge;
 };
 
+int32_t zmk_battery_state_of_charge();
```
petejohanson, line 17: This should be defined in a separate header, not the event header. We keep the event headers "self contained" and focused.

app/src/battery.c
```diff
@@ -36,17 +40,23 @@ static int zmk_battery_update(const struct device *battery) {
...
-        return rc;
+        if (rc != 0) {
+            LOG_WRN("Failed to set BAS GATT battery level (err %d)", rc);
+            return rc;
+        }
+
+        rc = ZMK_EVENT_RAISE(new_zmk_battery_state_changed(
+            (struct zmk_battery_state_changed){.state_of_charge = state_of_charge.val1}));
+
+        last_state_of_charge = state_of_charge.val1;
```
petejohanson, line 56: We should set the value *before* raising the event, in case any listeners ignore the event payload and instead call the newly added API to retrieve the value.

petejohanson (changes requested):
Thanks for the PR! A few tweaks needed.

app/include/zmk/events/battery_state_changed.h
ergodone: added header and changed zmk_battery_state_of_charge(); to uint8_t to match event instead of sensor.

app/src/battery.c
ergodone: done

petejohanson (approved):
Thanks!

PR #724, opened 2021-03-07
fix(behaviors): Fix timing of delayed hold-tap trigger

okke-formsma:
A hold-tap timer event would be triggered too soon if the hold-tap
was delayed for longer than its tapping-term. This may cause
accidental hold behavior when the correct behavior would be tap.

By queuing the timer event instead of executing it immediately,
other delayed events get a chance to be processed properly.

This fixes an edge case introduced by https://github.com/zmkfirmware/zmk/pull/694.


okke-formsma:
@petejohanson could you have a look at this?

petejohanson (approved):
Thanks!

PR #726, opened 2021-03-15
feat(hid): Configurable NKRO HID report support.

petejohanson:
* Add Kconfig settings for NKRO or HKRO (boot compatible), HID
  report types for keyboard page.
* Updated report storage and set/unset for each config.

### DIFF
--- a/app/src/hid.c
+++ b/app/src/hid.c
@@ -69,8 +69,30 @@ int zmk_hid_unregister_mods(zmk_mod_flags_t modifiers) {
     return 0;
 }
 
+#if IS_ENABLED(CONFIG_ZMK_HID_REPORT_TYPE_NKRO)
+
+#define TOGGLE_KEYBOARD(code, val) WRITE_BIT(keyboard_report.body.keys[code / 8], code % 8, val)
+
+static inline int select_keyboard_usage(zmk_key_t usage) {
+    if (usage > ZMK_HID_KEYBOARD_NKRO_MAX_USAGE) {
+        return -EINVAL;
+    }
+    TOGGLE_KEYBOARD(usage, 1);
+    return 0;
+}
+
+static inline int deselect_keyboard_usage(zmk_key_t usage) {
+    if (usage > ZMK_HID_KEYBOARD_NKRO_MAX_USAGE) {
+        return -EINVAL;
+    }
+    TOGGLE_KEYBOARD(usage, 0);
+    return 0;
+}
+
+#elif IS_ENABLED(CONFIG_ZMK_HID_REPORT_TYPE_HKRO)
+
 #define TOGGLE_KEYBOARD(match, val)                                                                \
-    for (int idx = 0; idx < ZMK_HID_KEYBOARD_NKRO_SIZE; idx++) {                                   \
+    for (int idx = 0; idx < CONFIG_ZMK_HID_KEYBOARD_REPORT_SIZE; idx++) {                          \
         if (keyboard_report.body.keys[idx] != match) {                                             \
             continue;                                                                              \
         }                                                                                          \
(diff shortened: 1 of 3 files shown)


app/src/hid.c
```diff
@@ -69,8 +69,30 @@ int zmk_hid_unregister_mods(zmk_mod_flags_t modifiers) {
...
+    if (usage > ZMK_HID_KEYBOARD_NKRO_MAX_USAGE) {
+        return -EINVAL;
+    }
+    TOGGLE_KEYBOARD(usage, 0);
+    return 0;
+}
+
+#elif IS_ENABLED(CONFIG_ZMK_HID_REPORT_TYPE_HKRO)
+
 #define TOGGLE_KEYBOARD(match, val)                                                                \
```
okke-formsma, line 94: wouldn't it be simpler to inline this macro into the `(de)select_keyboard_usage` methods?

okke-formsma (commented):
I found something interesting that shows how a board can support both the boot format as wel as nkro:

https://www.devever.net/~hl/usbnkro

> Workarounds for buggy implementations. In practice, the vast majority of cheap, non-NKRO USB keyboards use the “boot protocol” as their default and only reporting format. For this reason, it's not wholly implausible that there might be BIOSes out there which expect the boot protocol, but fail to send the command requesting it. I don't have any data on the existence or prevalence of such BIOSes, but I have heard of at least one keyboard using a clever workaround: the boot protocol format can be prepended to the NKRO-enabled input reports, but marked as padding in the Report Descriptors. Something like this:
```
  1  ui  Bitfield of Depressed Modifier Keys (LCtrl, RCtrl, LShift, RShift, LAlt, RAlt, LWindows, RWindows)
  1  ui  Padding
  1  ui  "Padding" (actually: Scancode 1)
  1  ui  "Padding" (actually: Scancode 2)
  1  ui  "Padding" (actually: Scancode 3)
  1  ui  "Padding" (actually: Scancode 4)
  1  ui  "Padding" (actually: Scancode 5)
  1  ui  "Padding" (actually: Scancode 6)
 13  ui  Bitfield of Depressed Non-Modifier Keys
```
> Since boot protocol hosts completely ignore report descriptors, they examine the “padding” and get the data they are expecting. Hosts which properly examine report descriptors will correctly ignore the “padding” and use the NKRO-enabled key state bitfields in the latter area of the input report.

petejohanson:
@okke-formsma We could consider that hack, but it increases our payload size not insignificantly, so this risks blowing away the BLE MTU on some older devices.

I'd rather we get basic "select at build time" in, then work on something more advanced.

joelspadin:
> We could consider that hack, but it increases our payload size not insignificantly, so this risks blowing away the BLE MTU on some older devices.

Assuming that hack only applies to USB, another option would be to keep storing the HID report in the format used for BLE, but then whenever we send it to USB, run it through a function that applies the hack to it and sends that.

petejohanson:
Ok, just tweaked this implementation a bit, and also added a new choice `ZMK_HID_CONSUMER_REPORT_USAGES` that can be either `ZMK_HID_CONSUMER_REPORT_USAGES_FULL`, for the current "too broad" consumer usage range (we currently do a full two byte 0xFFFF max value, which many OS don't like, causes issues w/ some versions of macOS, etc., or the more narrow `ZMK_HID_CONSUMER_REPORT_USAGES_BASIC` which only supports single byte consumer usage values, up to 0xFF.

The default for now will be `ZMK_HID_CONSUMER_REPORT_USAGES_FULL`, pending a bit more testing, and a notice to users about the switch. Switching seems to require re-pairing for hosts (at least Linux) to pick up the report descriptor changes and to work properly.

petejohanson:
Tested on:

* Linux (GNOME)
* Android (11)
* macOS (Big Sur)

Using the NKRO and basic CC page settings.

The consumer page setting also rids of us this existing `dmesg` warning on Linux:

```
[779815.647705] hid-generic 0005:0000:0000.0064: ignoring exceeding usage max
```

As this doesn't change any behavior unless you change default Kconfig values, merging this. We can discuss changing the defaults for these two new Kconfig choices after more testing.

petejohanson:
Important note: To test properly, you'll likely need to unpair and re-pair with hosts, who might not update the report descriptor on each connect in order to re-connect faster.

claviger-pc:
Just wanted to note: I can confirm that NKRO works on the BT60v1 by enabling the option. But NKRO does not have support for F13-F24. I believe that this is a limitation inherent in NKRO: https://github.com/dhowland/EasyAVR/issues/9. Otherwise, with 6KRO the F13+ keys work on Windows. Maybe it is worth noting this somewhere in the documentation.

dmaone:
Just a note - full-speed USB packet size is 64 bytes (even USB 1.1, so support isn't an issue as long as it's a full-speed device), so one can just use boot keyboard descriptor with 62-key array for "NKRO" instead of a crippled bitmap.
By spec, BIOS must ignore additional data, so the only problem is generating KRO event - but even then, only when BIOS was dumb enough to not request boot protocol (which it must).

PR #729, opened 2021-03-16
fix(docs): Use absolute path for new shield link

Nicell:


innovaker:
This fixes the immediate issue but I'd prefer us to solve the root problem as this will only raise its head again elsewhere.

petejohanson:
> This fixes the immediate issue but I'd prefer us to solve the root problem as this will only raise its head again elsewhere.

Did you have a suggestion? I'm not seeing anything in the Docusaurus docs that allows us to extend the validation easily.

Issue #554, opened 2020-12-29
per-key led feature


Retic1337:
creating this to start work on the per key led lighting.

joelspadin:
I think the first thing to do is to get a list of use cases for per key lighting. Here are a few I can think of:

1. Set different solid colors per key (e.g. mods one color and everything else a second color).
2. Dim/disable lighting for keys that have no function in the current layer.
3. Status LEDs (caps lock, num lock)
4. Static animations (e.g. color wave across keyboard)
5. Dynamic animations (e.g. ripples from pressed keys)

Other considerations:

It would be nice to have one system that manages all LEDs and then have mappings on top of that for per key LEDs, underglow LEDs, and standalone status LEDs. Having separate systems for each of those would likely lead to inconsistencies and make it difficult to do things like creating an animation that syncs between key and underglow LEDs.

Just like with the existing kscan and composite kscan drivers, we should support multiple LED drivers (e.g. direct GPIO, daisy chained ws2818, Lumissil matrix driver) and combinations of them controlling different subsets of LEDs.

Storing colors as red, green, and blue bytes is probably the format that works best across all LED drivers. We can still convert to/from HSV as needed for things like adjusting the underglow color. To support single color LEDs with the same system, the driver can simply read the red channel and adjust a PWM or use an on/off threshold.

We need to somehow support wireless splits. Central could send per LED color changes to the peripheral, but that probably won't perform well with fast animations. The peripheral likely needs to run the same LED control code as the central and just get high level commands like hue/brightness changes from central.

Retic1337:
for #1 and #2 we will need to import the keymap into the driver to determine colors.
#3 will require reading status of caps, num, etc..
#4 i think would have to be a map of the leds position x and y relative to their placement on the pcb. so esc on a tkl would be 0,255 and right arrow would be 255,0. 
#5 would be easy once #4 is done.
Currently have a background color working with per key color changing based upon keypress. Where are we going to want to be creating led maps and defining colors etc, board.keymap? or in the device-tree?

joelspadin:
> for #1 and #2 we will need to import the keymap into the driver to determine colors.
> #3 will require reading status of caps, num, etc..

I think we should structure things so that the low-level LED drivers don't have any knowledge of the keymaps. Instead, something built on top of that (not sure if it would also be considered a driver or just application code) should be deciding what color each LED should be and sending commands to the LED driver(s). It is this higher level code that would need to know about the keymap and lock statuses.

[...]

PR #730, opened 2021-03-16
Implement a driver for ISSI IS31FL3743A

ice9js:
This PR adds support for a IS31FL3743A LED matrix driver as a proposition for more energy efficient 'bling' on ZMK powered devices. Hopefully it can also be leveraged in implementing per-key RGB support (#554).

I took some notes from #555 and made the driver conform to Zephyr's `led_strip` API.

The driver works by setting all configuration and white-balance scaling registers on init, and only uses PWM registers for manipulating LED brightness and color.

My test configuration for reference:

```
@i2c0 {
	is31fl3743a@20 {
		compatible = "issi,is31fl3743a";
		label = "IS31FL3743A_0";
		reg = <0x20>;
		riset = <68>;
		sw-setting = <0x01>;
		led-max-current = <5>;
		cs-order = [01 02 00 04 05 03 07 06 08 0A 0B 09 0D 0E 0C 10 11 0F];
		sdb-gpios = <&gpio1 13 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>;
		sync = <0x11>;
	};
};
```

Food for thought: I wonder what the best way would be for handling software/hardware shutdown to preserve energy when LEDs aren't in use. Zephyr's `led_strip` API seems to leave this question open.


joelspadin:
> Food for thought: I wonder what the best way would be for handling software/hardware shutdown to preserve energy when LEDs aren't in use. Zephyr's `led_strip` API seems to leave this question open.

Zephyr has a power management API that we can implement for each driver: https://docs.zephyrproject.org/latest/reference/power_management/index.html#device-model-with-power-management-support

For example, putting this driver into `DEVICE_PM_LOW_POWER_STATE` could toggle the SDB pin. Then we could start doing things like putting LED drivers into low power state when all the LED colors are (0,0,0) or when the keyboard is idle.

app/dts/bindings/led_strip/issi,is31fl3743a.yaml
```diff
@@ -0,0 +1,65 @@
...
+  cs-order:
+    type: uint8-array
+    required: false
+    default: [0x02, 0x01, 0x00, 0x05, 0x04, 0x03, 0x08, 0x07, 0x06, 0x0B, 0x0A, 0x09, 0x0E, 0x0D, 0x0C, 0x11, 0x10, 0x0F]
+    description: |
+      Stores the configuration of CS* pins.
+      Each 3 consecutive numbers must be R, G and B pins for a single LED.
+      Matches the reference design from the datasheet by default.
+
+  riset:
```
joelspadin, line 40: Since we don't have capitalization to make it obvious the "R" is separate from "iset", maybe name this `r-iset`?

app/dts/bindings/led_strip/issi,is31fl3743a.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - 0x10
+      - 0x11
+    required: false
+    default: 0x00
+    description: |
+      Controls the SYNC function. Set to 0x11 to set the device as main or 0x10 for secondary.
+
+  sdb-gpios:
+    type: phandle-array
+    required: true
```
joelspadin, line 63: This one should probably be optional in case your hardware doesn't have a GPIO for the SDB pin and just has it wired so the chip is always active.

app/drivers/led_strip/is31fl3743a/Kconfig
```diff
@@ -0,0 +1,20 @@
...
+	help
+		Enable an IS31FL3743A LED Matrix driver.
+
+			The IS31FL3743A is a general purpose 18xn(n=1-11) LED Matrix
+			programmed via 1MHz I2C compatible interface. Each LED can be
+			dimmed individually with 8-bit PWM data and 8-bit DC scaling data
+			which allows 256 steps of linear PWM dimming and 256 steps of DC
+			current adjustable levels.
+
+config LED_STRIP_INIT_PRIORITY
```
joelspadin, line 16: Zephyr appears to define a config with the same name in its led_strip/Kconfig file: https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_LED_STRIP_INIT_PRIORITY.html

Can we just use that? If not, will this conflict with that one? If we need to define our own config for this, it should probably also live in a Kconfig file that is common to all led_strip drivers so it can be used by others later.

app/dts/bindings/led_strip/issi,is31fl3743a.yaml
ice9js: Yeah, I was wondering about that and there's another edge case: on my board, as well as what the reference designs suggest, if you have multiple of these running in sync you'd use a single SDB pin to control all of them.

So while we can make this optional to support designs with SDB connected to GPIO to have it work at all, it complicates things for taking advantage of it for power saving.

In that case, what I'd suggest is to use software shutdown by default and have another setting, say `enable-hardware-shutdown` which would then use hardware shutdown instead if it makes sense for the board.
And regardless of that, if `sdb-gpios` is defined, it'll always be pulled high on initial startup.

Does that make sense?

Sidenote: for this chip in particular, the datasheets suggests identical power usage in both software and hardware shutdown mode. For the larger IS31FL3741 there seems to be about 1uA difference. So maybe it's not worth it to bother with hardware shutdown at all?

app/drivers/led_strip/is31fl3743a/is31fl3743a.c
```diff
@@ -0,0 +1,249 @@
...
+ */
+static int is31fl3743a_strip_update_channels(const struct device *dev, uint8_t *channels,
+                                             size_t num_channels) {
+    const struct is31fl3743a_config *config = dev->config;
+
+    if (config->px_buffer_size < num_channels) {
+        return -ENOMEM;
+    }
+
+    return is31fl3743a_reg_burst_write(dev, 0x01, channels, num_channels);
```
ice9js, line 146: Note to self: ensure the page select register is set to the PWM page first.

mattvilim:
Are there plans to ever merge these LED matrix drivers into main? Would be very useful. 😄

joelspadin:
I think this driver needs some updates to fit with newer Zephyr, but I'm currently designing some hardware that will use one of these, so I will eventually be trying to get this through review.

One example of a change that should likely be made is to copy the `color-mapping` property from Zephyr's ws2812 driver, which gives the order of the color channels. If, for example, if you set `color-mapping = <LED_COLOR_ID_RED>`, then you should get a strip of up to 351 single color LEDs (controlled by the red channel of each pixel) instead of 117 RGB LEDs.

It might also be useful to be able to reorder the sw pins? Maybe that could be handled by a higher level transform similar to the kscan matrix transform though.

ice9js:
Just going to note here that this is somewhat outdated and I think my implementation for IS31FL3741 (#886) is much better in that regard. I was meaning to update this to whatever we do there eventually, but haven't gotten much feedback on it so far either.

Regarding the stuff you mentioned @joelspadin, the IS31FL3741 implementation does indeed allow to map individual channels, not just entire RGB LEDs like so:

```
map = <
    RGB(SW(1), CS(1), CS(2), CS(3))
    RGB(SW(7), CS(3), CS(2), CS(1))
>;
```

These macros expand to three integers which are the register numbers on the ISSI chip for the three LEDs. The driver does assume every three subsequent positions are the red, green and blue channels for a single RGB LED, and I think the macros go a long way of making that manageable so you can assign them in whatever order makes sense for your project.  
If you're not using RGB LEDs this still allows for full control over the order, but of course `update_rgb` will produce unexpected results in that case.

The `RGB()` macro work both ways, as in you can also do `RGB(CS(i), SW(j), SW(k), SW(l))` if you happen to be using common cathode LEDs. That said it probably still makes sense to have another macro that just takes a single pair if you're doing something really weird, and to make it easier to assign non-rgb LEDs.

Either way, we should probably continue that discussion in #886 when the time comes.

vladkvit:
For posterity - I think you meant issue #866 and not #886.

caksoylar:
(Disclaimer: opinion not related to being associated with ZMK project as accepting it isn't my decision to make anyway)

Would you consider making this a [module](https://zmk.dev/docs/features/modules) so that it can be used independently from a ZMK branch? The structure looks like it would cleanly transfer. Here is an example of such a driver: https://github.com/petejohanson/cirque-input-module

ice9js:
I think that's a good idea. My intention is to eventually push this to Zephyr itself, but it seems like a module would be a nice intermediate solution for anyone interested in the meantime.

It's been a while since I last touched it so for now I'm trying to make sense of it again, but I should be able to set it up in the next few weeks.

ice9js:
Just noting the implementation of this driver has been completed and is now available as [a standard zephyr module](https://github.com/crystalplanet/zephyr-led-strip-drivers).

nmunnich:
@ice9js would it be possible for you to try and upstream the drivers for those chips to Zephyr? That would be the ideal end result, and definitely beats having them in-tree of ZMK or in a module.

ice9js:
>@ice9js would it be possible for you to try and upstream the drivers for those chips to Zephyr?

Yeah, I did give it a look briefly and it's still my intention but I have no idea when I'll get around to it. The code will work as is, I just haven't gotten around to figuring out core Zephyr development for writing some tests or providing test instructions, which will probably be welcome. But maybe I can at least get a conversation going before EOY with what I already have.

PR #733, opened 2021-03-19
Fix &pro_micro_a map for A10 for BlueMicro840_v1

layertwo:
The pin for &pro_micro_a was incorrectly set to 1.13 when it should have been 1.11. See the relevant schematic as found [here](https://d3s5r33r268y59.cloudfront.net/datasheets/21640/2020-12-20-04-29-38/schematic_bluemicro840_1.0.pdf).

I've tested it on one of my boards, and confirmed that it is working as expected.


petejohanson (approved):
Good catch! Thanks.

PR #738, opened 2021-03-23
Docs: Add icon legend to the feature matrix

DarkFox:


docs/docs/intro.md
```diff
@@ -11,6 +11,10 @@ firmware built on the [Zephyr™ Project](https://zephyrproject.org/) Real Time
 
 ZMK is currently missing some features found in other popular firmware. This table compares the features supported by ZMK, BlueMicro and QMK:
 
+| Legend: | ✅ Done | 🚧 Under Development | 💡 Planning stage |
```
joelspadin, line 14: Minor inconsistency in capitalization here between "Under Development" and "Planning stage". I don't think it matters much whether we capitalize each word or just the first, but we should do the same everywhere.

Also, I'd suggest "Supported" instead of "Done", as I'm not sure it's immediately obvious what it means for a feature to be done.

Maybe "Planned" or "In planning" instead of "Planning stage" as well? "Planning stage" seems a little awkward to me, but I don't know what the clearest way to word that is.

DarkFox: I fully agree. I couldn't come up with good names for them at the time. :)

petejohanson (approved):
Thanks!

PR #742, opened 2021-03-27
Keymap adjustment: fixed right alt and added layer quick tap to the Jian

krikun98:
The quick tap feature was implemented after the Jian was merged, but it's an integral part of using the keymap. I've been using it for a couple months now, and it works great.

I've got a similar PR in the works for the Jorne, but that has some config issues regarding RGB.


krikun98:
If someone were to label this, `keymaps` and `shields` would be a great fit.

petejohanson (approved):
Thanks!

PR #744, opened 2021-03-28
docs: Fix keymap upgrader

joelspadin:
Bumped web-tree-sitter to v0.19.4 and added v0.4.0 of tree-sitter-devicetree, which had to be rebuilt to work in v0.19.x (https://github.com/joelspadin/tree-sitter-devicetree/releases/tag/v0.4.0)

Changed how we patch web-tree-sitter to correctly load tree-sitter.wasm to work with the latest version of Docusaurus. Including a copy of tree-sitter.wasm as a static resource is no longer needed.


Nicell (approved):
LGTM, thanks for fixing

Issue #651, opened 2021-01-30
Grave Escape/Mod-Morph documentation


Nicell:
#543 was recently merged, which adds support for grave escape through the more generic mod-morph behavior. We need to document both of these.

PR #749, opened 2021-04-02
Docs: added mod morph page

measlesbagel:
This is my first pull request and my first time writing documentation so let me know if there's any problems or anything I wrote is unclear.

Created in reference to issue #651

### DIFF
--- a/docs/docs/behaviors/mod-morph.md
+++ b/docs/docs/behaviors/mod-morph.md
@@ -1,0 +1,72 @@
+---
+title: Mod-Morph Behavior
+sidebar_label: Mod-Morph
+---
+
+## Summary
+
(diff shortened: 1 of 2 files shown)


okke-formsma (approved):
Thanks! Looks good to me (I think we need another reviewer to merge this)

caksoylar:
Would it be worth mentioning that the seconed keycode would be sent *with* the held modifiers (i.e. the held modifiers are not cleared)?

Nicell:
@caksoylar I think this looks good too, and I agree that adding a part about the fact that held modifiers are also sent and not cleared is an important piece to have in the docs.

measlesbagel:
added a note about how modifiers are sent along with the morphed keypress. hopefully everything looks good now.

Nicell (approved):
Thanks, @measlesbagel!

PR #750, opened 2021-04-03
feat: Add nrf VDDH battery driver

joelspadin:
Added a driver which uses the nRF52's ADC channel on the VDDH pin to read the battery voltage when using high voltage mode. 

The voltage divider driver technically could be configured to work for this, but the way you'd configure it would be really weird and arbitrary, so I figured a separate driver would be easier to use:

```
vbatt {
    // Channel must be SAADC_CH_PSELN_PSELN_VDDHDIV5 - SAADC_CH_PSELP_PSELP_AnalogInput0
    io-channels = <&adc (0x0D - 1)>;
    // Multiply ADC result by 5
    full-ohms = <5>;
    output-ohms = <1>;
};
```

### DIFF
--- a/app/drivers/sensor/battery/battery_nrf_vddh.c
+++ b/app/drivers/sensor/battery/battery_nrf_vddh.c
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2021 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * This is a simplified version of battery_voltage_divider.c which always reads
+ * the VDDHDIV5 channel of the &adc node and multiplies it by 5.
+ */
+
+#define DT_DRV_COMPAT zmk_battery_nrf_vddh
+
+#include <device.h>
+#include <devicetree.h>
+#include <drivers/adc.h>
+#include <drivers/sensor.h>
+#include <logging/log.h>
+
+#include "battery_common.h"
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#define VDDHDIV (5)
+
+static const struct device *adc = DEVICE_DT_GET(DT_NODELABEL(adc));
+
+struct vddh_data {
+    struct adc_channel_cfg acc;
+    struct adc_sequence as;
+    struct battery_value value;
+};
+
+static int vddh_sample_fetch(const struct device *dev, enum sensor_channel chan) {
+    // Make sure selected channel is supported
+    if (chan != SENSOR_CHAN_GAUGE_VOLTAGE && chan != SENSOR_CHAN_GAUGE_STATE_OF_CHARGE &&
+        chan != SENSOR_CHAN_ALL) {
+        LOG_DBG("Selected channel is not supported: %d.", chan);
+        return -ENOTSUP;
+    }
+
+    struct vddh_data *drv_data = dev->data;
+    struct adc_sequence *as = &drv_data->as;
+
+    int rc = adc_read(adc, as);
+    as->calibrate = false;
+
+    if (rc != 0) {
+        LOG_ERR("Failed to read ADC: %d", rc);
+        return rc;
+    }
+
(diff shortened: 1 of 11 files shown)


app/drivers/sensor/battery/battery_nrf_vddh.c
```diff
@@ -0,0 +1,147 @@
...
+        val->val1 = drv_data->state_of_charge;
+        val->val2 = 0;
+        break;
+
+    default:
+        return -ENOTSUP;
+    }
+
+    return 0;
+}
```
Nicell, line 89: It seems to me that this function is basically identical to the voltage divider driver. Can we bring this function to `battery_common.c`?

app/drivers/sensor/battery/battery_nrf_vddh.c
```diff
@@ -0,0 +1,147 @@
...
+    if (drv_data->adc == NULL) {
+        LOG_ERR("Failed to find &adc device");
+        return -ENODEV;
+    }
+
+    // TODO: (Zephyr 2.5.0)
+    // if (!device_is_ready(drv_data->adc)) {
+    //     LOG_ERR("ADC device is not ready %s", drv_data->adc->name);
+    //     return -ENODEV;
+    // }
```
Nicell, line 111: Do these same Zephyr 2.5.0 TODOs need to be complete on the voltage divider driver?

Nicell (changes requested):

app/drivers/sensor/battery/battery_nrf_vddh.c
joelspadin: The only reason I hadn't done that already was that the driver data structs are different. I could pull out the code that populates val given a state of charge or a millivolts value pretty easily though.

app/drivers/sensor/battery/battery_nrf_vddh.c
joelspadin: These aren't required when moving to 2.5.0, but they make use of a new API that I think simplifies driver initialization. Similar changes could likely be made to the voltage divider driver and throughout the rest of the codebase.

app/drivers/sensor/battery/battery_nrf_vddh.c
```diff
@@ -0,0 +1,147 @@
...
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#define VDDHDIV (5)
+
+struct vddh_data {
+    const struct device *adc;
+    struct adc_channel_cfg acc;
+    struct adc_sequence as;
+    uint16_t adc_raw;
+    uint16_t millivolts;
```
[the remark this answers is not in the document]
joelspadin: Will do.

app/drivers/sensor/battery/battery_nrf_vddh.c
joelspadin: Switch both battery drivers to using a common data struct and pulled this out into a common function.

app/drivers/sensor/battery/battery_nrf_vddh.c
joelspadin: I'm not sure how the IO channels for the voltage divider driver work, so I didn't mess with this. It might be possible to simplify it with something like this though?
```c
.adc =  DEVICE_DT_GET(DT_INST_IO_CHANNELS_LABEL(0)),
```

joelspadin:
I also included a change to find the battery sensor via chosen node instead of "BATTERY" label, mainly because that made testing things easier on my board that has two possible battery sensors. Let me know if you'd like me to split that out into a separate PR.

joelspadin:
Moved the battery chosen node to #868

Nicell (approved):
Code LGTM. I'll test this on the nice!nano v2 today or tomorrow to confirm it's working for me.

Update: Confirmed working with proper percentages on both this new VDDH driver as well as the voltage divider driver.

PR #752, opened 2021-04-05
fix(behaviors): Fix sticky keys quick-release for normal keypresses

### LINKED ISSUE #696: Bug: sticky key quick-release releases too quickly

okke-formsma:
Quick release for sticky keys failed for non-layer keys. The sticky key
was released just before the key that was supposed to be modified was
handled.

The issue was caused by an error in the sticky key logic, which released
the sticky key before handling the key up event.

Fixes #696.


petejohanson (approved):
Thanks!

PR #753, opened 2021-04-05
Testing: Use debug builds and high resolution timer

okke-formsma:
By setting CONFIG_DEBUG, the native_posix builds will not be optimized
which makes debugging them much nicer.

By setting CONFIG_SYS_CLOCK_TICKS_PER_SEC to 1000, debug prints have a
higher resolution and not always show up as multiples of 10ms.


petejohanson (approved):
Sweet!

PR #754, opened 2021-04-06
Fix nibble demux scan errors on encoder row

jaygreco:
I was getting erroneous presses when pressing the encoder. This PR fixes that.

+ Add a 1us sleep to let the column selection settle in order to avoid spurious keypresses when row capacitance is high (like on the encoder row)

### DIFF
--- a/app/drivers/kscan/kscan_gpio_demux.c
+++ b/app/drivers/kscan/kscan_gpio_demux.c
@@ -113,6 +113,8 @@ struct kscan_gpio_item_config {
                     &kscan_gpio_output_configs_##n(dev)[bit];                                      \
                 gpio_pin_set(out_dev, out_cfg->pin, state);                                        \
             }                                                                                      \
+            /* Let the col settle before reading the rows */                                       \
+            k_usleep(1);                                                                           \
                                                                                                    \
             for (int i = 0; i < INST_MATRIX_INPUTS(n); i++) {                                      \
                 /* Get the input device (port) */                                                  \


petejohanson:
@mcrosson can you test this?

mcrosson:
@petejohanson I can give it a test soon.

@jaygreco Can you explain how the error that this addresses can be re-produced? I don't use my encoder row much and could use some insights for verification.

jaygreco:
@mcrosson It's a hard error to replicate since it's marginal and varies board-to-board. If you're not having the issue with your NIBBLE as built, I don't know how easy it will be to repro. I think the important thing to test is the key matrix functionality works as expected with the change rolled in.

mcrosson:
@jaygreco could this also affect the bottom row of the board?

jaygreco:
> @jaygreco could this also affect the bottom row of the board?

It could happen on any row, the settling time of the demux is the same regardless of which column it's addressing. The delay between column addressing and row reading will prevent the issue from happening on any of the rows.

mcrosson:
I ask because I'm testing an MCU presently and am noticing mild hiccups with my nibble's bottom row suddenly ; I wanted to verify this may be a potential fix before digging into whether or not something else is happening.

jaygreco:
If it's using ZMK without this patch, then yes, it could be this issue.

mcrosson:
Thank you for the feedback ; I'll apply the patch semi-soon and re-verify.

petejohanson:
@mcrosson Have you had a chance to test this? Fixed for you?

mcrosson:
@petejohanson Tested and godo to go.

mcrosson (approved):

petejohanson (approved):

Issue #755, opened 2021-04-06
Docs Combo example does not compile


MvEerd:
The example for combos in the documentation does not seem to compile;
https://github.com/MvEerd/zmk-config/commit/c4fb41e1ddab4874b30cc40f8c0552ebff7e5f0b

It errors on the `layers = <-1>;` line.
After removing this line or [changing the value to a non-negative integer](https://github.com/MvEerd/zmk-config/commit/dc19e606d863c491bc1dd8947b0e0df0b96d66e5) it compiles fine.
Any other array containing a negative integer seems to fail as well; `layers = <0 -1>;`

Is this a devicetree syntax bug or limitation?
I'm not able to find much more on this than the following;
> 'type: array' is for properties that are assigned zero or more 32-bit values

https://docs.zephyrproject.org/2.5.0/guides/dts/bindings.html

okke-formsma:
Apparently the 'all layers' feature only works by not providing any value for the `layers` parameter, so it defaults to -1.

MvEerd:
The -1 is indeed superfluous in this case. I just wondered where the syntax error came from before offering a PR for the docs to say so.
Is it a devicetree limitation? I don't see a configuration affecting this apart from the default being set

okke-formsma:
https://docs.zephyrproject.org/latest/guides/dts/intro.html#syntax-and-structure 

> Devicetree nodes can also have properties. Properties are name/value pairs. Property values can be any sequence of bytes. In some cases, the values are an array of what are called cells. A cell is just a 32-bit unsigned integer.

MvEerd:
Ah, so they are indeed unsigned, that explains it. Thanks for pointing that part of the docs out!

Was indeed confused that you could still configure it as `-1` default

joelspadin:
> So it's actually kind of funky that the -1 default works :')

I'm pretty sure you're just getting `(uint32_t)(-1)` here, so the max possible uint32_t. Then since `combo_cfg::layers` is an array of signed integers, you get `(int8_t)(uint32_t)(4294967295)`, which fortunately happens to be -1.

MvEerd:
Fixed by #757

PR #757, opened 2021-04-06
fix(docs) omit layers -1 example, clarify it defaults when omitted

MvEerd:
A user on Discord was having trouble compiling their firmware with the combos example, this was due to the `<-1>` being invalid devicetree syntax for an array.
As @okke-formsma pointed out in #755 this is because array values are 32-bit unsigned.

I have removed the `layers = <-1>;` line in the combos example and clarified that it will default to the global scope when it is omitted.
I chose to remove the property because the example makes sense in a global scope and the property is optional, if the preference is to include it as example with valid layers let me know and I'll do that instead (e.g. `layers = <0 1>;`)

### DIFF
--- a/docs/docs/features/combos.md
+++ b/docs/docs/features/combos.md
@@ -18,7 +18,6 @@ Combos configured in your `.keymap` file, but are separate from the `keymap` nod
 			timeout-ms = <50>;
 			key-positions = <0 1>;
 			bindings = <&kp ESC>;
-			layers = <-1>;
 		};
 	};
 };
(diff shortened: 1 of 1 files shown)


docs/docs/features/combos.md
```diff
@@ -28,7 +27,7 @@ Combos configured in your `.keymap` file, but are separate from the `keymap` nod
 - The `compatible` property should always be `"zmk,combos"` for combos.
 - `timeout-ms` is the number of milliseconds that all keys of the combo must be pressed.
 - `key-positions` is an array of key positions. See the info section below about how to figure out the positions on your board.
-- `layers = <0 1...>` will allow limiting a combo to specific layers. this is an _optional_ parameter and defaults to `-1` which is global scope.
+- `layers = <0 1...>` will allow limiting a combo to specific layers. this is an _optional_ parameter, when omitted it defaults to `-1` which is global scope.
```
okke-formsma: Please capitalize `This` and don't mention the "defaults to -1" part. Simply "it defaults to global scope" is better because otherwise people will try to use the -1 value anyway :)

okke-formsma (commented):
Thanks! A small comment.

docs/docs/features/combos.md
MvEerd: Agreed, better to keep `-1` under the hood then. 
Fixed

joelspadin (approved):

okke-formsma (approved):

Nicell (approved):

PR #771, opened 2021-04-25
New shield Dactyl Manuform

DuMaM:
Hi Everyone, :wave: 

I would like to add my Dactyl Manuform implementation for this repository. :tada: 
It's almost finished but I need some advice, before I move forward. 

1. I have board specific settings and I don't know how I could add them in zmk/board folder. I would like to extend or overwrite settings already defined in this board inside Zephyr project. I was try to do so, but without success. Board name: Particle Xenon.
2. I know that this shield is split one type, I will add support for it in feature.
3. All current boards support only Arduino pico headers, so how I should handle Adafruit ones? I was contacting @petejohanson on discord long time ago about it. He said that there are plans to do, so in feature. Few months got passed and I haven't seen any support for it yet. I could even implement this by myself, but I need some advice how and If there is already some kind of roadmap for this feature. 
4. Dactyl Manuform has few variations in terms of size (4x5, 5x5, etc...). Do you have any idea how to handle this? I was thinking of just using multiple overlays, but maybe somebody knows cleaner way. :wink:  
5. In QMK there was problem of user pinout configuration for this shield. Each dactyl maintainer had his own pinout/layout, and most of them placed this in main repo dactyl folder. After some time it created quite a mess and maintenance problems, because we need to also update their configs... Is there any way of adding pinout config in zmk-config?

Sorry If I'm contacting in bad place.

Thanks for all support in advance. :yellow_heart:


DuMaM:
This shield is not according to standards used by ZMK :sob:

PR #776, opened 2021-05-02
Added support for prefers-color-scheme in docs

gbdlin:
Now, instead of always defaulting to light theme, documentation, docs
will default to the color scheme based by the user preference (reported
by web browser as a `prefers-color-scheme` media query).

It is still possible for user to change the theme by using the switch
next to the search box. His preference will be remembered.

I've tested all existing documentation pages for compatibility with the dark
color theme and I've found no major ones.


petejohanson (approved):
Simple enough to me. Tested from laptop and phone (dark mode Android), and it works as expected. Thanks!

PR #777, opened 2021-05-02
Readjusted colors for dark theme for OSes in docs

gbdlin:
Previously the colors of operating systems were the same both for dark
and light color scheme. That's not a major issue, but adjusting the
colors for the dark theme to better match the theme looks like an
improvement.

Colors were just darkened by reversing the lightness of each color taken
from the dark theme.


petejohanson:
Updated:



Vs

Current:



Looks much better to me. Thanks!

petejohanson (approved):
Thanks!

PR #802, opened 2021-05-19
Add missing semicolon to examples in doc

stuecklberger:
I was running into errors when copying example code from the documentation to my personal zmk-config. This was due to missing semicolons in at least two cases, so I want to fix those errors in the doc to maybe save others some time.

Not too experienced with contributing to open source projects. I ran through the Development Guidelines in the doc and executed the checks / linting / build, hope this is all that is required :)


petejohanson (approved):
Great catches, thanks!

PR #811, opened 2021-05-26
feat(behavior) Add Tap Dance

kurtis-lew:
Add tap dances as a behavior. Tap dances are designed to have the following characteristics:
- [x] Can include any ZMK keycode or behavior as a binding (tested regular keypresses, sticky keys, toggle layers, hold-taps as of initial PR)
- [x] Resolve immediately after the number of keypresses is equal to the number of bindings
- [x] Resolve immediately and end when the next keypress's position is not that of the last-pressed tap dance


app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,270 @@
...
+
+static int tap_dance_position_state_changed_listener(const zmk_event_t *eh);
+
+ZMK_LISTENER(behavior_tap_dance, tap_dance_position_state_changed_listener);
+ZMK_SUBSCRIPTION(behavior_tap_dance, zmk_position_state_changed);
+
+static int tap_dance_position_state_changed_listener(const zmk_event_t *eh) {
+    LOG_DBG("Position state changed");
+    struct zmk_position_state_changed *ev = as_zmk_position_state_changed(eh);
+    if (ev != NULL) {
```
okke-formsma, line 207: if you swap this around and return if `ev == NULL` you don't need as much indentations.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,270 @@
...
+        for (int i = 0; i < ZMK_BHV_TAP_DANCE_MAX_HELD; i++) {
+            struct active_tap_dance *tap_dance = &active_tap_dances[i];
+            if (tap_dance->position == ZMK_BHV_TAP_DANCE_POSITION_FREE) {
+                continue;
+            } else {
```
okke-formsma, line 212: you can get rid of the else here, because of the continue in the if-clause.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,270 @@
...
+}
+
+static int on_tap_dance_binding_pressed(struct zmk_behavior_binding *binding,
+                                        struct zmk_behavior_binding_event event) {
+    LOG_DBG("On Binding Pressed");
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_tap_dance_config *cfg = dev->config;
+    struct active_tap_dance *tap_dance;
+    tap_dance = find_tap_dance(event.position);
+    if (tap_dance != NULL) {
```
okke-formsma, line 132: if you reverse the if an else statements, you don't need the `else`.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,270 @@
...
+}
+
+static inline int press_tap_dance_behavior(struct active_tap_dance *tap_dance, int64_t timestamp) {
+    LOG_DBG("Press Tap Dance Behavior");
+    struct zmk_behavior_binding binding = tap_dance->config->behaviors[(tap_dance->counter) - 1];
+    struct zmk_behavior_binding_event event = {
+        .position = tap_dance->position,
+        .timestamp = timestamp,
+    };
+    if (tap_dance->counter <= tap_dance->config->behavior_count) {
```
okke-formsma, line 100: you can also trigger the 'final' tap dance item as soon as the tap dance is tapped N times.

 For example, a tap dance with 3 items would trigger immediately at the 3rd press.

okke-formsma (changes requested):
Thanks!

Please add tests to validate everything works as expected. Also, I think the code can be simplified a bit by changing some of the if-statement structures;

```
if(something != NULL) {
   do something
   and more 
} else {
   something else
   return;
}
```
can be rewritten to the simpler 
```
if(something != NULL) {
   something else
   return;
}
do something
and more 
```

PR #812, opened 2021-05-31
Add delay on initializing external power control for BlueMicro840

AnthonyAmanse:
This commits adds a delay of 100ms on initializing the external
power control driver. Previously, OLED's i2c driver is failing to
initialize the display. This commit fixes that issue.

Signed-off-by: Anthony Amanse <ghieamanse@gmail.com>

### DIFF
--- a/app/boards/arm/bluemicro840/bluemicro840_v1.dts
+++ b/app/boards/arm/bluemicro840/bluemicro840_v1.dts
@@ -29,6 +29,7 @@
 	ext-power {
 		compatible = "zmk,ext-power-generic";
 		label = "EXT_POWER";
+		init-delay-ms = <20>;
 		control-gpios = <&gpio0 12 GPIO_ACTIVE_HIGH>;
 	};
 


petejohanson:
We've seen this with other boards. Is the full 100ms really needed? Tested with shorter and not working consistently?

AnthonyAmanse:
I can do some more testing on it and see what's the least amount of delay needed

AnthonyAmanse:
@petejohanson 12ms seems to be the lowest. Tried resetting a few times and OLED is being initialized properly. 11 doesn't work at all even with a few resets

app/boards/arm/bluemicro840/bluemicro840_v1.dts
```diff
@@ -29,6 +29,7 @@
 	ext-power {
 		compatible = "zmk,ext-power-generic";
 		label = "EXT_POWER";
+		init-delay-ms = <12>;
```
petejohanson, line 32: Just to be safe, can we bump to 15 or 20?

AnthonyAmanse: Sure, bumped and rebased from main

petejohanson (approved):
Thanks!

PR #817, opened 2021-06-06
fix(ble): Ignore out of range profiles

joelspadin:
Don't allow selecting a BLE profile that is out of range to avoid reading/writing past the end of the profiles array.


petejohanson (approved):
Thanks!

PR #820, opened 2021-06-08
fix(combos): Avoid spuriously triggering combo (and stuck key down) when last key of long combo doesn't match

### LINKED ISSUE #818: Long combo incorrectly triggers when only last key mismatches
Filing after [discussion on ZMK Discord](https://discord.com/channels/719497620560543766/719909884769992755/851554113278836798).

I wanted Lower+Raise+Backspace to toggle a layer, so I put together a [keymap change](https://github.com/bcat/zmk-config/commit/0f0b57ff233a56345426d3139d3d26906ecd7f18). But I think I've found some sort of race condition between combo processing and momentary layer keys, because this change caused my keyboard (Corne with nice!nanos) to lock up if I pressed Lower+Raise+(AnythingButBackspace) within the 1000 ms combo timeout I set.

I tried to make a [unit test](https://github.com/bcat/zmk/commit/d83ef88640404fe4ec2ef8fa0ecbf18178d81258), with a (close to?) minimal repro. Unless I just messed up the test, I would expect 1) the combo not to trigger and 2) `F` (key at `(1, 1)` on `layer_2`) to have a press and release event.

Instead, I get a press (and no release) for `Z`, showing 1) the combo triggered even though not all its keys were pressed, and 2) there seems to be a key getting "stuck" down (no release). Here's what west test says:

```
$ west test tests/combo/press-release-long-combo-layer-overlap
Running tests/combo/press-release-long-combo-layer-overlap:
--- tests/combo/press-release-long-combo-layer-overlap/keycode_events.snapshot  2021-06-07 17:19:01.640000000 -0500
+++ build/tests/combo/press-release-long-combo-layer-overlap/keycode_events.log 2021-06-07 17:38:58.490000000 -0500
@@ -1,2 +1 @@
-pressed: usage_page 0x07 keycode 0x09 implicit_mods 0x00 explicit_mods 0x00
-released: usage_page 0x07 keycode 0x09 implicit_mods 0x00 explicit_mods 0x00
+pressed: usage_page 0x07 keycode 0x1d implicit_mods 0x00 explicit_mods 0x00
```

I don't understand why the combo seems to be triggering at all since the keys held down should be `{0, 1, 3}`, whereas the combo specifies `{0, 1, 2}`. It makes me suspect I just screwed up the test case, but the test passes if I remove the combo from the keymap, so I'm not sure.

**Edit:** I've simplified the test case further. Please see https://github.com/zmkfirmware/zmk/issues/818#issuecomment-856328283

bcat:
Fixes #818.

The current combo completion check only makes sure the last key in the combo is set. This works when the combo is typed correctly initially, or when reraising events in a combo of length two. However, it fails for longer combos since the last event in `pressed_keys` might be set, but the first (or subsequent) event in `pressed_keys` can be `NULL` thanks to `release_pressed_keys`.

I've unit tested this change, and I've also tested it manually in a few hours of usage on my nice!nano (Corne) board.


okke-formsma (approved):
Wow thanks! This looks really good :)

mcrosson:
Agreed, this looks awesome. Thank you for tracking this one down!~

petejohanson (approved):
Thanks!

PR #823, opened 2021-06-09
feat(behaviors): Add caps word (`&caps_word`).

petejohanson:
* Add new `&caps_word` behavior that acts like caps lock, but
  releases automatically when any "break" keycode is pressed.


weilbith:
Amazing!!! 🥳 
Thanks for tackling this feature! 🙏🏾

petejohanson:
Found one bug last night when testing on my actual keyboard, will push fixes in a bit.

app/dts/behaviors/caps_word.dtsi
```diff
@@ -0,0 +1,20 @@
...
+
+#include <dt-bindings/zmk/keys.h>
+
+/ {
+	behaviors {
+		/omit-if-no-ref/ caps_word: behavior_caps_word {
+			compatible = "zmk,behavior-caps-word";
+			label = "CAPS_WORD";
+			#binding-cells = <0>;
+			breaks = <SPACE COMMA DOT>;
```
petejohanson, line 15: Is this default list of space, comma, and dot reasonable? Others we should add?

petejohanson (commented):
The main item that I would like to probably adjust is the implementation which currently will also apply shift to non-alpha HID usage/keycodes, which means you can't use it to type sommething like `6SIGMA` as it will come out as `^SIGMA` instead.

I probably will refine this to instead of setting an explicit modifier when activated, intercept keycode events and enhance them with additional implicit modifiers if they fall in one of the "included usage ranges" configured.

app/dts/behaviors/caps_word.dtsi
weilbith: colon, semicolon, equals, line break, tab, ...
Maybe its easier to make it a whitelist? It would also address your above problem. So just note down the keycodes which should get the modifier applied until something else gets typed. I think this list will be always shorter as I barely see what else except alpha keycodes and underscore, dash that should be used here to write a caps word.

dxmh: > Is this default list of space, comma, and dot reasonable? Others we should add?

Escape is a must for vim users!

dxmh:
> The main item that I would like to probably adjust is the implementation which currently will also apply shift to non-alpha HID usage/keycodes, which means you can't use it to type sommething like `6SIGMA` as it will come out as `^SIGMA` instead.

Could this be resolved by applying capslock instead of shift? It would certainly be nice to be able to type an uppercase without shifting non-alpha characters (for the example you gave, as well many many others e.g. `USB-C`).

petejohanson:
> > The main item that I would like to probably adjust is the implementation which currently will also apply shift to non-alpha HID usage/keycodes, which means you can't use it to type sommething like `6SIGMA` as it will come out as `^SIGMA` instead.
> 
> Could this be resolved by applying capslock instead of shift? It would certainly be nice to be able to type an uppercase word without shifting non-alpha characters (for the example you gave, as well many many others e.g. `USB-C`).
> 
> Capslock also has the added benefit of illuminating capslock LEDs and similar indicators. :)

Yes! I would actually long term prefer to use that strategy, but doing so requires we finally implement the HID endpoint to have the OS send us the capslock state, so we can be sure to toggle on/off the caps lock as needed properly.

docs/docs/behaviors/caps-word.md
```diff
@@ -0,0 +1,74 @@
...
+
+If you want to use multiple caps breaks with different codes to break the caps, you can add additional caps words instances to use in your keymap:
+
+```
+/ {
+    prog_caps: behavior_prog_caps_word {
+        compatible = "zmk,behavior-caps-word";
+        label = "PROG_CAPS";
+        #binding-cells = <0>;
+        breaks = <SPACE>;
```
okke-formsma, line 67: I think this example would benefit from more breaks like <SPACE ENTER>

app/dts/behaviors/caps_word.dtsi
okke-formsma: I think ENTER and TAB should also be in the default list

theol0403:
I've never used ZMK, but I use a custom version of caps word on my qmk board. However, this implementation is missing a few features I use, if its helpful for me to share:
- the non-breaking keycodes are a whitelist containing A-Z, bspc, ent, del, unds, and mins (for CAPS_WORD or CAPS-WORD). 
- This means everything else (symbols, arrows, ect) will end caps word, which is useful
- There is also a list of what will be shifted, which is only A-Z. This means caps word does not affect symbols, quotes, ect. I only want caps word to act as a shift for alphas, since I have a specific spot for each symbol on a layer, I don't want caps word to change what the symbols do. Also, then I can type CAPS-WORD without the dash being shifted, as mentioned in this thread.
- Anything with a modifier applied to it will also end caps word. This lets me type THIs without needing to toggle caps word, I only manually shift the "I" and it ends caps word

petejohanson:
> I've never used ZMK, but I use a custom version of caps word on my qmk board. However, this implementation is missing a few features I use, if its helpful for me to share:

Thanks!

> - the non-breaking keycodes are a whitelist containing A-Z, bspc, ent, del, unds, and mins (for CAPS_WORD or CAPS-WORD). 

I'm leaning in the whitelist direction too, given the other feedback here, and my own continued brainstorming.

> - This means everything else (symbols, arrows, ect) will end caps word, which is useful

I can see the value in enter and backspace not breaking, now that you mention it. Why not left/right arrow as well?

> - There is also a list of what will be shifted, which is only A-Z. This means caps word does not affect symbols, quotes, ect. I only want caps word to act as a shift for alphas, since I have a specific spot for each symbol on a layer, I don't want caps word to change what the symbols do. Also, then I can type CAPS-WORD without the dash being shifted, as mentioned in this thread.

Yeah, that's definitely a flaw with the current implementation. Have you ever used the OS caps lock, instead of managing the shift state per character? That seems more reliable, than re-implementing the caps lock "caps list" yourself, and applying things manually.

> - Anything with a modifier applied to it will also end caps word. This lets me type THIs without needing to toggle caps word, I only manually shift the "I" and it ends caps word

Interesting. Is the use case here not interrupting your flow when typing something like "URLs"? I hadn't considered that as a feature.

theol0403:
> I can see the value in enter and backspace not breaking, now that you mention it. Why not left/right arrow as well?

I just want to minimize the chance of caps word still being on when I go and edit somewhere else. If I use arrows to navigate somewhere else in my text, my "caps word" is likely complete.

> That seems more reliable, than re-implementing the caps lock "caps list" yourself, and applying things manually.

I don't know about being more reliable, but I do think it's probably the better option.

> Interesting. Is the use case here not interrupting your flow when typing something like "URLs"? I hadn't considered that as a feature.

Yeah, or VSCode. 
`<caps_word>vs<shift-c>ode`

theol0403:
> Interesting. Is the use case here not interrupting your flow when typing something like "URLs"? I hadn't considered that as a feature.

I also remembered another use for this. If caps word is on, and I want to use a ctrl/gui modifier like ctrl-a, I want caps lock to turn off (I'm done typing the word), and I don't want it to send ctrl-shift-a.

weilbith:
> Anything with a modifier applied to it will also end caps word. This lets me type THIs without needing to toggle caps word, I only manually shift the "I" and it ends caps word

This feature sounds great! Nice idea.

> Also, some people have it in their implementations that when caps word is on, escape is intercepted and is used to turn off caps word without sending escape to the OS.

As a Vim-Everywhere-User: please don't. Or at least make it an option the user can toggle.

theol0403:
> As a Vim-Everywhere-User: please don't. Or at least make it an option the user can toggle.

I hear you 😛. I personally have it so that it both stops caps word and sends escape, but it might actually be better to choose one or the other.

weilbith:
> I hear you stuck_out_tongue. I personally have it so that it both stops caps word and sends escape, but it might actually be better to choose one or the other.

That would be my favorite. But actually that is the "normal" case again where escape is just not witelisted, cancels caps word and gets send as normal. Right?

petejohanson:
Ok, I just pushed a reworked version of this:

* Moved to `continue-list` that is the set of keycodes *in addition to the alphanumerics* that will preserve the caps word being on.
* Moved to implicit modifiers applied to the alpha keycodes only, so that `ROW6` will work as expected.

Not added: escape handling specially.

Test please?

petejohanson:
Happy w/ the implementation, had a few users report success. Merging.

PR #827, opened 2021-06-09
BT_RST | Clear bond information between the keyboard and host for all profiles. 

liuggio:


app/src/ble.c
```diff
@@ -548,6 +548,26 @@ static void zmk_ble_ready(int err) {
...
+        char setting_name[15];
+        sprintf(setting_name, "ble/profiles/%d", i);
+
+        int err = settings_delete(setting_name);
+        if (err) {
+            LOG_ERR("Failed to delete setting: %d", err);
+            ret -= 1;
+        }
+    }
+    return ret;
```
liuggio, line 568: do we need a `update_advertising();` here?

docs/docs/behaviors/bluetooth.md
```diff
@@ -36,6 +36,8 @@ Here is a table describing the command for each define:
 | `BT_NXT` | Switch to the next profile, cycling through to the first one when the end is reached.                                                                     |
 | `BT_PRV` | Switch to the previous profile, cycling through to the last one when the beginning is reached.                                                            |
 | `BT_SEL` | Select the 0-indexed profile by number. Please note: this definition must include a number as an argument in the keymap to work correctly. eg. `BT_SEL 0` |
+| `BT_CAL` | Clear bond information between the keyboard and host for all profiles.                                                                                    |
```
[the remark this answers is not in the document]
liuggio: thank you, i renamed to `BT_RST`

app/include/zmk/ble.h
```diff
@@ -13,6 +13,7 @@ int zmk_ble_clear_bonds();
 int zmk_ble_prof_next();
 int zmk_ble_prof_prev();
 int zmk_ble_prof_select(uint8_t index);
+int zmk_ble_reset_clear_all_bonds();
```
petejohanson, line 16: I'd rather a naming like:

```suggestion
int zmk_ble_clear_all_profile_bonds();
```

And move this above the existing `zmk_ble_clear_bonds()`.

app/src/ble.c
```diff
@@ -548,6 +548,28 @@ static void zmk_ble_ready(int err) {
     update_advertising();
 }
 
+int zmk_ble_reset_clear_all_bonds() {
+    LOG_WRN("Clearing all existing BLE bond information from the keyboard");
+
+    for (int i = 0; i < 10; i++) {
```
petejohanson, line 554: This really isn't necessary, this was a holdover from previously using multiple "IDs" for profiles. Can just be `bt_unpair(BT_ID_DEFAULT, null)` without the loop.

app/src/ble.c
```diff
@@ -548,6 +548,28 @@ static void zmk_ble_ready(int err) {
...
+        bt_unpair(i, NULL);
```
petejohanson, line 555: For *splits*, this *won't* work well for users. In particular, if we're invoking this from the keymap, we'll end up unpairing the other half in addition to the host bonds.

I think we need to instead do this inside the loop over the profiles below, and unpair the explicit address found in each profile before deleting the setting, so we don't mess w/ the split bonds when we do this.

And.... when we do this, we need to be careful to actually clear *everything* for the "settings reset Kconfig" setting though, since that *should* be clearing the split bonds too.

PR #835, opened 2021-06-12
New Feature: required keys for tap-hold behaviors

jmding8:
Adds the ability to for users to configure hold-tap behaviors so that the only resolve to hold behaviors if specified keys are pressed. Let's say that we set up a keymap this way:
```
keymap {
        compatible = "zmk,keymap";
        base_layer {
            bindings = <&cmt LSFT Q    &kp W    &kp E>
...
```

Where cmt is defined:
```
behaviors {
        cmt: conditional_mod_tap {
			label = "TAP_PREFERRED_MOD_TAP";
			#binding-cells = <2>;
			flavor = "tap-preferred";
			tapping_term_ms = <1>;
			bindings = <&kp>, <&kp>;
			required_keys_for_hold = <1>;
		};
    };
```

These are the expected behaviors:

key_0_down, key_1_down, key_1_up, key_0_up => "W"
Because the sequence is executed in longer than 1 ms, and key_1 is pressed when the hold/tap behavior is decided, the hold/tap behavior is resolved into a hold. This is the same behavior as would occur with the current logic as well.

0_down, 2_down, 2_up, 0_up => "qe"
Because the sequence is executed in longer than 1 ms, but key_1 is NOT pressed when the hold/tap behavior is decided, the hold/tap behavior is resolved into a tap. This differs from the behavior that would occur with the current logic, which would not have a requirement that key_1 be pressed to resolve into a hold behavior, and would therefore emit "E".

This is extremely useful for my situation:
* I have replaced Caps Lock on my keymap with ```&lt NUM CLCK```, such that the sequence <caps_down, j_down, j_up, caps_up> produces "4"
* Due to the position of the Caps Lock, it is difficult to press and hold LSFT while also holding Caps Lock down, so it becomes difficult to produce "$"
* To make it easier to apply a LSFT modifier while also holding Caps Lock down, I have replaced the A key on my keymap with ```&mt LSFT A```
* However, mod-tap behaviors can be difficult to habituate due to the tight timing constraints
* So instead of using a raw mod-tap, I condition the mod-tap A so that it can only produce a HOLD behavior if and only if the Caps Lock key is depressed at the time that the hold-tap is decided.


okke-formsma:
I like the general idea, but maybe we should change the meaning of the list from "required_keys_for_hold" to "hold-keys". Any combinations of a hold key + a key (position) from this list can trigger a hold, but keys not in the list won't.

That would enable us to do "left-side hold taps" and "right side hold taps", that would only trigger for keys on the opposite side of the board for example.

The current implementation requires all keys on the list to be pressed, for which I can't really find a usecase.

app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
```diff
@@ -30,3 +30,7 @@ properties:
       - "tap-preferred"
   retro-tap:
     type: boolean
+  required_keys_for_hold:
```
okke-formsma, line 33: `required-keys-for-hold`

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -90,6 +92,8 @@ struct last_tapped {
 
 struct last_tapped last_tapped;
 
+bool is_key_held[ZMK_KEYMAP_LEN] = {0};
```
okke-formsma, line 95: maybe rename to `key_position_pressed`?

jmding8:
I played around with this more, and totally agree. Instead of AND on all the keys in the list, OR is a lot more useful. As for naming, hold-behavior-enabler-keys?

app/src/behaviors/behavior_hold_tap.c
jmding8: how about is_key_position_pressed? I do like the "is_" prefix, IMO it helps imply bool

okke-formsma:
Great!

I think there's a bug in the current implementation. If one of the 'hold-enabled keys' is pressed _before_ the hold-tap, the hold will be triggered even though the actual key being pressed is not a hold-enabled key.

Let's say 1_down, 0_down, 2_down, 2_up, 0_up => "wE" instead of "wqe".

As a possible fix, we could keep the `is_key_held` array in the `active_hold_tap` struct (or rather, this should be the `first-key-pressed` int now instead of an array because we're only interested in the first key that was pressed after the hold-tap was enabled).

Please write some test to cover the scenarios you described above, and this one.

jmding8:
The original intent was indeed to produce "wE", and not "wqe". It was intended to only consider the pressed state, and be agnostic to the order of pressing. BTW, this is why the variable was named hold_enableR_keys, rather than hold_enableD_keys

I think it might be better to be order agnostic when using conditional hold/tap for situations like home-row mods. For example
```
/ {
	behaviors {
		cht: conditional_hold_tap {
			label = "cond_hold_tap";
			#binding-cells = <2>;
			flavor = "hold-preferred";
			tapping-term-ms = <10>;
			bindings = <&kp>, <&kp>;
			required_keys_for_hold = <0 1>;
		};
	};

	keymap {
		compatible = "zmk,keymap";
		label ="Default keymap";

		default_layer {
			bindings = <
				&cht LEFT_CONTROL Q    &cht LEFT_SHIFT W    & kp E
			>;
		};
	};
};
```

Then the sequence
0_down, 1_down, 2_down, 2_up, 0_up, 1_up => "Ctrl + Shift + e"
The 1_down resolves the 0 key into it's hold, Ctrl behavior. And because it is order agnostic, the 0_down also conditions the 1 key such that it is allowed to resolve into it's hold, Shift behavior.

Let me know what you think? These are tricky for me to reason about.

okke-formsma:
A concern I have with the proposal as it stands now, is that there are two distinct scenarios that are covered. I'd like to keep the set of behaviors and options we have as _orthogonal_ as possible and combinable (e.g., there should be one way to do something, and many 'simple' behaviors build up to a more complex whole)

Scenario A: conditional key first
* a key is pressed
* a conditional hold-tap is pressed
* the conditional hold-tap is decided

Scenario B: hold-tap first
* a conditional hold-tap is pressed
* a key is pressed
* the conditional hold-tap is decided

Scenario A could be covered by making the first key a layer key (if it needs to have other behaviors too that could be arranged with a macro-like behavior).

Scenario B, where it depends which key is presse to decide on the hold-tap, would be something that we can not do yet. It's also a often requested feature, for hold-taps that only hold for the other side of a split board for example.

jmding8:
Updated such that keypresses are order dependent.
Now, the hold-behavior is only enabled when the first other-key to be pressed is one of the configured, enabler keys. If the enabler key is pressed before the hold-tap key, it "does not count" and the hold behavior is not enabled.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -217,16 +230,67 @@ static void clear_hold_tap(struct active_hold_tap *hold_tap) {
     hold_tap->work_is_cancelled = false;
 }
 
-static void decide_balanced(struct active_hold_tap *hold_tap, enum decision_moment event) {
-    switch (event) {
+static bool
+does_position_of_first_other_keypress_permit_hold_behavior(struct active_hold_tap *hold_tap) {
```
okke-formsma, line 234: I would expect this function to take a `struct decision_trigger_event` as input. That way, the `position_of_first_other_key_pressed` doesn't have to be stored on the `hold_tap`, allowing us to dispose of the `store_position_if_is_first_other_key_pressed` method.

Something like`is_hold_enabler_keypress` may be a more concise name.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -217,16 +230,67 @@ static void clear_hold_tap(struct active_hold_tap *hold_tap) {
...
-        hold_tap->status = STATUS_HOLD_INTERRUPT;
+        if (does_position_of_first_other_keypress_permit_hold_behavior(hold_tap)) {
+            hold_tap->status = STATUS_HOLD_INTERRUPT;
+        } else {
+            hold_tap->status = STATUS_TAP;
+        }
         return;
     case HT_TIMER_EVENT:
-        hold_tap->status = STATUS_HOLD_TIMER;
+        if (does_position_of_first_other_keypress_permit_hold_behavior(hold_tap)) {
```
okke-formsma, line 289: We don't need to check this on the timer event, because the 'other keypress event' will already have set `STATUS_HOLD_INTERRUPT` if `does_position_of_first_other_keypress_permit_hold_behavior` is True.

(thus, for this check the global storage of `hold_tap->position_of_first_other_key_pressed` is not required either)

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -494,6 +579,9 @@ static const struct behavior_driver_api behavior_hold_tap_driver_api = {
 static int position_state_changed_listener(const zmk_event_t *eh) {
     struct zmk_position_state_changed *ev = as_zmk_position_state_changed(eh);
 
+    // Update key position state.
+    is_key_position_pressed[ev->position] = ev->state;
```
okke-formsma, line 583: this is dead code now

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -360,7 +438,7 @@ static int release_binding(struct active_hold_tap *hold_tap) {
 }
 
 static void decide_hold_tap(struct active_hold_tap *hold_tap,
-                            enum decision_moment decision_moment) {
+                            struct decision_trigger_event *decision_trigger_event) {
```
okke-formsma, line 441: I think I prefer passing three parameters into this method instead of the struct. _If_ we take a struct as input, I think we should include the `*hold_tap` in it too.

okke-formsma (changes requested):
in general, this looks good. I think we can simplify somewhat.

Haven't looked into the tests and documentation yet.

jmding8:
thanks! Based on your feedback, I was able to find an implementation that doesn't require us to store anything additional in the active_hold_tap struct, and simply sets active_hold_tap.status to STATUS_TAP under the appropriate conditions.

jmding8:
Tests pass:

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -370,19 +374,40 @@ static void decide_hold_tap(struct active_hold_tap *hold_tap,
...
-    switch (hold_tap->config->flavor) {
-    case FLAVOR_HOLD_PREFERRED:
-        decide_hold_preferred(hold_tap, decision_moment);
-    case FLAVOR_BALANCED:
-        decide_balanced(hold_tap, decision_moment);
-    case FLAVOR_TAP_PREFERRED:
-        decide_tap_preferred(hold_tap, decision_moment);
+    // For conditional hold-tap behaviors that have failed to meet the required prerequisites
+    // for a hold decision, force a tap decision.
+    if (decision_moment == HT_OTHER_KEY_DOWN && hold_tap->config->hold_enabler_keys_len > 0) {
```
okke-formsma, line 379: I don't think this if statement is correct for all 'flavors'. For example, the `balanced` never decides on HT_OTHER_KEY_DOWN; it decides on HT_OTHER_KEY_UP. I think we should put this code into a function `is_hold_enabled_key(int32 other_key_decision, struct active_hold_tap *hold_tap)`, and call it from each of the `decide_*` methods.

Somewhat tricky would be the behavior of `tap-preferred`. I think it should act like this:

* if the timer has not run out (e.g., we have not seen `HT_TIMER_EVENT`, always trigger tap. (doesn't matter which key is pressed)
* if the timer did run out (the HT_TIMER_EVENT was seen, maybe setting STATUS_UNDECIDED_TIMER_SEEN), trigger HOLD if the key is a hold_enabler_key, otherwise trigger TAP

For balanced, the decision_moment to check for hold_enabler_keys would be on HT_OTHER_KEY_UP.

jmding8: you're right, I'll fix. I'll also make the testing more robust to catch this.

jmding8: I think we'll still need to store position_of_first_other_key_pressed in the active_hold_tap struct, because in tap-preferred situation 4a, I don't see any other way of verifying whether the first other key that was pressed was an enabler key, especially if a second key is pressed before tapping term expires

image

jmding8:
rewrote the feature again. Did a couple things:

1) renamed it "positional hold-tap", to be more specific as to what conditions enable/disable
2) copy-pasted (most of) the tests for balanced twice, once positionally enabled, and once positionally disabled. If this approach looks good to you, I'll go ahead and do the same for the hold-preferred and tap-preferred tests as well.
3) rewrote the feature. Now it only overwrites hold behaviors at the time of decision (and prior to execution)

docs/docs/behaviors/hold-tap.md
```diff
@@ -95,6 +95,57 @@ This example configures a hold-tap that works well for homerow mods:
...
+
+  * Meanwhile, the sequence `(pht_down, E_down, E_up, W_down, W_up pht_up)` produces `qew`. The positional hold-tap is **NOT** permitted to produce a hold behavior, because the first key pressed after the hold-tap key (i.e. the E key, in position 2) is **NOT** one of the `hold-trigger-key-positions`.
+
+* Positional hold-taps can be useful with home-row modifiers. By using a positional hold-tap behavior for home-row modifiers on the left hand, and setting `hold-trigger-key-positions` to the keys under the right hand, positional hold-tap will only allow hold behaviors to occur with cross-hand keypresses.
+
+* Note that for regular hold-tap behaviors a shorter `tapping-term` encourages hold decisions. However the opposite is true for positional hold-tap behaviors, where a shorter `tapping-term` actually encourages tap decisions. This is because when the `tapping-term` expires, this triggers the behavior to decide as either a tap or a hold. But if the user has not yet had time to press one of the `hold-trigger-key-positions`, then with positional hold-tap the behavior will decide as a tap.
```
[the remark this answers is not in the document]
okke-formsma: Yeah we should probably check the position-hold-keys for both STATUS_HOLD_INTERRUPT and STATUS_HOLD_TIMER ....

app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
```diff
@@ -30,3 +30,7 @@ properties:
       - "tap-preferred"
   retro-tap:
     type: boolean
+  hold-trigger-key-positions:
```
jmding8, line 33: Is there a way I can create a compile-time constraints?
1. if `flavor = tap_positionally_preferred`, then `hold-trigger-key-positions` is included
2. if `hold-trigger-key-positions` is included, then `flavor = tap_positionally_preferred`

bortoz: You can but I don't recommend it.
I suggest adding something like this inside `KP_INST` for case 1:
```C
BUILD_ASSERT(!(DT_ENUM_IDX(DT_DRV_INST(n), flavor) == 3
                 && DT_INST_PROP_LEN(n, hold_trigger_key_positions) > 0),
             "flavor is 'tap-positionally-preferred' but 'hold-trigger-key-positions' is empty")
```
And something similar for case 2.

jmding8: why do you recommend against? given that this configuration option should be used if and only if the flavor is tap-positionally-preferred, it seems like it'd be good to have compile-time enforcement of that assumption?

bortoz: I think `BUILD_ASSERT` is a the best way to do that.
There are a few ways to get these checks done in the devicetree but they are much more complicated and I don't recommend them as they may break backward compatibility.

okke-formsma:
LGTM

@petejohanson ready from merge as far as I'm concerned.

jmding8:
not sure why `integration_test` are failing, they succeed on my machine?

petejohanson (approved):
Thanks, and thanks for the test coverage!

Issue #718, opened 2021-03-04
Peripheral for split randomly disconnects and doesn't auto-reconnect.


Goldie643:
Board(s): Nice!Nano
Repo: https://github.com/Goldie643/zmk-config
Shield: mDactyl (Dactyl Manuform Mini)

At seemingly random intervals, approximately once every hour or so, the right hand of the board will disconnect and not provide any output. The left hand continues to work, but to re-enable the right hand the board needs to be re-paired by resetting both sides at the same time. Restarting the LH or RH separately doesn't fix the issue.

I enabled USB logging and kept it running in the background. I grabbed these logs right after it disconnected.

Left hand
```
[00:40:02.491,729] <dbg> zmk.split_central_disconnected: Disconnected: d6:39:f3:4d:8a:08 (random) (reason 8)
[00:40:02.492,065] <dbg> zmk.start_scan: Scanning successfully started
[00:40:02.492,309] <dbg> zmk.disconnected: Disconnected from d6:39:f3:4d:8a:08 (random) (reason 0x08)
[00:40:02.492,340] <dbg> zmk.update_advertising: advertising from 0 to 2
[00:40:02.492,340] <err> zmk: Advertising failed to start (err -69)
[00:40:02.556,121] <dbg> zmk.split_central_device_found: [DEVICE]: 4d:5c:dc:3f:79:0c (random), AD evt type 2, AD data len 28, RSSI -53
[00:40:02.635,131] <dbg> zmk.split_central_device_found: [DEVICE]: 6c:5b:74:58:4f:02 (random), AD evt type 3, AD data len 31, RSSI -42
[00:40:02.863,342] <dbg> zmk.split_central_device_found: [DEVICE]: 44:f9:07:4a:b8:e0 (random), AD evt type 0, AD data len 15, RSSI -49
[00:40:02.863,342] <dbg> zmk.split_central_eir_found: [AD]: 1 data_len 1
[00:40:02.863,372] <dbg> zmk.split_central_eir_found: [AD]: 2 data_len 2
[00:40:02.863,372] <dbg> zmk.split_central_eir_found: [AD]: 255 data_len 6
[00:40:03.062,133] <dbg> zmk.split_central_device_found: [DEVICE]: 51:51:51:b6:6f:93 (random), AD evt type 2, AD data len 28, RSSI -58
[...]

Issue #609, opened 2021-01-15
Bluetooth central does not validate peripheral MAC address


xudongzheng:
The peripheral address is read from settings in `src/ble.c` but isn't checked against in `src/split/bluetooth/central.c`

xudongzheng:
This was resolved in #836.

Issue #314, opened 2020-10-31
Add support for using external dongle with split keyboard


xudongzheng:

innovaker:
The suspense is killing me.

xudongzheng:
Most of it has been implemented in `app/src/split/bluetooth/central.c` on https://github.com/ergoblue/zmk/commit/200f1452797c613ba5212a0a20c2f58bedabe516. I wrote that to run alongside my Nordic SDK stuff on the peripherals but should be pretty straightforward to use with a ZMK peripheral with minor changes:

- `static struct bt_uuid *service_uuid = BT_UUID_HIDS;` and `static struct bt_uuid *characteristic_uuid = BT_UUID_HIDS_REPORT;` should reflect the custom service
- removing `position = zmk_matrix_transform_row_column_to_position(position / 14, position % 14);`
- use the current `split_central_eir_found()` and identify by UUID instead of by name.

A couple of things that I think are necessary before a pull request:

- Change to central/peripheral identical logic #318 
- Something's been going wrong on the ZMK peripheral that kind of stops notifying after central disconnects/reconnects a few times. For this reason I haven't been able to actually test with a ZMK peripheral.
- Some way of defining "number of peripherals to try to connect to" and a decision on whether they should be identified by name
- Overall code cleanup

luigi311:
Im assuming we will have no bongo cat when using dongle mode based on this comment by @petejohanson about only the centeral side having WPM access or will the logic rework fix that?
https://www.reddit.com/r/ErgoMechKeyboards/comments/osr9zb/comment/i4qjb0j/?utm_source=share&utm_medium=web2x&context=3

xudongzheng:
> Im assuming we will have no bongo cat when using dongle mode based on this comment by @petejohanson about only the centeral side having WPM access or will the logic rework fix that?

The pull request just makes it possible to have multiple peripherals instead of just one. Any functionality that is only available on a central will still only be available on a central.

xudongzheng:
This is implemented in #836.

Issue #281, opened 2020-10-14
Periodic "stuck" keys from peripheral sides of splits


petejohanson:
Periodically, we are encountering issues w/ keys being stuck in a "held" state and not being released. This seems to *only* be from keys on the peripheral side of split keyboards. The suspicion is something along the lines of:

* Peripheral side notifies key position state pressed.
* HID report sent by central to host w/ keycode held.
* Peripheral side notifies of key position released, but that is dropped/lost by the central side.
* The HID report is never updated to release the keycode, so the host starts repeating the key infinitely, until the connection to the keyboard is lost.

The suggested fix would be to switch the peripheral to send "full state" updates, not just release/press events. If a certain notification was dropped, the *next* one would end up releasing the key position in question.

joelspadin:
I haven't checked if ZMK supports more than 256 keys, but assuming it does, https://en.wikipedia.org/wiki/Variable-length_quantity would be a good way to keep most reasonable keyboards' key positions down to one byte without limiting the number of keys you support.

innovaker:
@joelspadin, it's an interesting idea!  I'm not sure if we'll ever need to support nodes with more than 256 keys tho?

@petejohanson, were you initially thinking of bit flags or an integer array?

It might also need to handle outlier scenarios such as when the final key the user types is the one that gets lost.  That might require a timeout on the central combined with either:
a) requesting an update, or
b) mandating repeat/regular state updates for held keys

It reminds me of some of the strategies discussed in the ZeroMQ guide.

Of course, without nailing down the series of events, I guess we're prodding in the dark.

joelspadin:
I would guess that an integer array would result in the smaller average packet size, but bit flags would work well if there's any benefit to having a fixed packet size. Still, either way the packets are going to be tiny unless you're using an integer array and manage to hold down every key on a full-size split keyboard.

Slusho64:
I have noticed that the keyboard inputs lag/stutter for a few seconds before this happens. Sometimes both halves of my keyboard will stop inputting at all after some lag/stutter as well. Not clear if the two are related or separate issues. This is on win10 with an Intel bluetooth stack with a big antenna.

leejunseok:
There could be overlap with #221, since the symptoms look comparable (keys stuck held indefinitely until reset) and because I've personally found that #221 only happens on the peripheral side, similar to this issue.

Slusho64:
[...]

PR #836, opened 2021-06-12
feat(split): allow central to connect to multiple peripherals

xudongzheng:
This addresses #314 and #609.

This might also affect #281 and #718.

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -47,12 +47,6 @@ static uint8_t passkey_digit = 0;
 
 #endif /* IS_ENABLED(CONFIG_ZMK_BLE_PASSKEY_ENTRY) */
 
-#if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
-#define PROFILE_COUNT (CONFIG_BT_MAX_PAIRED - 1)
-#else
-#define PROFILE_COUNT CONFIG_BT_MAX_PAIRED
-#endif
-
 enum advertising_type {
     ZMK_ADV_NONE,
     ZMK_ADV_DIR,
@@ -84,7 +78,7 @@ static const struct bt_data zmk_ble_ad[] = {
 
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
 
-static bt_addr_le_t peripheral_addr;
+static bt_addr_le_t peripheral_addrs[ZMK_SPLIT_BLE_PERIPHERAL_COUNT];
 
 #endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL) */
 
@@ -283,9 +277,34 @@ char *zmk_ble_active_profile_name() { return profiles[active_profile].name; }
 
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
 
-void zmk_ble_set_peripheral_addr(bt_addr_le_t *addr) {
-    memcpy(&peripheral_addr, addr, sizeof(bt_addr_le_t));
-    settings_save_one("ble/peripheral_address", addr, sizeof(bt_addr_le_t));
+int zmk_ble_put_peripheral_addr(const bt_addr_le_t *addr) {
+    for (int i = 0; i < ZMK_SPLIT_BLE_PERIPHERAL_COUNT; i++) {
+        // If the address is recognized and already stored in settings, return
+        // index and no additional action is necessary.
+        if (!bt_addr_le_cmp(&peripheral_addrs[i], addr)) {
+            return i;
+        }
+
+        // If the peripheral address slot is open, store new peripheral in the
+        // slot and return index. This compares against BT_ADDR_LE_ANY as that
+        // is the zero value.
+        if (!bt_addr_le_cmp(&peripheral_addrs[i], BT_ADDR_LE_ANY)) {
+            char addr_str[BT_ADDR_LE_STR_LEN];
+            bt_addr_le_to_str(addr, addr_str, sizeof(addr_str));
+            LOG_DBG("Storing peripheral %s in slot %d", addr_str, i);
+            bt_addr_le_copy(&peripheral_addrs[i], addr);
+
+            char setting_name[32];
+            sprintf(setting_name, "ble/peripheral_addresses/%d", i);
+            settings_save_one(setting_name, addr, sizeof(bt_addr_le_t));
+
+            return i;
+        }
+    }
+
+    // The peripheral does not match a known peripheral and there is no
+    // available slot.
+    return -ENOMEM;
 }
 
 #endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL) */
@@ -340,15 +359,20 @@ static int ble_profiles_handle_set(const char *name, size_t len, settings_read_c
         }
     }
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
-    else if (settings_name_steq(name, "peripheral_address", &next) && !next) {
+    else if (settings_name_steq(name, "peripheral_addresses", &next) && next) {
         if (len != sizeof(bt_addr_le_t)) {
             return -EINVAL;
         }
 
-        int err = read_cb(cb_arg, &peripheral_addr, sizeof(bt_addr_le_t));
-        if (err <= 0) {
-            LOG_ERR("Failed to handle peripheral address from settings (err %d)", err);
-            return err;
+        int i = atoi(next);
+        if (i < 0 || i >= ZMK_SPLIT_BLE_PERIPHERAL_COUNT) {
+            LOG_ERR("Failed to store peripheral address in memory");
+        } else {
+            int err = read_cb(cb_arg, &peripheral_addrs[i], sizeof(bt_addr_le_t));
+            if (err <= 0) {
+                LOG_ERR("Failed to handle peripheral address from settings (err %d)", err);
+                return err;
+            }
         }
     }
 #endif
--- a/app/src/split/bluetooth/central.c
+++ b/app/src/split/bluetooth/central.c
@@ -26,7 +26,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/event_manager.h>
 #include <zmk/events/position_state_changed.h>
 
-static int start_scan(void);
+static int start_scanning(void);
 
 #define POSITION_STATE_DATA_LEN 16
 
@@ -48,6 +48,8 @@ struct peripheral_slot {
 };
 
 static struct peripheral_slot peripherals[ZMK_SPLIT_BLE_PERIPHERAL_COUNT];
+
+static bool is_scanning = false;
 
 static const struct bt_uuid_128 split_service_uuid = BT_UUID_INIT_128(ZMK_SPLIT_BT_SERVICE_UUID);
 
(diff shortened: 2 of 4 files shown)


app/src/ble.c
```diff
@@ -337,15 +362,20 @@ static int ble_profiles_handle_set(const char *name, size_t len, settings_read_c
         }
     }
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL)
-    else if (settings_name_steq(name, "peripheral_address", &next) && !next) {
```
DuMaM, line 340: I think this line will solve multiple problems i have currently with my split setup

xudongzheng:
Most likely some additional changes would be needed for settings reset. It's not clear how that should be implemented, perhaps just loop and delete `ble/peripheral_addresses/[0,7]` or just call `nvs_clear()` to handle peripheral addresses and everything else.

app/src/ble.c
```diff
@@ -280,9 +274,34 @@ char *zmk_ble_active_profile_name() { return profiles[active_profile].name; }
...
-void zmk_ble_set_peripheral_addr(bt_addr_le_t *addr) {
-    memcpy(&peripheral_addr, addr, sizeof(bt_addr_le_t));
-    settings_save_one("ble/peripheral_address", addr, sizeof(bt_addr_le_t));
+int zmk_ble_put_peripheral_addr(const bt_addr_le_t *addr) {
+    for (int i = 0; i < ZMK_BLE_SPLIT_PERIPHERAL_COUNT; i++) {
+        // If the address is recognized and already stored in settings, return
+        // index and no additional action is necessary.
+        if (!bt_addr_le_cmp(&peripheral_addrs[i], addr)) {
+            return i;
+        }
```
LSChyi, line 283: What if the recorded peripheral appears after slot `i+1`, and takes `i`th slot?

xudongzheng: I don't quite understand what `i+1` you are referring to.

LSChyi: I wonder if the `addr` is actually recorded at some slot `j`, where `j` > `i`, but somehow the `i`th slot is open, so the `addr` takes away the `i`th slot as well.

This may be a rare case, and `j`th slot anyhow could be recycled for the `addr` should trigger `disconnected` function in normal cases.

xudongzheng: Slot 0 is filled, then slot 1 is filled, and so on. There is no code to delete just a `ble/peripheral_addresses/%d` setting so it's not possible to have `i` open and `j` occupied where `i < j`.

app/src/split/bluetooth/central.c
```diff
@@ -438,14 +447,33 @@ static void split_central_device_found(const bt_addr_le_t *addr, int8_t rssi, ui
...
+    }
+    if (!has_unconnected) {
+        LOG_DBG("All devices are connected, scanning is unnecessary");
+        return 0;
+    }
 
-    err = bt_le_scan_start(BT_LE_SCAN_PASSIVE, split_central_device_found);
+    // Start scanning otherwise.
+    is_scanning = true;
+    int err = bt_le_scan_start(BT_LE_SCAN_PASSIVE, split_central_device_found);
```
petejohanson, line 476: This is fine for now, but ideally a future enhancement will add a backoff on scanning, since multi-peripheral setups are more likely to have one of them only on some of the time.

petejohanson (changes requested):
Couple minor items, but on a whole, LGTM.

xudongzheng:
@petejohanson

Is there a reason for this section of code https://github.com/zmkfirmware/zmk/blob/main/app/src/split/bluetooth/central.c#L403-L411? It's not clear to me why `split_central_process_connection()` would be called there as it's already called from the connect event handler.

I would consider getting rid of the lookup and having https://github.com/slicemk/zmk/commit/44e0e7b158b08d5f014dc813741f72e578eac04f#diff-18de66983adf7197deaf9e826c4087b486beb147e0ab19445d41a85ba7be9fe0R374-R384.

petejohanson:
> @petejohanson
> 
> Is there a reason for this section of code https://github.com/zmkfirmware/zmk/blob/main/app/src/split/bluetooth/central.c#L403-L411? It's not clear to me why `split_central_process_connection()` would be called there as it's already called from the connect event handler.
> 
> I would consider getting rid of the lookup and having [slicemk@44e0e7b#diff-18de66983adf7197deaf9e826c4087b486beb147e0ab19445d41a85ba7be9fe0R374-R384](https://github.com/slicemk/zmk/commit/44e0e7b158b08d5f014dc813741f72e578eac04f#diff-18de66983adf7197deaf9e826c4087b486beb147e0ab19445d41a85ba7be9fe0R374-R384).

IIRC, at least with Zephyr from several versions ago when I first wrote this, some times we would end up having an *existing* connection, after the scanning first discovers the advertising info, we and we needed to handle both scenarios:

1) No connect, create it.
2) Already connected, process the connection.

If we can test this thoroughly and see this no longer happening with the stack, we should simplify this, for sure.

app/src/ble.c
```diff
@@ -337,15 +356,20 @@ static int ble_profiles_handle_set(const char *name, size_t len, settings_read_c
...
         if (len != sizeof(bt_addr_le_t)) {
             return -EINVAL;
         }
 
-        int err = read_cb(cb_arg, &peripheral_addr, sizeof(bt_addr_le_t));
-        if (err <= 0) {
-            LOG_ERR("Failed to handle peripheral address from settings (err %d)", err);
-            return err;
+        int i = atoi(next);
+        if (i >= ZMK_BLE_SPLIT_PERIPHERAL_COUNT) {
```
petejohanson, line 365: To be bullet proof here, should check for `i < 0` as well.

app/src/split/bluetooth/central.c
```diff
@@ -344,9 +347,66 @@ static void split_central_process_connection(struct bt_conn *conn) {
...
+    // Restart scanning if necessary.
+    start_scanning();
+}
+
+static int stop_scanning() {
+    LOG_DBG("Stopping peripheral scanning");
+    is_scanning = false;
+
+    int err = bt_le_scan_stop();
+    if (err) {
```
petejohanson, line 360:
```suggestion
    if (err < 0) {
```

We should be more diligent with these checks... Some Zephyr APIs in the past have been refactored across versions and return positive values for different success states.

app/src/split/bluetooth/central.c
```diff
@@ -344,9 +347,66 @@ static void split_central_process_connection(struct bt_conn *conn) {
...
+
+    return 0;
+}
+
+static bool split_central_eir_found(const bt_addr_le_t *addr) {
+    LOG_DBG("Found the split service");
+
+    // Stop scanning so we can connect to the peripheral device.
+    int err = stop_scanning();
+    if (err) {
```
petejohanson, line 373:
```suggestion
    if (err < 0) {
```

app/src/split/bluetooth/central.c
```diff
@@ -344,9 +347,66 @@ static void split_central_process_connection(struct bt_conn *conn) {
...
+        if (err) {
+            LOG_ERR("Update phy conn failed (err %d)", err);
+        }
+    } else {
+        struct bt_le_conn_param *param = BT_LE_CONN_PARAM(0x0006, 0x0006, 30, 400);
+
+        LOG_DBG("Initiating new connnection");
+
+        err = bt_conn_le_create(addr, BT_CONN_LE_CREATE_CONN, param, &slot->conn);
+        if (err) {
```
petejohanson, line 399:
```suggestion
        if (err < 0) {
```

app/src/split/bluetooth/central.c
```diff
@@ -344,9 +347,66 @@ static void split_central_process_connection(struct bt_conn *conn) {
...
+    }
+
+    struct peripheral_slot *slot = &peripherals[slot_idx];
+
+    slot->conn = bt_conn_lookup_addr_le(BT_ID_DEFAULT, addr);
+    if (slot->conn) {
+        LOG_DBG("Found existing connection");
+        split_central_process_connection(slot->conn);
+        err = bt_conn_le_phy_update(slot->conn, BT_CONN_LE_PHY_PARAM_2M);
+        if (err) {
```
petejohanson, line 390:
```suggestion
        if (err < 0) {
```

petejohanson (changes requested):
Few final code nitpicks. I will try to get to testing this soon.

xudongzheng:
My suggestion is to only handle a new connection when the `connected` callback happens. This should happen exactly once per new connection.

If there is an existing connection, `reserve_peripheral_slot()` should fail because each peripheral will resolve to the same index each time - that's in `zmk_ble_put_peripheral_addr()`.

Will have an updated PR shortly.

app/src/split/bluetooth/central.c
bezmi: I'm exploring this by changing this to,
```c
bt_le_scan_start(&test_scan_params, split_central_device_found)
```
where,
```c
static struct bt_le_scan_param test_scan_params = {
    .type = BT_LE_SCAN_TYPE_PASSIVE,
    .options = BT_LE_SCAN_OPT_FILTER_DUPLICATE,
    .interval = BT_GAP_SCAN_FAST_INTERVAL,
    .window = BT_GAP_SCAN_FAST_WINDOW,
    .timeout = 1500,
    .interval_coded = 0,
    .window_coded = 0,
};
```

Setting any `timeout` > 0 will cause zmk to hang. `.timeout = 0` will work just fine,
```
[00:00:00.770,935] <dbg> zmk: zmk_usb_get_conn_state: state: 1
[00:00:00.770,965] <dbg> zmk: get_selected_endpoint: No endpoints are ready.
[00:00:01.081,146] <inf> usb_hid: Device configured
[00:00:01.081,207] <dbg> zmk: zmk_usb_get_conn_state: state: 3
[00:00:01.081,237] <dbg> zmk: zmk_usb_get_conn_state: state: 3
[00:00:01.081,237] <dbg> zmk: get_selected_endpoint: Only USB is ready.
[00:00:01.081,237] <dbg> zmk: zmk_endpoints_send_report: usage page 0x07
[00:00:01.081,268] <dbg> zmk: zmk_endpoints_send_report: usage page 0x0C
[00:00:01.081,298] <inf> zmk: Endpoint changed: 0
[00:00:01.081,359] <dbg> zmk: destination_connection: Address pointer 0x20008fdf
[00:00:01.081,359] <wrn> zmk: Not sending, no active address for current profile
[00:00:01.081,390] <dbg> zmk: destination_connection: Address pointer 0x20008fdf
[00:00:01.081,390] <wrn> zmk: Not sending, no active address for current profile
```

Any ideas? I was hoping to register scan callbacks where after the 15s timeout, scanning is restarted with new params with a longer interval and window and an indefinite timeout.

bezmi:
I have been testing this (on the latest one as of this comment) and it's working well. @xudongzheng would you please consider pushing on top of the current commit instead of re-writing history with the force pushes? Makes it a little easier to explore the changes and can always be squashed when merging. Also, [see my comment above](https://github.com/zmkfirmware/zmk/pull/836#discussion_r1018572634) RE scan timeout.

xudongzheng:
@bezmi good point, you can see today's changes in https://github.com/zmkfirmware/zmk/commit/d26b0ff8747476620764da27a401aa1083515286

I'm not familiar with the scan timeout. I'd consider explicitly stopping the scanning using a custom timer, which might be necessary regardless for maximum flexibility.

bezmi:
@xudongzheng, I have been testing this PR on my daily driver split keyboard for pretty much as long as it has existed. I often run into a particular latency issue where keypresses from each half will arrive out of order, making it difficult to type quickly. It is very similar in description to #1633. Have you had any such issues and any success in troubleshooting them?

xudongzheng:
@bezmi The first thing to try is placing the dongle right between the two keyboard halves. If it works fine there, it's likely a signal issue and not a firmware issue. You can bump the Bluetooth TX power as is typically recommended for ZMK.

petejohanson:
@xudongzheng Can you please rebase now that your refactor #1815 is in? Code looks good w/ your latest push, will test here then merge, barring any problems appearing. Thanks!

petejohanson:
@xudongzheng See the pre-commit check issue please on this.

xudongzheng:
@petejohanson Oops, somehow missed that. The new commit seems to pass all the commit checks.

petejohanson (approved):
Tested on a pair of ZMK Uno/DK kits set up as splits. Works fine, and I know others have leveraged for a while for "dongle mode" as well.

This is a great first step toward fully supported BLE dongle mode, which will need still more docs, and build fixes to make doing so more seamless.

Thanks for sticking w/ this!

PR #736, opened 2021-03-22
Zephyr 2.5 upgrade

petejohanson:
The start of the Zephyr 2.5.0 upgrade. I need to do more real world testing, so far it works for basics on my Zaphod prototype, but I am having some issue w/ deadlocks w/ the display enabled, which may or may not be Zaphod specific.

Issue #120, opened 2020-08-22
Onboarding: Missing bzip Dependency 


cdarwin:
While setting up a brand new Fedora based development environment based on the included documentation, I noticed a failure in the Zephyr SDK installation step related to a missing bzip executable.

```
[vagrant@fedora32 ~]$ sh "zephyr-toolchain-arm-${ZSDK_VERSION}-setup.run" --quiet -- -d ~/.local/zephyr-sdk-${ZSDK_VERSION}
Installing SDK to /home/vagrant/.local/zephyr-sdk-0.11.2
Creating directory /home/vagrant/.local/zephyr-sdk-0.11.2
Success
 [|] Installing arm tools...tar (child): bzip2: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting no
```

I was able to resolve the issue by installing bzip manually and after following the rest of the documentation I can successfully build firmware with the `west` command.

This is obviously a trivial issue but I find that a seamless dev onboarding makes for a more welcoming experience for potential contributors

PR #838, opened 2021-06-14
Tidy and synchronize host dependencies with Zephyr

innovaker:


innovaker:
I'd suggest we put this in just before #736 to combine the testing cycles and give `2.5` a solid footing.

innovaker:
@petejohanson, I traced Fedora back to dca81c6579460e2693d07ecaf08cb5ad766484fe.  It seems quite different to Zephyr's recommendations - even 2.3.0.

Are you happy with the updated list please?  I don't have Fedora to test it nor the experience to judge it.

innovaker:
@petejohanson, I've eyeballed all the relevant commits.

- `autoconf`, `automake`, `build-essential`, `g++` and `libtool` was added by @petejohanson in a709e956802b872421d373bb98816c1e75ca3805
- `gcc-multilib` was originally moved to `native_posix` by @petejohanson in 766eedebe53c1146ce4ea9db73d659d12e49b53a
- `bzip2` was added by #121 / 0c255aee84155405741ba8fa0c81c6d711106033

So I guess ...
- Is `bzip2` still required?  If so, is it all platforms or only Fedora (see #120 and #121)?
- Generally speaking, are you happy with the other synchronization changes please?

Thanks!

petejohanson:
@innovaker `bzip2` definitely still needed. Let me audit the rest.

innovaker:
> @innovaker `bzip2` definitely still needed. Let me audit the rest.

I've just removed `bzip2` from the `build` stage of `zmk-docker` and it builds fine on Ubuntu.

I've also confirmed that `bzip2` needed to be explicitly specified on Debian.

innovaker:
I was curious so I took a deeper dive into Fedora.  I've created a test branch of zmk-docker for Fedora 34.

- https://github.com/innovaker/zmk-docker/commits/test/fedora
- https://github.com/innovaker/zmk-docker/actions/runs/950782211

For reference, Zephyr 2.4.0 and 2.5.0 have the same list for Fedora:
```
sudo dnf group install "Development Tools" "C Development Tools and Libraries"
dnf install git cmake ninja-build gperf ccache dfu-util dtc wget \
  python3-pip python3-tkinter xz file glibc-devel.i686 libstdc++-devel.i686 \
  SDL2-devel
```

The only changes that ZMK seems to need beyond this list is:
- add `clang-tools-extra` for `clang-format`.
- add `python38` for Zephyr SDK's `arm-zephyr-eabi-gdb`.
  - Zephyr 2.6.0 adds this to the Zephyr documentation.

Everything else - `bzip2` etc. - is included by the `Developer Tools` and `C Development Tools and Libraries` groups.

Please can you confirm my findings @petejohanson?

We *could* skip the groups and define each dependency explicitly, but it'll require significantly more maintainence.  I personally don't think it's worth the trouble for Fedora.

innovaker:
Based on the zmk-docker work, we should probably consider adding these non-Zephyr dependencies to Ubuntu as well:
- `clang-format`
- `nodejs` (from `nodesource`)
- `tio`
- `gdb` (for debugging of unit tests)

To minimize stumbling blocks for budding contributors.

petejohanson:
@innovaker Is this ready to not be draft? Can we add the other deps like `tio` you mentioned and get this in?

innovaker:
> @innovaker Is this ready to not be draft? Can we add the other deps like `tio` you mentioned and get this in?

As previously posted, it was/is waiting for someone to confirm my Fedora findings.  I've only "run" Fedora via zmk-docker builds - I haven't tried developing with it.  I would assume Zephyr's instructions work, but they're different to your original instructions.

I've also just noticed that we've ignored the other platforms, so I've done a quick comparison.

### ZMK vs Zephyr

This seems to be the state of each platform (compared to Zephyr's instructions) as of b3b1bbd:
- Ubuntu
  - ZMK is missing `--no-install-recommends` - shall we add it?
- Debian
  - Isn't supported by Zephyr!
  - ZMK currently bundles it with Ubuntu but I don't think we should.
  - Given that neither Zephyr nor zmk-docker supports it, I suggest we remove Debian from the documentation entirely.  Thoughts?
- Windows
  - ZMK is missing `choco feature enable -n allowGlobalConfirmation` - shall we add it?
- macOS
  - ZMK adds:
    - `dfu-util`
    - `git`
    - `wget`
  - ZMK doesn't have:
    - `gperf`
    - `qemu`
  - I suggest we add the missing dependencies as we've effectively done with Ubuntu.  Thoughts?
  - Zephyr has explicit instructions for Homebrew: `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
- Raspberry OS
  - Isn't supported by Zephyr!
  - It looks like I synced it with Ubuntu for b3b1bbd, but in hindsight I think that's a mistake because Zephyr's instructions are only for Ubuntu.
  - I suggest we remove Raspberry OS entirely given that nobody is maintaining it.  Thoughts?
- Fedora
  - ZMK has an extra `sudo` (can be ignored)
- VS Code & Docker
   - N/A

It's weird that ...
- Zephyr includes `git` for Ubuntu and Windows but not for macOS.
- Zephyr includes `wget` for Ubuntu but nothing else.

`python38` (Fedora) for `arm-zephyr-eabi-gdb` will officially be added by Zephyr 2.6.0 and I'm happy ignoring it for now given you're probably the only person using it (otherwise we would've seen complaints).  I don't know why it wasn't included with Zephyr 2.5 but zmk-docker has an integration test for `arm-zephyr-eabi-gdb` so presumably it's fine on Ubuntu.

### ZMK specific

Although I touched on the non-Zephyr dependencies in previous messages (because I was working on related zmk-docker improvements at the time), I'd rather not get into them at the moment.  They're technically outside the scope of this PR and would require further testing time - especially if we consider all the development platforms.  To be clear these are ...
- `clang-tools-extra` (Fedora)
- `clang-format` (Ubuntu)
- `nodejs`
- `tio`
- `gdb`

### Future

Going forward, I highly recommend that we find a better approach than trying to keep both zmk's documentation and zmk-docker in sync with Zephyr's documentation.  My current favoured approach is just to point at Zephyr's documentation and then add extra steps for Zephyr's holes plus anything ZMK specific.  Still not ideal, but it should at least reduce the pain.

PR #864, opened 2021-07-07
a link and a note on how to install west

pvinis:


pvinis:
I might close this and do it differently?
I ended up running it like this instead:
```
docker run --rm -it -v $(pwd):/usr/src/project zmkfirmware/zmk-build-arm:2.4
```
from within my `zmk-config`.

docs/docs/customization.md
```diff
@@ -50,6 +50,10 @@ For instance, building kyria firmware from a user `myUser`'s `zmk-config` folder
 west build -b nice_nano -- -DSHIELD=kyria_left -DZMK_CONFIG="C:/Users/myUser/Documents/Github/zmk-config/config"
 ```
 
+:::note
+You can install `west` using `pip3 install -U west`. Take a look [here](https://docs.zephyrproject.org/latest/guides/west/install.html) and [here](https://docs.zephyrproject.org/latest/getting_started/index.html#getting-started) for more info.
```
petejohanson, line 54: Why not just point back to our ownr developer setup documentation that covers general setup of ZMK, including Zephyr dependencies?

pvinis:
I didn't want to install all deps in the end, so the docker way was muuuuuuch better.

PR #866, opened 2021-07-09
Implement a driver for ISSI IS31FL3741

ice9js:
This PR adds support for a IS41FL3741 matrix LED driver, enabling efficient control for up to 351 single color, or 117 RGB LEDs.  
In turn, this should allow for power efficient underglow or per-key RGB (#554).

The driver implements Zephyr's [led strip API](https://docs.zephyrproject.org/apidoc/latest/led__strip_8h.html). In order to map, RGB pixels to correct channels, one should use the `map` config property.  
For convenience, I also defined a set of macros that'd allow you to correctly set it up based on a schematic and pin numbers versus having to manually calculate the register index for each pixel. For example:

```
map = <
    RGB(SW(1), CS(1), CS(2), CS(3))
    RGB(SW(7), CS(3), CS(2), CS(1))
>;
```

Translates to:

```
map = <
    0x00 0x01 0x02
    0xb6 0xb5 0xb4
>;
```

Although the configuration can get a bit lengthy considering it supports up to 117 RGB LEDs, there are significant advantages of handling it this way:

- Mixing different color channels on the same source or sink rail is possible.
- It supports both common-anode as well as common-cathode LEDs. In which case you'd just turn it around: `RGB(CS(1), SW(1), SW(2), SW(3))`.
- Should you need to do anything more exotic you can still add a manual definition for the map or add new macros.

All in all, the configuration looks something like this:

```
#include <dt-bindings/zmk/issi_transform.h>

@i2c0 {
    ...
    led_matrix: is31fl3741@30 {
        compatible = "issi,is31fl3741";
        label = "IS31FL3741";
        reg = <0x30>;
        r-ext = <68>;
        led-max-current = <5>;
        sdb-gpios = <&gpio1 6 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>;

        map = <
            RGB(SW(7), CS(14), CS(13), CS(15))
            RGB(SW(9), CS(14), CS(13), CS(15))
            RGB(SW(5), CS(14), CS(13), CS(15))
            ....
        >;
}
```

I'm not sure I put the transform macros in the right place, so some guidance there would be appreciated.
Finally, a short animation of the driver in action using the above configuration:

https://user-images.githubusercontent.com/8056203/125053970-f9cb2d80-e0a5-11eb-8919-3608557e2354.mov


ice9js:
Rebased.

I also added gamma settings to better control the percepted brightness of the LEDs as well as each channel relative to each other in order to adjust for any unwanted color-skew. As these things are usually very specific to the particular LEDs being used, I believe the driver is the best place to adjust for that.

[Click here](https://github.com/muriworks/zmk-config/blob/master/config/boards/arm/muri_one/muri_one.dts#L239) for an example of the complete configuration for the driver.

hackrmomo:
This seems to have been ready already but wasn't merged. Is there something this was missing? I might be wrong but this seems like pretty core functionality for a lot of potential zmk users (I might be biased because I fall into this category though)

PR #867, opened 2021-07-14
Add nice!nano v2

Nicell:
The next version of the nice!nano is here. In terms of firmware, the two versions are nearly compatible. The only change is how the battery voltage is read and that the external VCC switch is flipped. I've used inheritance to reduce the number of changes.


.github/workflows/build.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       matrix:
         board:
           - bluemicro840_v1
-          - nice_nano
```
caksoylar, line 22: Can we keep the v1 in the build actions for existing users, so that they can access the settings reset uf2 and default shield keymaps for testing?

Nicell: @petejohanson What are your thoughts on this? The reset uf2s are interchangeable, but the default shield builds are definitely up for consideration.

petejohanson: I would say we keep it in the list for now. At least until we can get more modular stuff in, and our refinements to the builds for PRs versus `main` for which shield/board combos we build.

app/boards/arm/nice_nano/nice_nano.dtsi
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/dts-v1/;
```
petejohanson, line 7: As I look at this, we might be doing this elsewhere already for splits, but I wonder if this early directive should be kept in the toplevel `dts` file, not in any sub-includes. Thoughts?

app/boards/arm/nice_nano/nice_nano_v2.dts
```diff
@@ -0,0 +1,25 @@
...
+		compatible = "zmk,ext-power-generic";
+		label = "EXT_POWER";
+		control-gpios = <&gpio0 13 GPIO_ACTIVE_HIGH>;
+		init-delay-ms = <10>;
+	};
+
+	vbatt {
+		compatible = "zmk,battery-voltage-divider";
+		label = "BATTERY";
+		io-channels = <&adc (0x0D - 1)>;
```
petejohanson, line 20: Is this effectively the VDDH hack that's similar to the driver from @joelspadin in #750 ?

petejohanson (commented):
Couple small questions.

app/boards/arm/nice_nano/nice_nano.dtsi
Nicell: Looking at Zephyr, they seem to have this in both `dtsi`s and `dts` files for some boards and just in the `dts` for others. Most seem to use just once in the `dts`, so I'll go that route.

app/boards/arm/nice_nano/nice_nano_v2.dts
Nicell: This is correct. It's not too much of a hack, as the VDDH internal ADC works similarly to the AIN pins just with fewer options.

joelspadin: Yeah, the only really hacky things about it are the seemingly arbitrary numbers needed to get it to work. 0x0D is the number of the channel we want, but the driver adds one to it so we have to subtract one. Then we don't really have a voltage divider but the ADC needs to be rescaled, so we just use the voltage divider resistances to get the multiplier we need.

petejohanson (approved):

PR #868, opened 2021-07-17
feat!: Add chosen node for battery

joelspadin:
battery.c now uses the `zmk,battery` chosen node to select a battery sensor. Using the node labeled `"BATTERY"` is maintained as a fallback, but is now deprecated. Custom boards/shields should switch to using the chosen node.

Also changed to fetch all channels to support fuel gauge sensors that don't support fetching individual channels like max17055.

### DIFF
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -26,6 +26,8 @@ uint8_t zmk_battery_state_of_charge() { return last_state_of_charge; }
 #if DT_HAS_CHOSEN(zmk_battery)
 static const struct device *const battery = DEVICE_DT_GET(DT_CHOSEN(zmk_battery));
 #else
+#warning                                                                                           \
+    "Using a node labeled BATTERY for the battery sensor is deprecated. Set a zmk,battery chosen node instead. (Ignore this if you don't have a battery sensor.)"
 static const struct device *battery;
 #endif
 
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -5,6 +5,7 @@
  */
 
 #include <device.h>
+#include <devicetree.h>
 #include <init.h>
 #include <kernel.h>
 #include <drivers/sensor.h>
@@ -18,11 +19,15 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/battery.h>
 #include <zmk/events/battery_state_changed.h>
 
-const struct device *battery;
-
 static uint8_t last_state_of_charge = 0;
 
 uint8_t zmk_battery_state_of_charge() { return last_state_of_charge; }
+
+#if DT_HAS_CHOSEN(zmk_battery)
+static const struct device *const battery = DEVICE_DT_GET(DT_CHOSEN(zmk_battery));
+#else
+static const struct device *battery;
+#endif
 
 static int zmk_battery_update(const struct device *battery) {
     struct sensor_value state_of_charge;
@@ -75,10 +80,18 @@ static void zmk_battery_timer(struct k_timer *timer) { k_work_submit(&battery_wo
 K_TIMER_DEFINE(battery_timer, zmk_battery_timer, NULL);
 
 static int zmk_battery_init(const struct device *_arg) {
+#if !DT_HAS_CHOSEN(zmk_battery)
     battery = device_get_binding("BATTERY");
 
     if (battery == NULL) {
-        LOG_DBG("No battery device labelled BATTERY found.");
+        return -ENODEV;
+    }
+
+    LOG_WRN("Finding battery device labeled BATTERY is deprecated. Use zmk,battery chosen node.");
+#endif
+
+    if (!device_is_ready(battery)) {
+        LOG_ERR("Battery device \"%s\" is not ready", battery->name);
         return -ENODEV;
     }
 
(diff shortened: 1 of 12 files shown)


app/src/battery.c
```diff
@@ -18,16 +19,20 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
 
-const struct device *battery;
-
 static uint8_t last_state_of_charge = 0;
 
 uint8_t zmk_battery_state_of_charge() { return last_state_of_charge; }
 
+#if DT_HAS_CHOSEN(zmk_battery)
+
+#define BATTERY_DEV_LABEL DT_LABEL(DT_CHOSEN(zmk_battery))
```
petejohanson, line 28: We should be able to use some condition defines magic to fallback to the hardcoded label if a chosen node isn't found.

Ideally, we can also *warn* and add a deprecation notice, since there are quite a few out of tree boards that will break if we don't leave compatibility for a while.

petejohanson (changes requested):
Backwards compatibility will be needed. See my comment for thoughts/details. 

Thanks!

app/src/battery.c
joelspadin: I can add a fallback, but I don't think I can make a compile time warning about deprecation. I could add a log warning (that probably nobody will ever see). I don't think it is possible find a node with a given label property at compile time, which means I can't identify at compile time whether you are using `label = "BATTERY"` or you just don't have a battery sensor.

petejohanson: You should be able to do something like:

```
#if DT_HAS_CHOSEN{zmk_battery)
#define BATTERY_NODE_ID DT_CHOSEN(zmk_battery)
#else
#warn "Use of 'BATTERY' node label is deprecated, use `zmk,battery` chosen node instead."
#define BATTERY_NODE_ID DT_NODELABEL("BATTERY")
#endif
```

And then use the define in the rest of the code.

joelspadin: Summarized discussion from Discord: DT_NODELABEL() is for the other kind of label. There is no way to get a node by its label property at compile time, so unfortunately this is the best we can do.

Darryldh:
I'll stay out of the "chosen" conversation, but in terms of your battery.c recommendation, I have definitely been able to fetch individual channels from the MAX17055... I think you just have to do a sensor_sample_fetch first. I cannot imagine any fuel gauge that would not allow calling specific channels.... seems like kind of an important thing ;-)

joelspadin:
> I'll stay out of the "chosen" conversation, but in terms of your battery.c recommendation, I have definitely been able to fetch individual channels from the MAX17055... I think you just have to do a sensor_sample_fetch first. I cannot imagine any fuel gauge that would not allow calling specific channels.... seems like kind of an important thing ;-)

My change does switch to using `sensor_sample_fetch()`. Note the difference between "fetch" and "get". The MAX17055 will assert if you try to fetch anything other than all channels. Once you have fetched all channels you can get any channel you want.

Nicell (approved):
Definitely overdue to get this changed

Nicell:
@petejohanson Could you give this a look when you get the chance? Pretty straightforward change now without the sensor fetch parts.

app/src/battery.c
```diff
@@ -18,12 +19,16 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
-const struct device *battery;
-
 static uint8_t last_state_of_charge = 0;
 
 uint8_t zmk_battery_state_of_charge() { return last_state_of_charge; }
 
+#if DT_HAS_CHOSEN(zmk_battery)
+static const struct device *const battery = DEVICE_DT_GET(DT_CHOSEN(zmk_battery));
+#else
+static const struct device *battery;
```
petejohanson, line 29: Can we add a small warn here:

```suggestion
#warning "Using specially labeled BATTERY node is deprecated, in favor of setting a zmk,battery chosen node."
static const struct device *battery;
```

petejohanson (changes requested):
Just one minor request here. Otherwise looks good.

app/src/battery.c
joelspadin: Only if you can either find a macro to get a device by label at compile time (I couldn't) or are okay with that warning appearing for every keyboard that doesn't have a battery sensor at all.

petejohanson: We only compile `battery.c` if ZMK_BLE is enabled:

```
```

So I think that's reasonable/safe for now?

joelspadin: That means you probably have a battery, but it doesn't necessarily mean your hardware has a way to *check* the battery. Do you think that scenario is unlikely that it's okay to give a meaningless warning in it?

petejohanson: I think we *do* warn here, because I think our *next* step is to do something like:

```
DT_CHOSEN_ZMK_BATTERY := zmk,battery

config ZMK_BATTERY_READ
    bool "Read battery levels to report them"
    default $(dt_chosen_enabled,DT_CHOSEN_ZMK_BATTERY)
```

Then use that Kconfig in CMakeLists.txt to determine if we build this.

joelspadin: Done.

joelspadin:
I also added a page to the docs to explain the `zmk,battery` chosen node.

petejohanson (approved):
Perfect! Thanks Joel.

PR #872, opened 2021-07-18
refactor(display): Saner font selection/defaults.

petejohanson:
* Only select fonts for the default built in status screen
* Leverage theme default fonts, instead of hardcoding theme
  details in each component.


petejohanson:
Tested on OLEDs and Zaphod sharp memory display w/ custom status screen, working as expected for both defaults and custom display code now.

PR #875, opened 2021-07-23
Update to Corne default keymap

devonallie:
Updated default corne layout.
Reference: https://github.com/qmk/qmk_firmware/blob/master/keyboards/crkbd/keymaps/default/keymap.c


petejohanson (approved):
Thanks!

PR #876, opened 2021-07-24
Refactor Pro Micro pin mapping

Nicell:
This PR refactors the Pro Micro pin mapping:

- Adds in the missing pins 18-21
- Renames `pro_micro_d` to `pro_micro` (`pro_micro_d` alias included for compatibility)
- Removes all uses of `pro_micro_a` in favor of `pro_micro` 
- Updates documentation image + description of Pro Micro pins


docs/docs/development/boards-shields-keymaps.md
```diff
@@ -47,6 +47,6 @@ in the `app/boards/shields/${board_name}` directory, e.g. `app/boards/shields/cl
...
-  - A [chosen](https://docs.zephyrproject.org/2.5.0/guides/dts/intro.html#aliases-and-chosen-nodes) node named `zmk,kscan` which references the configured KSCAN driver (usually a GPIO matrix). For these keyboards, to be compatible with any Pro Micro compatible boards, the KSCAN configuration should reference the [nexus node](https://docs.zephyrproject.org/2.5.0/guides/porting/shields.html#gpio-nexus-nodes) that ZMK has standardized on. In particular, the `&pro_micro_a` and `&pro_micro_d` aliases can be used to reference the standard `A#` and `D#` pins in shields.
+  - A [chosen](https://docs.zephyrproject.org/2.5.0/guides/dts/intro.html#aliases-and-chosen-nodes) node named `zmk,kscan` which references the configured KSCAN driver (usually a GPIO matrix). For these keyboards, to be compatible with any Pro Micro compatible boards, the KSCAN configuration should reference the [nexus node](https://docs.zephyrproject.org/2.5.0/guides/porting/shields.html#gpio-nexus-nodes) that ZMK has standardized on. In particular, the `&pro_micro` aliases can be used to reference the standard pins of a Pro Micro for shields.
```
petejohanson, line 50:
```suggestion
  - A [chosen](https://docs.zephyrproject.org/2.5.0/guides/dts/intro.html#aliases-and-chosen-nodes) node named `zmk,kscan` which references the configured KSCAN driver (usually a GPIO matrix). For these keyboards, to be compatible with any Pro Micro compatible boards, the KSCAN configuration should reference the [nexus node](https://docs.zephyrproject.org/2.5.0/guides/porting/shields.html#gpio-nexus-nodes) that ZMK has standardized on. In particular, the `&pro_micro` aliases can be used to reference the standard digital pins of a Pro Micro for shields.
```

petejohanson (changes requested):
Minor nitpick. Appreciate the backwards compatible approach, and getting all existing shields updated to help set proper examples moving forward.

docs/docs/development/boards-shields-keymaps.md
Nicell: Added, thanks!

petejohanson (commented):
Thoughts on how to prevent new shields getting merged in w/ the old names? Add an item to the check-list to make sure we're keeping an eye out for it?

Nicell:
> Thoughts on how to prevent new shields getting merged in w/ the old names? Add an item to the check-list to make sure we're keeping an eye out for it?

It doesn't look like there's a way to add a deprecation warning to our nexus nodes in DT, so no way to really detect usage. A check-list item seems like the best idea for now.

petejohanson (approved):
Thanks! This is much better.

PR #878, opened 2021-07-25
refactor(kscan): Demacroify GPIO matrix driver

joelspadin:
Refactored the GPIO matrix kscan driver so that only the data and config structures are defined in the foreach macro. Functionality is unchanged except for the addition of DT properties to adjust polling speed.

This matches how most other drivers in Zephyr are written, and it should make it easier to add other enhancements later, like improved and customizable debounce behavior.

### DIFF
--- a/app/drivers/kscan/kscan_gpio_matrix.c
+++ b/app/drivers/kscan/kscan_gpio_matrix.c
@@ -1,48 +1,158 @@
 /*
- * Copyright (c) 2020 The ZMK Contributors
+ * Copyright (c) 2020-2021 The ZMK Contributors
  *
  * SPDX-License-Identifier: MIT
  */
 
-#define DT_DRV_COMPAT zmk_kscan_gpio_matrix
-
 #include <device.h>
-#include <drivers/kscan.h>
+#include <devicetree.h>
 #include <drivers/gpio.h>
+#include <drivers/kscan.h>
 #include <logging/log.h>
+#include <sys/__assert.h>
+#include <sys/util.h>
 
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
+#define DT_DRV_COMPAT zmk_kscan_gpio_matrix
+
 #if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
 
-struct kscan_gpio_item_config {
-    char *label;
+#define INST_DIODE_DIR(n) DT_ENUM_IDX(DT_DRV_INST(n), diode_direction)
+#define COND_DIODE_DIR(n, row2col_code, col2row_code)                                              \
+    COND_CODE_0(INST_DIODE_DIR(n), row2col_code, col2row_code)
+
+#define INST_ROWS_LEN(n) DT_INST_PROP_LEN(n, row_gpios)
+#define INST_COLS_LEN(n) DT_INST_PROP_LEN(n, col_gpios)
+#define INST_MATRIX_LEN(n) (INST_ROWS_LEN(n) * INST_COLS_LEN(n))
+#define INST_INPUTS_LEN(n) COND_DIODE_DIR(n, (INST_COLS_LEN(n)), (INST_ROWS_LEN(n)))
+
+#define USE_POLLING IS_ENABLED(CONFIG_ZMK_KSCAN_MATRIX_POLLING)
+#define USE_INTERRUPTS (!USE_POLLING)
+
+#define COND_INTERRUPTS(code) COND_CODE_1(CONFIG_ZMK_KSCAN_MATRIX_POLLING, (), code)
+#define COND_POLL_OR_INTERRUPTS(pollcode, intcode)                                                 \
+    COND_CODE_1(CONFIG_ZMK_KSCAN_MATRIX_POLLING, pollcode, intcode)
+
+// TODO (Zephr 2.6): replace the following
+// kscan_gpio_dt_spec -> gpio_dt_spec
+// KSCAN_GPIO_DT_SPEC_GET_BY_IDX -> GPIO_DT_SPEC_GET_BY_IDX
+// gpio_pin_get -> gpio_pin_get_dt
+// gpio_pin_set -> gpio_pin_set_dt
+// gpio_pin_interrupt_configure -> gpio_pin_interrupt_configure_dt
+struct kscan_gpio_dt_spec {
+    const struct device *port;
     gpio_pin_t pin;
-    gpio_flags_t flags;
+    gpio_dt_flags_t dt_flags;
 };
 
-#define _KSCAN_GPIO_ITEM_CFG_INIT(n, prop, idx)                                                    \
+#define KSCAN_GPIO_DT_SPEC_GET_BY_IDX(node_id, prop, idx)                                          \
     {                                                                                              \
-        .label = DT_INST_GPIO_LABEL_BY_IDX(n, prop, idx),                                          \
-        .pin = DT_INST_GPIO_PIN_BY_IDX(n, prop, idx),                                              \
-        .flags = DT_INST_GPIO_FLAGS_BY_IDX(n, prop, idx),                                          \
-    },
+        .port = DEVICE_DT_GET(DT_GPIO_CTLR_BY_IDX(node_id, prop, idx)),                            \
+        .pin = DT_GPIO_PIN_BY_IDX(node_id, prop, idx),                                             \
+        .dt_flags = DT_GPIO_FLAGS_BY_IDX(node_id, prop, idx),                                      \
+    }
 
-#define _KSCAN_GPIO_ROW_CFG_INIT(idx, n) _KSCAN_GPIO_ITEM_CFG_INIT(n, row_gpios, idx)
-#define _KSCAN_GPIO_COL_CFG_INIT(idx, n) _KSCAN_GPIO_ITEM_CFG_INIT(n, col_gpios, idx)
+#define KSCAN_GPIO_ROW_CFG_INIT(idx, inst_idx)                                                     \
+    KSCAN_GPIO_DT_SPEC_GET_BY_IDX(DT_DRV_INST(inst_idx), row_gpios, idx),
+#define KSCAN_GPIO_COL_CFG_INIT(idx, inst_idx)                                                     \
+    KSCAN_GPIO_DT_SPEC_GET_BY_IDX(DT_DRV_INST(inst_idx), col_gpios, idx),
 
-#if !defined(CONFIG_ZMK_KSCAN_MATRIX_POLLING)
-static int kscan_gpio_config_interrupts(const struct device **devices,
-                                        const struct kscan_gpio_item_config *configs, size_t len,
-                                        gpio_flags_t flags) {
-    for (int i = 0; i < len; i++) {
-        const struct device *dev = devices[i];
-        const struct kscan_gpio_item_config *cfg = &configs[i];
+enum kscan_diode_direction {
+    KSCAN_ROW2COL,
+    KSCAN_COL2ROW,
+};
 
-        int err = gpio_pin_interrupt_configure(dev, cfg->pin, flags);
+struct kscan_matrix_irq_callback {
+    const struct device *dev;
+    struct gpio_callback callback;
+    struct k_delayed_work *work;
+};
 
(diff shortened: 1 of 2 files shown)


app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -51,257 +159,301 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
...
-                    data->callback(dev, r, c, pressed);                                            \
-                }                                                                                  \
-            }                                                                                      \
-        }                                                                                          \
-        COND_INTERRUPTS(                                                                           \
-            if (submit_follow_up_read) {                                                           \
-                COND_CODE_0(DT_INST_PROP(n, debounce_period), ({ k_work_submit(&data->work); }),   \
-                            ({                                                                     \
-                                k_delayed_work_cancel(&data->work);                                \
-                                k_delayed_work_submit(&data->work, K_MSEC(5));                     \
```
joelspadin, line 190: Was this an error in the original code? This is conditional based on there being a debounce period, but then it uses a hard-coded 5 ms instead of the debounce period. I added a separate DT property to control this, but if this was supposed to be the debounce time I can remove it.

joelspadin: I removed the separate property assuming that this was intended to use the debounce period.

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -51,257 +158,300 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
...
+}
+
+static int kscan_matrix_init_input_inst(const struct device *dev,
+                                        const struct kscan_gpio_dt_spec *gpio, const int index) {
+    if (!device_is_ready(gpio->port)) {
+        LOG_ERR("GPIO is not ready: %s", gpio->port->name);
+        return -ENODEV;
+    }
+
+    int err = gpio_pin_configure(gpio->port, gpio->pin, GPIO_INPUT | gpio->dt_flags);
```
Nicell, line 326: What are your thoughts on inferring all of the GPIO flags for inputs (and outputs)? Instead of manually defining them in our DT, we could infer inputs to have `GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN` and outputs to have `GPIO_ACTIVE_HIGH`. We know this based on the `diode-direction` property. To me it seems that `diode-direction` and these GPIO flags explain the same thing, so we should only need one. This would result in a cleaner keyboard DT.

joelspadin: Yeah, I was considering adding that here, but figured it would be better to do that as a separate commit. I think we'd want to infer the flags only if `gpio->dt_flags` is 0 so you can still use active high, pull up if needed.

app/drivers/kscan/kscan_gpio_matrix.c
petejohanson: Yeah, this was a bug.

app/drivers/kscan/kscan_gpio_matrix.c
petejohanson: I would rather we do this as a follow up. We've discussed this previously, and I'm *mostly* open to it, but we definitely have one scenario where that is *flipped*, which is Ferris, where the IO expander only has a pull down, so the diode direct does *not* actually infer those flags. Not saying we don't do it, just needs to be clearly documented if we do.

Nicell: I was going to suggest that we do `GPIO_ACTIVE_LOW` + `GPIO_ACTIVE_LOW | GPIO_PULL_UP` by default since I knew that atmegas only have pull-ups, but since we don't support atmegas, I didn't. I guess this still applies to things like these IO expanders and other older/simpler ICs. Generally, it seems that pull-ups are found everywhere though, so it might be a saner default.

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -1,48 +1,155 @@
...
+#define INST_DIODE_DIR(n) DT_ENUM_IDX(DT_DRV_INST(n), diode_direction)
+#define COND_DIODE_DIR(n, row2col_code, col2row_code)                                              \
+    COND_CODE_0(INST_DIODE_DIR(n), row2col_code, col2row_code)
+
+#define INST_ROWS_LEN(n) DT_INST_PROP_LEN(n, row_gpios)
+#define INST_COLS_LEN(n) DT_INST_PROP_LEN(n, col_gpios)
+#define INST_MATRIX_LEN(n) (INST_ROWS_LEN(n) * INST_COLS_LEN(n))
+#define INST_INPUTS_LEN(n) COND_DIODE_DIR(n, (INST_COLS_LEN(n)), (INST_ROWS_LEN(n)))
+
+#define USE_POLLING IS_ENABLED(CONFIG_ZMK_KSCAN_MATRIX_POLLING)
```
petejohanson, line 30: Not needed *this PR*, but the reality, I'm not convinced this really should be a global Kconfig setting, versus a per-driver instance property. Will review more to see if that really *matters* in the scope of these changes or not.

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -1,48 +1,155 @@
...
 };
 
-#define _KSCAN_GPIO_ITEM_CFG_INIT(n, prop, idx)                                                    \
+#define KSCAN_GPIO_DT_SPEC_GET_BY_IDX(node_id, prop, idx)                                          \
     {                                                                                              \
-        .label = DT_INST_GPIO_LABEL_BY_IDX(n, prop, idx),                                          \
-        .pin = DT_INST_GPIO_PIN_BY_IDX(n, prop, idx),                                              \
-        .flags = DT_INST_GPIO_FLAGS_BY_IDX(n, prop, idx),                                          \
-    },
+        .port = DEVICE_DT_GET(DT_GPIO_CTLR_BY_IDX(node_id, prop, idx)),                            \
```
petejohanson, line 51: Yeah, this is much nicer. Thanks.

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -1,48 +1,155 @@
...
+    const struct device *dev;
+    kscan_callback_t callback;
+    struct k_delayed_work work;
+#if USE_POLLING
+    struct k_timer poll_timer;
+#else
+    /** Array of length config->inputs.len */
+    struct kscan_matrix_irq_callback *irqs;
+#endif
+    /** Current state of the matrix as a flattened 2D array. */
```
petejohanson, line 82:
```suggestion
    /** Current state of the matrix as a flattened 2D array of length (config->rows.len * config->cols.len). */
```

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -1,48 +1,155 @@
...
+    bool *current_state;
+    /** Temporary buffer for reading in the next matrix state. */
```
petejohanson, line 84: Can we note the 2D/length details here as well please?

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -1,48 +1,155 @@
...
+    const struct kscan_matrix_config *config = dev->config;
 
-        int err = gpio_pin_interrupt_configure(dev, cfg->pin, flags);
+    for (int i = 0; i < config->inputs.len; i++) {
+        const struct kscan_gpio_dt_spec *gpio = &config->inputs.gpios[i];
 
+        int err = gpio_pin_interrupt_configure(gpio->port, gpio->pin, flags);
         if (err) {
-            LOG_ERR("Unable to enable matrix GPIO interrupt");
+            LOG_WRN("Unable to configure interrupt for pin %u on %s", gpio->pin, gpio->port->name);
```
petejohanson, line 152: Any reason this isn't still an `LOG_ERR`?

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -51,257 +158,300 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
...
+        return err;
+    }
+
+    // While interrupts are disabled, set all outputs inactive so
+    // kscan_matrix_read() can scan them one by one.
+    return kscan_matrix_set_all_outputs(dev, 0);
+}
+#endif
+
+#if USE_INTERRUPTS
```
petejohanson, line 187: We're already in an `#if USE_INTERRUPTS` block. Any reason you've chosen to duplicate the preprocessor blocks around each function?

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -51,257 +158,300 @@ static int kscan_gpio_config_interrupts(const struct device **devices,
...
+            LOG_ERR("Failed to set output %i active: %i", o, err);
+            return err;
+        }
+
+        for (int i = 0; i < config->inputs.len; i++) {
+            const struct kscan_gpio_dt_spec *in_gpio = &config->inputs.gpios[i];
+
+            const int index = state_index_io(config, i, o);
+            const int pressed = gpio_pin_get(in_gpio->port, in_gpio->pin);
+            data->next_state[index] = pressed;
```
petejohanson, line 222: Perhaps:

```suggestion
            data->next_state[index] = gpio_pin_get(in_gpio->port, in_gpio->pin);
```

petejohanson (commented):
In general... awesome! A few comments/questions on some of the details.

Thanks for working on this!

app/drivers/kscan/kscan_gpio_matrix.c
joelspadin: Yeah, I had considered that, but decided to just match the existing behavior for now.

app/drivers/kscan/kscan_gpio_matrix.c
joelspadin: After looking at this more, I think we'd need to create our own "auto" define. `GPIO_ACTIVE_HIGH` is 0, so using 0 would make it impossible to select just `GPIO_ACTIVE_HIGH` (e.g. when you have an external pull down) instead of `GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN`.

app/drivers/kscan/kscan_gpio_matrix.c
joelspadin: No idea why I changed this. Changed it back.

app/drivers/kscan/kscan_gpio_matrix.c
joelspadin: IMO it makes it easier to tell that you're in a `USE_INTERRUPTS` block, and you can reorder functions as needed without accidentally inserting something into the `USE_INTERRUPTS` block that you didn't intend to be interrupt-specific.

app/drivers/kscan/kscan_gpio_matrix.c
joelspadin: Done. (At one point line 222 was a function call like in the old code, but I changed it and missed simplifying it further.)

app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -1,48 +1,158 @@
 /*
- * Copyright (c) 2020 The ZMK Contributors
+ * Copyright (c) 2020-2021 The ZMK Contributors
```
joelspadin, line 2: How do we want to handle copyright dates? Should I leave this as the original author date, or use a date range like this?

app/drivers/kscan/kscan_gpio_matrix.c
joelspadin: Done.

app/drivers/kscan/kscan_gpio_matrix.c
joelspadin: Done.

joelspadin:
Rebased. No changes.

app/drivers/kscan/kscan_gpio_matrix.c
petejohanson: Since it's a bit field... we would risk trying to use an unused bit here to flag a "auto". Let's discuss further on a follow up.

petejohanson (approved):
Code looks good, tested on my Ferris and working great.

PR #879, opened 2021-07-26
docker-compose development workflow

kitchen:
I don’t use VSCode. Nothing against it, I just use a different environment. And I don’t like globally installing build deps and such, so I much prefer containerized environments! So I’ve tried to reproduce the VSCode devcontainer workflow here as much as possible.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+---
+version: "3.9"
+services:
+    shell:
+        ports:
+            - 3000
+        build:
+            context: .devcontainer/
+            dockerfile: Dockerfile
+        volumes:
+            - $PWD:/workspace
+            - zmk-root-user:/root
+            - zmk-zephyr:/workspace/zephyr
+            - zmk-zephyr-modules:/workspace/modules
+            - zmk-zephyr-tools:/workspace/tools
+
+volumes:
+    zmk-root-user:
+    zmk-zephyr:
+    zmk-zephyr-modules:
+    zmk-zephyr-tools:
--- a/docs/docs/development/build-flash.md
+++ b/docs/docs/development/build-flash.md
@@ -90,7 +90,11 @@ This produces `left` and `right` subfolders under the `build` directory and two 
 
 ### Building from `zmk-config` Folder
 
-Instead of building .uf2 files using the default keymap and config files, you can build directly from your [`zmk-config` folder](../user-setup.md#github-repo) by adding
+Getting your `zmk-config` to be used by your build environment differs based on what type of build environment you are using.
+
+#### non-containerized environment
+
+Instead of building .uf2 files using the default keymap and config files, you can build directly from your [`zmk-config` folder](../user-setup#github-repo) by adding
 `-DZMK_CONFIG="C:/the/absolute/path/config"` to your `west build` command. **Notice that this path should point to the folder labelled `config` within your `zmk-config` folder.**
 
 For instance, building kyria firmware from a user `myUser`'s `zmk-config` folder on Windows 10 may look something like this:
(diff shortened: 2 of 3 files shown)


kitchen:
draft comment: still working on the docs and such.

meta comment: the dev environment docs are visually quite difficult to navigate, imo, and we might want to re-envision them somehow. I have some ideas, but I think they would be out of scope for this PR. They could certainly be a dependency for this PR, however :)

docs/docs/development/build-flash.md
```diff
@@ -90,6 +90,10 @@ This produces `left` and `right` subfolders under the `build` directory and two
 
 ### Building from `zmk-config` Folder
 
+Getting your `zmk-config` to be used by your build environment differs based on what type of build environment you are using.
+
+#### non-containerized environment
```
petejohanson, line 95: This needs proper capitalization, also for anyone that isn't a developer, or doesn't know what a "containerized environment" is, this might be really confusing.

Perhaps something like:

```suggestion
#### Local/Host Build Environment.
```

docs/docs/development/build-flash.md
```diff
@@ -116,6 +122,16 @@ docker volume create --driver local -o o=bind -o type=none -o \
 
 Now start VSCode and rebuild the container after being prompted. You should be able to see your zmk-config mounted to `/workspaces/zmk-config` inside the container. So you can build your custom firmware with `-DZMK_CONFIG="/workspaces/zmk-config/config"`.
 
+#### docker compose build environment
```
petejohanson, line 125: Headers should use caps:

```suggestion
#### Docker Compose Build Environment
```

docs/docs/development/setup.md
```diff
@@ -16,6 +16,7 @@ values={[
 {label: 'Raspberry OS', value: 'raspberryos'},
 {label: 'Fedora', value: 'fedora'},
 {label: 'VS Code & Docker', value: 'docker'},
+{label: 'docker compose', value: 'dockercompose'},
```
petejohanson, line 19:
```suggestion
{label: 'Docker Compose', value: 'dockercompose'},
```

petejohanson (changes requested):
A few thoughts on the docs.

kitchen:
ok, this should be ready to roll now, I just tested the `docker compose run -v` on my mac and it seems to be working as expected now!

smeikx:
What is holding this back? Just merged it locally with `main`, read the docs and successfully built the .uf2 files for kyria. 
Works perfectly fine, I am really glad this work has already been done.

docs/docs/development/build-flash.md
```diff
@@ -90,7 +90,11 @@ This produces `left` and `right` subfolders under the `build` directory and two
 
 ### Building from `zmk-config` Folder
 
-Instead of building .uf2 files using the default keymap and config files, you can build directly from your [`zmk-config` folder](../user-setup.md#github-repo) by adding
+Getting your `zmk-config` to be used by your build environment differs based on what type of build environment you are using.
+
+#### Local/Host Build Environment.
+
+Instead of building .uf2 files using the default keymap and config files, you can build directly from your [`zmk-config` folder](../user-setup#github-repo) by adding
```
caksoylar, line 97:
```suggestion
Instead of building .uf2 files using the default keymap and config files, you can build directly from your [`zmk-config` folder](../user-setup.md#github-repo) by adding
```
I think this was changed by a PR by @dxmh, so the diff should be cleaner with the `.md` added back.

caksoylar (approved):
Since this is confirmed working by an additional user, I think it's good to merge pending this tiny fix.

LeducH:
@kitchen Why are you removing the container after execution `docker compose run --rm .....` ? 

It seems to me it could be good the keep them up since maybe a accidental `crtl + d` throws you out then . 

In any case great job with the suggestions. All the power to docker

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+---
+version: "3.9"
```
chrissicool, line 2: My Debian 11.5 complains about the version being too high. 

```
$ docker-compose -v
docker-compose version 1.25.0, build unknown
$ docker --version
Docker version 20.10.5+dfsg1, build 55c4c88
```

Using version: "3" does the trick for me. Works like a charm. And it also has to be 'docker-compose' in the docs.

I assume you are using a newer version of Docker. Does it work for you, if you switch these bits, so that older systems can use this as well?

Nick-Munnich:
Closing as I believe this was made redundant by #2548. Feel free to reopen if you feel otherwise.

Issue #502, opened 2020-12-11
ci(build): distill workflow


innovaker:
I recently increased our ci test coverage to include all boards and shields because we had some blind spots which needed to be covered for verifying #401.  However, GitHub Action's random `ECONNRESET` issue (during the archiving step) is subsequently hitting us harder than before and we're having to restart runs more often.

I've [suggested](https://discord.com/channels/719497620560543766/719544230497878116/787032584873443429) we explore a new strategy:
a) PRs and merges trigger testing of a select (small) group of boards/shields as well as build-args/configs, targeting broad feature coverage.
b) Nightly builds for every other board/shield as a safety net.

I'm happy to implement this change, but I don't have sufficient insight to design the specifics of (a).  We need from all the contributors.

Please use this issue to discuss the approach and suggest ideas, hopefully leading to a design for (a).

okke-formsma:
I would like to propose the following:
* build all boards for a small set of shields (proton-c and nice nano for example)
* build all shields for a small set of boards (kyria and qaz for example)

This would cover the most used combinations.

This would require us to define multiple jobs, as there is only one matrix definition per job possible.

Nicell:
> * build all boards for a small set of shields (proton-c and nice nano for example)
> * build all shields for a small set of boards (kyria and qaz for example)

This seems like a good start. I think we can also keep the current include list since that's quite short. I'd like to see #500 be done for underglow and other future features. 

The only potential issues I see for this small coverage case is when a new shield is submitted that works on the nice!nano, but it breaks on other boards due to missing a board specific overlay or enabling a feature that isn't supported. These shields should be able to work without a board specific overlay and shouldn't have bad config options enabled, but it will be harder to catch that error without testing a few boards (or all). See this commit as an example: https://github.com/zmkfirmware/zmk/pull/429/commits/a50fdeca575e7eee91e622b7f6a9dd0a3e40a8cb

It would be neat if we could have a way of building a shield on every supported board for PRs, but I'm thinking GitHub actions may be a bit too limited to do this. This page has some information on labels be used to trigger certain builds, but I'm not sure if we can work with this: https://github.community/t/do-something-if-a-particular-label-is-set/17149

innovaker:
There's also the option of writing a dedicated script (i.e. Python) which generates a tailored build matrix.  Would that be more useful @Nicell?

mcrosson:
[...]

PR #883, opened 2021-07-27
Add hardware metadata files, generate setup scripts and hardware docs from them.

petejohanson:
Draft PR for the ongoing work. I am approaching a reasonable *minimal* metadata format to address our immediate pain points, that I believe can be used as a starting point for our further work on this, more advanced setup web utility from @joelspadin , etc.

## Overview

The PR proposes a standard `${item_id}.zmk.yml` filenaming convention, where the item might be a board, shield, or interconnect. Those files contain a standard YAML structure, and I have also created a JSON schema that can be used to *validate* these files, e.g. `west metadata check`, including a GH Action to perform these checks.

Each item has a required `type` property of `board`, `shield`, or `interconnect`. By having a dedicated interconnect file, we can include information about the interconnect without duplicating it, and also achieve "promote a board ID to an interconnect ID" and keep the metadata about both uses distinct.

### Examples

`boards/arm/nrfmicro/nrfmicro_13.zmk.yml`:

```
id: nrfmicro_13
name: nRFMicro 1.3/1.4
type: board
arch: arm
outputs:
  - usb
  - ble
url: https://github.com/joric/nrfmicro/
exposes: [pro_micro]
```

`boards/shields/corne/corne.zmk.yml`:

```
id: corne
name: Corne
type: shield
url: https://github.com/foostan/crkbd/
requires: [pro_micro]
exposes: [i2c_oled]
features:
  - keys
  - display
siblings:
- corne_left
- corne_right
```

`boards/interconnects/pro_micro.zmk.yml`:

```
id: pro_micro
name: Pro Micro
type: interconnect
url: https://www.sparkfun.com/products/12640
manufacturer: SparkFun
description: |
  The SparkFun Pro Micro grew popular as a low cost ATmega32U4 board with sufficient GPIO and peripherals
  to work for many keyboard needs. Since the original Pro Micro, many pin compatible boards have appeared,
  with various changes or improvements, such as the Elite-C w/ USB-C, nice!nano with nRF52840 wireless.
```

### Hardware Docs Page

Link: https://deploy-preview-883--zmk.netlify.app/docs/hardware

Screenshot:

Screenshot from 2021-07-26 22-27-48

## TODO

- [ ] Ensure everyone is happy w/ current minimalist format.
- [ ] Investigate "layout" metadata, and if we need to account for that in this first pass at metadata + schema.
- [ ] Add docs for how to create such a file when writing a board or shield.
- [ ] Complete metadata for existing boards/shield.
- [ ] Determine regression w/ styling of code blocks being lost.


Nicell:
Generally, I think this metadata format is a good start. I like the `exposes`/`requires` stuff, nice. Should things like the `pro_micro` expose `pro_micro`? or is that inherent to its ID? 

Maybe we can consider rolling all the power profiler stuff into here too. Although that may require a bit of a refactor on how I currently do it, including getting readings of different SoCs.

Do we have a plan on how to update/extend this metadata? I can list 100 more things I could see us having on this, but I don't think we need them _right now_.

Nicell:
Besides getting rid of the hardcoded setup scripts and hardware docs page, we also need to update the GitHub workflows to not require hardcoding every board and shield as well.  Or maybe we can try doing an overhaul with only the minimum being built with PRs (#502).

petejohanson:
> Generally, I think this metadata format is a good start. I like the `exposes`/`requires` stuff, nice. Should things like the `pro_micro` expose `pro_micro`? or is that inherent to its ID? 

Hmm... Good question. Let me think on that.

> 
> Maybe we can consider rolling all the power profiler stuff into here too. Although that may require a bit of a refactor on how we currently do it, including getting readings of different SoCs.

I could see getting some power related stuff in here, yeah, e.g. quiescent at idle, perhaps add a `vcc_cutoff` option to the features list, etc
> 
> Do we have a plan on how to update/extend this metadata? I can list 100 more things I could see us having on this, but I don't think we need them _right now_. 

A few thoughts:

1. We can add a `version: 1` to these right now and make it required. That acknowledges we'll have future versions.
2. We can add additional *optional* properties easily, and then backfill additional values.
3. When we settle on the final new set of required items, we bump to a new major version, and update those to add to the required list 
4. Any code in the meantime leverages schema to add null checks to verify optional props they want to use.

docs/package.json
```diff
@@ -11,7 +11,8 @@
     "clear": "docusaurus clear",
     "lint": "eslint . --ext js,jsx,md,mdx",
     "prettier:check": "prettier --check .",
-    "prettier:format": "prettier --write ."
+    "prettier:format": "prettier --write .",
+    "typecheck": "tsc"
```
joelspadin, line 15: If you just intend to check types and not compile anything, I think you want to add `--noEmit` here. (I would be totally in favor of switching all the JS files to TS though.)

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,156 @@
...
+        {shields.map((s) => (
+          <HardwareLineItem key={s.id} item={s} />
+        ))}
+      </ul>
+    </div>
+  );
+}
+
+function HardwareList({ items }: HardwareListProps) {
+  let grouped: {
```
joelspadin, line 82: Minor: would probably be easier to read if you broke this out into an interface.

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,156 @@
...
+} from "../hardware-metadata";
+
+interface HardwareListProps {
+  items: HardwareMetadata[];
+}
+
+function itemIds(item: HardwareMetadata) {
+  if (item.type == "board" || item.type == "shield") {
+    let nodes = (item.siblings || [item.id])
+      .map<JSX.Element | string>((id) => <code key={id}>{id}</code>)
```
joelspadin, line 17: Could make a `type Foo = JSX.Element | string` to avoid repetition here.

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,156 @@
...
+    </li>
+  );
+}
+
+function mapInterconnect({
+  interconnect,
+  boards,
+  shields,
+}: {
+  interconnect?: Interconnect;
```
joelspadin, line 53: Minor: would probably be easier to read if you broke this out into an interface.

docs/package.json
petejohanson: All of that is in the `tsconfig.json` file.

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,156 @@
...
+      </ul>
+    </div>
+  );
+}
+
+function HardwareList({ items }: HardwareListProps) {
+  let grouped: {
+    onboard: Array<Board>;
+    interconnects: {
+      [key: string]: {
```
joelspadin, line 85: Repeated type from lines 53-55. Should probably be broken out into an interface. Then you can also simplify this to

```
interconnects: Record<string, Interconnect>;
```

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,156 @@
...
+        (prev, curr, index) => [...prev, index > 0 ? " and " : "", curr],
+        [] as Array<string | JSX.Element>
+      );
+    return <span key={item.id}>{nodes}</span>;
+  } else {
+    return <code key={item.id}>{item.id}</code>;
+  }
+}
+
+function typeLabel(type: HardwareMetadata["type"]): string {
```
joelspadin, line 28: I think you could simplify this to just a record like this:

```ts
// Type alias not necessary, but maybe easier to read
type HardwareType = HardwareMetadata['type'];

const TYPE_LABELS: Record<HardwareType, string> = {
  board: 'Board: ',
  shield: 'Shield: ',
  interconnect: 'Interconnect: ',
};
```

docs/package.json
joelspadin: Oh, I missed that. If we did want to use TypeScript elsewhere I think we'd need to remove that from tsconfig.json and then use --noEmit here, but that would be something for later.

schema/hardware-metadata.schema.json
```diff
@@ -0,0 +1,261 @@
...
+      "$ref": "#/$defs/shield"
+    },
+    {
+      "$ref": "#/$defs/interconnect"
+    }
+  ],
+  "$defs": {
+    "id": {
+      "type": "string",
+      "pattern": "^[a-z0-9_]+$"
```
joelspadin, line 19: Is it okay for an ID to start with a number? (Not sure if these IDs need to match C identifier rules.)

schema/hardware-metadata.schema.json
```diff
@@ -0,0 +1,261 @@
...
+        "file_format",
+        "id",
+        "name",
+        "url",
+        "type"
+      ],
+      "properties": {
+        "file_format": {
+          "type": "string",
+          "const": "1"
```
joelspadin, line 101: If the intent is to increment this any time we make a change to the schema, would a number be better? Then we'd be able to specify a range of allowed versions if that became useful.

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,156 @@
...
+          if (hm.features?.includes("keys")) {
+            agg.onboard.push(hm);
+          } else if (hm.exposes) {
+            hm.exposes.forEach((element) => {
+              let ic = agg.interconnects[element] || {
+                boards: [],
+                shields: [],
+              };
+              ic.boards.push(hm);
+              agg.interconnects[element] = ic;
```
Nicell, line 104: Thinking a bit about requires/exposes... Can't boards expose something like `i2c_oled`? This seems like a case where we may need to differentiate between interconnect requires/exposes and feature requires/exposes. In this case, `i2c_oled` isn't an interconnect, so this shouldn't be valid.

petejohanson: But it *is* an interconnect, just not one that a shields with the `keys` feature would likely use.

I think the interconnect is truly generic, and what features show up with the selected shield you add to it drives the "did we get what we need?"

petejohanson: Further example... That interconnect gets hijacked for pointers by modders...

Nicell: I see what you're saying, and I think it makes sense. The trouble I was having with this was during my build workflow rewrite, where I am compiling a similar list of all boards, shields, and interconnects for the nightly builds. The issue that crossed my mind was that I can't build the `i2c_oled` interconnect with a board and shield, but I think this is remedied by only building interconnects that have boards exposing it and shields requiring it. This would be a board + shield combo interconnect to build. Does that logic actually hold up?

petejohanson: Only thing I would add is building those combinations that resulting in the final set of `features` including the `keys` feature. Otherwise currently ZMK fails without a kscan chosen.

petejohanson: That additional check isn't needed *today* but might be needed later as we expand the use of interconnects more.

docs/package.json
petejohanson: Ok, good note, thanks.

schema/hardware-metadata.schema.json
petejohanson: These do *not* need to be C identifiers, and we should be a bit more flexible here, for when someone has a keyboard named funky things like _33ble (Looking at you, @tominabox1 )

schema/hardware-metadata.schema.json
petejohanson: I was modeling this off of other similar schemas like the docker-compose YAML files, for instance, which use strings so they can do semantic versioning w/ minor versions that don't break compat.

docs/src/components/hardware-list.tsx
petejohanson: Indeed. Pulled this out into a new `InterconnectDetails` type.

docs/src/components/hardware-list.tsx
petejohanson: Much nicer! Refactored to this approach.

docs/src/components/hardware-list.tsx
petejohanson: Indeed. Thanks!

docs/src/components/hardware-list.tsx
petejohanson: Yup!

docs/src/components/hardware-list.tsx
petejohanson: Yup. Done.

docs/docs/development/hardware-metadata-files.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+### Features
+
+Boards and shields should document the sets of hardware features found on them using the `features` array. There is a fixed enum of possible values to use here, which will be expanded over time. The current set of possible `features` values is:
+
+- `keys` - Any board or shield that contains keyboard keys should include this feature. It is a central feature used to determine if we have a "complete combination" for ZMK to produce a keyboard firmware when performig setup.
+- `display` - Indicates the hardware includes a display for use with the ZMK display functionality.
+- `encoder` - Indicates the hardware contains one or more (possibly optional) rotary encoders.
+- `underglow` - Indicatges the hardware includes underglow LEDs.
```
Nicell, line 96:
```suggestion
- `underglow` - Indicates the hardware includes underglow LEDs.
```

docs/docs/development/hardware-metadata-files.md
```diff
@@ -0,0 +1,111 @@
...
+
+Each metadata file includes a `type` property uniquely identifying the type of item, be it `board`, `shield`, or the less frequently needed `interconnect` (which is used to document generic hardware interconnects like the Pro Micro format):
+
+```yaml
+type: shield
+```
+
+### URL
+
+The `url` property should contain the canonical URL used to learn more about the board/shield, e.g. the main vendor website, or GitHub repository for a given keyboard/controller:
```
Nicell, line 70:
```suggestion
The `url` property should contain the canonical URL used to learn more about the board/shield, e.g. the main vendor website or GitHub repository for a given keyboard/controller:
```

docs/docs/development/hardware-metadata-files.md
```diff
@@ -0,0 +1,111 @@
...
+
+```yaml
+url: https://github.com/foostan/crkbd/
+```
+
+### Interconnect Requires/Exposes
+
+For boards and shields, one of the key pieces of high level information is compatibility between the two items. In particular, a board usually exposes one ore more "interconnects", the physical location/type of connections available, and their assigned possible uses (e.g. GPIO, power, ground, i2c, etc). Similarly, a shield is usually designed around one (or sometimes more) "interconnects" that allow it to connect to one of those boards.
+
+In ZMK, we encode both of those scenarios with the `exposes` and `requires` properties, respectfully. For example, for a Corne shield that requires a Pro Micro compatible controller to function, and simultaneously exposes a four pin header to be used by standard i2c OLED modules, the metadata file contains:
```
Nicell, line 80:
```suggestion
In ZMK, we encode both of those scenarios with the `exposes` and `requires` properties, respectively. For example, for a Corne shield that requires a Pro Micro compatible controller to function and simultaneously exposes a four pin header to be used by standard i2c OLED modules, the metadata file contains:
```

docs/docs/development/hardware-metadata-files.md
```diff
@@ -0,0 +1,111 @@
...
+
+- `keys` - Any board or shield that contains keyboard keys should include this feature. It is a central feature used to determine if we have a "complete combination" for ZMK to produce a keyboard firmware when performig setup.
+- `display` - Indicates the hardware includes a display for use with the ZMK display functionality.
+- `encoder` - Indicates the hardware contains one or more (possibly optional) rotary encoders.
+- `underglow` - Indicatges the hardware includes underglow LEDs.
+- `pointer` (future) - Used to indicate the hardware includes one or more pointer inputs, e.g. joystick, touchpad, or trackpoint.
+
+### Siblings
+
+The `siblings` array is used to identify multiple hardware items designedc to be used together as one logical device. Right now, that primarily is used to identify the two halves of a split keyboard, but future enhancement will include more complicated and flexible combinations.
```
Nicell, line 101:
```suggestion
The `siblings` array is used to identify multiple hardware items designed to be used together as one logical device. Right now, that primarily is used to identify the two halves of a split keyboard, but future enhancements will include more complicated and flexible combinations.
```

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+type ElementOrString = JSX.Element | string;
+
+function itemIds(item: HardwareMetadata) {
+  if (item.type == "board" || item.type == "shield") {
+    let nodes = (item.siblings || [item.id])
+      .map<ElementOrString>((id) => <code key={id}>{id}</code>)
+      .reduce(
+        (prev, curr, index) => [...prev, index > 0 ? " and " : "", curr],
```
Nicell, line 21: Future-proof for more than 2 siblings. Feel free to ignore.
```suggestion
        (prev, curr, index) => [...prev, index > 0 ? ", " : "", curr],
```

docs/src/templates/setup.sh.mustache
```diff
@@ -65,71 +65,73 @@ fi
...
+                board=${board_ids[$keyboard_index]}
+                board_title=${options[$keyboard_index]}
+                break
+            ;;
+
+        esac
+    done
+else
+    board=${keyboard}
+    echo "Support for onboard keyboard keyboards is still a work in progress."
```
Nicell, line 132: keyboard keyboard!
```suggestion
    echo "Support for onboard microcontroller keyboards is still a work in progress."
```

Nicell (commented):
Some small proofreading suggestions. Also, can we use package-lock.json version 2 for this PR?

schema/hardware-metadata.schema.json
joelspadin: That works. Just wanted to make sure we didn't need this to be `[a-z_][a-z0-9_]*`

joelspadin:
> Also, can we use package-lock.json version 2 for this PR?

Agreed. It would be nice to stop completely rewriting the lock file every few changes. (Upgrade to Node 14.x which includes NPM 7.x, or just upgrade NPM on its own.)

schema/hardware-metadata.schema.json
joelspadin: That's reasonable. I think the fact that you just used "1" instead of something like "1.0.0" made me think this was just an integer version and not a semantic version.

docs/src/components/hardware-list.tsx
petejohanson: Fixed.

docs/src/templates/setup.sh.mustache
petejohanson: Indeed! (fixed)

docs/docs/development/hardware-metadata-files.md
petejohanson: Good catches. Fixed.

docs/docs/development/hardware-metadata-files.md
petejohanson: Thanks.

petejohanson:
> > Also, can we use package-lock.json version 2 for this PR?
> 
> Agreed. It would be nice to stop completely rewriting the lock file every few changes. (Upgrade to Node 14.x which includes NPM 7.x, or just upgrade NPM on its own.)

This was done from our dev container. I have a feeling, dependabot keeps upgrading it, causing the flip floppiness.

petejohanson:
> > Also, can we use package-lock.json version 2 for this PR?
> 
> Agreed. It would be nice to stop completely rewriting the lock file every few changes. (Upgrade to Node 14.x which includes NPM 7.x, or just upgrade NPM on its own.)

Fixed this up to avoid the lock file version flip flopping.

app/boards/arm/nrfmicro/nrfmicro_11.zmk.yml
```diff
@@ -0,0 +1,10 @@
+file_format: "1"
+id: nrfmicro_11
+name: nRFMicro 1.1/1.2
+type: board
+arch: arm
+outputs:
+  - usb
+  - ble
+url: https://github.com/joric/nrfmicro/
+exposes: [pro_micro]
```
krikun98, line 10:
```suggestion
exposes: [pro_micro]

```

docs/docs/hardware.mdx
```diff
@@ -0,0 +1,26 @@
+---
+title: Supported Hardware
+sidebar_label: Supported Hardware
+---
+
+import HardwareList from "@site/src/components/hardware-list";
+import Metadata from "@site/src/data/hardware-metadata.json";
+
+With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets.
+That being said, there are currently only a few specific [boards](/docs/faq#what-is-a-board)/[shields](faq.md#what-is-a-shield) that have been written and tested by the ZMK contributors.
```
krikun98, line 10:
```suggestion
That being said, there are currently only a few specific [boards](/docs/faq#what-is-a-board)/[shields](faq.md#what-is-a-shield) that have been implemented and tested by the ZMK contributors.
```
I think the sentence flows a bit better like this.

docs/src/.gitignore
```diff
@@ -0,0 +1 @@
+hardware-metadata.d.ts
```
krikun98, line 1:
```suggestion
hardware-metadata.d.ts

```

docs/src/data/.gitignore
```diff
@@ -0,0 +1 @@
+hardware-metadata.json
```
krikun98, line 1:
```suggestion
hardware-metadata.json

```

schema/hardware-metadata.schema.json
```diff
@@ -0,0 +1,261 @@
...
+        "requires": {
+          "$ref": "#/$defs/interconnects"
+        },
+        "exposes": {
+          "$ref": "#/$defs/interconnects"
+        }
+      }
+    }
+  }
+}
```
krikun98, line 261:
```suggestion
}

```

docs/docs/development/hardware-metadata-files.md
```diff
@@ -0,0 +1,111 @@
...
+  - pro_micro
+exposes:
+  - i2c_oled
+```
+
+### Features
+
+Boards and shields should document the sets of hardware features found on them using the `features` array. There is a fixed enum of possible values to use here, which will be expanded over time. The current set of possible `features` values is:
+
+- `keys` - Any board or shield that contains keyboard keys should include this feature. It is a central feature used to determine if we have a "complete combination" for ZMK to produce a keyboard firmware when performig setup.
```
krikun98, line 93:
```suggestion
- `keys` - Any board or shield that contains keyboard keys should include this feature. It is a central feature used to determine if we have a "complete combination" for ZMK to produce a keyboard firmware when performing setup.
```

docs/.prettierignore
```diff
@@ -1,3 +1,5 @@
 node_modules
 build
-.docusaurus
\ No newline at end of file
+.docusaurus
+*.mustache
+hardware-metadata.json
```
krikun98, line 5:
```suggestion
hardware-metadata.json

```

app/boards/interconnects/pro_micro/pro_micro.zmk.yml
```diff
@@ -0,0 +1,10 @@
+file_format: "1"
+id: pro_micro
+name: Pro Micro
+type: interconnect
+url: https://www.sparkfun.com/products/12640
+manufacturer: SparkFun
+description: |
+  The SparkFun Pro Micro grew popular as a low cost ATmega32U4 board with sufficient GPIO and peripherals
+  to work for many keyboard needs. Since the original Pro Micro, many pin compatible boards have appeared,
+  with various changes or improvements, such as the Elite-C w/ USB-C, nice!nano with nRF52840 wireless.
```
krikun98, line 10:
```suggestion
  with various changes or improvements, such as the Elite-C w/ USB-C, nice!nano with nRF52840 wireless.
  
```

.github/workflows/hardware-metadata-validation.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: West init
+        run: west init -l app
+      - name: Update modules (west update)
+        run: west update
+      - name: Export Zephyr CMake package (west zephyr-export)
+        run: west zephyr-export
+      - name: Validate Hardware Metadata
+        run: |
+          cd app
+          west metadata check
```
krikun98, line 35:
```suggestion
          west metadata check
          
```

krikun98 (commented):
Mostly a bunch of newlines and a typo.

docs/src/hardware-metadata-collection-plugin/index.js
```diff
@@ -0,0 +1,33 @@
+/* TODO: COPYRIGHT AND LICENSE HEADER */
```
Nicell, line 1: License! (And in a couple other files)

Nicell (commented):
This is looking like an almost ready first pass on the metadata. Very exciting! Let me know if I can help at all by helping to make some of the initial `zmk.yml` files.

docs/src/hardware-metadata-collection-plugin/index.js
petejohanson: Added.

app/boards/shields/cradio/cradio.zmk.yml
```diff
@@ -0,0 +1,12 @@
+file_format: "1"
+id: cradio
+name: Cradio/Ferris Sweep
```
dxmh, line 3: At some point, [Sweep](https://github.com/davidphilipbarr/Sweep) dropped "Ferris" from the name to avoid potential confusion with the original [Ferris](https://github.com/pierrechevalier83/ferris). It's now referred to simply as Sweep in its documentation: https://github.com/davidphilipbarr/Sweep.

For consistency, and given that original Ferris will also be in ZMK someday with https://github.com/zmkfirmware/zmk/pull/642, would it make sense to go with simply "Sweep" here also?

```suggestion
name: Cradio/Sweep
```

petejohanson: Good catch. Fixed.

app/boards/shields/jian/jian.zmk.yml
```diff
@@ -0,0 +1,12 @@
+file_format: "1"
+id: jian
+name: Jian
+type: shield
+url: https://github.com/KGOH/Jian-Info
+requires: [pro_micro]
+features:
+  - keys
+  - underglow
```
krikun98, line 9: The Jian doesn't feature underglow, only single-color backlighting (which, AFAIK, isn't supported yet).

petejohanson: Thanks. Fixed.

Nicell (approved):
🎉

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+interface HardwareListProps {
+  items: HardwareMetadata[];
+}
+
+type ElementOrString = JSX.Element | string;
+
+function itemHasMultiple(item: HardwareMetadata) {
+  return (
+    (item.type == "board" || item.type == "shield") &&
+    (item.siblings?.length || 1) > 1
```
joelspadin, line 19:
```suggestion
    item.siblings?.length > 1
```
The above code works, but `undefined > 1` is also false, so you could simplify this

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+  );
+}
+
+function itemIds(item: HardwareMetadata) {
+  if (item.type == "board" || item.type == "shield") {
+    let nodes = (item.siblings || [item.id])
```
joelspadin, line 25:
```suggestion
    let nodes = (item.siblings ?? [item.id])
```
Kinda pedantic, but IMO ?? makes it more obvious you're using a fallback rather than a boolean operator

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+      <a href={item.url}>{item.name}</a> (
+      {TYPE_LABELS[item.type][itemHasMultiple(item) ? "plural" : "singular"]}{" "}
+      {itemIds(item)})
+    </li>
+  );
+}
+
+interface InterconnectDetails {
+  interconnect?: Interconnect;
+  boards: Array<Board>;
```
joelspadin, line 58:
```suggestion
  boards: Board[];
```
Not necessary to change these, but in my experience `T[]` is more conventional than `Array<T>` in the TS code I've seen.

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+    interconnects: Record<string, InterconnectDetails>;
+  } = items.reduce(
+    (agg, hm) => {
+      switch (hm.type) {
+        case "board":
+          if (hm.features?.includes("keys")) {
+            agg.onboard.push(hm);
+          } else if (hm.exposes) {
+            hm.exposes.forEach((element) => {
+              let ic = agg.interconnects[element] || {
```
joelspadin, line 103:
```suggestion
              let ic = agg.interconnects[element] ?? {
```
see earlier comment on ??

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+function HardwareList({ items }: HardwareListProps) {
+  let grouped: {
+    onboard: Array<Board>;
+    interconnects: Record<string, InterconnectDetails>;
+  } = items.reduce(
+    (agg, hm) => {
+      switch (hm.type) {
+        case "board":
+          if (hm.features?.includes("keys")) {
```
joelspadin, line 99: I think I'd find this a lot easier to follow if you broke these cases out into separate functions, e.g.

```ts
switch (hm.type) {
  case 'board':
    groupBoard(hm, agg);
    break;

  case 'shield':
    groupShield(hm, agg);
    break;
  // ...
}

```

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+              ic.boards.push(hm);
+              agg.interconnects[element] = ic;
+            });
+          } else {
+            console.error("Board without keys or interconnect");
+          }
+          break;
+        case "shield":
+          hm.requires.forEach((id) => {
+            let ic = agg.interconnects[id] || { boards: [], shields: [] };
```
joelspadin, line 116:
```suggestion
            let ic = agg.interconnects[id] ?? { boards: [], shields: [] };
```

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+            ic.shields.push(hm);
+            agg.interconnects[id] = ic;
+          });
+          break;
+        case "interconnect":
+          let ic = agg.interconnects[hm.id] || { boards: [], shields: [] };
```
joelspadin, line 122:
```suggestion
          let ic = agg.interconnects[hm.id] ?? { boards: [], shields: [] };
```

docs/src/components/hardware-list.tsx
```diff
@@ -0,0 +1,160 @@
...
+      <h2>Keyboards</h2>
+      <h3>Onboard Controller Boards</h3>
+      <p>
+        Keyboards with onboard controllers are single PCBs that contain all the
+        components of a keyboard, including the controller chip, switch
+        footprints, etc.
+      </p>
+      <ul>
+        {grouped["onboard"]
+          .sort((a, b) => a.name.localeCompare(b.name))
```
joelspadin, line 143: If we ever start getting into things with double digit version numbers, we might want to pull in something like https://www.npmjs.com/package/string-natural-compare to sort numbers embedded in strings numerically instead of lexicographically.

docs/src/hardware-schema-typescript-plugin/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+var PrebuildPlugin = require("prebuild-webpack-plugin");
+const fs = require("fs");
+const { compileFromFile } = require("json-schema-to-typescript");
+
+function generateHardwareMetadataTypescript() {
+  compileFromFile("../schema/hardware-metadata.schema.json")
+    .then((ts) => fs.writeFileSync("src/hardware-metadata.d.ts", ts))
+    .catch((e) => console.error(e));
+}
```
joelspadin, line 15: It looks like you can return a promise from the `build` function: https://github.com/hashicorp/prebuild-webpack-plugin/blob/d4dc160fcbe9ef5a66068169331bbf1c0969cc2f/index.js#L84

```suggestion
function generateHardwareMetadataTypescript() {
  return compileFromFile("../schema/hardware-metadata.schema.json")
    .then((ts) => fs.writeFileSync("src/hardware-metadata.d.ts", ts))
    .catch((e) => console.error(e));
}
```
or
```suggestion
async function generateHardwareMetadataTypescript() {
  try {
    const ts = await compileFromFile("../schema/hardware-metadata.schema.json");
    fs.writeFileSync("src/hardware-metadata.d.ts", ts));
  } catch (e) {
    console.error(e);
  }
}
```

Also, is `console.error()` sufficient on failure, or do we actually want to throw an exception though? If throwing an exception is better;

```suggestion
async function generateHardwareMetadataTypescript() {
  const ts = await compileFromFile("../schema/hardware-metadata.schema.json");
  fs.writeFileSync("src/hardware-metadata.d.ts", ts));
}
```

docs/src/setup-script-generation-plugin/index.js
```diff
@@ -0,0 +1,66 @@
...
+      "utf8"
+    );
+    const script = Mustache.render(templateBuffer, data);
+    fs.writeFileSync("static/setup.sh", script);
+    const powershellTemplateBuffer = fs.readFileSync(
+      "src/templates/setup.ps1.mustache",
+      "utf8"
+    );
+    const psScript = Mustache.render(powershellTemplateBuffer, data);
+    return fs.writeFileSync("static/setup.ps1", psScript);
```
joelspadin, line 49: Minor: not much of an issue with only two scripts, but you could define an array of the scripts to generate and loop through them, since the render code is the same for them all aside from template file and output file.

app/scripts/west_commands/metadata.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2021 The ZMK Contributors
```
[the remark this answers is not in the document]
petejohanson: Fixed.

docs/src/hardware-schema-typescript-plugin/index.js
petejohanson: Indeed. Much cleaner. Updated. Thanks!

PR #884, opened 2021-07-27
Fix power management to trigger low power devices

### LINKED ISSUE #882: Device power management no long works since Zephyr 2.5.0 upgrade
It seems the way we were updating device power management state no longer works in 2.5.0. This results in external VCC not turning off during sleep.

~~It looks like `pm_policy_low_power_devices` is no longer valid.~~
On further inspection it seems that `pm_policy_low_power_devices` is still ok, but the issue is that `pm_low_power_devices()` isn't called by the `PM_STATE_SOFT_OFF` state, where it's called by the other standby/suspend states.

Relevant docs: https://docs.zephyrproject.org/latest/reference/power_management/index.html#device-model-with-power-management-support

See it being removed from nrf connect SDK: https://github.com/nrfconnect/sdk-nrf/commit/2ad0361fb4aaba7ee22f0146253b9740a0aedcd3

Nicell:
Fixes #882 

`PM_STATE_SOFT_OFF` doesn't tell devices to go into low power mode, so for now, I've added a line before going into sleep to force the standby state, which does tell devices to go into low power mode.

Related Zephyr fix: https://github.com/zephyrproject-rtos/zephyr/pull/37256

I've tested this with a nice!nano v2, and now VCC properly powers off during sleep.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -267,7 +267,7 @@ choice SYS_PM_POLICY
 	default PM_POLICY_APP
 endchoice
 
-config DEVICE_POWER_MANAGEMENT
+config PM_DEVICE
 	default y
 
 config ZMK_IDLE_SLEEP_TIMEOUT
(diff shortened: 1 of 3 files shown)


petejohanson (approved):
Thanks!

PR #888, opened 2021-07-31
Debounce rework

joelspadin:
Depends on #878


app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-matrix.yaml
```diff
@@ -15,8 +15,22 @@ properties:
     type: phandle-array
     required: true
   debounce-period:
+    type: int
+    required: false
+    deprecated: true
+    description: Deprecated. Use debounce-on-ms and debounce-off-ms instead.
```
lyallcooper, line 21:
```suggestion
    description: Deprecated. Use debounce-down-ms and debounce-up-ms instead.
```

lyallcooper:
First impressions: with the default config, bouncing seems to actually be worse on my kb. I'm frequently getting bounces on keydown whereas that was pretty rare before, and bounces on keyup still happen.

That being said, being able to set the down/up periods separately is a big win, especially since it seems my kb has more of a keyup bounce issue. So with some tuning I think I've landed on something that is still usable latency-wise but doesn't bounce, which is really all I need.

So in summary I'm super happy (thank you!) but think it's probably worth testing this with some less bounce-prone keyboards to figure out if this algo works generally for folks and what a sane default config is. In the meantime I might play around with `debounce.c` and see if any other approaches produce better results at shorter debounce intervals.

app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-matrix.yaml
joelspadin: Nice catch. I renamed those properties but forgot to update here.

joelspadin:
> With debounce-up-msset to 127, I can consistently get keypresses that are as short as ~30ms. If I change the config struct to allow for a value of 1023 the shortest keypress I can get is a bit more than 200ms. So at least with my board and its current config, every ~5 "ms" of debounce config equals one real ms of debounce. This explains why I was getting so much bounce with the default values.

Interesting. I hadn't yet verified the timings. It sounds like it might be scanning faster than the 1 ms per scan I intended.

app/drivers/kscan/debounce.c
```diff
@@ -0,0 +1,38 @@
...
+                                    const struct debounce_config *config) {
+    return state->pressed ? config->debounce_up_ms : config->debounce_down_ms;
+}
+
+bool debounce_update(struct debounce_state *state, const bool active,
+                     const struct debounce_config *config) {
+    // This uses a variation of the integrator debouncing described at
+    // https://www.kennethkuhn.com/electronics/debounce.c
+    // We may have different thresholds for press and release, so we reset the
+    // counter and begin counting up again every time the state flips.
```
lyallcooper, line 19: Since even with the comment in the link as reference it took me a little bit to figure out how this works, might be worth adding a little more description to the comment here. This is how I've been thinking about it if it's helpful:

> There is a counter that represents how close we are to transitioning away from the current state--it starts at 0. On every tick, if the new state (`active`) is different from the current state we increment the counter, otherwise we decrement it. When the counter reaches the configured threshold (can be different for key down vs. up) we transition to the new state and reset the counter back to zero.

joelspadin:
I think I see the problem. I'm setting a timeout for an absolute time of 1 ms, not an absolute time of 1 ms from now. Therefore the timeout expires immediately and it just scans as fast as possible.

joelspadin:
@lyallcooper Try it now. It should be properly spacing scans at 1 ms apart now.

joelspadin:
(Build is unsurprisingly broken with polling because I haven't tested that at all yet.)

lyallcooper:
Yep that seems to have done it. Timings line up as they should for me. As a result the default config is much less bouncy, so probably good enough for boards that are well constructed, unlike mine 😅

joelspadin:
Thanks for testing it!

Fixed and tested with polling now too, plus I added a bit more documentation to the debounce code.

app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-matrix.yaml
```diff
@@ -15,8 +15,22 @@ properties:
...
+    default: 5
+    description: Debounce time for key press in milliseconds. Must be <= 127. Use 0 for eager debouncing.
+  debounce-up-ms:
     type: int
     default: 5
+    description: Debounce time for key release in milliseconds. Must be <= 127.
+  poll-period-ms:
+    type: int
+    default: 10
+    description: Time between reads in milliseconds when polling is enabled
```
Nicell, line 33: Can we make Kconfig options to override these? We definitely don't _need_ Kconfig options to override these, especially since it would globally override all kscans, but they would also be more accessible for people working with zmk-config repos where they may not know how to overlay the kscan definition. Alternatively we could write up documentation, but I generally see personally that ZMK users are much more comfortable updating their .conf files than updating DT. I'd like to hear others' thoughts though!

lyallcooper: +1. It's not hard to override the device tree stuff once you know how to do it but it's definitely not well documented right now. Plus I'm generally in favor of putting config all in one place whenever possible.

joelspadin: After attempting to write documentation that explained how to find the correct node to modify, I'm on board with this. I couldn't find a simple way to create an optional `int` setting in Kconfig though, so I'm not sure if what I came up with is the best way to do this.

joelspadin:
Added global Kconfig options for debouncing and added documentation for them.

Renamed `debounce-down-ms` and `debounce-up-ms` to `debounce-press-ms` and `debounce-release-ms` for consistency with the rest of the code.

Added `debounce_is_pressed()` so that you don't need to access the debounce state struct directly for anything. That way we can modify `debounce_state` later without needing to change the scanning code.

joelspadin:
Tweaked the scheduling a bit to eliminate any timer drift. Instead of setting the scheduled time for the next scan to be relative to the time at the start of the current scan, it is now relative to when we scheduled the current scan to run.

joelspadin:
Prettier check is failing because encoders.md isn't formatting correctly. I didn't change that here. I assume that was caused in https://github.com/zmkfirmware/zmk/commit/47109641d8d257ee8124fecd5977d618fa5839b4

Nicell:
I've been testing this PR for about a week now, and the debouncing seems much more reliable. Great job on this @joelspadin. I'm looking to do some power profiling soon to see the impact this work has on battery life. Hopefully, it is minimal.

app/drivers/kscan/debounce.h
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @returns whether the switch is latched as pressed.
+ */
+bool debounce_is_pressed(const struct debounce_state *state);
+
+/**
+ * @returns the "changed" flag and clears it.
+ */
+bool debounce_did_change(struct debounce_state *state);
```
joelspadin, line 60: I'm not sure if it makes more sense to have one function that reads and clears the flags, or separate read and clear functions. At least in the matrix driver, one function that does both is easier to use, though it might be confusing if you tried to call `debounce_did_change()` twice in a row and got different results.

joelspadin:
Added another commit on top that slightly reworks how the kscan driver reads things. Previously it did:

1. Scan the matrix and store the results in an array
2. Loop through the results and debounce. If anything changes, send a key event.

Now it does:

1. Scan the matrix and debounce the results. If anything changes, set a flag on the debounce state.
2. Loop through the states. If a "changed" flag is set, send a key event and clear the flag.

That gets rid of a parallel array, and it gives us 14 bits for the counter (which isn't terribly useful now, but it could be later if we wanted to scan faster than 1000 Hz, plus we don't need to document a max value for it since no usable debounce period could possibly need 14 bits right now).

app/drivers/kscan/debounce.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <sys/util.h>
+
+#define DEBOUNCE_COUNTER_BITS 14
+#define DEBOUNCE_COUNTER_MAX BIT_MASK(DEBOUNCE_COUNTER_BITS)
```
lyallcooper, line 14: You mentioned not needing to document the max debounce period since 14 bits is so much, but AFAICT it'll still cause a pretty cryptic build error to exceed that (even though it wouldn't make sense to) so might want to either add back the documentation or change the check. Maybe taking the min of the config value and `2^14 - 1` could avoid the issue.

joelspadin: Sure, it is a somewhat cryptic error (it just says the value is too large but doesn't say what the max is), but if you are trying to set a debounce time of over 16 seconds I have questions.

app/drivers/kscan/debounce.c
```diff
@@ -0,0 +1,52 @@
...
+    const uint32_t flip_threshold = get_threshold(state, config);
+
+    if (state->counter < flip_threshold) {
+        state->counter++;
+        return;
+    }
+
+    state->pressed = !state->pressed;
+    state->counter = 0;
+    state->changed = true;
```
elagil, line 37: I would suggest not to return in the middle of the function, but to employ else statements instead.

```suggestion
    if (active == state->pressed) {
        if (state->counter > 0) {
            state->counter--;
        }
    } else {
        const uint32_t flip_threshold = get_threshold(state, config);
  
        if (state->counter < flip_threshold) {
            state->counter++;
        } else {
            state->pressed = !state->pressed;
            state->counter = 0;
            state->changed = true;
        }
    }
```

app/drivers/kscan/debounce.h
elagil: I would leave it as is and potentially rename it to something like `debounce_get_and_clear_change_flag` for avoiding confusion. There is probably a better name than what I suggested.

elagil (approved):
Some minor comments only. I tested your changes for a few days without negative effects. Debouncing is effective.

petejohanson:
@joelspadin Needs a rebase on `main` now that the initial refactor is in.

joelspadin:
> @joelspadin Needs a rebase on `main` now that the initial refactor is in.

Done.

app/drivers/kscan/debounce.c
joelspadin: > As an example, consider a function that uses dynamic memory allocation and you free at the end, a premature return skips the freeing step and breaks the functionality. You would have to replicate it before each of the exit points of your function.

That is a very good argument (and one I agree with) for only having a single return point in a function that allocates resources, but this function doesn't (and likely never will).

This function is simple enough that I think it's pretty readable either way, but I personally try to avoid having many levels of nested control structures. With a return, I know the function is done and don't have to check if there's any code after the if/else block.

elagil: > but this function doesn't (and likely never will)

That is true. The function is indeed simple enough to be readable this way.

I agree that the early return makes handling of if/else blocks simpler and possibly more robust.

joelspadin:
The failed checks do not appear to be related to this PR. Netlify is complaining that the package-lock.json is using version 2 while it's trying to use NPM v6, while this PR does not change the lock file.

app/drivers/kscan/Kconfig
```diff
@@ -14,6 +14,27 @@ config ZMK_KSCAN_MATRIX_POLLING
 config ZMK_KSCAN_DIRECT_POLLING
 	bool "Poll for key event triggers instead of using interrupts on direct wired boards."
 
+config ZMK_KSCAN_GLOBAL_DEBOUNCE_PRESS
+	bool
```
petejohanson, line 18: This should get a string label as well, e.g.


```suggestion
	bool "Global kscan debounce support"
```

app/drivers/kscan/debounce.c
petejohanson: Good nuance here, would agree with the opinion about *this* function that the early returns lead to clearer code.

app/drivers/kscan/debounce.c
```diff
@@ -0,0 +1,52 @@
...
+}
+
+bool debounce_is_active(const struct debounce_state *state) {
+    return state->pressed || state->counter > 0;
+}
+
+bool debounce_is_pressed(const struct debounce_state *state) { return state->pressed; }
+
+bool debounce_did_change(struct debounce_state *state) {
```
petejohanson, line 46: Is there a reason we don't just return this from the update function directly? This two phase, and the *query* also being a *mutate* can easily lead to bugs later.

app/drivers/kscan/debounce.h
```diff
@@ -0,0 +1,60 @@
...
+
+struct debounce_state {
```
petejohanson, line 16: Are you trying to pack this to avoid extra storage overhead?

app/drivers/kscan/debounce.h
```diff
@@ -0,0 +1,60 @@
...
+    /** Duration a switch must be pressed to latch as pressed. */
+    uint32_t debounce_press_ms;
+    /** Duration a switch must be released to latch as released. */
+    uint32_t debounce_release_ms;
+};
+
+/**
+ * Debounces one switch.
+ *
+ * @note For the timings in "config" to be accurate, this currently assumes that
```
petejohanson, line 32: Don't love this gotcha. Can we pass a duration to this function as well for "time elapsed"?

app/drivers/kscan/debounce.h
petejohanson: See my other comment. I'm not a fan of functions that query and mutate like this.

petejohanson (commented):
Some initial thoughts, will review the rest of the kscan changes later.

app/drivers/kscan/debounce.c
joelspadin: I wanted to preserve the way the scanning currently works, so the scan is all done all at once, and then the results are processed all at once. If you sent the key press event within the scan loop, then long-running behaviors could make the timing inconsistent.

joelspadin: Also, if I returned from the update function directly, and we wanted to keep all the events being sent at the end, then the scan function needs a parallel array of bools the same size as the matrix, which uses an extra byte per key compared to using a single bit of the debounce_state struct.

app/drivers/kscan/debounce.h
joelspadin: Correct. I figured I might as well use as little space as possible for the best chance of being able to run things on cheaper processors.

app/drivers/kscan/debounce.h
joelspadin: Yes. I assume we would want to be able to scan faster than 1000 Hz, since having something capable of running Zephyr but not capable of scanning at 1000 Hz seems unlikely?

If we want that, then we can make the counter be a fixed point number of milliseconds, but then we need to decide how many bits to use for the fractional component. Assuming we kept the 14-bit counter, some reasonable splits:

Integer bits | Fractional bits | Max debounce time (ms) | Min timer resolution (ms)
------------|------------------|------|-----| 
14 (current) | 0 | 16,383 | 1 |
10 | 4 | 1023 | 0.0625
9 | 5 | 511 | 0.0313
8 | 6 | 255 | 0.0156
7 | 7 | 127 | 0.0078
6 | 8 | 63 | 0.0039
5 | 9 | 31 | 0.0020

Then the scanning routine needs to measure how long it's been in the appropriate fraction of a millisecond instead of just counting scans. This could lead to rounding errors depending on your scan rate though. 

For an exaggerated example, say you chose to scan at 3000 Hz with a minimum timer resolution of 1/16 ms and a debounce period of 1 ms. You would expect it to take 3 scans to debounce a key press since each scan would be 1/3 ms apart, but that rounds down to 5/16. After three scans the counter is now 15/16 ms, so it actually would take 4 scans. 

Maybe this is all too minor to matter, but the reason I did it the way it is now is because I didn't want to have to deal with that. Any suggestions on how to do this better are welcome.

app/drivers/kscan/Kconfig
joelspadin: I think I forgot to remove this. It doesn't appear to be used anywhere.

Nicell:
I just ran some power profiling on this PR. As expected the typing power draw is much higher.

With a single key held down:
 - on main I'm getting ~300uA, ~150uA over idle
 - on this PR I'm getting ~750uA, ~600uA over idle

Considering how little time keys are held down while using a keyboard, this seems reasonable. The ZMK power profiler should probably be updated, but it should have a total remeasure anyways, as ZMK has changed a lot since its readings were originally captured.

PR #889, opened 2021-08-02
feat(ci): Use metadata for builds

Nicell:
Refactors the build workflow to use hardware metadata to accomplish the following:
 - Nightly, build every single onboard MCU board and board+shield combo possible
 - Whenever app/** is changed (minus app/boards), run a short, premade core builds list
 - Whenever app/boards is changed, build every board+shield combo that's affected by the changes

Here's a nightly build: https://github.com/Nicell/zmk/actions/runs/1519228121
Here's a boards change run: https://github.com/Nicell/zmk/actions/runs/1517831312
Here's a core run: https://github.com/Nicell/zmk/actions/runs/1517789825

Todo:

- [x] make the final compile-matrix output remove all duplicates includes.
- [x] add cmake-args to artifact name in some fashion
- [x] figure out 256 matrix job limit (do one job per board rather than one job per board + shield combo)
- [x] get-changed-files seems to fail after squashes and PRs (alternative action with fixes: https://github.com/Ana06/get-changed-files)
- [x] don't inline the core build list (use a separate .yml file)
- [ ] make sure to build setting-reset shield (update link to them on docs)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,75 +9,19 @@ on:
     paths:
       - ".github/workflows/build.yml"
       - "app/**"
+  schedule:
+    - cron: '22 4 * * *'
 
 jobs:
   build:
+    if: ${{ always() }}
     runs-on: ubuntu-latest
     container:
       image: docker.io/zmkfirmware/zmk-build-arm:2.5
+    needs: compile-matrix
     strategy:
       matrix:
-        board:
-          - bluemicro840_v1
-          - nice_nano
-          - nice_nano_v2
-          - nrfmicro_13
-          - proton_c
-        shield:
-          - bfo9000_left
-          - bfo9000_right
-          - boardsource3x4
-          - corne_left
-          - corne_right
-          - cradio_left
-          - cradio_right
-          - crbn
-          - eek
-          - helix_left
-          - helix_right
-          - iris_left
-          - iris_right
-          - jian_left
-          - jian_right
-          - jorne_left
-          - jorne_right
-          - kyria_left
-          - kyria_right
-          - lily58_left
-          - lily58_right
-          - microdox_left
-          - microdox_right
-          - nibble
-          - qaz
-          - quefrency_left
-          - quefrency_right
-          - reviung41
-          - romac
-          - romac_plus
-          - settings_reset
-          - sofle_left
-          - sofle_right
-          - splitreus62_left
-          - splitreus62_right
-          - tg4x
-          - tidbit
-        cmake-args: [""]
-        include:
-          - board: bdn9_rev2
-          - board: dz60rgb_rev1
-          - board: nrf52840_m2
-            shield: m60
-          - board: planck_rev6
-          - board: proton_c
-            shield: clueboard_california
-          - board: nice_nano_v2
-            shield: kyria_left
-            cmake-args: -DCONFIG_ZMK_DISPLAY=y
-            skip-archive: true
-          - board: nice_nano_v2
-            shield: kyria_right
-            cmake-args: -DCONFIG_ZMK_DISPLAY=y
-            skip-archive: true
+        include: ${{ fromJSON(needs.compile-matrix.outputs.include-list) }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


app/scripts/west_commands/metadata.py
```diff
@@ -0,0 +1,51 @@
...
+                except yaml.YAMLError as exc:
+                    failure = True
+                    print("Failed loading metadata yaml: " + file)
+                    print(exc)
+                except ValidationError as vexc:
+                    failure = True
+                    print("Failed validation of: " + file)
+                    print(vexc)
+
+        exit(1 if failure else 0)
```
joelspadin, line 51: exit() is supposed to only be used from within the Python interactive interpreter from what I understand. You probably want sys.exit() instead.

app/scripts/west_commands/metadata.py
```diff
@@ -0,0 +1,51 @@
...
+        schema = json.load(open("../schema/hardware-metadata.schema.json", 'r'))
+        failure = False
+        for file in glob.glob("boards/**/*.zmk.yml", recursive=True):
+            print("Validating: " + file)
+            with open(file, 'r') as stream:
+                try:
+                    metadata = yaml.safe_load(stream)
+                    ret = validate(metadata, schema)
+                except yaml.YAMLError as exc:
+                    failure = True
```
joelspadin, line 43: IMO it would be simpler to just
```python
print(exc)
sys.exit(1)
```
instead of having a common exit point. You've already hit an error, so you should probably bail immediately.

Alternatively, you could wrap all of this up in another function and have it re-raise the exceptions with added info, then have a single handler that prints the error and exits. It might look something like this:

```python
class MetadataError(Exception):
    pass

def do_run(self, args, unknown_args):
    try:
        self.validate_files()
    except MetadataError as e:
        print(e)
        sys.exit(1)
        
def validate_files(self):
    # ...
    try:
        # ...
    except yaml.YAMLError as e:
        # Use .format() instead of f-strings if we care about Python < 3.6
        raise MetadataError(f'Failed loading YAML file "{file}": {e}') from e
    except ValidationError as e:
        raise MetadataError(f'Failed validation of file "{file}": {e}') from e
```
...which I kinda like because all the logic for printing an error is in one spot instead of spread across multiple exception handlers.

app/scripts/west_commands/metadata.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def do_run(self, args, unknown_args):
+        schema = json.load(open("../schema/hardware-metadata.schema.json", 'r'))
+        failure = False
+        for file in glob.glob("boards/**/*.zmk.yml", recursive=True):
+            print("Validating: " + file)
+            with open(file, 'r') as stream:
+                try:
+                    metadata = yaml.safe_load(stream)
+                    ret = validate(metadata, schema)
```
joelspadin, line 41: `ret` appears unused?

docs/docs/development/hardware-metadata-files.md
```diff
@@ -0,0 +1,111 @@
+---
+title: Hardware Metadata Files
+---
+
+## Overview
+
+ZMK leverages an additional metadata YAML file for all boards and shields to provide high level information about the hardware to be leveraged by setup scripts/utilities, proper inclusion in the website hardware list, etc.
```
joelspadin, line 7: Super minor: using "leverage" twice makes this read a bit awkwardly to me. Maybe just use "use" and/or "to be read by setup scripts" to keep things easy to understand?

(If you change this, Ctrl+F "leverage" since it's used in a few other places too.)

joelspadin:
Whoops. I am on the wrong PR. Those comments apply to #883.

app/scripts/west_commands/metadata.py
petejohanson: Oh, the point is to continue, so we can check *all* the files. Then if any fail... we return a non-zero exit code.

petejohanson: We specifically *don't* want to exit early.

.github/workflows/build.yml
```diff
@@ -130,3 +74,359 @@ jobs:
...
+            let include = [];
+            
+            coreCoverage.boards.forEach(b => {
+              coreCoverage.shields.forEach(s => {
+                include.push({
+                  board: b,
+                  shield: s
+                });
+              });
+            });
```
petejohanson, line 171: The purist in me says, do it functionally, darn it!

```suggestion
            let include = coreCoverage.boards.flatMap(board => coreCoverage.shields.map(shield => ({ board, shield }));
```

(I may have the JS syntax slightly wrong)

.github/workflows/build.yml
```diff
@@ -130,3 +74,359 @@ jobs:
...
+
+            for (const bc of boardChanges) {
+              const globber = await glob.create(bc + "/*.zmk.yml");
+              const files = await globber.glob();
+
+              const aggregated = files.flatMap((f) =>
+                yaml.loadAll(fs.readFileSync(f, "utf8"))
+              );
+
+              aggregated.forEach(hm => {
```
petejohanson, line 210: Similar grumblings about using `forEach` here instead of mapping, but this also works fine.

petejohanson (commented):
Seems like a good start. I think in general, the code could be cleaned up, even if the you can't declare functions in github script, you could define lambdas and use them, to make the code clearer, e.g. `const mapBoard = (metadata)  => foo`.

.github/workflows/build.yml
```diff
@@ -104,29 +48,365 @@ jobs:
...
+              } catch (e) {
+                console.error(`Failed to build or upload ${{ matrix.board }} ${shieldArgs.shield} ${shieldArgs['cmake-args']}`);
+                console.error(e);
+                error = true;
+              }
+            }
+
+            if (error) {
+              throw new Error('Failed to build one or more configurations');
+            }
```
Nicell, line 100: Throwing this error later does let us build the rest of the combinations if one fails, but then it's quite a pain to search through the logs after for the error since they can contain potentially 100 builds.

.github/workflows/build.yml
```diff
@@ -104,29 +48,365 @@ jobs:
...
+                  .map(extension => "build/zephyr/zmk." + extension)
+                  .filter(path => fs.existsSync(path));
+
+                const rootDirectory = 'build/zephyr';
+                const options = {
+                    continueOnError: true
+                }
+                const artifactName = `${{ matrix.board }}${shieldArgs.shield ? '-' + shieldArgs.shield : ''}-zmk`;
+
+                await artifactClient.uploadArtifact(artifactName, files, rootDirectory, options);
```
Nicell, line 90: Currently, this throws an error if any file fails to upload. I believe our current build script doesn't error if the upload fails. Is it worth wrapping this in a try block too?

.github/workflows/build.yml
```diff
@@ -104,29 +48,355 @@ jobs:
...
+
+            const execSync = require('child_process').execSync;
+            
+            const buildShieldArgs = JSON.parse(`${{ matrix.shieldArgs }}`);
+
+            let error = false;
+
+            for (const shieldArgs of buildShieldArgs) {
+              try {
+                const output = execSync(`west build -s app -p -b ${{ matrix.board }} -- ${shieldArgs.shield ? '-DSHIELD=' + shieldArgs.shield : ''} ${shieldArgs['cmake-args'] || ''}`);
```
petejohanson, line 74: We should group the log by each entry we build, to make it easier to see each combination we build. See https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines

.github/workflows/build.yml
```diff
@@ -104,29 +48,355 @@ jobs:
...
+
+                const rootDirectory = 'build/zephyr';
+                const options = {
+                    continueOnError: true
+                }
+                const artifactName = `${{ matrix.board }}${shieldArgs.shield ? '-' + shieldArgs.shield : ''}-zmk`;
+
+                await artifactClient.uploadArtifact(artifactName, files, rootDirectory, options);
+              } catch (e) {
+                console.error(`Failed to build or upload ${{ matrix.board }} ${shieldArgs.shield} ${shieldArgs['cmake-args']}`);
```
petejohanson, line 92: Perhaps better to use: https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-error-message

petejohanson (changes requested):
This seems really close! A few tweaks to make the experience a tad nicer needed before we merge.

Nicell:
@petejohanson Improved the logging with groups/errors as suggested. Also simplified the combining code using your suggestion, thanks. Let me know what you think now. I still want to add settings-reset to the nightly build and also add cmake args to the artifacts, but I'm not sure the best way to do either if you have suggestions. Have an idea for the settings-reset at the very least.

petejohanson:
> 

@Nicell Perhaps for the cmake-args case, we expect you to *also* pass a `build-nickname` or `build-name` that's used, and if someone *fails* to provide that, just do something like (pseudo-code): `build-args-${m5dsum(args-string)}` to give it a unique name?

Nicell:
@petejohanson I added nicknames, you can see it on the core build of this PR. Works well. I've simply removed the spaces from cmake-args as the alternate rather than doing a has, since it's more readable this way and just a fallback.

I had some thoughts about settings-reset I think we need to address, so I'm not sure if I'm going to add that to this PR. I think this is mostly ready to go then from my point of view.

.github/workflows/build.yml
```diff
@@ -104,29 +48,359 @@ jobs:
...
+
+            let error = false;
+
+            for (const shieldArgs of buildShieldArgs) {
+              try {
+                const output = execSync(`west build -s app -p -b ${{ matrix.board }} -- ${shieldArgs.shield ? '-DSHIELD=' + shieldArgs.shield : ''} ${shieldArgs['cmake-args'] || ''}`);
+
+                console.log(`::group::${{ matrix.board}} ${shieldArgs.shield} Build`)
+                console.log(output.toString());
+                console.log('::endgroup::');
```
petejohanson, line 78: We should put this endgroup into a `finally` block so it there are any exceptions logged in the `catch` it's part of the group.

petejohanson (changes requested):
One minor tweak needed, then good to go. Thanks!

petejohanson (approved):
Awesome, thanks!

PR #897, opened 2021-08-06
Add new sensor channels to battery.c

LOWPROKB:
This PR introduces a bunch of sensor channels that can be used by Zephyr-supported fuel gauge ICs on keyboards. It should still work fine if a board is using the local voltage divider driver, but expands the functionality for drivers that can provide more than just State Of Charge data.

**The gist of what the major changes consist of:**
1. The struct that currently holds the `zmk_battery_state_changed` event has been expanded to handle more data channels. The event is only raised after all the possible channel data is collected.
2. The code in batttery.c now checks to see if a channel is supported by the sensor, so that a "not supported" response from the driver does not constitute a failure.
3.  Errors while retrieving sensor channels now get counted and the total errors get returned to the main function call, rather than returning after the first error. This allow the debug to catch the state of all the channels, not just the ones prior to the error.

**I got stuck on one thing, and could use help from someone with better C skills than me:**

The sensor data gets returned from the driver in two parts. The whole integer part and the fractional part (the parts before and after the decimal in a float). These parts have to be combined back into one value again, and maybe rounded to two decimal places, so that the full value can be passed back to the `zmk_battery_state_changed` event in one piece. I just don't know how to achieve that. Right now I am simply returning the whole integer part. I have marked each place this needs to be changed with this comment: `/*combine val1 and val2 here*/`

### DIFF
--- a/app/include/zmk/events/battery_state_changed.h
+++ b/app/include/zmk/events/battery_state_changed.h
@@ -10,8 +10,40 @@
 #include <zmk/event_manager.h>
 
 struct zmk_battery_state_changed {
-    // TODO: Other battery channels
+    
+    /* Remaining capacity as a %age */
     uint8_t state_of_charge;
+
+    /* State of Health */
+    int16_t state_of_health;
+
+    /*  Current cell voltage in units of 1.25/16mV */
+    uint32_t voltage;
+
+    /* Average current in units of 1.5625uV / Rsense */
+    uint32_t current;
+
+    /* Standby current in mA? uA? */
+    int32_t current_standby;
+
+    /* Maximum Load Current in mA? uA? */
+    int32_t current_max_load;
+
+    /* Full charge capacity in 5/Rsense uA */
+    int32_t full_charge_capacity;
+
+    /* Remaining capacity in 5/Rsense uA */
+    int32_t remaining_charge_capacity;
+
+    int32_t nominal_available_capacity;
+
+    int32_t full_available_capacity;
+
+    /* Average power consumption in mA? uA? */
+    int32_t avg_power;
+
+    /* Internal temperature in units of 1/256 degrees C */
+    int32_t int_temp;
 };
 
 ZMK_EVENT_DECLARE(zmk_battery_state_changed);
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -110,7 +110,7 @@ static int zmk_battery_update(const struct device *battery) {
         LOG_DBG("The configured battery sensor does not support the channel: Voltage");
     } else {
         /* append this value to the battery state changed event */
-        bat_state_data.voltage = voltage.val1; /*combine val1" and "val2" here*/
+        bat_state_data.voltage = voltage.val1; /*combine val1 and val2 here*/
         LOG_DBG("-----> battery voltage: %d.%dV", voltage.val1, voltage.val2);
     }
 
@@ -125,7 +125,7 @@ static int zmk_battery_update(const struct device *battery) {
         LOG_DBG("The battery sensor does not support the channel: Average Current");
     } else {
         /* append this value to the battery state changed event */
-        bat_state_data.current = current.val1; /*combine val1" and "val2" here*/
+        bat_state_data.current = current.val1; /*combine val1 and val2 here*/
         LOG_DBG("-----> battery Average Current: %d.%dA", current.val1, current.val2);
     }
 
@@ -140,7 +140,7 @@ static int zmk_battery_update(const struct device *battery) {
         LOG_DBG("The battery sensor does not support the channel: Standby Current");
     } else {
         /* append this value to the battery state changed event */
-        bat_state_data.current_standby = current_standby.val1; /*combine val1" and "val2" here*/
+        bat_state_data.current_standby = current_standby.val1; /*combine val1 and val2 here*/
         LOG_DBG("-----> battery Standby Current: %d.%dA", current_standby.val1,
                 current_standby.val2);
     }
@@ -156,7 +156,7 @@ static int zmk_battery_update(const struct device *battery) {
         LOG_DBG("The battery sensor does not support the channel: Maximum Load Current");
     } else {
         /* append this value to the battery state changed event */
-        bat_state_data.current_max_load = current_max_load.val1; /*combine val1" and "val2" here*/
+        bat_state_data.current_max_load = current_max_load.val1; /*combine val1 and val2 here*/
         LOG_DBG("-----> battery Maximum Load Current: %d.%dA", current_max_load.val1,
                 current_max_load.val2);
     }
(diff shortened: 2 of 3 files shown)


app/src/battery.c
```diff
@@ -85,7 +287,7 @@ static int zmk_battery_init(const struct device *_arg) {
     int rc = zmk_battery_update(battery);
 
     if (rc != 0) {
-        LOG_DBG("Failed to update battery value: %d.", rc);
+        LOG_WRN("Failed to update %d battery value(s) which the configured sensor supports.", rc);
```
joelspadin, line 290: I'm not sure knowing how many channels failed is useful information. It would be much easier to know went wrong if you just logged which channels failed right after attempting to read them.

app/src/battery.c
```diff
@@ -25,40 +25,242 @@ static uint8_t last_state_of_charge = 0;
 uint8_t zmk_battery_state_of_charge() { return last_state_of_charge; }
 
 static int zmk_battery_update(const struct device *battery) {
-    struct sensor_value state_of_charge;
+    struct sensor_value state_of_charge,
+                        state_of_health,
```
joelspadin, line 29: Since you're writing to bat_state_data after each channel read, you don't need a separate variable for each channel. You can just have one sensor value and keep reusing it.

joelspadin (changes requested):

app/include/zmk/events/battery_state_changed.h
```diff
@@ -10,8 +10,40 @@
 #include <zmk/event_manager.h>
 
 struct zmk_battery_state_changed {
-    // TODO: Other battery channels
+
```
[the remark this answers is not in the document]
LOWPROKB: Yeah I know... I meant to go back into the driver and find out for sure what units were being output, but I can still do that. There are only a couple of channels I am using atm, which is why I have not deeply investigated the others. Wanted to get other people's eyes on this sooner rather than later.

app/src/battery.c
```diff
@@ -25,40 +25,242 @@ static uint8_t last_state_of_charge = 0;
 uint8_t zmk_battery_state_of_charge() { return last_state_of_charge; }
 
 static int zmk_battery_update(const struct device *battery) {
-    struct sensor_value state_of_charge;
+    struct sensor_value state_of_charge,
```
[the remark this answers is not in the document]
LOWPROKB: Most of this code derives from the Zephyr sample provided by one of the driver's authors for the TI BQ274xx chips. I tried to make is as compatible as possible with the ZMK existing code. I appreciate your opinion about the length of functions, but I do not have time to refactor it, and it is logical to be held in one function.

app/src/battery.c
LOWPROKB: This is true. the combined bat_state_data was added to this code after I had written the rest of it.

app/src/battery.c
LOWPROKB: you are right, its not useful... except the code that calls this function is expecting a 0 for success... so if that number is non-zero, the calling code handles the error. It seemed like a simple solution

LOWPROKB:
I should also mention that I have had a heck of a time migrating the code that talks to the BQ274xx driver to Zephyr 2.5.
It seems that `sensor_sample_fetch(battery)` worked fine in the older version of Zephyr, but in 2.5, it does not (for me anyway).
I am still digging into it, and looking at all the PR's in Zephyr that affect this driver, but right now, I can only fetch individual channels before getting the values.
This affects this PR in that I would have to add a  `sensor_sample_fetch_chan(battery, SENSOR_CHAN_GAUGE_XSENSORNAMEX)` in each block that gets a sensor channel value.
@joelspadin I know this conflicts with your PR on how the fetch/get has to work for your IC
I am still digging.

joelspadin:
Interesting. That would also heavily conflict with the max17055 driver, as it still asserts if you try to fetch anything other than all channels. https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/sensor/max17055/max17055.c#L239

LOWPROKB:
I have put absolutely everything I can think of that is related to I2C, sensors, and nrf_TWIM, into debug mode, and no errors appear in the logs, it (the call to sensor_sample_fetch) simply fails with a negative errno 35, which translates to "Unexpected message type"

LOWPROKB:
Grrrr... ok I found the issues... it is definitely a driver issue, but it was resolved lazily.

- PR submitted indicating that the driver does not work for fetching (https://github.com/zephyrproject-rtos/zephyr/issues/28383)
- PR to solve the issue (https://github.com/zephyrproject-rtos/zephyr/pull/28429)

So the two drivers (MAX17055 & BQ274xx) have diverged enough that it is going to get complicated to have one battery file that works for all drivers. Perhaps a benign config flag that identifies which method to use, and base logic in battery.c on that flag...

joelspadin:
Yeah, I think the proper solution to this is to fix the bq274 driver to support fetching all channels (and probably to fix the max17055 one to support fetching individual channels).

Until then, the loop idea I suggested would make it much easier to work around this in ZMK. You could have something like

```c
#if !SOME_CONFIG_TO_FETCH_INDIVIDUAL_CHANNELS
  rc = sensor_sample_fetch();
  if (rc != 0) {
    return rc;
  }
#endif

for (int i = 0; i < ARRAY_SIZE(battery_channels); i++) {
#if SOME_CONFIG_TO_FETCH_INDIVIDUAL_CHANNELS
  rc = sensor_sample_fetch_chan(battery, battery_channels[i].channel);
  if (rc != 0) {
    continue;
  }
#endif

  rc = sensor_channel_get(battery, battery_channels[i].channel, &value);
  ...
}
```

petejohanson:
> Yeah, I think the proper solution to this is to fix the bq274 driver to support fetching all channels (and probably to fix the max17055 one to support fetching individual channels).
> 
> Until then, the loop idea I suggested would make it much easier to work around this in ZMK. You could have something like
> 
> ```c
> #if !SOME_CONFIG_TO_FETCH_INDIVIDUAL_CHANNELS
>   rc = sensor_sample_fetch();
>   if (rc != 0) {
>     return rc;
>   }
> #endif
> 
> for (int i = 0; i < ARRAY_SIZE(battery_channels); i++) {
> #if SOME_CONFIG_TO_FETCH_INDIVIDUAL_CHANNELS
>   rc = sensor_sample_fetch_chan(battery, battery_channels[i].channel);
>   if (rc != 0) {
>     continue;
>   }
> #endif
> 
>   rc = sensor_channel_get(battery, battery_channels[i].channel, &value);
>   ...
> }
> ```

Yeah, I was thinking something along these same lines. The struggle I have is determining how we settle on the set of channels to fetch in the "fetch individual channels" case.

We could just always try the complete set, or possibly add a Kconfig bool setting for each channel we want to support/fetch/include.

Pondering this more.

PR #899, opened 2021-08-07
Feature/usb report leds

ci-bus:
Feature: Read report leds from usb

Example to test it

1. In board_file.conf add: `CONFIG_ZMK_USB_REPORT_LEDS=y`
2. In rgb_underglow.c add: 

```
int zmk_rgb_hid_listener(const zmk_event_t *eh) {
    const struct zmk_keycode_state_changed *ev = as_zmk_keycode_state_changed(eh);
    uint8_t receive_report[2] = { 0, 0 };
    if (ev && !ev->state) {
        zmk_usb_hid_receive_report((uint8_t *)receive_report, 2);   
        if (receive_report[0]) {
            if (receive_report[1] == HID_USAGE_LED_CAPS_LOCK) {
                zmk_rgb_underglow_on();
            } else {
                zmk_rgb_underglow_off();
            }
        }
    }
    return 0;
}

ZMK_LISTENER(zmk_rgb_hid_listener, zmk_rgb_hid_listener);
ZMK_SUBSCRIPTION(zmk_rgb_hid_listener, zmk_keycode_state_changed);`
```

this example code will turn on the LEDs when caps is locked

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -424,6 +424,10 @@ choice CBPRINTF_IMPLEMENTATION
 
 endchoice
 
+config ZMK_USB_REPORT_LEDS
+  bool "Enable usb report leds"
+  select ENABLE_HID_INT_OUT_EP
+
 module = ZMK
 module-str = zmk
 source "subsys/logging/Kconfig.template.log_config"
--- a/app/include/zmk/hid.h
+++ b/app/include/zmk/hid.h
@@ -135,6 +135,38 @@ static const uint8_t zmk_hid_report_desc[] = {
     ZMK_HID_CONSUMER_NKRO_SIZE,
     HID_MI_INPUT,
     0x00,
+
+#ifdef CONFIG_ZMK_USB_REPORT_LEDS
+    /* LED */
+    0x85, 0x01,
+    /* REPORT_COUNT (5) */
+    HID_GI_REPORT_COUNT,
+    0x05,
+    /* REPORT_SIZE (1) */
+	  HID_GI_REPORT_SIZE,
+    0x01,
+    /* USAGE_PAGE (Page# for LEDs) */
+	  HID_GI_USAGE_PAGE,
+    0x08,
+    /* USAGE_MINIMUM (1) */
+	  0x19, 0x01,
+    /* USAGE_MAXIMUM (5) */
+	  0x29, 0x05,
+    /* OUTPUT (Data, Variable, Absolute), */
+	  HID_MI_OUTPUT,
+    0x02,
+	  /* LED REPORT */
+    /* REPORT_COUNT (1) */
+	  HID_GI_REPORT_COUNT,
+    0x01,
+    /* REPORT_SIZE (3) */
+	  HID_GI_REPORT_SIZE,
+    0x03,
+    /* OUTPUT (Data, Variable, Absolute), */
+	  HID_MI_OUTPUT,
+    0x01,
+#endif /* CONFIG_ZMK_USB_REPORT_LEDS */
+
     /* END COLLECTION */
     HID_MI_COLLECTION_END,
 };
--- a/app/include/zmk/usb.h
+++ b/app/include/zmk/usb.h
@@ -26,4 +26,7 @@ static inline bool zmk_usb_is_hid_ready() { return zmk_usb_get_conn_state() == Z
 
 #ifdef CONFIG_ZMK_USB
 int zmk_usb_hid_send_report(const uint8_t *report, size_t len);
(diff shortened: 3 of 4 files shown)


app/src/usb.c
```diff
@@ -38,6 +38,22 @@ static const struct hid_ops ops = {
     .int_in_ready = in_ready_cb,
 };
 
+#ifdef CONFIG_ZMK_USB_REPORT_LEDS
+
+int zmk_usb_hid_receive_report(uint8_t *report, size_t len) {
+    k_sem_take(&hid_sem, K_MSEC(30));
```
joelspadin, line 44: This looks copied from `zmk_usb_hid_send_report()`, but I'm not sure that code makes sense. 

If `hid_int_ep_read()` requires the semaphore to be taken, then I would expect us to check the return value of `k_sem_take()` to make sure we actually took the semaphore. As is, it looks like this code will try to take the semaphore for 30 ms, then continue regardless of whether it succeeded. Then if `hid_int_ep_read()` fails, it could give a semaphore it never took.

@petejohanson could you explain how the semaphore code in `zmk_usb_hid_send_report()` is supposed to work?

app/Kconfig
```diff
@@ -424,6 +424,10 @@ choice CBPRINTF_IMPLEMENTATION
 
 endchoice
 
+config ZMK_USB_REPORT_LEDS
```
joelspadin, line 427: Maybe it would make more sense to just always support reading the report? Unless there's some reason that the HID report descriptor might need to be kept short, I think making this conditional would just make things more confusing. Any features that depend on this would also need to be conditional, and then anyone who wants to use it in their code has to know to turn it on.

app/src/usb.c
```diff
@@ -38,6 +38,22 @@ static const struct hid_ops ops = {
...
+    int err = hid_int_ep_read(hid_dev, report, len, NULL);
```
joelspadin, line 45: Is the `ret_bytes` parameter to `hid_int_ep_read()` not necessary? I would expect calling code to need to be able to check how many bytes were actually read, since it might be smaller than `len`.

app/include/zmk/hid.h
```diff
@@ -135,6 +135,38 @@ static const uint8_t zmk_hid_report_desc[] = {
...
     0x00,
+
+#ifdef CONFIG_ZMK_USB_REPORT_LEDS
+    /* LED */
+    0x85, 0x01,
+    /* REPORT_COUNT (5) */
+    HID_GI_REPORT_COUNT,
+    0x05,
+    /* REPORT_SIZE (1) */
+	  HID_GI_REPORT_SIZE,
```
joelspadin, line 146: You should install clang-format (https://releases.llvm.org/download.html) and run your code through it. Otherwise the automated checks will complain that this isn't formatted correctly.

If you're using VS Code and clang-format is installed, Microsoft's C/C++ extension can format files with Shift+Alt+F or automatically on save by setting the `editor.formatOnSave` setting.

app/src/usb.c
```diff
@@ -38,6 +38,22 @@ static const struct hid_ops ops = {
     .int_in_ready = in_ready_cb,
 };
 
+#ifdef CONFIG_ZMK_USB_REPORT_LEDS
+
+int zmk_usb_hid_receive_report(uint8_t *report, size_t len) {
```
joelspadin, line 43: Question for someone more familiar with USB/HID than I am: is it okay to attempt to read regardless of the USB status, or do we need a `switch` like the one in `zmk_usb_hid_send_report()`?

app/include/zmk/usb.h
```diff
@@ -26,4 +26,7 @@ static inline bool zmk_usb_is_hid_ready() { return zmk_usb_get_conn_state() == Z
 
 #ifdef CONFIG_ZMK_USB
 int zmk_usb_hid_send_report(const uint8_t *report, size_t len);
+#ifdef CONFIG_ZMK_USB_REPORT_LEDS
+int zmk_usb_hid_receive_report(uint8_t *report, size_t len);
```
joelspadin, line 30: Minor: it would be nice if the declarations in the .h file were in the same order as the definitions in the .c file.

joelspadin (commented):
Is there a callback that gets called when there is new data to read? Possibly `struct hid_ops::get_report`? If so, it would be nice if we could create a ZMK event from that, as that would be more reliable than reading after every key press. (For example, what if you have a keyboard and a separate numpad, and the OS shares num lock state between keyboards? It would be nice if the keyboard could immediately get the new num lock state without waiting for you to press something on it.)

petejohanson:
@ci-bus Thoughts on the feedback from Nick and Joel?

Issue #115, opened 2020-08-21
Feature: Consuming "LED status" state from HID hosts.


petejohanson:
HID hosts will *write* to a USB device that supports and advertises in the report descriptor the status of the caps lock, etc. LEDs so they can enable/disable those for display on the device.

For folks that have keyboards with those lights, or who want to show that in the OLEDs, we need to add support for that to our HID layer.

peterjc:
#2041 added the code to record and track the "LED status" indicators whose state is controlled by the host computer and passed to the keyboard as the HID indicators. Does that cover the scope of this issue, meaning it can be closed?

If so, actually using the HID indicator status in lighting would be a new issue...

caksoylar:
This issue seems resolved, yes, since `CONFIG_ZMK_HID_INDICATORS` setting is implemented: https://zmk.dev/docs/config/system#hid

PR #904, opened 2021-08-09
Feature backlight

bortoz:
This PR adds support for single color LED backlighting. The current implementation supports basic behaviors, such as increase/decrease brightness and toggle on/off.

I also added some useful configurations to increase power efficiency such as turning off the backlight when the USB is disconnected or when the board goes into idle state.

Finally I updated the documentation and the power profiler with the new changes.


app/Kconfig
```diff
@@ -248,6 +248,39 @@ config ZMK_RGB_UNDERGLOW_ON_START
...
+
+config ZMK_BACKLIGHT_LEVELS
+	int "Number of brightness levels"
+    range 2 100
+	default 6
+
+config ZMK_BACKLIGHT_LEVEL_START
+	int "Default brightness level"
+    range 1 ZMK_BACKLIGHT_LEVELS
+	default 3
```
Nicell, line 267:
```suggestion
	default ZMK_BACKLIGHT_LEVELS
```
Since 2 is a valid amount of backlight levels, 3 would be out of range, right?

app/Kconfig
```diff
@@ -248,6 +248,39 @@ config ZMK_RGB_UNDERGLOW_ON_START
 #ZMK_RGB_UNDERGLOW
 endif
 
+menuconfig ZMK_BACKLIGHT
+	bool "LED backlight"
+	select PWM
+	select LED
+    select LED_PWM
```
Nicell, line 255: Could you fix the mixed spaces/tabs in your configs?

app/Kconfig
```diff
@@ -248,6 +248,39 @@ config ZMK_RGB_UNDERGLOW_ON_START
...
+config ZMK_BACKLIGHT_LEVEL_START
+	int "Default brightness level"
+    range 1 ZMK_BACKLIGHT_LEVELS
+	default 3
+
+config ZMK_BACKLIGHT_ON_START
+	bool "Default backlight state"
+	default y
+
+config ZMK_BACKLIGHT_SLEEP
```
Nicell, line 273: Thoughts on an option for idle too?

docs/src/components/power-estimate.js
```diff
@@ -180,6 +186,21 @@ function PowerEstimate({
...
+      backlightPower.ledOn;
+    const backlightMicroW = backlightMicroA * voltageEquivalent;
+    const backlightUsage = backlightMicroW * (1 - usage.percentAsleep);
+
+    totalUsage += backlightUsage;
+    powerUsage.push({
+      title: "Backlight",
+      usage: backlightUsage,
+    });
+  }
```
Nicell, line 202: Maybe instead of setting a blanket 20mA per LED, instead the user can input the forward voltage of the LED and the resistors in use? I think this could help quite a bit with accuracy. 

Then the current would be something like `(powerSupply.outputVoltage - backlight.forwardVoltage) / backlight.resistance * backlight.backlightQuantity`. Definitely more work for the user to input, but LEDs can vary quite a bit in power usage. A blue LED hooked up to 3.3V with only 100 ohms of resistance is about 3mA for reference rather than 20mA.

app/src/backlight.c
```diff
@@ -0,0 +1,233 @@
...
+static void zmk_backlight_save_state_work() {
+    settings_save_one("backlight/state", &state, sizeof(state));
+}
+
+static struct k_delayed_work backlight_save_work;
+#endif // IS_ENABLED(CONFIG_SETTINGS)
+
+static int zmk_backlight_save_state() {
+#if IS_ENABLED(CONFIG_SETTINGS)
+    return settings_save_one("backlight/state", &state, sizeof(state));
```
Nicell, line 86: Instead of saving settings directly, could you debounce the save similarly to how it's done on these two lines in RGB underglow? https://github.com/zmkfirmware/zmk/blob/main/app/src/rgb_underglow.c#L261

The reason for this is to reduce flash operations. People often will be cycling backlight up, down, on, and off. This results in a pretty big amount of flash wear, and only saving after a minute of no changes will help us to reduce that.

docs/src/components/power-estimate.js
Nicell: Another item we need to consider is the PWM power usage. I can help test to see how much power this uses. IIRC PWM power usage is quite high on nRF52.

app/src/backlight.c
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    int b = state.brightness + direction;
+    if (b < 0) {
+        b = 0;
+    } else if (b >= BACKLIGHT_LEVELS) {
+        b = BACKLIGHT_LEVELS - 1;
+    }
+
+    // turn on backlight when change brightness
```
Nicell, line 150: Nice touch!

app/src/backlight.c
```diff
@@ -0,0 +1,233 @@
...
+
+int zmk_backlight_on() { return zmk_backlight_set_on(true); }
+
+int zmk_backlight_off() { return zmk_backlight_set_on(false); }
+
+int zmk_backlight_inc() { return zmk_backlight_change_brt(1); }
+
+int zmk_backlight_dec() { return zmk_backlight_change_brt(-1); }
+
+#if IS_ENABLED(ZMK_BACKLIGHT_SLEEP) || IS_ENABLED(ZMK_TODO)
```
Nicell: `ZMK_TODO` ? 😄

app/Kconfig
Nicell: I see now in the code that this lumps idle/sleep together. This should probably let the user know it's on idle too, or it could be split up.

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,68 @@
...
+| --------------------------- | ------------------------------------------------------- | ------- |
+| `ZMK_BACKLIGHT_LEVELS`      | Number of brightness levels                             | 6       |
+| `ZMK_BACKLIGHT_LEVEL_START` | Default brightness level (1 turn off backlight)         | 3       |
+| `ZMK_BACKLIGHT_ON_START`    | Default backlight state                                 | y       |
+| `ZMK_BACKLIGHT_SLEEP`       | Turn off backlight when keyboard goes into sleep status | y       |
+| `ZMK_BACKLIGHT_USB`         | Turn off backlight when USB is disconnected             | n       |
+
+## Adding Backlight to a Board
+
+First, you need to enable PWM by adding the following lines to your `.overlay` file:
```
Nicell, line 32: A similar description as to why this needs to go into an overlay/board configuration like the RGB underglow one might be useful.

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,68 @@
...
+For example, _P1.13_ would give you _32 \* 1 + 13_ = `<45>` and _P0.15_ would give you _32 \* 0 + 15_ = `<15>`.
+
+Then you have to add the following lines to your `.dtsi` file inside the root devicetree node:
+
+```
+backlight: pwmleds {
+    compatible = "pwm-leds";
+    label = "Backlight LEDs";
+    pwm_led_0 {
+        pwms = <&pwm0 45>;
```
Nicell, line 53: Is it necessary to put 45 again if it was done in the `&pwm0` definition? Seems odd, but I've never used Zephyr's PWM.

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+The value inside `pwm_led_0` must be the same as you used before.
+
+Finally you need to add backlight to the `chosen` element of the root devicetree node:
+
+```
+chosen {
+    ...
+    zmk,underglow = &backlight;
```
Nicell, line 66:
```suggestion
    zmk,backlight = &backlight;
```

Nicell (commented):
Nice job on this @bortoz, clean implementation of the feature. Here are some initial changes/thoughts on this 👍

app/Kconfig
bortoz: It should automatic change if you use `west build -t menuconfig`, but that's a good idea.

docs/docs/features/backlight.md
bortoz: It should. A PWM module can have multiple channels, so the meaning of that line is that the led is using the channel which has pin 45.
These things are poorly documented in Zephyr, I found this [blog](https://medium.com/home-wireless/using-a-pwm-device-in-zephyr-7100d089f15c) really useful.

bortoz:
@Nicell I have updated the code based on your comments.

docs/src/data/power.js
```diff
@@ -71,6 +71,18 @@ export const underglowPower = {
   ledOff: 460, // Quiescent current of a WS2812B
 };
 
+export const backlightLEDs = {
+  White: 3.2,
+  Blue: 3.0,
+  Green: 2.2,
+  Yellow: 2.1,
+  Red: 1.8,
+};
```
Nicell, line 80: I like this, nice!

docs/src/pages/power-profiler.js
```diff
@@ -207,6 +223,68 @@ function PowerProfiler() {
...
+                {backlightEnabled && (
+                  <>
+                    <div className="profilerInput">
+                      <label>LED Quantity</label>
+                      <div className="inputBox">
+                        <input {...bindBacklightQuantity} type="number" />
+                      </div>
+                    </div>
+                    <div className="profilerInput">
+                      <label>LED Voltage</label>
```
Nicell, line 249:
```suggestion
                      <label>LED Forward Voltage</label>
```
Let's be explicit here

docs/src/data/power.js
```diff
@@ -71,6 +71,18 @@ export const underglowPower = {
...
+
+export const backlightPower = {
+  pwmPower: 500, // Estimated power consumption of PWM module
```
Nicell, line 83: I'll test this soon and get some real-world numbers for you.

docs/src/pages/power-profiler.js
```diff
@@ -207,6 +223,68 @@ function PowerProfiler() {
...
+                        <div className="profilerInput">
+                          <div className="inputBox">
+                            <input {...bindBacklightVoltage} type="number" />
+                            <span>V</span>
+                          </div>
+                        </div>
+                      )}
+                    </div>
+                    <div className="profilerInput">
+                      <label>LED Resistence</label>
```
Nicell, line 268:
```suggestion
                      <label>LED Resistance</label>
```
Will probably want to update all other uses of resistance -> resistance

Nicell (commented):
A few more things I noticed. Excited to get this added to ZMK. Thanks for the quick update on this.

app/src/backlight.c
```diff
@@ -0,0 +1,234 @@
...
+
+    int b = state.brightness + direction;
+    if (b < 0) {
+        b = 0;
+    } else if (b >= BACKLIGHT_LEVELS) {
+        b = BACKLIGHT_LEVELS - 1;
+    }
+
+    // turn on backlight when change brightness
+    state.on = true;
```
Nicell, line 152:
```suggestion
    state.on = b != 0;
```
This would make it so that when brightness is 0 we turn off the PWM and save power.

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,72 @@
...
+Backlight is always added to a board, not a shield.
+If you have a shield with backlight, you must add a `boards/` directory within your shield folder to define the backlight individually for each board that supports the shield.
+Inside the `boards/` folder, you define a `<board>.overlay` for each different board.
+
+First, you need to enable PWM by adding the following lines to your `.overlay` file:
+
+```
+&pwm0 {
+	status = "okay";
+	ch0-pin = <45>;
```
Nicell, line 41: Should we note the `ch0-inverted` flag? I'm not super familiar with backlights, but from my understanding you might need it inverted.

htang014:
FWIW, I built your fork to my board, works like a charm

docs/src/data/power.js
Nicell: Just tested this. These results are specific to nrf52840. At 100% and 0% brightness, PWM uses around 0-2uA (negligible basically and probably off due to a smart LED implementation). At anything in between, it's around 510uA. Not sure if you want to add in the logic to check if brightness is 100% or 0%.

bortoz:
Thanks for your feedback @htang014!

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Backlight
+sidebar_label: Backlight
+---
+
+Backlight is a feature used to control array of LEDs, usually placed through or under switches. Unlike [RGB Underglow](./underglow), backlight currently allows only one color per LED, also LEDs are not addressable, so you can't control individual LEDs.
```
Nicell, line 6:
```suggestion
Backlight is a feature used to control array of LEDs, usually placed through or under switches. Unlike [RGB Underglow](underglow.md), backlight currently allows only one color per LED, also LEDs are not addressable, so you can't control individual LEDs.
```

docs/docs/intro.md
```diff
@@ -28,6 +28,7 @@ ZMK is currently missing some features found in other popular firmware. This tab
 | [Encoders](features/encoders)[^1]                                                                                         | ✅  |    ✅     | ✅  |
 | [Display Support](features/displays)[^2]                                                                                  | 🚧  |    🚧     | ✅  |
 | [RGB Underglow](features/underglow)                                                                                       | ✅  |    ✅     | ✅  |
+| [Backlight](features/backlight)                                                                                           | ✅  |    ✅     | ✅  |
```
Nicell, line 31:
```suggestion
| [Backlight](features/backlight.md)                                                                                           | ✅  |    ✅     | ✅  |
```

Nicell (commented):
Two small link changes (#911 related) and a merge conflict fix. Otherwise, this all looks good to me. I'd like to have one more set of eyes over it before a merge though.

app/src/backlight.c
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        return 0;
+    }
+#endif // IS_ENABLED(ZMK_BACKLIGHT_SLEEP)
+
+#if IS_ENABLED(ZMK_BACKLIGHT_USB)
+    struct zmk_usb_conn_state_changed *ev_usb = as_zmk_usb_conn_state_changed(eh);
+    if (ev_usb != NULL) {
+        switch (ev_usb->state) {
```
joelspadin, line 205: Has this been tested? It doesn't look like this will compile. `zmk_usb_conn_state_changed` has a member named `conn_state`, not `state`.

Also, you could maybe simplify this whole thing to just

```c
if (zmk_usb_is_powered()) {
  zmk_backlight_set_on();
} else {
  zmk_backlight_set_off();
}
```
since the value of `ev_usb->conn_state` comes from the state variable that `zmk_usb_is_powered()` uses.

Since `ev_usb` wouldn't be used at all except to check the event type in that case, you could also simplify the `if` check to 

```c
if (as_zmk_usb_conn_state_changed(eh))
```
which admittedly is a bit awkward to read. You could instead set up a separate `ZMK_LISTENER()` for each event so you don't have to check the event type at all.

app/src/behaviors/behavior_backlight.c
```diff
@@ -0,0 +1,57 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+#define DT_DRV_COMPAT zmk_behavior_backlight
+
+#include <device.h>
+#include <drivers/behavior.h>
+#include <logging/log.h>
+
+#include <dt-bindings/zmk/bl.h>
```
joelspadin, line 13: Non system libraries should use `""` instead of `<>`.

app/src/behaviors/behavior_backlight.c
```diff
@@ -0,0 +1,57 @@
...
+    default:
+        LOG_ERR("Unknown backlight command: %d", binding->param1);
+    }
+
+    return -ENOTSUP;
+}
+
+static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
+                                      struct zmk_behavior_binding_event event) {
+    return ZMK_BEHAVIOR_OPAQUE;
```
joelspadin, line 46: Not something to change in this commit, but it seems a bit silly that we require a released callback for behaviors that only do something on press.

@petejohanson maybe we could change `z_impl_behavior_keymap_binding_released` so it returns `ZMK_BEHAVIOR_OPAQUE` instead of `-ENOTSUP` if no release callback is set?

docs/docs/behaviors/lighting.md
```diff
@@ -3,12 +3,27 @@ title: Lighting Behavior
 sidebar_label: Lighting
 ---
 
+import Tabs from '@theme/Tabs';
```
joelspadin, line 6: I'm not sure it's a good idea to use tabs here. If I give you a link to something inside the second tab like https://deploy-preview-904--zmk.netlify.app/docs/behaviors/lighting/#behavior-binding-1, it won't automatically switch to the correct tab. The table of contents on the right is also confusing because it doesn't see the tabs as a header level.

docs/docs/features/backlight.md
joelspadin, line 6:
```suggestion
Backlight is a feature used to control array of LEDs, usually placed through or under switches. Unlike [RGB Underglow](underglow.md), backlight currently allows only one color per LED, also LEDs are not addressable, so you can't control individual LEDs.
```
We recently discovered that links break unless you use the exact path to the .md file.

joelspadin (changes requested):
I think there may be some off by one errors in the brightness level configs and/or calculations.

Also, check that `ZMK_BACKLIGHT_USB` works. I might be missing something, but it looks like the current code won't compile with that option enabled.

app/Kconfig
```diff
@@ -248,6 +248,39 @@ config ZMK_RGB_UNDERGLOW_ON_START
...
+	int "Number of brightness levels"
+	range 2 100
+	default 6
+
+config ZMK_BACKLIGHT_LEVEL_START
+	int "Default brightness level"
+	range 1 ZMK_BACKLIGHT_LEVELS
+	default ZMK_BACKLIGHT_LEVELS
+
+config ZMK_BACKLIGHT_ON_START
```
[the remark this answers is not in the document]
joelspadin: Ignore this. I forgot to update it after seeing parts of the code below. Doing this would make it so if you started with backlighting off and toggled it on, it would go to the initial brightness level instead of a probably more reasonable brightness.

app/Kconfig
```diff
@@ -248,6 +248,39 @@ config ZMK_RGB_UNDERGLOW_ON_START
...
+	bool "Default backlight state"
+	default y
+
+config ZMK_BACKLIGHT_SLEEP
+	bool "Turn off backlight when keyboard goes into idle or sleep status"
+	default y
+
+config ZMK_BACKLIGHT_USB
```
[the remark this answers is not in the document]
bortoz: I know it's not the best name, but I haven't found a better one. Any suggestions are welcome

app/include/dt-bindings/zmk/bl.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
```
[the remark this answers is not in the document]
bortoz: These files only contain macros and it shouldn't be a problem to redefine a macro as long as the definitions are the same.

app/src/backlight.c
```diff
@@ -0,0 +1,237 @@
...
+
+#include <device.h>
+#include <init.h>
+#include <kernel.h>
+
+#include <drivers/led.h>
+#include <logging/log.h>
+#include <settings/settings.h>
+
+#include <zmk/backlight.h>
```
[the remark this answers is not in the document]
bortoz: I don't understand, `""` makes sense for a header, but in a `.c` file there is no local file to include. Also I don't see any files that use the `""` syntax.

app/Kconfig
```diff
@@ -248,6 +248,39 @@ config ZMK_RGB_UNDERGLOW_ON_START
...
+if ZMK_BACKLIGHT
+
+config ZMK_BACKLIGHT_LEVELS
+	int "Number of brightness levels"
+	range 2 100
+	default 6
+
+config ZMK_BACKLIGHT_LEVEL_START
+	int "Default brightness level"
+	range 1 ZMK_BACKLIGHT_LEVELS
```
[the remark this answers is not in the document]
bortoz: Adding `range 0 (ZMK_BACKLIGHT_LEVELS-1)` to the Kconfig file generates an error, so the best compromise I have found is to use values in range [1, ZMK_BACKLIGHT_LEVELS] in the config file and values in range [0, ZMK_BACKLIGHT_LEVELS-1] in the `.c` file. Do you know any workaround?

app/Kconfig
```diff
@@ -248,6 +248,39 @@ config ZMK_RGB_UNDERGLOW_ON_START
...
+	select LED_PWM
+
+if ZMK_BACKLIGHT
+
+config ZMK_BACKLIGHT_LEVELS
+	int "Number of brightness levels"
+	range 2 100
+	default 6
+
+config ZMK_BACKLIGHT_LEVEL_START
```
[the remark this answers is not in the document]
bortoz: I agree it's better, but I prefer to leave it the same as underglow

app/src/backlight.c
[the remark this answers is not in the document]
joelspadin: After looking at the C standard, what I said is not technically accurate. Search order for both quotes and brackets is implementation defined. https://stackoverflow.com/a/77092

However, if we go by what GCC says, bracket should be used for system header files, while quotes should be "used for header files of your own program". https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

app/Kconfig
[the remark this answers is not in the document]
joelspadin: The problem with that is you are initializing the brightness to `CONFIG_ZMK_BACKLIGHT_LEVEL_START - 1`, which will be 0 if you set an initial brightness of 1. Two solutions I can think of:

1. Set the range to be 1-100, and initialize the brightness to `MIN(CONFIG_ZMK_BACKLIGHT_LEVEL_START, CONFIG_ZMK_BACKLIGHT_LEVELS - 1)` to silently ignore values that are too large and clamp them.
2. Set the range to be 1-100, and initialize the brightness to `CONFIG_ZMK_BACKLIGHT_LEVEL_START`. Add a `BUILD_ASSERT(CONFIG_ZMK_BACKLIGHT_LEVEL_START < CONFIG_ZMK_BACKLIGHT_LEVELS, "CONFIG_ZMK_BACKLIGHT_LEVEL_START is too large")`  to cause a build error if the value is too large.
3. Keep the range as it is, but then add something like
  ```c
  #define MIN_BRIGHTNESS 1
  ```
  and then treat `state.brightness` as an offset above the minimum brightness instead of an absolute brightness.

I think options 1 and 2 would make the code simpler, because then you don't have to remember that 0 actually means 1.

app/src/backlight.c
```diff
@@ -0,0 +1,237 @@
...
+
+#define BACKLIGHT_LABEL DT_LABEL(DT_CHOSEN(zmk_backlight))
+#define BACKLIGHT_LEVELS CONFIG_ZMK_BACKLIGHT_LEVELS
+
+struct backlight_state {
+    uint8_t brightness;
+    bool on;
+};
+
+static const struct device *led_bl;
```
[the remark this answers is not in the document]
bortoz: I'm trying to implement it, unfortunately I'm getting a weird compiler error. Could you try to build and see what I'm missing?

app/Kconfig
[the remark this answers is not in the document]
bortoz: I have replaced `range 1 ZMK_BACKLIGHT_LEVELS` with `range 2 ZMK_BACKLIGHT_LEVELS` in the Kconfig file, this should fix that.

app/src/backlight.c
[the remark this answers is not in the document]
bortoz: I think if you want to use `""` instead of `<>` you should refactor the whole project and possibly in a separate PR.
Using different styles in different files is bad and I want to be consistent with the rest of the project.

app/Kconfig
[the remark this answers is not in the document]
joelspadin: Maybe something like `ZMK_BACKLIGHT_AUTO_OFF_IDLE` and `ZMK_BACKLIGHT_AUTO_OFF_USB`?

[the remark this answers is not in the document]
bortoz: I like it!

app/src/backlight.c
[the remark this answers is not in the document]
joelspadin: Okay, looking at the rest of the codebase, `""` is used, but only in one or two spots. This is something that should be discussed separately from this PR.

app/include/dt-bindings/zmk/backlight.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#define BL_TOG 0
```
joelspadin, line 7: Would it be useful to have an option to set the backlight to a specific level with a key? (If so, you'd need to do something like the `&bt` where the behavior takes two params, and the macros are things like

```c
#define BL_DEC 4 0
#define BL_SET 5
```
```c

&bl BL_SET 100
```

app/src/backlight.c
```diff
@@ -0,0 +1,180 @@
...
+    if (rc != 0) {
+        return rc;
+    }
+
+    return zmk_backlight_save_state();
+}
+
+static int zmk_backlight_adjust_brt(int direction) {
+    int b = state.brightness + direction;
+    if (b < 1) {
```
joelspadin, line 124: I think this logic might not do what you want for `|direction| > 1`

I suggest something more like
```c
const int new_brightness = state.brightness + direction
state.on = new_brightness > 0;
state.brightness = CLAMP(new_brightness, 1, BACKLIGHT_LEVELS);
```

joelspadin (commented):
The way ZMK_BACKLIGHT_LEVEL_START works still seems weird to me, but otherwise this looks pretty good to me. I added a few more minor suggestions.

app/src/backlight.c
bortoz: I think this works correctly. Can you give me an example?

joelspadin: You would expect decrementing by one twice would give the same result as decrementing by two, but it might not.

Scenario 1:

1. Start with `on = true, brightness = 2`.
2. Call `zmk_backlight_adjust_brt(-1)`. Now `on = true, brightness = 1`.
3. Call `zmk_backlight_adjust_brt(-1)` again. Now `on = false, brightness = 1`.

Scenario 2:

1. Start with `on = true, brightness = 2`.
2. Call `zmk_backlight_adjust_brt(-2)`. Now `on = false, brightness = 2`.

bortoz: I don't think it's a problem. Also your suggestion doesn't work:

1. Start with `on = true, brightness = 1`
1. Call `zmk_backlight_dec()`: `on = false, brightness = 1`
1. Call `zmk_backlight_inc()`, expected: `on = true, brightness = 1`, actual: `on = true, brightness = 2`

app/include/dt-bindings/zmk/backlight.h
bortoz: I don't find it particularly useful, but other opinions would be helpful

app/src/backlight.c
joelspadin: If we're going to provide an API that allows for increments larger than 1, I think it needs to correctly handle them.

Here are a couple ideas that would handle incrementing from off properly. I'm not sure what people would expect the behavior to be if brightness > 1, you turn the backlight off, and then you increment the brightness though.

It could revert to the previous brightness plus the increment:

```c
// Let brightness decrease to 0, but set it back to 1 if turning on and the brightness is 0.
int zmk_backlight_adjust_brt(int direction) {
  const int new_brightness = state.brightness + direction
  state.on = new_brightness > 0;
  state.brightness = CLAMP(new_brightness, 0, BACKLIGHT_LEVELS);
  ...
}

int zmk_backlight_set_on(bool on) {
  state.on = on;
  if (on && state.brightness == 0) {
    state.brightness = 1;
  }
}
```

It could increment from zero:

```c
// Keep brightness >= 1, but if backlight is off, treat brightness as if it were 0.
int zmk_backlight_adjust_brt(int direction) {
  const int current_brightness = state.on ? state.brightness : 0;
  const int new_brightness = current_brightness + direction
  state.on = new_brightness > 0;
  state.brightness = CLAMP(new_brightness, 1, BACKLIGHT_LEVELS);
  ...
}
```

It could also just revert to the previous brightness without incrementing, then increment on the following change. I think if this was the desired behavior, we would want to handle that at a higher level though, as baking that into the adjust API would be confusing. 

```c
case BL_INC:
  if (zmk_backlight_is_on()) {
    zmk_backlight_inc();
  } else {
    zmk_backlight_set_on(true);
  }
```

joelspadin (commented):
Initial brightness level changes look good to me. I think @petejohanson should review the API though.

app/src/backlight.c
joelspadin: Given the chosen meaning of BACKLIGHT_LEVELS to include zero, things should be clamped to BACKLIGHT_LEVELS - 1 in the above examples.

Or if we were to allow setting arbitrary brightness, those examples would look more like this:

```c
// Allow brightness = 0
int zmk_backlight_get_brt(void) { return state.brightness; }
int zmk_backlight_set_brt(int brightness) {
  state.on = brightness > 0;
  state.brightness = CLAMP(brightness, 0, BACKLIGHT_LEVELS - 1);

  return zmk_backlight_update();
}

int zmk_backlight_adjust_brt(int direction) {
  return zmk_backlight_set_brt(zmk_backlight_get_brt() + direction);
}
```

```c
// Keep brightness >= 1
int zmk_backlight_get_brt(void) {
  return state.on ? state.brightness : 0; 
}

int zmk_backlight_set_brt(int brightness) {
  state.on = brightness > 0;
  state.brightness = CLAMP(brightness, 1, BACKLIGHT_LEVELS - 1);

  return zmk_backlight_update();
}

int zmk_backlight_adjust_brt(int direction) {
  return zmk_backlight_set_brt(zmk_backlight_get_brt() + direction);
}
```

ReFil:
Very excited to see this merged, Polarity works have got a line of products on the way designed with exactly this feature in mind

bortoz:
@ReFil hope to see your products soon!

ReFil:
I've been trying to build this and I've been getting linker errors relating to precompiled submodules. 
cmd.exe /C "cd . && C:\gnu_arm_embedded\bin\arm-none-eabi-gcc.exe   zephyr/CMakeFiles/zephyr_prebuilt.dir/misc/empty_file.c.obj -o zephyr\zephyr_prebuilt.elf  -Wl,-T  zephyr/linker.cmd  -Wl,-Map=C:/Users/Harry/Documents/zmk/app/build/zephyr/zephyr_prebuilt.map  -Wl,--whole-archive  app/libapp.a  zephyr/libzephyr.a  zephyr/arch/common/libarch__common.a  zephyr/arch/arch/arm/core/aarch32/libarch__arm__core__aarch32.a  zephyr/arch/arch/arm/core/aarch32/cortex_m/libarch__arm__core__aarch32__cortex_m.a  zephyr/arch/arch/arm/core/aarch32/cortex_m/mpu/libarch__arm__core__aarch32__cortex_m__mpu.a  zephyr/lib/libc/minimal/liblib__libc__minimal.a  zephyr/lib/posix/liblib__posix.a  zephyr/soc/arm/common/cortex_m/libsoc__arm__common__cortex_m.a  zephyr/soc/arm/nordic_nrf/nrf52/libsoc__arm__nordic_nrf__nrf52.a  zephyr/subsys/bluetooth/common/libsubsys__bluetooth__common.a  zephyr/subsys/bluetooth/host/libsubsys__bluetooth__host.a  zephyr/subsys/bluetooth/controller/libsubsys__bluetooth__controller.a  zephyr/subsys/net/libsubsys__net.a  zephyr/subsys/usb/class/hid/libsubsys__usb__class__hid.a  zephyr/subsys/random/libsubsys__random.a  zephyr/drivers/adc/libdrivers__adc.a  zephyr/drivers/gpio/libdrivers__gpio.a  zephyr/drivers/pwm/libdrivers__pwm.a  zephyr/drivers/spi/libdrivers__spi.a  zephyr/drivers/flash/libdrivers__flash.a  zephyr/drivers/entropy/libdrivers__entropy.a  modules/hal_nordic/libmodules__hal_nordic.a  modules/drivers/kscan/libzmk__drivers__kscan.a  modules/drivers/sensor/battery/lib..__app__drivers__sensor__battery.a  modules/drivers/sensor/ec11/lib..__app__drivers__sensor__ec11.a  -Wl,--no-whole-archive  zephyr/kernel/libkernel.a  zephyr/CMakeFiles/offsets.dir/./arch/arm/core/offsets/offsets.c.obj  -L"c:/gnu_arm_embedded/bin/../lib/gcc/arm-none-eabi/9.3.1/thumb/v7e-m/nofp"  -LC:/Users/Harry/Documents/zmk/app/build/zephyr  -lgcc  zephyr/arch/common/libisr_tables.a  -mcpu=cortex-m4  -mthumb  -mabi=aapcs  -Wl,--gc-sections  -Wl,--build-id=none  -Wl,--sort-common=descending  -Wl,--sort-section=alignment  -Wl,-u,_OffsetAbsSyms  -Wl,-u,_ConfigAbsSyms  -nostdlib  -static  -no-pie  -Wl,-X  -Wl,-N  -Wl,--orphan-handling=warn && cd ."
c:/gnu_arm_embedded/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld.exe: app/libapp.a(backlight.c.obj): in function `zmk_backlight_update':
C:\Users\Harry\Documents\zmk\app\build/../src/backlight.c:53: undefined reference to `__device_dts_ord_33'
c:/gnu_arm_embedded/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld.exe: app/libapp.a(backlight.c.obj): in function `zmk_backlight_init':
C:\Users\Harry\Documents\zmk\app\build/../src/backlight.c:117: undefined reference to `__device_dts_ord_33'
collect2.exe: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
FATAL ERROR: command exited with status 1: 'C:\Program Files\CMake\bin\cmake.EXE' --build 'C:\Users\Harry\Documents\zmk\app\build'

The thing is I can see references to __device_dts_ord_33 in the libapp.a file so I don't know where it's getting stuck

bortoz:
@ReFil This is a Zephyr bug that should be fixed when ZMK is upgraded to Zephyr 2.6 (hopefully soon), in the meantime you can try this [patch](https://github.com/bortoz/zephyr/commit/78a9c4c7000266a3ce928300143ad291f70118f5).

ReFil:
We've been testing this pr on our Dev boards and have spotted the brightness control doesn't seem to work. Putting a scope on the pin shows it's not getting a pwm signal out the microcontroller

htang014:
On my board (n!n) brightness control is working fine.  However `ZMK_BACKLIGHT_AUTO_OFF_IDLE` and `ZMK_BACKLIGHT_AUTO_OFF_USB` do not appear to do anything.

bortoz:
> We've been testing this pr on our Dev boards and have spotted the brightness control doesn't seem to work. Putting a scope on the pin shows it's not getting a pwm signal out the microcontroller

@ReFil can you check you have configured it correctly?

ReFil:
> > We've been testing this pr on our Dev boards and have spotted the brightness control doesn't seem to work. Putting a scope on the pin shows it's not getting a pwm signal out the microcontroller
> 
> @ReFil can you check you have configured it correctly?

Followed the docs to the letter, toggling it on or off works perfectly, it's just brightness that doesn't work. We're using a high frequency drive pin on the nrf

bortoz:
> Followed the docs to the letter, toggling it on or off works perfectly, it's just brightness that doesn't work. We're using a high frequency drive pin on the nrf

It's weird, can you share your config folder?

bortoz:
> On my board (n!n) brightness control is working fine. However `ZMK_BACKLIGHT_AUTO_OFF_IDLE` and `ZMK_BACKLIGHT_AUTO_OFF_USB` do not appear to do anything.

@htang014 I should have fix it, can you check if it's working now?

app/src/backlight.c
```diff
@@ -137,15 +141,17 @@ static int zmk_backlight_adjust_brt(int direction) {
     return zmk_backlight_save_state();
 }
 
+int zmk_backlight_get_brt() { return state.brightness; }
```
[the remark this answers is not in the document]
bortoz: I know it seems strange, my thoughts:
- `zmk_backlight_set_brt(-1)` also results in `zmk_backlight_get_brt() != -1` and I think this is expected.
- `state.on == false` does not imply that the brightness is 0 (or 1), so checking if `state.on == false` and then returning `0` is not a good solution.
Maybe there is a better solution.

ReFil:
> > Followed the docs to the letter, toggling it on or off works perfectly, it's just brightness that doesn't work. We're using a high frequency drive pin on the nrf
> 
> It's weird, can you share your config folder?

https://github.com/ReFil/zmk/backlight

htang014:
> > On my board (n!n) brightness control is working fine. However `ZMK_BACKLIGHT_AUTO_OFF_IDLE` and `ZMK_BACKLIGHT_AUTO_OFF_USB` do not appear to do anything.
> 
> @htang014 I should have fix it, can you check if it's working now?

Both features working on latest commit 👍

htang014:
Found a small bug:
Given `ZMK_BACKLIGHT_AUTO_OFF_USB` is toggled on and USB is disconnected
When I let the keyboard idle
Then LEDs are reenabled when I return

bortoz:
> Followed the docs to the letter, toggling it on or off works perfectly, it's just brightness that doesn't work. We're using a high frequency drive pin on the nrf

@ReFil I did some testing but I wasn't able to reproduce your bug. The only weird thing I've seen is that you are using the same pin for both the pwm and the i2c but I'm not sure that's the cause. Can you try a different pin? Also it's not necessary to use a high frequency pin to control pwm.

> Found a small bug:
> Given `ZMK_BACKLIGHT_AUTO_OFF_USB` is toggled on and USB is disconnected
> When I let the keyboard idle
> Then LEDs are reenabled when I return

@htang014 I'm aware of this, the current logic does not maintain the state before the auto off. I might try adding more logic in a later PR.

ReFil:
> > Followed the docs to the letter, toggling it on or off works perfectly, it's just brightness that doesn't work. We're using a high frequency drive pin on the nrf
> 
> @ReFil I did some testing but I wasn't able to reproduce your bug. The only weird thing I've seen is that you are using the same pin for both the pwm and the i2c but I'm not sure that's the cause. Can you try a different pin? Also it's not necessary to use a high frequency pin to control pwm.
> 
> > Found a small bug:
> > Given `ZMK_BACKLIGHT_AUTO_OFF_USB` is toggled on and USB is disconnected
> > When I let the keyboard idle
> > Then LEDs are reenabled when I return
> 
> @htang014 I'm aware of this, the current logic does not maintain the state before the auto off. I might try adding more logic in a later PR.
> 

Good idea, the i2c should just be assigned to an unused pin, I'll do that instead

ReFil:
Still no dice unfortunately, tried with multiple boards and it still doesn't dim it just shuts off

ReFil:
It's something to do with that pin, we changed to pin 22 and it works flawlessly

ReFil:
@bortoz my sincere apologies, we've traced the issue back to an obscure hardware problem and have rectified it. By way of an apology for wasting your time we'd like to send you a 60% pcb if you want one 😊

bortoz:
@ReFil No problem, glad it works!
If it's something useful I can add a note to the documentation

ReFil:
Because of differing voltage levels we were using an n channel mosfet controlling a p channel, the p channel mosfet needs a pullup but because of gate capacitance the pullup needs to be very low value, typically on our boards we use 100k pull-ups but this meant that the control voltage couldn't rise quickly enough, replacing it with a 100 ohm solved the problem, 10k and 2k didn't help enough

bortoz:
I found a little problem. If we implement the status LEDs (#115), the user should be able to choose whether a specific LED works as a status LED or as a backlight LED. With the current implementation this is not possible. I'm working on it.

bortoz:
Hello, I made some changes:
- I added a PWM LED driver directly in ZMK, so no more need to patch zephyr to build,
- it's now possibile to define multiple LEDs ([docs](https://deploy-preview-904--zmk.netlify.app/docs/features/backlight#multiple-backlight-leds)),
- from @joelspadin's feedback I realized the brightness levels were confusing, I've changed from levels to percentage, it should be similar to underglow and hopefully more intuitive,
- I've added more logic to auto off, it should be more usable.

htang014:
Auto-off behavior seems to be perfect now.  State is maintained :)
In combination with https://github.com/zmkfirmware/zmk/pull/685, I can implement a pretty effective "soft lock" behavior

htang014:
Any plans to approve this soon? Looks like a few backlight-related features in the works would benefits from having this as a foundation first

ReFil:
Has anyone got this working on a split? I'm getting build errors when I try and use this on a split setup
image

purple-rw:
I just pulled in this branch and the build went fine.  I'm testing on my newly-built Dilly with nice!nano v2.  On/off/toggle, and brightness inc/dec all work as expected.  Yay!

ReFil:
@bortoz I just completed a significant refactor of this, changing it to match the current RGB underglow driver more closely, changing the auto off to incorporate the changes pete requested for the RGB auto off as well as supporting the in progress split locality work. It can be found in the backlight-refactor branch of my zmk fork, might be worth merging my changes into yours :)

bortoz:
> @bortoz I just completed a significant refactor of this, changing it to match the current RGB underglow driver more closely, changing the auto off to incorporate the changes pete requested for the RGB auto off as well as supporting the in progress split locality work. It can be found in the backlight-refactor branch of my zmk fork, might be worth merging my changes into yours :)

Thanks! Please open a pull request on my repository so that I can merge your changes.

ReFil:
I tried but it wouldn't let me unfortunately

bortoz:
> I tried but it wouldn't let me unfortunately

I updated the branch to resolve a conflict, you need to rebase your branch.

ReFil:
> > I tried but it wouldn't let me unfortunately
> 
> I updated the branch to resolve a conflict, you need to rebase your branch.

It doesn't say conflict, your repo just doesn't show up in the list of base repos even if I search. Maybe just cherry pick the commit from my repo?

bortoz:
> It doesn't say conflict, your repo just doesn't show up in the list of base repos even if I search. Maybe just cherry pick the commit from my repo?

You should be able to open a pull request here: [https://github.com/bortoz/zmk/compare/backlight...ReFil:backlight-refactor](https://github.com/bortoz/zmk/compare/backlight...ReFil:backlight-refactor), but since there are some conflicts, you need to rebase your branch on my latest commit.

petejohanson:
> Hello, I made some changes:
> 
>     * I added a PWM LED driver directly in ZMK, so no more need to patch zephyr to build,
> 
>     * it's now possibile to define multiple LEDs ([docs](https://deploy-preview-904--zmk.netlify.app/docs/features/backlight#multiple-backlight-leds)),
> 
>     * from @joelspadin's feedback I realized the brightness levels were confusing, I've changed from levels to percentage, it should be similar to underglow and hopefully more intuitive,
> 
>     * I've added more logic to auto off, it should be more usable.

Since the LED driver is fixed in Zephyr 2.7, which we'll move to before too long, instead of duplicating *all* the driver stuff (which I believe is *only* for the macro call to register the driver instance, correct?), I would match rather add another patch to our existing zephyr fork, that we can then just skip when we do the Zephyr 2.7 update.

bortoz:
Yes, it was just a small bug in the macros that was fixed in Zephyr 2.6. I also added the LED GPIO driver to implement the tests which was added in Zephyr 2.6, so upgrading to Zephyr 2.7 should fix all the problems.

bortoz:
@petejohanson I made a PR: https://github.com/zmkfirmware/zephyr/pull/7

app/src/backlight.c
```diff
@@ -0,0 +1,180 @@
...
+
+static int zmk_backlight_update() {
+    uint8_t brt = zmk_backlight_get_brt();
+    LOG_DBG("Update backlight brightness: %d%%", brt);
+
+    for (int i = 0; i < BACKLIGHT_NUM_LEDS; i++) {
+        int rc = led_set_brightness(backlight_dev, i, brt);
+        if (rc != 0) {
+            return rc;
+        }
```
petejohanson, line 53: Do we really want to exit the loop early if the update to one fails?

Perhaps better to `LOG_ERR` all failures, and return some generic errno if any failed.

app/src/behaviors/behavior_backlight.c
petejohanson: I think perhaps returning that is still valid, but the keymap code could better handle that errno specifically.

app/west.yml
```diff
@@ -6,9 +6,11 @@ manifest:
       url-base: https://github.com/zmkfirmware
     - name: microsoft
       url-base: https://github.com/microsoft
+    - name: bortoz
+      url-base: https://github.com/bortoz
   projects:
     - name: zephyr
-      remote: zmkfirmware
+      remote: bortoz
```
petejohanson, line 13: Making a note this needs reverting before merge.

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,117 @@
...
+
+```
+&pwm0 {
+	status = "okay";
+	ch0-pin = <45>;
+	/* ch0-inverted; */
+};
+```
+
+The value `ch0-pin` represents the pin that controls the LEDs. To calculate the value to use, you need a bit of math. You need the hardware port and run it through a function.
```
petejohanson, line 48: We should note this is nRF52 specific, and if this is for a shield, this should go in a board specific file in `<shield>/boards/nice_nano_v2.overlay`, for instance.

petejohanson (changes requested):
On a whole, this looks great!

A few minor comments.

app/west.yml
petejohanson: Also, just merged zmkfirmware/zephyr#7 so this can be reverted now, I believe.

app/src/backlight.c
bortoz: I can change it, however, if the update fails it's most likely due to these two reasons:

* The device is configured incorrectly (e.g. not a real LED PWM peripheral)
* The LED index is out of range (again, due to an incorrect configuration)

In the first case, every update fails, so it makes sense to stop at the first error.
In the second case, every update with an index out of range fails, again it makes sense to stop at the first error.

docs/docs/features/backlight.md
bortoz: Added a note about nRF52, there was already a note about boards and shields

app/src/backlight.c
petejohanson: WFM. Thanks.

app/dts/behaviors/backlight.dtsi
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2021 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+ / {
+	behaviors {
+		/omit-if-no-ref/ bl: behavior_backlight {
+			compatible = "zmk,behavior-backlight";
+			label = "BACKLIGHT";
```
petejohanson, line 11: To avoid having to just change it later, the upcoming locality work in #547 has a limit on the label length (to avoid larger GATT payloads), so can we shorten this, perhaps:

```suggestion
			label = "BCKLGHT";
```

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,117 @@
...
+| `CONFIG_ZMK_BACKLIGHT_BRT_START`     | Default brightness in percent                         | 40      |
+| `CONFIG_ZMK_BACKLIGHT_ON_START`      | Default backlight state                               | y       |
+| `CONFIG_ZMK_BACKLIGHT_AUTO_OFF_IDLE` | Turn off backlight when keyboard goes into idle state | y       |
+| `CONFIG_ZMK_BACKLIGHT_AUTO_OFF_USB`  | Turn off backlight when USB is disconnected           | n       |
+
+## Adding Backlight to a Board
+
+Backlight is always added to a board, not a shield.
+If you have a shield with backlight, you must add a `boards/` directory within your shield folder to define the backlight individually for each board that supports the shield.
+Inside the `boards/` folder, you define a `<board>.overlay` for each different board.
```
petejohanson, line 36: Thinking about this more, since we have the GPIO backed driver available that should be board agnostic, I wonder if we should be documenting and suggesting:

1. Add GPIO backlight nodes/compatibles to shields.
2. For tested boards used w/ the shield, add a `shield/foo/boards/bar.overlay` to override and add the board-specific PWM nodes.

Doing so will ensure a shield would have reasonable "fallback" backlight on some new board used w/ a given shield.

Thoughts?

petejohanson (changes requested):
A few minor follow up thoughts. Thanks again!

I'll have test hardware for this soon.

ReFil:
> A few minor follow up thoughts. Thanks again!
> 
> I'll have test hardware for this soon.

Hi Pete,  we can send you one of our boards equipped for this backlighting stuff if you want

petejohanson:
> > A few minor follow up thoughts. Thanks again!
> > 
> > I'll have test hardware for this soon.
> 
> Hi Pete,  we can send you one of our boards equipped for this backlighting stuff if you want

I actually snagged a revlp recently so that:

* I have a non-split for testing non-wireless controllers like kb2040
* Has backlighting!

I have the resistors en route right now so I can solder the final piece and test this.

Thanks though!

docs/docs/features/backlight.md
bortoz: I think it's a good idea

bortoz: Added a note about that

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,213 @@
...
+
+<Tabs
+defaultValue="shield"
+values={[
+{label: 'Adding to a board', value: 'board'},
+{label: 'Adding to a shield', value: 'shield'},
+]}>
+<TabItem value="board">
+
+First, you must enable PWM by adding the following lines to your `.conf` file:
```
petejohanson, line 43: For this and the shield, I think we might want this in the `Kconfig` files so we can enable these *only* it the backlight config is enabled.

Thoughts?

docs/docs/features/backlight.md
```diff
@@ -0,0 +1,213 @@
...
+    chosen {
+        zmk,backlight = &backlight;
+    };
+}:
+```
+
+</TabItem>
+<TabItem value="shield">
+
+First, you must enable PWM by adding the following lines to your `.conf` file:
```
petejohanson, line 101: This would usually go in the shield's `boards/<board>.conf` file.

petejohanson (changes requested):
Few follow up thoughts, tested this and it works great!

docs/docs/features/backlight.md
bortoz: Not sure where it should be added, inside the `Kconfig.defconfig` file?
And should it be added to a `board/Kconfig.<board>` file for shields?

docs/docs/features/backlight.md
bortoz: Fixed

bortoz:
Rebased to resolve a conflict

docs/docs/features/backlight.md
petejohanson: For boards, this should definitely go in the `Kconfig.defconfig` file.

For shields, according to https://docs.zephyrproject.org/2.5.0/guides/porting/shields.html#board-specific-shield-configuration this would go into `<board>/<board>.defconfig`

ReFil: We put these in board_defconfig for our boards

petejohanson: That works if you don't want users the option to disable a certain feature. But if a user might want to disable backlight, and not pull in other drivers unnecessarily, then the Kconfig.defaults approach actually allows that nuance/dependency chain.

bortoz: Fixed

petejohanson (approved):
Thanks so much for your work on this!

PR #228, opened 2020-10-01
Add support for the Preonic rev3 board to ZMK

shurikai:
This pull request is to add support for the Preonic rev 3. The preonic is a board with built in MCU, and not a shield. The included default keymap is based on the one found on the QMK configurator, but will not be fully functional until shifted keycodes are available. 

To build: `west build -b preonic_rev3`

Note that there is one file touched below (kscan_gpio_matrix.c) that was strictly a change to line formatting in the C file. Nothing changed functionally.

PR #909, opened 2021-08-11
fix: Planck rev 6.1

aumuell:
My Planck rev 6.1 would not register any keys without "col2row".
I also had to apply a matrix_transform in order to move them to the
correct position. Peeking into #228 helped for finding this.

### DIFF
--- a/app/boards/arm/planck/planck_rev6.dts
+++ b/app/boards/arm/planck/planck_rev6.dts
@@ -6,6 +6,7 @@
 
 /dts-v1/;
 #include <st/f3/stm32f303Xc.dtsi>
+#include <dt-bindings/zmk/matrix_transform.h>
 
 / {
 	model = "Plack PCD, rev6";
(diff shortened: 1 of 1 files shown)


petejohanson:
Thanks! This is a board I never actually got my hands on, just threw together best I could to prove out the idea early in ZMK development.

With these changes, is this now fully working for you for basic functionality? (obviously no RGB, etc)

Thanks!

aumuell:
If basic functionality means keyboard input, then it is now working for me. But I only have switches on 36 of the 48 positions (none in the center columns and none in the two outer positions on each side of the bottom row). So I couldn't check that completely.
I also have an encoder on the board, but I don't have that working yet.

aumuell:
Came back home to my tweezers and checked the matrix transform: now it's correct for all three common spacebar layouts. Actually, the previous push did not contain the intended commit and the transform was wrong...

petejohanson (approved):
Thanks!

PR #914, opened 2021-08-15
feat(docs) Clarify Bluetooth profile pairing

MvEerd:
As discussed here;
https://ptb.discord.com/channels/719497620560543766/719909884769992755/876550012153233420

It seems like the bluetooth pairing mechanism ZMK uses isn't always intuitive to new users who may expect a bluetooth device to replace the current bond when it is advertising. However since ZMK does not do this there are frequent questions on the discord and other things may be assumed at fault.

I've tried clarifying this behaviour and also moved the note about number of profiles down to reduce the amount of text in the note
There's some repetition with the `Bluetooth Pairing and Profiles` section but I think this is a tldr that will grab users attention a bit more

I've also fixed a small typo that existed here 1c05f37

See the preview here; https://deploy-preview-914--zmk.netlify.app/docs/behaviors/bluetooth
I'm open to any improvements in phrasing or layout to clarify this further

Cheers!

### DIFF
--- a/docs/docs/behaviors/bluetooth.md
+++ b/docs/docs/behaviors/bluetooth.md
@@ -13,8 +13,6 @@ computer/laptop/keyboard should receive the keyboard input; many of the commands
 When pairing to a host device ZMK saves bond information to the selected profile. It will not replace this when you initiate pairing with another device. To pair with a new device select an unused profile with `BT_SEL`, `BT_NXT` or by clearing an existing profile using `BT_CLR`.
 
 A ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
-
-Please note there are only five available Bluetooth profiles by default. If you need to increase the number of available profiles you can set `CONFIG_BT_MAX_CONN` in your `zmk-config` `.conf` file.
 :::
 
 ## Bluetooth Command Defines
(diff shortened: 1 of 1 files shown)


docs/docs/behaviors/bluetooth.md
```diff
@@ -9,12 +9,12 @@ The bluetooth behavior allows management of various settings and states related
 between the keyboard and the host. By default, ZMK supports five "profiles" for selecting which bonded host
 computer/laptop/keyboard should receive the keyboard input; many of the commands here operation on those profiles.
 
-:::note Number of Profiles
-Please note there are only five available Bluetooth profiles by default. If you need to increase the number of available profiles you can set `CONFIG_BT_MAX_CONN` in your `zmk-config` `.conf` file.
-:::
-
 :::note Connection Management
+When pairing to a host device ZMK saves bond information to the selected profile. It will not replace this when you initiate pairing with another device. To pair with a new device select an unused profile with `BT_SEL`, `BT_NXT` or by clearing an existing profile using `BT_CLR`.
```
caksoylar, line 13:
```suggestion
When pairing to a host device ZMK saves bond information to the selected profile. It will not replace this when you initiate pairing with another device. To pair with a new device select an unused profile with `BT_SEL`, `BT_NXT` or `BT_PRV` bindings, or by clearing an existing profile using `BT_CLR`.
```

MvEerd: Was worried this would get too verbose but it makes sense to list them all, this looks good to me so 👍🏻  thanks!

caksoylar (approved):

petejohanson (approved):

PR #915, opened 2021-08-15
feat(docs): Add note to clarify shared .conf files do not apply outside zmk-config

### LINKED ISSUE #786: Shared .conf file for split shields is ignored if in shield directory
Contrary to what is documented in https://zmkfirmware.dev/docs/development/new-shield/#conf-files-split-shields, `<shield>.conf` that is supposed to include shared configuration for two split halves is ignored if it is in the shield directory. Note that it is processed correctly if it is in a `ZMK_CONFIG` directory, such as the `config/<shield>.conf` file in a Github user config repo.

There are two ways to resolve this:
1. Change building so that `path/to/shield/<shield>.conf` is also processed (is this logic in https://github.com/zmkfirmware/zmk/blob/main/app/cmake/zmk_config.cmake?)
2. Fix the documentation to note that this only works in a config folder, not shield

caksoylar:
Fixes #786, until the time the behavior might be changed.

### DIFF
--- a/docs/docs/development/new-shield.md
+++ b/docs/docs/development/new-shield.md
@@ -286,6 +286,10 @@ In most case you'll only need to use the .conf file that affects both halves of 
 CONFIG_ZMK_SLEEP=y
 ```
 
+:::note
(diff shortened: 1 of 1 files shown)


MvEerd (approved):
I think this is a useful note to add as I've seen it arise in Discord multiple times, if this ends up being an implementation bug that gets fixed we can remember to remove this 👍🏻

MvEerd (approved):

petejohanson (approved):

PR #919, opened 2021-08-22
Update user-setup.md

tammingaj:
Make explicit that the new github project must be checked out locally, and that the setup script should be executed in the root folder of the new project.


caksoylar:
I tested setting up with the user script right now and it handles creating a local repo and setting the upstream to the created Github one. So I don't believe the local checkout step is necessary.

Issue #278, opened 2020-10-13
Mac OS BLE stack is a trashcan: possible workarounds.


Na-Cly:
Pairing with Mac OS has been an overall negative experience that requires multiple resets of the keyboard and multiple resets of bluetooth settings on the mac, and when it does start working you've tried so many things you don't even remember how you got to where you are. 

Here's behavior I've seen when pairing and unpairing from mac os:

1. Initial connection looks like it establishes fine but no key presses are registered.
2. Disconnecting the keyboard by clearing the bond causes issues reconnecting the keyboard to the mac.

Since the improved bluetooth profiles were released this has been significantly helped as now if the connection is established and working it will usually continue working and an occasional reset of central is required after a reboot. 

There are ways to get it working with a couple of reboots but I wanted a method of restoring connectivity without rebooting. So far this is the method I've come up with:

1. Clear bond on keyboard
2. Remove device from mac from debug menu
3. Reset bluetooth device from debug menu 
4. rm ~/Library/Preferences/ByHost/*Bluetooth* && sudo pkill blued

This has been the best to get the keyboard working again without rebooting.

Na-Cly:
It should be noted this workaround didn’t work the last time I tried it, so it’s definitely not a fix every time.

pluskid:
> 2. Remove device from mac from debug menu

Where can I find the "debug menu"?

adamcstephens:
> Where can I find the "debug menu"?

Hold shift and option while clicking the bluetooth menu bar icon

![image](https://user-images.githubusercontent.com/2071575/97934524-f677e900-1d43-11eb-8126-22d9f255ba35.png)

j1n6:
another tip shared @petejohanson 
![unknown](https://user-images.githubusercontent.com/37033/99235603-f836b780-27ed-11eb-93c8-a7cb5dc55d0c.png)

pluskid:
![mac-bluetooth](https://user-images.githubusercontent.com/3542/99350249-c0159000-2852-11eb-8490-7a27cb45beb0.png)

My Macbook constantly wake itself up from sleep, and the battery drains over night. I suspect it is waken up by battery report or some other notification but it wake up even when I switch the keyboard to a different ZMK BLE profile. Unchecking the checkbox that allows bluetooth to wake up computer solves the issue, but for some mysterious reason my MacOS automatically re-check that box for me from time to time.

petejohanson:
[...]

PR #921, opened 2021-08-24
Some initial macOS pairing improvements

petejohanson:
Draft PR to explore some options for improving our pairing experience w/ macOS. Hoping to try to help w/ many of the issues reported in #278 

In particular, this:

* Defers connection parameter update until much later, leveraging the built in Zephyr support for this instead (@Nicell thoughts?)
* Limits our consumer page range in our HID report descriptor. This is a compatibility thing that others have reported on some OS versions, and seems to prevent *some* HID issues w/ connected macOS hosts.

This is *not* perfect yet, in particular, I'm still seeing logs where the BLE "security request" we initiate from the peripheral side isn't logged/detected on the macOS side early on (why, not sure yet), which causes it to defer pairing until too late in the HID/connection setup process.


app/Kconfig
```diff
@@ -78,6 +78,18 @@ config ZMK_BLE_CLEAR_BONDS_ON_START
 	bool "Configuration that clears all bond information from the keyboard on startup."
 	default n
 
+config BT_PERIPHERAL_PREF_MIN_INT
+	default 6
```
xwu, line 82: Apple’s guide states:

> The connection parameter request may be rejected if it does not meet the following guidelines:
> - Peripheral Latency of up to 30 connection intervals.
> - Supervision Timeout from 2 seconds to 6 seconds.
> - Interval Min of at least 15 ms.
> - Interval Min is a multiple of 15 ms.
> - One of the following:
>   - Interval Max at least 15 ms greater than Interval Min.
>   - Interval Max and Interval Min both set to 15 ms.
> - Interval Max * (Peripheral Latency + 1) of 2 seconds or less.
> - Supervision Timeout greater than Interval Max * (Peripheral Latency + 1) * 3.

So I think this means that, to match that guidance, this parameter should be 12 (in units of 1.25 ms).

sorokinDev:
I had problems connecting nrfmicro to MacBook air m1 with latest big sur. Tried to reset macbook, keyboard and everything I could. 
This PR fixed the problem!

peauters:
Tried out this PR and had great success, including connecting to multiple profiles.

Only thing I found was that I had to manually select the BLE output.

Thanks Pete for looking into this!!!

Nicell:
Are you sure that these BLE connection settings are being applied with this PR? See #862, I believe `BT_GAP_AUTO_UPDATE_CONN_PARAMS` needs to be turned on for those values to be used. Can we work off of #862 and add on the HID changes in another PR?

xwu:
Very interesting. So I have findings to report:

Based on what @NiceII points out, I applied this PR but also set `BT_GAP_AUTO_UPDATE_CONN_PARAMS` to `default y`, and then for good measure adhered to Apple's advice to set `BT_PERIPHERAL_PREF_MIN_INT` to `default 12`. And...it didn't work. Flashing the new firmware resulted in no change in behavior: my keyboard didn't connect to my laptop running Monterey within a reasonable amount of time, and there was no input even via USB.

As a last ditch, I _reverted_ my change, setting `BT_GAP_AUTO_UPDATE_CONN_PARAMS` back to `default n` (which, if I understand correctly, means the BLE connection settings aren't being applied at all, and we're using Zephyr's defaults. _And that worked!_ After flashing the firmware, I could instantly input via USB (which I could never do before), and after disconnecting the USB I could use the Bluetooth connection. _So_, fortuitously, this draft PR appears to work _because_ the intended BLE connection settings aren't being applied!

Nicell:
@xwu I suspected this was the case. It's rather odd that MacOS does not want us to use our custom parameters. They all should be in spec especially with your last change. Is it possible that you could check what parameters MacOS negotiates by default when we don't set anything? I imagine they're not ideal for a keyboard experience.

xwu:
@NiceII I'd need some pointers on where to check; I know my way around a compiler but BLE stuff I haven't debugged before :)

xwu:
One more separate piece of feedback: I think the HID changes here have disabled `C_VOLUME_UP` and `C_VOLUME_DOWN`, among others.

malinges:
I might be wrong, but it looks like `BT_GAP_AUTO_UPDATE_CONN_PARAMS` is already enabled on the main branch. If I checkout main, run `west update`, and pristine-build zmk, then despite the `n` default value in `app/Kconfig` I can always find:
- a `CONFIG_BT_GAP_AUTO_UPDATE_CONN_PARAMS=y` line in `app/build/zephyr/.config`
- a `#define CONFIG_BT_GAP_AUTO_UPDATE_CONN_PARAMS 1` line in `app/build/zephyr/include/generated/autoconf.h`

It seems that whatever default value is set for `BT_GAP_AUTO_UPDATE_CONN_PARAMS` in `app/Kconfig` is always overridden to `y` because of [`zephyr/subsys/bluetooth/host/Kconfig.gatt`](https://github.com/zmkfirmware/zephyr/blob/68c11efba6e1eb2d689b3039553868b1b1c57f94/subsys/bluetooth/host/Kconfig.gatt#L137-L144).

Menuconfig also looks a bit confused regarding the conflicting default values, which might explain all this:

image

judahrand:
Is this something we're likely to see progress on soon @Nicell?

musl:
I've been using firmware built off of this branch for a couple of weeks now on my corne clone. It's had a positive effect on both connection stability and possibly battery life with 3 BT profiles in use.

Just one data point, anecdotal for sure.

In any case - Thank you!  - I appreciate the exploration and look forward to this and other split keyboard improvements. 🚀  ❤️

petejohanson:
Ok, I've pushed another hopeful fix for this, based on a small note found scanning https://docs.silabs.com/bluetooth/2.13/code-examples/applications/ble-hid-keyboard#security

I've updated the permissions on a few of our HID service characteristics to also require encryption, which *seems* to force macOS to pair earlier, and avoid the weird timing issues encountered previously.

With this change in place, connections have been reliable.

Needs some further testing, and backing out the HID consumer report changes to try to really dial in this fix more specifically.

Test please?

malinges:
Seeing _massive_ improvements here on macOS Big Sur (11.5.2). The pairing now only takes a couple of seconds, and more importantly, always succeeds (I tried maybe a dozen times). On the main branch I was averaging maybe 1 success out of 6 attempts. This is extremely promising!

Although I don't know if it's just me, but the keyboard feels a bit laggy at times. Might be unrelated, or caused by the additional encryption perhaps? Hard to tell, it might also be caused by my setup, I'm sure having more testers will help.

Anyway, this is definitely going in the right direction. Once the HID report size issue is fixed, there won't be much to complain about ;-) Thank you!

vaheder:
> Although I don't know if it's just me, but the keyboard feels a bit laggy at times. Might be unrelated, or caused by the additional encryption perhaps? Hard to tell, it might also be caused by my setup, I'm sure having more testers will help.

How's your signal strength? I had very low signal quality that caused repeated keypresses and lag until I added this config in my board's config file. This comes per dnaq on Discord: `CONFIG_BT_CTLR_TX_PWR_PLUS_8=y`. https://zmk.dev/docs/troubleshooting/#connectivity-issues

To add to this PR, I'm seeing good results after this commit as well. I had to delete `/Library/Preferences/com.apple.Bluetooth.plist` and restart, possibly due to previous connection issues, but it could also have had to do with signal strength. I'll keep testing and report any issues. I was able to use three profiles to connect to a Macbook Air on Big Sur, a Mac Mini on Big Sur, and an iPhone 11.

malinges:
> How's your signal strength? I had very low signal quality that caused repeated keypresses and lag until I added this config in my board's config file. This comes per dnaq on Discord: `CONFIG_BT_CTLR_TX_PWR_PLUS_8=y`.

It swings a lot between -90dBm and -50 something. I don't know why. I already boosted TX power a little bit (+4dBm, not +8) and it got rid of the occasional lags. But this PR _seems_ to have brought them back a little. It might be unrelated, but I figured I'd mention it so that other testers can keep an eye on this symptom, just in case.

petejohanson:
> > How's your signal strength? I had very low signal quality that caused repeated keypresses and lag until I added this config in my board's config file. This comes per dnaq on Discord: `CONFIG_BT_CTLR_TX_PWR_PLUS_8=y`.
> 
> It swings a lot between -90dBm and -50 something. I don't know why. I already boosted TX power a little bit (+4dBm, not +8) and it got rid of the occasional lags. But this PR _seems_ to have brought them back a little. It might be unrelated, but I figured I'd mention it so that other testers can keep an eye on this symptom, just in case.

Can you post the final `.config` file from your build?

malinges:
> Can you post the final `.config` file from your build?

Here you go: https://pastebin.com/6uw1cNMn

Although I'm not sure my previous comment was very clear: I used to have some lags _before trying this PR_ so this is why I increased the TX power to +4 a couple of weeks ago.

But this PR definitely feels like it has some impact on link quality for me, my keyboard became unresponsive 2 or 3 times in the last 10 minutes. I did reset the board everytime (without even looking if macOS still saw it connected, silly me...) and now the RSSI sits at -57...

I think I'll enable USB logging, add some output switching bindings, and see if something interesting comes up in the logs while using BLE.

vaheder:
I've started to get big RSSI swings as well. Apple has a tool called PacketInspector in the Additional Xcode Tools downloads on https://developer.apple.com. I'm seeing periodic connection timeouts and disconnects followed by reconnects. I'm also occasionally getting `Unknown Connection Identifier` after a `LMP Response Timeout` (usually it's just a `Response Timeout`) that keeps repeating until I reset the board with the reset button.

My configs can be found here: https://github.com/vaheder/zmk-config/tree/testing-pairing-improvements

What kind of logging would be helpful here @petejohanson? I can give you the packet logs from PacketInspector, or I can try to hook up the USB debugging and see what comes from there?

petejohanson:
> I've started to get big RSSI swings as well. Apple has a tool called PacketInspector in the Additional Xcode Tools downloads on https://developer.apple.com. I'm seeing periodic connection timeouts and disconnects followed by reconnects. I'm also occasionally getting `Unknown Connection Identifier` after a `LMP Response Timeout` (usually it's just a `Response Timeout`) that keeps repeating until I reset the board with the reset button.
> 
> My configs can be found here: https://github.com/vaheder/zmk-config/tree/testing-pairing-improvements
> 
> What kind of logging would be helpful here @petejohanson? I can give you the packet logs from PacketInspector, or I can try to hook up the USB debugging and see what comes from there?

Oh! It's the too aggressive connection parameters, that macOS is rejecting probably. I forgot to update the code here, had tweaked manually locally. Give me a few to push a new version.

petejohanson:
Ok, I've just tweaked this to set the interval minimum according to the Apple guidelines, which may result in macOS actually *accepting* the values. please pull again and try again. Thanks!

malinges:
Interestingly, what I see in the logs seems to indicate a higher interval on this PR (it's the same before and after your latest push) vs. the main branch.

main branch:

    <dbg> zmk.le_param_updated: a4:83:e7:da:37:85 (public): interval 12 latency 30 timeout 400

This PR:

    <dbg> zmk.le_param_updated: a4:83:e7:da:37:85 (public): interval 36 latency 30 timeout 400

It's very consistent too, never changes unless I switch branches. If I'm right (I don't really know what I'm talking about to be honest) I guess this could explain the "laggy feel" I've been experiencing?

If I can provide more helpful feedback, please, do ask :)

petejohanson:
> Interestingly, what I see in the logs seems to indicate a higher interval on this PR (it's the same before and after your latest push) vs. the main branch.
> 
> main branch:
> 
>     <dbg> zmk.le_param_updated: a4:83:e7:da:37:85 (public): interval 12 latency 30 timeout 400
> 
> This PR:
> 
>     <dbg> zmk.le_param_updated: a4:83:e7:da:37:85 (public): interval 36 latency 30 timeout 400
> 
> It's very consistent too, never changes unless I switch branches. If I'm right (I don't really know what I'm talking about to be honest) I guess this could explain the "laggy feel" I've been experiencing?
> 
> If I can provide more helpful feedback, please, do ask :)

This branch changes the *defaults*, so could you please be sure you get the updated values in your `.config` or build from a fresh build dir?

malinges:
> This branch changes the _defaults_, so could you please be sure you get the updated values in your `.config` or build from a fresh build dir?

I pristine-build every time just to be safe, and it does looks like I'm getting the updated values in my `.config`:

    [...]
    # CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START is not set
    CONFIG_BT_PERIPHERAL_PREF_MIN_INT=12
    CONFIG_BT_PERIPHERAL_PREF_MAX_INT=12
    CONFIG_BT_PERIPHERAL_PREF_SLAVE_LATENCY=30
    CONFIG_BT_PERIPHERAL_PREF_TIMEOUT=400
    # CONFIG_BT_GATT_NOTIFY_MULTIPLE is not set
    [...]

`CONFIG_BT_GAP_AUTO_UPDATE_CONN_PARAMS` is enabled too...

malinges:
I just installed the PacketLogger tool @vaheder was talking about, and something's definitely wrong (or at least unexpected) with the way connection parameters are handled:

image

Zeyphyr doesn't use the configured `BT_PERIPHERAL_PREF_*_INT` parameters... This explains the "36" I saw in the logs, 36*1.25 being 40ms, exactly halfway between the min and max intervals that macOS sees in the params update request.

Looks like this one's caused by Zephyr!

petejohanson:
@malinges Can you please give my new branch a try, over at ` ble/hog-characteristic-perm-tweaks-macos`? This includes *only* the HID service permission tweaks, after I inspected the Zephyr connection code a bit more, it seems that even if you call the `bt_conn_le_param_update` function right on connection event, they simply store the new params, and still defer sending the update request for `CONFIG_BT_CONN_PARAM_UPDATE_TIMEOUT` milliseconds *anyways*.

So, that branch has a more selective fix, let's see how that goes please.

(Thanks for testing!)

malinges:
Ok, the screenshot in my previous comment wasn't of the right message, here's a better one, for the `ble/defer-connection-param-upgarade-while-pairing` branch:

image

45ms connection interval... No wonder the board feels "laggy".

Now with your new `ble/hog-characteristic-perm-tweaks-macos` branch:

image

Back to 15ms! Also in the logs:

    <dbg> zmk.le_param_updated: a4:83:e7:da:37:85 (public): interval 12 latency 30 timeout 400 // <-- back to interval 12

Connections (pairing) still seem very fast and reliable, the "laggy feeling" is gone, even consumer page keycodes still work! The best of both worlds if you ask me :)

vaheder:
Same here, seems more reliable and fast. Here are my numbers for comparison:
Screen Shot 2021-09-15 at 9 43 33 PM

malinges:
11.25ms! ~Who did you bribe~ I mean what kind of Mac / version of macOS are you using?

vaheder:
> 11.25ms! ~Who did you bribe~ I mean what kind of Mac / version of macOS are you using?

This is on a 2018 Intel Mac Mini running 11.6.

malinges:
> This is on a 2018 Intel Mac Mini running 11.6.

Ok so probably not the same performance vs. power usage heuristics tuning than a laptop (2019 Intel MacBook Pro in my case - still at 15ms after updating to 11.6). Good to know, thanks!

hahuang65:
The `ble/hog-characteristic-perm-tweaks-macos` worked like charm for my BT60.

It was previously paired to my Mac, but I had to re-pair it. For 2 weeks it wouldn't pair no matter what I tried. It was still successfully paired a Windows and Linux machine, but wouldn't work with the Mac.

My nice!60 was successfully paired to all 3 and working.

Finally, using the branch with the fix, the BT60 paired up in 1 second.

Thanks for the hard work!

bwhelm:
Using `ble/hog-characteristic-perm-tweaks-macos` works for me with one keyboard, but not with another; both keyboards have a nice!nano, and both successfully connect to iPhone/iPad, but one of them does not connect to a Mac running MacOS 11.6. In both cases I hit the "connect" button in the Bluetooth SysPrefs pane and was asked if I wanted to connect. The keyboard that does not work shows up in the Bluetooth pane as "connected", but it also has the spinning wheel on the right in its device line of the Bluetooth pane, and no keypresses are registered.

What diagnostic information can I provide?

malinges:
@bwhelm did you both remove the device from the macOS Bluetooth manager and press BT_CLR on the keyboard? If you did, then you might want to try deleting the file `/Library/Preferences/com.apple.Bluetooth.plist` like @vaheder mentioned above and try again: https://github.com/zmkfirmware/zmk/pull/921#issuecomment-920447175

bwhelm:
@malinges -- Yes, I had done BT_CLR, many times. I tried again deleting that file and rebooting, with no change: still not working on the one keyboard.

malinges:
Doing some tests on a Windows laptop. The initial connection (pairing) works greats, just like with macOS, but re-connections (like after a board reset) are a bit hit-or-miss. It often requires some time and/or a few board resets to get things working again, whereas the same scenario generally works flawlessly with macOS. Switching Bluetooth profiles on the keyboard also seems to help sometimes. I can provide Bluetooth traces if needed. But please note I never really used any ZMK boards with this laptop until today, and I didn't even try the main branch before, so this might be unrelated. I'll investigate a bit more.

What I can say for sure is that I'm getting 15ms intervals, just like on macOS:
image

petejohanson:
> @malinges -- Yes, I had done BT_CLR, many times. I tried again deleting that file and rebooting, with no change: still not working on the one keyboard.

Does the same thing work properly on `main`?

bwhelm:
> Does the same thing work properly on `main`?

Nope. I get the same result: it indicates it's connected, but has the spinning wheel, and no keypresses are registered.

petejohanson:
> > Does the same thing work properly on `main`?
> 
> Nope. I get the same result: it indicates it's connected, but has the spinning wheel, and no keypresses are registered.

Can you please double check, and make sure you remove the keyboard from macOS *first*, then BT_CLR, then add to macOS again?

bwhelm:
> Can you please double check, and make sure you remove the keyboard from macOS _first_, then BT_CLR, then add to macOS again?

Tried again, both with `main` and `ble/hog-characteristic-perm-tweaks-macos`. From the BT menu (holding <Opt><Shift>), selected "Remove all devices", which nonetheless leaves the keyboard "Connected" in the BT SysPref pane. In that pane, I manually disconnect it. With my keyboard switched off, I turn BT off and back on again, confirming that the keyboard does not appear in the BT pane. Then switch the keyboard on, and it immediately appears in the BT pane. Then select BT_CLR on the keyboard and try connecting, and I still get the same result.

petejohanson:
> > Can you please double check, and make sure you remove the keyboard from macOS _first_, then BT_CLR, then add to macOS again?
> 
> Tried again, both with `main` and `ble/hog-characteristic-perm-tweaks-macos`. From the BT menu (holding <Opt><Shift>), selected "Remove all devices", which nonetheless leaves the keyboard "Connected" in the BT SysPref pane. In that pane, I manually disconnect it. With my keyboard switched off, I turn BT off and back on again, confirming that the keyboard does not appear in the BT pane. Then switch the keyboard on, and it immediately appears in the BT pane. Then select BT_CLR on the keyboard and try connecting, and I still get the same result.

Since this happens also on `main`, I don't think this is a regression from these changes. Have you tried switching to each and every profile (except maybe those taken properly by other hosts), and clearing *all* the other profiles? Occasionally we've had users with stuff lingering in those other profiles that cause issues. If that doesn't work, I'd recommend setting up logging, and continuing working on this in the #help discord channel.

jamesramsay:
@petejohanson thanks for your work on this! Rebuilding using `ble/hog-characteristic-perm-tweaks-macos` means I am now able to paid with MacOS! ⚡

petejohanson:
Thanks everyone for testing, I'm closing this for now in favor of #946 that I just opened, and will likely merge soon.

PR #922, opened 2021-08-25
Add optional dedicated work queue for display updates

petejohanson:
For displays that take a lot longer to push updates out, e.g. EPD devices, during the blocking calls to push those updates, the system work queue is currently blocked from processing other key inputs, sending HID updates, etc.

This PR adds a new Kconfig setting to allow instead using a dedicated work queue for all UI updates.

Doing so requires that all UI updates be done through `k_work` submitted to the UI work q, instead of assuming we can do that directly from the main event callbacks.

Tested on a waveshare EPD module, will test on standard OLEDs soon.

### DIFF
--- a/app/include/zmk/display.h
+++ b/app/include/zmk/display.h
@@ -6,4 +6,6 @@
 
 #pragma once
 
+struct k_work_q *zmk_display_work_q();
+
 int zmk_display_init();
--- a/app/include/zmk/display.h
+++ b/app/include/zmk/display.h
@@ -4,8 +4,57 @@
  * SPDX-License-Identifier: MIT
  */
 
+/** @file display.h
+ *  @brief Display functions and macros.
+ */
+
 #pragma once
 
 struct k_work_q *zmk_display_work_q();
 
+bool zmk_display_is_initialized();
 int zmk_display_init();
+
+/**
+ * @brief Macro to define a ZMK event listener that handles the thread safety of fetching
+ * the necessary state from the system work queue context, invoking a work callback
+ * in the display queue context, and properly accessing that state safely when performing
+ * display/LVGL updates.
+ *
+ * @param listener THe ZMK Event manager listener name.
+ * @param state_type The struct/enum type used to store/transfer state.
+ * @param cb The callback to invoke in the dispaly queue context to update the UI. Should be `void
+ * func(state_type)` signature.
+ * @param state_func The callback function to invoke to fetch the updated state from ZMK core.
+ * Should be `state type func(const zmk_event_t *eh)` signature.
+ * @retval listner##_init Generates a function `listener##_init` that should be called by the widget
+ * once ready to be updated.
+ **/
+#define ZMK_DISPLAY_WIDGET_LISTENER(listener, state_type, cb, state_func)                          \
+    K_MUTEX_DEFINE(listener##_mutex);                                                              \
+    static state_type __##listener##_state;                                                        \
+    static state_type listener##_get_local_state() {                                               \
+        k_mutex_lock(&listener##_mutex, K_FOREVER);                                                \
+        state_type copy = __##listener##_state;                                                    \
+        k_mutex_unlock(&listener##_mutex);                                                         \
+        return copy;                                                                               \
+    };                                                                                             \
+    static void listener##_work_cb(struct k_work *work) { cb(listener##_get_local_state()); };     \
+    K_WORK_DEFINE(listener##_work, listener##_work_cb);                                            \
+    static void listener##_refresh_state(const zmk_event_t *eh) {                                  \
+        k_mutex_lock(&listener##_mutex, K_FOREVER);                                                \
+        __##listener##_state = state_func(eh);                                                     \
+        k_mutex_unlock(&listener##_mutex);                                                         \
+    };                                                                                             \
+    static void listener##_init() {                                                                \
+        listener##_refresh_state(NULL);                                                            \
+        listener##_work_cb(NULL);                                                                  \
+    }                                                                                              \
+    static int listener##_cb(const zmk_event_t *eh) {                                              \
+        if (zmk_display_is_initialized()) {                                                        \
+            listener##_refresh_state(eh);                                                          \
+            k_work_submit_to_queue(zmk_display_work_q(), &listener##_work);                        \
+        }                                                                                          \
+        return ZMK_EV_EVENT_BUBBLE;                                                                \
+    }                                                                                              \
+    ZMK_LISTENER(listener, listener##_cb);
--- a/app/src/display/Kconfig
+++ b/app/src/display/Kconfig
@@ -29,6 +29,29 @@ config ZMK_DISPLAY_STATUS_SCREEN_CUSTOM
 
 endchoice
 
+choice ZMK_DISPLAY_WORK_QUEUE
+    prompt "Work queue selection for UI updates"
+
+config ZMK_DISPLAY_WORK_QUEUE_SYSTEM
+    bool "Use default system work queue for UI updates"
+
+config ZMK_DISPLAY_WORK_QUEUE_DEDICATED
+    bool "Use dedicated work queue for UI updates"
+
+endchoice
+
+if ZMK_DISPLAY_WORK_QUEUE_DEDICATED
+
+config ZMK_DISPLAY_DEDICATED_THREAD_STACK_SIZE
+    int "Stack size for dedicated UI thread/queue"
+    default 2048
+
+config ZMK_DISPLAY_DEDICATED_THREAD_PRIORITY
+    int "Thread priority for dedicated UI thread/queue"
+    default 5
+
+endif # ZMK_DISPLAY_WORK_QUEUE_DEDICATED
+
 rsource "widgets/Kconfig"
 
 endif
--- a/app/src/display/main.c
+++ b/app/src/display/main.c
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: MIT
  */
 
+#include <kernel.h>
 #include <init.h>
 #include <device.h>
 #include <devicetree.h>
@@ -32,19 +33,41 @@ void display_tick_cb(struct k_work *work) { lv_task_handler(); }
 
 K_WORK_DEFINE(display_tick_work, display_tick_cb);
 
-void display_timer_cb() {
-    lv_tick_inc(TICK_MS);
-    k_work_submit(&display_tick_work);
+#if IS_ENABLED(CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED)
+
+K_THREAD_STACK_DEFINE(display_work_stack_area, CONFIG_ZMK_DISPLAY_DEDICATED_THREAD_STACK_SIZE);
+
+static struct k_work_q display_work_q;
+
+#endif
+
+struct k_work_q *zmk_display_work_q() {
+#if IS_ENABLED(CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED)
+    return &display_work_q;
+#else
+    return &k_sys_work_q;
+#endif
 }
 
+void display_timer_cb() {
+    lv_tick_inc(TICK_MS);
+    k_work_submit_to_queue(zmk_display_work_q(), &display_tick_work);
+}
+
+void blank_display_cb(struct k_work *work) { display_blanking_on(display); }
+
+void unblank_display_cb(struct k_work *work) { display_blanking_off(display); }
+
 K_TIMER_DEFINE(display_timer, display_timer_cb, NULL);
+K_WORK_DEFINE(blank_display_work, blank_display_cb);
+K_WORK_DEFINE(unblank_display_work, unblank_display_cb);
 
 static void start_display_updates() {
     if (display == NULL) {
         return;
     }
 
-    display_blanking_off(display);
+    k_work_submit_to_queue(zmk_display_work_q(), &unblank_display_work);
 
     k_timer_start(&display_timer, K_MSEC(TICK_MS), K_MSEC(TICK_MS));
 }
@@ -54,7 +77,7 @@ static void stop_display_updates() {
         return;
     }
 
-    display_blanking_on(display);
+    k_work_submit_to_queue(zmk_display_work_q(), &blank_display_work);
 
     k_timer_stop(&display_timer);
 }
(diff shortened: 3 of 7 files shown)


app/src/display/Kconfig
```diff
@@ -29,6 +29,29 @@ config ZMK_DISPLAY_STATUS_SCREEN_CUSTOM
 
 endchoice
 
+choice ZMK_DISPLAY_WORK_QUEUE
+    prompt "Work queue selection for UI updates"
```
joelspadin, line 33: Should we default this one way or the other? I'm not sure if most displays update quickly or slowly.

app/src/display/main.c
```diff
@@ -32,19 +33,43 @@ void display_tick_cb(struct k_work *work) { lv_task_handler(); }
 
 K_WORK_DEFINE(display_tick_work, display_tick_cb);
 
+#if IS_ENABLED(CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED)
+
+K_THREAD_STACK_DEFINE(display_work_stack_area, CONFIG_ZMK_DISPLAY_DEDICATED_THREAD_STACK_SIZE);
+
+static struct k_work_q display_work_q;
+
+struct k_work_q *zmk_display_work_q() {
```
joelspadin, line 42: Since the function signature doesn't depend on CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED, this might be a little easier to maintain if you made it be 
```c
struct k_work_q *zmk_display_work_q() {
#if IS_ENABLED(CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED)
    return &display_work_q;
#else
    return &k_sys_work_q;
#endif
}
```

app/src/display/widgets/battery_status.c
```diff
@@ -33,16 +35,29 @@ void battery_status_init() {
...
     char text[2] = "  ";
-    uint8_t level = bt_bas_get_battery_level();
+    k_mutex_lock(&battery_status_mutex, K_FOREVER);
+
+    uint8_t level = battery_status_state.level;
 
 #if IS_ENABLED(CONFIG_USB)
-    if (zmk_usb_is_powered()) {
+    if (battery_status_state.usb_present) {
         strcpy(text, LV_SYMBOL_CHARGE);
```
joelspadin, line 55: Not relevant to this change, and it's currently not an issue since the code can only write two characters to the string, but note that strcpy and strcat will happily overflow the buffer if one of those `LV_` constants happens to be longer than a single character or someone adds another strcat call later.

joelspadin (approved):
I added a couple of suggestions but otherwise this looks good to me.

app/src/display/main.c
krikun98: Wouldn't this approack create a `static struct k_work_q display_work_q;` in both `if` branches?

joelspadin: I'm not sure I understand what you mean, but probably no? You would statically define the work queue in a separate #if block, or you could even define it inside the function itself if there wasn't any code outside the function that needed to reference it.

krikun98: I'm implementing a simular solution for the mousekeys PR, so the question just kind of sprung up, and I thought of the separate `#if` block right after the question. Thanks! This looks a tad more elegant.

petejohanson: Yeah, I like that refactor to avoid the duplication of the function signature. Pushed that.

app/src/display/Kconfig
petejohanson: So, the default, if not set, is the first defined in the choice. By not *explicitly* setting a default, then it's easier to default it in `Kconfig.defconfig` files for shields, etc.

So, the *implicit* default is the system work queue, which I think is reasonable.

joelspadin: Ah. I didn't realize it worked that way. (Doesn't help that I cannot find a single easy to understand resource on how kconfig works.)

petejohanson:
@joelspadin I refacted this to extract the mutex/queue/work logic into a re-usable macro in `app/include/zmk/display.h` and refactored the widgets accordingly. Got a chance for a follow up review?

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+ * display/LVGL updates.
+ * @param listener THe ZMK Event manager listener name
+ * @param state_type The struct/enum type used to store/transfer state.
+ * @param cb The callback to invoke in the dispaly queue context to update the UI
+ * @param state_func The callback function to invoke to fetch the updated state from ZMK core
+ **/
+#define ZMK_DISPLAY_WIDGET_LISTENER(listener, state_type, cb, state_func)                          \
+    K_MUTEX_DEFINE(listener##_mutex);                                                              \
+    void listener##_work_cb(struct k_work *work) {                                                 \
+        LOG_DBG("");                                                                               \
```
joelspadin, line 27: Leftover debugging code?

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+    };                                                                                             \
+    K_WORK_DEFINE(listener##_work, listener##_work_cb);                                            \
+    void listener##_refresh_state(const zmk_event_t *eh) {                                         \
+        k_mutex_lock(&listener##_mutex, K_FOREVER);                                                \
+        state_type new_state = state_func(eh);                                                     \
+        memcpy(&state, &new_state, sizeof(state));                                                 \
+        k_mutex_unlock(&listener##_mutex);                                                         \
+    };                                                                                             \
+    int listener##_cb(const zmk_event_t *eh) {                                                     \
+        LOG_DBG("");                                                                               \
```
joelspadin, line 46: Leftover debugging code?

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+    K_MUTEX_DEFINE(listener##_mutex);                                                              \
+    void listener##_work_cb(struct k_work *work) {                                                 \
+        LOG_DBG("");                                                                               \
+        cb();                                                                                      \
+    };                                                                                             \
+    static state_type state;                                                                       \
+    static state_type listener##_get_local_state() {                                               \
+        state_type copy;                                                                           \
+        k_mutex_lock(&listener##_mutex, K_FOREVER);                                                \
+        memcpy(&copy, &state, sizeof(state));                                                      \
```
joelspadin, line 34: Should be able to simplify to 
```suggestion
        copy = state; \
```

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+ * @param cb The callback to invoke in the dispaly queue context to update the UI
+ * @param state_func The callback function to invoke to fetch the updated state from ZMK core
+ **/
+#define ZMK_DISPLAY_WIDGET_LISTENER(listener, state_type, cb, state_func)                          \
+    K_MUTEX_DEFINE(listener##_mutex);                                                              \
+    void listener##_work_cb(struct k_work *work) {                                                 \
+        LOG_DBG("");                                                                               \
+        cb();                                                                                      \
+    };                                                                                             \
+    static state_type state;                                                                       \
```
joelspadin, line 30: Adding a static variable named just `state` is likely to cause conflicts with any other variable named `state` (such as multiple listeners), and it won't be obvious why because one of the variables is hidden inside a macro.

```suggestion
    static state_type listener##_state;                                                                       \
```

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+        k_mutex_lock(&listener##_mutex, K_FOREVER);                                                \
+        memcpy(&copy, &state, sizeof(state));                                                      \
+        k_mutex_unlock(&listener##_mutex);                                                         \
+        return copy;                                                                               \
+    };                                                                                             \
+    K_WORK_DEFINE(listener##_work, listener##_work_cb);                                            \
+    void listener##_refresh_state(const zmk_event_t *eh) {                                         \
+        k_mutex_lock(&listener##_mutex, K_FOREVER);                                                \
+        state_type new_state = state_func(eh);                                                     \
+        memcpy(&state, &new_state, sizeof(state));                                                 \
```
joelspadin, line 42: Should be able to simplify to 

```suggestion
        state = state_func(eh);                                                     \
```

app/src/display/widgets/layer_status.c
```diff
@@ -31,29 +38,44 @@ void layer_status_init() {
...
-    const char *layer_label = zmk_keymap_layer_label(active_layer_index);
-    if (layer_label == NULL) {
+static void set_layer_symbol(lv_obj_t *label) {
+    struct layer_status_state local_state = widget_layer_status_get_local_state();
+
+    if (local_state.label == NULL) {
         char text[6] = {};
 
-        sprintf(text, LV_SYMBOL_KEYBOARD "%i", active_layer_index);
+        sprintf(text, LV_SYMBOL_KEYBOARD "%i", local_state.index);
```
joelspadin, line 64:
```suggestion
        snprintf(text, sizeof(text), LV_SYMBOL_KEYBOARD "%i", local_state.index);
```

app/src/display/widgets/layer_status.c
```diff
@@ -31,29 +38,44 @@ void layer_status_init() {
...
 
         lv_label_set_text(label, text);
     } else {
         char text[12] = {};
 
-        snprintf(text, 12, LV_SYMBOL_KEYBOARD "%s", layer_label);
+        snprintf(text, 12, LV_SYMBOL_KEYBOARD "%s", local_state.label);
```
joelspadin, line 70:
```suggestion
        snprintf(text, sizeof(text), LV_SYMBOL_KEYBOARD "%s", local_state.label);
```

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+ * @param listener THe ZMK Event manager listener name
+ * @param state_type The struct/enum type used to store/transfer state.
+ * @param cb The callback to invoke in the dispaly queue context to update the UI
+ * @param state_func The callback function to invoke to fetch the updated state from ZMK core
+ **/
+#define ZMK_DISPLAY_WIDGET_LISTENER(listener, state_type, cb, state_func)                          \
+    K_MUTEX_DEFINE(listener##_mutex);                                                              \
+    void listener##_work_cb(struct k_work *work) {                                                 \
+        LOG_DBG("");                                                                               \
+        cb();                                                                                      \
```
joelspadin, line 28: Would it make more sense to pass the state into the callback function? Then users of this macro wouldn't need to know about the (currently undocumented) `listener##_get_local_state` function.

petejohanson: Indeed! Just pushed an updated that refactors that to make that a bit more "pure" and not have to worry about the locally generated functions.

app/src/display/widgets/layer_status.c
petejohanson: Done.

app/include/zmk/display.h
petejohanson: Yeah, renamed, and included some underscore prefixes to be safe.

app/include/zmk/display.h
petejohanson: Yup. Removed.

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+int zmk_display_init();
+
+/**
+ * @brief Macro to define a ZMK event listener that handles the thread safety of fetching
+ * the necessary state from the system work queue context, invoking a work callback
+ * in the display queue context, and properly accessing that state safely when performing
+ * display/LVGL updates.
+ * @param listener THe ZMK Event manager listener name
+ * @param state_type The struct/enum type used to store/transfer state.
+ * @param cb The callback to invoke in the dispaly queue context to update the UI
```
joelspadin, line 21: Would be helpful to list the expected function signatures for `cb` and `state_func`.

app/include/zmk/display.h
```diff
@@ -6,4 +6,48 @@
...
+ * @param state_func The callback function to invoke to fetch the updated state from ZMK core
+ * @returns Generates a function listener##_init that should be called by the widget once ready to
```
joelspadin, line 23: Seems a bit weird to list this in `@returns` since it doesn't behave like a function and doesn't "return" the _init() function.

app/src/display/widgets/output_status.c
```diff
@@ -36,62 +45,73 @@ void output_status_init() {
...
+#if defined(CONFIG_USB)
+ZMK_SUBSCRIPTION(widget_output_status, zmk_usb_conn_state_changed);
+#endif
+#if defined(CONFIG_ZMK_BLE)
+ZMK_SUBSCRIPTION(widget_output_status, zmk_ble_active_profile_changed);
+#endif
+
 int zmk_widget_output_status_init(struct zmk_widget_output_status *widget, lv_obj_t *parent) {
     output_status_init();
+    // widget_output_status_refresh_state(NULL);
```
joelspadin, line 101: Are these intended to be left commented out?

app/src/display/widgets/wpm_status.c
```diff
@@ -31,37 +36,42 @@ void wpm_status_init() {
...
+    SYS_SLIST_FOR_EACH_CONTAINER(&widgets, widget, node) { set_wpm_symbol(widget->obj, state); }
+}
+
+ZMK_DISPLAY_WIDGET_LISTENER(widget_wpm_status, struct wpm_status_state, wpm_status_update_cb,
+                            wpm_status_get_state)
+ZMK_SUBSCRIPTION(widget_wpm_status, zmk_wpm_state_changed);
+
 int zmk_widget_wpm_status_init(struct zmk_widget_wpm_status *widget, lv_obj_t *parent) {
     wpm_status_init();
+    // widget_wpm_status_refresh_state(NULL);
```
joelspadin, line 63: Are these intended to be left commented out?

app/include/zmk/display.h
petejohanson: Yeah, I didn't find anything authoritative on how to document functions generated by a macro, and fell back to this. Is there something here I missed?

petejohanson: Switched this to at least use `@retval` which let's you *name* the returned value you're documenting, which seems better. Thoughts?

app/include/zmk/display.h
petejohanson: Added, as best I could w/ Doxygen syntax.

app/src/display/widgets/output_status.c
petejohanson: Nope, removed.

app/src/display/widgets/wpm_status.c
petejohanson: Nope, removed.

Issue #928, opened 2021-09-03
Tidbit missing semicolon


kevinmgee:
Default build script fails when using board Nice!Nano v2 and Tidbit shield.

https://github.com/zmkfirmware/zmk/blob/e491c282e74559d10818394feae3d64c7a559acf/docs/static/setup.sh#L121
[logs_1.zip](https://github.com/zmkfirmware/zmk/files/7102986/logs_1.zip)

Missing semicolon after split = "n"
(Same issue with missing semicolon on next line)

Build fails because Tidbit is not a split keyboard shield.

dxmh:
@kevinmgee This has now been fixed in https://github.com/zmkfirmware/zmk/pull/931.

PR #931, opened 2021-09-04
fix: Add missing semicolons for tidbit and eek in setup script

caksoylar:
This results in these two boards getting recognized as split, since that is the default. Credit goes to @dxmh for noticing it on Discord :)


petejohanson (approved):
Thanks!

dxmh:
This fixes #928.

PR #917, opened 2021-08-21
Update docs for OSX use cases

muppetjones:
* Added and clarified some OSX-specific use cases.
* Added a .remarkrc file to prevent usage of personal settings.

PR #932, opened 2021-09-06
Fix macOS setpup docs

rtitmuss:


docs/docs/development/setup.md
```diff
@@ -343,11 +343,12 @@ The install command is:
 brew install --cask gcc-arm-embedded
 ```
 
-:::warning Security Controls Workaround
+Next, we'll configure Zephyr™ with some extra environment variables needed to find the cross compiler by running the following commands:
 
-Please be sure to read the [additional setup instructions](https://docs.zephyrproject.org/2.5.0/getting_started/installation_mac.html#mac-gatekeeper) needed to address security controls found in macOS 10.15 Catalina and newer
```
dxmh, line 348: I think it would be good if another macOS user could chip in regarding the removal of the security notice in https://github.com/rtitmuss/zmk/commit/3c098d065ea9c0067df2854ba2a4541cefd7330d, but it does seem to make sense as the linked page states the following:

> Note that this section does not apply to executables installed with Homebrew, since those are automatically un-quarantined by brew itself. This is however relevant for most 3rd Party Toolchains.

dxmh:
Just wanted to leave a reference here to https://github.com/zmkfirmware/zmk/pull/917 which has some similar changes to https://github.com/rtitmuss/zmk/commit/f50976f865863a6cda5f50af7dc63b7820a0b3bf.

docs/docs/development/setup.md
dxmh: Ah, so it seems this security workaround is not necessary _if_ the GNU ARM Embedded is installed via the `brew` command, but the warning _is_ necessary if users install “manually” via the steps on the page linked in the previous paragraph.

rtitmuss:
> Just wanted to leave a reference here to #917 which has some similar changes to [rtitmuss@f50976f](https://github.com/rtitmuss/zmk/commit/f50976f865863a6cda5f50af7dc63b7820a0b3bf).

I added the similar changes as suggestions to #917 and removed them from this commit.

docs/docs/development/setup.md
rtitmuss: Good catch. I have updated the PR and also removed the page linked. This is not required when following these instructions. Note also in the 'Prerequisites' section it states 'Homebrew is required to install the system dependencies'. If you choose to complete the setup without brew, then this security warning would apply earlier in the process too.

dxmh (approved):
Thanks. Given the above discussion (https://github.com/zmkfirmware/zmk/pull/932#pullrequestreview-747430665), these changes seem sensible to me. 👍

caksoylar (approved):

petejohanson (approved):

PR #935, opened 2021-09-07
feat(underglow): battery state of charge (SoC) animation

lewurm:
This change adds an underglow effect that visualizes the state of charge. If no battery is installed, blue will be sent to the LEDs. Otherwise the SoC is used to calculate a color with a green and red portion (green = full, red = empty).

 Note on testing: If USB is connected, the battery is charged and therefore `bt_bas_get_battery_level()` returns 100%. This is fine but when disconnecting the actual battery level needs up to a minute to be updated due to https://github.com/zmkfirmware/zmk/blob/2a9ab828b53b70539621ccb4ec4fa0cd47afc959/app/src/battery.c#L90-L95
 
 Is there a way to force a battery update whenever USB is disconnected? Same problem for connecting.

### DIFF
--- a/app/src/rgb_underglow.c
+++ b/app/src/rgb_underglow.c
@@ -12,6 +12,8 @@
 #include <math.h>
 #include <stdlib.h>
 
+#include <bluetooth/services/bas.h>
+
 #include <logging/log.h>
 
 #include <drivers/led_strip.h>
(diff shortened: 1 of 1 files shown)


DarkFox:
Just because no-one's working on it, it doesn't mean it's not still wanted.

PR #951, opened 2021-09-22
feat(shield): Add A. Dux shield

dxmh:
This PR adds a new shield for [Architeuthis Dux by Tapi][1] (aka A. Dux, A.D., "Giant squid").

The shield has been available at [`dxmh/zmk-architeuthis-dux`](https://github.com/dxmh/zmk-architeuthis-dux) for a couple of months already, but it would be great for Architeuthis Dux to be supported in ZMK properly.

![Wireless Architeuthis Dux with nice!nano controllers][2]

[1]: https://github.com/tapioki/cephalopoda/tree/main/Architeuthis%20dux
[2]: https://media.discordapp.net/attachments/855822038287908864/866315666802081792/image0.jpg


.github/workflows/build.yml
```diff
@@ -24,6 +24,8 @@ jobs:
           - nrfmicro_13
           - proton_c
         shield:
+          - architeuthis_dux_left
```
petejohanson, line 27: We are no longer adding to the `build.yml` in new shields, since it causes merge conflicts. We'll eventually drive the build from the metadata files as well.

app/boards/shields/architeuthis_dux/architeuthis_dux_left.conf
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) 2021 The ZMK Contributors
```
petejohanson, line 1: We usually add one single .conf file, not left and right, since that's the file our setup scripts look for.

petejohanson (changes requested):
A few initial comments, thanks!

app/boards/shields/architeuthis_dux/architeuthis_dux.keymap
```diff
@@ -0,0 +1,26 @@
...
+		base_layer {
+			bindings = <
+				&kp Q     &kp W     &kp E     &kp R     &kp T     &kp Y     &kp U     &kp I     &kp O     &kp P
+				&kp A     &kp S     &kp D     &kp F     &kp G     &kp H     &kp J     &kp K     &kp L     &kp SEMI
+				&kp Z     &kp X     &kp C     &kp V     &kp B     &kp N     &kp M     &kp COMMA &kp DOT   &kp SLASH
+				                              &kp TAB   &kp BSPC  &kp SPACE &kp ENTER
+			>;
+		};
+
+	};
```
dxmh, line 25: I've included this simple keymap comprised of only a QWERTY layer, because Architeuthis Dux doesn't have an official keymap. I am happy to put together a more featureful keymap if necessary.

.github/workflows/build.yml
dxmh: Ah, I thought so. Thanks. I've now removed the changes to `build.yml` from this PR.

I've also just raised a small PR aligning the New Shield documentation: https://github.com/zmkfirmware/zmk/pull/952

app/boards/shields/architeuthis_dux/architeuthis_dux_left.conf
dxmh: Got it, thanks! The left and right .conf files were from previously distributing the shield in a `zmk-config`-style repo. I've now removed them from this PR.

dxmh:
Thanks @petejohanson, I've updated the PR based on your comments. :+1:

caksoylar:
Would it be worth keeping the machine name ‘a_dux‘ similar to the [QMK PR](https://github.com/qmk/qmk_firmware/pull/14320)? It is pretty subjective, but might be user friendlier.

petejohanson:
> Would it be worth keeping the machine name `a_dux` similar to the [QMK PR](https://github.com/qmk/qmk_firmware/pull/14320)? It is pretty subjective, but might be user friendlier.

Given most of the conversations on the LPK have always used the shorthand, and the QMK naming is also doing so, I'm leaning towards agreeing w/ this. And indeed, *much* easier to type as needed if editing build.yml files, etc.

 Thoughts @dxmh?

dxmh:
Thanks @caksoylar, @petejohanson - using the short name for user-friendliness is an excellent point. 

I've just updated the PR so the shield now uses `a_dux` (and "A. Dux" for consistency).

app/boards/shields/architeuthis_dux/architeuthis_dux.keymap
petejohanson: I don't think this is a blocker for merging this, especially if upstream doesn't have a "stock" keymap. 

At some point, it might be nice for us to have a standard "default layout" for various stndard-ish layouts, like 36- and 34-key splits, that could be `#include`'d for this exact use case, but this is fine for now.

app/boards/shields/a_dux/a_dux.zmk.yml
```diff
@@ -0,0 +1,11 @@
+file_format: "1"
+id: a_dux
+name: A. Dux
+type: shield
+url: https://github.com/tapioki/cephalopoda/
```
petejohanson, line 5: Can we actually use the more direct URL here as well, same one from the README file:

```suggestion
url: https://github.com/tapioki/cephalopoda/
```
```suggestion
url: https://github.com/tapioki/cephalopoda/tree/main/Architeuthis%20dux
```

petejohanson (changes requested):
Just one more minor fix found while verifying the metadata work.

Thanks!

app/boards/shields/a_dux/a_dux.zmk.yml
dxmh: Absolutely! I've just updated the PR with this change to the URL. :+1:

petejohanson (approved):
LGTM!

PR #960, opened 2021-09-26
Add board/shield Check-list

Nicell:


.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+<!-- If you're adding a board/shield please fill out this check-list, otherwise you can delete it -->
+## Board/Shield Check-list
+ - [ ] This board/shield is tested working on real hardware
+ - [ ] Definitions follow the general style of other shields/boards upstream ([Reference](https://zmk.dev/docs/development/new-shield))
+ - [ ] `.zmk.yml` metadata file added
+ - [ ] Proper Copyright + License headers added to applicable files
+ - [ ] General consistent formatting of DeviceTree files
+ - [ ] Keymaps do not use deprecated key defines (Check using the [upgrader tool](https://zmk.dev/docs/codes/keymap-upgrader))
+ - [ ] `&pro_micro` used in favor of `&pro_micro_d/a` if applicable
+ - [ ] If split, no name added for the right/peripheral half
```
krikun98, line 10: Wasn't the new style to delete names on the central? This way, the only thing that advertises and is visible would have the "Left" added.

petejohanson: That's right, and that's what this line item is suggesting.

petejohanson (approved):
I like this! Would be nice if we could somehow have a selection of new PR templates, based on the kind of PR, but this will do for now. Thanks!

PR #961, opened 2021-09-28
Pimoroni PIM447 trackball support

cdc-mkb:
I wrote a Zephyr driver for the famous Pimoroni PIM447 trackball and added the necessary glue on top of the "mouse-pr" development branch.  This glue will likely change in the future since mouse support in ZMK is not stabilized yet; I actually open this pull-request to let other users play with this driver.

Note: thanks to a neat trick, small displacements are very precise whereas large displacements are quite fast; it never feels choppy at all. This has been successfully tested with the Zephirum, but it should work with any ZMK keyboard.


krikun98:
Huh, this looks sweet! You might as well PR this into the main mousekeys branch. Also, could you please record a GIF or a video?

app/src/mouse/pim447_trackball.c
```diff
@@ -0,0 +1,124 @@
...
+        LOG_ERR("Cannot get PIM447_TRACKBALL device");
+        return;
+    }
+
+    /* Event loop. */
+
+    bool button_press_sent   = false;
+    bool button_release_sent = false;
+
+    while (true) {
```
krikun98, line 57: Is this constantly polling the sensor? That would be quite a hit on power draw, and wouldn't take the configurable polling rate into account.

app/src/mouse/pim447_trackball.c
```diff
@@ -0,0 +1,124 @@
...
+#include <drivers/sensor.h>
+#include <logging/log.h>
+
+LOG_MODULE_REGISTER(PIM447, CONFIG_SENSOR_LOG_LEVEL);
+
+/*
+ * It feels more natural and more confortable to square the speed
+ * reported by the PIM447 trackball.
+ */
+static int16_t convert_speed(int32_t value)
```
krikun98, line 16: Did you try plugging this into the existing acceleration code?

app/src/mouse/pim447_trackball.c
```diff
@@ -0,0 +1,124 @@
...
+}
+
+#define STACK_SIZE 1024
+
+static K_THREAD_STACK_DEFINE(thread_stack, STACK_SIZE);
+static struct k_thread thread;
+
+int zmk_pim447_trackball_init()
+{
+    k_thread_create(&thread, thread_stack, STACK_SIZE, thread_code,
```
krikun98, line 119: Huh, this is indeed a neat trick. So that's how you run a specific function in a dedicated thread. 

I was thinking of converting the whole mouse emulation polling system to this principle instead of a workqueue, but the thread probably does not need to be constantly active? For emulation, it's trivial - we could wake the thread up only when a key is pressed, but here... Maybe we could check every two or three polling intervals, and only engage the trackball constantly if it is indeed moving?

cdc-mkb:
> Huh, this looks sweet! You might as well PR this into the main mousekeys branch. Also, could you please record a GIF or a video?

Of course, here it is (the link will expire after 190 days): https://webmshare.com/play/ew0rP

krikun98:
> > Huh, this looks sweet! You might as well PR this into the main mousekeys branch. Also, could you please record a GIF or a video?
> 
> Of course, here it is (the link will expire after 190 days): https://webmshare.com/play/ew0rP

Excellent! So running everything in a dedicated thread (and not a workqueue) is indeed the answer! 
Nevertheless, I still think power consumption needs to be taken into account. Are you on the ZMK Discord? I'd love to have an architecture discussion over there.

app/src/mouse/pim447_trackball.c
cdc-mkb: It is polling the sensor every 10 ms; see `k_sleep(K_MSEC(10))` at the end of this loop.  Anyway, it tried to write a driver in interrupt-mode using the INT line of this sensor, but this is not possible because:
- this INT signal toggles during I2C transferts: 
- this INT signal is triggered (low level) without any reason a couple of ms after being cleared:

The official "driver" from Pimoroni uses this INT signal, but it turns out it uses it the wrong way:

        while not trackball.get_interrupt():
            time.sleep(0.001)
        [...]

        def get_interrupt(self):
            if self._interrupt_pin is not None:
                return GPIO.input(self._interrupt_pin) == 0 # [my comment] poll pin value
            else:                                           # [my comment] or
                value = self.i2c_rdwr([REG_INT], 1)[0]      # [my comment] poll register value
                return value & MSK_INT_TRIGGERED

Maybe I should contact Pimoroni to know if there is a way to get a clean INT signal.

I didn't take into account the power consumption because I only use the USB HID, but you are right: this might not be a good driver for wireless keyboard.

app/src/mouse/pim447_trackball.c
cdc-mkb: No I didn't because I developed this driver independently from your branch initially.

app/src/mouse/pim447_trackball.c
cdc-mkb: I don't know if a workqueue is cheaper than a thread in terms of memory or power, but it seems a bit more complex to write/maintain.

krikun98: A workqueue is the way everything is done now, and it's not suitable for mouse polling because they're not meant for time-sensitive tasks. The whole polling system needs to be rewritten to use a bare thread instead.

However, that bare thread has to work smartly. I don't think we should run the timer and the thread constantly. With emulation, we could send it to sleep when no keys are pressed and only wake it up when we have something to send (with a mutex and something like `.await()/.notify`, or however it works in Zephyr).

Here, it's more difficult to think of something, but maybe we could define a "passive state" that only polls every few intervals and wakes things up if there's any movement?

app/src/mouse/pim447_trackball.c
krikun98: I'd appreciate it if you contacted them, but we should still think of something. There's a possibility we can contact Pimoroni, but we're unlikely to get through to the creators of old Thinkpad trackpoints, and the constant polling issue is likely to pop up there too.

cdc-mkb: Ok. I've just asked for support on Pimoroni forum: https://forums.pimoroni.com/t/pim447-trackball-how-to-get-a-clean-int-signal/17968

cdc-mkb:
> Are you on the ZMK Discord? I'd love to have an architecture discussion over there.

Sadly I don't have enough time to be useful in this regard.  Sorry.

app/drivers/sensor/pim447_trackball/Kconfig
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2021 Cedric VINCENT
+# SPDX-License-Identifier: Apache-2.0
+
+config ZMK_PIM447_TRACKBALL
+	bool "ZMK PIM447 trackball"
+	depends on I2C
+	help
+	  Enable I2C-based driver for Pimorono PIM447 trackball.
```
krikun98, line 8: Shouldn't it be "Pimoroni"?

cdc-mkb: Oh right, thanks for the review!  I'll fix that when I rebase this driver on top of the mouse emulation branch once it gets integrated officially.

grasegger:
@cdc-mkb do you have an example shield and keymap configuration? I coudn't find a repository for the zephirium where the trackball driver was used. I'd like to use a trackball with my zmk setup, but even when using your fork I am a bit lost on how to implement it in a keymap.

cdc-mkb:
@grasegger you can take a look at my configuration here: https://github.com/cdc-mkb/zmk/commit/b2a172b219b0de95e3ca2de39d680a4ab7db270c Be careful: you only need to adjust your .conf and .overlay files, there is no need to change anything in the .keymap file (my configuration also uses the "macro" WIP feature, that's why the .keymap is modified here but this is not related to the PIM447 trackball driver).

grasegger:
> @grasegger you can take a look at my configuration here: [cdc-mkb@b2a172b](https://github.com/cdc-mkb/zmk/commit/b2a172b219b0de95e3ca2de39d680a4ab7db270c) Be careful: you only need to adjust your .conf and .overlay files, there is no need to change anything in the .keymap file (my configuration also uses the "macro" WIP feature, that's why the .keymap is modified here but this is not related to the PIM447 trackball driver).

Thanks - that helped a lot. 

There is no possibility to rebind the actions of the trackball to e.g. scroll, if I see that correctly?

cdc-mkb:
@grasegger you are right, this is not implemented yet.  Using this tiny trackball to scroll on documents/webpages is a feature (in QMK) I like a lot, so I'll surely add this feature to this pull request soon.

grasegger:
@cdc-mkb One last questions, since I don't have the hardware to test here yet: Would it be possible to support two trackballs on two split keeb halfs? One for scrolling, one for cursor movements, and with different bindings (left click, right click)? **Is there any way I can help?** I know a *bit* of C/C++, but never worked with firmware.

cdc-mkb:
> @cdc-mkb One last questions, since I don't have the hardware to test here yet: Would it be possible to support two trackballs on two split keeb halfs? One for scrolling, one for cursor movements, and with different bindings (left click, right click)? **Is there any way I can help?** I know a _bit_ of C/C++, but never worked with firmware.

@grasegger So far I assumed only one trackball is used, but I'm pretty sure it should be technically feasible to make a driver that works for several trackballs at the same time.  I didn't think about this use case so far, so feel free to send me ideas, patches, ... for review if you wish.

Anyway, I added a couple of new features to this PIM447 driver for ZMK, for instance it it now possible to use it as a mouse pointer and as mouse wheels (the mode can be changed dynamically).  It is also possible to place the breakout board in any position thanks invert/swap transformation on axes.  See https://github.com/cdc-mkb/zmk/commit/45e1ae7f29abf37dd96387983107370e28e9560c for information on how to use these new features.

wizar:
@krikun98 @cdc-mkb Any idea of how badly it would affect battery life? I really would like to use this trackball but a little bit concened about how much it will drain.

cdc-mkb:
New version:
- it is now possible to use mouse emulation key behaviors (ex. &mkp, &mwh) and the trackball at the same time
- the trackball key behaviors support (&pim447*) has been reworked; it was buggy and clumsy.

cdc-mkb:
> @krikun98 @cdc-mkb Any idea of how badly it would affect battery life? I really would like to use this trackball but a little bit concened about how much it will drain.

@wizar I have no idea since I'm only using the USB HID.  Pimoroni hasn't replied to my question about the INT signal so far :( (https://forums.pimoroni.com/t/pim447-trackball-how-to-get-a-clean-int-signal/17968)

app/src/mouse/pim447_trackball.c
isteven: Hi both, really appreciate the works done so far on this trackball implementation. I'm one of those who look forward to see this work properly & get merged to master :)

Regarding the clean INT signal issue, is it a software or hardware limitation of the trackball? 

I'm asking this because I am thinking to build a wireless keyboard (on ZMK obviously) with Pimoroni trackball and OLED. OLED is already something that could easily drain battery, and if the trackball does as well, I may have to re-consider my plan. 

Might worth a shot checking in their unofficial discord (https://discord.gg/hr93ByC) or raising an issue in their Github page (https://github.com/pimoroni/trackball-python).

grasegger:
@cdc-mkb Can you take a look at my shield config? Keys are working, but the trackball isn't. 

SDA is connected to pin 006 (D0) and SCL to pin 008 (D1)

https://github.com/grasegger/keyboard-layout/tree/anna/config/boards/shields/anna

I tried adding ```&i2c0 {
	compatible = "nordic,nrf-twi";
	sda-pin = <0>;
	scl-pin = <1>;
};``` to the board/nice_nano.overlay

cdc-mkb:
@grasegger My reply here: https://github.com/grasegger/keyboard-layout/pull/1

genkobar:
Hey @cdc-mkb,

I'm trying to build a custom shield with pimoroni support off of your branch, but it fails because it doesn't recognize the flag CONFIG_ZMK_TRACKBALL_PIM447... can you point me in the right direction for what needs to be fixed? I'm completely new to ZMK still, so I'd appreciate the help :)

Here's the line that fails in the Github Action:
https://github.com/genkobar/zmk-config/runs/4342798400?check_suite_focus=true#step:9:35

Thansk for working on this feature support!

fossephate:
A github thread I managed to find with some potentially helpful info on interrupt pin / usage:

https://github.com/pimoroni/trackball-python/issues/5

minjaep:
Thanks! I just managed to build a PMW3360 based trackball, just replacing the driver in this PR by that implemented in nrf sdk (which cannot be used directly in zmk unfortunately due to the license issue). I'm now trying to modify it to irq based.

grasegger:
@cdc-mkb Sadly I can't get it working the the n!nv2 I have. After I tried it on my on board without luck I connected the wires to the kyria, which has dedicated sda, scl, vcc and gnd on the pcb. 

I played around with the config a bit and stumbled upon `Check these unsatisfied dependencies: I2C` which I worked around with a `CONFIG_I2C`, after that I got no errors or warnings regarding the 447 anymore. Sadly neither via usb nor bluetooth the cursor moves or scrolls. 

https://github.com/grasegger/keyboard-layout/tree/main/nice_nano_v2/kyria

w84miracle:
@cdc-mkb the documentation is different from the Kconfig you wrote
should be `CONFIG_ZMK_TRACKBALL_PIM447=y` instead of `CONFIG_ZMK_PIM447_TRACKBALL=y`

devenv:
@grasegger I made it work with a nano (https://github.com/devenv/zmk-config/tree/trackball)... For some time (like 10 seconds), then it fails on this error and stops working:

```
<err> i2c_nrfx_twi: Error 195952642 occurred for message 0
<err> trackball_pim447: Sensor reg read byte failed
<err> PIM447: Failed to fetch TRACKBALL_PIM447 sample
```

app/src/mouse/pim447_trackball.c
olidacombe: If you're prepared to do something so dirty, you can debounce the interrupt signal with a couple of passive components, keeping the int line from passing the low threshold during i2c noise, and dropping fast enough on a true interrupt event:

Channel 2 here is the treated interrupt signal during i2c:
DS1054Z_DS1ZA233608483_2022-03-09_21 12 03

And here it is when the pimoroni fires a true interrupt:
DS1054Z_DS1ZA233608483_2022-03-09_21 15 43

If that's not too gross for the community (and there's not a whole ton of i2c activity on a given board that might eventually drive the level low enough for a false trigger), then is this enough to progress with a power-friendly version?

eviallet (approved):

alexmaloteaux:
hi regarding the INT  pin, it looks like you need to init the trackball with a special i2c write for it to work. At least there is an example here : https://github.com/pimoroni/pimoroni-pico/blob/3ce347799b74b81bc18611867dea202e6e1f875f/drivers/trackball/trackball.cpp#L89

app/src/mouse/pim447_trackball.c
DelusionalLogic: I've been looking into this INT problem, and at least on the specimen I've got the only problem is the noise during I2C. If you just guard the interrupt with a semaphore that you release when you're done reading the values, the INT noise doesn't really matter.

One problem you'll have then is that the falling edge of the interrupt may happen before you release the semaphore, which would block you from ever discovering that the PIM has data to read. The workaround there is to manually check if the interrupt pin is low after releasing the semaphore and in that case retaking it and processing the interrupt.

The final catch is that you'll have to wait for around 8 microseconds from when you end i2c communication before releasing the semaphore to avoid the final INT noise.

FearlessSpiff:
I am tinkering with a Pimoroni atm. Would love to see this on ZMK. Just found out that there is a new firmware for it which fixes the interrupt issue! See the OP post on their forum: https://forums.pimoroni.com/t/pim447-trackball-how-to-get-a-clean-int-signal/17968/10
Hope this gives some new life into this PR!

app/drivers/sensor/trackball_pim447/trackball_pim447.c
```diff
@@ -0,0 +1,167 @@
...
+}
+
+static struct trackball_pim447_data trackball_pim447_data;
+
+static const struct sensor_driver_api trackball_pim447_api = {
+    .sample_fetch = trackball_pim447_sample_fetch,
+    .channel_get  = trackball_pim447_channel_get,
+};
+
+DEVICE_DT_INST_DEFINE(0, &trackball_pim447_init, device_pm_control_nop,
```
daniel2887, line 165:
```suggestion
DEVICE_DT_INST_DEFINE(0, &trackball_pim447_init, NULL,
```

`device_pm_control_nop` has been deprecated in favor of NULL, per [these Zephy release notes](https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/releases/release-notes-2.6.rst).

daniel2887:
Question to anyone who has this up and running:

Does moving the trackball around feel a bit... "stiff", or is it just my keeb/me? Moving the trackball in circles generates a cursor movement that feels like discrete up/down/left/right movements, rather than the cursor moving diagonally (i.e., both axes updating simultaneously). Reading through the pim447 app loop, I see all 3 axes states (x, y, button) being set within the same mouse report, so I'm wondering if I'm observing a HW limitation (which would be quite discouraging) or something that still needs to be improved in SW.

For reference, I have this up on a nice!nano v2 (driving a Corne shield, if it matters), connected over USB to a Windows 10 host.

DarkFox:
Just because no-one's working on it, it doesn't mean it's not still wanted.

twinoner:
do anyone know how to integrate this pim447 with nice!view?
I cannot do it for some reason or do I have to rewire the pins for any both of them? (the pim447 module or the pins for the nice!view)

twinoner:
> do anyone know how to integrate this pim447 with nice!view? I cannot do it for some reason or do I have to rewire the pins for any both of them? (the pim447 module or the pins for the nice!view)

so.. I'll answer myself: I've added `nice_view_adapter nice_view` to the shields on the `build.yaml`
it successfully builds and the nice-view works but the pim447 doesn't work and I can't seem to figure out how to make both work. I didn't do any changes to nice-view and the pim447 is connected to it's needed pins ("VCC, 17, 20 and Gnd" on nice!nano for "3.5v, scl, sda and gnd" on pim module)

caksoylar:
We have no idea what your hardware setup is like. If you use the same pins for nice!view and pim447 they won't work, because one uses SPI and other I2C. If you ask in the [Discord server](https://zmk.dev/community/discord/invite) you can provide more information, and reach more people that might have experience with this.

twinoner:
> We have no idea what your hardware setup is like. If you use the same pins for nice!view and pim447 they won't work, because one uses SPI and other I2C. If you ask in the [Discord server](https://zmk.dev/community/discord/invite) you can provide more information, and reach more people that might have experience with this.

Yeah! Been looking on discord and didn't found anything helpful.

https://github.com/twinoner/zmk-config-trackboll this is what I'm working on now. It's a new fork again.. if you're curious, go ahead.

Thanks! :)

caksoylar:
As an aside, ZMK [pointing device feature](https://zmk.dev/docs/development/hardware-integration/pointing) supports using generic Zephyr input drivers, such as the Cirque one [here](https://github.com/petejohanson/cirque-input-module) or PMW3610 [here](https://github.com/badjeff/zmk-pmw3610-driver). You'll notice that there's barely any ZMK-related code in either. This, along with [module support](https://zmk.dev/docs/features/modules), means that driver code like this doesn't need to be integrated into ZMK and likely[^1] will not be.

[^1]: I am not the decision maker but pushing such functionality to modules is the direction we've been heading to.

twinoner:
Yeap! I'm familiar with pointing official support.
But, at the time, I don't plan on using a cirque or a larger trackball.
I like the pimoroni because is pretty small and can use it when I'm away (I'm aware of battery issues but my keyboard uses PS3 controller batteries, so I don't mind charging the pim447 side every week or even add a second battery in series)
I have a "regular size" trackball (Logitech ergo don't remember the model, mx570 maybe)

caksoylar:
> Yeap! I'm familiar with pointing official support.

Right, that wasn't addressed to you specifically but was a general comment.

> Yeah! Been looking on discord and didn't found anything helpful.
> 
> https://github.com/twinoner/zmk-config-trackboll this is what I'm working on now. It's a new fork again.. if you're curious, go ahead.
> 
> Thanks! :)

You can make a help-forum post, or just post in the pointing-devices channel. To be clear, I am not sure how likely you are to get help given the very custom things needed to get such a driver working. But it certainly beats debugging over GH PR comments.

PR #965, opened 2021-10-02
docs(docs): Document up-to-date Node.js repo

bcat:
For example, Debian Bullseye (current stable) packages Node.js version
12, which is too old to build the docs successfully. At least version 14
is required, and version 16 is current. General advice seems to be to
install from the NodeSource repo instead of your distro's repo, so I
added a suggestion to the docs.


caksoylar (approved):
Thanks, this is very good to have in the docs! Unless someone would like us to recommend a specific version or a different installation method, I think this looks good.

Issue #665, opened 2021-02-04
Refactor battery_voltage_divider Kconfig to not select ADC


Nicell:
Currently the `battery_voltage_divider` sensor selects `ADC`, which prevents us from putting our sensor Kconfig in an if block.

It should be refactored so that it only depends on it, and then boards should enable `ADC` as they currently do for `GPIO`. We'll only need nRF52 boards to enable `ADC` though as currently the driver doesn't support other types of ADC.

petejohanson:
@Nicell Is this really needed any more now that we default this based on the DTS, e.g. `default $(dt_compat_enabled,$(DT_COMPAT_ZMK_BATTERY_VOLTAGE_DIVIDER))` ?

myst729:
Is this refactor eventually aiming for a generic ADC driver? I'm looking forward to seeing that happen, as I would like to try plenty of analog peripherals, such as joystick, linear slider, hall effect sensor, etc.

PR #966, opened 2021-10-02
refactor(boards): Refactor battery_voltage_divider Kconfig

careyk007:
As requested in #665, `ZMK_BATTERY*` Kconfig options rely on `ADC` being enabled. I have also added `CONFIG_ADC=y` to all nrf52 boards.

For verification, I ran `west build -p -b proton_c -t menuconfig -- -DSHIELD=kyria_left` and verified that `Modules->drivers` did not contain any references to battery monitoring. Likewise, I ran `west build -p -b nice_nano -t menuconfig -- -DSHIELD=kyria_left` and verified that the battery monitoring options were present.


petejohanson:
This can be closed, we've moved to the way upstream Zephyr does this, enabling automatically only if we find a device in the DT w/ the compatible value that is enabled. Thanks for the PR, regardless!

PR #969, opened 2021-10-05
feat(display): Add optional full refresh timer.

petejohanson:
* Useful for EPD that ghost after several partial updates. Only a
  temporary fix, tracking number of partial updates and triggering
  full refreshes based on that would be better.

### DIFF
--- a/app/src/display/Kconfig
+++ b/app/src/display/Kconfig
@@ -71,6 +71,13 @@ endchoice
 
 endif # ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
 
+config ZMK_DISPLAY_FULL_REFRESH_PERIOD
+    int "(Optional) Period to issue a full refresh to the display (in seconds)"
+    default 0
+    help
+      Period in seconds for how often to completely refresh/redraw the whole screen.
+      Most useful for e-ink/EPD displays that require occasional full redraws.
+
 rsource "widgets/Kconfig"
 
 endif
--- a/app/src/display/main.c
+++ b/app/src/display/main.c
@@ -50,6 +50,17 @@ struct k_work_q *zmk_display_work_q() {
 #endif
 }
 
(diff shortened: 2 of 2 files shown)


caksoylar:
I have been testing this on my Corne-ish Zen and it works as expected. I used timeouts of 60 and 120 seconds so far.

One behavior I thought interesting -- which might have been expected -- is that the idle timer (the 30 second one) delays the timer. So the refresh happens when each half wakes from idle if the refresh timer expired during that time. I think it's probably a good thing since the display wouldn't be needlessly refreshing while the keyboard sits idle.

Given this behavior, one thing I have been meaning to check is if it increases the number of refreshes after waking from deep sleep as well. It wouldn't be a big deal but I can update when I do.

caksoylar:
I also wanted to note that my subjective experience with the e-ink screens of the Zen is already much improved with this PR, since practically whenever I need to look at the screen it no longer has dim portions. (I think this is also helped by the fact that I disabled layer widget updates for my non-base layers.) I also don't particularly notice the periodic full refreshes with the timeouts I use. Thank you for your work!

caksoylar:
I noticed an interesting interaction with this period value and the idle timeout: If you set it high enough and higher than the idle timeout (which is 30s by default), it pretty much never triggers the refresh because going into idle stops the timer, then waking up restarts it from zero (or that's my guess). In order to trigger the refresh you have to prevent going into idle state for the full period. Maybe making it so the timer isn't reset when going into idle would help?

Side note: This didn't used to happen until I rebased my Zen on the `main` branch and applied patch as-is, but it is probably not worth chasing the difference. (What used to happen was that display refreshed whenever it came out of the idle state, I am guessing if the period had elapsed by then.)

caksoylar:
We have a few people now using this PR through one of my branches on either v1 or v2 Corne-ish Zens. Is there a plan to replace this mechanism in the short term? If not, I don't see an issue with publishing and merging this.

(I no longer have any issues with idle timeout that was described in the above comment.)

PR #975, opened 2021-10-12
refactor: Config setup scripts to support onboard keyboards.

petejohanson:
* Use unified config template repo that uses an external
  build matrix YAML file.
* Proper handling for onboard keyboards, including splits, and
 supports for appending the right build matrix once selected.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -13,7 +13,9 @@
         "@fortawesome/fontawesome-svg-core": "^1.2.32",
         "@fortawesome/free-solid-svg-icons": "^5.15.3",
         "@fortawesome/react-fontawesome": "^0.1.14",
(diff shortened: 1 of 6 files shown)


petejohanson:
Ok, setup scripts for the test deploy are up at:

* powershell: `iex ((New-Object System.Net.WebClient).DownloadString('https://deploy-preview-975--zmk.netlify.app/setup.ps1'))` 
* bash: `bash -c "$(curl -fsSL https://deploy-preview-975--zmk.netlify.app/setup.sh)" `

I've created a few when testing these scripts:

* https://github.com/petejohanson/corne-user-repo-test
* https://github.com/petejohanson/nice60-zmk-config

And they all seemed to do the right thing, including for boards!

I also really like that we can use this as a basis for making the scripts re-executable, for adding additional setups to existing repos.

PR #984, opened 2021-10-17
fix(split): Properly rediscover peripherals.

petejohanson:
* Handle disconnects of peripherals, and properly clean up
  and resume discovering for when the peripheral re-appears.


petejohanson (commented):
Had a few other testers on Discord note this worked for them. Merging.

PR #990, opened 2021-10-20
Add support for Knob Goblin shield

lucasuyezu:
## TODO before opening for reviews
- [x] Add scaffolding code
- [x] Copy keymap from qmk
- [x] Flash on a nice!nano
- [x] Test on cable
- [x] Test on Bluetooth

## Board/Shield Check-list

### DIFF
--- a/app/boards/shields/knob_goblin/Kconfig.defconfig
+++ b/app/boards/shields/knob_goblin/Kconfig.defconfig
@@ -1,0 +1,45 @@
+# Copyright (c) 2022 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+if SHIELD_KNOB_GOBLIN
+
+config ZMK_KEYBOARD_NAME
+  default "Knob Goblin"
+
+if ZMK_DISPLAY
+
+config I2C
+  default y
+
+config SSD1306
+  default y
+
+config SSD1306_REVERSE_MODE
+  default y
+
+endif # ZMK_DISPLAY
+
+if LVGL
+
+config LVGL_HOR_RES_MAX
+  default 128
+
+config LVGL_VER_RES_MAX
+  default 32
+
+config LVGL_VDB_SIZE
+  default 64
+
+config LVGL_DPI
+  default 148
+
+config LVGL_BITS_PER_PIXEL
+  default 1
+
+choice LVGL_COLOR_DEPTH
+  default LVGL_COLOR_DEPTH_1
+endchoice
+
+endif # LVGL
+
+endif # SHIELD_KNOB_GOBLIN
--- a/app/boards/shields/knob_goblin/knob_goblin.conf
+++ b/app/boards/shields/knob_goblin/knob_goblin.conf
@@ -1,0 +1,9 @@
+# Copyright (c) 2022 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+# Uncomment to enable Encoders
+CONFIG_EC11=y
+CONFIG_EC11_TRIGGER_GLOBAL_THREAD=y
+
+# Uncomment the following line to enable the Knob Goblin OLED Display
+CONFIG_ZMK_DISPLAY=y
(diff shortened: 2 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -41,6 +41,7 @@ jobs:
           - jian_right
           - jorne_left
           - jorne_right
+          - knob_goblin
```
Nicell, line 44: Please don't edit this file, we're changing how we do builds very soon!

app/boards/arm/nice_nano/nice_nano_v2.yaml
```diff
@@ -8,6 +8,7 @@ toolchain:
   - xtools
 supported:
   - adc
+  - encoders
```
Nicell, line 11: The nice!nano itself doesn't support encoders, no need to change this. A shield that the nice!nano plugs into may support encoders though! This also is a Zephyr specific file, which doesn't know anything about these encoders.

app/boards/shields/knob_goblin/Kconfig.defconfig
```diff
@@ -0,0 +1,45 @@
+if SHIELD_KNOB_GOBLIN
+
+config ZMK_KEYBOARD_NAME
+  default "Knob Goblin"
+
+config ZMK_USB
+  default y
+
```
Nicell, line 8: You don't need to add this configuration. The board that is plugged in will dictate whether this is needed.
```suggestion
```

app/boards/shields/knob_goblin/knob_goblin.overlay
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (c) 2021 The ZMK Contributors
```
Nicell, line 2:
```suggestion
 * Copyright (c) 2022 The ZMK Contributors
```

Nicell (changes requested):
Thanks for this PR! Have a few comments

lucasuyezu:
@Nicell comments addressed.

app/boards/shields/knob_goblin/Kconfig.defconfig
```diff
@@ -0,0 +1,42 @@
+if SHIELD_KNOB_GOBLIN
```
Nicell, line 1: Could you add the copyright header to here too?
```suggestion
# Copyright (c) 2022 The ZMK Contributors
# SPDX-License-Identifier: MIT

if SHIELD_KNOB_GOBLIN
```

app/boards/shields/knob_goblin/knob_goblin.conf
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2022 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+# Uncomment to enable Encoders
+CONFIG_EC11=y
+CONFIG_EC11_TRIGGER_GLOBAL_THREAD=y
+
+# Uncomment the following line to enable the Knob Goblin OLED Display
+CONFIG_ZMK_DISPLAY=y
```
Nicell, line 9: These should probably be commented out by default, unless every knob goblin has these

Nicell (changes requested):

app/boards/shields/knob_goblin/knob_goblin.conf
lucasuyezu: Every knob goblin ships with those, but the user may choose to not solder the screen for example. What should I do in this case?

Nicell: In this case, I think the decision should be up to you on whether they should be turned on by default. I would say it seems sane either way.

lucasuyezu: In that case I'd rather leave it like this, that way the default route to install ZMK would work with zero code changes.

Nicell (approved):
Thanks!

PR #999, opened 2021-10-23
feature: LED indicators

bortoz:
#115

This PR adds support for LED indicators, such as Caps Lock LED and Num Lock LED.


joelspadin:
FYI, this seems to have some overlap with #904. It would be nice to make sure we don't end up with competing PWM LED systems.

bortoz:
> FYI, this seems to have some overlap with #904. It would be nice to make sure we don't end up with competing PWM LED systems.

LED indicators do not require PWM LEDs, you can use any LED driver supported by Zephyr. I added the PWM LED driver because the one in Zephyr 2.5 is not working properly but it is only a temporary solution until Zephyr is upgraded.

bortoz:
> I think it would be better to do two prs, a pr to read the report and another for the implementations

Is there a particular reason for doing this?

bortoz:
> For two reasons, first so that this is faster and easier to do the code review and second because of what has been working in backlight, the implementation may change, in my design the caplock led is the last led of the underglow, is it possible configure it with your implementation?

This PR and backlight PR are independent of each other, you can use only one of them or both together without any problem. As I said earlier this PR does not depend on PWM LEDs, any LED driver can be used.

Regarding your behavior, I don't think it's achievable as the underglow API doesn't allow changing individual LEDs (and modifying those API is beyond the intentions of this PR). However the current implementation of LED indicators allows you to implement a custom behavior, you can find more details in the [documentation](https://deploy-preview-999--zmk.netlify.app/docs/features/led-indicators#custom-behavior).

I think this implementation is fine as it is, most users will use the standard implementation, users like you who want different behaviors can implement custom behavior without modifing ZMK codebase.

app/include/zmk/events/led_indicator_changed.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2021 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#pragma once
+
+#include <zmk/led_indicators.h>
```
petejohanson, line 9: Since we only need this for the flags type, not any implementation of the functions, can we move the flag type alias to something like:

```suggestion
#include <zmk/led_indicators_types.h>
```

app/src/hog.c
```diff
@@ -85,6 +91,31 @@ static ssize_t read_hids_input_report(struct bt_conn *conn, const struct bt_gatt
...
+        LOG_ERR("LED report is malformed: length=%d", len);
+        return BT_GATT_ERR(BT_ATT_ERR_INVALID_ATTRIBUTE_LEN);
+    }
+
+    struct zmk_hid_led_report_body led_report;
+    memcpy(&led_report, (uint8_t *)buf + offset, sizeof(struct zmk_hid_led_report_body));
+
+    struct zmk_host_report_source source = {
+        .endpoint = ZMK_ENDPOINT_BLE,
+        .profile = bt_conn_index(conn),
```
petejohanson, line 111: The connection index isn't the same as the profile #. Why is this info needed? Do we switch the LED indicator based on the active profile/connection? (If not, we should)

app/src/host.c
```diff
@@ -0,0 +1,64 @@
...
+        break;
+    default:
+        LOG_WRN("Unsupported host report: %d", report_id);
+        break;
+    }
+}
+
+#endif /* CONFIG_ZMK_USB */
+
+struct zmk_host_report_source zmk_host_current_source() {
```
petejohanson, line 45: This seems more appropriate for the endpoint API (endpoint.c) than tied to the host report code here. Why here?

app/src/host.c
```diff
@@ -0,0 +1,64 @@
...
+#include <zmk/host.h>
+#include <zmk/led_indicators.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#ifdef CONFIG_ZMK_USB
+
+#define LED_REPORT_ID 1
+
+void zmk_host_process_usb_report(struct zmk_host_usb_report report) {
```
petejohanson, line 22: Why is this here, and not in `app/src/usb.c`?

Seems like USB specifics leaking in here.

petejohanson (changes requested):
A few initial thoughts. If you don't want to do this as two PRs, please at least split the code additions into two parts:

* The generic state management and ties into USB and GATT HID systems.
* The LED display handling of the new state.

That will make this way easier to review.

Thanks!

app/src/hog.c
bortoz: Yes, we switch the LED indicators only on the active profile but we still receive reports from other profiles and we should handle them.

ReFil:
Been testing this over the past little while, usb behaviour is flawless after i added a subscriber to update when it's connected to handle state changes whislt disconnected, although i'm having significant trouble getting it to work over BLE, tried windows mac and linux and i cannot make it work, have you got any suggestions?

bortoz:
> Been testing this over the past little while, usb behaviour is flawless after i added a subscriber to update when it's connected to handle state changes whislt disconnected, although i'm having significant trouble getting it to work over BLE, tried windows mac and linux and i cannot make it work, have you got any suggestions?

I'm not sure why it doesn't work, but I haven't had a lot of time in the past months to test it properly

ReFil:
i've found the problem, it's that the source is not being set properly in your indexing function in led_indicators.c. i'd reccomend replacing the source index setup with what i've done here: https://github.com/ReFil/zmk/commit/47330785161df4b16bc60ff86ff8d0c787b39fe4 which i've been testing across multiple bluetooth profiles as well as usb

bortoz:
> i've found the problem, it's that the source is not being set properly in your indexing function in led_indicators.c. i'd reccomend replacing the source index setup with what i've done here: [ReFil@4733078](https://github.com/ReFil/zmk/commit/47330785161df4b16bc60ff86ff8d0c787b39fe4) which i've been testing across multiple bluetooth profiles as well as usb

Thanks! I will look into that

bortoz:
@ReFil I made a lot of improvements, can you test the latest version?

ReFil:
Spotted another bug, currently consumer reports dont write. the fix entailed changing https://github.com/zmkfirmware/zmk/blob/025bfd92900eab6d186a8dac2cc4cff5fc6e68b7/app/src/hog.c#L261 the size from 10 to 12

app/include/dt-bindings/zmk/led_indicators.h
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#pragma once
+
+/* Led Num Lock */
+#define LED_NUMLOCK (0)
```
petejohanson, line 10: Any reason we're adding this and duplicating values already defined in `app/include/dt-bindings/zmk/hid_usage.h`

app/dts/bindings/zmk,led-indicators.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2021 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+description: LED indicators
+
+compatible: "zmk,led-indicators"
```
petejohanson, line 6: Is there a reason not to use the `gpio-leds` bindings and just leverage that?

Seems like we could do:

```
/ {
  chosen {
    zmk,led-indicators = &indicators;
  };

  indicators: gpio_indicators {
    compatible = "gpio-leds";
      led_numlock {
        gpios = <&gpio1 1 GPIO_ACTIVE_HIGH>;
     };
  };
};
```

And leverage it directly. Check out `app/src/backlight.c` for an example where we are already using this.

petejohanson (commented):
A few questions.

app/dts/bindings/zmk,led-indicators.yaml
bortoz: There are two downsides with your proposal:

* How to find which indicator corresponds to a LED? We can use the child name, as in your example, but then we'll have to find a way to translate the name from a string to the indicator. This approach also doesn't allow multiple LEDs with the same indicator.
* This approach works for LED_GPIO and LED_PWM drivers, but it doesn't work with other drivers, however I guess very few users have other drivers.

If you think this approach is better than the current one, I can change it.

app/src/usb.c
```diff
@@ -34,8 +37,83 @@ static K_SEM_DEFINE(hid_sem, 1, 1);
...
 
+static void out_ready_cb(const struct device *dev) {
+    size_t report_length;
+    int rc = hid_int_ep_read(hid_dev, NULL, 0, &report_length);
+    if (rc != 0) {
+        LOG_ERR("Failed to read USB report length: %d", rc);
+        return;
+    }
+
+    uint8_t *report = k_malloc(report_length);
```
petejohanson, line 48: Ideally, we should try to avoid dynamic `k_alloc` like this if at all avoidable.  Until we have more than one output report, just use a stack allocated array using `sizeof(struct zmk_hid_led_report)` for the length.

app/include/dt-bindings/zmk/led_indicators.h
bortoz: These values are not exactly the same, they represent the index of the indicator within the bitmask. Having these values can be useful in the future if we want to add other indicators whose HID values are not consecutive.

app/dts/bindings/zmk,led-indicators.yaml
bortoz: I like this approach and I agree that having multiple LEDs for the same indicator is not very useful.

app/src/usb.c
bortoz: Although we only have one report output, the device can still receive packets of arbitrary length and if these packets are not handled correctly they could cause inconsistency in the USB buffer.

app/dts/bindings/zmk,led-indicators.yaml
petejohanson: I would suggest we go this route then, for the DTS side of things.

app/include/dt-bindings/zmk/led_indicators.h
petejohanson: Ok, I would suggest:

1. Don't create duplicates.
2. If they're not needed in any DTS files (don't see why they would be, given our conversation about chosen nodes), then don't nest within `dt-bindings/`.
3. I might suggest using a naming convention like `ZMK_LED_NUMLOCK_BIT` for these.

I don't actually see these being *used* anywhere in the code. Am I missing something?

app/src/usb.c
petejohanson: Are we sure the USB buffer stuff won't handle this? I'm not seeing anything in the docs.

bortoz: If we received a 3-byte report and read only the first 2 bytes, 1 byte would remain in the buffer. Doing so, on the next reading we would first read the previous byte remaining in the buffer causing incorrect behavior.

app/include/dt-bindings/zmk/led_indicators.h
bortoz: These values should be used in the devicetree nodes, currently no boards/shields support LED indicators so you won't find these values in the code.

```
led_indicators {
    compatible = "zmk,led-indicators";
    led_numlock {
        ctrl = <&pwmleds>;
        index = <0>;
        binding = <LED_NUMLOCK>; 
    };
};
```

Since we have changed the devicetree definition, I will delete this file and move these definitions somewhere else.

moergo-sc:
It looks like currently it is not possible to overlay the caps/scroll/num lock onto underglow LEDs. What would it take to make that possible?

bortoz:
> It looks like currently it is not possible to overlay the caps/scroll/num lock onto underglow LEDs. What would it take to make that possible?

As I said before, that is a limitation of the underglow APIs and modifying those API is beyond the intentions of this PR.

ReFil:
I did the same thing here https://github.com/ReFil/zmk/blob/adv360-beta/app/src/rgb_underglow.c

docs/docs/features/led-indicators.md
```diff
@@ -0,0 +1,183 @@
...
+&pwm0 {
+	status = "okay";
+	ch0-pin = <33>;
+	/* ch0-inverted; */
+	ch1-pin = <35>;
+	/* ch1-inverted; */
+};
+```
+
+You have to declere a channel for each LED. A single PWM module has a fixed number of channels depending on your SoC, if you have many LEDs you may want to use `&pwm1` or ` &pwm2` as well.
```
kylemccreery, line 65:
```suggestion
You have to declare a channel for each LED. A single PWM module has a fixed number of channels depending on your SoC, if you have many LEDs you may want to use `&pwm1` or ` &pwm2` as well.
```

bortoz: Fixed, thanks!

alinelena:
i have tested this one and seems to work.

one thing i noticed scroll lock does not work but i suspect this is a more general issue on modern computers.
did not test kana and compose.

now with the exception of capslock... all the others are kind of academic... i wonder if will not be more useful to add a layer indicator feature in here... more in line with modern firmware

bortoz:
@alinelena thanks for testing!

> i wonder if will not be more useful to add a layer indicator feature in here... more in line with modern firmware

Can you elaborate more?

alinelena:
Practically I would like to have leds that indicate which layer is active this is additional to what you did already.

Imagine 4 layers more or less standard layering these days default raise lower and adjust. A simple approach is 1 indicator layer led per non default layer. This gives you 3. Of course if adj is dependent of raise and lower you get away with only 2.

Also these are similar with what you did for lock keys... With the only difference they are local to the keyboard.  We can elaborate in discord if interested.

ReFil:
Layer indication is tricky to do with only single colour LEDs, although you can indicate one additional layer I guess. Polarity works have done full colours per layer using extra props in the layer to signal to the RGB code what it should be doing

alinelena:
The only tricky bit is layers like adjust the rest you know the state and you just put the right led on or off.you just need a way to track the state of the layer. 

One can argue that caps lock is a layer indicator.

kylemccreery:
> The only tricky bit is layers like adjust the rest you know the state and you just put the right led on or off.you just need a way to track the state of the layer.
> 
> One can argue that caps lock is a layer indicator.

To clarify, caps lock is a host status and not a layer as it is handled on the keyboard. It is handled differently between mac and windows, as well. (IIRC windows is system wide where mac is board specific but still tied to the host)

They are interacted with pretty differently is the only reason I bring this up.

kylemccreery:
I am unable to get it to compile using PWM due to a linking issue (SEEMS to point at cmakelist but I am unsure on how to work that one out)

As for the GPIO LED, it updates on my board if I am plugged in via USB but if I disconnect, it no longer updates and will remain in the state that it was in when the board was unplugged and switched to bluetooth only

bortoz:
> I am unable to get it to compile using PWM due to a linking issue (SEEMS to point at cmakelist but I am unsure on how to work that one out)

Did you follow all the steps in the documentation?
Can you post the compilation output?

alinelena:
ok one issue I have got today... 
playing with a simple xiao ble today I noticed that using the user led(red/green/blue) on the board seems to have inversed logic... on when caps is off, off when caps is on

you can see an example in here 
https://github.com/alinelena/zmk/tree/mlego/app/boards/shields/mlego2x2

bortoz:
@alinelena Can you try this change?

```diff
 led_capslock: led_capslock {
-    gpios = <&gpio0 26 0>;
+    gpios = <&gpio0 26 GPIO_ACTIVE_HIGH>;
     label = "Caps lock LED";
 };
```

alinelena:
yes I can but makes no difference, that was the first version I got

bortoz:
Have you tried `GPIO_ACTIVE_LOW`?

alinelena:
> Have you tried `GPIO_ACTIVE_LOW`?

Yes... I think all is needed for these user led is inverted logic can you point to me where you set the state of the led according to the state of capslock? I can change by hand and test. In long term you want an option to account for this behaviour

alinelena:
> > Have you tried `GPIO_ACTIVE_LOW`?
> 
> Yes... I think all is needed for these user led is inverted logic can you point to me where you set the state of the led according to the state of capslock? I can change by hand and test. In long term you want an option to account for this behaviour

indeed I was right

```
 git diff  src/led_indicators_generic.c                                                                                                                                                   29/08/22 09:47:52
diff --git a/app/src/led_indicators_generic.c b/app/src/led_indicators_generic.c
index f8d58a17..128cea3a 100644
--- a/app/src/led_indicators_generic.c
+++ b/app/src/led_indicators_generic.c
@@ -46,7 +46,7 @@ zmk_leds_flags_t zmk_leds_get_current_flags() {
 
 zmk_leds_flags_t zmk_leds_get_flags(enum zmk_endpoint endpoint, uint8_t profile) {
     size_t index = profile_index(endpoint, profile);
-    return led_flags[index];
+    return ~led_flags[index];
 }
 
 static void raise_led_changed_event(struct k_work *_work) {
```

something like this fixes the behaviour, of course the initial state is still wrong. there are quite few places where these can be added.

bortoz:
The current logic is correct, this is a problem of your hardware or your devicetree configuration. If you want inverted logic you should use `GPIO_ACTIVE_LOW` instead of `GPIO_ACTIVE_HIGH`.

alinelena:
i am confused... this is standard seeduino xiao ble and normal zephyr zmk tree so I am pretty sure there is nothing wrong with them. 
have you tried the code on seeduino xiao and got different behaviour?

bortoz:
Either `GPIO_ACTIVE_HIGH` or `GPIO_ACTIVE_LOW` should give you the behaviour you want. If `GPIO_ACTIVE_HIGH` doesn't work for you, you should use `GPIO_ACTIVE_LOW` and vice versa. `GPIO_ACTIVE_HIGH` should work for most users, but the correct value depends on how you conntected your LEDs. The current code works fine with both configurations.

chrisandreae:
Adding an OUT interrupt endpoint to a keyboard, as done in this PR, hurts compatibility with some BIOS/UEFI environments.

While we were looking at the boot protocol support PR (#1140) over the weekend, we noticed that to get it to work in the UEFI environment particularly on Macs additionally required disabling the OUT interrupt endpoint that was added by this PR: https://github.com/moergo-sc/zmk/commit/d6b135521eff1b170634b66d2202de9e5d381b3a

To go into a bit more detail why this is the case: it's very uncommon for keyboards to use an interrupt endpoint for receiving LED reports from the host, because it LED updates happen so rarely that it's not worth consuming the USB bandwidth with an interrupt time slot. It's more common to have only an IN endpoint, and then the LED states are set by `Set_Report` class requests. Because it's uncommon, UEFI/BIOS/embedded environments often don't handle it correctly. 

[This Linux commit](https://github.com/torvalds/linux/commit/d7428bc26fc767942c38d74b80299bcd4f01e7cb) goes into more detail on why, but in short several environments fail to either initialize or use the device if its descriptor contains an OUT endpoint. The Mac UEFI environment (e.g. the Filevault unlock screen) is like this: a keyboard with an OUT endpoint fails to work at all, unless you flip the order of the in and out endpoints in the USB descriptor (which would be a spec violation, and Zephyr wouldn't make easy anyway).

The `Set_Report` class request is always sufficient for receiving LED state, so I'd suggest dropping the additional OUT interrupt endpoint from this PR to maximise compatibility.

bortoz:
@chrisandreae thanks for your feedback!

I have updated the code, can you test that everything is working correctly?

ReFil:
Been using this with RGB LEDs as the indicators for several months now on windows and Linux and it's been working excellently

chrisandreae:
Hi, thanks for the update, and sorry for the slow response! I've tried out the latest revision with my branch. The indicators are now working over USB, and the branch no longer seems to interfere with functioning in the BIOS/UEFI environment. However, the indicators don't seem to be working for me over Bluetooth. I haven't spent any time debugging this yet; I'll see if I can find some time over the next few days to look at it in more detail.

bortoz:
> just to reexplain myself. problem is the leds have the on state 0 and off state 1... kind of the opposite one expects. seen it in quite few boards. setting high or low at start does not help... cause the moment you sync the led with the state of caps for example does exactly the opposite.

This PR use the driver abstraction, if the indicator value is 1 then [len_on](https://docs.zephyrproject.org/latest/hardware/peripherals/led.html#c.led_on) is called, otherwise [led_off](https://docs.zephyrproject.org/latest/hardware/peripherals/led.html#c.led_off) is called. It is up to the LED driver to handle the call correctly.

ZMK does not know the type of LED you use and should not know it. The driver abstraction allows the code to be generic and to work with all types of LED drivers.

If your LED isn't working correctly, you should configure your LED driver properly. There is no need to change the code of this PR.

alinelena:
> > just to reexplain myself. problem is the leds have the on state 0 and off state 1... kind of the opposite one expects. seen it in quite few boards. setting high or low at start does not help... cause the moment you sync the led with the state of caps for example does exactly the opposite.
> 
> This PR use the driver abstraction, if the indicator value is 1 then [len_on](https://docs.zephyrproject.org/latest/hardware/peripherals/led.html#c.led_on) is called, otherwise [led_off](https://docs.zephyrproject.org/latest/hardware/peripherals/led.html#c.led_off) is called. It is up to the LED driver to handle the call correctly.
> 
> ZMK does not know the type of LED you use and should not know it. The driver abstraction allows the code to be generic and to work with all types of LED drivers.
> 
> If your LED isn't working correctly, you should configure your LED driver properly. There is no need to change the code of this PR.

there is little I can do to configure, this happens on xiao boards or blackpill boards, which are zephyr supported. I am just pointing out that the pr how it is does not account for these situations. of course you can choose to ignore it but that does not make the code right  or makes the boards stop existing.

bortoz:
> there is little I can do to configure, this happens on xiao boards or blackpill boards, which are zephyr supported. I am just pointing out that the pr how it is does not account for these situations. of course you can choose to ignore it but that does not make the code right or makes the boards stop existing.

I am not ignoring your situation, just that changing this code is the wrong way to solve your problem.

What you are looking for is [`GPIO_ACTIVE_LOW`](https://docs.zephyrproject.org/latest/hardware/peripherals/gpio.html#c.GPIO_ACTIVE_LOW), there is no need to duplicate that functionality in this PR. If `GPIO_ACTIVE_LOW` doesn't work, you should open an issue on Zephyr repo.

alinelena:
> What you are looking for is [`GPIO_ACTIVE_LOW`](https://docs.zephyrproject.org/latest/hardware/peripherals/gpio.html#c.GPIO_ACTIVE_LOW), there is no need to duplicate that functionality in this PR. If `GPIO_ACTIVE_LOW` doesn't work, you should open an issue on Zephyr repo.

is not my problem is the nature of encoding states in 0 and 1... it is arbitrary.. and assuming one is more valid than other is not ok... at least not in programming. funny bit is pwm that I am using now has an option to account for this. 
anyhow since I stopped using this I do not know why I bother.

bortoz:
> assuming one is more valid than other is not ok...

I am not assuming anything, I'm just calling [`led_on`](https://docs.zephyrproject.org/latest/hardware/peripherals/led.html#c.led_on) and [`led_off`](https://docs.zephyrproject.org/latest/hardware/peripherals/led.html#c.led_off). `led_on` should turn the LED on and `led_off` should turn the LED off. Calling `led_on` to turn the LED off is absolutely wrong.

> funny bit is pwm that I am using now has an option to account for this.

As I have said several times before, GPIO driver has the same functionality: [`GPIO_ACTIVE_LOW`](https://docs.zephyrproject.org/latest/hardware/peripherals/gpio.html#c.GPIO_ACTIVE_LOW).

yurikhan:
Sorry, I’m a novice to ZMK. How does this work with split keyboards?

If I add `CONFIG_ZMK_LED_INDICATORS=y` to my (peripheral) `<shield>_right.conf`, I get link errors about `zmk_event_zmk_led_changed`. If I only add that to the (central) `<shield>_left.conf`, I get a working indicator LED on the left side but nothing on the right side.

ReFil:
I think peripheral side splits are currently unsupported

bortoz:
> Sorry, I’m a novice to ZMK. How does this work with split keyboards?

I'm sorry, but I don't have the hardware to test split keyboards.  If anyone knows which changes are needed, let me know and I'll update the code and the docs.

ReFil:
I have made changes to allow for the indicator information to be sent to the peripheral side of a split. By spawning an additional ble service in central.c you can send led between sides in the same way behaviour information is. The branch is adv360_z3, I'd be happy to collaborate on adding these features in if you'd like 🙂

yurikhan:
> I have made changes to allow for the indicator information to be sent to the peripheral side of a split. By spawning an additional ble service in central.c you can send led between sides in the same way behaviour information is. The branch is adv360_z3, I'd be happy to collaborate on adding these features in if you'd like

I might try to lift some things out of that branch later if I find one LED insufficient. Thank you.

app/src/led_indicators_generic.c
```diff
@@ -0,0 +1,81 @@
...
+
+    CODE_UNREACHABLE;
+}
+
+zmk_leds_flags_t zmk_leds_get_current_flags() {
+    enum zmk_endpoint endpoint = zmk_endpoints_selected();
+    uint8_t profile = 0;
+
+#if IS_ENABLED(CONFIG_ZMK_BLE)
+    if (endpoint == ZMK_ENDPOINT_USB) {
```
snoyer, line 40:
```suggestion
    if (endpoint == ZMK_ENDPOINT_BLE) {
```

bortoz: Thanks!

snoyer:
Has anyone successfully tested this on mac with bluetooth? I get state change events (ie. `zmk_leds_process_report: Update LED indicators: endpoint=1, profile=4, flags=2`) correctly from an iMac running Big Sur, but not from a Macbook running Monterey.

ReFil:
Setting the USB vid and pid using CONFIG_USB_VID and CONFIG_USB_PID causes the board to lock up about half the time when plugging in the usb, it appears when i use this branch and disappears when i switch to zmk main. It's even worse on splits where plugging the main half in when the peripheral half is connected causes the main half to lock up all the time. I've been unable to capture logs as it seems to happen at the moment of enumeration. removing those lines from the defconfig fixes the problem

You mentioned you dont have a split board to test on, I can provide a pair of PCBs paired together for testing purposes at no cost at all as I'm very interested in seeing this feature working. I'll also provide all the board definitions and test code suitable for enumerating the indicator state over the RGB LEDs as a commit you can cherry pick into a test branch as you work. email me at team@polarityworks.com if you're interested

bortoz:
> ~Setting the USB vid and pid using CONFIG_USB_VID and CONFIG_USB_PID causes the board to lock up about half the time when plugging in the usb,~ it appears when i use this branch and disappears when i switch to zmk main. It's even worse on splits where plugging the main half in when the peripheral half is connected causes the main half to lock up all the time. I've been unable to capture logs as it seems to happen at the moment of enumeration. removing those lines from the defconfig fixes the problem

That's weird, which configuration and which branch are you using?

> You mentioned you dont have a split board to test on, I can provide a pair of PCBs paired together for testing purposes at no cost at all as I'm very interested in seeing this feature working. I'll also provide all the board definitions and test code suitable for enumerating the indicator state over the RGB LEDs as a commit you can cherry pick into a test branch as you work. email me at [team@polarityworks.com](mailto:team@polarityworks.com) if you're interested

Thank you, it would help me a lot to test split keyboards!

bortoz:
> Has anyone successfully tested this on mac with bluetooth? I get state change events (ie. `zmk_leds_process_report: Update LED indicators: endpoint=1, profile=4, flags=2`) correctly from an iMac running Big Sur, but not from a Macbook running Monterey.

@snoyer I might have found the issue, can you try again? Remember to unpair and pair back before testing.

snoyer:
> > Has anyone successfully tested this on mac with bluetooth? I get state change events (ie. `zmk_leds_process_report: Update LED indicators: endpoint=1, profile=4, flags=2`) correctly from an iMac running Big Sur, but not from a Macbook running Monterey.
> 
> @snoyer I might have found the issue, can you try again? Remember to unpair and pair back before testing.

Thanks @bortoz! I applied your change to a branch where I refactored out the host state tracking without the hardware stuff (https://github.com/snoyer/zmk/tree/led-indicators) and it works on both Macs now

ReFil:
> > ~Setting the USB vid and pid using CONFIG_USB_VID and CONFIG_USB_PID causes the board to lock up about half the time when plugging in the usb,~ it appears when i use this branch and disappears when i switch to zmk main. It's even worse on splits where plugging the main half in when the peripheral half is connected causes the main half to lock up all the time. I've been unable to capture logs as it seems to happen at the moment of enumeration. removing those lines from the defconfig fixes the problem
> 
> That's weird, which configuration and which branch are you using?
> 
> > You mentioned you dont have a split board to test on, I can provide a pair of PCBs paired together for testing purposes at no cost at all as I'm very interested in seeing this feature working. I'll also provide all the board definitions and test code suitable for enumerating the indicator state over the RGB LEDs as a commit you can cherry pick into a test branch as you work. email me at [team@polarityworks.com](mailto:team@polarityworks.com) if you're interested
> 
> Thank you, it would help me a lot to test split keyboards!
> 

Updated to the latest branch with the latest updates and it still happens. Didn't have any configuration changes set as I was using the RGB LEDs to display the state but even with my RGB features commented out the lock up still happened. I connected an SWD debugger and found that it was entering a fatal kernel error, but because the issue is quite sensitive to timing if I step through instructions it doesn't seem to manifest so I can't trace where the fatal error occurs

bortoz:
> Updated to the latest branch with the latest updates and it still happens. Didn't have any configuration changes set as I was using the RGB LEDs to display the state but even with my RGB features commented out the lock up still happened. I connected an SWD debugger and found that it was entering a fatal kernel error, but because the issue is quite sensitive to timing if I step through instructions it doesn't seem to manifest so I can't trace where the fatal error occurs

Does it happen only on split keyboards?

ReFil:
Certainly seems to be that way, I've been unable to replicate the lockup on a non split

tyalie:
This PR looks awesome, but I would like to propose a change in how it is configured. 

Right now the indicator types are not only responsible for on/off behaviour, but they also handle properties that make assumptions about the attached LED, ... I noticed this mainly with the brightness being configured through a config flag.

In order to be more flexible, I'd suggest separating these two features from another and introduce another level of abstraction. In the end e.g. zmk,caps-lock would require a device as input that exposes an on and off method. At this point the concept of an LED, brightness & co would be lost.
To make this change we would also need to introduce a new device type / interface, but in the end we would have far greater flexibility.

One example application would be having multiple LEDs with differing brightness (e.g. you use different LED colors and type). 
In my case I have WS2812 leds on my keyboard, and with this change I could write a driver that handles colors and such and integrate it neetly into your work.

*note*: I'd be very happy to help you here.

tyalie:
This PR looks awesome, but I would like to propose a change in how it is configured. 

Right now the indicator types are not only responsible for on/off behaviour, but they also handle properties that make assumptions about the attached LED, ... I noticed this mainly with the brightness being configured through a config flag.

In order to be more flexible, I'd suggest separating these two features from one another and introduce another level of abstraction. In the end e.g. zmk,caps-lock would require a device as input that exposes an on and off method. At this point the concept of an LED, brightness & co would be lost.
To make this change we would also need to introduce a new device type / interface, but in the end we would have far greater flexibility.

One example application would be having multiple LEDs with differing brightness (e.g. you use different LED colors and type). 
In my case I have WS2812 leds on my keyboard, and with this change I could write a driver that handles colors and such and integrate it neetly into your work.

*note*: I'd be very happy to help you here.

joelspadin:
I think the way I would design this is:

1. Add separate code to read the HID LED states for USB and BLE.
2. Add a generic HID function which retrieves the HID LED state for the active output.
3. Add an event that fires whenever the HID LED state for the active output changes.
4. For each type of indicator, write a driver (or non-driver code) that listens for the event and updates itself accordingly. Examples might include:
  1. A driver that maps an HID LED state to anything implementing Zephyr's LED driver API.
  2. A driver that maps an HID LED state to anything implementing Zephyr's RGB strip API.
  3. A display widget that displays HID LED states.
  4. An RGB animation driver (e.g. #1046) that sets specific pixel colors according to the HID LED state.

I think this is effectively the same thing as you suggested, but with one less layer of abstraction. I'm not sure I see much of a need for inserting a special on/off driver in the middle.

From skimming the code here, it looks like the core code here is very similar to what I would suggest, but I rather than using chosen nodes, I think I would want to be able to configure things more like (pseudo-devicetree, not necessarily exactly how things would work):

```
/ {
    leds {
        compatible = "pwm-leds";
        ...
        led_capslock: led_capslock { ... };
    };
   
    capslock {
        compatible = "zmk,indicator-led";
        indicator = <HID_LED_CAPSLOCK>;
        led = &led_capslock;
        // Brightness as a property of each LED lets me set different brightnesses for different LEDs.
        // Could still have a kconfig option to set a global default though.
        brightness = <80>;
    };

    scroll_lock {
        compatible = "zmk,indicator-led-strip";
        indicator = <HID_LED_SCROLLLOCK>;
        leds = <0 1>;
        on-color = <255 0 0>;
        // off-color = <0 0 0>;
    };

    // Maybe I have some GPIO I want to turn on when a state is set?
    compose {
        compatible = "zmk,indicator-gpio";
        indicator = <HID_LED_COMPOSE>;
        led-gpios = <&gpio0 42>;
    };

    // Or I have an RGB matrix and we've written an animation system for it.
    chosen { 
        zmk,animation = &animation; 
    };
    animation: animation {
        compatible = "zmk,animation-composite";
        animations = <&base_animation &indicator_animation>;
    };
    base_animation: base_animation { ... };
    indicator_animation: indicator_animation {
        indicators = <HID_LED_NUMLOCK HID_LED_KANA>;
        leds = <0 1>;
        colors = <255 0 0>, <0 0 255>;
    };
};

&spi1 {
    ...
    led_strip: ws2812@0 { ... };
};
```

Note that there are a *lot* of different indicators in the LED page, so having a separate chosen node for each one sounds like it would be a lot harder to maintain that just having you specify in the driver which LED usage ID you want to use.

snoyer:
> I think the way I would design this is:
> 
>     1. Add separate code to read the HID LED states for USB and BLE.
>     2. Add a generic HID function which retrieves the HID LED state for the active output.
>     3. Add an event that fires whenever the HID LED state for the active output changes.

This is part should really be done first and independently of the custom hardware considerations, as already pointed out since the first few comments on this PR.

@bortoz, could you make a separate PR implementing only the HID state tracking part so it can get past @petejohanson ?

I would suggest some minor refactoring such as:
- using more consistent `zmk_led_indicators` prefixes
- change the `NUM_...PROFILES` constants to `ZMK_...PROFILE_COUNT` to match and reuse the existing `ZMK_BLE_PROFILE_COUNT`
- move the profile indexing stuff to separate files as it can be useful for other features/behaviors

I have already implemented these changes [in a branch on my fork](https://github.com/snoyer/zmk/tree/led-indicators) if that can save you some work.

yurikhan:
I was thinking about a different LED configuration mechanism.

For the key matrix, we have a way to define a matrix (with row pins and column pins or direct connection pins), a way to combine central and peripheral matrices into one, and a separate way to assign behaviors to each key of the combined matrix.

One could imagine a similar mechanism with LEDs. E.g. a Seeed XIAO-based split keyboard could declare three on-board LEDs and one dedicated on-shield LED on each side, combine them into a single 2×4 matrix, and assign behaviors. Possible behaviors could be:

* HID indicator (parametrized by the HID usage code)
* Battery charge < (%). E.g. light the red LED when discharged below 10%.
* Battery charge ≥ (%). E.g. 4 LEDs in a line with threshold set to 20%, 40%, 60%, 80% would implement a bar graph indicator.
* Layer active (ID). (Alternatively, LED behaviors could be assigned per layer.)
* Bluetooth profile active (ID) or paired (ID).
* Output active (Bluetooth or USB).
* …

(no, I’m not volunteering, sorry)

tyalie:
@snoyer I could do that. But just to not mix stuff up: the PR should essentially only contain a mechanism to retrieve the HID. So no output, LEDs or anything, just that ZMK is aware of the HID LED states?

snoyer:
> @snoyer I could do that. But just to not mix stuff up: the PR should essentially only contain a mechanism to retrieve the HID. So no output, LEDs or anything, just that ZMK is aware of the HID LED states?

@bortoz, correct. Basically [this](https://github.com/zmkfirmware/zmk/compare/main...snoyer:zmk:led-indicators): setting up the proper hid descriptors, reading the reports, firing the `zmk_led_indicators_changed` events, providing `zmk_led_indicators_flags` and `zmk_led_indicators_get_current_flags` methods.

This could then be used by any of the many possible hardware output method (leds, rgb, displays, underglow, ...), behavior, or any functionality that need to know about or react to capslock/numlock/... state.

joelspadin:
> 3. Add an event that fires whenever the HID LED state for the active output changes.

Just in case it wasn't obvious, I would also expect this event to fire any time the active output or active BLE profile changes, and this results in a change to the HID LED indicator state.

joelspadin:
Also, agreed on splitting this up into to PRs. I think that would help to get this through review, as we'd be able to independently review getting the HID LED state (which seems pretty straightforward) from making the keyboard indicate that state (which could be implemented many different ways).

bortoz:
@tyalie @snoyer sorry for the late response, here are my thoughts:

Many of the feature you mentioned are already implemented:
- you can use `zmk_leds_get_current_flags()` to retrive the current LED indicators;
- if you need the indicators for another profile you can use  `zmk_leds_get_flags(...)`;
- there is the `zmk_led_changed` event which is fired every time an indicator changes (and also when the profile changes!).

On top of this, I implemented a generic interface using Zephyr LED drivers. This interface is what most users will use, but if you don't want to use this interface, you don't have to! You can just use this PR and define whatever you want using the functions I mentioned before. It's also explained in the documentation (["other" tab](https://deploy-preview-999--zmk.netlify.app/docs/features/led-indicators#adding-led-indicators-to-a-board))!

This PR was opened more than a year ago and many people have already test it. It makes no sense to me to remove this feature, open a new PR and wait for another few years to be tested and reviewed again.

Of course we can instead develop a much better implementation that meets everyone's needs, but that would take a lot of time and meanwhile people who actually need this simple behavior (which is already implemented) would have to wait with nothing.

I'm open to make changes, but keep in mind that I've already accounted for a lot of different things when designing this PR. For example we can't use `ZMK_BLE_PROFILE_COUNT` directly since it is not defined when bluetooth is disabled.

joelspadin:
I suspect that part of the reason this is taking so long to get reviewed is because of the size of the change. Creating the API should be pretty straightforward, but there are many different ways to design the code for controlling LEDs, so reviewing that part is more difficult.

On that note, please see [my previous comment](https://github.com/zmkfirmware/zmk/pull/999#issuecomment-1320792760) regarding how things are configured. I think it would be a lot more flexible to create a node that maps each indicator to an LED rather than to have a separate chosen node for each indicator. The benefits to this I see are:

- It allows you to connect multiple LEDs to the same indicator.
- It allows you to configure indicators individually, for example if you have two LEDs of different colors, and one appears brighter than the other, you might want to change the brightness of only one LED.
- It could simplify the code, because instead of having to define an array of all the indicators in led_indicators.c and then loop over them at init, you just let Zephyr's device driver instancing handle it.
- The configuration stays consistent if we add new types of display later. The existing chosen nodes are tied to the GPIO/PWM LED implementation.

Here's a comparison:

```c
// Common to both implementations
/ {
    leds {
        compatible = "pwm-leds";
        led_capslock1: led_capslock1 { ... };
        led_capslock2: led_capslock2 { ... };
        led_compose: led_compose { ... };
    };
};
&spi1 {
    ...
    led_strip: ws2812@0 { ... };
};

// With chosen nodes
/ {
    chosen {
        // Problem 1: what if I want both &led_capslock1 and &led_capslock2
        // to light up with capslock? 
        zmk,led-capslock = &led_capslock1;

        // Problem 2: what if my compose key LED is a different color and needs
        // to be a different brightness to match? A single kconfig option for the
        // brightness won't work, as it would affect capslock too.
        zmk,led-compose = &led_compose;

        // Problem 3: zmk,led-numlock only handles GPIO LEDs. What if I want
        // LEDs 2 and 3 of &led_strip to turn red when numlock is on? That behavior
        // can't be described with just a chosen node that points to &led_strip.
    };
};

// With device instances
// Need a header to define the indicator values.
#include <zmk/dt-bindings/indicators.h>

/ {
    capslock {
        compatible = "zmk,indicator-led";
        indicator = <INDICATOR_CAPSLOCK>;
        // Solution 1: the LED driver could take a list of LEDs, or you could just 
        // define multiple instances of it.
        leds = <&led_capslock1, &led_capslock2>;
    };
    compose {
        compatible = "zmk,indicator-led";
        indicator = <INDICATOR_COMPOSE>;
        leds = <&led_compose>;
        // Solution 2: each instance of the driver can have its own brightness.
        // (A kconfig option could still be provided to set a default brightness.)
        brightness = <50>;
    };
    // Solution 3: configuring an LED strip is roughly the same.
    numlock {
        compatible = "zmk,indicator-led-strip";
        indicator = <INDICATOR_NUMLOCK>;
        device = <&led_strip>;
        leds = <2 3>;
        on-color = <255 0 0>;
        off-color = <0 0 0>;
    };
};
```

tyalie:
I need to agree here with @joelspadin 

Splitting this PR up will presumably ease the way to get the functionality into upstream, as LED indicators per se aren't that much of a controversial topic.

Additionally the strength of ZMK is that it is configurable without needing to write C code. Forcing users here to put code into their custom configs in order to access functionality that could easily be wrapped up into a device driver doesn't feel right to me. Especially when those features entail common use cases like they have been mentioned before in this thread.

snoyer:
@bortoz,

> @tyalie @snoyer sorry for the late response, here are my thoughts:
> 
> Many of the feature you mentioned are already implemented:
> 
>     * you can use `zmk_leds_get_current_flags()` to retrive the current LED indicators;
> 
>     * if you need the indicators for another profile you can use  `zmk_leds_get_flags(...)`;
> 
>     * there is the `zmk_led_changed` event which is fired every time an indicator changes (and also when the profile changes!).

I know, I've looked at your code, I've used it, I've tested it, I've built on top of it, and I'm trying to help get it adopted.
My suggestion was to use more explicit `zmk_led_INDICATORS` prefixes to make it clearer the functions relate to the HID led indicators rather than any random leds that could be on the keyboard. Have a look through [my branch](https://github.com/zmkfirmware/zmk/compare/main...snoyer:zmk:led-indicators)

> On top of this, I implemented a generic interface using Zephyr LED drivers. This interface is what most users will use, but if you don't want to use this interface, you don't have to! You can just use this PR and define whatever you want using the functions I mentioned before. It's also explained in the documentation (["other" tab](https://deploy-preview-999--zmk.netlify.app/docs/features/led-indicators#adding-led-indicators-to-a-board))!

"**On top of this**", yes, exactly, that is the point, there is a layer of ground work and then more stuff **on top**. That line is important and useful. That's the line where you split the PR.

> This PR was opened more than a year ago and many people have already test it. It makes no sense to me to remove this feature, open a new PR and wait for another few years to be tested and reviewed again.

And you were asked more than a year ago, including by the main contributor of the project, to please split your PR to facilitate its review and adoption.

> Of course we can instead develop a much better implementation that meets everyone's needs, but that would take a lot of time and meanwhile people who actually need this simple behavior (which is already implemented) would have to wait with nothing.

I don't think this is the point people are making.

> I'm open to make changes, 

Then prove it ;)

> but keep in mind that I've already accounted for a lot of different things when designing this PR.

Yes, you did a lot of useful work that many people, including me, want to see merged, but you can't expect the maintainers to just "take it or leave it" (I mean, you could, but that's a gamble). People here are not ganging up on you to make you destroy your efforts, quite the opposite actually.

> For example we can't use `ZMK_BLE_PROFILE_COUNT` directly since it is not defined when bluetooth is disabled.

Then why not [fix that](https://github.com/snoyer/zmk/blob/e224e6e22a028788b1d650096f3356b154f4b09b/app/include/zmk/ble.h#L22) instead of creating a duplicated variable with a different naming convention?

petejohanson:
Yes, this whole thing would make more progress if split into two PRs. We could have a focused review on the core HID bits, state, and events, and then continue iterating on the display bits layered on top.

I'm tentatively liking the proposal from @joelspadin but definitely want to consider it further before suggesting you refactor to that approach.

In the meantime, a PR for the core bits I would love to have, so I could review and we could merge that sooner.

bortoz:
As requested, I have removed the "display" part of this PR and left only the "core" part.
I also added support for LED indicators on the peripheral side.

joelspadin:
I wonder, do people think it might be confusing to have things named "LED indicators" throughout the code which don't directly control LEDs? I know the HID spec does call them LED indicators, but maybe we should just call them "indicators" since the indicators don't have to be LEDs?

snoyer:
> I wonder, do people think it might be confusing to have things named "LED indicators" throughout the code which don't directly control LEDs? I know the HID spec does call them LED indicators, but maybe we should just call them "indicators" since the indicators don't have to be LEDs?

Just calling them "indicators" may be pushing the problem to the other extreme as some other features (layers, endpoints, ...) may need to be "indicated" by unrelated means (displays, ...)?

Maybe _"HID indicators"_ would be more descriptive? _"Lock indicators"_ could be more intuitive/user-friendly as the main targets are caps/num/scroll lock, but not technically correct as the spec covers more than these 3.

Personally I'd get behind any name that would put this PR back on track because ZMK needs this feature. If the indicators have to be called _"blinkydoodles"_ to be merged in main I'm 100% on board with that too :)

joelspadin:
I like "HID indicators". I've been busy with other things recently, but I'll try to make some time to review this soon.

joelspadin:
"Lock indicators" has the problem that the LED page has many more indicators than just lock states, for example Compose and Kana (and a ton of other things that probably nobody uses).

snoyer:
> I like "HID indicators". I've been busy with other things recently, but I'll try to make some time to review this soon.

Great thing to hear!

> "Lock indicators" has the problem that the LED page has many more indicators than just lock states, for example Compose and Kana (and a ton of other things that probably nobody uses).

Yeah it does go on for [almost 60 lines after `HID_USAGE_LED_NUM_LOCK` in `hid_usage.h`](https://github.com/zmkfirmware/zmk/blob/main/app/include/dt-bindings/zmk/hid_usage.h#L512-L568). Note that the `// OOC` comments come from this part of the spec: "An LED or indicator is implemented as an On/Off Control", so _"OOC indicators"_ is another naming option if you need/want to be extra specific (_"HID indicators"_ is probably a better balance of correctness and usability though).

purdeaandrei:
There's a small conflict in central.c now due to adjacent line having been edited, and zephyr/ prefixes will need to be added to a bunch of includes to make it all compile with the recent zephyr 3.2 update.

```diff
diff --git a/app/src/events/led_indicator_changed.c b/app/src/events/led_indicator_changed.c
index 47cc5f6c..4569bfa2 100644
--- a/app/src/events/led_indicator_changed.c
+++ b/app/src/events/led_indicator_changed.c
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  */
 
-#include <kernel.h>
+#include <zephyr/kernel.h>
 #include <zmk/events/led_indicator_changed.h>
 
 ZMK_EVENT_IMPL(zmk_led_changed);
diff --git a/app/src/led_indicators.c b/app/src/led_indicators.c
index 5acab0c8..c5e08d83 100644
--- a/app/src/led_indicators.c
+++ b/app/src/led_indicators.c
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: MIT
  */
 
-#include <kernel.h>
-#include <logging/log.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
 
 #include <zmk/ble.h>
 #include <zmk/endpoints.h>

```

app/src/split/bluetooth/central.c
```diff
@@ -265,9 +267,14 @@ static uint8_t split_central_chrc_discovery_func(struct bt_conn *conn,
                             BT_UUID_DECLARE_128(ZMK_SPLIT_BT_CHAR_RUN_BEHAVIOR_UUID))) {
         LOG_DBG("Found run behavior handle");
         slot->run_behavior_handle = bt_gatt_attr_value_handle(attr);
+    } else if (!bt_uuid_cmp(((struct bt_gatt_chrc *)attr->user_data)->uuid,
+                            BT_UUID_DECLARE_128(ZMK_SPLIT_BT_UPDATE_LEDS_UUID))) {
+        LOG_DBG("Found update LEDs handle");
+        slot->update_leds_handle = bt_gatt_attr_value_handle(attr);
```
joelspadin, line 273: `release_peripheral_slot()` clears `run_behavior_handle` to 0. Does that need to happen for `update_leds_handle` as well?

joelspadin (changes requested):

app/include/zmk/led_indicators.h
```diff
@@ -0,0 +1,18 @@
...
+#include <zmk/endpoints_types.h>
+#include <zmk/hid.h>
+#include <zmk/led_indicators_types.h>
+
+zmk_leds_flags_t zmk_leds_get_current_flags();
+zmk_leds_flags_t zmk_leds_get_flags(enum zmk_endpoint endpoint, uint8_t profile);
+void zmk_leds_update_flags(zmk_leds_flags_t leds, enum zmk_endpoint endpoint, uint8_t profile);
+
+void zmk_leds_process_report(struct zmk_hid_led_report_body *report, enum zmk_endpoint endpoint,
+                             uint8_t profile);
```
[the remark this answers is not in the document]
snoyer: @joelspadin, should this be looked at as something that ZMK could/should provide independently of this PR?

The general idea is that some behaviors/functionalities may want to store data separately for each connected device: USB, BLE(0), BLE(1), ... so having a function to index the endpoints+profiles would be convenient.
For example I ran into this when looking into saving (and automatically restoring) base layers per device.

Basically we'd want something like:
- `zmk_endpoint_index_t zmk_endpoint_index(zmk_endpoint_select)`
- `zmk_endpoint_select zmk_current_endpoint()`
- maybe `zmk_endpoint_index_t zmk_current_endpoint_index()` to shortcut `zmk_endpoint_index(zmk_current_endpoint())`
- `ZMK_ENDPOINT_COUNT` (`ZMK_USB_PROFILE_COUNT + ZMK_BLE_PROFILE_COUNT` + possibly future endpoint count)

I don't know enough to tell what the right way to do it is, just pointing out that it's worth considering the wider pattern/use-case.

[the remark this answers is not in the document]
joelspadin: @petejohanson Any thoughts on how to best design this?

app/src/led_indicators.c
```diff
@@ -0,0 +1,87 @@
...
+void zmk_leds_process_report(struct zmk_hid_led_report_body *report, enum zmk_endpoint endpoint,
+                             uint8_t profile) {
+    zmk_leds_flags_t leds = report->leds;
+    zmk_leds_update_flags(leds, endpoint, profile);
+
+    LOG_DBG("Update LED indicators: endpoint=%d, profile=%d, flags=%x", endpoint, profile, leds);
+}
+
+static int endpoint_listener(const zmk_event_t *eh) {
+    raise_led_changed_event(NULL);
```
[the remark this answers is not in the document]
bortoz: This function is always called from the main thread while `zmk_leds_update_flags` could be called from other threads.

app/src/led_indicators.c
```diff
@@ -0,0 +1,87 @@
...
+#endif
+}
+
+static K_WORK_DEFINE(led_changed_work, raise_led_changed_event);
+
+void zmk_leds_update_flags(zmk_leds_flags_t leds, enum zmk_endpoint endpoint, uint8_t profile) {
+    size_t index = profile_index(endpoint, profile);
+    led_flags[index] = leds;
+
+    k_work_submit(&led_changed_work);
```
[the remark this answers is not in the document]
bortoz: I can add a mutex or an atomic variable but I don't see why that is necessary. In this function we never read the value, we just write it and writing operations are already atomic.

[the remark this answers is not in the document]
joelspadin: I was more concerned about the other functions that do read the value.

Also, do you think it matters whether two changes here in quick succession result in a single notification of the final state or separate notifications for each change? If we want the latter, we really need to use a message queue (or similar) so the main thread can process each individual change.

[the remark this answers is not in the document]
joelspadin: Okay. This is fine then, though I would still be concerned about someone later adding code which reads the variable and not realizing that it gets written on another thread. Adding a comment or making it an atomic would at least be a clue that more care needs to be taken because multiple threads are involved.

app/include/zmk/led_indicators.h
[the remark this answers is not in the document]
petejohanson: So, I generally really like the idea of a tagged union for this, in terms of the data type for identifying a given endpoint (active or otherwise). 

As I think on it, I think there's maybe use in having clear concept of "endpoint transport", e.g. USB, BLE, and then figuring out a way to have "per transport selections/preferences" that would allow us to capture BLE profile index, etc.

Doing it this way would allow us to have both a "preferred endpoint" and "active endpoint" nicely modeled, and also open up the way to have each transport offer an away to enumerate possible endpoints available on that transport.

And I do agree, this is something that should be really nailed down first in a separate PR, for focus and clarity while we independently review and work out issues w/ LED indicators.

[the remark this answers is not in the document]
joelspadin: For the terminology, what do you think of "endpoint transport" (USB/BLE) and "endpoint instance" (USB, BLE profile 0, etc.)?

I could put together a PR for that if you'd like.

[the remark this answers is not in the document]
petejohanson: Seems good to me.

app/src/led_indicators.c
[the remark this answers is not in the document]
bortoz: I added a comment.

app/src/split/bluetooth/central.c
```diff
@@ -584,6 +591,31 @@ int zmk_split_bt_invoke_behavior(uint8_t source, struct zmk_behavior_binding *bi
...
+            LOG_ERR("Failed to write the leds characteristic (err %d)", err);
+        }
+    }
+}
+
+static K_WORK_DEFINE(split_central_update_leds, split_central_update_leds_callback);
+
+int zmk_split_bt_update_leds(zmk_leds_flags_t leds) {
+    led_flags = leds;
+    return k_work_submit_to_queue(&split_central_split_run_q, &split_central_update_leds);
```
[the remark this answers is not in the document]
bortoz: I added a variable at the start of `split_central_update_leds_callback` to avoid reading the variable at each iteration of the loop. Correct me if I'm wrong, but as in the other comment, I think it shouldn't be necessary to use a lock or an atomic variable.

app/src/usb_hid.c
```diff
@@ -23,8 +24,41 @@ static K_SEM_DEFINE(hid_sem, 1, 1);
...
+    if ((setup->wValue & HID_GET_REPORT_TYPE_MASK) != HID_REPORT_TYPE_OUTPUT) {
+        LOG_ERR("Unsupported report type %d requested",
+                (setup->wValue & HID_GET_REPORT_TYPE_MASK) >> 8);
+        return -ENOTSUP;
+    }
+
+    switch (setup->wValue & HID_GET_REPORT_ID_MASK) {
+    case HID_REPORT_ID_LEDS:
+        if (*len != sizeof(struct zmk_hid_led_report)) {
+            LOG_ERR("LED set report is malformed: length=%d", *len);
```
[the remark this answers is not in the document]
bortoz: Fixed

app/include/zmk/led_indicators.h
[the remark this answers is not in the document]
joelspadin: @bortoz Try this and let me know if it works for you: https://github.com/zmkfirmware/zmk/pull/1803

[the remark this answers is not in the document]
bortoz: Thank you! I've rebased on your branch.

app/include/zmk/hid_indicators_types.h
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#pragma once
+
+typedef uint8_t zmk_hid_indicators;
```
petejohanson, line 9: We usually follow the convention of using a `_t` suffix for types like these:

```suggestion
typedef uint8_t zmk_hid_indicators_t;
```

app/src/hid_indicators.c
```diff
@@ -0,0 +1,69 @@
...
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+static zmk_hid_indicators hid_indicators[ZMK_ENDPOINT_COUNT];
+
+zmk_hid_indicators zmk_hid_indicators_get_current_profile(void) {
+    return zmk_hid_indicators_get_profile(zmk_endpoints_selected());
+}
+
+zmk_hid_indicators zmk_hid_indicators_get_profile(struct zmk_endpoint_instance endpoint) {
+    int profile = zmk_endpoint_instance_to_index(endpoint);
```
petejohanson, line 26:
```suggestion
    const int profile = zmk_endpoint_instance_to_index(endpoint);
```

app/src/hid_indicators.c
```diff
@@ -0,0 +1,69 @@
...
+    return hid_indicators[profile];
+}
+
+static void raise_led_changed_event(struct k_work *_work) {
+    zmk_hid_indicators indicators = zmk_hid_indicators_get_current_profile();
```
petejohanson, line 31:
```suggestion
    const zmk_hid_indicators indicators = zmk_hid_indicators_get_current_profile();
```

app/src/hid_indicators.c
```diff
@@ -0,0 +1,69 @@
...
+    // or writes only one entry at a time, so it is safe to do these operations without a lock.
+    hid_indicators[profile] = indicators;
+
+    k_work_submit(&led_changed_work);
+}
+
+void zmk_hid_indicators_process_report(struct zmk_hid_led_report_body *report,
+                                       struct zmk_endpoint_instance endpoint) {
+    uint8_t indicators = report->leds;
+    zmk_hid_indicators_set_profile(indicators, endpoint);
```
petejohanson, line 58:
```suggestion
    zmk_hid_indicators_t indicators = (zmk_hid_indicators_t)report->leds;
    zmk_hid_indicators_set_profile(indicators, endpoint);
```

app/include/zmk/hid.h
```diff
@@ -17,11 +17,15 @@
 
 #define COLLECTION_REPORT 0x03
 
+#define HID_REPORT_ID_KEYBOARD 0x01
+#define HID_REPORT_ID_LEDS 0x01
+#define HID_REPORT_ID_CONSUMER 0x02
```
petejohanson, line 22: I missed you had added these here. I had rolled a similar tweak into #1991 with different naming conventions. Let me pull that out into a small refactor PR (where it should be anyways, and we can then build on it for this and that PR as well.

petejohanson (changes requested):
@bortoz Thank you for your patience on this! One or two minor comments on the coding conventions, but otherwise this looks great. I'll get the HID tweaks in a PR right away to get that in, while you work on the couple other comments here.

petejohanson:
@bortoz See #1993 for the small `hid.h` refactor.

app/src/split/bluetooth/central.c
```diff
@@ -685,6 +693,32 @@ int zmk_split_bt_invoke_behavior(uint8_t source, struct zmk_behavior_binding *bi
...
+static void split_central_update_indicators_callback(struct k_work *work) {
+    zmk_hid_indicators indicators = hid_indicators;
+    for (int i = 0; i < ZMK_SPLIT_BLE_PERIPHERAL_COUNT; i++) {
+        if (peripherals[i].state != PERIPHERAL_SLOT_STATE_CONNECTED) {
+            continue;
+        }
+
+        int err = bt_gatt_write_without_response(peripherals[i].conn,
+                                                 peripherals[i].update_hid_indicators, &indicators,
+                                                 sizeof(indicators), true);
```
petejohanson, line 707: See https://github.com/chrisandreae/zmk/commit/bc824de85811a62018c33fe3828a1f0b51453a3f from the MoErgo folks for a small safety check that would be good here, if this gets called mid-attribute discovery.

app/CMakeLists.txt
```diff
@@ -37,6 +37,8 @@ target_sources(app PRIVATE src/behaviors/behavior_reset.c)
 target_sources_ifdef(CONFIG_ZMK_EXT_POWER app PRIVATE src/behaviors/behavior_ext_power.c)
 if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
   target_sources(app PRIVATE src/hid.c)
+  target_sources_ifdef(CONFIG_ZMK_USB app PRIVATE src/usb_hid.c)
```
bryanforbes, line 40: This is no longer needed after #2006 was merged since `CONFIG_ZMK_USB` now depends on the `if` condition

petejohanson:
@bortoz Thank you so much for the work on this! I've merged #2041 just now that includes all your core work with the minor changes/additions as noted there. I am truly sad this has taken so long, and I appreciate your persistence in sticking with this. I'm hoping we can build out various status/display bits on top of this now that the core has been merged.

:heart:

PR #1000, opened 2021-10-24
feat(shield): add Horizon shield

skarrmann:
Hello! I would like to add a ZMK shield for my open source keyboard, [Horizon](https://github.com/skarrmann/horizon).
Thank you very much, and please let me know if you have any questions.

## Board/Shield Check-list
    * Note: Not applicable
    * Note: Not split
    * Note: `.conf` file is intentionally empty - this board only supports keys


Nicell:
@skarrmann Sorry about the late response here. Currently, we're accepting boards that are commercially available, have a completed GB, or are otherwise widely available to lighten the review load for contributors. This shield could use a ZMK config repo instead with the custom definition inside so that users who build themselves can fork and adjust the keymap etc. I would recommend that route for now!

skarrmann:
Thank you for clarifying @Nicell. I will host this shield in a separate ZMK config repo for the time being. If/when this board has a commercial release, I will submit a PR at that time. Thank you for your time!

PR #1007, opened 2021-10-31
Bugfix: decide_hold_tap switch statement needs breaks after cases

jmding8:
Not sure why this bug didn't affect anything in the past, perhaps the ordering of the cases is special and lucky?


petejohanson (approved):
Good catch! Thanks!

PR #1010, opened 2021-11-02
feat(underglow): Add RGB auto off timeout on idle and on usb disconnect

ReFil:
This is based off the excellent work by @bortoz but modified to work with the RGB underglow feature

if the config settings are enabled the lights will shut off when the board goes into idle and resumes when you push a key and wake the board.

Tested and working on my personal BT75 board, it's just a minor change


app/Kconfig
```diff
@@ -316,6 +316,14 @@ config ZMK_RGB_UNDERGLOW_ON_START
 	bool "RGB underglow starts on by default"
 	default y
 
+config ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE
+	bool "Turn off RGB underglow when keyboard goes into idle state"
+	default y
+
+config ZMK_RGB_UNDERGLOW_AUTO_OFF_USB
+	bool "Turn off RGB underglow when USB is disconnected"
+	default n
```
petejohanson, line 325: This should skip the redundant `default n` (see https://docs.zephyrproject.org/2.5.0/guides/kconfig/tips.html#redundant-defaults) and include a depends on USB, e.g.:

```suggestion
config ZMK_RGB_UNDERGLOW_AUTO_OFF_USB
	bool "Turn off RGB underglow when USB is disconnected"
	depends on USB
```

app/src/rgb_underglow.c
```diff
@@ -434,4 +440,67 @@ int zmk_rgb_underglow_change_spd(int direction) {
...
 }
 
+#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE)
+static bool auto_off_idle_prev_state = false;
+#endif
+
+#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_USB)
+static bool auto_off_usb_prev_state = false;
+#endif
+
```
petejohanson, line 450: I'm not sure we actually need dedicated *different* variables for each condition, instead of one general `auto_off_prev_state` boolean.

app/src/rgb_underglow.c
```diff
@@ -434,4 +440,67 @@ int zmk_rgb_underglow_change_spd(int direction) {
...
+        }
+        if (new_state) {
+            state.on = auto_off_idle_prev_state;
+            auto_off_idle_prev_state = false;
+            return zmk_rgb_underglow_on();
+        } else {
+            state.on = false;
+            auto_off_idle_prev_state = true;
+            return zmk_rgb_underglow_off();
+        }
```
petejohanson, line 469: If we have one common storage, we can move this logic into some new function, e.g.:

```
static void zmk_rgb_underglow_auto_state(bool new_state) {
    if (new_state) {
            state.on = auto_off_prev_state;
            auto_off_prev_state = false;
            return zmk_rgb_underglow_on();
        } else {
            state.on = false;
            auto_off_prev_state = true;
            return zmk_rgb_underglow_off();
        }
}
```

And then the implementation of these two cases in the event handler can be drastically simplified.

petejohanson (changes requested):
Thanks for the PR! A few items from my first review of this.

app/Kconfig
bortoz: We cannot use a single variable. Consider the following situation when both `AUTO_OFF_IDLE` and `AUTO_OFF_USB` are enabled.

* initially `state=on`
* unplug USB: `state=off`, `prev_state=on`
* idle state: nothing happens
* back from idle state
  expected behaviour: nothing happens
  actual behaviour: RGB turns on since `prev_state=on`

app/src/rgb_underglow.c
bortoz: We cannot use a single variable. Consider the following situation when both `AUTO_OFF_IDLE` and `AUTO_OFF_USB` are enabled.

* initially `state=on`
* unplug USB: `state=off`, `prev_state=on`
* idle state: nothing happens
* back from idle state
  expected behaviour: nothing happens
  actual behaviour: RGB turns on since `prev_state=on`

ReFil: Yeah I think separating the two is better because of how the two states can conflict and interact with each other. I'm happy to implement it and test it but I'm pretty sure @bortoz is correct in that unexpected behaviour can happen

petejohanson: You should still be able to avoid the duplication then, by making the function take a pointer to the state variable you should use for comparison and update. That helps bring the code size down, while avoiding the edge cases you note. Thanks!

berfarah:
This is super cool and something I'll definitely set up for myself.

I was wondering: Since we're seeing multiple instances of wanting to disable power-intensive behavior when not connected to USB, would it be worth centralizing this behavior under EXT_POWER (I'll do my best in cobbling together a PR)?

app/Kconfig
```diff
@@ -317,6 +317,14 @@ config ZMK_RGB_UNDERGLOW_ON_START
 	bool "RGB underglow starts on by default"
 	default y
 
+config ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE
+	bool "Turn off RGB underglow when keyboard goes into idle state"
+	default y
```
petejohanson, line 322: Sorry I missed this before. I would say this being default on changes previous behavior. Can we remove this for now?

```suggestion
```

petejohanson (changes requested):
Sorry, one minor tweak left.

berfarah:
> That's strange, we've been testing this on a split without issue, what setup are you using?

Sorry for the delay - had a baby and it took up some mental time!

In my case, I have a [corne v3](https://keyhive.xyz/shop/corne-v3) with 5050 LEDs and a 110mAh battery using [this config](https://github.com/berfarah/zmk-config/tree/master/config). It's possible the battery just can't supply the wattage required for the LEDs and this logic doesn't have a chance to run before the nice nano is severely underpowered.

ReFil:
> > That's strange, we've been testing this on a split without issue, what setup are you using?
> 
> Sorry for the delay - had a baby and it took up some mental time!
> 
> In my case, I have a [corne v3](https://keyhive.xyz/shop/corne-v3) with 5050 LEDs and a 110mAh battery using [this config](https://github.com/berfarah/zmk-config/tree/master/config). It's possible the battery just can't supply the wattage required for the LEDs and this logic doesn't have a chance to run before the nice nano is severely underpowered.
> 

Well congratulations on the baby firstly! 5050 LEDs are very thirsty, approx 50ma at full brightness but I can't imagine it'd kill it that quickly. I tried but couldn't replicate the specific issue you were describing

ReFil:
Hi, just wondering if we can get this merged soon? a lot of our customers are interested in this feature getting merged

app/Kconfig
```diff
@@ -325,6 +325,13 @@ config ZMK_RGB_UNDERGLOW_ON_START
 	bool "RGB underglow starts on by default"
 	default y
 
+config ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE
+	bool "Turn off RGB underglow when keyboard goes into idle state"
+
+config ZMK_RGB_UNDERGLOW_AUTO_OFF_USB
+	bool "Turn off RGB underglow when USB is disconnected"
+	depends on USB
```
petejohanson, line 333:
```suggestion
	depends on USB_DEVICE_STACK
```

Needed w/ the Zephyr 3.0 update.

petejohanson (changes requested):
One code comment, and then a general item:

When testing this, the *initial* auto state isn't set for USB. If I've enabled the auto off USB setting, and I start the firmware from reset w/o USB connected, the RGB starts on, instead of off.

Need to update the auto stuff at the end of the init callback, I imagine.

app/Kconfig
ReFil: Rectified both issues

app/src/rgb_underglow.c
```diff
@@ -265,6 +271,11 @@ static int zmk_rgb_underglow_init(const struct device *_arg) {
     settings_load_subtree("rgb/underglow");
 #endif
 
+#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_USB)
+    if (!zmk_usb_is_powered())
+        state.on = false;
```
petejohanson, line 276:
```suggestion
    state.on = zmk_usb_is_powered();
```

petejohanson (changes requested):
Minor style tweak. This should get fixed for the backlight logic as well, I'll create an issue for that.

ReFil:
Once this is merged i can open a pr to sort that backlight thing as well

ReFil:
> @ReFil Tested with the latest changes, and even though the animation isn't running, if I have a saved state it seems to still initialize the underglow w/ a fixed color at reset.

by this do you mean it's overriding the animation state or it's not shutting off properly? on my little test board it was initialising with lights off properly

petejohanson:
Actually, looks like it is fine, I think the way I had done a reset through the programmer hadn't done the same as a full reset.

Only lingering issue: This seems to still start the *timer* for the animation, even if we start off. That timer start shuld be conditional.

ReFil:
Should we change it to call zmk_rgb_underglow_off and on then? that properly handles all the timer stopping and starting without intriducing superfluous code

petejohanson:
> Should we change it to call zmk_rgb_underglow_off and on then? that properly handles all the timer stopping and starting without intriducing superfluous code

I had considered that. The risk is that also persists the settings afterwards, which seems superfluous, and extra flash wear to be avoided.

Likely needs a small refactor to extract the "turn it off" from the "queue settings save" so we can invoke just the "turn it off" code.

ReFil:
What if i just wrap the k_timer_start in the if enabled block with an else when it's not enabled?

petejohanson:
> What if i just wrap the k_timer_start in the if enabled block with an else when it's not enabled?

I'd totally support that, yeah.

app/src/rgb_underglow.c
```diff
@@ -265,7 +271,13 @@ static int zmk_rgb_underglow_init(const struct device *_arg) {
     settings_load_subtree("rgb/underglow");
 #endif
 
+#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_USB)
+    state.on = zmk_usb_is_powered();
+    if (state.on)
+        k_timer_start(&underglow_tick, K_NO_WAIT, K_MSEC(50));
+#else
     k_timer_start(&underglow_tick, K_NO_WAIT, K_MSEC(50));
+#endif
```
petejohanson, line 280: Can be simplified, I believe:

```suggestion
#if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_USB)
    state.on = zmk_usb_is_powered();
#endif

    if (state.on) {
        k_timer_start(&underglow_tick, K_NO_WAIT, K_MSEC(50));
    }
```

petejohanson (changes requested):
Testing was looking good now. One minor style tweak needed for the init changes.

app/src/rgb_underglow.c
ReFil: I didnt want to do that bc i wasn't sure if the timer had to be running for other stuff and didn't want to interfere but if you're ok with it then that is definitely the better way

ReFil: This was sorta a bug, because the timer was running even if the leds are off on init, so probably goodto fix that, saves a bit of power

petejohanson: Yeah, that was my thinking as well. Thanks!

petejohanson (approved):
Thanks!

PR #1005, opened 2021-10-31
New feature: hold/tap flavor tap-unless-interrupted

jmding8:
While discussing https://github.com/zmkfirmware/zmk/pull/835, it became clear that `Positional Hold/Tap` would likely be quite useful in combination with a new flavor of hold/tap, `tap-unless-interrupted`

This flavor produces a tap behavior in all situations, except if another key is pressed before `tapping-term-ms` expires.

This is very useful with `Positional Hold/Tap` because:
- decide-as-hold-upon-other key-down is, in some ways, a strongly hold-preferring pattern because it is timing agnostic
- however, `Position Hold/Tap` gives us a mechanism to reject unwanted decide-as-hold-upon-other key-down events on the basis of position, which is independent from timing

PR #1018, opened 2021-11-06
New feature: hold/tap flavor tap-unless-interrupted (squashed)

jmding8:
# Description:

Same as https://github.com/zmkfirmware/zmk/pull/1005, just squashed and with better commit message.

### Description from original PR (1005):

While discussing https://github.com/zmkfirmware/zmk/pull/835, it became clear that `Positional Hold/Tap` would likely be quite useful in combination with a new flavor of hold/tap, `tap-unless-interrupted`

This flavor produces a tap behavior in all situations, except if another key is pressed before `tapping-term-ms` expires.

This is very useful with `Positional Hold/Tap` because:
- decide-as-hold-upon-other key-down is, in some ways, a strongly hold-preferring pattern because it is timing agnostic
- however, `Position Hold/Tap` gives us a mechanism to reject unwanted decide-as-hold-upon-other key-down events on the basis of position, which is independent from timing

### DIFF
--- a/app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
@@ -28,6 +28,7 @@ properties:
       - "hold-preferred"
       - "balanced"
       - "tap-preferred"
+      - "tap-unless-interrupted"
   retro-tap:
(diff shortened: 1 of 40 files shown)


petejohanson (approved):
Thanks!

PR #1022, opened 2021-11-10
feat(shield) Add Chalice support

joshajohnson:
Adding support for [Chalice](https://customkbd.com/products/chalice-pre-order).
This is my first ZMK PR so apologies for any issues and thanks in advance :)

## Board/Shield Check-list

### DIFF
--- a/app/boards/shields/chalice/chalice.zmk.yml
+++ b/app/boards/shields/chalice/chalice.zmk.yml
@@ -6,3 +6,4 @@ url: https://customkbd.com/products/chalice-pre-order
 requires: [pro_micro]
 features:
   - keys
+  - underglow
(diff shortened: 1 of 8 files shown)


app/boards/shields/chalice/chalice.zmk.yml
```diff
@@ -0,0 +1,8 @@
+file_format: "1"
+id: chalice
+name: Chalice
+type: shield
+url: https://customkbd.com/products/chalice-pre-order
+requires: [pro_micro]
+features:
+  - keys
```
Nicell, line 8:
```suggestion
  - keys
  - underglow
```

Nicell (changes requested):
Generally looks great, one small change, and then I was curious when this shield will start to ship, as we'd like to wait until then to merge it in.

joshajohnson:
Thanks for the review @Nicell 😄

Chalice is currently shipping, so if it's possible to merge now that would be great.
We caught the split backspace issue in the field from one of the early units.

Nicell (approved):

PR #1031, opened 2021-11-25
Fix docs for west command for custom config repo

issmirnov:
Updating docs, per discussion on https://discordapp.com/channels/719497620560543766/719909884769992755/913258608983166996

### DIFF
--- a/docs/docs/development/build-flash.md
+++ b/docs/docs/development/build-flash.md
@@ -99,6 +99,8 @@ For instance, building kyria firmware from a user `myUser`'s `zmk-config` folder
 west build -b nice_nano -- -DSHIELD=kyria_left -DZMK_CONFIG="C:/Users/myUser/Documents/Github/zmk-config/config"
 ```
 
+Note: The above command must be invoked from the `zmk/app` directory, or alternatively you will need to add the `-s /path/to/zmk/app` flag to your `west` command. Otherwise, you will see errors such as `ERROR: source directory "." does not contain a CMakeLists.txt; is this really what you want to build?`
+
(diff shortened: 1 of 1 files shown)


docs/docs/development/build-flash.md
```diff
@@ -99,6 +99,8 @@ For instance, building kyria firmware from a user `myUser`'s `zmk-config` folder
 west build -b nice_nano -- -DSHIELD=kyria_left -DZMK_CONFIG="C:/Users/myUser/Documents/Github/zmk-config/config"
 ```
 
+Note: The above command must be invoked from the `zmk/app` directory, or alternatively you will need to add the `-s /path/to/zmk/app` flag to your `west` command. Otherwise, you will see errors such as `ERROR: source directory "." does not contain a CMakeLists.txt; is this really what you want to build?`
```
caksoylar, line 102:
```suggestion
:::caution
The above command must still be invoked from the `zmk/app` directory as noted above, rather than the config directory. Otherwise, you will encounter errors such as `ERROR: source directory "." does not contain a CMakeLists.txt; is this really what you want to build?`. Alternatively you can add the `-s /path/to/zmk/app` flag to your `west` command.
:::
```
Thanks! A suggestion above to address two things:
- Converting `Note:` to a caution box
- Acknowledging that this was already noted at the beginning of the page, but repeated here for emphasis

caksoylar (approved):

docs/docs/development/build-flash.md
issmirnov: Sure thing! Thank you for the enhancement.

dxmh (approved):

Issue #853, opened 2021-06-23
Feature request: “Repeat key” behavior


dxmh:
A “repeat key” sends the previous keycode.

I believe this has been popularised by NotGate’s post on minimising single finger bigrams:

[NotGate’s SFB Minimized Layout](https://notgate.github.io/layout/#bonus):

> 2) Repeat Key: Bound to right alt (by default) and repeats the last key pressed. I suggest you try to use the thumb that doesn’t press space for this. The repeat key fits the theme of reducing SFBs by as much as possible and has the bonus of including the 10th finger. Words like follow become ‘fol[rep]ow’ so your flow is never broken by the speed of a single finger. It might seem overkill but you quickly get used to it and it’s great for keeping a steady pace. There are actually far more double letters than SFBs in typing, so technically this matters more for preserving flow than the layout’s SFB minimization 🙃.

However, a “repeat key” behavior would also be useful for more than typing double letters; for example, to repeat commonly repeated chorded shortcuts like:

* `ctrl` +  `n` (or even `ctrl` + `shift` + `g`) to cycle through occurrences of a word in a document (especially useful when one’s modifiers are in a separate layer!)
* digging into the adjust layer and sending `shift` + `alt` + `volume up` to increase volume in minor increments 

Below is a previous short discussion between boogerlad and petejohanson on discord regarding this feature:

boogerlad:
> can zmk assign a button to input the previous keycode? That is, if I pressed the "e" key, and then pressed the "repeat" button, another "e" would show up

petejohanson:
> No, but wouldn't be too hard of a behavior to write. What's the use case?

boogerlad:
> in the interest of reducing same finger bigrams, the isrt layout assigns the underutilized second thumb to the repeat function
> 
> for example, if you type "follow", you don't need to type "l" twice with the same finger, potentially improving efficiency
> 
> https://notgate.github.io/layout/#bonus

petejohanson:
> Interesting. To do that, you'd add a behavior that adds an event listener for the keycode state change event, and just capture the keycode of the most recent event to fire a pressed state.
> 
> Then in the behavior activated callback, you'd raise a new keycode state event w/ that same usage.
> 
> (some house keeping necessary to also send the right release codes, which complicates it a wee bit, but not massively so.

Source: https://discord.com/channels/719497620560543766/719909884769992755/841849093067636738

boogerlad:
Related: https://sites.google.com/alanreiser.com/handsdown/home#h.rt23wndkh65l
[...]

PR #1034, opened 2021-11-29
feat(behaviors): `&key_repeat` behavior + tests.

petejohanson:
* Add new `&key_repeat` behavior that captures and re-sends
  the most recently triggered keycode.


petejohanson:
Closes #853

dxmh:
Nice @petejohanson!

I have noticed that it doesn't currently repeat keys with modifiers applied:

* `alt`+`left` will only repeat `left`
* `cmd`+`space` will only repeat `space`
* `shift`+`t` will only repeat a lowercase `t`, etc. 

However, it does repeat modified keys when [modifier functions](https://zmk.dev/docs/codes/modifiers/#modifier-functions) are used:

* tap `&kp LG(V)` and it will repeat `cmd`+`v` (paste)
* hold `&kp LGUI`, tap `&kp V` and it will only repeat `v` 
* tap `&sk LGUI`, tap `&kp V` and it will only repeat `v` 

Is this intentional with the current implementation?

jcmkk3:
> Nice @petejohanson!
> 
> I have noticed that it doesn't currently repeat keys with modifiers applied:
> 
> * `alt`+`left` will only repeat `left`
> * `cmd`+`space` will only repeat `space`
> * `shift`+`t` will only repeat a lowercase `t`, etc.
> 
> However, it does repeat modified keys when [modifier functions](https://zmk.dev/docs/codes/modifiers/#modifier-functions) are used:
> 
> * tap `&kp LG(V)` and it will repeat `cmd`+`v` (paste)
> * hold `&kp LGUI`, tap `&kp V` and it will only repeat `v`
> * tap `&sk LGUI`, tap `&kp V` and it will only repeat `v`
> 
> Is this intentional with the current implementation? My initial expectation was that all three of the above scenarios would repeat `cmd`+`v` (paste).

This is a good callout. I was wondering if it would be applied this way or not. One of the reasons that it has always been desirable is to help repeating some mod-based shortcuts that can be a pain to repeat using callum-style mods or at the very least require planning to know that you're going to repeat it. `ctrl-d` for multiple cursors or `ctrl-s` for progressing to the next search candidate in emacs are examples of where I could see using it. I will probably start by putting it on my nav layer, but may later try to incorporate it into the alphas somehow.

petejohanson:
@dxmh @jcmkk3 Please pull the latest and try again. Added code to capture the current mods, along w/ corresponding test case.

PR #1037, opened 2021-11-29
fix(keymaps): Locate shared conf and overlay for split boards.

caksoylar:
Previously shared `<board>.conf` files in the ZMK config folder were not considered in the build for split boards (no shields), but only `<board>_left.conf`,  `<board>_right.conf`. It would make sense for them to be since `<board>.keymap` can now be used as a keymap and split shields also consider `<shield>.conf`. The shared `.conf` will also override any existing left- or right-specific `.conf` similar to how it works for split shields.

Since `.conf` and `.overlay` discovery logic seem to be very similar, I also added the same change for `<board>.overlay`.

Tested with the `corne-ish_zen` left and right boards in https://github.com/LOWPROKB/zmk/tree/Board-Corne-ish-Zen-dedicated-work-queue,

### DIFF
--- a/app/cmake/zmk_config.cmake
+++ b/app/cmake/zmk_config.cmake
@@ -105,8 +105,10 @@ if (ZMK_CONFIG)
 		endif()
 
 		# TODO: Board revisions?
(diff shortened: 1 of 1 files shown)


petejohanson (approved):
Thanks!

PR #1039, opened 2021-11-30
Core encoder/sensor refactor

petejohanson:
Some major refactors to sensors:

* Refactor ec11 sensor driver to align w/ Zephyr degree channel convention, that sensor data for that channel is in degrees of change since the last event.
* Refactor to allow specitying what angle delta triggers a given behavior.
* Move channel data into the sensor event, instead of requiring listeners to probe the sensor themselves upon events.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -190,18 +190,6 @@ rsource "src/split/Kconfig"
 #Basic Keyboard Setup
 endmenu
 
-menu "Encoders"
-
-config ZMK_ENCODERS_DEFAULT_TRIGGERS_PER_ROTATION
-    int "Default behavior triggers per rotation"
-    help
-      Unless overridden for a specific behavior in the keymap/devicetree, this value
-      determines how many times to trigger the bound behavior per full rotation.
-      For tactile encoders with detents, this usually should match the number of
-      detents per rotation of the encoder.
-    default 30
-
-endmenu
 menu "Display/LED Options"
 
 rsource "src/display/Kconfig"
@@ -544,6 +532,11 @@ if ZMK_KEYMAP_SENSORS
 
 config ZMK_KEYMAP_SENSORS_DEFAULT_TRIGGERS_PER_ROTATION
     int "Default triggers per rotation"
+    help
+      Unless overridden for a sensor in the board/shield/devicetree, this value
+      determines how many times to trigger the bound behavior per full rotation.
+      For tactile encoders with detents, this usually should match the number of
+      detents per rotation of the encoder.
     default 20
 
 endif # ZMK_KEYMAP_SENSORS
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -131,7 +131,7 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
 
 config ZMK_BLE_THREAD_STACK_SIZE
     int "BLE notify thread stack size"
-    default 512
+    default 768
 
 config ZMK_BLE_THREAD_PRIORITY
     int "BLE notify thread priority"
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -190,6 +190,18 @@ rsource "src/split/Kconfig"
 #Basic Keyboard Setup
 endmenu
 
+menu "Encoders"
+
+config ZMK_ENCODERS_DEFAULT_TRIGGERS_PER_ROTATION
+    int "Default behavior triggers per rotation"
+    help
+      Unless overridden for a specific behavior in the keymap/devicetree, this value
+      determines how many times to trigger the bound behavior per full rotation.
+      For tactile encoders with detents, this usually should match the number of
+      detents per rotation of the encoder.
+    default 30
+
+endmenu
 menu "Display/LED Options"
 
 rsource "src/display/Kconfig"
@@ -518,6 +530,14 @@ config ZMK_WPM
 
 config SENSOR
     default y
+
+if ZMK_KEYMAP_SENSORS
+
+config ZMK_KEYMAP_SENSORS_DEFAULT_TRIGGERS_PER_ROTATION
+    int "Default triggers per rotation"
+    default 20
+
+endif # ZMK_KEYMAP_SENSORS
 
 choice CBPRINTF_IMPLEMENTATION
     default CBPRINTF_NANO
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -114,9 +114,9 @@ menuconfig ZMK_BLE
     select BT_SMP_APP_PAIRING_ACCEPT
     select BT_PERIPHERAL
     select BT_DIS
-    select BT_BAS
     select BT_SETTINGS
     select SETTINGS
+    imply ZMK_BATTERY_REPORTING
 
 if ZMK_BLE
 
@@ -317,6 +317,12 @@ endif
 endmenu
 
 menu "Power Management"
+
+config ZMK_BATTERY_REPORTING
+    bool "Battery level detection/reporting"
+    default n
+    select SENSOR
+    select BT_BAS if ZMK_BLE
 
 config ZMK_IDLE_TIMEOUT
     int "Milliseconds of inactivity before entering idle state (OLED shutoff, etc)"
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -528,8 +528,11 @@ config ZMK_WPM
     bool "Calculate WPM"
     default n
 
-config SENSOR
+config ZMK_KEYMAP_SENSORS
+    bool "Enable Keymap Sensors support"
     default y
+    depends on DT_HAS_ZMK_KEYMAP_SENSORS_ENABLED
+    select SENSOR
 
 if ZMK_KEYMAP_SENSORS
 
--- a/app/boards/shields/zmk_uno/zmk_uno.overlay
+++ b/app/boards/shields/zmk_uno/zmk_uno.overlay
@@ -40,10 +40,10 @@
 
     // Commented out until we add more powerful power domain support
     // external_power {
-    //  compatible = "zmk,ext-power-generic";
-    //  label = "EXT_POWER";
-    //  init-delay-ms = <200>;
-    //  control-gpios = <&arduino_header 1 GPIO_ACTIVE_LOW>;
+        //  compatible = "zmk,ext-power-generic";
+        //  label = "EXT_POWER";
+        //  init-delay-ms = <200>;
+        //  control-gpios = <&arduino_header 1 GPIO_ACTIVE_LOW>;
     // };
 
     rgb_power {
@@ -128,14 +128,14 @@
         diode-direction = "col2row";
 
         col-gpios
-            = <&arduino_header 10 GPIO_ACTIVE_HIGH>
-            , <&arduino_header 9  GPIO_ACTIVE_HIGH>
-            ;
+        = <&arduino_header 10 GPIO_ACTIVE_HIGH>
+        , <&arduino_header 9  GPIO_ACTIVE_HIGH>
+        ;
 
         row-gpios
-            = <&arduino_header 13 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>
-            , <&arduino_header 11 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>
-            ;
+        = <&arduino_header 13 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>
+        , <&arduino_header 11 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>
+        ;
 
     };
 
@@ -144,11 +144,11 @@
         status = "disabled";
 
         input-gpios
-            = <&arduino_header 10 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
-            , <&arduino_header 9  (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
-            , <&arduino_header 13 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
-            , <&arduino_header 11 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
-            ;
+        = <&arduino_header 10 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+        , <&arduino_header 9  (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+        , <&arduino_header 13 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+        , <&arduino_header 11 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+        ;
 
     };
 
@@ -157,23 +157,26 @@
         toggle-mode;
 
         input-gpios
-            = <&arduino_header 4 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
-            , <&arduino_header 3 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
-            , <&arduino_header 2 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
-            ;
+        = <&arduino_header 4 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+        , <&arduino_header 3 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+        , <&arduino_header 2 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>
+        ;
     };
 
     encoder: encoder {
         label = "ENCODER";
-        resolution = <4>;
+        steps = <80>;
         compatible = "alps,ec11";
-        a-gpios = <&arduino_header 14 GPIO_PULL_UP>;
-        b-gpios = <&arduino_header 15 GPIO_PULL_UP>;
+        a-gpios = <&arduino_header 15 GPIO_PULL_UP>;
+        b-gpios = <&arduino_header 14 GPIO_PULL_UP>;
     };
-
     sensors {
         compatible = "zmk,keymap-sensors";
         sensors = <&encoder>;
+        triggers-per-rotation = <20>;
+        left {
+            triggers-per-rotation = <20>;
+        };
     };
 
 };
--- a/app/drivers/sensor/ec11/ec11.c
+++ b/app/drivers/sensor/ec11/ec11.c
@@ -16,6 +16,8 @@
 
 #include "ec11.h"
 
+#define FULL_ROTATION 360
+
 LOG_MODULE_REGISTER(EC11, CONFIG_SENSOR_LOG_LEVEL);
 
 static int ec11_get_ab_state(const struct device *dev) {
@@ -59,9 +61,14 @@ static int ec11_sample_fetch(const struct device *dev, enum sensor_channel chan)
     drv_data->pulses += delta;
     drv_data->ab_state = val;
 
-    drv_data->ticks = drv_data->pulses / drv_cfg->resolution;
-    drv_data->delta = delta;
-    drv_data->pulses %= drv_cfg->resolution;
+    // TODO: Temporary code for backwards compatibility to support
+    // the sensor channel rotation reporting *ticks* instead of delta of degrees.
+    // REMOVE ME
+    if (drv_cfg->steps == 0) {
+        drv_data->ticks = drv_data->pulses / drv_cfg->resolution;
+        drv_data->delta = delta;
+        drv_data->pulses %= drv_cfg->resolution;
+    }
 
     return 0;
 }
@@ -69,13 +76,26 @@ static int ec11_sample_fetch(const struct device *dev, enum sensor_channel chan)
 static int ec11_channel_get(const struct device *dev, enum sensor_channel chan,
                             struct sensor_value *val) {
     struct ec11_data *drv_data = dev->data;
+    const struct ec11_config *drv_cfg = dev->config;
+    int32_t pulses = drv_data->pulses;
 
     if (chan != SENSOR_CHAN_ROTATION) {
         return -ENOTSUP;
     }
 
-    val->val1 = drv_data->ticks;
-    val->val2 = drv_data->delta;
+    drv_data->pulses = 0;
+
+    if (drv_cfg->steps > 0) {
+        val->val1 = (pulses * FULL_ROTATION) / drv_cfg->steps;
+        val->val2 = (pulses * FULL_ROTATION) % drv_cfg->steps;
+        if (val->val2 != 0) {
+            val->val2 *= 1000000;
+            val->val2 /= drv_cfg->steps;
+        }
+    } else {
+        val->val1 = drv_data->ticks;
+        val->val2 = drv_data->delta;
+    }
 
     return 0;
 }
@@ -132,7 +152,8 @@ int ec11_init(const struct device *dev) {
     const struct ec11_config ec11_cfg_##n = {                                                      \
         .a = GPIO_DT_SPEC_INST_GET(n, a_gpios),                                                    \
         .b = GPIO_DT_SPEC_INST_GET(n, b_gpios),                                                    \
-        COND_CODE_0(DT_INST_NODE_HAS_PROP(n, resolution), (1), (DT_INST_PROP(n, resolution))),     \
+        .resolution = DT_INST_PROP_OR(n, resolution, 1),                                           \
+        .steps = DT_INST_PROP_OR(n, steps, 0),                                                     \
     };                                                                                             \
     DEVICE_DT_INST_DEFINE(n, ec11_init, NULL, &ec11_data_##n, &ec11_cfg_##n, POST_KERNEL,          \
                           CONFIG_SENSOR_INIT_PRIORITY, &ec11_driver_api);
--- a/app/drivers/sensor/ec11/ec11.h
+++ b/app/drivers/sensor/ec11/ec11.h
@@ -14,6 +14,7 @@ struct ec11_config {
     const struct gpio_dt_spec a;
     const struct gpio_dt_spec b;
 
+    const uint16_t steps;
     const uint8_t resolution;
 };
 
--- a/app/drivers/zephyr/dts/bindings/sensor/alps,ec11.yaml
+++ b/app/drivers/zephyr/dts/bindings/sensor/alps,ec11.yaml
@@ -18,4 +18,9 @@ properties:
   resolution:
     type: int
     description: Number of pulses per tick
+    deprecated: true
+    required: false
+  steps:
+    type: int
+    description: Number of pulses in one full rotation
     required: false
--- a/app/include/drivers/behavior.h
+++ b/app/include/drivers/behavior.h
@@ -23,9 +23,17 @@
  * (Internal use only.)
  */
 
+enum behavior_sensor_binding_process_mode {
+    BEHAVIOR_SENSOR_BINDING_PROCESS_MODE_TRIGGER,
+    BEHAVIOR_SENSOR_BINDING_PROCESS_MODE_DISCARD,
+};
+
 typedef int (*behavior_keymap_binding_callback_t)(struct zmk_behavior_binding *binding,
                                                   struct zmk_behavior_binding_event event);
-typedef int (*behavior_sensor_keymap_binding_callback_t)(
+typedef int (*behavior_sensor_keymap_binding_process_callback_t)(
+    struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,
+    enum behavior_sensor_binding_process_mode mode);
+typedef int (*behavior_sensor_keymap_binding_data_callback_t)(
     struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,
     const struct zmk_sensor_config *sensor_config, size_t channel_data_size,
     const struct zmk_sensor_channel_data channel_data[channel_data_size]);
@@ -41,7 +49,8 @@ __subsystem struct behavior_driver_api {
     behavior_keymap_binding_callback_t binding_convert_central_state_dependent_params;
     behavior_keymap_binding_callback_t binding_pressed;
     behavior_keymap_binding_callback_t binding_released;
-    behavior_sensor_keymap_binding_callback_t sensor_binding_triggered;
+    behavior_sensor_keymap_binding_data_callback_t sensor_binding_data;
+    behavior_sensor_keymap_binding_process_callback_t sensor_binding_process;
 };
 /**
  * @endcond
@@ -151,7 +160,7 @@ static inline int z_impl_behavior_keymap_binding_released(struct zmk_behavior_bi
 }
 
 /**
- * @brief Handle the a sensor keymap binding being triggered
+ * @brief Handle the a sensor keymap binding processing any incoming data from the sensor
  * @param binding Sensor keymap binding which was triggered.
  * @param sensor Pointer to the sensor device structure for the sensor driver instance.
  * @param virtual_key_position ZMK_KEYMAP_LEN + sensor number
@@ -160,12 +169,12 @@ static inline int z_impl_behavior_keymap_binding_released(struct zmk_behavior_bi
  * @retval 0 If successful.
  * @retval Negative errno code if failure.
  */
-__syscall int behavior_sensor_keymap_binding_triggered(
+__syscall int behavior_sensor_keymap_binding_data(
     struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,
     const struct zmk_sensor_config *sensor_config, size_t channel_data_size,
     const struct zmk_sensor_channel_data *channel_data);
 
-static inline int z_impl_behavior_sensor_keymap_binding_triggered(
+static inline int z_impl_behavior_sensor_keymap_binding_data(
     struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,
     const struct zmk_sensor_config *sensor_config, size_t channel_data_size,
     const struct zmk_sensor_channel_data *channel_data) {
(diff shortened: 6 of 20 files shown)


ReFil:
If it's alright I'll send you a revised rotr folder with a better keymap and a few definition tweaks to match the bt60s

app/Kconfig
```diff
@@ -482,8 +494,12 @@ config ZMK_WPM
 	bool "Calculate WPM"
 	default n
 
-config SENSOR
+DT_COMPAT_ZMK_KEYMAP_SENSORS := zmk,keymap-sensors
```
joelspadin, line 497: Does battery.c require `SENSOR` be enabled?

petejohanson: Sure does. Added a new commit here that creates a new `ZMK_BATTERY_REPORTING` Kconfig symbol:

```
config ZMK_BATTERY_REPORTING
	bool "Battery level detection/reporting"
	default n
	select SENSOR
	select BT_BAS if ZMK_BLE
```

That is `imply`d by `ZMK_BLE` automatically.

Thoughts?

app/drivers/sensor/ec11/ec11.c
```diff
@@ -59,23 +61,38 @@ static int ec11_sample_fetch(const struct device *dev, enum sensor_channel chan)
...
         return -ENOTSUP;
     }
 
-    val->val1 = drv_data->ticks;
-    val->val2 = drv_data->delta;
+    drv_data->pulses = 0;
+
+    if (drv_cfg->steps > 0) {
+        val->val1 = (pulses * FULL_ROTATION) / drv_cfg->steps;
+        val->val2 = (pulses * FULL_ROTATION) - (val->val1 * drv_cfg->steps);
```
joelspadin, line 87: It took me a while to figure out that this is just getting the remainder. Consider switching to
```suggestion
        val->val2 = (pulses * FULL_ROTATION) % drv_cfg->steps;
```
for better readability. 

Doing a division gets you the remainder for free, so this gets optimized to a single operation either way you write it: https://godbolt.org/z/Y9rcdP844

app/include/zmk/events/sensor_event.h
```diff
@@ -6,12 +6,21 @@
...
+#include <zmk/event_manager.h>
+#include <zmk/sensors.h>
+
+// TODO: Move to Kconfig when we need more than one channel
+#define ZMK_SENSOR_EVENT_MAX_CHANNELS 1
+
 struct zmk_sensor_event {
-    uint8_t sensor_number;
-    const struct device *sensor;
+    uint8_t sensor_position;
```
joelspadin, line 19: Sensors can be things that sense a change in physical position, so this name might be a bit confusing. If this is the sensor's index within the list of sensors, maybe call it an index instead of a position? If this is the virtual key position assigned to the sensor, maybe use `key_position` or `virtual_key_position`?

Also, are we actually saving any space by representing sensor positions as `uint8_t` instead of just `int`? At least in this struct, it's just going to get padded out to the size of an int anyways: https://godbolt.org/z/5YrKqPvKh

app/include/zmk/sensors.h
```diff
@@ -6,6 +6,9 @@
 
 #pragma once
 
+#include <zephyr/drivers/sensor.h>
+
+#define _SENSOR_CHILD_LEN(node) 1 +
```
joelspadin, line 11: Identifiers beginning with an underscore followed by an uppercase letter are reserved for the compiler.

app/src/behaviors/behavior_sensor_rotate_common.c
```diff
@@ -5,48 +5,92 @@
...
+        remainder.val1 += value.val1;
+        remainder.val2 += value.val2;
+
+        if (remainder.val2 >= 1000000 || remainder.val2 <= 1000000) {
+            remainder.val1 += remainder.val2 / 1000000;
+            remainder.val2 %= 1000000;
+        }
+
+        triggers = remainder.val1 / (360 / sensor_config->triggers_per_rotation);
+        remainder.val1 -= triggers * (360 / sensor_config->triggers_per_rotation);
```
joelspadin, line 42: As above, consider using modulo to get the remainder.

app/src/behaviors/behavior_sensor_rotate_common.c
```diff
@@ -5,48 +5,92 @@
...
-    const struct behavior_sensor_rotate_config *cfg = dev->config;
+    struct behavior_sensor_rotate_data *data = dev->data;
+
+    const struct sensor_value value = channel_data[0].value;
+    int triggers;
+    int sensor_position = ZMK_SENSOR_POSITION_FROM_VIRTUAL_KEY_POSITION(event.position);
+
+    // Some funky special casing for "old encoder behavior" where ticks where reported in val2 only,
+    // instead of rotational degrees in val1.
+    if (value.val1 == 0) {
```
joelspadin, line 28: Doesn't the new encoder driver report degrees in val1 and fractional degrees in val2? If so, won't that mean that if I position my knob at exactly 0 degrees and then rotate it by less than one degree, it will generate somewhere up to a million triggers?

As much as I'd like to maintain backwards compatibility, as I'm reading through this, I'm feeling more and more like it would be simpler to just add forced compiler errors if you haven't set `steps` or `triggers_per_rotation`, and have the error messages link to ZMK documentation describing how to set them (or describe inline if we can find a way to make that short enough).

joelspadin (changes requested):

app/drivers/sensor/ec11/ec11.c
```diff
@@ -59,23 +61,38 @@ static int ec11_sample_fetch(const struct device *dev, enum sensor_channel chan)
     drv_data->pulses += delta;
     drv_data->ab_state = val;
 
-    drv_data->ticks = drv_data->pulses / drv_cfg->resolution;
-    drv_data->delta = delta;
-    drv_data->pulses %= drv_cfg->resolution;
+    if (drv_cfg->steps == 0) {
+        drv_data->ticks = drv_data->pulses / drv_cfg->resolution;
+        drv_data->delta = delta;
```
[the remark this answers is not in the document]
petejohanson: It definitely would. The expectation is you "enable" both the new `steps` at the same time you enable the new config in the ZMK sensors node so both sets of "new code" are turned on at the same time.

app/boards/shields/zmk_uno/zmk_uno.overlay
```diff
@@ -157,23 +157,26 @@
...
+        b-gpios = <&arduino_header 14 GPIO_PULL_UP>;
     };
-
     sensors {
         compatible = "zmk,keymap-sensors";
         sensors = <&encoder>;
+        triggers-per-rotation = <20>;
+        left {
+            triggers-per-rotation = <20>;
+        };
```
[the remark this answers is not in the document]
petejohanson: The config here allows setting a global value, or override the setting per sensor. Needed in case, say, you have a tactile encoder with a certain number of detents per rotation, and next to it a linear encoder that you want to trigger behaviors more frequently. This gets pulled in by the sensor code and leveraged by the keymap.

app/Kconfig
```diff
@@ -305,6 +317,12 @@ endmenu
 
 menu "Power Management"
 
+config ZMK_BATTERY_REPORTING
```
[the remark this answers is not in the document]
petejohanson: I didn't find any code to do so in Zephyr's python Kconfig code, sadly.

app/include/zmk/events/sensor_event.h
petejohanson: Refactored to the terminology "sensor index" everywhere.

Storage in the struct will be padded, sure, but elsewhere for stack allocation, etc. it'll be lower. Also, if the event struct has more added later it might not actually grow, and can use some of the previous padding.

app/include/zmk/sensors.h
petejohanson: Isn't actually used any more, so removed it.

app/src/behaviors/behavior_sensor_rotate_common.c
petejohanson: Indeed. Fixed.

app/src/behaviors/behavior_sensor_rotate_common.c
petejohanson: I don't think we *yet* have any encoders in use that have that high a resolution, so I think the risk here is low. I do think we need to get folks migrated as quickly as possible, but would rather not hard break right away. I'd imagine:

* Merge this PR, but include config docs and blog post outlining the config changes needed
* Add `#warning` in the code that would trigger when using the old deprecated prop(s)
* Roll this out, announce in Discord, social media, etc.
* Give it some time (month?) then remove the old code/configs.

Thoughts?

joelspadin: I think if we just use a `#warning`, almost nobody will ever see that (why would you go looking through the logs on your GH actions for a _successful_ build?). Anyone who's not watching Discord etc. will just see a hard break later anyways.

petejohanson: > I think if we just use a `#warning`, almost nobody will ever see that (why would you go looking through the logs on your GH actions for a _successful_ build?). Anyone who's not watching Discord etc. will just see a hard break later anyways.

I think just because we might miss *some* folks with our announcements/warnings, doesn't mean we shouldn't make the effort to reduce it as much as we can. Agreed that the warning might be a waste of time though, we'll get more bang for our buck w/ blog post that's posted to Twitter, Discord, etc. to catch as many as we can that way.

joelspadin: I think as long as we clearly document in the code that this needs to be changed, and we plan to force a compiler error when we eventually change it, then that would be fine.

app/boards/shields/zmk_uno/zmk_uno.overlay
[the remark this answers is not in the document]
joelspadin: I'm just confused by the overriding of a "left" encoder on a thing that only has one encoder as far as I can tell.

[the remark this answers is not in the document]
petejohanson: For example and testing, more than requirement. Made it easy to comment out portions locally when testing this.

app/drivers/sensor/ec11/ec11.c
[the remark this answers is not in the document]
joelspadin: Can we add a comment explaining that this is temporary code for backwards compatibility?

app/include/zmk/events/sensor_event.h
joelspadin: > Also, if the event struct has more added later it might not actually grow, and can use some of the previous padding.

The compiler can't reorder structs, so you should place the smaller fields at the end if you want that to happen.

app/drivers/sensor/ec11/ec11.h
```diff
@@ -15,6 +15,7 @@ struct ec11_config {
     const struct gpio_dt_spec b;
 
     const uint8_t resolution;
+    const uint8_t steps;
```
joelspadin, line 18: I haven't seen such an encoder yet, but it seems like it should be possible for an encoder to exists that has more than 255 steps.

app/drivers/zephyr/dts/bindings/sensor/alps,ec11.yaml
```diff
@@ -19,3 +19,7 @@ properties:
     type: int
     description: Number of pulses per tick
     required: false
```
joelspadin, line 21: Is there a way we can mark this as deprecated?

app/drivers/zephyr/dts/bindings/sensor/alps,ec11.yaml
```diff
@@ -19,3 +19,7 @@ properties:
...
+  steps:
```
joelspadin, line 22: Should we call this `steps-per-rotation` to match `triggers-per-rotation`?

app/drivers/zephyr/dts/bindings/sensor/alps,ec11.yaml
petejohanson: Looks like it: https://docs.zephyrproject.org/3.2.0/build/dts/bindings.html#property-entry-syntax

Will do.

app/src/sensors.c
```diff
@@ -18,65 +18,144 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+        CONTAINER_OF(trigger, struct sensors_item_cfg, trigger);
+    int sensor_index = test_item - sensors;
+
+    if (sensor_index < 0 || sensor_index >= ARRAY_SIZE(sensors)) {
+        LOG_ERR("Invalid sensor item triggered our callback");
+        return;
+    }
+
+    if (k_is_in_isr()) {
+        if (k_sem_take(&pending_sensors_sem, K_MSEC(1)) >= 0) {
```
joelspadin, line 128: The documentation says "The kernel does allow an ISR to take a semaphore, however the ISR must not attempt to wait if the semaphore is unavailable." Does that mean that the wait time should be 0? It seems like that could lead to dropped sensor events if run_sensors_data_trigger() was running.

I think you could do this more reliably with an atomic variable instead. This part would just be
```c
atomic_set_bit(&pending_sensors, sensor_index);
```
and you could either change run_sensors_data_trigger to use
```c
pd = atomic_set(&pending_sensors, 0);
for (...)
```
or simplify the loop to
```c
for (int i = 0; i < ARRAY_SIZE(sensors); i++) {
    if (atomic_test_and_clear_bit(&pending_sensors, i)) {
        trigger_sensor_data_for_position(i);
    }
}
```
(this should work just the same, though it isn't _exactly_ the same since if a bit gets set while it's iterating through the loop, it could get immediately handled instead of waiting until the next call to run_sensors_data_trigger.)

app/src/sensors.c
```diff
@@ -18,65 +18,144 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+#if ZMK_KEYMAP_SENSORS_CHILD_COUNT > 0
 
-static void zmk_sensors_trigger_handler(const struct device *dev,
-                                        const struct sensor_trigger *trigger) {
-    int err;
-    const struct sensors_data_item *item = CONTAINER_OF(trigger, struct sensors_data_item, trigger);
+#define SENSOR_CHILD_ITEM(node)                                                                    \
+    {                                                                                              \
+        .triggers_per_rotation = DT_PROP_OR(                                                       \
+            node, triggers_per_rotation, DT_PROP(ZMK_KEYMAP_SENSORS_NODE, triggers_per_rotation))  \
```
joelspadin, line 45: With this using DT_PROP instead of DT_PROP_OR, does that mean this property is required to exist on either the sensors node or the child node, even once we support sensors that aren't rotational?

app/src/sensors.c
```diff
@@ -18,65 +18,144 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+static struct zmk_sensor_config configs[] = {
 
-static struct sensors_data_item sensors[] = {LISTIFY(ZMK_KEYMAP_SENSORS_LEN, SENSOR_ITEM, (, ), 0)};
+#if ZMK_KEYMAP_SENSORS_CHILD_COUNT > 0
 
-static void zmk_sensors_trigger_handler(const struct device *dev,
-                                        const struct sensor_trigger *trigger) {
-    int err;
-    const struct sensors_data_item *item = CONTAINER_OF(trigger, struct sensors_data_item, trigger);
+#define SENSOR_CHILD_ITEM(node)                                                                    \
```
joelspadin, line 42: Minor: `#define`ing something in the middle of an array initializer is a bit confusing. Would be a bit easier to read if this was

```c
#define SENSOR_CHILD_ITEM ...
#define SENSOR_CHILD_DEFAULTS ...

static struct zmk_sensor_config configs[] = {
#if ZMK_KEYMAP_SENSORS_CHILD_COUNT > 0
    DT_FOREACH_CHILD_SEP(ZMK_KEYMAP_SENSORS_NODE, SENSOR_CHILD_ITEM, (, ))
#else
    LISTIFY(ZMK_KEYMAP_SENSORS_LEN, SENSOR_CHILD_DEFAULTS, (, ), 0)
#endif
};
```

Not sure if you can do math to determine the LISTIFY count, but if so, you could also maybe write it as 

```c
static struct zmk_sensor_config configs[] = {
    DT_FOREACH_CHILD_SEP(ZMK_KEYMAP_SENSORS_NODE, SENSOR_CHILD_ITEM, (, )),
    LISTIFY(ZMK_KEYMAP_SENSORS_LEN - ZMK_KEYMAP_SENSORS_CHILD_COUNT, SENSOR_CHILD_DEFAULTS, (, ), 0)
};
```

app/src/sensors.c
```diff
@@ -18,65 +18,144 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+        LOG_WRN("Failed to set sensor trigger (%d)", err);
+    }
 }
 
-#define _SENSOR_INIT(node)                                                                         \
-    zmk_sensors_init_item(DT_PROP(node, label), local_index++, absolute_index++);
-#define SENSOR_INIT(idx, _i)                                                                       \
-    COND_CODE_1(DT_NODE_HAS_STATUS(ZMK_KEYMAP_SENSORS_BY_IDX(idx), okay),                          \
-                (_SENSOR_INIT(ZMK_KEYMAP_SENSORS_BY_IDX(idx))), (absolute_index++;))
+#define _SENSOR_INIT(idx, _t) zmk_sensors_init_item(idx);
```
joelspadin, line 154: Underscore followed by capital letter is reserved for use by the compiler.

app/drivers/zephyr/dts/bindings/sensor/alps,ec11.yaml
petejohanson: The `steps` naming was chosen to match the NRFX qdec driver that uses the same property name.

app/src/sensors.c
joelspadin: As a bonus, `ATOMIC_DEFINE(<sensor count>)` would automatically expand to the needed size no matter how many sensors there are instead of silently breaking when you go beyond 32.

app/src/sensors.c
petejohanson: Correct, although when we do that, my thinking is to make this particular property optional, and add other settings here that might be relevant to those, e.g. acceleration, inverted-X, etc.

app/src/sensors.c
petejohanson: I prefer the a bit-more explicit first version. Good catch.

app/drivers/zephyr/dts/bindings/sensor/alps,ec11.yaml
petejohanson: Added.

app/drivers/sensor/ec11/ec11.h
petejohanson: Yes, the upcoming ma730 hall effect encoder has super high resolution. But ec11 encoders with that high a resolution... Haven't seen any *yet*... I suppose they could exist though.

app/src/keymap.c
```diff
@@ -252,41 +252,59 @@ int zmk_keymap_position_state_changed(uint8_t source, uint32_t position, bool pr
...
+            .layer = layer,
+            .position = ZMK_VIRTUAL_KEY_POSITION_SENSOR(sensor_index),
+            .timestamp = timestamp,
+        };
+
+        ret = behavior_sensor_keymap_binding_data(binding, event,
+                                                  zmk_sensors_get_config_at_index(sensor_index),
+                                                  channel_data_size, channel_data);
+
+        if (ret > 0) {
```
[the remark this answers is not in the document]
petejohanson: Nope. Typo. Good catch.

app/include/drivers/behavior.h
```diff
@@ -158,14 +169,50 @@ static inline int z_impl_behavior_keymap_binding_released(struct zmk_behavior_bi
  * @retval 0 If successful.
  * @retval Negative errno code if failure.
  */
-__syscall int behavior_sensor_keymap_binding_triggered(struct zmk_behavior_binding *binding,
-                                                       const struct device *sensor,
-                                                       struct zmk_behavior_binding_event event);
+__syscall int behavior_sensor_keymap_binding_data(
```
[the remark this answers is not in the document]
petejohanson: Renamed to `behavior_sensor_keymap_binding_accept_data`. Thoughts?

app/src/keymap.c
```diff
@@ -252,41 +252,59 @@ int zmk_keymap_position_state_changed(uint8_t source, uint32_t position, bool pr
...
+                            const struct zmk_sensor_channel_data channel_data[channel_data_size],
+                            int64_t timestamp) {
+    bool opaque_response = false;
 
-            LOG_DBG("layer: %d sensor_number: %d, binding name: %s", layer, sensor_number,
-                    binding->behavior_dev);
+    for (int layer = ZMK_KEYMAP_LAYERS_LEN - 1; layer >= 0; layer--) {
+        struct zmk_behavior_binding *binding = &zmk_sensor_keymap[layer][sensor_index];
+        const struct device *behavior;
+        int ret;
```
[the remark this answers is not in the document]
petejohanson: Done.

app/src/keymap.c
```diff
@@ -252,41 +252,59 @@ int zmk_keymap_position_state_changed(uint8_t source, uint32_t position, bool pr
...
 #if ZMK_KEYMAP_HAS_SENSORS
-int zmk_keymap_sensor_triggered(uint8_t sensor_number, const struct device *sensor,
-                                int64_t timestamp) {
-    for (int layer = ZMK_KEYMAP_LAYERS_LEN - 1; layer >= _zmk_keymap_layer_default; layer--) {
-        if (zmk_keymap_layer_active(layer)) {
-            struct zmk_behavior_binding *binding = &zmk_sensor_keymap[layer][sensor_number];
-            const struct device *behavior;
-            int ret;
+int zmk_keymap_sensor_event(uint8_t sensor_index, size_t channel_data_size,
+                            const struct zmk_sensor_channel_data channel_data[channel_data_size],
```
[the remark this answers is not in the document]
petejohanson: Nope. Changed to pointer for consistency.

app/src/keymap.c
```diff
@@ -252,41 +252,59 @@ int zmk_keymap_position_state_changed(uint8_t source, uint32_t position, bool pr
...
+            .position = ZMK_VIRTUAL_KEY_POSITION_SENSOR(sensor_index),
+            .timestamp = timestamp,
+        };
+
+        ret = behavior_sensor_keymap_binding_data(binding, event,
+                                                  zmk_sensors_get_config_at_index(sensor_index),
+                                                  channel_data_size, channel_data);
+
+        if (ret > 0) {
+            LOG_DBG("behavior processing to continue to next layer");
```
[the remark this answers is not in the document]
petejohanson: Added, with the error code.

app/include/zmk/events/sensor_event.h
petejohanson: Done.

app/drivers/sensor/ec11/ec11.h
joelspadin: By way of DigiKey search, I found one optical encoder with quadrature output at 512 pulses per revolution, and two magnetic encoders with quadrature output and 256 pulses per revolution. Seems unlikely that they'd be used, but you did make triggers_per_rotation be a uint16, so it seems reasonable to me for steps to also be one.

app/include/drivers/behavior.h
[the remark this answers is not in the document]
joelspadin: I think that explains it pretty well.

app/src/sensors.c
joelspadin: Assuming I have correctly determine which DT bindings map to this, https://github.com/zmkfirmware/zmk/pull/1039/files#diff-d2a624d29650518030a8f943eeffa937b4a0162e9d8ea333d231116fb0d3f2abR22 says this property isn't required. Could lead to some difficult to parse compile errors.

app/src/sensors.c
petejohanson: Switched to the atomic approach, tested on ZMK Uno w/ the QDEC driver that causes the ISR code to be hit, working fine. Thanks!

app/src/sensors.c
petejohanson: Added a new default fallback value in Kconfig. Thoughts?

app/drivers/sensor/ec11/ec11.c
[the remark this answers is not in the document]
petejohanson: Added.

app/drivers/sensor/ec11/ec11.h
petejohanson: Yeah, fair. Bumped it to 16-bit.

app/src/behaviors/behavior_sensor_rotate_common.c
petejohanson: Added a blog post to document the changes: https://deploy-preview-1039--zmk.netlify.app/blog/2023/06/10/encoder-refactors

Thoughts?

docs/blog/2023-06-10-encoder-refactors.md
```diff
@@ -0,0 +1,119 @@
...
+
+1. Split configuration of hardware details, and behavior configuration to allow more flexible functionality of sensors/encoders, in particular linear encoders that lack detents/"clicks" as they rotate.
+2. Support for upstream Zephyr sensor drivers, including the NRFX QDEC driver that can be used on nRF52 based keyboards.
+3. Sensor data handling changes that pave the way for split sensor handling easily.
+
+## Configuration Changes
+
+The major changes to configuration in the devicetree files relates to how the number of steps/triggers for a given encoder are set. In particular, the number of pulses/steps for a given encoder is configured first, allowing ZMK to determine the exact angular degrees of change that is represented by a single pulse on the data lines to that encoder.
+
+Once that angular degrees mapping is completed, now independently there is a configuration setting to control how many triggers of the behavior in the keymap should occur for each full rotation of the sensor. Another way to think of this is "how many degrees of rotation results in a triggering of the sensor behavior in your keymap layer.
```
caksoylar, line 30:
```suggestion
Once that angular degrees mapping is completed, now independently there is a configuration setting to control how many triggers of the behavior in the keymap should occur for each full rotation of the sensor. Another way to think of this is "how many degrees of rotation results in a triggering of the sensor behavior in your keymap layer".
```

docs/blog/2023-06-10-encoder-refactors.md
```diff
@@ -0,0 +1,119 @@
...
+        sensors = <&left_encoder &right_encoder>;
+        triggers-per-rotation = <20>;
+    };
+```
+
+For tactile encoders that have detents, the `triggers-per-rotation` would match the number of detents on the encoder. For linear encoders, the value can be chosen to suit your needs.
+
+## Zephyr Sensor Drivers
+
+The configuration changes bring ZMK's code in line with how upstream Zephyr sensor drivers handle rotations. This has the added advantage of allowing us to leverage other sensor drivers. On Nordic MCUs, like nRF52840, the NRFX QDEC driver can be used, for example:
```
caksoylar, line 85: I guess this is not strictly a docs question, but it might be good to clarify here: Can you define a "generic" encoder node for a shield like the current implementation, without being nrf-specific? Since we now support RP2040 controllers in pro micro form factor, it might be an issue and a regression compared to the status quo.

If you can still define a generic version, maybe that can be overridden in `boards/nice_nano.overlay` etc. for nRF boards.

docs/blog/2023-06-10-encoder-refactors.md
```diff
@@ -0,0 +1,119 @@
...
+    pinctrl-0 = <&qdec_default>;
+    pinctrl-names = "default";
+};
+```
+
+The NRFX QDEC driver has the advantage of supporting optical encoders as well, and although it polls, it does so in hardware without waking the MCU core; initial basic power profiling is promising.
+
+## Split Sensor/Encoder Support
+
+In addition to the refactors for splitting the configuration, the changes merged included refactors designed to simplify and move forward with the long outstanding feature of supporting encoders on the peripheral side of split keyboards. That work will is planned as a next follow up.
```
caksoylar, line 112:
```suggestion
In addition to the refactors for splitting the configuration, the changes merged included refactors designed to simplify and move forward with the long outstanding feature of supporting encoders on the peripheral side of split keyboards. That work is planned as a follow up.
```

docs/blog/2023-06-10-encoder-refactors.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Deprecation
+
+The old configuration will be supported for a period of one month, and then remove, giving users a grace period to complete the migration to the new separated configuration.
```
caksoylar, line 116:
```suggestion
The old configuration will be supported for a period of one month, and then removed, giving users a grace period to complete the migration to the new separated configuration.
```

docs/blog/2023-06-10-encoder-refactors.md
petejohanson: The *current* `alps,ec11` (which is admittedly a bad compatible name, since quadrature encoders are not specific to alps *at all*) driver is MCU agnostic, and preferable. If you want to use the NRFX QDEC driver for a generic shield, you most definitely would use the `alps,ec11` driver in the base `<shield>.overlay`, and then override in a board specific file.

You can see an example of this in the ZMK Uno shield changes in this PR.

caksoylar: Thanks, that makes sense. I misread the blog post and conflated the update for the generic driver with the *suggestion* of utilizing the NRFX QDEC driver.

app/Kconfig
```diff
@@ -515,8 +533,19 @@ config ZMK_WPM
...
-config SENSOR
+config ZMK_KEYMAP_SENSORS
+    bool "Enable Keymap Sensors support"
     default y
+    depends on DT_HAS_ZMK_KEYMAP_SENSORS_ENABLED
+    select SENSOR
+
+if ZMK_KEYMAP_SENSORS
+
+config ZMK_KEYMAP_SENSORS_DEFAULT_TRIGGERS_PER_ROTATION
```
joelspadin, line 544: What is the difference between this and `ZMK_ENCODERS_DEFAULT_TRIGGERS_PER_ROTATION`? Do we need two different configs (with different defaults) that seem to be for the same thing?

app/include/drivers/behavior.h
```diff
@@ -39,7 +49,8 @@ __subsystem struct behavior_driver_api {
     behavior_keymap_binding_callback_t binding_convert_central_state_dependent_params;
     behavior_keymap_binding_callback_t binding_pressed;
     behavior_keymap_binding_callback_t binding_released;
-    behavior_sensor_keymap_binding_callback_t sensor_binding_triggered;
+    behavior_sensor_keymap_binding_accept_data_callback_t sensor_binding_data;
```
joelspadin, line 52: Minor: should this callback be named "accept data"?

app/src/behaviors/behavior_sensor_rotate_common.c
```diff
@@ -5,48 +5,94 @@
...
+#include <zmk/virtual_key_position.h>
 
 #include "behavior_sensor_rotate_common.h"
 
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
-int zmk_behavior_sensor_rotate_common_trigger(struct zmk_behavior_binding *binding,
-                                              const struct device *sensor,
-                                              struct zmk_behavior_binding_event event) {
+int zmk_behavior_sensor_rotate_common_data(struct zmk_behavior_binding *binding,
```
joelspadin, line 14: As above, should this be "accept data"?

app/src/sensors.c
```diff
@@ -18,65 +18,128 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #if ZMK_KEYMAP_HAS_SENSORS
 
-struct sensors_data_item {
-    uint8_t sensor_number;
+struct sensors_item_cfg {
+    uint8_t sensor_index;
```
joelspadin, line 22: Minor: can move this to the end to maybe avoid unnecessary padding.

app/src/sensors.c
```diff
@@ -18,65 +18,128 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+#define ZMK_KEYMAP_SENSORS_CHILD_COUNT (0 DT_FOREACH_CHILD(ZMK_KEYMAP_SENSORS_NODE, PLUS_ONE))
+#define SENSOR_CHILD_ITEM(node)                                                                    \
+    {                                                                                              \
+        .triggers_per_rotation =                                                                   \
+            DT_PROP_OR(node, triggers_per_rotation,                                                \
+                       DT_PROP_OR(ZMK_KEYMAP_SENSORS_NODE, triggers_per_rotation,                  \
+                                  CONFIG_ZMK_KEYMAP_SENSORS_DEFAULT_TRIGGERS_PER_ROTATION))        \
+    }
+#define SENSOR_CHILD_DEFAULTS(idx, arg)                                                            \
+    { .triggers_per_rotation = DT_PROP_OR(ZMK_KEYMAP_SENSORS_NODE, triggers_per_rotation, 20) }
```
joelspadin, line 46: Should the `20` here be `CONFIG_ZMK_KEYMAP_SENSORS_DEFAULT_TRIGGERS_PER_ROTATION`?

app/src/sensors.c
```diff
@@ -18,65 +18,128 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+    err = sensor_channel_get(item->dev, item->trigger.chan, &value);
+
+    if (err) {
+        LOG_WRN("Failed to get channel data from device %d", err);
+        return;
+    }
+
+    ZMK_EVENT_RAISE(new_zmk_sensor_event(
+        (struct zmk_sensor_event){.sensor_index = item->sensor_index,
+                                  .channel_data = {(struct zmk_sensor_channel_data){
```
joelspadin, line 88: Missing `.channel_data_size = 1`?

docs/blog/2023-06-10-encoder-refactors.md
```diff
@@ -0,0 +1,119 @@
...
+
+### New Configuration
+
+```
+    left_encoder: encoder_left {
+        compatible = "alps,ec11";
+        label = "LEFT_ENCODER";
+        a-gpios = <&pro_micro 21 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
+        b-gpios = <&pro_micro 20 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
+        steps = <20>;
```
joelspadin, line 67: I think this should be 80 to match the original `resolution = <4>` given the `triggers-per-rotation = <20>` below?

app/src/sensors.c
petejohanson: Done.

app/src/sensors.c
petejohanson: Indeed. Fixed.

app/src/sensors.c
petejohanson: Yup. Added.

docs/blog/2023-06-10-encoder-refactors.md
petejohanson: Yeah, good call.

joelspadin (approved):

caksoylar (approved):

PR #1054, opened 2021-12-11
Some light refactoring of the test code 

okke-formsma:
* Rewrote some python code to be more pythonic, and applied 'black' formatter
* Tests now output "FAILED" instead of "FAIL" and "PENDING" instead of "PEND" so it's easier to scan the output when the messages are not all 4 characters long
* Tests now output the results while they're running so screwups can be identified faster locally


petejohanson (approved):
Thanks!

PR #1059, opened 2021-12-13
fix(keymaps): add row-offset property to matrix-transform

### LINKED ISSUE #721: row-offset property missing from matrix transform bindings
https://github.com/zmkfirmware/zmk/blob/a7c6e080a7def74d8bdc9c0b326a911bf5ceec4a/app/src/matrix_transform.c#L29 supports a `row-offset` property, but https://github.com/zmkfirmware/zmk/blob/main/app/dts/bindings/zmk%2Cmatrix-transform.yaml does not define it.

RocketLL:
Closes #721.
Adds `row-offset` property to `matrix-transform`.

First fixed in #735 by @devriesp, but the PR went stale and is not atomic, so submitting a separate fix just for #721.

### DIFF
--- a/app/dts/bindings/zmk,matrix-transform.yaml
+++ b/app/dts/bindings/zmk,matrix-transform.yaml
@@ -13,6 +13,9 @@ properties:
   col-offset:
     type: int
     default: 0
+  row-offset:
+    type: int
+    default: 0
   map:
     type: array
     required: true


joelspadin (approved):

petejohanson (approved):
Thanks!

PR #1062, opened 2021-12-16
Add Osprette to build targets

mrlinuxfish:
Osprette is in master but does not build with the rest of the boards when the build workflow is invoked


Nicell:
@mrlinuxfish We're no longer updating the build list due to the number of merge conflicts it creates. Instead, we're moving towards using shield/board metadata to run workflows through #889

PR #1066, opened 2021-12-21
Stickier keys

snoyer:
This makes sticky keys able to hold for more that one subsequent key press.

The `count` parameter can be used to create `&sk` variations that will stick for a given number of presses. 
The parameter defaults to 1 so the change is backward compatible (does not seem to break any of the existing tests).

Example use cases:
- "two shot mods", VS Code shortcuts such as `CTRL+K+C` have been mentioned on Discord
- unicode macros on Mac which require typing a 4 digits hex code while holding `ALT` (eg. `ALT+1234`)

The main remaining question for me is how to deal with the timer. With multiple expected key presses we need to decide if the timer should be cancelled on the first or last one, it could also be extended after each new one.


snoyer:
Closing as the original use-case is now covered by macros

PR #1079, opened 2021-12-30
Fix typos and inconsistent spellings

Apsu:
Saw some typos when reading one of the doc pages, and ran a spellcheck against the rest. This is an American English dictionary, so it flagged `licence` and `behaviour`.

The former is valid in British English, though still was used incorrectly in that doc; `license` is the verb and `licence` the noun, so `licencing` is less valid.

`behaviour` is the British spelling as well, but since there aren't any others in all the docs, I assumed it should be called out.

If you'd rather these stay as-is, I'll pare this down to just the typos.

### DIFF
--- a/docs/docs/development/build-flash.md
+++ b/docs/docs/development/build-flash.md
@@ -26,7 +26,7 @@ From here on, building and flashing ZMK should all be done from the `app/` subdi
 cd app
 ```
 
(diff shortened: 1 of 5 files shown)


caksoylar (approved):
Looks good to me, I lean towards using US spellings in general, unless there is an objection.

dxmh (approved):

Issue #997, opened 2021-10-22
Feature request: Automatically held/released mods for certain keycodes for Alt-Tab window switching


caksoylar:
This is an attempt to formalize the discussion at https://discord.com/channels/719497620560543766/719565084208398406/900412989033828444 with @bcat and @dxmh.

There are a few use cases around tapping a sequence of keys while another key (usually a modifier) is being held that folks, especially working with small keyboards, have been using with custom code in other firmware like QMK:
  1. Having a single key to do window switching with `Alt+Tab` or `Cmd+Tab`. The usual way this is implemented is with a timer (out of scope here), or first press includes the mod, layer switching and/or pressing/releasing any other key releases the mod.
 The most popular implementation is probably `swapper` feature by Callum Oakley in their [QMK workspace](https://github.com/qmk/qmk_firmware/tree/master/users/callum#swapper). There are other bespoke implementations that are independently developed or extension to multiple keys (`Tab`, `Shift+Tab` on separate keys). Another use case is switching keyboard layouts in Windows/Mac OS with `Ctrl+Space` or `GUI+Space`.
  2. ~~Having a "text selection mode" similar to the visual mode in Vim. In this case usually arrows keys are pressed while Shift is automatically held.~~ This is not a good example since shift doesn't need to stay held between key presses typically, see below comments.

In terms of implementation there can be two approaches:
1. A behavior that can be assigned to a single key (like for `Alt+Tab` window switching). It can keep track of its own modifier state and listen to layer change or key press/release events to release it. It could look something like this in the keymap:
```devicetree
/ {
    behaviors {
        at: alt_tabber {
            compatible = "zmk,mod-holder";
            label = "ALT_TABBER";
            #binding-cells = <1>;
            mods = <MOD_LALT>;
        };
    };
    keymap {
    [...] &at TAB [...]
    }
};
```
2. A keymap node similar to a combo where you specify the modifier, (a list of) layer(s) and a list of key positions. The first press on any of the positions on that layer would press the modifier key, along with triggering the usual behavior of that key. On layer change or another position press/release the modifier is released.
This could look like:
```devicetree
/ {
    modholders {
        compatible = "zmk,mod-holder";
        window_switcher {
            layers = <0>;
            key-positions = <0 1>;
            mods = <MOD_LALT>;
        };
        visual_mode {
            layers = <1>;
            key-positions = <0 1 2 3>;
            mods = <MOD_LSHIFT>;
        };
    };
    keymap {
[...]

PR #1080, opened 2021-12-30
feat(supertab): modkey timeout for rotate key press

lewurm:
_Snippet from documentation diff:_

In addition to the keycodes above a modifier (`modifier-key`) can be configured that is pressed simultaneously with the configured keycodes, and then held until another keycode is send via an encoder action or a timeout (`mod-timeout-ms`) is reached.

Example for application switching:
```
sm: supermods {
    compatible = "zmk,behavior-sensor-rotate-key-press";
    label = "encoder supermod";
    #sensor-binding-cells = <2>;
    modifier-key = <LWIN>;
    mod-timeout-ms = <1200>;
};
[...]
sensor-bindings = <&sm TAB LS(TAB)>;
```
Application switching with rotary encoder

Example for selecting characters in an text editor:
```
hlm: highlightmods {
    compatible = "zmk,behavior-sensor-rotate-key-press";
    label = "encoder highlighter";
    #sensor-binding-cells = <2>;
    modifier-key = <LSHFT>;
    mod-timeout-ms = <500>;
};
[...]
sensor-bindings = <&hlm RIGHT LEFT>;
```
Select characters with rotary encoder

### DIFF
--- a/app/dts/bindings/behaviors/zmk,behavior-sensor-rotate-key-press.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-sensor-rotate-key-press.yaml
@@ -13,6 +13,12 @@ properties:
     type: int
     required: true
     const: 2
+  modifier-key:
+    type: int
+    default: -1
+  mod-timeout-ms:
+    type: int
+    default: -1
 
 sensor-binding-cells:
   - param1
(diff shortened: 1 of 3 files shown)


caksoylar:
This seems pretty useful! Related: #997

Regarding documentation, I am curious if we should have a policy about including gifs. In my opinion it might be too distracting (and also cause large size page loads). @zmkfirmware/docs what do you think?

lewurm:
> This seems pretty useful! Related: #997

Thanks, I haven't seen this yet. Reading that, it makes more sense to split the idea of this PR into two separate things:

1. As suggested in #997, implement another behavior that has this "mod holder" functionality.
2. Add a new behavior for`sensor-bindings` such that not only the `&kp` behavior can be used, but any arbitrary one. Apart from the suggested "mod holder" behavior I would like to be able to map `&rgb_ub` to the rotatry encoders (mentioned in #65 ).

Of course, I don't know how to actually implement those nicely 😉  I'll keep this PR open for visibility, so if someone wants the funcitionality today it can be cherry-picked, but it should probably not be merged as-is.

PR #1082, opened 2022-01-03
Bump init delay to 50ms for nico-nano

midgethoen:
Bump init delay to 50ms for nico-nano to prevent issues with OLED not being found. i.e.:

```
[00:00:00.303,375] <err> zmk: Failed to find display device
```


midgethoen:
conversation for reference: https://discord.com/channels/719497620560543766/719544230497878116/927252960306679808

Nicell (approved):
Thanks for this change. This should allow ample time for the VCC pin to get power out to the external peripherals like OLEDs, and it makes startup time 40ms longer, which is a fine tradeoff IMO.

PR #1084, opened 2022-01-05
fix(keymaps): Handle matching `then-layer`s.

petejohanson:
* Proporly handle multiple  conditonal layers w/ the same target
  `then-layer` values.
* Move handling to work callback, to avoid re-entrance for cascading
  layers enabling other layers.

### DIFF
--- a/app/src/conditional_layer.c
+++ b/app/src/conditional_layer.c
@@ -7,6 +7,7 @@
 #define DT_DRV_COMPAT zmk_conditional_layers
 
 #include <stdint.h>
+#include <kernel.h>
 
 #include <devicetree.h>
 #include <logging/log.h>
@@ -18,6 +19,8 @@
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
+
+static K_SEM_DEFINE(conditional_layer_sem, 1, 1);
 
 // Conditional layer configuration that activates the specified then-layer when all if-layers are
 // active. With two if-layers, this is referred to as "tri-layer", and is commonly used to activate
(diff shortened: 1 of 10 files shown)


petejohanson:
@bcat Tagging you here, as the original author.

Thoughts?

caksoylar:
I see different behavior with this patch when using a single tri-layer accessed through layer-tap keys. What I see is the following: hold down two keys (with layer-taps going to `if-layers`) then immediately tap a key twice. The first tap always produces the keycode from the one of the `if-layers`, second tap produces the keycode from the then-layer.

Here is a patch adding a new test case `tri-layer-lt` (which is just `tri-layer` with balanced `&lt` instead of `&mo`) that passes with `main` and fails with this branch:
```patch
diff --git a/app/tests/conditional-layer/tri-layer-lt/events.patterns b/app/tests/conditional-layer/tri-layer-lt/events.patterns
new file mode 100644
index 0000000..14ded79
--- /dev/null
+++ b/app/tests/conditional-layer/tri-layer-lt/events.patterns
@@ -0,0 +1,3 @@
+s/.*mo_keymap_binding/mo/p
+s/.*conditional_layer/cl/p
diff --git a/app/tests/conditional-layer/tri-layer-lt/keycode_events.snapshot b/app/tests/conditional-layer/tri-layer-lt/keycode_events.snapshot
new file mode 100644
index 0000000..b23e61b
--- /dev/null
+++ b/app/tests/conditional-layer/tri-layer-lt/keycode_events.snapshot
@@ -0,0 +1,8 @@
+mo_pressed: position 2 layer 1
+mo_pressed: position 3 layer 2
+cl_activate: layer 3
+kp_pressed: usage_page 0x07 keycode 0x0a implicit_mods 0x00 explicit_mods 0x00
+kp_released: usage_page 0x07 keycode 0x0a implicit_mods 0x00 explicit_mods 0x00
+mo_released: position 3 layer 2
+cl_deactivate: layer 3
+mo_released: position 2 layer 1
diff --git a/app/tests/conditional-layer/tri-layer-lt/native_posix.keymap b/app/tests/conditional-layer/tri-layer-lt/native_posix.keymap
new file mode 100644
index 0000000..7a09160
--- /dev/null
+++ b/app/tests/conditional-layer/tri-layer-lt/native_posix.keymap
@@ -0,0 +1,56 @@
+#include <behaviors.dtsi>
+#include <dt-bindings/zmk/kscan_mock.h>
+
+&lt {
+    flavor = "balanced";
+};
+
+/ {
+    conditional_layers {
+        compatible = "zmk,conditional-layers";
+        tri_layer {
+            if-layers = <1 2>;
+            then-layer = <3>;
+        };
+    };
+
+    keymap {
+        default_layer {
+            bindings = <
+                &kp A &kp B
+                &lt 1 I &lt 2 J
+            >;
+        };
+        layer_1 {
+            bindings = <
+                &kp C &kp D
+                &trans &trans
+            >;
+        };
+        layer_2 {
+            bindings = <
+                &kp E &kp F
+                &trans &trans
+            >;
+        };
+        layer_3 {
+            bindings = <
+                &kp G &kp H
+                &trans &trans
+            >;
+        };
+    };
+};
+
+&kscan {
+    events = <
+        ZMK_MOCK_PRESS(1,0,10)
+        ZMK_MOCK_PRESS(1,1,10)
+        ZMK_MOCK_PRESS(0,0,10)
+        ZMK_MOCK_RELEASE(0,0,10)
+        ZMK_MOCK_RELEASE(1,1,10)
+        ZMK_MOCK_RELEASE(1,0,10)
+    >;
+};
```

bcat:
Sorry for the delay in looking at this! I honestly didn't consider multiple conditional layer configs sharing the same `then-layer`. In retrospect, yeah, that won't work today. I'll take a look at the actual changes as soon as I can, but in general, I don't see any reason to object to supporting this. :)

app/tests/conditional-layer/same-layer-reached-differently/native_posix.keymap
```diff
@@ -0,0 +1,69 @@
...
+        layer_4 {
+            bindings = <
+                &kp E &trans
+                &trans &trans
+            >;
+        };
+    };
+};
+
+&kscan {
```
bcat, line 53: This behavior LGTM. Do you mind also adding a test case where all three `&mo` bindings are held at the same time, causing layer 4 to be activated through both the `<1 2>` and `<1 3>` configs?

app/src/conditional_layer.c
```diff
@@ -68,20 +68,40 @@ static void conditional_layer_deactivate(int8_t layer) {
...
-static int layer_state_changed_listener(const zmk_event_t *ev) {
+static void update_layer_state_work_handler(struct k_work *work) {
+    int8_t max_then_layer = -1;
+    uint32_t then_layers = 0;
+    uint32_t then_layer_state = 0;
     for (int i = 0; i < NUM_CONDITIONAL_LAYER_CFGS; i++) {
         const struct conditional_layer_cfg *cfg = CONDITIONAL_LAYER_CFGS + i;
         zmk_keymap_layers_state_t mask = cfg->if_layers_state_mask;
+        then_layers |= BIT(cfg->then_layer);
+        max_then_layer = MAX(max_then_layer, cfg->then_layer);
```
bcat, line 80: I feel like there's probably a macro-y way to compute `then_layers` and `max_then_layer` as compile-time constants, but maybe that's overkill?

bcat (approved):

bcat:
I mean, this looks fine to me, but I don't really know enough about work queues or how ZMK uses them to know what to do about the layer tap race condition @caksoylar described....

app/src/conditional_layer.c
petejohanson: I'd say overkill for what we're doing.

petejohanson:
> ```diff
> +s/.*hid_listener_keycode/kp/p
> +s/.*mo_keymap_binding/mo/p
> +s/.*conditional_layer/cl/p
> ```

Thanks for the test case! Added the test case, and implemented the re-entrancy handling without using the work queue, which makes the processing *immediate* but also avoids re-entrance.

Test now passes, and I've added it as part of this PR.

@bcat Thanks for the review! See my approach to handling the processing stuff.

caksoylar:
Thank you for fixing it! I tested on real hardware as well and it now works as expected with layer-taps.

bcat:
Awesome! I'm super happy this is useful and y'all are making it way better. :)

@petejohnson: Maybe it got lost in the shuffle, but I do think it'd be useful to add an extra test next to `same-layer-reached-differently` verify that having both configs activate simultaneously (by holding down all three `&mo` keys at the same time) still works as expected. (Unless there's already a test for that case and I've just forgotten.)

petejohanson:
> Awesome! I'm super happy this is useful and y'all are making it way better. :)
> 
> @petejohnson: Maybe it got lost in the shuffle, but I do think it'd be useful to add an extra test next to `same-layer-reached-differently` to verify that having both configs activate simultaneously (by holding down all three `&mo` keys at the same time) still works as expected. (Unless there's already a test for that case and I've just forgotten.)

Hadn't forgotten, just haven't had a chance to add it.

> 
> Also, +1 to the semaphore to deal with reentrancy. I am (still) not really familiar with the actual processing mode of ZMK/Zephyr, but it makes sense reading the code as a ZMK novice. :)

Great!

petejohanson:
@bcat Added as `tests/conditional-layer/same-layer-reached-both-ways`.

PR #1087, opened 2022-01-11
Fix ext-power in Mikoto board definition

zhiayang:
Fix the devicetree for the mikoto board definition. If my understanding of `GPIO_ACTIVE_HIGH` is correct (ie. it follows the name), then to enable external power, the pin should be driven high. This follows the schematic here:

image

The current active-low definition was probably an oversight from the initial commit. Anyway, I also added the `init-delay-ms` line from #1082, since whatever applies to the n!n probably applies here as well.

Disclaimer: I don't use zmk, and I don't have any spare boards to test with right now. However, this *is* my board, so I feel a little responsibility to fix it.

### DIFF
--- a/app/boards/arm/mikoto/mikoto_520.dts
+++ b/app/boards/arm/mikoto/mikoto_520.dts
@@ -29,7 +29,8 @@
 	ext-power {
 		compatible = "zmk,ext-power-generic";
 		label = "EXT_POWER";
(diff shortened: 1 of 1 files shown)


petejohanson (approved):
Yup, this makes sense. Apologies for the oversight there, I don't have hardware on which to test this, but that all looks right, and aligns w/ the `nice_nano_v2` definition, which has a similar schematic.

PR #1095, opened 2022-01-20
GPIO direct driver bugfix

megamind4089:
For POLLING mode gpio direct driver, debounce value is not considered on timer_handler to schedule a delayed work

## Board/Shield Check-list

### DIFF
--- a/app/drivers/kscan/kscan_gpio_direct.c
+++ b/app/drivers/kscan/kscan_gpio_direct.c
@@ -111,8 +111,9 @@ static void kscan_gpio_irq_callback_handler(const struct device *dev, struct gpi
 
 static void kscan_gpio_timer_handler(struct k_timer *timer) {
     struct kscan_gpio_data *data = CONTAINER_OF(timer, struct kscan_gpio_data, poll_timer);
+    const struct kscan_gpio_config *cfg = data->dev->config;
 
-    kscan_gpio_direct_queue_read(&data->work, 0);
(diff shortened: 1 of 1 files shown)


app/drivers/kscan/kscan_gpio_direct.c
```diff
@@ -111,8 +111,9 @@ static void kscan_gpio_irq_callback_handler(const struct device *dev, struct gpi
 
 static void kscan_gpio_timer_handler(struct k_timer *timer) {
     struct kscan_gpio_data *data = CONTAINER_OF(timer, struct kscan_gpio_data, poll_timer);
+    const struct kscan_gpio_config *cfg = data->dev->config;
 
-    kscan_gpio_direct_queue_read(&data->work, 0);
+    kscan_gpio_direct_queue_read(&data->work, cfg->debounce_period);
```
petejohanson, line 116: I'm not sure this is 100% right. For polling mode we have a fixed timer if every 10ms, which them fires this callback. Until we.have the more nuanced debounce work from @joelspadin perhaps that timer interval should honor debounce period instead?

petejohanson (commented):
One question. Thanks for looking at this.

app/drivers/kscan/kscan_gpio_direct.c
megamind4089: Ah okie. I will close this now if there is advanced debounce is on the work

PR #1097, opened 2022-01-24
fix(ble): Restore BLE SC passkey entry for pairing.

petejohanson:
* Handle capturing numeric inputs while pairing and sending
  final passkey once six digits entered.

This restores the basic functioning of the BLE Secure Connect numeric comparison by allowing the user to type the digits displayed on the host into their keyboard during pairing. This secures against active MITM attacks during pairing.

To test, enable the `CONFIG_ZMK_BLE_PASSKEY_ENTRY=y` config and rebuild.

So far, I've tested w/ Fedora, Android 12, and macOS Catalina.

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -36,9 +36,23 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/event_manager.h>
 #include <zmk/events/ble_active_profile_changed.h>
 
+#define IS_HOST_PERIPHERAL                                                                         \
+    (!IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL))
+#define IS_SPLIT_PERIPHERAL                                                                        \
+    (IS_ENABLED(CONFIG_ZMK_SPLIT) && !IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL))
+
+#define DO_PASSKEY_ENTRY (IS_ENABLED(CONFIG_ZMK_BLE_PASSKEY_ENTRY) && !IS_SPLIT_PERIPHERAL)
+
+#if DO_PASSKEY_ENTRY
+#include <zmk/events/keycode_state_changed.h>
+
+#define PASSKEY_DIGITS 6
+
 static struct bt_conn *auth_passkey_entry_conn;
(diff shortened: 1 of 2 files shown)


app/src/ble.c
```diff
@@ -612,32 +622,69 @@ int zmk_ble_unpair_all() {
...
+    }
+
+    if (key == HID_USAGE_KEY_KEYBOARD_ESCAPE) {
+        bt_conn_auth_cancel(auth_passkey_entry_conn);
+        return ZMK_EV_EVENT_HANDLED;
     }
 
-    if (key < NUMBER_1 || key > NUMBER_0) {
-        return true;
+    if (key < HID_USAGE_KEY_KEYBOARD_1_AND_EXCLAMATION ||
```
caksoylar, line 646: Is it worth checking for numpad digits as well? Not a big deal either way, but I can see folks trying to use them on a 100% keyboard.

petejohanson: Perhaps. The host OS doesn't treat them as numeric unless caps lock is enabled, but I suppose we could ignore that requirement for our use.

petejohanson: @caksoylar Added.

petejohanson:
Pushed some fixes for splits.

adamcstephens:
I was able to successfully test this on MacOS 12.2, iOS 15.3 and Archlinux. After forgetting the previous pairing on both ends, all three prompted for a passcode and paired/connected.

On Linux I did have to reset the primary side of the keyboard after pairing in order for it to connect. Not sure if this is a Linux issue or not. After the reset everything works as expected.

petejohanson:
> On Linux I did have to reset the primary side of the keyboard after pairing in order for it to connect. Not sure if this is a Linux issue or not. After the reset everything works as expected.

@adamcstephens Have you encountered that 'reset to connect' issue previously?

adamcstephens:
> > On Linux I did have to reset the primary side of the keyboard after pairing in order for it to connect. Not sure if this is a Linux issue or not. After the reset everything works as expected.
> 
> @adamcstephens Have you encountered that 'reset to connect' issue previously?

@petejohanson not that I recall, but it's a rare occurrence for me to have to pair and it has been many months since the last time I did so.

Nicell:
Tested on Windows and Android. Windows worked flawlessly and quickly. Android I always have to click the device twice in the device list for some reason. The first press fails right away, and the second press works as expected, but then it takes a few seconds to fully pair after typing.

Nicell (approved):

petejohanson:
Ok, merging, we can continue to evaluate what's up w/ Android pairing.

PR #1098, opened 2022-01-24
fix(display): Initialize display on queue as well.

petejohanson:
A hopeful fix for slow initialization of devices with displays with slow response times, by moving the init to the display queue, which may be a secondary display queue, to free up system queue for handling inputs right away.

### DIFF
--- a/app/src/display/main.c
+++ b/app/src/display/main.c
@@ -85,26 +85,20 @@ static void stop_display_updates() {
 
 int zmk_display_is_initialized() { return initialized; }
 
-int zmk_display_init() {
+void initialize_display(struct k_work *work) {
     LOG_DBG("");
 
     display = device_get_binding(ZMK_DISPLAY_NAME);
     if (display == NULL) {
         LOG_ERR("Failed to find display device");
-        return -EINVAL;
+        return;
     }
-
-#if IS_ENABLED(CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED)
-    k_work_queue_start(&display_work_q, display_work_stack_area,
-                       K_THREAD_STACK_SIZEOF(display_work_stack_area),
-                       CONFIG_ZMK_DISPLAY_DEDICATED_THREAD_PRIORITY, NULL);
-#endif
 
     screen = zmk_display_status_screen();
 
     if (screen == NULL) {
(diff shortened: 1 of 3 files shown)


app/src/display/main.c
```diff
@@ -85,33 +85,41 @@ static void stop_display_updates() {
...
 
     start_display_updates();
 
     initialized = true;
+}
+
+K_WORK_DEFINE(init_work, initialize_display);
+
+int zmk_display_init() {
+    LOG_DBG("");
```
Nicell, line 114: What are the blank debug messages used for?

Nicell (commented):
I can't test this since I don't have a keyboard with a display right now, but the reasoning and code look sane!

mloffer:
This works on the Cornish Zen.

petejohanson:
Just pushed a new version that moves the display clearing/initialization into the unblank callback of the driver, keeping it from blocking the actual *driver* init, and we are moving the unblank to a dedicated queue, so this frees up system work queue during that initial display clear/initialization that was hanging things up previously.

Ready for re-test.

caksoylar:
I had commented on Discord but I have been using this for a day and thought I would comment here as well.

I tested this on the Corne-ish Zen with e-paper displays: The startup delay from reset or wake from deep sleep is much quicker with this change, whereas the previous version of the PR hadn't helped. I used to measure about 8 seconds from when central half is reset to sending keystrokes over USB and with this version adapted to Zen's current branch, I measure about 2 seconds. About a similar speed up seems to apply to waking from deep sleep as well.

app/src/display/main.c
petejohanson: Silliness! Removed.

danielebra:
A much needed improvement! Thanks for putting this together 🚀

caksoylar:
I had a chance to test this with the Corne-ish Zen rebase on ZMK `main` as well (where the patch without additional modifications). I still see the same benefit, where delay from central-side reset to first keypress is about 2 seconds on USB output, and 3-4 over BLE. Do we need confirmation from anyone else to merge, like with a different display?

petejohanson (commented):
Tested on ZMK Uno w/ an SSD1306 with both system work queue and dedicated work queue, both start and run just fine.

PR #1099, opened 2022-01-26
feat(behaviors): AdD `&sarcasm` BehAviOR.

petejohanson:
New behavior, inspired by https://www.reddit.com/r/funny/comments/scykg9/i_built_a_sarcasm_converter_for_my_keyboard_to/

Toggled on/off w/ `&sarcasm`

WORKs LIke tHis wheN aCTivE.

Cribbed most of the `&caps_word` logic for this.

See https://youtu.be/wfsQ4tNFS8I for it in action.

### DIFF
--- a/app/dts/behaviors/sarcasm.dtsi
+++ b/app/dts/behaviors/sarcasm.dtsi
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2021 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <dt-bindings/zmk/keys.h>
+
+/ {
+	behaviors {
+		/omit-if-no-ref/ sarcasm: behavior_sarcasm {
+			compatible = "zmk,behavior-sarcasm";
+			label = "SARCASM";
+			#binding-cells = <0>;
+		};
+	};
+};
+
--- a/app/dts/bindings/behaviors/zmk,behavior-sarcasm.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-sarcasm.yaml
@@ -1,0 +1,8 @@
+# Copyright (c) 2021 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+description: SaRcASm BeHaViOr
+
+compatible: "zmk,behavior-sarcasm"
+
+include: zero_param.yaml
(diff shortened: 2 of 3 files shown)


petejohanson:
jUsT puSHED a fiX tO nOt TrIggEr wHEN ExplIciT MODs ARe HeaLd, so SHOrTcutS sTilL WIll work.

AlexCzar:
This is pure evil.

petejohanson:
> This is pure evil.

tHis is pUre eViL

ReFil:
I cannot emphasise how much i want this

Percentnineteen:
Man....I literally implemented the same thing today but worried it was too dumb to make a pr

petejohanson:
> Man....I literally implemented the same thing today but worried it was too dumb to make a pr

I am definitely not convinced this should be merged to `main`, but it was too fun to pass up sharing.

How was it to implement for you?

Percentnineteen:
> > Man....I literally implemented the same thing today but worried it was too dumb to make a pr
> 
> I am definitely not convinced this should be merged to `main`, but it was too fun to pass up sharing.
> 
> How was it to implement for you?

Not too bad, once I wrapped my head around what was happening in the behavior definitions. I haven't written c in at least a dozen years.

Like you, I also based it off of the caps_word but stored an extra state instead of your mod(2) approach. Probably less "good" but it also stayed closer to caps_word (i.e. automatically toggle off on non-alphanumeric characters that aren't in a user modifiable permitted list that defaults to space, underscore and minus)

The biggest problem I faced was that I didn't realize I had to modify CMakeList.txt.

petejohanson:
> 
> Not too bad, once I wrapped my head around what was happening in the behavior definitions. I haven't written c in at least a dozen years.

This was me when I started ZMK. It does come back to you.

> 
> Like you, I also based it off of the caps_word but stored an extra state instead of your mod(2) approach. Probably less "good" but it also stayed closer to caps_word (i.e. automatically toggle off on non-alphanumeric characters that aren't in a user modifiable permitted list that defaults to space, underscore and minus)
> 

I wasn't sure if the auto-off like caps word would be best or not... Maybe you're meme-ing hard and want to keep it on for a while ...


> Oh, I called mine "sponge lock" in honor of the SpongeBob memes that I always associate with the text style..

Much better name!

PR #1113, opened 2022-02-03
Feature: Add setting to only allow external power while on USB power

berfarah:
Inspired by these PRs:
https://github.com/zmkfirmware/zmk/pull/904
https://github.com/zmkfirmware/zmk/pull/1010

In order to prevent battery drain from external screens, backlights or underglow, allow users to use a setting to block power usage when not connected to USB power.

--

A few notes about this work that may help add context:
- I'm honestly not sure if the file I put this logic in is where this belongs. I don't fully understand the power management system - I think this would play nice, but not confident.
- I'm happy to add an disable setting similar to the PRs I mention above - would love to get feedback on whether this is the right direction before I do that
- For testing: my batteries died, which is actually what motivated me to make this PR. I'll be looking to replace my batteries, but in the meantime I'd love any help testing this I can get.
- I'm not super well versed with C, so please feel free to tell me if I'm doing something wrong / non-idiomatic


petejohanson (commented):
It's an interesting idea... The implementation is a partial solution, we'd still need the underglow to react to USB events or we'll still have animations running, etc.

Also, seems odd to put this in the generic ext power driver... Perhaps a better place is the general USB code, or in an additional dedicated handler glue code that listens for the events and toggles the ext-power "from outside".

Thanks for exploring this!

PR #778, opened 2021-05-02
feat(mouse): Added mouse emulation

krikun98:
First of all, I'd like to thank @dtsykunov - this feature was our hackathon project.

Mouse button clicks, movement and scrolling from your keymap! There's even documentation available! This is still in alpha, though. Further development needs some architectural decisions, and some bugs need to be ironed out.

- Mouse movement over BLE is ~~very~~ (I've tried it, and it's actually usable. A MacOS tester on Discord reports not noticing a difference to USB) choppy. It's smooth at https://github.com/zmkfirmware/zmk/commit/bd108c5d9027bff194545d5c7fa1607fdfb2e24b, but it has frequent unpredictable deadlocks there, so it's unusable. 
- The binding format needs to be decided on. Keypresses are fine, but the movement/scroll speeds would ideally be user-adjustable, and that requires wrangling signed/unsigned integers (which is tricky with DeviceTree) and deciding on the granularity (0-100 would lose a lot of the available precision).
- Currently, only mouse movement is handled via events. Ideally, all three behaviors would be, and the event structure needs to be decided on. Should all three mouse behaviours (click, movement, wheel) be consolidated into a single event and go straight into the report, or shoud there be three separate events that will go into the same report (easier to handle "hold-to-repeat" granularly)?
- On that point, try scrolling while holding a mouse movement button. It's even faster than an MX Master.
- The whole additive system with mouse movements and scrolling was basically thrown together in five minutes and should be re-examined. ~~I have a feeling it should be rewritted to `int16_t`s and overflow checks should be implemented there.~~ The rewrite is done, and the usable value range is small enough that I don't think overflow checks are necessary. Still may need testing, though.
- Mouse click code is also basically copied from modifiers. I'm pretty sure it's fine that way, but some parts may be excessive. Buttons 9-16 also don't work for some reason.
- Should movement press repeat timers be made user-configurable?  I would imagine it'd be similar to mouse sensitivity adjustments.
- Some thought should also be given to future hardware support implementation. "hold-to-repeat" is not needed for actual mice, they also don't tend to "press" or "release" anything. By the way, if anyone can send a PS/2 trackpoint to Russia - we can work on integrating that too.
- Mouse actions should also be mappable to encoders.

I'll add to the list if I can think of anything else. In the meantime, testing and feedback would be very appreciated.

PR #1116, opened 2022-02-03
add: Cirque Pinnacle trackpad driver

crides:
This adds the driver for Cirque Pinnacle trackpads. Note this is _only the driver_, support for fetching the data and binding behaviours to it should be added later as it depends on reworking the sensor subsystem.

The driver uses only the SPI bus and the `dr` data ready pin if available.

The implementation of the driver follows the [Cirque documentation](https://www.mouser.com/pdfDocs/InterfaceSPIandI2C.pdf). While the docs mentions using the sleep mode with high SPI speeds can be a problem, this is not observed in reality (I'm using a SPI clock of 10MHz).


crides:
For transparency, I've been using this driver on my Fusion board for a while, and it seems to work fine for the most part, but sometimes it can cause weird bugs. I'm only starting to debug this, and I have no idea why (so it could be related to my driver or not) but when it happens it just messes up the screen and the system freezes.

crides:
Okay, so this is weird... Cirque has some interesting [data format](https://www.mouser.com/pdfDocs/InterfaceSPIandI2C.pdf) for the relative mode (which is all I use; see page 10). It would be helpful if others can help me see if there's some interpretation problem here or it's just cirque being weird.

So sometimes the driver will output some spuriously large data, and here's some of the instances. `handle_trackpad` just prints out the signed delta X, delta Y and the button flags, and we only need to focus on the deltas. The `pak` logs the raw packet data (`PacketByte_0` to `PacketByte_2` inclusive) if the deltas are too big, and is for the line _below_. In these testings I just move my finger roughly in a circle.

```
[00:04:13.368,774] <dbg> trackpad: handle_trackpad: trackpad -3 13 00
[00:04:13.383,758] <dbg> trackpad: handle_trackpad: trackpad 0 10 00
[00:04:13.398,742] <dbg> trackpad: handle_trackpad: trackpad 4 6 00
[00:04:13.413,757] <dbg> trackpad: handle_trackpad: trackpad 4 6 00
[00:04:13.428,741] <dbg> trackpad: handle_trackpad: trackpad 9 6 00
[00:04:13.443,725] <dbg> trackpad: handle_trackpad: trackpad 9 6 00
[00:04:13.458,740] <dbg> trackpad: handle_trackpad: trackpad 12 5 00
[00:04:13.473,724] <wrn> pinnacle: pak 0x08 0c ff
[00:04:13.473,724] <dbg> trackpad: handle_trackpad: trackpad 12 255 00
[00:04:13.488,739] <dbg> trackpad: handle_trackpad: trackpad 12 -1 00
[00:04:13.503,723] <dbg> trackpad: handle_trackpad: trackpad 12 -6 00
[00:04:13.518,707] <dbg> trackpad: handle_trackpad: trackpad 7 -6 00
[00:04:13.533,721] <dbg> trackpad: handle_trackpad: trackpad 1 -9 00
[00:04:13.563,690] <dbg> trackpad: handle_trackpad: trackpad -3 -5 00
[00:04:13.578,704] <dbg> trackpad: handle_trackpad: trackpad -11 -5 00
```

```
[00:03:56.470,428] <dbg> trackpad: handle_trackpad: trackpad 7 3 00
[00:03:56.480,072] <dbg> trackpad: handle_trackpad: trackpad 6 2 00
[00:03:56.480,346] <dbg> trackpad: handle_trackpad: trackpad 6 2 00
[00:03:56.488,891] <dbg> trackpad: handle_trackpad: trackpad 5 1 00
[00:03:56.489,135] <dbg> trackpad: handle_trackpad: trackpad 5 1 00
[00:03:56.499,267] <dbg> trackpad: handle_trackpad: trackpad 6 0 00
[00:03:56.514,190] <dbg> trackpad: handle_trackpad: trackpad 6 0 00
[00:03:56.529,205] <wrn> pinnacle: pak 0x08 08 fc
[00:03:56.529,205] <dbg> trackpad: handle_trackpad: trackpad 8 252 00
[00:03:56.529,815] <dbg> trackpad: handle_trackpad: trackpad 16 -4 00
[00:03:56.539,001] <dbg> trackpad: handle_trackpad: trackpad 15 -5 00
[00:03:56.539,245] <dbg> trackpad: handle_trackpad: trackpad 15 -5 00
[00:03:56.549,530] <dbg> trackpad: handle_trackpad: trackpad 12 -6 00
[00:03:56.549,774] <dbg> trackpad: handle_trackpad: trackpad 12 -6 00
[00:03:56.559,173] <dbg> trackpad: handle_trackpad: trackpad 11 -7 00
[00:03:56.574,188] <dbg> trackpad: handle_trackpad: trackpad 9 -8 00
```

```
[00:03:48.185,211] <dbg> trackpad: handle_trackpad: trackpad 19 8 00
[00:03:48.193,664] <dbg> trackpad: handle_trackpad: trackpad 24 7 00
[00:03:48.193,939] <dbg> trackpad: handle_trackpad: trackpad 24 7 00
[00:03:48.206,878] <dbg> trackpad: handle_trackpad: trackpad 25 4 00
[00:03:48.221,862] <dbg> trackpad: handle_trackpad: trackpad 24 2 00
[00:03:48.236,846] <wrn> pinnacle: pak 0x08 14 ff
[00:03:48.236,877] <dbg> trackpad: handle_trackpad: trackpad 20 255 00
[00:03:48.237,457] <dbg> trackpad: handle_trackpad: trackpad 17 -1 00
[00:03:48.240,661] <dbg> trackpad: handle_trackpad: trackpad 12 -3 00
[00:03:48.240,905] <dbg> trackpad: handle_trackpad: trackpad 12 -3 00
[00:03:48.251,708] <dbg> trackpad: handle_trackpad: trackpad 7 -3 00
[00:03:48.252,014] <dbg> trackpad: handle_trackpad: trackpad 7 -3 00
[00:03:48.266,845] <dbg> trackpad: handle_trackpad: trackpad 4 -6 00
[00:03:48.281,860] <dbg> trackpad: handle_trackpad: trackpad 0 -11 00
```

petejohanson:
@crides Sure seems to be at the "transition point" from positive to negative values for the delta. Is there a moment when the sign bit and the value are "out of sync" perhaps?

crides:
Hmm that is very likely.. But the datasheet/comm format doesn't specify anything at all about syncing data, and I'm assuming auto-inc reads should make sure that it doesn't read across update epochs. Seems like something worth asking Cirque sales

MikaelElkiaer:
I am building a [ffkb](https://fingerpunch.xyz/product/faux-fox-keyboard-v2/) with a Nice!nano and a Cirque trackpad, so I could help trying this out - if I could just get a few pointers on how to get started. :)

crides:
@MikaelElkiaer Do you have any specific directions you want to look at? If you just wanna use it you should just merge this PR, the [mouse PR](https://github.com/zmkfirmware/zmk/pull/778), and then some version of the example "configuration" code linked in the start. Look around at the configs in the `fusion` board.

MikaelElkiaer:
> @MikaelElkiaer Do you have any specific directions you want to look at? If you just wanna use it you should just merge this PR, the [mouse PR](https://github.com/zmkfirmware/zmk/pull/778), and then some version of the example "configuration" code linked in the start. Look around at the configs in the `fusion` board.

Yeah, I just wanted to try it out and use it.
So this branch is not built on top of the mouse changes?
Thanks for some useful info.

drzony:
@crides The fetching is incorrect in your case. The sign is not in the data[1] and data[2] it's in data[0]. So:
```cpp
    if ((data[0] & 0x10) && data[1] != 0) {
        xDelta = -((int16_t)256 - (int16_t)(data[1]));
    } else {
        xDelta = data[1];
    }
    if ((data[0] & 0x20) && data[2] != 0) {
        yDelta = ((int16_t)256 - (int16_t)(data[2]));
    } else {
        yDelta = -((int16_t)data[2]);
    }
    wheelCount = ((int8_t*)data)[3];
```

crides:
I know that. The problem is that the sign bits and the data bytes may not be synced, and cause weird problems. I've contacted cirque about it, never got a response

drzony:
Weird, I'm running relative mode on QMK (just got it merged into develop) and I don't seem to have this problem. (The code above is taken from my test code not QMK, just to be on safe side in terms of licensing, I ported it to QMK afterwards). Maybe try running QMK to compare results (might be some HW issue). Also there is SW_DATA_READY bit, maybe it's worth checking it after hardware pin is asserted.

drzony:
Also maybe I'm missing something in the code, but the SW_DR flag in REG 0x02 should be cleared after every read.

crides:
I'm not sure why one would clear DR when the trigger is not set, cuz the DR status is not used anyway

drzony:
From the spec:
```
To interrupt or alert the host when data is ready, Pinnacle uses a Hardware Data Ready (HW_DR) signal (active HIGH) that is triggered by either the Command Complete (SW_CC) flag or the Software Data Ready (SW_DR) flag in the Status1 register address 0x02. When either software flag is set, the HW_DR (pin 36 PA2) is also asserted. The SW_DR and SW_CC flags must be manually cleared by the host. The HW_DR signal remains asserted while either SW_DR or SW_CC is asserted
```
So my understanding is that SW_DR flag is set by the chip when data is available, this causes HW_DR to be asserted. To clear HW_DR you need to clear both SW_CC and SW_DR flags. I do not have a pin available for use with HW_DR, so I'm using SW_DR to check for data, and then clear it after each read. I'm saying that this might be the cause for out-of-sync data.

crides:
Added the I2C part of the driver. Both the I2C and SPI versions use the same devicetree `compatible` instance, just on different buses.

app/drivers/sensor/cirque_trackpad/Kconfig
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2022 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+menuconfig PINNACLE
+	bool "PINNACLE Incremental Encoder Sensor"
+	depends on GPIO
+	depends on SPI
```
petejohanson, line 7:
```suggestion
	depends on SPI || I2C
```

app/drivers/sensor/cirque_trackpad/Kconfig
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2022 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+menuconfig PINNACLE
```
petejohanson, line 4: Ideally the folder/file naming should match the Kconfig naming... Which is the authoritative/canonical name for this hardware?

app/drivers/sensor/cirque_trackpad/cirque_trackpad.c
```diff
@@ -0,0 +1,256 @@
...
+    .invert_x = DT_INST_PROP(0, invert_x),
+    .invert_y = DT_INST_PROP(0, invert_y),
+    .sleep_en = DT_INST_PROP(0, sleep),
+    .no_taps = DT_INST_PROP(0, no_taps),
+#ifdef CONFIG_PINNACLE_TRIGGER
+    .dr = GPIO_DT_SPEC_GET(DT_DRV_INST(0), dr_gpios),
+#endif
+};
+
+DEVICE_DT_INST_DEFINE(0, pinnacle_init, device_pm_control_nop, &pinnacle_data, &pinnacle_config, POST_KERNEL, CONFIG_SENSOR_INIT_PRIORITY, &pinnacle_driver_api);
```
petejohanson, line 256: Do we want to assume from the get go that we only want to support one instance of this driver?

Seems like a bit of a risk.

petejohanson (commented):
Thanks for the driver work! A few comments/questions for a first pass.

app/drivers/sensor/cirque_trackpad/Kconfig
crides: I'm not sure, which is why it's inconsistent. The manufacturer is Cirque, the product is called Glide Point (circle trackpad; though nobody seems to refer it by that name), and the ASIC used on the version sold at Mouser is Pinnacle. There are newer versions of the ASIC, not sure if they are sold in a complete product (so I'm not sure if Glide Point is specific enough, if we want to settle on a name).

app/drivers/sensor/cirque_trackpad/cirque_trackpad.c
```diff
@@ -0,0 +1,256 @@
+#define DT_DRV_COMPAT cirque_pinnacle
+
+#include <init.h>
+#include <drivers/sensor.h>
+#include <zmk/sensors.h>
```
mike1808, line 5: Hey, I'm curious, how do you compile it when `zmk` includes are not included in the CMakeLists?

crides: Can you clarify your problem? I'm having trouble understanding

app/drivers/sensor/cirque_trackpad/cirque_trackpad.c
```diff
@@ -0,0 +1,256 @@
...
+        k_sem_take(&data->gpio_sem, K_FOREVER);
+        pinnacle_int_cb(dev);
+        pinnacle_write(dev, PINNACLE_STATUS1, 0);   // Clear SW_DR
+    }
+}
+#elif defined(CONFIG_PINNACLE_TRIGGER_GLOBAL_THREAD)
+static void pinnacle_work_cb(struct k_work *work) {
+    struct pinnacle_data *data = CONTAINER_OF(work, struct pinnacle_data, work);
+    pinnacle_int_cb(data->dev);
+    pinnacle_write(dev, PINNACLE_STATUS1, 0);   // Clear SW_DR
```
mike1808, line 165:
```suggestion
    pinnacle_write(data->dev, PINNACLE_STATUS1, 0);   // Clear SW_DR
```

app/drivers/sensor/cirque_trackpad/cirque_trackpad.c
mike1808: When I'm trying to compile with rebased main the compiler fails to find `zmk/sensors.h` include. And I was wondering if it's the case for you.

crides: Absolutely not. I'm guessing you somehow have a ZMK version that's way too old, or maybe the CMake stuff got messed up somehow?

FrostKiwi:
How is this branch supposed to be used?
In the beginning this commit https://github.com/crides/zmk/commit/bdc8b9ba2d7e4bcee1f742de623a222ec2ead254 added the file `app/driver/sensor/cirque_trackpad/Kconfig`  but removed it in a later commit. Now the reference of that Kconfig points to nowhere, leading to a compile error:
https://github.com/FrostKiwi/zmk-config/actions/runs/4793167453/jobs/8525377273#step:9:42
```console
/__w/zmk-config/zmk-config/zephyr/scripts/kconfig/kconfig.py: /__w/zmk-config/zmk-config/zmk/app/drivers/sensor/Kconfig:6: '/__w/zmk-config/zmk-config/zmk/app/drivers/sensor/cirque_trackpad/Kconfig' not found (in 'rsource "cirque_trackpad/Kconfig"').
```
How are you supposed to use that branch to add Cirque Touchpad i2c support? Anything I should add?

crides:
I missed that in 4b5efc0. otherwise can't test/support things until at least next weekend

FrostKiwi:
> I missed that in [4b5efc0](https://github.com/zmkfirmware/zmk/commit/4b5efc0278502856b91134b6b569184bd12c807a). otherwise can't test/support things until at least next weekend

Compiled successfully ヽ(*・ω・)ﾉ
https://github.com/FrostKiwi/zmk-config/actions/runs/4793167453/jobs/8574209991
My frankenstein merge of: Kyria rev3 code + https://github.com/krikun98/zmk/tree/mouse-pr + this PR

alyx-brett:
Hi! Is there a roadmap to getting this functionality merged? It seems there's a lot of appetite and enthusiasm for it and I'd love if there was a way to make it easier.

FearlessSpiff:
> Whether it works or not, I'm about to find out I guess.

So, how did it go? Thinking about merging everything for my Aurora Corne... Or is there somewhere a usable/updated fork/branch for it? Thanks

FearlessSpiff:
> This adds the driver for Cirque Pinnacle trackpads. Note this is _only the driver_, support for fetching the data and binding behaviours to it should be added later as it depends on reworking the sensor subsystem.
> 
> The driver uses only the SPI/I2C bus and the `dr` data ready pin if available.
> 
> The implementation of the driver follows the [Cirque documentation](https://www.mouser.com/pdfDocs/InterfaceSPIandI2C.pdf). While the docs mentions using the sleep mode with high SPI speeds can be a problem, this is not observed in reality (I'm using a SPI clock of 10MHz).
> 
> Usage example: https://github.com/crides/zmk/blob/fusion-led-c-config/app/boards/arm/fusion/trackpad.c

Hi. I am trying to add this to my Aurora Corne Nicenano board. I came pretty far I think by merging your PR into urobs zmk base and fixing a few things. But now I am stuck at a compile error in trackpad.c:
```
/workspaces/zmk/app/boards/arm/nice_nano/trackpad.c:1:10: fatal error: drivers/sensor.h: No such file or directory
    1 | #include <drivers/sensor.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```
I think it has to do with the CMakeLists.txt which I took from elsewhere and added the stuff from your fusion board:
```
if(CONFIG_PINNACLE)

zephyr_library()
zephyr_library_sources(trackpad.c)
zephyr_library_include_directories(${ZEPHYR_BASE}/drivers)
zephyr_library_include_directories(${CMAKE_SOURCE_DIR}/include)
zephyr_library_include_directories(${CMAKE_SOURCE_DIR}/../include)

endif()
```

I can't find `sensor.h` in any of these directories and I have no clue, what `zephyr_library_include_directories(${CMAKE_SOURCE_DIR}/../include)` is supposed to be?

Thanks for any help. My code is at https://github.com/FearlessSpiff/zmk/tree/cirque-trackpad-1116 if that helps.

crides:
I'll have to look at this on my PC later (and I don't remember what zephyr version we're on), but I suspect that should be `zephyr/drivers/sensor.h` (due to the namespace change)

FearlessSpiff:
> I'll have to look at this on my PC later (and I don't remember what zephyr version we're on), but I suspect that should be `zephyr/drivers/sensor.h` (due to the namespace change)

This was it! 1000x thanks. I am now at the next error which I will check tomorrow when I have time again.

FearlessSpiff:
Again I don't know how to proceed... :) Don't know if this is the right place to ask or if I should ask on Discord or something. But here it is:
I am now stuck with:
```
/workspaces/zmk/zephyr/include/zephyr/device.h:104:39: error: '__device_dts_ord_DT_N_NODELABEL_cirque_pinnacle_ORD' undeclared here (not in a function)
  104 | #define DEVICE_NAME_GET(name) _CONCAT(__device_, name)
      |                                       ^~~~~~~~~
/workspaces/zmk/zephyr/include/zephyr/toolchain/common.h:132:26: note: in definition of macro '_DO_CONCAT'
  132 | #define _DO_CONCAT(x, y) x ## y
      |                          ^
/workspaces/zmk/zephyr/include/zephyr/device.h:104:31: note: in expansion of macro '_CONCAT'
  104 | #define DEVICE_NAME_GET(name) _CONCAT(__device_, name)
      |                               ^~~~~~~
/workspaces/zmk/zephyr/include/zephyr/device.h:264:37: note: in expansion of macro 'DEVICE_NAME_GET'
  264 | #define DEVICE_DT_NAME_GET(node_id) DEVICE_NAME_GET(Z_DEVICE_DT_DEV_NAME(node_id))
      |                                     ^~~~~~~~~~~~~~~
/workspaces/zmk/zephyr/include/zephyr/device.h:282:34: note: in expansion of macro 'DEVICE_DT_NAME_GET'
  282 | #define DEVICE_DT_GET(node_id) (&DEVICE_DT_NAME_GET(node_id))
      |                                  ^~~~~~~~~~~~~~~~~~
/workspaces/zmk/app/boards/arm/nice_nano/trackpad.c:14:33: note: in expansion of macro 'DEVICE_DT_GET'
   14 | const struct device *trackpad = DEVICE_DT_GET(DT_NODELABEL(cirque_pinnacle));
      |                                 ^~~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.
```
After checking Google and Zephyr (https://docs.zephyrproject.org/latest/build/dts/troubleshooting.html) I think it is because the driver doesn't go into the devicetree. I can't find any reference in `<build>/zephyr/include/generated/devicetree_generated.h` of it.

But I enabled it in my config if I'm not mistaken: https://github.com/FearlessSpiff/aurora-corne-zmk-config/blob/feature/testing-mouse-buttons/config/splitkb_aurora_corne.conf

The whole error can also be seen in the last run action: https://github.com/FearlessSpiff/aurora-corne-zmk-config/actions/runs/6730988507/job/18294788040

Thanks!

crides:
I don't see any instantiations of the pinnacle device here: https://github.com/FearlessSpiff/zmk/tree/cirque-trackpad-1116/app/boards/arm/nice_nano

FearlessSpiff:
> I don't see any instantiations of the pinnacle device here: https://github.com/FearlessSpiff/zmk/tree/cirque-trackpad-1116/app/boards/arm/nice_nano

Thanks, after reading your comment and reading some more documentation I now added the trackpad to the aurora corne shield: https://github.com/FearlessSpiff/zmk/blob/cirque-trackpad-1116/app/boards/shields/splitkb_aurora_corne/splitkb_aurora_corne.dtsi 

I know get another problem about "[undefined node label 'gpio0'](https://github.com/FearlessSpiff/zmk/actions/runs/6742567217/job/18328903041)". This seams to be for DR Pin, which I don't have as I will use I2C on the Aurora Corne. Can I just skip this and loose battery or what should I do here? Thanks!

FearlessSpiff:
Ah, I think I can disable it with `CONFIG_PINNACLE_TRIGGER=n`, right? Gonna try that...

FearlessSpiff:
Did some progress again. Now it fails when linking. And different errors when building the left side and right side. On the left side I have absolutely no clue what might go wrong. On the right side I have at least a problem description which would suggest, that the mouse code is not incorporated. Which might even be true on the peripheral side?!? Any ideas would be nice. [Here's the failed build.](https://github.com/FearlessSpiff/aurora-corne-zmk-config/actions/runs/6743689183/job/18332090251) Thanks!

crides:
The right side problems are easy, either the mouse stuff wasn't merged properly, or you didn't enable the mouse driver.

For the left side, I don't have any ideas, and you'll have to consult the ZMK/zephyr servers

FearlessSpiff:
> The right side problems are easy, either the mouse stuff wasn't merged properly, or you didn't enable the mouse driver.
> 
> For the left side, I don't have any ideas, and you'll have to consult the ZMK/zephyr servers

After many hours searching I could fix it with [this](https://github.com/FearlessSpiff/zmk/commit/5ba653fe49a8a5a900147a242f0cff7dbc758329). Coming from the Java world I have no idea how this can compile but then fail linking... Now have to add the actual trackpad... :)

crides:
Oh, because all of the initialization code in zephyr is stored in order of the priority in the binary. The macros just generate linker sections

FearlessSpiff:
After everything out of the way I was able to install the firmware. Already fixed the wrong i2c address but still init fails:

```
[00:00:00.383,422] <wrn> pinnacle: pinnacle start
[00:00:00.383,453] <err> pinnacle: i2c addr: 2a
[00:00:00.883,514] <err> i2c_nrfx_twi: Error on I2C line occurred for message 0
[00:00:00.883,636] <err> pinnacle: can't reset -5
```
I know the trackpad is working with an Elite-C on QMK.

Shy question: Was this tested with I2C? I can only find working SPI code... Or what am I missing? Also checked QMK code which looks similar to this... Will dig deeper tomorrow. Thanks!

crides:
Not sure what's happening, but given that this is the first I2C call, it's likely that there's some problem on your bus. I would check that there's pullup on the bus, you didn't swap `SDA` and `SCL`, and continuity is good. Oscilloscope/logic analyzer would be the next step.

> Shy question: Was this tested with I2C?

Valid question. Not me. I think 1 or 2 people on the ZMK discord tested and had success with it

> I can only find working SPI code... Or what am I missing? 

[There's definitely I2C code](https://github.com/crides/zmk/blob/cirque-trackpad/app/drivers/sensor/pinnacle/pinnacle.c#L42-L43)

FearlessSpiff:
> Not sure what's happening, but given that this is the first I2C call, it's likely that there's some problem on your bus. I would check that there's pullup on the bus, you didn't swap `SDA` and `SCL`, and continuity is good. Oscilloscope/logic analyzer would be the next step.
> 
Yeah, I know it is working because of the tests with an Elite-C MCU and QMK.
> [There's definitely I2C code](https://github.com/crides/zmk/blob/cirque-trackpad/app/drivers/sensor/pinnacle/pinnacle.c#L42-L43)

Sure, I meant in the open wild. I can see the I2C code in the driver.

Will investigate further. Thanks!

FearlessSpiff:
After a lot of debugging and some help from SplitKB Discord I had external power turned off. :( So now I know I2C is working also tested with an OLED on ZMK on the keyboard. I now get a new error for the first call to I2C `i2c_nrfx_twi: Error 0x0BAE0001 occurred for message 0` This seams to be an NACK for the address. So I checked there and found that in the [Cirque Documentation](https://www.mouser.com/pdfDocs/InterfaceSPIandI2C.pdf) Chapter 7.1 they talk about shifting bits for the address and then ORing with 1 when reading. I tried that for writing only atm on my branch with:
```
static int pinnacle_write(const struct device *dev, const uint8_t addr, const uint8_t val) {
    const struct pinnacle_config *config = dev->config;
    const struct i2c_dt_spec *spec = &config->bus;

    return i2c_reg_write_byte(spec->bus, spec->addr << 1, PINNACLE_WRITE | addr, val);
}
```
But still the same problem. Any idea what could be wrong?
Thanks again for looking into it...

FearlessSpiff:
As a follow up I tried to write the reset command to every I2C address on the bus in [this commit](https://github.com/urob/zmk/commit/32f34d75d8080cb6d0eb5c9b713f52529a9eca62) but it can't do it anywhere. So something seams to be off unfortunately.

crides:
I don't remember the address needs to be shifted. Also I can't really help much without an oscilloscope

Just checking: have you desoldered `R1` on the cirque, for switching to I2C mode?

FearlessSpiff:
> Just checking: have you desoldered `R1` on the cirque, for switching to I2C mode?

Yes, i switched to I2C and 5V. For anybody interested [here's a great resource](https://keycapsss.com/help/cirque-trackpad/) explaining it.  It works fine in QMK. 

I can't see anything on my oscilloscope. I ordered a logic analyzer from Aliexpress. This means a 2+ weeks break from the project...

I read the I2C Cirque documentation a lot and can't figure out the START command in the code. Just the STOP as a flag. But I guess this is done deeper in the I2C driver... I think without logic analyzer I am blind atm. :(

drzony:
As far as I remember ZMK has some issues with reinitializing I2C after sleep (this might have been fixed) and when the ext power is toggled. So you need to make sure that the ext power is ON on startup. I would try first on the master side with USB plugged in, that might save you some debugging time.

crides:
> I can't see anything on my oscilloscope. I ordered a logic analyzer from Aliexpress. This means a 2+ weeks break from the project...

@FearlessSpiff if you're still having issues, you could just post a scope screenshot here. If it proves to be too complex to debug, we could move to some other channel to debug and summarize back here

FearlessSpiff:
> @FearlessSpiff if you're still having issues, you could just post a scope screenshot here. If it proves to be too complex to debug, we could move to some other channel to debug and summarize back here

Thanks. Long time since I worked with these things. I just can't capture the start. And then it's just 5V...
As I am an inpatient person I now ordered a logic analyzer from a local dealer. It should arrive this week if I'm lucky.

FearlessSpiff:
I'm back... ;) Although without good news but a LA. I could now capture I2C init done by QMK where the trackpad is working and by ZMK where it doesn't. Unfortunatley I can't make a lot out of it. I just see that it is different and the trackpad returns NACK so it seams. Same as the error code told me in the running code. Maybe somebody knows more than me when having a look at this. Thanks

### QMK, working
QMK-working
---
### ZMK, not working
ZMK-not-working

crides:
I think it should be pretty clear then. Check that it's powered and the pins are indeed wired correctly

FearlessSpiff:
> I think it should be pretty clear then. Check that it's powered and the pins are indeed wired correctly

Unfortunately it is. I can just switch the MCU to the Elite-C with QMK and everything works.

FearlessSpiff:
> I think it should be pretty clear then. Check that it's powered and the pins are indeed wired correctly

Oh, I only have 3.3V on the trackpad. Shouldn't this be 5V? Have to check it...

drzony:
@FearlessSpiff I think on 3.3V MCUs you need to leave the 5V resistors (R7, R8) on the trackpad

FearlessSpiff:
Finally it is initalizing! Thanks to everybody! I am now at the next problem by not beeing able to set a trigger. I changed `SENSOR_TRIG_DATA_READY` to `SENSOR_TRIG_TIMER`:
```
static int trackpad_init() {
    struct sensor_trigger trigger = {
        .type = SENSOR_TRIG_TIMER,
        .chan = SENSOR_CHAN_ALL,
    };
    printk("trackpad");
    if (sensor_trigger_set(trackpad, &trigger, handle_trackpad) < 0) {
        LOG_ERR("can't set trigger");
        return -EIO;
    };
    return 0;
}
```
I can't find much in the docu, neither when googling for examples. Any idea? Thanks!

drzony:
> Finally it is initalizing! Thanks to everybody! I am now at the next problem by not beeing able to set a trigger. I changed `SENSOR_TRIG_DATA_READY` to `SENSOR_TRIG_TIMER`:
> 
> ```
> static int trackpad_init() {
>     struct sensor_trigger trigger = {
>         .type = SENSOR_TRIG_TIMER,
>         .chan = SENSOR_CHAN_ALL,
>     };
>     printk("trackpad");
>     if (sensor_trigger_set(trackpad, &trigger, handle_trackpad) < 0) {
>         LOG_ERR("can't set trigger");
>         return -EIO;
>     };
>     return 0;
> }
> ```
> 
> I can't find much in the docu, neither when googling for examples. Any idea? Thanks!

PINNACLE_TRIGGER_NONE is probably what you are looking for

FearlessSpiff:
AFAIK `PINNACLE_TRIGGER_NONE` is to disable DR functionality, which is the default kconfig setting. Therefore it does not compile with this.

FearlessSpiff:
Btw, I now logged the errorcode when calling `sensor_trigger_set` which is `-88`

drzony:
> AFAIK `PINNACLE_TRIGGER_NONE` is to disable DR functionality, which is the default kconfig setting. Therefore it does not compile with this.

When `PINNACLE_TRIGGER_NONE` is used, then `trigger_set` is not available, see around line 264 in `pinnacle.c`. When not using DR pin, you'll need to poll the sensor, I think the code for the mouse you are using needs the trigger. From what I understand, you would need to set up a timer or fetch data in some already running task.

FearlessSpiff:
Thanks. I get it now. I guess I will need to create a thread and poll the sensor there. Or check schematics of my Aurora and maybe add a DR line somewhere. What advantages would it have to have a DR line? Is this just better for battery-life and that's it?

FrostKiwi:
Does anyone know what has to be done to get this done on a theoretical level, considering the new ZMK firmware structure? My ZMK config for reference: https://github.com/FrostKiwi/zmk-config

FrostKiwi:
I merged this PR into `zmkfirmware/main` here: https://github.com/FrostKiwi/zmk (`main` branch) with this commit: https://github.com/FrostKiwi/zmk/commit/03a055d6f4825bd09d5768f142f4b635a27adf8b and these correction commits: https://github.com/FrostKiwi/zmk/commit/72200705a47cc7b1e8211ac43567a1597977f9ea  https://github.com/FrostKiwi/zmk/commit/dfb326fbc9d0c16b594a206108eadfb014897b7f

The firmware builds successfully without the driver being included: https://github.com/FrostKiwi/zmk-config/actions/runs/11378828087 but fails with the driver being included via `CONFIG_PINNACLE=y`: https://github.com/FrostKiwi/zmk-config/actions/runs/11379092737/job/31656277722

```c
/tmp/zmk-config/zmk/app/module/drivers/sensor/pinnacle/pinnacle.c:3:10: fatal error: init.h: No such file or directory
    3 | #include <init.h>
      |          ^~~~~~~~
```

I have a Kyria rev3, nice!nano2, Cirque trackpad on both sites hooked up to i2c.
When this PR was created, Mouse button support didn't yet hit `main` yet. But now this is the case and mouse buttons work just fine on my Kyria rev3 via https://zmk.dev/docs/keymaps/behaviors/mouse-emulation Now to the trackpad and this PR.

- What has to be done to make this build in modern ZMK?
- Via GitHub search I think I found a user of the cirque trackpad: https://github.com/lazy-dinosaur/zmk-config-LazyDino/blob/master/config/lazycorne.keymap @lazy-dinosaur does the trackpad work for you? What did you have to do in order for it to work?
- Can this driver move the mouse on it's own or will https://github.com/zmkfirmware/zmk/pull/2027 as opened by @petejohanson be required?

caksoylar:
You should try using https://github.com/petejohanson/cirque-input-module (which is a [module](https://zmk.dev/docs/features/modules)) along with either #2027 or #2477. I am not sure if the current state of this PR is compatible with these two, but I am sure that you need one of those branches for the Cirque to work.

FrostKiwi:
@caksoylar Ohh, I never knew of this module system! So the way I understand is: Go with https://github.com/zmkfirmware/zmk/pull/2027 as the base, it's only 2 months behind `main` and https://github.com/zmkfirmware/zmk/pull/2477 is in early infancy. @petejohanson do you agree with the assessment of those PRs?

Include the module. This seems easy and I can follow. But how do you glue it together?
The module describes a couple of settings: https://github.com/petejohanson/cirque-input-module/blob/main/dts/bindings/input/cirque%2Cpinnacle-common.yaml but where do I set them and how does ZMK know which I2C address to use? In the keymap like:
```
&trackpad {
    compatible = "cirque,pinnacle";
    reg = <???>;  # I2C address for the trackpad
    dr-gpios = <???>;
    x-axis-z-min = <5>;
    y-axis-z-min = <4>;
    sensitivity = "2x";
    rotate-90;
};
```

Looking at the source code of https://github.com/zmkfirmware/zmk/pull/2027 and seeing `CONFIG_ZMK_MOUSE=y`, I guess this is an extension of https://zmk.dev/docs/keymaps/behaviors/mouse-emulation. So the cirque module just assumes mouse emulation to be present and works or does there some configuration need to be done gluing both together?

PR #1117, opened 2022-02-04
Add wired split support using UART communication

megamind4089:
Initial wired split support. Raised this PR to get feedback on initial changes and design

**Changes:**
* Updated Kconfig to allow configuring split communication type using Kconfig choice
* Type can either be BLE or SERIAL for now
* BLE is selected as default when board supports BLE
* Serial is selected when there is only USB support
* Used choice for central/peripheral instead of using single flag. This is experimental. Want to get feedback on this
* Use chosen node to select UART dev for serial split driver
* Serial central & service behaves similarly to the current ble split comm
* Only addition is the CRC to split position packet which i added as a defensive check
* Please look at the a_dux Kconfig.defconfig changes and let me know whether it makes sense to use like this.

**Note:**
* I tried to make changes to be not intrusive, i.e does not change/affect existing design to make PR merge smoother
* Ignore the west.yml change. Cherrypicked one PR from zephyr main for STM32 single wire UART and will revert it
* Ignore the stemcell.overlay file. I added this to give an idea about the change needed. Will remove this later.
* Will update other split keyboard's defconfig once the design seems acceptable

SteMCell board config:
https://github.com/megamind4089/zmk-config/tree/main/config/boards/arm/stemcell

Cherrypicked zephyr PR:
https://github.com/zephyrproject-rtos/zephyr/pull/40525


app/Kconfig
```diff
@@ -217,6 +218,56 @@ endif
 #ZMK_SPLIT_BLE
 endif
 
+# ZMK_BLE
+endif
+
+if ZMK_USB
```
petejohanson, line 224: I'm not sure this really needs to be behind this condition, TBH.

app/boards/shields/a_dux/Kconfig.defconfig
```diff
@@ -1,19 +1,41 @@
 # Copyright (c) 2021 The ZMK Contributors
 # SPDX-License-Identifier: MIT
 
+if SHIELD_A_DUX_LEFT || SHIELD_A_DUX_RIGHT
+
+if ZMK_BLE || ZMK_USB
```
petejohanson, line 6: Why isn't this just defaulted on? Why behind this check?

app/boards/shields/a_dux/boards/stemcell.overlay
```diff
@@ -0,0 +1,37 @@
...
+	chosen {
+		zmk,split-serial = &usart1;
+	};
+};
+
+&usart1 {
+	status = "okay";
+};
+
+&kscan0 {
```
petejohanson, line 17: Why would the pins for a kscan device need overriding for just one board?

app/src/split/bluetooth/service.c
```diff
@@ -141,12 +142,12 @@ int send_position_state() {
     return 0;
 }
 
-int zmk_split_bt_position_pressed(uint8_t position) {
+int zmk_split_position_pressed(uint8_t position) {
```
petejohanson, line 145: It this is common, it shouldn't, IMHO, be duplicated for both, but should really be a reusable common core that then delegates to the transport to send the state.

app/src/split/serial/central.c
```diff
@@ -0,0 +1,190 @@
...
+
+#include <logging/log.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#include <zmk/event_manager.h>
+#include <zmk/events/position_state_changed.h>
+#include <zmk/matrix.h>
+
+#if DT_HAS_CHOSEN(zmk_split_serial)
```
petejohanson, line 22: If we don't have a chosen, we should just `#error` instead of wrapping this all in a conditional.

app/src/split/serial/central.c
```diff
@@ -0,0 +1,190 @@
...
+
+#define UART_NODE1 DT_CHOSEN(zmk_split_serial)
+const struct device *serial_dev = DEVICE_DT_GET(UART_NODE1);
+static int uart_ready = 0;
+
+#define CONFIG_ZMK_SPLIT_SERIAL_CENTRAL_POSITION_QUEUE_SIZE 5
+
+#define CONFIG_SERIAL_THREAD_STACK_SIZE 128
+#define CONFIG_SERIAL_THREAD_PRIORITY 5
```
petejohanson, line 31: Why not actual Kconfig settings?

app/src/split/serial/central.c
```diff
@@ -0,0 +1,190 @@
...
+    int ret = uart_callback_set(serial_dev, uart_callback, NULL);
+    if (ret == -ENOTSUP || ret == -ENOSYS) {
+        LOG_WRN("UART device:%s ASYNC not supported", serial_dev->name);
+        return;
+    }
+
+    uart_ready = 1;
+    LOG_DBG("UART device:%s ready", serial_dev->name);
+
+    while(true) {
```
petejohanson, line 168: Do we really need to hard loop here? Can't we just enable the RX here, and then in the notify callback, do more? Not familiar w/ the UART API in detail, but this seems off.

app/src/split/serial/service.c
```diff
@@ -0,0 +1,106 @@
...
+#include <device.h>
+#include <drivers/uart.h>
+
+#include <logging/log.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#include <zmk/matrix.h>
+
+#if DT_HAS_CHOSEN(zmk_split_serial)
```
petejohanson, line 21: Ditto, should `#error` if no chosen node, not wrap all this in the `#if`.

app/src/split/serial/service.c
```diff
@@ -0,0 +1,106 @@
...
+              CONFIG_ZMK_SPLIT_SERIAL_PERIPHERAL_POSITION_QUEUE_SIZE, 4);
+
+
+void send_data_via_uart(const struct device *dev, char *data, size_t len) {
+    if (!uart_ready) {
+        return;
+    }
+
+    for (int i=0; i < len; i++) {
+        uart_poll_out(serial_dev, data[i]);
```
petejohanson, line 43: Why not `uart_tx` for the whole buffer?

app/west.yml
```diff
@@ -4,12 +4,14 @@ manifest:
...
+    - name: megamind4089
+      url-base: https://github.com/megamind4089
     - name: microsoft
       url-base: https://github.com/microsoft
   projects:
     - name: zephyr
-      remote: zmkfirmware
-      revision: v2.5.0+zmk-fixes
+      remote: megamind4089
+      revision: v2.5.0+zmk-fixes-stm32f4
```
petejohanson, line 14: Anything here actually relevant to this work that this is needed?

petejohanson (changes requested):
First pass w/ some initial questions. Thanks for the work on this!

app/west.yml
megamind4089: No, please ignore this file. Will remove this before final cleanup.
Just added here to show that i had to cherry-pick this PR to make it work in half duplex mode.
https://github.com/zephyrproject-rtos/zephyr/pull/40525

app/src/split/serial/service.c
megamind4089: I have done the TX and RX part first using sync mode and then updated RX to async using DMA.
Will update TX also to use same apis

app/boards/shields/a_dux/boards/stemcell.overlay
megamind4089: Existing split keyboards use any one of four ports D0, D1, D2 and D3 for split communication
Ex: Lily58 uses D2 (promicro RX pin), A_Dux uses D1 (I2C SDA pin) etc

STM32 chips support half duplex UART comm only on TX pins. 
So STeMCell has jumpers on back side to swap the TX/RX pins.
This depends completely on the shield which uses the pin for communication.

Other way is to use some KConfig variable to detect and swap pins. 
But i have not figured out how to do it exactly using Kconfig and device tree yet,

So in the meantime, i have just juggling the pins inside the shields directly.

app/Kconfig
megamind4089: Do you mean we can remove this check ?
I added this check for cases where we have controllers with chips without having USB peripheral like nrf52832 chip.
If we are making an assumption that we support only USB controllers, we can remove this.
Let me know if i mis-understood the question

app/boards/shields/a_dux/Kconfig.defconfig
megamind4089: Yes, does not makes much sense. No idea why i added this.
Will remove this check.

app/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ target_sources_ifdef(CONFIG_ZMK_BLE app PRIVATE src/events/battery_state_changed
 target_sources_ifdef(CONFIG_USB app PRIVATE src/events/usb_conn_state_changed.c)
 target_sources(app PRIVATE src/behaviors/behavior_reset.c)
 target_sources_ifdef(CONFIG_ZMK_EXT_POWER app PRIVATE src/behaviors/behavior_ext_power.c)
-if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL)
+if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL OR CONFIG_ZMK_SPLIT_SERIAL_ROLE_CENTRAL)
```
petejohanson, line 48: See the refactor in #1314 which will massively simplify things drastically and avoid this change being needed.

app/Kconfig
```diff
@@ -155,10 +155,14 @@ config ZMK_SPLIT
 
 if ZMK_SPLIT
 
+choice ZMK_SPLIT_TYPE
```
petejohanson, line 158: Ditto on #1314

app/boards/shields/a_dux/Kconfig.defconfig
```diff
@@ -1,19 +1,39 @@
...
+
 if SHIELD_A_DUX_LEFT
 
 config ZMK_KEYBOARD_NAME
 	default "A. Dux"
 
+if ZMK_BLE
 config ZMK_SPLIT_BLE_ROLE_CENTRAL
 	default y
+endif
```
petejohanson, line 17: This should be easier w/ the agnostic role config in #1314

marcoster:
Should really the Async-API of zephyr's uart be used here?
As far as I've seen the API is only supported by STM32, NRF and sam0.
Or is the API the future 'standard' for uart in zephyr.

At least for now I'm not able to 'simply' use this on a Pi Pico... maybe I'll looker deeper on how the async api has to be implemented or find a workaround for this.

Maybe at least a Kconfig error should be generated if *SERIAL_SUPPORT_ASYNC* of a board is set to *n*?

petejohanson:
I do think using the async API is probably preferred, which means we should work on the RP2040 UART driver for Zephyr to add support for it.

app/Kconfig
genius-ua: I want to use nice nano in USB mode without bluetooth

marcoster:
@mbrgm Sorry, never wrote an answer - as I don't have any nice!nanos around I don't really know, but since the nRF52840 in Zephyr supports async uart it normally should work.
@nov1n Well I kinda don't need wireless split and some quite cheap and available controller-options (RP2040 boards i.e.) don't have any wireless capabilities.
If done right I guess it should be possible to connect to the host per BLE.

jesseleite:
Am I right to assume that UART support in Zephyr would allow for wired split communication between peripheral and central over TRRS? 🙏

petejohanson:
> Am I right to assume that UART support in Zephyr would allow for wired split communication between peripheral and central over TRRS? 🙏

This PR definitely needs a rebase into latest ZMK `main`, but yes*.


* this assumes a proper hardware uart, not any single wire bit-banging approach, etc as is often used in many three wire TRRS splits.

M1cha:
I have a corne cherry with sparkfun_pro_micro_rp2040 and will make this work with the latest main branch within the next few days. Shall I open a new PR once done?

petejohanson:
> I have a corne cherry with sparkfun_pro_micro_rp2040 and will make this work with the latest main branch within the next few days. Shall I open a new PR once done?

Sure! See my previous note about the current state needing a working UART device (in terms of Zephyr device w/ associated UART driver) so any single wire work will need to either:

* Adapt this code to use a higher level abstraction than the Zephyr UART driver one, *OR*
* Create a UART driver that works for single wire hardware, maybe using PIO, etc. Doing this likely requires experimental Zephyr version that recently added PIO driver support.

jonathanfv:
Hey there, I was looking at implementing "just this" (wired connection between halves, likely using TRRS and UART TTL) on the Glove80 (dual nRF52840, harware UART). Any way I can help?

thomasbaart:
Am also interested in seeing this happen. If there's a contributor who wants to pick this up but needs hardware to do so, please post a comment and I'll sponsor the needful :)

erickearns:
> Also, I wonder what the main benefits are of wiring the two halves. Does it reduce power consumption or improve latency?

I have personally worked for several organizations that prohibit the use of input devices that utilize BT or other wireless connectivity. And from what I've seen (discussion around this topic from prospective Adv360/Glove80 customers), others are dealing with substantially similar restrictions.

daniel-frak:
> Also, I wonder what the main benefits are of wiring the two halves. Does it reduce power consumption or improve latency?

I would also add improved robustness and reliability. Bluetooth can be unreliable and I personally sometimes have problems with the two halves recognizing each other. This problem goes away with a fully wired setup.

klardotsh:
Voicing my huge support for this to land. ZMK is still to my knowledge the only major firmware that doesn't support wired split keyboards, which make up a large amount (I'm stopping shy of saying "overwhelming majority" due to selection bias; I prefer them so maybe I notice more of them, but I definitely notice more of them) of the ergo mech KB scene. Wired splits have been around and stable for a long time, and they work great and reliably.

Further, some keyboards *only* exist in fully-wired form. I've been looking at buying a [Svalboard](https://svalboard.com/), for example, which is all USB-C everywhere, powered by Bluetooth-less RP2040s. To use that board, I'd have to go back to QMK and write a bunch of custom C to replicate ZMK features I've enjoyed, like the enhanced Mod-Tap key system and declarative homerow mods.

I don't have the available time to help write the code, but I'm happy to try to help in other ways to get this to land if help is needed.

kumekay:
To me, the main concern of not having wired split is battery consumption. I'm working on ultra-compact portable keyboards, and always listening left half is a serious problem when tiny batteries are used. Furthermore, I would prefer to put battery only to one half of the keyboard to save the size. Now I'm using FFC with big number of lanes between half, but if this will be merged I'll happily put a second controller to the right half, but will use narrow FFC. I can also help to test the solution.

nmunnich:
Superseded by #2766

PR #1120, opened 2022-02-05
add: LED widgets

### LINKED ISSUE #1003: Feature: LED widgets
This is sort of a term I coined, but the idea is to have LEDs act as displays. The widgets would be parameterized by PWM brightness, LED number (as in which LED to use; like colors or just linear position), and on/off length. The basic features would just be an API to preset several widgets and display them (with some spacing between them) in order when they are requested.

Just like actual screen display widgets, this'll be hard to customize for various different events so we shouldn't go there (unless someone wants to group all core events into a 32bit flag, which would be cool but unnecessary).

This is to mainly put the idea out here. I will need this at some point in the near future (need as in nice to have), so if nobody else want this I'll probably implement this anyway (just really not sure _when_ it will be); otherwise, others can take a stab if they want to

crides:
Fixes #1003 

Currently has the config in C, see example [here](https://github.com/crides/zmk/blob/fusion-led-c-config/app/boards/arm/fusion/led_widgets.c). You also need to setup a LED device as shown [here](https://github.com/crides/zmk/blob/fusion-led-c-config/app/boards/arm/fusion/fusion.dts#L23). It would be better to put the config in devicetree, however there are some macro issues to be resolved before that could be done. Having the config in devicetree means that the config is better structured, and is less janky (see below) compared to the current C implementation. However, it also means that rebuilding is slower due to the nature of processing devicetrees in Zephyr.

Issues:
- [ ] The C config is janky, and uses embedded arrays to allow for easier config and macro generation. There are magic numbers for the (max) size of those embedded arrays. This could either be solved by having actual configuration items for them, or have smarter macros (probably hard)
- [ ] Currently this only supports single LEDs, whether PWM or not should work fine; support for addressable LED isn't a technical problem, and I'll see what needs to be done to adapt the config/interface
- [ ] No animations supported

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -101,6 +101,7 @@ target_sources_ifdef(CONFIG_ZMK_RGB_UNDERGLOW app PRIVATE src/rgb_underglow.c)
 target_sources_ifdef(CONFIG_ZMK_BACKLIGHT app PRIVATE src/backlight.c)
 target_sources_ifdef(CONFIG_ZMK_LOW_PRIORITY_WORK_QUEUE app PRIVATE src/workqueue.c)
 target_sources(app PRIVATE src/main.c)
+target_sources_ifdef(CONFIG_ZMK_LED_WIDGETS app PRIVATE src/led_widgets.c)
 
 add_subdirectory(src/display/)
 add_subdirectory_ifdef(CONFIG_SETTINGS src/settings/)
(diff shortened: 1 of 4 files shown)


alinelena:
i have played with the layer led widget and works for "normal" layers but if you use a conditional layer the indicator does not work when this layer is activated.

example here
https://github.com/alinelena/zmk/tree/mlego2/app/boards/shields/mlego2x2

PR #1127, opened 2022-02-12
fix(docs): Fix issues with positional hold-tap example

### LINKED ISSUE #646: Docs contain keymaps with non-existent and deprecated key defines.
Currently our example keymaps on the docs contain key defines that are deprecated or don't even exist. All keymaps in our docs should be run through the keymap upgrader. 

Example: `QUOTE` in https://zmkfirmware.dev/docs/features/keymaps#layers-1

caksoylar:
This PR replaces the deprecated `SCLN` keycode in the positional hold-tap example, which fixes #646 (I found no other deprecated keycodes in the docs).

Also I tweaked the label of the behavior in the example. We see folks copying this behavior node and duplicating it for the right hand, then forgetting to modify the "label" value and getting confusing runtime ehavior. If we modify the label to be left-specific like this it might be a better hint to change it when duplicating it.

Two unrelated changes in one PR isn't ideal but they were small enough that I thought I shouldn't split them up; ideally we would not squash this if we merge the PR.

### DIFF
--- a/docs/docs/behaviors/hold-tap.md
+++ b/docs/docs/behaviors/hold-tap.md
@@ -115,7 +115,7 @@ The following are suggested hold-tap configurations that work well with home row
 	behaviors {
 		lh_pht: left_hand_positional_hold_tap {
 			compatible = "zmk,behavior-hold-tap";
-			label = "POSITIONAL_HOLD_TAP";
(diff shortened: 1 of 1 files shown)


dxmh (approved):
LGTM! Thanks @caksoylar 🙂

PR #1128, opened 2022-02-12
fix(docs): Improve powershell command for setup script for failure cases

### LINKED ISSUE #1206: Powershell setup script command closes terminal window
The command on https://zmk.dev/docs/user-setup for running the setup script from PowerShell runs the commands in the current interpreter's context, so any return statements in the setup script for bailing on error just close the entire terminal window.

One solution that was suggested on Discord is to change the command to wrap everything in another instance of PowerShell:

```ps
powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('https://zmk.dev/setup.ps1'))"
```

caksoylar:
This change tweaks the `powershell` setup script invocation in the docs so that if it exits with a failure it will not close the current terminal window. We saw quite a few folks in Discord confused because of this behavior.

From what I have tried this helps the current shell stay alive for error cases and doesn't persist a nested shell even after the command runs successfully. I tried with Windows 11 and Powershell version 5.1, but continues working if I try different versions down to `3` with `powershell -Version 3`.

That being said I am not very knowledgeable in Powershell and I don't know how backwards-compatible the `-Command` switch is. I would appreciate any advice or testing from folks with different Windows versions.

Fixes #1206.

### DIFF
--- a/docs/docs/user-setup.md
+++ b/docs/docs/user-setup.md
@@ -82,7 +82,7 @@ bash -c "$(wget https://zmk.dev/setup.sh -O -)" '' --wget
 <TabItem value="PowerShell">
 
 ```
-iex ((New-Object System.Net.WebClient).DownloadString('https://zmk.dev/setup.ps1'))
+powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('https://zmk.dev/setup.ps1'))"
 ```
 
 </TabItem>


dxmh:
[Report from user on Discord ](https://discord.com/channels/719497620560543766/719909884769992755/946655429952434218) confirming this change was helpful. Windows 10 64 bit, Powershell 5.1.19041.1320.

caksoylar:
Another user reported this workaround working for them: https://discord.com/channels/675924128108118016/741329942771662909/959479545906495538. We also haven't had reports of this not working, so I think it is safe to merge. @dxmh what do you think?

dxmh (approved):
Thanks @caksoylar 🙂 I think it should be safe to merge based on the feedback we've seen in Discord and #1206.

PR #1132, opened 2022-02-14
Fix tidbit_19key

luigi311:
Fixed issue with tidbit_19key not being handled correctly. Without this fix the board was never being given a name so bluetooth was not functioning correctly.


Nick-Munnich:
Closed by #2369

PR #831, opened 2021-06-10
feat(behavior) Add Tap Dance (Updated)

kurtis-lew:
Add tap dances as a behavior. Tap dances are designed to have the following characteristics:
- [x] Can include any ZMK keycode or behavior as a binding (tested regular keypresses, sticky keys, toggle layers, hold-taps as of initial PR)
- [x] Resolve immediately after the number of keypresses is equal to the number of bindings
- [x] Resolve immediately and end when the next keypress's position is not that of the last-pressed tap dance

As of this PR, the comments by @okke-formsma on PR #811 from 12:43 PDT today (June 9, 2021) and observations from @dxmh's testing on March 31, 2021, have both been addressed.

PR #1139, opened 2022-02-19
feat(behavior) Add Tap-Dance (2022)

kurtis-lew:
Consists of the following commits to `tapdance_2022_working` branch in local repo:
**Updated Tap-Dances to enable hold functionality**
**Update Documentation**
- Edited for clarity and added Images

**Uploaded New Tests**
**Updates according to test results of PR[#1139](https://github.com/zmkfirmware/zmk/issues/1139)** (Updated at 14:45 Pacific Time)

### Updated Design Criteria:
- [x] Can include any ZMK keycode or behavior as a binding (tested regular keypresses, sticky keys, toggle layers, hold-taps as of initial PR)
- [x] Resolve immediately after the number of keypresses is equal to the number of bindings
- [x] Resolve immediately ~~and end~~ when the next keypress's position is not that of the last-pressed tap dance 
- [x] Support holds (NEW)

 Further updates to #831 according to testing from testers of tap-dances since original work since June 2021.


app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,347 @@
...
+// release_tap_dance_behavior
+
+// Returns behavior_keymap_binding_pressed to ensure that keycode is actually released
+static inline int release_tap_dance_behavior(struct active_tap_dance *tap_dance,
+                                             int64_t timestamp) {
+    // Release chosen tap dance binding
+    struct zmk_behavior_binding binding =
+        (tap_dance->counter < tap_dance->config->behavior_count)
+            ? tap_dance->config->behaviors[(tap_dance->counter) - 1]
+            : tap_dance->config->behaviors[(tap_dance->config->behavior_count) - 1];
```
petejohanson, line 161: Can't this be simplified to:

```suggestion
            tap_dance->config->behaviors[MIN(tap_dance->counter, tap_dance->config->behavior_count) - 1];
```

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,347 @@
...
+
+// press_tap_dance_behavior
+
+// Returns behavior_keymap_binding_pressed to ensure that keycode is actually pressed
+static inline int press_tap_dance_behavior(struct active_tap_dance *tap_dance, int64_t timestamp) {
+    // Press chosen tap dance binding
+    struct zmk_behavior_binding binding =
+        (tap_dance->counter < tap_dance->config->behavior_count)
+            ? tap_dance->config->behaviors[(tap_dance->counter) - 1]
+            : tap_dance->config->behaviors[(tap_dance->config->behavior_count) - 1];
```
petejohanson, line 144: Ditto:

```suggestion
        tap_dance->config->behaviors[MIN(tap_dance->counter, tap_dance->config->behavior_count) - 1];
```

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,347 @@
...
+}
+
+// *new_tap_dance
+
+// Searches through array of all active tap-dances (active_tap_dances)
+// for first element with FREE position. Set its variables to default
+// values as to overwrite any settings from previous tap-dances.
+// Returns pointer to reset active_tap_dance created by this function.
+static struct active_tap_dance *new_tap_dance(uint32_t position,
+                                              const struct behavior_tap_dance_config *config) {
```
petejohanson, line 76: Since this could fail if we hit the max held, perhaps better to be:

```suggestion
static int new_tap_dance(uint32_t position,
                                              const struct behavior_tap_dance_config *config, struct active_tap_dance **tap_dance) {
```

And update the pointer w/ the selected active tap dance if we have an open slot, return 0, otherwise return `-ENOMEM`.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,347 @@
...
+static int on_tap_dance_binding_pressed(struct zmk_behavior_binding *binding,
+                                        struct zmk_behavior_binding_event event) {
+    // Code executes when the tap dance key is pressed.
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_tap_dance_config *cfg = dev->config;
+    struct active_tap_dance *tap_dance;
+    tap_dance = find_tap_dance(event.position);
+    // No active tap dance found on keypress location, create a new one
+    if (tap_dance == NULL) {
+        tap_dance = new_tap_dance(event.position, cfg);
```
petejohanson, line 187: Needs some error checking in case we hit the max held.

docs/docs/behaviors/tap-dance.md
```diff
@@ -0,0 +1,69 @@
+---
+title: Tap-Dance Behavior
+sidebar_label: Tap-Dance
+---
+
+## Summary
+
+A tap-dance key outputs a keycode or behavior corresponding to how many times it is pressed.
```
petejohanson, line 8: Perhaps a minor doc tweak:

```suggestion
A tap-dance key invokes a different behavior (e.g. `kp`) corresponding to how many times it is pressed.
```

docs/docs/behaviors/tap-dance.md
```diff
@@ -0,0 +1,69 @@
...
+
+This example configures a tap-dance that outputs the number of keypresses from 1-3.
+
+```
+#include <behaviors.dtsi>
+#include <dt-bindings/zmk/keys.h>
+
+/ {
+	behaviors {
+		td: tap_dance {
```
petejohanson, line 39: Perhaps number this to reinforce the need for uniqueness:

```suggestion
		td0: tap_dance_0 {
```

docs/docs/behaviors/tap-dance.md
```diff
@@ -0,0 +1,69 @@
...
+            compatible = "zmk,behavior-tap-dance";
+            label = "TAP_TEST";
```
petejohanson, line 41: Ditto:

```suggestion
            label = "TAP_DANCE_0";
```

docs/docs/behaviors/tap-dance.md
```diff
@@ -0,0 +1,69 @@
...
+            bindings = <&kp N1>, <&kp N2>, <&kp N3>;
+        };
+	};
+
+	keymap {
+		compatible = "zmk,keymap";
+
+		default_layer {
+			bindings = <
+	            &td
```
petejohanson, line 53:
```suggestion
	            &td0
```

petejohanson (changes requested):
Thanks for working on this! A few review comments.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,347 @@
...
+                return ZMK_EV_EVENT_CAPTURED;
+            }
+            // Tap dance will release on key release, allowing for decided tap-dances
+            // to be held, like those using mod keys
+        }
+    }
+    return ZMK_EV_EVENT_BUBBLE;
+}
+
+#define _TRANSFORM_ENTRY(idx, node)                                                                \
```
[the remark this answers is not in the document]
kurtis-lew: Just to confirm, am I replacing `ZMK_EV_EVENT_BUBBLE` or `_TRANSFORM_ENTRY`? Currently assuming it's the latter that needs to be updated.

app/src/behaviors/behavior_tap_dance.c
kurtis-lew: `static int new_tap_dance`, formerly `static struct active_tap_dance *new_tap_dance` already had error checking for the max held. I've added a few extra comments and another `LOG_ERR` statement to make this clear. 
```
static int new_tap_dance(uint32_t position,
                         const struct behavior_tap_dance_config *config,
                         struct active_tap_dance **tap_dance) {
    // Attempt to create a new instance of a tap-dance
    for (int i = 0; i < ZMK_BHV_TAP_DANCE_MAX_HELD; i++) {
        struct active_tap_dance *const tap_dance = &active_tap_dances[i];
        // Skip any elements in active_tap_dances[] that should not be altered (active/paused tap-dances)
        if (tap_dance->position != ZMK_BHV_TAP_DANCE_POSITION_FREE || tap_dance->timer_cancelled) {
            continue;
        }
        tap_dance->counter = 1;
        tap_dance->position = position;
        tap_dance->config = config;
        tap_dance->release_at = 0;
        tap_dance->is_pressed = true;
        tap_dance->timer_started = true;
        tap_dance->timer_cancelled = false;
        tap_dance->tap_dance_decided = false;
        return 0;
    }
    // Unable to store new tap-dance in array active_tap_dances[].
    LOG_ERR("Unable to store new tap-dance. Insufficient space.");
    return -ENOMEM;
}
```

kurtis-lew: I realize that the `for` loop could just as easily be rewritten with flipped logic. Something like:
```
static int new_tap_dance(uint32_t position,
                         const struct behavior_tap_dance_config *config,
                         struct active_tap_dance **tap_dance) {
    for (int i = 0; i < ZMK_BHV_TAP_DANCE_MAX_HELD; i++) {
        struct active_tap_dance *const tap_dance = &active_tap_dances[i];
        if (tap_dance->position == ZMK_BHV_TAP_DANCE_POSITION_FREE) {
            // Found a free space in active_tap_dances[], create new tap-dance
            tap_dance->counter = 1;
            tap_dance->position = position;
            tap_dance->config = config;
            tap_dance->release_at = 0;
            tap_dance->is_pressed = true;
            tap_dance->timer_started = true;
            tap_dance->timer_cancelled = false;
            tap_dance->tap_dance_decided = false;
            return 0;
        }
    }
    // Unable to store new tap-dance in array active_tap_dances[].
    LOG_ERR("Unable to store new tap-dance. Insufficient space.");
    return -ENOMEM;
}
```
The `if` statement in the loop shouldn't need to consider `tap_dance->timer_cancelled` because `clear_tap_dance` doesn't alter the value of `tap_dance->timer_cancelled`. It's implied that the status of the timer is irrelevant if the position of the tap-dance is `ZMK_BHV_TAP_DANCE_POSITION_FREE`.

petejohanson: Maybe I wasn't clear, I mean that this code calls `new_tap_dance` but doesn't actually verify it was given back a new tap dance struct to work with.

app/src/behaviors/behavior_tap_dance.c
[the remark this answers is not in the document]
petejohanson: Just the `_TRANSFROM_ENTRY` call.

app/src/behaviors/behavior_tap_dance.c
kurtis-lew: I understand. I'll keep that in mind for today's revisions.

docs/docs/behaviors/tap-dance.md
```diff
@@ -0,0 +1,69 @@
...
+	            &td0
+			>;
+		};
+	};
+};
+```
+
+The following image describes the behavior of this particular tap-dance.
+
+![Timing Diagram](../assets/tap-dance/timing_diagram.svg)
```
caksoylar, line 62: image
I think the intention here is to show an example where the tap dance key is held (blue `td` block) until `j` key is tapped. However this doesn't seem to match the diagram where the blue block cuts off before the orange block starts.

docs/docs/behaviors/tap-dance.md
```diff
@@ -0,0 +1,69 @@
...
+
+:::note
+Since all `bindings` used for `&td` are [key presses](../behaviors/key-press),
+they will release as soon as an interrupting key press occurs. If a modifier key like
+`LSHIFT` were to replace the `N1` binding, it would remain pressed until
+`&td`'s binding is released. That is to say, the output would instead be `J`.
```
caksoylar, line 68: I am a bit confused about the interrupt behavior as described here -- I wonder if this could be made clearer. 

Modifiers like `LSHIFT` are technically key press behaviors as well, so there is a contradiction with the first sentence. In addition the last two sentences seem to imply there is special logic for modifiers, if that is the case it can be made clearer.

docs/docs/behaviors/tap-dance.md
kurtis-lew: If we read this part of the diagram with only the code in mind, the `td` block doesn't necessarily need to be held down when the `j` is tapped since the tap-dance has technically not been decided yet since the `tapping-term-ms` has not fully elapsed.
However, I definitely see your point about the confusion there since the documentation currently implies that interrupting key-bindings only trigger the tap-dance when the tap-dance key is also held down. Not sure at the moment if adding a note on the image or just putting it directly in the docs would be better. Thanks for the feedback!

caksoylar: I see, it makes sense that we call it an "interrupt" also when we are waiting for a potential next tap, and not only while the key is held down (which would be the case for e.g. a hold-tap). I am also not sure how to explain it better; I can add a suggestion if I come up with one.

docs/docs/behaviors/tap-dance.md
caksoylar: After thinking about this one more: I wonder if the above diagram is slightly incorrect with regards to the timing of `j` output? This paragraph makes sense to me if `1` and `j` were output at the same time in the above diagram, rather than `j` being output after `1` is.

kurtis-lew: I'll check in the serial monitor if they output at the same time. At least visually, the output in say, a text editor, `1` comes before `j`.

kurtis-lew:
```
[00:00:09.845,397] [0m<dbg> zmk.on_keymap_binding_pressed: position 1 keycode 0x70004[0m
[00:00:09.845,428] [0m<dbg> zmk.hid_listener_keycode_pressed: usage_page 0x07 keycode 0x04 implicit_mods 0x00 explicit_mods 0x00[0m
[00:00:09.845,428] [0m<dbg> zmk.zmk_hid_implicit_modifiers_press: Modifiers set to 0x00[0m
[00:00:09.845,428] [0m<dbg> zmk.zmk_endpoints_send_report: usage page 0x07[0m
[00:00:09.845,489] [0m<dbg> zmk.release_captured_events: Releasing key position event for position 0 pressed[0m
[00:00:09.845,520] [0m<dbg> zmk.position_state_changed_listener: 0 bubble (no undecided hold_tap active)[0m
[00:00:09.845,520] [0m<dbg> zmk.zmk_keymap_apply_position_state: layer: 0 position: 0, binding name: KEY_PRESS[0m
[00:00:09.845,550] [0m<dbg> zmk.on_keymap_binding_pressed: position 0 keycode 0x70027[0m

```
They output on the same millisecond, but `1` definitely comes before `j`, as expected. Will update the timing diagram to reflect this.

kurtis-lew: Draft 4 (Main)@1 5x

Made some small tweaks to the timing diagram according to your suggestions. If you have any further comments, feel free to share them.

kurtis-lew: I realize that the `tap, tap, tap` is missing an additional extension to the dashed red line indicating the cancelled tapping-term from the third press of `td`. This updated image would be more accurate.
Draft 4 (Main)@1 5x

caksoylar: > Made some small tweaks to the timing diagram according to your suggestions. If you have any further comments, feel free to share them.

This looks good to me, thanks!

> I realize that the tap, tap, tap is missing an additional extension to the dashed red line indicating the cancelled tapping-term from the third press of td. This updated image would be more accurate.

I see, tapping term gets refreshed after the second tap's press so this shows that extension. I assume all timers are killed after the third tap anyway since tap-dance only has three items: I guess it is still good to illustrate the extension in this case.

kurtis-lew: > I am a bit confused about the interrupt behavior as described here -- I wonder if this could be made clearer.
> 
> Modifiers like `LSHIFT` are technically key press behaviors as well, so there is a contradiction with the first sentence. In addition the last two sentences seem to imply there is special logic for modifiers, if that is the case we should state it explicitly.

```
Alphanumeric [`key press`](key-press.md) bindings, like those used for `td0`,
will release as soon as an interrupting alphanumeric key press occurs.
If a modifier key like `LSHIFT` were to replace the `N1` binding, it would
remain pressed until `td0`'s binding is released.
That is to say, the output would instead be `J` and any following alphanumeric
key presses would be capitalized as long as `&td0` is held down.
```
Just got around to addressing your concerns regarding the wording surrounding key presses too. How does this look?

caksoylar: I think that's better, thanks! In "as soon as an interrupting alphanumeric key press occurs", is the fact that the interrupting key press is alphanumeric important? If not we can remove that to simplify. Also it would be good to point out that this example refers to the last case in the diagram, e.g. we can augment the last part to be

> For instance, if a modifier key like `LSHIFT` were to replace the `N1` binding in the last example above, it would
remain pressed until `td0`'s binding is released and the output would instead be `J`. Any following alphanumeric
key presses would also be shifted while `&td0` is held down.

kurtis-lew: Specification of the alphanumeric nature of the keypress is important because like you said previously, modifiers are technically keypresses too. I also realized that due to the limited scope of our findings this far, I'm not sure if other types of behavior bindings interrupt tap-dances, so that might be something for other testers to discover as well.

kurtis-lew: Actually, I think I see your point a little better. I used "alphanumeric" twice in the same sentence which reads a little weird. We probably don't need the second copy of the word for readability.

docs/docs/behaviors/tap-dance.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Tap-Dance Behavior
+sidebar_label: Tap-Dance
+---
+
+## Summary
+
+A tap-dance key invokes a different behavior (e.g. `kp`) corresponding to how many times it is pressed.
+Tap-dances are completely custom, so for every unique tap-dance key,
+a new tap-dance must be defined in your keymap's `behaviors`.
```
dxmh, line 10: Nice work! 

I think a brief example could be useful here, before getting into the technical stuff:

```suggestion
A tap-dance key invokes a different behavior (e.g. `kp`) corresponding to how many times it is pressed. For example, you could configure a tap-dance key that toggles [Caps Word](../caps-word) if tapped once, or Caps _Lock_ if tapped twice.

Tap-dances are completely custom, so for every unique tap-dance key,
a new tap-dance must be defined in your keymap's `behaviors`.
```

I'm not sure if that's the best example, but I tried to think of something practical for those who might not understand the potential use cases of tap-dance.

kurtis-lew: > For example, you could configure a tap-dance key that toggles
> [Caps Word](../caps-word) if tapped once, or Caps _Lock_ if tapped twice.
> The expandability of the number of [`bindings`](#bindings) attached to a
> particular tap-dance is a great way to add more functionality to a single key,
> especially for keyboards with a limited number of keys.

Expanding upon your work here with an extra qualitative statement before the technical stuff

okke-formsma:
I'm a little confused why there is a need for multiple tap dances to be active at the same time - I would expect that a tap dance resolves as soon as another key is pressed.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    // Timer Data
+    bool timer_started;
+    bool timer_cancelled;
+    bool tap_dance_decided;
+    int64_t release_at;
+    struct k_delayed_work release_timer;
+};
+
+// Array to store all active_tap_dances
+struct active_tap_dance active_tap_dances[ZMK_BHV_TAP_DANCE_MAX_HELD] = {};
```
okke-formsma, line 52: I don't think this active_tap_dances is necessary, as there is only a single tap-dance that can be active at the same time?

If multiple tap dances can be active at the same time, please write a test showing behavior.

okke-formsma: Oh, right... We need this information when the behavior is being released.

app/tests/tap-dance/3b-hold-int/native_posix.keymap
```diff
@@ -0,0 +1,12 @@
...
+#include <behaviors.dtsi>
+#include <dt-bindings/zmk/kscan_mock.h>
+#include "../behavior_keymap.dtsi"
+
+&kscan {
+	events = <
+		ZMK_MOCK_PRESS(1,0,400)
+        ZMK_MOCK_PRESS(1,1,10)
+        ZMK_MOCK_RELEASE(1,1,10)
+	>;
```
okke-formsma, line 11: most tests include the release of all keys at the end - I'd add a `ZMK_MOCK_RELEASE(1,0,100)` at the end.

app/tests/tap-dance/behavior_keymap.dtsi
```diff
@@ -0,0 +1,45 @@
...
+        };
+    };
+
+	keymap {
+		compatible = "zmk,keymap";
+		label ="Default keymap";
+
+		default_layer {
+			bindings = <
+				&tdb   &kp N0
```
okke-formsma, line 41: `&tdb` is not exercised in any test?

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+        press_tap_dance_behavior(tap_dance, tap_dance->release_at);
+        tap_dance->tap_dance_decided = true;
+        // Tap dance is currently pressed, prepare for release of binding
+        // before actual behavior release
+        if (tap_dance->is_pressed)
+            return;
+        // Tap dance is not currently pressed. Release the tap dance.
+        LOG_DBG("Tap dance releases immediately! Counter reached: %d", tap_dance->counter);
+        release_tap_dance_behavior(tap_dance, tap_dance->release_at);
+        return;
```
okke-formsma, line 258: unnecessary return

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+
+// Determines what happens when a tap-dance's timer expires
+void behavior_tap_dance_timer_handler(struct k_work *item) {
+    // Timer for tap dance has ran out.
+    struct active_tap_dance *tap_dance = CONTAINER_OF(item, struct active_tap_dance, release_timer);
+    if (tap_dance->position == ZMK_BHV_TAP_DANCE_POSITION_FREE) {
+        return;
+    }
+    if (tap_dance->timer_cancelled) {
+        tap_dance->timer_cancelled = false;
```
okke-formsma, line 247: if you return in this block you don't need the else and it's more in line with the rest of the coding standard

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    } else {
+        press_tap_dance_behavior(tap_dance, tap_dance->release_at);
+        tap_dance->tap_dance_decided = true;
+        // Tap dance is currently pressed, prepare for release of binding
+        // before actual behavior release
+        if (tap_dance->is_pressed)
+            return;
```
okke-formsma, line 254: our coding style asks for `{}` around all code blocks

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    LOG_DBG("%d tap-dance ms_left = %d", tap_dance->position, ms_left);
+    if (ms_left > 0) {
+        k_delayed_work_submit(&tap_dance->release_timer, K_MSEC(ms_left));
+    }
+}
+
+// press_tap_dance_behavior
+
+// Returns behavior_keymap_binding_pressed to ensure that keycode is actually pressed
+static inline int press_tap_dance_behavior(struct active_tap_dance *tap_dance, int64_t timestamp) {
```
okke-formsma, line 140: I'd set `press_tap_dance_behavior->behavior_decided` to true in this method, as all call sites now have to make sure they update the value

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+            ref_dance->tap_dance_decided = false;
+            *tap_dance = ref_dance;
+            return 0;
+        }
+    }
+    // Unable to store new tap-dance in array active_tap_dances[].
+    return -ENOMEM;
+}
+
+// clear_tap_dance
```
okke-formsma, line 97: not sure what the function of these comments is

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    if (tap_dance == NULL) {
+        LOG_ERR("ACTIVE TAP DANCE CLEARED TOO EARLY");
+        return ZMK_BEHAVIOR_OPAQUE;
+    }
+    tap_dance->is_pressed = false;
+    // An active tap dance has already been decided and is registering keypresses.
+    // Release tap dance on binding release.
+    if (tap_dance->tap_dance_decided) {
+        release_tap_dance_behavior(tap_dance, event.timestamp);
+        tap_dance->tap_dance_decided = false;
```
okke-formsma, line 232: `release_tap_dance_behavior` already calls `clear_tap_dance` so this is unnecessary

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    struct zmk_position_state_changed *ev = as_zmk_position_state_changed(eh);
+    if (ev == NULL) {
+        return ZMK_EV_EVENT_BUBBLE;
+    }
+    for (int i = 0; i < ZMK_BHV_TAP_DANCE_MAX_HELD; i++) {
+        struct active_tap_dance *tap_dance = &active_tap_dances[i];
+        if (tap_dance->position == ZMK_BHV_TAP_DANCE_POSITION_FREE) {
+            continue;
+        }
+        if (tap_dance->position != ev->position && tap_dance->timer_started) {
```
okke-formsma, line 289: if you invert the check in this if to the following, the code has a more consistent style
```
if (tap_dance->position == ev->position && tap_dance->timer_started) {
    continue;
}
```

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+            // Press tap dance immediately before new keypress.
+            stop_timer(tap_dance);
+            LOG_DBG("Tap dance interrupted, position changed. %d %d", tap_dance->position,
+                    ev->position);
+            if (!tap_dance->tap_dance_decided) {
+                tap_dance->tap_dance_decided = true;
+                // Ensure tap-dance resolves before next keycode
+                ZMK_EVENT_RAISE(press_tap_dance_behavior(tap_dance, ev->timestamp));
+                ZMK_EVENT_RAISE(eh);
+                return ZMK_EV_EVENT_CAPTURED;
```
okke-formsma, line 299: instead of
```
               ZMK_EVENT_RAISE(eh);
                return ZMK_EV_EVENT_CAPTURED;
```

I think you could simply
```
                return ZMK_EV_EVENT_BUBBLE;
```

okke-formsma: This made me realize there's probably a bug in hold-tap which does the same 😨

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    return ZMK_EV_EVENT_BUBBLE;
+}
+
+static int behavior_tap_dance_init(const struct device *dev) {
+    static bool init_first_run = true;
+    if (init_first_run) {
+        for (int i = 0; i < ZMK_BHV_TAP_DANCE_MAX_HELD; i++) {
+            k_delayed_work_init(&active_tap_dances[i].release_timer,
+                                behavior_tap_dance_timer_handler);
+            active_tap_dances[i].position = ZMK_BHV_TAP_DANCE_POSITION_FREE;
```
okke-formsma, line 314: call `clear_tap_dance` instead of setting the position here manually. That way tapdances are always cleared in the same way.

okke-formsma:
The overall architecture and implementation seems pretty great! I have added a bunch of comments for minor nitpicks.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+// for first element with FREE position. Set its variables to default
+// values as to overwrite any settings from previous tap-dances.
+// Returns -ENOMEM if unable to store tap-dance, 0 otherwise.
+static int new_tap_dance(uint32_t position, const struct behavior_tap_dance_config *config,
+                         struct active_tap_dance **tap_dance) {
+    // Create a new instance of a tap-dance
+    for (int i = 0; i < ZMK_BHV_TAP_DANCE_MAX_HELD; i++) {
+        struct active_tap_dance *const ref_dance = &active_tap_dances[i];
+        if (ref_dance->position == ZMK_BHV_TAP_DANCE_POSITION_FREE) {
+            ref_dance->counter = 1;
```
okke-formsma, line 81: if you start counting at 0, you don't have to subtract 1 everywhere else.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+        LOG_DBG("%d created new tap dance", event.position);
+        reset_timer(tap_dance, event);
+        return ZMK_BEHAVIOR_OPAQUE;
+    }
+    // An active tap dance was found at the keypress location. Stop the timer.
+    tap_dance->is_pressed = true;
+    LOG_DBG("%d tap dance re-pressed", event.position);
+    stop_timer(tap_dance);
+    // Increment the counter. Check if the counter has reached the maximum behavior number.
+    tap_dance->counter++;
```
okke-formsma, line 200: if you check the maximum before incrementing the count, (e.g., move this line inside of the if statement and modify the if statement slightly) you don't run the risk of having a higher counter than there are behavior_counts, removing the need to use `MIN(tap_dance->counter, tap_dance->config->behavior_count) - 1)` everywhere.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+
+// release_tap_dance_behavior
+
+// Returns behavior_keymap_binding_pressed to ensure that keycode is actually released
+static inline int release_tap_dance_behavior(struct active_tap_dance *tap_dance,
+                                             int64_t timestamp) {
+    // Release chosen tap dance binding
+    struct zmk_behavior_binding binding =
+        tap_dance->config
+            ->behaviors[MIN(tap_dance->counter, tap_dance->config->behavior_count) - 1];
```
okke-formsma, line 160: In what scenario is it possible for `counter` to become greater than `behavior_count`?  I think the rest of the code should make sure that `counter` can not exceed `behavior_count`, instead of defensively preventing that from occurring here.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    const struct behavior_tap_dance_config *cfg = dev->config;
+    struct active_tap_dance *tap_dance;
+    tap_dance = find_tap_dance(event.position);
+    // No active tap dance found on keypress location, attempt to create a new one
+    if (tap_dance == NULL) {
+        if (new_tap_dance(event.position, cfg, &tap_dance) == -ENOMEM) {
+            LOG_ERR("Unable to create new tap dance. Insufficient space in active_tap_dances[].");
+            return ZMK_BEHAVIOR_OPAQUE;
+        }
+        // Tap dance successfully created. Reset timer.
```
okke-formsma, line 190: I would remove this entire block of code and the return, so the same flow is taken for both the first key press and any consecutive presses.

```
        // Tap dance successfully created. Reset timer.
        LOG_DBG("%d created new tap dance", event.position);
        reset_timer(tap_dance, event);
        return ZMK_BEHAVIOR_OPAQUE;
```

(I think there is a bug in the current implementation of the tap-dance - tap_dance->is_pressed is not set to true on the first keypress, causing the key to be released when the timer expires instead of on release).

okke-formsma: Actually there is a bug - Tap dances with 1 behavior will never trigger the behavior keypress because we're returning early here.

app/tests/tap-dance/behavior_keymap.dtsi
kurtis-lew: `&tdb` is not used in testing since `&tdm` gives a better indication of held keys working properly with tap-dances. Further updates on my local repo remove `&tdb` and sets `&tdm` to the `row 0` `col 0` position (position 0) in the keymap.

okke-formsma: I think you need to have multiple tapdances in the keymap to prove there are no bugs when two are used shortly after each other.

app/src/behaviors/behavior_tap_dance.c
kurtis-lew: Confirming that the single-keypress bug is resolved in updated squashed commit 22118f5

app/src/behaviors/behavior_tap_dance.c
kurtis-lew: Using `ZMK_EVENT_RAISE` caused segmentation faults according to testing. `return ZMK_EV_EVENT_BUBBLE` has been implemented as suggested, allowing all tests to function as expected.
It was originally hypothesized that `ZMK_EVENT_RAISE` was necessary to ensure that the tap-dance being interrupted would finish first, although it was revealed in testing that the issues of tap-dances in progress that involved hold-taps would require the hold-taps to be decided first, causing the anomalies in timing.

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -0,0 +1,340 @@
...
+    bool timer_cancelled;
+    bool tap_dance_decided;
+    int64_t release_at;
+    struct k_delayed_work release_timer;
+};
+
+// Array to store all active_tap_dances
+struct active_tap_dance active_tap_dances[ZMK_BHV_TAP_DANCE_MAX_HELD] = {};
+
+// *find_tap_dance
```
okke-formsma, line 54: these extra comments with the function names should be removed, as they are not according to our style guide.

kurtis-lew: Comments have been removed in commit [218c1c3](https://github.com/zmkfirmware/zmk/commit/218c1c3f7e52b602fe46dab6432f1875c6aa2b7b)

okke-formsma (approved):
I think this is good enough to be merged.

You may consider cleaning up the comments. The comments before the functions repeating the function name are quite unnecessary and not in line with our code style. Additionally, quite a few comments just repeat what the code does (increment counter, etc). Try to only keep comments that explain /why/ something is done, not /how/ (unless it's especially tricky).

petejohanson (approved):
Thanks so much for working on this! Great contributions.

PR #1140, opened 2022-02-21
Boot protocol support

### LINKED ISSUE #913: ZMK Lacks USB HID Boot Protocol (Needed For FileVault or BIOS)
If FileVault (full disk encryption) is turned on in macOS, ZMK over USB currently cannot be used on the boot login screen before the disk is decrypted. This is because ZMK does not implement the boot HID protocol for USB and FileVault does not use the report protocol.

rbitanga:
This PR makes ZMK optionally compliant with the USB HID v1.11 spec per `Appendix F.3` except for:
- indicator LED handling
- `Get_Descriptor` requests for the configuration descriptor. Zephyr 2.5 handles `HID_CLASS_DESCRIPTOR_HID` and `HID_CLASS_DESCRIPTOR_REPORT` but doesn’t expose a means to return the configuration descriptor.

This will fix #913 

6KRO configurations will be the most efficient. NKRO and HKRO with report sizes not equal to 6 have overhead due to `zmk_hid_keyboard_report_body` not matching the expected structure. Boot/report protocol usage is specific to USB and logic was moved from `endpoints.c` to `usb.c` and `hog.c` to localize handling of the required reports.

Handling non-spec conformant BIOS implementations (e.g. those that expect the keyboard to start in boot protocol mode) is beyond the scope of this PR. Multiple strategies exist for handling non-conformant BIOS implementations in conjunction with NKRO (e.g. declaring bytes 3-7 to be padding in the report descriptor, having separate endpoints, etc.) but are not handled by this PR.

This has been tested with a 2018 Mac Mini, 2021 MBP Max, and Asus  ROG Strix X570-I Gaming motherboard. Tested with 5KRO, 6KRO, and NKRO configurations.


app/include/zmk/hid.h
```diff
@@ -237,5 +255,5 @@ int zmk_hid_consumer_press(zmk_key_t key);
 int zmk_hid_consumer_release(zmk_key_t key);
 void zmk_hid_consumer_clear();
 
-struct zmk_hid_keyboard_report *zmk_hid_get_keyboard_report();
-struct zmk_hid_consumer_report *zmk_hid_get_consumer_report();
+uint8_t *zmk_hid_get_keyboard_report(zmk_hid_report_request_t req_t, uint8_t proto);
```
rbitanga, line 258: I think I'd be happier with `int zmk_hid_get_keyboard_report(zmk_hid_report_request_t req_t, uint8_t proto, uint8_t **report, size_t *len)`. All the code in usb.c and hog.c care about is there's n bytes of data and data length calculation is better served inside of hid.c.

typorian:
Is there an easy way for me to integrate this in my github action? It would be really useful if I could use the keyboard to enter the bitlocker pin (I hope that is what this fixes)

caksoylar:
> Is there an easy way for me to integrate this in my github action? It would be really useful if I could use the keyboard to enter the bitlocker pin (I hope that is what this fixes)

See the instructions at https://zmk.dev/docs/features/beta-testing, but note that this branch is based off of an older ZMK revision and it doesn't include features like macros, tap dances, hold-tap fixes etc.

moergo-sc:
Just like to add our vote on this PR. It is a super important PR for ZMK in terms of general usability, 
in our opinion. Thanks rbitanga.

It appears some bit-rot has set in. What would it take to get this PR accepted into mainline ZMK?

rbitanga:
Spent part of labor day to pull in several months worth of commits into the branch. #1025 broke USB connections for me because the default changed from `ZMK_USB_CONN_HD` to `ZMK_USB_CONN_POWERED` and `USB_DC_SOF` wasn't handled so that was handled here. Unlike my earlier tests, I've only tested this with 6KRO and my M1 Max MBP. If someone else is willing to test and validate NKRO and HKRO as well as functionality in PCs that would be great

chrisandreae:
Have you tested with in the UEFI on your Mac, for example on the Filevault unlock screen? We were also looking at rebasing these changes over the weekend, and discovered that in addition to supporting boot protocol and the `Get_Report` class request, to work in the the mac pre-boot environment additionally required disabling the OUT interrupt endpoint. https://github.com/moergo-sc/zmk/commit/d6b135521eff1b170634b66d2202de9e5d381b3a

To go into a bit more detail: it's very uncommon for keyboards to use an interrupt endpoint for receiving LED reports from the host, because it happens so rarely that it's not worth consuming the USB bandwidth with an interrupt time slot. It's more common to have only an IN endpoint, and then LED states are set by `Set_Report` class requests. Because it's uncommon, bios/embedded environments often don't handle it correctly. [This Linux commit](https://github.com/torvalds/linux/commit/d7428bc26fc767942c38d74b80299bcd4f01e7cb) goes into more detail, but in short they can not only fail to receive the interrupt reports, but some even fail to use the device entirely. Macs' UEFI environment is like this: it fails to work at all, unless you flip the order of the in and out endpoints in the USB descriptor (which would be a spec violation, and Zephyr wouldn't make easy anyway).

app/include/zmk/hid.h
chrisandreae: We followed this suggestion when refactoring the patch here: https://github.com/moergo-sc/zmk/commit/8f078c8cc1b296aa8e0cb772e79f4b6edd183b12. Note that that branch isn't yet fully tested; there may still be some regressions.

chrisandreae:
Incidentally, could you go into more detail why SOF events are required to handle idle events? I was expecting to see an `on_idle` callback that submitted a report when the timeout occurred, similar to Zephyr's example code here: https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/usb/hid/src/main.c#L92-L96. 

I do however notice that that sample project also enables SOF events. Does Zephyr's implementation of idle internally require counting SOF events to determine the idle timeout?

rbitanga:
> Have you tested with the UEFI on your Mac, for example on the Filevault unlock screen? We were also looking at rebasing these changes over the weekend, and discovered that in addition to supporting boot protocol and the `Get_Report` class request, to work in the the mac pre-boot environment additionally required disabling the OUT interrupt endpoint. [moergo-sc@d6b1355](https://github.com/moergo-sc/zmk/commit/d6b135521eff1b170634b66d2202de9e5d381b3a)
> 

Yes, I specifically made these changes to work with the FileVault unlock screen of a Mac Mini which I no longer have. I typically only suspend my MBP but it was working when I tried it after updating my branch. I don't have indicator LEDs on my keyboard so I never bothered looking into the OUT interrupt endpoint. You're more likely to run into issues with PCs with the wide variety of BIOS implementations that you'll need to support.

>Incidentally, could you go into more detail why SOF events are required to handle idle events? I was expecting to see an on_idle callback that submitted a report when the timeout occurred, similar to Zephyr's example code here: https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/usb/hid/src/main.c#L92-L96.

Appendix C on page 62 of the [HID spec](https://www.usb.org/sites/default/files/hid1_11.pdf) requires keyboards support the Idle request but [Zephyr returns -ENOTSUP for `Get_Idle` and `Set_Idle` unless `USB_DEVICE_SOF` is enabled](https://github.com/zephyrproject-rtos/zephyr/blob/e9e7ad388da7122a6193845be4eff48c19bbe618/subsys/usb/device/class/hid/core.c#L158).

I don't think supporting idle requests is specific to boot keyboards and it very well may not be necessary in most cases but I enabled it to adhere to the spec as closely as possible.

chrisandreae:
> I don't have indicator LEDs on my keyboard so I never bothered looking into the OUT interrupt endpoint. 

Ah, right, I hadn't noticed that the out endpoint was actually added by PR #999, which I'd already integrated. I'll make a comment over there about adding that endpoint being dangerous in BIOS-like environments.

chrisandreae:
> Appendix F.3 on page 74 of the [HID spec](https://www.usb.org/sites/default/files/hid1_11.pdf) requires keyboards support the Idle request but [Zephyr returns -ENOTSUP for `Get_Idle` and `Set_Idle` unless `USB_DEVICE_SOF` is enabled](https://github.com/zephyrproject-rtos/zephyr/blob/e9e7ad388da7122a6193845be4eff48c19bbe618/subsys/usb/device/class/hid/core.c#L158).

I think this is missing one of two parts of this equation - to support idle behaviour:
* The keyboard needs to support responding to the `Get_Idle` and `Set_Idle` class requests: this is handled by Zephyr (which requires `USB_DEVICE_SOF`, something that I'd missed)
* If a non-zero idle time has been set, whenever the configured idle period expires, the keyboard needs to send another report, even if there have been no changes since the last report (see details in section 7.2.4 pp52-53). This re-sending of the report, if I understand correctly, is not handled internally by Zephyr: it looks like when the idle timeout has expired, Zephyr simply [calls the application's onIdle callback](https://github.com/zephyrproject-rtos/zephyr/blob/e9e7ad388da7122a6193845be4eff48c19bbe618/subsys/usb/device/class/hid/core.c#L341).

So I think we still need to define an `onIdle` callback which schedules a `zmk_usb_hid_send_report`, similar to the [Zephyr example](https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/usb/hid/src/main.c#L92-L96).

thefloweringash:
Hi @rbitanga, I've been working with @chrisandreae on updating this patch. I think I was probably working on it around the same time you were. I hope I'm not stepping on your toes here. I made a few changes that I think are improvements:
- Minimize the api changes for bluetooth by keeping `zmk_hid_get_keyboard_report()` and returning the `.body` like before.
- Keeping the logic about "full" vs "body" entirely within the usb code path. 
  - The endpoint now calls either `zmk_usb_hid_send_keyboard_report()` and `zmk_usb_hid_send_consumer_report()` instead of `zmk_usb_hid_send_report(...)`. 
  - These functions now internally dispatch on protocol to either `zmk_hid_get_keyboard_report()` or a new function for boot reports `zmk_hid_get_boot_report()`. 
  - There's a change here from your version in the behavior of `get_report()`: when in boot protocol, don't include the report id. I believe this is correct, in that implicit boot protocol report descriptor does not use multiple reports, so any boot protocol report should not include the report id field.
- Use a single definition of a boot report, used for regular reports in non-6KRO, and for rollover in all branches.
- I've also included your new fix for SOF support, and using the observer to reset the state on startup.
- Handle gaps in the zmk report when producing a boot report in HKRO mode. For example, if it was 8KRO, it would be possible to have the state `0 0 0 0 0 0 0 17` (by pressing 8 keys, and letting go of the first 7). Copying the first 6 bytes would not show up the single pressed key.

[Comparison view](https://github.com/zmkfirmware/zmk/compare/main...thefloweringash:boot?expand=1).

typorian:
not sure if this is useful, but I merged the PR into the fork in my repo and used the beta testing instructions to build the firmware. It doesn't work with the bitlocker pin on my thinkpad yoga. If there is additional testing I could do, I'll do my best.

app/src/usb_hid.c
```diff
@@ -61,4 +136,16 @@ static int zmk_usb_hid_init(const struct device *_arg) {
     return 0;
 }
 
+#if IS_ENABLED(CONFIG_ZMK_USB_BOOT)
+static int usb_hid_listener(const zmk_event_t *eh) {
+    if (as_zmk_usb_conn_state_changed(eh) && zmk_usb_get_status() == USB_DC_RESET) {
```
chrisandreae, line 141: I think there's a race condition here: by the time this event is delivered to the listener, the status is no longer guaranteed to be `USB_DC_RESET`, even if that was the status change that triggered the event to be sent. In practice, it doesn't work for me at all, while the previous version that set the protocol directly in `usb_status_cb` did.

joelspadin: That seems like a reasonable theory. `usb_status_cb()` sets `usb_status` on some arbitrary thread before scheduling this event on the main thread. This seems bad, as `usb_status` may be accessed by different threads with no synchronization.

A couple ideas for improving this:

1. Do not directly set `usb_status` inside `usb_status_cb()`. Instead pass the status to the main thread with a [FIFO or message queue](https://docs.zephyrproject.org/3.0.0/reference/kernel/index.html#data-passing), then read it on the main thread and set `usb_status` there immediately before sending the `zmk_usb_conn_state_changed` event.
2. Add a `usb_dc_status_code` field to `struct zmk_usb_conn_state_changed` so you can query it directly from the event instead of calling `zmk_usb_get_status()` (though doing 1 would make this unnecessary to prevent a race condition).

Na-Cly:
Any chance on this getting merged anytime soon?

app/src/hid.c
```diff
@@ -122,13 +184,33 @@ static inline bool check_keyboard_usage(zmk_key_t usage) {
         }                                                                                          \
     }
 
+#if IS_ENABLED(CONFIG_ZMK_USB_BOOT)
+static zmk_hid_boot_report_t *get_boot_report() {
+    HID_CHECK_ROLLOVER_ERROR();
+#if CONFIG_ZMK_HID_KEYBOARD_REPORT_SIZE != HID_BOOT_KEY_LEN
+    boot_report.modifiers = keyboard_report.body.modifiers;
+    int len = CONFIG_ZMK_HID_KEYBOARD_REPORT_SIZE < HID_BOOT_KEY_LEN ? CONFIG_ZMK_HID_KEYBOARD_REPORT_SIZE : HID_BOOT_KEY_LEN;
+    memcpy(&boot_report.keys, keyboard_report.body.keys, len);
```
joelspadin, line 193: If using memcpy, it would also be a good idea to `BUILD_ASSERT(sizeof(boot_report.keys[0]) == sizeof(keyboard_report.body.keys[0]), ...)` to catch if someone changes these types to be incompatible.

app/src/hid.c
```diff
@@ -82,15 +93,63 @@ int zmk_hid_unregister_mods(zmk_mod_flags_t modifiers) {
...
 #if IS_ENABLED(CONFIG_ZMK_HID_REPORT_TYPE_NKRO)
 
 #define TOGGLE_KEYBOARD(code, val) WRITE_BIT(keyboard_report.body.keys[code / 8], code % 8, val)
 
+#if IS_ENABLED(CONFIG_ZMK_USB_BOOT)
+static zmk_hid_boot_report_t *get_boot_report() {
+    HID_CHECK_ROLLOVER_ERROR();
+
+    boot_report.modifiers = keyboard_report.body.modifiers;
+    memset(&boot_report.keys, 0, HID_BOOT_KEY_LEN);
```
joelspadin, line 124: This will break if `keys` is changed to not be one byte per element.
```suggestion
    memset(&boot_report.keys, 0, sizeof(boot_report.keys));
```

app/src/hid.c
```diff
@@ -82,15 +93,63 @@ int zmk_hid_unregister_mods(zmk_mod_flags_t modifiers) {
...
+    int ix = 0;
```
joelspadin, line 125: Not sure what to call this, but I recommend finding a more descriptive name. Having both `i` and `ix` is a bit confusing.

app/src/hid.c
```diff
@@ -82,15 +93,63 @@ int zmk_hid_unregister_mods(zmk_mod_flags_t modifiers) {
...
+    uint8_t base_code = 0;
+    for (int i = 0; i < (ZMK_HID_KEYBOARD_NKRO_MAX_USAGE + 1) / 8; ++i) {
+        if (ix == keys_held) {
+            break;
+        }
+        if (!keyboard_report.body.keys[i]) {
+            continue;
+        }
+        base_code = i * 8;
```
joelspadin, line 134: This isn't used outside of this loop. It could be declared here instead of line 126.

app/src/usb_hid.c
```diff
@@ -15,19 +15,73 @@
...
+
+    uint8_t *report = NULL;
+    switch (setup->wValue & HID_GET_REPORT_ID_MASK) {
+        case HID_REPORT_ID_KEYBOARD:
+            report = zmk_hid_get_keyboard_report(HID_REPORT_FULL, HID_PROTOCOL_REPORT);
+            *len = sizeof(struct zmk_hid_keyboard_report);
+            break;
+        case HID_REPORT_ID_CONSUMER:
+            report = zmk_hid_get_consumer_report(HID_REPORT_FULL);
+            *len = sizeof(struct zmk_hid_consumer_report);
```
joelspadin, line 52: This is making some assumptions about the return values that might easily be changed to not be true. If these functions are going to return byte buffers instead of pointers with known types, then they should really provide the sizes to go along with them. As I mentioned above though, I feel like it would make more sense to just return pointers with types so you can directly get the size from the data.

app/src/hid.c
```diff
@@ -246,10 +328,17 @@ bool zmk_hid_is_pressed(uint32_t usage) {
     return false;
 }
 
-struct zmk_hid_keyboard_report *zmk_hid_get_keyboard_report() {
-    return &keyboard_report;
+uint8_t *zmk_hid_get_keyboard_report(zmk_hid_report_request_t req_t, uint8_t proto) {
+    if (proto == HID_PROTOCOL_REPORT) {
+        return req_t == HID_REPORT_FULL ? (uint8_t*)&keyboard_report : (uint8_t*)&keyboard_report.body;
```
joelspadin, line 333: I feel like it would be simpler to just have separate functions for getting the full report and just the body. Every location this gets used, the value of `req_t` is hard-coded. Similarly, there's only one place that actually dynamically sets the value of `proto`. 

Maybe it would be better to keep these functions returning typed pointers, and in the one place that needs to switch on the protocol, just do something like this?

```c
void* data;
size_t size;
if (protocol == HID_PROTOCOL_REPORT) {
  data = zmk_hid_get_keyboard_report();
  size = sizeof(*zmk_hid_get_keyboard_report());
} else {
  data = get_boot_report();
  size = sizeof(*zmk_hid_get_boot_report());
}
```

app/src/hid.c
```diff
@@ -246,10 +328,17 @@ bool zmk_hid_is_pressed(uint32_t usage) {
     return false;
 }
 
-struct zmk_hid_keyboard_report *zmk_hid_get_keyboard_report() {
-    return &keyboard_report;
+uint8_t *zmk_hid_get_keyboard_report(zmk_hid_report_request_t req_t, uint8_t proto) {
```
joelspadin, line 331: `req_t` looks more like a type name than a parameter name?

app/src/usb_hid.c
```diff
@@ -38,6 +92,27 @@ int zmk_usb_hid_send_report(const uint8_t *report, size_t len) {
         return -ENODEV;
     default:
         k_sem_take(&hid_sem, K_MSEC(30));
+        uint8_t *report;
+        size_t len;
+
+        if (report_id == HID_REPORT_ID_KEYBOARD) {
+#if IS_ENABLED(CONFIG_ZMK_USB_BOOT)
+            report = zmk_hid_get_keyboard_report(HID_REPORT_FULL, hid_protocol);
```
joelspadin, line 100: Same as above. If the implementations of these functions were changed, this code could easily become wrong without causing any obvious compiler error.

joelspadin (changes requested):
Gave this a more thorough review since someone was asking about it on Discord.

chrisandreae:
I think it was missed [earlier](https://github.com/zmkfirmware/zmk/pull/1140#issuecomment-1237655563), but @thefloweringash took a stab at refactoring this PR here: https://github.com/zmkfirmware/zmk/commit/b599aadb629b85c72a8fdc2548a2b2f8b8bb9919.

The changes made are described in the linked commit and the earlier comment, but overall they're aimed to simplify the API changes and improve correctness without substantially changing the approach.

joelspadin:
Ah. I did miss that. I'll give that a review once it's been submitted as a PR then.

petejohanson:
Merged in via #1793 

Thanks for the work on this!

Issue #1076, opened 2021-12-27
Feature Request: Better handling of keys with the same keycode but different modifiers


rayduck:
A common bigram in programming is `+=`, which may exist as two separate keys on a custom ZMK keyboard layout. Rolling the keys `+` and `=` will result in one of the key being ignored.

Underlying `+` and `=` is the same key, with the former under the shift modifier and latter under no modifiers. When a user types `+`, a keydown of `+` i.e a keydown of `Shift` then `=` is sent. If the user then presses the `=` key before `+` is released, it will be treated as a duplicate keydown event. The end result is `+` being typed, quite contrary to what the user have in mind (`+=`). Typing `=+` in the same manner also results in `=`, which I suppose reflects the sequence `=` keydown, `Shift` keydown, `=` keydown (ignored as `=` is already held down).

On a normal keyboard each key is, well, just one key. But the extensibility of custom keymaps means that the same keycode can exist in different forms (i.e. with different mods) on different key locations.

QMK has [fixed](https://github.com/qmk/qmk_firmware/pull/2710) this by forcing a key (and held modifiers) release when a new key press for the same keycode is registered.

innhyu:
Definitely something I hope is addressed. Doing any variable adds with += results in the = sign getting ignored!

pujvz:
I also hope that this is addressed at some point. I would like to help, but I do not know how.

aaronkollasch:
A solution that seems to work is to add a [key combo](https://zmk.dev/docs/features/combos) for the '+' and '=' keys, and bind it to a macro that taps '+' and '='. The combo can be specific to your symbol layer as well.

caksoylar:
Closing since #1828 solves this.

PR #1142, opened 2022-02-21
Marcoster/behavior key tap

marcoster:
This pull request implements a key-tap behavior which immediately sends a released event immediately after the pressed event.

Background: I made a keymap for myself which had
`&kp LBRC &kp LBKT &kp RBKT &kp RBRC`
directly side-by side. This way it when holding i.e. LBRC while using LBKT the keypress was not recognized.

In #1076 a similar problem with += which is commonly used was reported.


okke-formsma:
LGTM

app/src/behaviors/behavior_key_tap.c
```diff
@@ -0,0 +1,47 @@
...
+
+static int behavior_key_tap_init(const struct device *dev) { return 0; };
+
+static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
+                                     struct zmk_behavior_binding_event event) {
+    LOG_DBG("position %d keycode 0x%02X", event.position, binding->param1);
+    int ret = 0;
+    ret = ZMK_EVENT_RAISE(zmk_keycode_state_changed_from_encoded(binding->param1, true, event.timestamp));
+    if(ret != 0) {
+        return 0;
```
okke-formsma, line 27: probably want to return `ret` here instead of 0

app/src/behaviors/behavior_key_tap.c
```diff
@@ -0,0 +1,47 @@
...
+#include <zmk/behavior.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+static int behavior_key_tap_init(const struct device *dev) { return 0; };
+
+static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
+                                     struct zmk_behavior_binding_event event) {
+    LOG_DBG("position %d keycode 0x%02X", event.position, binding->param1);
+    int ret = 0;
```
okke-formsma, line 24: simpler to combine this line with the next;

`int ret = ZMK_EVENT_RAISE(zmk_keycode_state_changed_from_encoded(binding->param1, true, event.timestamp));`

marcoster:
Sorry for the late response - both absolutely correct. Must have been the late hour and excitement that it worked (adding such things to zephyr was kinda new for me).

marcoster:
Rebased and applied clang-format on the failed file.

dxmh:
Would it be possible to get a section in the docs for this new functionality please @marcoster?

docs/docs/behaviors/key-tap.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Key Tap Behavior
+sidebar_label: Key Tap
+---
+
+## Summary
+
+The key tap behavior is a combination of key press and release on just pressing the key.
```
dxmh, line 8:
```suggestion
The key tap behavior is similar to [key press](key-press.md), but key tap will press _and then immediately release_ the key (even if the physical key remains held down).
```

docs/docs/behaviors/key-tap.md
```diff
@@ -0,0 +1,27 @@
...
+
+All keycodes including modifiers can be used the same way as with the key press behavior (so standard keycodes are sent for press/release).
```
dxmh, line 10:
```suggestion
All [keycodes](../codes/index.mdx) (including modifiers) can be used the same way as with the [key press](key-press.md) behavior.
```

docs/docs/behaviors/key-tap.md
```diff
@@ -0,0 +1,27 @@
...
+
+It is usefull if you want to make a layout that has the same keycode with and without a modifier close together, e.g. `[` and `{`.
+
+Since `LBRC` is basically just `LSHIFT(LBKT)` it can happen, that you still hold `{` while using `[` which leads to a missed keycode since the base-key is still pressed.
+
+By immediately releasing the key with key tap this is circumvented.
```
dxmh, line 16: This is a great tip that I think would be nice wrapped up inside a "tip" box, with a link to the GitHub thread tracking the issue. 

I personally find the simplified `+=` example from the GitHub thread is a little clearer to comprehend than the more detailed `{[` brackets example, but it's also then consistent when clicking through to GitHub.

```suggestion
:::tip

Key tap is useful if you experience issues when rolling keys that share a base keycode.

For example, a common programming bigram is `+=`. Behind the scenes, `+` is produced with `SHIFT` and `=`. So typing the keys `+` and `=` too quickly can sometimes result in one of them being ignored since the base-key is still pressed. However, by immediately releasing the key with key tap this issue is circumvented.

See [zmkfirmware/zmk #1076](https://github.com/zmkfirmware/zmk/issues/1076) for a thread tracking this.

:::
```

The above suggestion looks like this on my local environment:

image

docs/docs/behaviors/key-tap.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Behavior Binding
+
+- Reference: `&kt`
+- Parameter: The keycode usage ID from the usage page, e.g. `LBTK` or `A`
```
dxmh, line 21: Typo here? (`LBTK` is not a valid keycode.)

However, if we're changing the above tip we could keep this example consistent? e.g.:

```suggestion
- Parameter: The keycode usage ID from the usage page, e.g. `PLUS` or `A`
```

dxmh (commented):
Docs look good, thanks!

I do have a few minor suggestions below to simplify things and link off to related resources.

Please let me know what you think?

caksoylar:
I think this behavior having a `tap-ms` property similar to the `macro_tap` implementation would make sense, but it can be considered a future improvement as well. (Would it require using the behavior queue?)

marcoster:
@dxmh thanks for the great feedback and improvements. I totally forgot about the += example since my problem came from bracket/brace placement...

marcoster:
@caksoylar Good question - I'm not sure why this tap-time is needed for the macros and I'm not sure, what improvement this would bring to this simple key-tap? Is something like this sometimes needed for slowly reacting usb-hosts?

PR #1147, opened 2022-02-25
Behaviors: Fix memory leak in sticky keys

okke-formsma:
Some handlers returned 'ZMK_EV_EVENT_CAPTURED' instead of
'ZMK_EV_EVENT_HANDLED' as they should, and did not free
the memory.


okke-formsma:
Oops, there is no memory leak. I thought ZMK_EVENT_RAISE_AFTER would not release the memory but it does.

Issue #1149, opened 2022-02-27
Consecutive sticky layers not releasing correctly


dlip:
- press sticky layer key to go to layer 1
- press sticky layer key to go to layer 2
- press number 1 key in 'a' position
- press 'a' key

expected output:

1a

actual output:

11

okke-formsma:
Interesting. It's probably caused by this code:

<img width="878" alt="image" src="https://user-images.githubusercontent.com/486179/155883497-e42fe0d6-ed6e-4c8c-95da-746d39f3c2eb.png">

This code is in a loop, and it causes the second sticky key not to release. The return should be removed and all items in the loop should be processed.

dlip:
@okke-formsma I just tested your fix and it works beautifully! :clap: Thank you so much for helping so quickly! :pray:

dlip:
I may have found another small bug, see if you can reproduce:

- press sticky layer key to go to layer 1
- press sticky control key
- press bbbbb quickly

this results in multiple ctrl+b when it should just be 1

dlip:
hmm, maybe you can ignore my last bug, I've re-flashed and its not happening for me anymore

dlip:
I have found another related issue:

- press sticky layer key to go to layer 1
- hold sticky layer key to go to layer 2 (keep holding)
- press number 1 key in 'a' position
- release sticky layer 2 key
- press 'a' key

expected output:

1a

actual output:

11 (you can keep pressing a to output 1 until the sticky timeout completes)

PR #1154, opened 2022-02-28
Behaviors: Fix bug in nested sticky keys

okke-formsma:
If multiple sticky keys with quick release were nested, only the first one
was properly released. This fix makes sure all of them are released properly.

Fixes https://github.com/zmkfirmware/zmk/issues/1149


petejohanson (approved):
Thanks!

PR #1157, opened 2022-03-03
Update with split keyboard issues

methoxid:
Adding section ### Split keyboard example and default behaviour


caksoylar:
Sorry this fell through the cracks, I think the issue is sufficiently covered by the warning box in the summary section of https://zmk.dev/docs/behaviors/outputs now.

PR #1164, opened 2022-03-08
fix(docs): Update instructions to find settings reset firmware

cgoates:
Updated the instructions to find the settings reset firmware, since it isn't built on every "Build" GitHub Action any more.  Tested locally and ran prettier.


petejohanson (approved):
Better than nothing, given the lack of more specific search capabilities from GH.

PR #1166, opened 2022-03-12
feat(behaviors): Add macro support.

petejohanson:
* Fine grained press/release/tap actions.
* TIming between actions can be controlled.
* Processed async, to avoid blocking.

This is a stab at more nuanced macros, building on the initial macros work in #685 

# Outstanding

- [x] Add docs
- [x] Config option for max queue size
- [ ] Calculate queue size from largest macro, of not configured explicitly.
- [x] Test on hardware


app/src/behavior_queue.c
```diff
@@ -0,0 +1,64 @@
...
+#define CONFIG_ZMK_BEHAVIOS_QUEUE_SIZE 20
+
+struct q_item {
+    uint32_t position;
+    struct zmk_behavior_binding binding;
+    bool press;
+    uint8_t wait;
+};
+
+K_MSGQ_DEFINE(zmk_behavior_queue_msgq, sizeof(struct q_item), CONFIG_ZMK_BEHAVIOS_QUEUE_SIZE, 4);
```
okke-formsma, line 19: `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE`

okke-formsma (approved):
looks great!

app/include/dt-bindings/zmk/macros.h
```diff
@@ -0,0 +1,12 @@
+
+#define MAC_TAP_BIT (1 << 31)
+#define MAC_PRESS_BIT (1 << 30)
+#define MAC_REL_BIT (1 << 29)
+
+#define MAC_TAP_WAIT(ms) (MAC_TAP_BIT + ms)
+#define MAC_PRESS_WAIT(ms) (MAC_PRESS_BIT + ms)
+#define MAC_REL_WAIT(ms) (MAC_REL_BIT + ms)
+
+#define MAC_TAP MAC_TAP_WAIT(100)
```
joelspadin, line 10: If you want to be able to override the default delay, you could maybe use `MAC_REL_BIT - 1` as a special "default" value that gets swapped out for a DT property value when the macros code finds it.

petejohanson:
Proposal after some awesome brainstorming with @okke-formsma and @joelspadin :

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,149 @@
...
+            LOG_DBG("macro mode set: tap");
+        } else if (strcmp(cfg->items[i].binding.behavior_dev, PRESS_MODE) == 0) {
+            LOG_DBG("macro mode set: press");
+            mode = MACRO_MODE_PRESS;
+        } else if (strcmp(cfg->items[i].binding.behavior_dev, REL_MODE) == 0) {
+            LOG_DBG("macro mode set: release");
+            mode = MACRO_MODE_RELEASE;
+        } else if (press && strcmp(cfg->items[i].binding.behavior_dev, WAIT_REL) == 0) {
+            LOG_DBG("macro: stopping until macro is released");
+            state->wait_release_index = i;
```
okke-formsma, line 73: It might be even simpler to set "wait_release_seen" to `True`:

```
         } else if (strcmp(cfg->items[i].binding.behavior_dev, WAIT_REL) == 0) {
             wait_release_seen = true;
             if(press) { break; }
        }
```
Then change this:
```
         } else if (press || (state->wait_release_index > 0 && i > state->wait_release_index)) {
```
to 
```
         } else if (press || wait_release_seen)) {
```
That way no state needs to be kept at all.

okke-formsma (commented):
some small ideas to simplify.

app/src/behaviors/behavior_macro.c
petejohanson: I was thinking of a further refinement to actually store the wait/tap milliseconds to the state as well, so that on release, we actually would start our for loop at the previous stop point, avoiding extra unnecessary looping/state changes we've already done on press.

Thoughts?

joelspadin: Can the same macro be used on multiple keys? If so, what happens if you press the macro, then while still holding it press it again on a different key? Do the two instances have separate state, or does saving any state to the macro cause it to be overwritten in that case?

petejohanson: That would definitely be a limitation of the implementation, I'm not sure if there are any valid use cases for that, that we should worry about it for our first pass of this. Anybody got one?

app/dts/behaviors/macros.dtsi
```diff
@@ -0,0 +1,45 @@
...
+			#binding-cells = <1>;
+		};
+
+		macro_wait_time: behavior_macro_wait_time {
+			compatible = "zmk,behavior-macro-wait-time";
+			label = "MAC_WAIT_TIME";
+			#binding-cells = <1>;
+		};
+
+		macro_wait_rel: behavior_macro_wait_rel {
```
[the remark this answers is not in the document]
petejohanson: How about `macro_pause_until_release`?

app/dts/behaviors/macros.dtsi
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2021 The ZMK Contributors
```
joelspadin, line 2: Should all these copyright dates be 2022?

app/dts/bindings/behaviors/zmk,behavior-macro-press.yaml
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2020 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+description: Macro Press Mode Behavior
```
joelspadin, line 4: Maybe people will just look at the docs website instead of the code for this info, but something more like "sets macro to press mode" would probably be more helpful.

app/dts/bindings/behaviors/zmk,behavior-macro-press.yaml
```diff
@@ -0,0 +1,8 @@
...
+
+compatible: "zmk,behavior-macro-press"
```
joelspadin, line 6: This won't do anything if you try to put it directly in a keymap's bindings, so I wonder if calling it a "behavior" might be confusing. Should we come up with a new name for these? Maybe something like "behavior modifier" or "behavior config"?

app/dts/bindings/behaviors/zmk,behavior-macro.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+compatible: "zmk,behavior-macro"
+
+include: zero_param.yaml
+
+properties:
+  bindings:
+    type: phandle-array
+    required: true
+  wait-ms:
```
joelspadin, line 14: Should we put help text on these to clarify what the tap time and wait time do?

app/src/behavior_queue.c
```diff
@@ -0,0 +1,64 @@
...
+
+        if (item.press) {
+            behavior_keymap_binding_pressed(&item.binding, event);
+        } else {
+            behavior_keymap_binding_released(&item.binding, event);
+        }
+
+        k_delayed_work_submit(&queue_work, K_MSEC(item.wait));
+    }
+    return;
```
joelspadin, line 44: Minor: unnecessary return?

app/tests/macros/behavior_keymap.dtsi
```diff
@@ -0,0 +1,71 @@
...
+			#binding-cells = <0>;
+			bindings
+				= <&macro_press &kp LSHFT>
+				, <&macro_tap>
+				, <&kp D &kp O &kp G>
+				, <&macro_release &kp LSHFT>
+				;
+		};
+
+		custom_timing: cutom_timing_macro {
```
joelspadin, line 26:
```suggestion
		custom_timing: custom_timing_macro {
```

app/dts/bindings/behaviors/zmk,behavior-macro-press.yaml
petejohanson: How about `zmk,macro-control-press` ?

app/dts/bindings/behaviors/zmk,behavior-macro.yaml
petejohanson: Added `descriptions` to each of these properties. Thoughts on the specific descriptions?

app/dts/bindings/behaviors/zmk,behavior-macro-press.yaml
joelspadin: That seems pretty good to me.

app/dts/bindings/behaviors/zmk,behavior-macro.yaml
joelspadin: You might explicitly specify "time in milliseconds" in case the -ms suffix isn't obvious, but otherwise those seem understandable to me.

app/dts/bindings/behaviors/zmk,behavior-macro-press.yaml
petejohanson: I went with `zmk,macro-control-mode-press`.

app/dts/bindings/behaviors/zmk,behavior-macro.yaml
petejohanson: Added the units in the descriptions.

app/dts/behaviors/macros.dtsi
petejohanson: Fixed.

petejohanson:
@dxmh I added the initial docs for this: https://deploy-preview-1166--zmk.netlify.app/docs/behaviors/macros

Any feedback/improvements you can suggest would be appreciated.

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+static int on_macro_binding_pressed(struct zmk_behavior_binding *binding,
+                                    struct zmk_behavior_binding_event event) {
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_macro_config *cfg = dev->config;
+    struct behavior_macro_state *state = dev->data;
+
+    LOG_DBG("");
```
joelspadin, line 144: Are these empty logging statements intentional?

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,192 @@
...
+struct behavior_macro_state {
+    struct behavior_macro_trigger_state release_state;
+
+    int8_t last_press_binding;
+};
+
+struct behavior_macro_config {
+    uint32_t default_wait_ms;
+    uint32_t default_tap_ms;
+    uint8_t count;
```
joelspadin, line 43: Using a uint8 over a uint32 probably doesn't save any memory space unless you explicitly pack this struct. The compiler is most likely going to pad this out to 32-bits.

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,192 @@
...
+    MACRO_MODE_PRESS,
+    MACRO_MODE_RELEASE,
+};
+
+struct behavior_macro_trigger_state {
+    uint32_t wait_ms;
+    uint32_t tap_ms;
+    enum behavior_macro_mode mode;
+    int8_t start_index;
+    int8_t end_index;
```
joelspadin, line 31: This allows for 127 item macros, which is certainly a lot but still seems within the realm of possibility that someone would run past that limit. Some alternative ideas:

1. Use int16_t instead.
2. Use uint8_t and see my comment below about changing "end" to mean one past the last item.

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,192 @@
...
+};
+
+struct behavior_macro_state {
+    struct behavior_macro_trigger_state release_state;
+
+    int8_t last_press_binding;
```
joelspadin, line 37: Maybe rename this to something like `press_end_index` for consistency with `release_state.end_index`? Also see comment above about types.

Alternatively, be even more consistent and have `struct behavior_macro_trigger_state press_state` which you set up once on init instead of in every `on_macro_binding_pressed()` call.

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,192 @@
...
+
+#define TAP_MODE DT_LABEL(DT_INST(0, zmk_macro_control_mode_tap))
+#define PRESS_MODE DT_LABEL(DT_INST(0, zmk_macro_control_mode_press))
+#define REL_MODE DT_LABEL(DT_INST(0, zmk_macro_control_mode_release))
+
+#define TAP_TIME DT_LABEL(DT_INST(0, zmk_macro_control_tap_time))
+#define WAIT_TIME DT_LABEL(DT_INST(0, zmk_macro_control_wait_time))
+#define WAIT_REL DT_LABEL(DT_INST(0, zmk_macro_pause_for_release))
+
+#define _IS_NODE_MATCH(a, b) (strcmp(a, b) == 0)
```
joelspadin, line 55: Any symbol starting with an underscore followed by a capital letter is technically reserved for the compiler in C.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+        ab_macro: ab_macro {
+			label = "AB_MAC";
+			compatible = "zmk,behavior-macro";
+			#binding-cells = <0>;
+            wait-ms = <30>;
+			bindings
+				= <&kp A &kp B>
+				;
+		};
+    };
```
dxmh, line 61: There is a mixture of tabs vs spaces for indentation here (and in the other code blocks in this file).

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+			bindings
+				= <&macro_press &kp LALT>
+				, <&macro_tap>
+				, <&kp TAB>
+                , <&macro_pause_for_release>
+				, <&macro_release &kp LSHFT>
+				;
+		};
+    };
+};
```
dxmh, line 141: In addition to my above comments, I think there's a lot going on in this example that isn't necessarily relevant? Could this example focus solely on the `&macro_pause_for_release`, similar to the others examples in this document? Something like this:

~~~suggestion
```
bindings
    = <&kp D &kp O &kp W &kp N>
    , <&macro_pause_for_release>
    , <&kp U &kp P>
    ;
```
~~~

docs/sidebars.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
       "behaviors/misc",
       "behaviors/hold-tap",
       "behaviors/mod-tap",
+      "behaviors/macros",
```
dxmh, line 28: Don't forget to update the status of Macros in `/docs/docs/intro.md` 😃

image

dxmh (commented):
Exciting stuff! 

I’ve not yet tried out the functionality in this PR, as I thought it might be advantageous to do a first-pass review of the documentation from the perspective of a newcomer - so apologies in advance if some of my feedback/questions are due to any current misunderstanding I may have.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+The macro can then be bound in your keymap by referencing it by the label `&alt_tab`.
```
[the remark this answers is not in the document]
joelspadin: That macro is not quite the same as `&kp LA(TAB)` but also not a single-key "swapper" behavior. It is effectively `&kp LA(TAB)` when tapped, but it allows you to keep cycling forward by tapping Tab while holding the macro. On press it would cycle forward one window, then release Tab but not Alt, so you can keep holding the macro and tap Tab to cycle forward another window. 

I agree that the introductory example to press/release modes should be something simpler like holding shift while typing letters. We do need a good example for `&macro_pause_for_release` though. Can anyone come up with a better use case for that?

docs/sidebars.js
joelspadin, line 28: Sticking macros right between mod-tap and mod-morph seems a bit odd. Should we maybe just alphabetize everything after the really common ones (key press, layers, misc.)?

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Bindings
+
+Like hold-taps, macros are created by composing other behaviors, and any of those behaviors can
+be added to the `bindings` list.
+
+### Wait Time
```
joelspadin, line 44: I think it would make more sense to describe `wait-time-ms` and `&macro_wait_time` in the same section (and the same for tap time). That way it would be easier to understand that `wait-time-ms` sets the initial value, and `&macro_wait_time` changes it in the middle of the macro.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+## Macro Definition
+
+Each macro you want to use in your keymap gets defined first, then bound in your keymap.
+
+A macro definition looks like:
+
+```
+/ {
+    behaviors {
+        alt_tab: alt_tab_macro {
```
joelspadin, line 21: Since I know this stuff was basically magic to me before I started to understand DeviceTree, it may be helpful to clarify somewhere (in a comment inline in the code?) that the part before the colon is the label you will use in keymap bindings, and the part after the colon doesn't matter as long as it's unique.

(That suggestion also applies to the other docs pages like hold-tap and mod-morph.)

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+		};
+    };
+};
+```
+
+The macro can then be bound in your keymap by referencing it by the label `&alt_tab`.
+
+### Bindings
+
+Like hold-taps, macros are created by composing other behaviors, and any of those behaviors can
```
joelspadin, line 41: Linking "hold-taps" to the hold-tap docs may be helpful.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+be added to the `bindings` list.
```
joelspadin, line 42: An example with some different types of behavior bindings could be helpful, e.g.

```
bindings
    = <&to 1>
    , <&bl BL_ON>
    , <&kp Z &kp M &kp K &kp EXCLAMATION>
    ;
```

docs/docs/behaviors/macros.md
joelspadin: Suggestion for a real-world use case for this:

```
bindings 
    = <&macro_press &mo 1 &kp LSHFT>
    , <&macro_wait_for_release>
    , <&macro_release &mo 1 &kp LSHFT>
    ;

Comment (dxmh) [reply docs/docs/behaviors/macros.md:141]: I think that is a great suggestion for a real-world use case – and one that would possibly satisfy https://github.com/zmkfirmware/zmk/issues/968.

File: docs/docs/behaviors/macros.md [hunk 24]
```diff
@@ -0,0 +1,168 @@
...
+### Tap Time
+
+The default tap time is 100ms, unless explicitly set by the `tap-ms` property on the macro. If you want to update the tap time at any
+point in the middle of a macro sequence, use `&macro_tap_time`, e.g. `&macro_tap_time 30`. A full example:
+
+```
+bindings
+    = <&kp W &kp A &kp I &kp T>
+    , <&macro_tap_time 500>
+    , <&kp I &kp T>
```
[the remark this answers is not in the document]
dxmh: I like this pattern of up/down, fast/slow, short/long. 👍

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+bindings
+    = <&macro_press &kp LSHFT>
+    , <&macro_tap &kp Z &kp M &kp K>
+    , <&macro_release &kp LSHFT>
+    ;
+```
+
+### Processing Continuation on Release
+
+The macro sequence can be paused, so that only part of the sequence is processed when the macro is pressed, and the remainder is processed once
```
joelspadin, line 120:
```suggestion
The macro can be paused so that only part of the `bindings` list is processed when the macro is pressed, and the remainder is processed once
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+the macro itself is released. When processing is resumed, the current activation mode, wait time, and tap times will be preserved and used.
```
joelspadin, line 121:
```suggestion
the macro itself is released.
```
Do we need this last sentence? Since it's all just one bindings list, I feel like it should be fairly obvious that the state carries over. Explicitly stating what carries over might make you wonder what state *doesn't* carry over.

app/src/behavior_queue.c
```diff
@@ -0,0 +1,65 @@
...
+
+    LOG_DBG("");
+
+    ret = k_msgq_put(&zmk_behavior_queue_msgq, &item, K_NO_WAIT);
+    if (ret < 0) {
+        return ret;
+    }
+
+    if (!k_delayed_work_pending(&queue_work)) {
+        k_delayed_work_submit(&queue_work, K_MSEC(1));
```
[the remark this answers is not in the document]
petejohanson: Nope. Switching to `K_NO_WAIT`.

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    return 0;
+};
+
+static void queue_macro(uint32_t position, const struct zmk_behavior_binding (*bindings)[],
+                        struct behavior_macro_trigger_state trigger_state) {
+    LOG_DBG("Iterating macro bindings from %d-%d", trigger_state.start_index,
+            trigger_state.end_index);
+    for (int i = trigger_state.start_index; i <= trigger_state.end_index; i++) {
```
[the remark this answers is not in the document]
petejohanson: Fixed, using "start" + "count" variables now, which seems cleaner, and avoids the `<=` which I had originally done here.

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,192 @@
...
+            break;
+        } else {
+            // Ignore regular invokable bindings
+        }
+    }
+
+    return 0;
+};
+
+static void queue_macro(uint32_t position, const struct zmk_behavior_binding (*bindings)[],
```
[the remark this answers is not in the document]
petejohanson: I shouldn't code after midnight. Simplified.

app/src/behaviors/behavior_macro.c
petejohanson: I sometimes find them useful for certain scenarios to just see entry into functions happening. Removed.

app/src/behaviors/behavior_macro.c
petejohanson: Switched to `uint8_t`.

app/src/behaviors/behavior_macro.c
petejohanson: Kept this, but renamed. Storing a whole struct to state that's always initialized to the same starter values seems wasteful, allocated on the stack when calling `queue_macro` instead.

app/src/behaviors/behavior_macro.c
petejohanson: Thanks. Renamed.

docs/sidebars.js
petejohanson: I tried to roughly do this, but splitting up hold- and mod-tap seemed wrong.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+There are a set of special macro controls that can be included in the `bindings` list to modify the
+way the macro sequence is processed.
+
+### Binding Activation Mode
+
+Bindings in a macro sequence are activated differently, depending on the current "activation mode" of the macro.
+
+Available modes:
+
+- Tap - The default mode; when in this mode, the macro processing will press, then release, each behavior the macro sequence. This mode is useful for
```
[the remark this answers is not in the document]
petejohanson: Agreed, pushed tweaks to make this more consistent.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,168 @@
...
+				= <&kp A &kp B>
+				;
+		};
+    };
+};
+```
+
+### Tap Time
+
+The tap time setting controls how long of a delay is introduced between press and release of a tapped behavior in the macro bindings sequence.
```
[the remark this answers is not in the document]
petejohanson: Agreed. Updated.

docs/docs/behaviors/macros.md
petejohanson: Removed.

app/src/behavior_queue.c
```diff
@@ -0,0 +1,65 @@
...
+                                                   .timestamp = k_uptime_get()};
+
+        if (item.press) {
+            behavior_keymap_binding_pressed(&item.binding, event);
+        } else {
+            behavior_keymap_binding_released(&item.binding, event);
+        }
+
+        LOG_DBG("Processing next queued behavior in %dms", item.wait);
+        k_delayed_work_submit(&queue_work, K_MSEC(item.wait));
```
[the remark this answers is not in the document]
petejohanson: Indeed. I dig it. Updated.

docs/docs/behaviors/macros.md
petejohanson: Updated.

docs/docs/behaviors/macros.md
petejohanson: Linked.

docs/docs/behaviors/macros.md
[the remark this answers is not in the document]
petejohanson: Tweaked to use this approach in most of the examples.

docs/docs/behaviors/macros.md
petejohanson: Combined them into one.

docs/sidebars.js
petejohanson: Updated.

docs/docs/behaviors/macros.md
petejohanson: Yup, agreed that's a good example. Moved to this.

docs/docs/behaviors/macros.md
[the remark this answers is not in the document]
petejohanson: Made this a simpler intro example.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,136 @@
...
+    , <&macro_release &kp LSHFT>
+    ;
+```
+
+### Processing Continuation on Release
+
+The macro can be paused so that only part of the `bindings` list is processed when the macro is pressed, and the remainder is processed once
+the macro itself is released.
+
+To pause the macro until release, use `&macro_pause_for_release`, like in this example that will press a modifier and activate a layer when the macro is pressed, and once the macro is released, release the modifier and deactivate the layer by releasing the `&mo`:
```
jmding8, line 98: code snippet shows `macro_wait_for_release`. Should this be `pause` or `wait`?

petejohanson: Thanks, good catch.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,136 @@
...
+Each macro you want to use in your keymap gets defined first, then bound in your keymap.
+
+A macro definition looks like:
+
+```
+/ {
+    macros {
+        zed_em_kay: zmk_macro {
+        label = "ZMK_MAC";
+        compatible = "zmk,behavior-macro";
```
bwhelm, line 23: I've been looking at these instructions to make sure I understood how to define macros. For the most part they are clear, and I had no difficulty. A couple of questions, though:

1. Must this `compatible` line (and perhaps the following `#binding-cells` line) occur inside each particular macro, or can in be placed outside of the individual macros, but still under `macro { ... }` (as is true with combos)? That is, is the following permissible?---

        / {
            macros {
                compatible = "zmk,behavior-macro";
                #binding-cells = <0>;
                zed_em_kay: zmk_macro {
                    label = "ZMK_MAC";
                    bindings =
                        ...
                };
                kay_em_zed: kmz_macro {
                    label = "KMZ_MAC";
                    bindings =
                        ...
                };
            };
        };

2. I'm uncertain what the point of the `label` item is or whether it matters that it be assigned to anything in particular or just any unique string. (Does it have to be all caps, for example?) From the perspective of a user, that `label` item seems superfluous, and might just be made identical to the "node label".

petejohanson: Thanks for reviewing!

1. Not easily. At a minimum, the `binding-cells` still needs to be on the node. The compatible we could *maybe* get away w/ moving. Let me ponder more.
2.The label needs to be unique, it's used internally by our Zephyr RTOS... But perhaps we could auto-generate it...

petejohanson: Sadly, it really is required. We can definitely consider a C macro to simplify this.

petejohanson: Hmm... Thoughts on using this macro to simplify defining them:

```
		ZMK_MACRO(
			dual_sequence_macro,
			wait-ms = <10>;
			bindings
				= <&macro_press &kp LALT>
				, <&macro_tap>
				, <&kp TAB>
				, <&macro_pause_for_release>
				, <&macro_release &kp LALT>
				;
		)
```

petejohanson: @okke-formsma @joelspadin ^

filterpaper:
The `wait-ms` and `tap-ms` properties are not effective when configured in this manner:
```c
	macros {
		dir_up: dir_up_macro {
			compatible = "zmk,behavior-macro";
			label = "dir_up";
			#binding-cells = <0>;
			wait-ms = <5>;
			tap-ms = <5>;
			bindings = <&kp DOT &kp DOT &kp FSLH>;
		};
	};
```
Instead, they have be part of `bindings` to take effect:
```c
	macros {
		dir_up: dir_up_macro {
			compatible = "zmk,behavior-macro";
			label = "dir_up";
			#binding-cells = <0>;
			bindings = <&macro_wait_time 5>, <&macro_tap_time 5>, <&kp DOT &kp DOT &kp FSLH>;
		};
	};
```

petejohanson:
> The `wait-ms` and `tap-ms` properties are not effective when configured in this manner (intervals remained at default 100ms):
> 
> ```c
> macros {
>   dir_up: dir_up_macro {
>     compatible = "zmk,behavior-macro";
>     label = "dir_up";
>     #binding-cells = <0>;
>     wait-ms = <5>;
>     tap-ms = <5>;
>     bindings = <&kp DOT &kp DOT &kp FSLH>;
>   };
> };
> ```
> 
> Instead, they have be part of `bindings` to take effect:
> 
> ```c
> macros {
>   dir_up: dir_up_macro {
>     compatible = "zmk,behavior-macro";
>     label = "dir_up";
>     #binding-cells = <0>;
>     bindings = <&macro_wait_time 5>, <&macro_tap_time 5>, <&kp DOT &kp DOT &kp FSLH>;
>   };
> };
> ```

Found the bug, and fixed it, and tweaked our tests to be sure they captured this properly.

docs/docs/behaviors/macros.md
petejohanson: I haven't documented it yet, but I've added this macro now, if folks are happy with it, I'll add a note at the end of the docs w/ some examples of using the macro to simplify defining them.

app/Kconfig
```diff
@@ -410,6 +410,14 @@ config ZMK_COMBO_MAX_KEYS_PER_COMBO
 #Combo options
 endmenu
 
+menu "Behavior Options"
+
+config ZMK_BEHAVIORS_QUEUE_SIZE
+	int "Maximum number of behaviors to allow queueing from a macro or other complex behavior"
+	default 20
```
joelspadin, line 417: Should we default this to a number larger than we expect most people to ever use, like 64/128? Boards with less memory can turn it down as needed.

Otherwise, we're going to run into problems when we add things like a send string behavior and someone tries to send a string more than 20 characters long.

app/dts/behaviors/macros.dtsi
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#define ZMK_MACRO_STRINGIFY(x) #x
+#define ZMK_MACRO(name,...) \
+  name: name { \
```
joelspadin, line 9: Mixed spaces and tabs here

app/src/behavior_queue.c
```diff
@@ -0,0 +1,70 @@
...
+#include <logging/log.h>
+#include <drivers/behavior.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct q_item {
+    uint32_t position;
+    struct zmk_behavior_binding binding;
+    bool press;
+    uint32_t wait;
```
joelspadin, line 19: The compiler is probably going to pad `press` out to 32 bits (https://godbolt.org/z/PnhMazzhq). If you want to be really space-efficient, you could give `press` one bit and `wait` the remaining 31 to pack them both into 32-bits (https://godbolt.org/z/44EevY8a8).

app/src/behavior_queue.c
```diff
@@ -0,0 +1,70 @@
...
+};
+
+K_MSGQ_DEFINE(zmk_behavior_queue_msgq, sizeof(struct q_item), CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE, 4);
+
+static void behavior_queue_process_next(struct k_work *work);
+static K_DELAYED_WORK_DEFINE(queue_work, behavior_queue_process_next);
+
+static void behavior_queue_process_next(struct k_work *work) {
+    struct q_item item = {.wait = 0};
+    int ret;
```
joelspadin, line 29: Unused?

app/src/behavior_queue.c
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+}
+
+int zmk_behavior_queue_add(uint32_t position, const struct zmk_behavior_binding binding, bool press,
+                           uint32_t wait) {
+    struct q_item item = {.press = press, .binding = binding, .wait = wait};
+    int ret;
+
+    LOG_DBG("");
```
joelspadin, line 58: Empty log comment?

app/src/behavior_queue.c
```diff
@@ -0,0 +1,70 @@
...
+
+    ret = k_msgq_put(&zmk_behavior_queue_msgq, &item, K_NO_WAIT);
```
joelspadin, line 60: Could simplify to 
```suggestion
    const int ret = k_msgq_put(&zmk_behavior_queue_msgq, &item, K_NO_WAIT);
```
or
```suggestion
    int ret = k_msgq_put(&zmk_behavior_queue_msgq, &item, K_NO_WAIT);
```

app/src/behaviors/behavior_macro.c
```diff
@@ -0,0 +1,187 @@
...
+    state->release_state.start_index = cfg->count;
+    state->release_state.count = 0;
+
+    LOG_DBG("Precalculate initial release state:");
+    for (int i = 0; i < cfg->count; i++) {
+        if (handle_control_binding(&state->release_state, &cfg->bindings[i])) {
+            // Updated state used for initial state on release.
+        } else if (IS_PAUSE(cfg->bindings[i].behavior_dev)) {
+            state->release_state.start_index = i + 1;
+            state->release_state.count = cfg->count - i - 1;
```
joelspadin, line 101: maybe simplification? IMO this is more obvious than having a `- 1` in the expression.
```suggestion
            state->release_state.count = cfg->count - state->release_state.start_index;
```

docs/docs/behaviors/macros.md
joelspadin: As noted on Discord, I find it weird that it appears as if the first line is delimited with a comma but all others with semicolons. Maybe it would be less confusing if you formatted these like

```c
ZMK_MACRO(abc_macro,
    wait-ms = <10>;
    tap-ms = <20>;
    bindings = <&kp A &kp B &kp C>;
)
```
to differentiate the macro name from its contents?

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Binding Activation Mode
+
+Bindings in a macro are activated differently, depending on the current "activation mode" of the macro.
+
+Available modes:
+
+- Tap - The default mode; when in this mode, the macro will press, then release, each behavior in the `bindings` list. This mode is useful for
+  basic keycode output to hosts, i.e. when activating a `&kp` behavior.
+- Press - In this mode, the macro processing will only trigger a press on the behaviors in the macro. This is useful for holding down modifiers for some duration of a macro, e.g. `&kp LALT`.
```
joelspadin, line 74: suggestion for consistency with above
```suggestion
- Press - In this mode, the macro will only trigger a press on each behavior in the `bindings` list. This is useful for holding down modifiers for some duration of a macro, e.g. `&kp LALT`.
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,136 @@
...
+- Release - In this mode, the macro processing will only trigger a release on the behaviors in the macro. This is useful for releasing modifiers previously pressed earlier in the macro processing, e.g. `&kp LALT`.
```
joelspadin, line 75: suggestion for consistency with above
```suggestion
- Release - In this mode, the macro will only trigger a release on each behavior in the `bindings` list. This is useful for releasing modifiers previously pressed earlier in the macro, e.g. `&kp LALT`.
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,136 @@
...
+
+To modify the activation mode, macro controls can be added at any point in the bindings list.
```
joelspadin, line 77: Do we want to consistently monospace "bindings"?
```suggestion
To modify the activation mode, macro controls can be added at any point in the `bindings` list.
```

docs/docs/behaviors/macros.md
joelspadin, line 98: Feels a little bit like a run on sentence. Maybe split this up?
```suggestion
To pause the macro until release, use `&macro_pause_for_release`. For example, this macro will press a modifier and activate a layer when the macro is pressed. Once the macro is released, it will release the modifier and deactivate the layer by releasing the `&mo`:
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,136 @@
...
+### Tap Time
+
+The tap time setting controls how long a tapped behavior is held in the `bindings` list. The initial tap time for a macro, 100ms by default, can
+be set by assigning a value to the `tap-ms` property of the macro, e.g. `tap-ms = <20>;`. If you want to update the tap time at any
+point in a macro bindings list, use `&macro_tap_time`, e.g. `&macro_tap_time 30`. A full example:
+
+```
+bindings
+    = <&macro_tap_time 10>
+	, <&kp S &kp H &kp O &kp R &kp T>
```
joelspadin, line 132: Mixed tabs and spaces.

docs/sidebars.js
joelspadin: At the very least I think we shouldn't split up mod-tap and mod-morph.

filterpaper:
Thank you @petejohanson, `wait-ms` and `tap-ms` are taking effect correctly with [530301a](https://github.com/zmkfirmware/zmk/pull/1166/commits/530301a3a07c5344aa061b74c061452383a911de)

app/src/behavior_queue.c
petejohanson: Removed.

app/src/behavior_queue.c
petejohanson: Removed.

app/src/behavior_queue.c
petejohanson: Thanks! Fixed.

docs/sidebars.js
petejohanson: Moved after mod-morph.

app/src/behaviors/behavior_macro.c
petejohanson: Tweaked a bit.

petejohanson:
Added some common uses docs: https://deploy-preview-1166--zmk.netlify.app/docs/behaviors/macros/#common-patterns

And a small section at the end documenting the C macro.

I believe this is ready for merging, asking for one more review pass for anything I missed.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+### Unicode Sequences
+
+Many operating systems allow a special sequence to input unicode characters, e.g. [Windows alt codes](https://support.microsoft.com/en-us/office/insert-ascii-or-unicode-latin-based-symbols-and-characters-d13f58d3-7bcb-44a7-a4d5-972ee12e50e0). You can use macros to insert there automatically, e.g.:
+
+```
+wait-ms = <40>;
+tap-ms = <40>;
+bindings
+    = <&macro_press   &kp LALT>
+    , <&marco_tap     &kp N0 &kp N1 &kp N6 &kp N3>
```
joelspadin, line 193:
```suggestion
    , <&macro_tap     &kp N0 &kp N1 &kp N6 &kp N3>
```
Polo!

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+wait-ms = <0>;
+bindings
+    = <&macro_press &mo 1 &kp LSHFT>
+    , <&macro_pause_for_release>
+    , <&macro_release &mo 1 &kp LSHFT>;
+```
+
+#### Layer + Underglow Color
+
+To trigged a different underglow when the macro is pressed, and when it is released, we use the macro "press" activation mode whenever triggering the `&rgb_ug` behavior:
```
joelspadin, line 159:
```suggestion
To trigger a different underglow when the macro is pressed, and when it is released, we use the macro "press" activation mode whenever triggering the `&rgb_ug` behavior:
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+
+```
+wait-ms = <0>;
+bindings
+    = <&macro_press   &mo 1              &rgb_ug RGB_COLOR_HSB(128,100,100)>
+    , <&macro_pause_for_release>
+    , <&macro_release &mo 1 &macro_press &rgb_ug RGB_COLOR_HSB(300,100,50)>;
```
joelspadin, line 166: Is it okay to press &rgb_ug twice but never release it?

dxmh: This is a good point that I think we should include in the documentation.

What is the expected outcome when a key is pressed within a macro but not released via `&macro_release`? Should the macro release any pressed keys upon completion regardless of whether `&macro_release` is used or not?

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+
+```
+    ZMK_MACRO(my_macro,
+        wait-ms = <20>;
+        tap-ms = <40>;
+        bindings = <&kp Z &kp M &kp K>;
+    )
+```
+
+The macro above would produce the following complete macro definition:
```
dxmh, line 210: I think there could be some benefit in clarifying that the complete macro definition is not required, and that the complete macro definition will be produced only during compilation (meaning the user should not expect to see complete version). Something like this perhaps?

```suggestion
This can be used instead of a complete macro definition. During the firmware build process, the example above would produce the complete macro definition below:
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+#### Layer + Underglow Color
+
+To trigged a different underglow when the macro is pressed, and when it is released, we use the macro "press" activation mode whenever triggering the `&rgb_ug` behavior:
+
+```
+wait-ms = <0>;
+bindings
+    = <&macro_press   &mo 1              &rgb_ug RGB_COLOR_HSB(128,100,100)>
```
dxmh, line 164:
```suggestion
    = <&macro_press &mo 1 &rgb_ug RGB_COLOR_HSB(128,100,100)>
```

I appreciate that things are aligned with the lines below, but when viewing the documentation on a mobile device, it looks like this line contains only `<&macro_press &mo 1` which might be confusing (this kind of alignment is also not consistent with the other examples).

---

image

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+
+Each macro you want to use in your keymap gets defined first, then bound in your keymap.
+
+A macro definition looks like:
+
+```
+/ {
+    macros {
+        zed_em_kay: zmk_macro {
+            label = "ZMK_MAC";
```
dxmh, line 22: Now that we have the C macro, we could make this example more consistent. 

Also, I personally like things repeating throughout the binding, name, and label. As a user, it's helpful for a copy/paste find/replace workflow.

```suggestion
        zed_em_kay: zed_em_kay {
            label = "ZMK_zed_em_kay";
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
+---
+title: Macro Behavior
+sidebar_label: Macros
+---
+
+## Summary
+
+The macro behavior allows configuring a list of other behaviors to invoke
+when the macro is pressed, including support for completing part of the binding list
+once the macro is released.
```
dxmh, line 10: Now that we've thorough documentation below, could this introduction be simplified? 

I think we already simplified the first half of this in previous revisions, but we still refer to "the binding list" in the second half and I'm not sure it's totally necessary. Something like the suggestion below is less for a new user to wrap their head around, whilst still hinting about the awesome release functionality (which is covered thoroughly later in the doc):

```suggestion
The macro behavior allows configuring a list of other behaviors to invoke
when the macro is pressed and/or released.
```

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+bindings
+    = <&macro_tap_time 10>
+    , <&kp S &kp H &kp O &kp R &kp T>
+    , <&macro_tap_time 500>
+    , <&kp L &kp O &kp N &kp G>
+    ;
+```
+
+## Common Patterns
+
```
dxmh, line 139: I think a brief introduction to this section is needed:

```suggestion
## Common Patterns

Below are some examples of how the Macros behavior can be used to achieve various different functionality. 
```

dxmh (commented):
I left a few minor comments below.

I think the documentation is great overall though – it's really thorough. The addition of the included C macro is a nice touch (something most people are currently defining themselves for setting up combos). I also like the Common Patterns section - it really demonstrates how powerful the macros functionality is. 👍

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+The macro above would produce the following complete macro definition:
+
+```
+    my_macro: my_macro {
+        compatible = "zmk,behavior-macro";
+        label = "ZMK_my_macro";
+        #binding-cells = <0>;
+        waif-ms = <20>;
```
bwhelm, line 217: Should be `wait-ms = <20>;`.

bwhelm: Also: given the recommendation that `wait-ms` and `tap-ms` be at least 30, shouldn't this example (both here and in the above C macro, of course) follow those recommendations?

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,223 @@
...
+bindings
+    = <&macro_press   &kp LALT>
+    , <&marco_tap     &kp N0 &kp N1 &kp N6 &kp N3>
+    , <&macro_release &kp LALT>
+    ;
+```
+
+## Convenience C Macro
+
+To avoid repitition or possible typos when declaring a macro, a conenience _C_ macro, named `ZMK_MACRO(name, props)` can be used to simplify things:
```
joelspadin, line 200:
```suggestion
To avoid repetition or possible typos when declaring a macro, a conenience _C_ macro, named `ZMK_MACRO(name, props)` can be used to simplify things:
```

docs/docs/behaviors/macros.md
joelspadin: It could potentially be useful to define a macro that never releases a key and have a different macro that does release it, though having keys be pressed and never released in a macro is probably more likely to be accidental.

MatCyg:
Thank you a lot for all the hard work! This finally allowed me to move my complex configuration from karabiner elements to zmk. I am currently leveraging this branch for my keymap containing 300+ macros, featuring things like 'vim like' [selection/visual mode](https://github.com/MatCyg/zmk-config/blob/standalone/config/macros/selection_mode.dtsi) and [custom auto shift](https://github.com/MatCyg/zmk-config/blob/standalone/config/macros/auto_shift.dtsi). 

Auto shift is different from the usual implementation. It always sends lowercase key on key down (without repeat). Once held long enough it sends backspace + uppercase letter. This way there is no noticeable delay during typing. To achieve this I had to combine two macros and a hold tap:
```c
ZMK_MACRO(letter_A,
    wait-ms = <0>;
    tap-ms = <10>;
    bindings
        = <&kp A>
        , <&macro_press &hold_letter_A 0 0>
        , <&macro_pause_for_release>
        , <&macro_release &hold_letter_A 0 0>
        ;
)

hold_letter_A: hold_letter_A {
    label = "hold_letter_A";
    #binding-cells = <2>;
    tapping-term-ms = <180>;
    flavor = "tap-preferred";
    bindings = <&capital_letter_A>, <&none>;
};

ZMK_MACRO(capital_letter_A,
    wait-ms = <0>;
    tap-ms = <10>;
    bindings = <&kp BSPC &kp LS(A)>;
)

// Use &letter_A in a keymap
```

The trick is to combine `macro press` with hold tap like this:
`<&macro_press &hold_tap_A 0 0>`

where `&hold_tap_A` is defined with the following binding:
`bindings = <&capital_letter_A>, <&none>;`

where `&capital_letter_A` is defined as macro with binding:
`bindings = <&kp BSPC &kp LS(A)>;`

With this we achieve the following behavior:
- tap -> send 'a'
- hold -> send 'a', wait for `tapping-term-ms`, send 'BSPC', send 'A'

The point why I am explaining is to show that macro properly propagates press (hold) to `hold-tap` behavior. However, I noticed that the same is not true when we put the above macro into another `hold-tap` with `quick_tap_ms`. Like this: 

```c
hm_letter_A: hm_letter_A { \
    compatible = "zmk,behavior-hold-tap"; \
    label = "hm_letter_A"; \
    #binding-cells = <2>; \
    tapping-term-ms = <120>; \
    quick_tap_ms = <180>; \
    flavor = "tap-preferred"; \
    bindings = <&kp>, <&letter_A>; \
};
// Use &hm_letter_A LSFT 0 in a keymap
```

I believe that double tap on `hm_letter_A` should propagate hold to `<&macro_press &hold_letter_A 0 0>`. It doesn't work like that right now. This might be an issue with how `quick_tap_ms` is implemented.

I can imagine my example might be a bit complex/out of scope for this PR. I am happy to provide more details and better explanation if needed.

docs/docs/behaviors/macros.md
petejohanson: I updated these, and noted this in the docs covering this. Marking resolved.

docs/docs/behaviors/macros.md
petejohanson: Indeed. Fixed the typo, and adjusted the value. Thanks!

docs/docs/behaviors/macros.md
petejohanson: Added a small intro, worded slightly differently.

docs/docs/behaviors/macros.md
petejohanson: Yeah, totally agree. Added the simplified version.

docs/docs/behaviors/macros.md
petejohanson: Yup. Good call. Fixed.

docs/docs/behaviors/macros.md
petejohanson: Tweaked this to make it more legible on mobile. Great catch!

petejohanson:
> ```c
> 
>         , <&macro_press &hold_letter_A 0 0>
>         , <&macro_pause_for_release>
>         , <&macro_release &hold_letter_A 0 0>
>         ;
> ```

@MatCyg It's an interesting use case.... let me try to put that together an an integration test, see what bubbles out of this.

docs/docs/behaviors/macros.md
```diff
@@ -0,0 +1,227 @@
...
+
+Each macro you want to use in your keymap gets defined first, then bound in your keymap.
+
+A macro definition looks like:
+
+```
+/ {
+    macros {
+        zed_em_kay: zed_em_kay {
+            label = "ZM_zed_em_kay";
```
dxmh, line 21: Typo?

```suggestion
            label = "ZMK_zed_em_kay";
```

petejohanson: No, was going for the initials for Zmk Macro -> "Z.M." -> `ZM_`

dxmh: Oh nice, I missed that 👍

dxmh: > Which is in fact what the macro generates.

Does [this example](https://github.com/zmkfirmware/zmk/pull/1166/files#diff-bb8c34fc4eb07da51cc48c0999057d683fed48c7c51997521ada620789efe88aR219) need changing at the end of the doc then? (`ZMK_my_macro`→`ZM_my_macro`)

```
    my_macro: my_macro {
        compatible = "zmk,behavior-macro";
        label = "ZMK_my_macro";
        #binding-cells = <0>;
        wait-ms = <30>;
        tap-ms = <40>;
        bindings = <&kp Z &kp M &kp K>;
    };
```

MatCyg:
I found a case that could be considered a bug.

```c 
keymap {
    compatible = "zmk,keymap";
    
    test { bindings = <
        &unshiftedA &kp LSFT &kp B
    >;};
};
ZMK_MACRO(
    unshiftedA,
    wait-ms = <0>;
    tap-ms = <0>;
    bindings
        = <&macro_release &kp LSFT>
        , <&macro_tap &kp A>
        , <&macro_press &kp LSFT>
        ;
)
```

Pressing `unshiftedA` sets shift as pressed. After that, `&kp B` will produce uppercase `B` letter. I would expect that taping `&kp LSFT` would release the `<&macro_press &kp LSFT>` which it doesn't. Not sure if this is by design.

In other words a tapping sequence:
&unshiftedA -> &kp LSFT -> &kp B 
produces uppercase B, while I would expect shift to no longer be pressed after `&kp LSFT` and a lowercase b produced.

petejohanson:
> I found a case that could be considered a bug.
> 
> ```c 
> keymap {
>     compatible = "zmk,keymap";
>     
>     test { bindings = <
>         &unshiftedA &kp LSFT &kp B
>     >;};
> };
> ZMK_MACRO(
>     unshiftedA,
>     wait-ms = <0>;
>     tap-ms = <0>;
>     bindings
>         = <&macro_release &kp LSFT>
>         , <&macro_tap &kp A>
>         , <&macro_press &kp LSFT>
>         ;
> )
> ```
> 
> Pressing `unshiftedA` sets shift as pressed. After that, `&kp B` will produce uppercase `B` letter. I would expect that taping `&kp LSFT` would release the `<&macro_press &kp LSFT>` which it doesn't. Not sure if this is by design.
> 
> In other words a tapping sequence:
> &unshiftedA -> &kp LSFT -> &kp B 
> produces uppercase B, while I would expect shift to no longer be pressed after `&kp LSFT` and a lowercase b produced. 

Why are you doing that order in the macro? You shouldn't leave that hanging like that. IIRC, we record the *count* of presses/releases to not unset a mod if a user presses both shift keys on a standard layout, then releases only one of them.

MatCyg:
> Why are you doing that order in the macro? You shouldn't leave that hanging like that. IIRC, we record the _count_ of presses/releases to not unset a mod if a user presses both shift keys on a standard layout, then releases only one of them.

To be honest I found it by a mistake, and got my keyboard stuck in permanently enabled shift. I tried pressing shift to release it, but it didn't work so decided to mention it. If you think it works as expected please ignore this.

petejohanson:
The issue with your shifted A example is failing because hold-tap currently doesn't differentiate multiple hold-taps and their triggering from the "same position" even though they're unique/different because of macro invocation.

I will consider how we might fix this, but I'm not sure it's in scope for the initial pass at macros.

docs/docs/behaviors/macros.md
MatCyg: FYI: One of the good use cases for not releasing a key in macro is to keep is opening app switcher on Mac:
```c
ZMK_MACRO(
        app_switcher,
        wait-ms = <0>;
        tap-ms = <0>;
        bindings
            = <&macro_press &kp LGUI>
            , <&macro_tap &kp TAB>
            ;
    )
```
This opens  app switcher and keeps it open. With right programming of other keys, it would release GUI key i.e., on enter key.

docs/docs/behaviors/macros.md
petejohanson: Indeed. Fixed.

PR #1167, opened 2022-03-12
fix(boards): Adjust matrix for BT60v1 to fix the broken right Shift key

LiquidPL:
This PR fixes a badly configured matrix for the BT60v1 board (ANSI layout), where the right Shift was configured to use a wrong row/column pair. According to the [diagram](https://github.com/ReFil/zmk-config#more-advanced) from the board designers, it should be using `RC(3,12)`, instead of `RC(3,13)` it's using currently.


Nicell:
@LiquidPL, thanks for fixing this. Can you confirm that this change works properly on real hardware? Just want to confirm the diagram matches the real thing.

LiquidPL:
Yeah, I have tested a firmware built off this on my own board, and I can confirm that the key now works correctly.

PR #1169, opened 2022-03-15
feat: append the current profile number to the advertised name

Percentnineteen:
This commit adds the string "~#" to the end of the keyboard name for advertising purposes. If the name is too long to fit this string, the end of the keyboard name is overwritten. KeyboardName would become KeyboardName~2 while LongKeyboardName would become LongKeyboardNa~2 if profile 2 is active.

I did make the assumption that no (sane) person would have more than 99 profiles (i.e. the length of the active profile would never be more than 2 characters).

I've tested this on Windows, macOS and Android and all reflected the name change (albeit with various amounts of delay and macOS seems to like to store the old name).

There may be better ways to do this - I haven't written in C (or really any non-scripting language) in at least a dozen years.

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -78,11 +78,13 @@ static uint8_t active_profile;
 #define DEVICE_NAME CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
 
+static uint8_t device_name[17] = DEVICE_NAME;
+
 BUILD_ASSERT(DEVICE_NAME_LEN <= 16, "ERROR: BLE device name is too long. Max length: 16");
 
-static const struct bt_data zmk_ble_ad[] = {
+static struct bt_data zmk_ble_ad[] = {
 #if IS_HOST_PERIPHERAL
-    BT_DATA(BT_DATA_NAME_COMPLETE, DEVICE_NAME, DEVICE_NAME_LEN),
+    BT_DATA(BT_DATA_NAME_COMPLETE, device_name, 16),
     BT_DATA_BYTES(BT_DATA_GAP_APPEARANCE, 0xC1, 0x03),
 #endif
     BT_DATA_BYTES(BT_DATA_FLAGS, (BT_LE_AD_GENERAL | BT_LE_AD_NO_BREDR)),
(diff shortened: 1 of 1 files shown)


app/src/ble.c
```diff
@@ -183,6 +185,12 @@ int update_advertising() {
     struct bt_conn *conn;
     enum advertising_type desired_adv = ZMK_ADV_NONE;
 
+    uint8_t max_device_name_length = (active_profile > 9) ? 13 : 14;
+    // need to store the temporary device name because Zephyr *printf doesn't support string precision
+    char temp_device_name[14];
+    snprintf(temp_device_name, max_device_name_length, "%s", DEVICE_NAME);
+    snprintf(device_name, 16, "%s~%d", temp_device_name, active_profile);
```
petejohanson, line 192: We can probably avoid the double copy here with something like:

```
uint8_t prof_num_len = active_profile > 9 : 3 : 2;
snprintf(device_name, MAX_NAME - prof_num_le, "%s", DEVICE_NAME);
sprintf(device_name + MAX_NAME - prof_num_len, prof_number_len, "-%d", active_profile);
```

Just specifying a different start address for the snprintf for the second call, to just print it into the "end" of the string.

petejohanson (changes requested):
Thanks for working on this!

A few comments.

app/src/ble.c
```diff
@@ -78,11 +78,13 @@ static uint8_t active_profile;
 #define DEVICE_NAME CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
 
+static uint8_t device_name[17] = DEVICE_NAME;
```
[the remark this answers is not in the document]
Percentnineteen: Done.

app/src/ble.c
```diff
@@ -77,12 +77,15 @@ static uint8_t active_profile;
...
+
+static uint8_t device_name[DEVICE_NAME_ARRAY_MAX_SIZE] = DEVICE_NAME;
 
 BUILD_ASSERT(DEVICE_NAME_LEN <= 16, "ERROR: BLE device name is too long. Max length: 16");
 
-static const struct bt_data zmk_ble_ad[] = {
+static struct bt_data zmk_ble_ad[] = {
 #if IS_HOST_PERIPHERAL
-    BT_DATA(BT_DATA_NAME_COMPLETE, DEVICE_NAME, DEVICE_NAME_LEN),
+    BT_DATA(BT_DATA_NAME_COMPLETE, device_name, 16),
```
petejohanson, line 88:
```suggestion
    BT_DATA(BT_DATA_NAME_COMPLETE, device_name, DEVICE_NAME_ARRAY_MAX_SIZE-1),
```

app/src/ble.c
```diff
@@ -183,6 +186,13 @@ int update_advertising() {
     struct bt_conn *conn;
     enum advertising_type desired_adv = ZMK_ADV_NONE;
 
+    uint8_t max_device_name_length = (active_profile > 9) ? 13 : 14;
```
petejohanson, line 189:
```suggestion
    uint8_t max_device_name_length = DEVICE_NAME_ARRAY_MAX_SIZE - ((active_profile > 9) ? 4 : 3);
```

petejohanson (changes requested):

PR #1175, opened 2022-03-20
Test macros that include modifiers and layers

jmding8:
Adds a test (`basic`) with a macro that activates a momentary layer and a modifier, and does not release them until `macro_pause_for_release`. Test expects that the outcome should be a shifted, layered key. Unfortunately, this test fails and emits a NON-shifted, layered key.

Adds a second test (`macro_from_tap_hold`) that is very similar to the first, except that the macro is triggered from within a tap/hold instead of by a basic keypress. This test expects the same kind of outcome as the first. Unfortunately, it also fails by emitting a NON-shifted, NON-layered key.

All changes are built on top of petejohanson_behaviors/macros-take-two, and are contained within the `/workspaces/zmk/app/tests/macros/mod-layer-from-macro` directory


petejohanson:
I merged in some tests based on this, closing.

PR #1179, opened 2022-03-21
RGB Underglow: Auto off on idle

infused-kim:
This PR adds an auto off on idle feature to the underglow. It's already supported for the backlight and this adds a similar setting to the underglow.

I have tested it on my sofle choc and it's working great.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -341,6 +341,9 @@ config ZMK_RGB_UNDERGLOW_ON_START
 	bool "RGB underglow starts on by default"
 	default y
 
+config ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE
+	bool "Turn off RGB underglow when keyboard goes into idle state"
+
 #ZMK_RGB_UNDERGLOW
 endif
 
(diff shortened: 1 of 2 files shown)


infused-kim:
I realized my initial commit had a bug.

`zmk_rgb_underglow_off`and `zmk_rgb_underglow_on` save the state to the config. So if the user has underglow on, then idles  and the underglow is auto-turned off, that setting saved into the flash. If the user turns off and on their keyboard, it starts in the underglow off state even though the user never changed that setting.

The new commit ensures the auto-off feature is only active if the underglow is on and that it doesn't save the temporary off state into the config.

ReFil:
I believe this is a duplicate of my PR #1010

ReFil:
#1010 has been merged so this is superfluous now

petejohanson:
Thanks for the PR, @infused-kim!

Can you please try with the latest main? You'll need to enable the new Kconfig option to enable the functionality.

autoferrit:
This seems like a smaller change. is anything other than review needed to get this in? is there any way I can help? I am new to c/zmk but could try.

Issue #1178, opened 2022-03-21
[Suggestion] Reuse Github CI definition for zmk_config repositories


elagil:
Currently, each zmk_config repository that a user creates, includes its own Github actions pipeline definition.

As described in [this document](https://docs.github.com/en/actions/using-workflows/reusing-workflows), the job definition could be imported from a global place, instead of copying it each time. Thus, it would remain up to date and would not have to be adjusted manually (e.g. for things like [these](https://zmk.dev/blog/2022/03/08/zephyr-3-0-upgrade-prep/)). It looks like these job definitions are always the same, and if not, the imported job can be extended.

If this suggestion makes sense to you, I can create a pull request for this functionality.

joelspadin:
Seems like a good idea to me now that we pull configuration from build.yaml in the root of the repo instead of from the workflow filed directly.

elagil:
I made two PRs. 

1. Add a `build-user-config.yaml` that closely resembles the current template for user configurations, but with an added header
2. Update the template, so that it uses the new reusable workflow

PR #1183, opened 2022-03-22
ci: add reusable user-config workflow

elagil:
Closes https://github.com/zmkfirmware/zmk/issues/1178

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -64,14 +64,15 @@ jobs:
           echo ::set-output name=extra-cmake-args::${EXTRA_CMAKE_ARGS}
           echo ::set-output name=artifact-name::${ARTIFACT_NAME}
           echo ::set-output name=display-name::${DISPLAY_NAME}
+          echo ::set-output name=zephyr-version::${ZEPHYR_VERSION}
 
       - name: Checkout
         uses: actions/checkout@v2
 
       - name: Cache west modules
-        uses: actions/cache@v2
+        uses: actions/cache@v3.0.1
         env:
-          cache-name: cache-zephyr-modules
+          cache-name: cache-zephyr-${{ steps.variables.outputs.zephyr-version }}-modules
         with:
           path: |
             modules/
@@ -97,12 +98,7 @@ jobs:
       - name: West Build (${{ steps.variables.outputs.display-name }})
         run: |
           west build -s zmk/app -b ${{ matrix.board }} -- -DZMK_CONFIG=${GITHUB_WORKSPACE}/${{ inputs.config_path }} ${{ steps.variables.outputs.extra-cmake-args }} ${{ matrix.cmake-args }}
-      - name: ${{ steps.variables.outputs.display-name }} DTS File
-        if: ${{ always() }}
-        run: |
-          if [ -f "build/zephyr/${{ matrix.board }}.dts.pre.tmp" ]; then cat -n build/zephyr/${{ matrix.board }}.dts.pre.tmp; fi
-          if [ -f "build/zephyr/zephyr.dts" ]; then cat -n build/zephyr/zephyr.dts; fi
-      
+
       - name: ${{ steps.variables.outputs.display-name }} Kconfig file
         run: cat build/zephyr/.config | grep -v "^#" | grep -v "^$"
       
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -79,7 +79,7 @@ jobs:
             zephyr/
             bootloader/
             zmk/
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('config/west.yml') }}
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('${{ inputs.config_path }}/west.yml') }}
           restore-keys: |
             ${{ runner.os }}-build-${{ env.cache-name }}-
             ${{ runner.os }}-build-
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -1,0 +1,96 @@
+name: Reusable user config build
+
+on: 
+  workflow_call:
+
+jobs:
+  matrix:
+    runs-on: ubuntu-latest
+    name: Fetch Build Keyboards
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install yaml2json
+        run: python3 -m pip install remarshal
+      - id: set-matrix
+        name: Fetch Build Matrix
+        run: |
+          matrix=$(yaml2json build.yaml | jq -c .)
+          yaml2json build.yaml
+          echo "::set-output name=matrix::${matrix}"
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: zmkfirmware/zmk-build-arm:stable
+    needs: matrix
+    name: Build
+    strategy:
+      fail-fast: false
+      matrix: ${{fromJson(needs.matrix.outputs.matrix)}}
+    steps:
+      - name: Prepare variables
+        id: variables
+        run: |
+          if [ -n "${{ matrix.shield }}" ]; then
+            EXTRA_CMAKE_ARGS="-DSHIELD=${{ matrix.shield }}"
+            ARTIFACT_NAME="${{ matrix.shield }}-${{ matrix.board }}-zmk"
+            DISPLAY_NAME="${{ matrix.shield }} - ${{ matrix.board }}"
+          else
+            EXTRA_CMAKE_ARGS=
+            DISPLAY_NAME="${{ matrix.board }}"
+            ARTIFACT_NAME="${{ matrix.board }}-zmk"
+          fi
+          echo ::set-output name=extra-cmake-args::${EXTRA_CMAKE_ARGS}
+          echo ::set-output name=artifact-name::${ARTIFACT_NAME}
+          echo ::set-output name=display-name::${DISPLAY_NAME}
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Cache west modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-zephyr-modules
+        with:
+          path: |
+            modules/
+            tools/
+            zephyr/
+            bootloader/
+            zmk/
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('manifest-dir/west.yml') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: West Init
+        run: west init -l config
+      - name: West Update
+        run: west update
+      - name: West Zephyr export
+        run: west zephyr-export
+      - name: West Build (${{ steps.variables.outputs.display-name }})
+        run: |
+          west build -s zmk/app -b ${{ matrix.board }} -- -DZMK_CONFIG=${GITHUB_WORKSPACE}/config ${{ steps.variables.outputs.extra-cmake-args }} ${{ matrix.cmake-args }}
+      - name: ${{ steps.variables.outputs.display-name }} DTS File
+        if: ${{ always() }}
+        run: |
+          if [ -f "build/zephyr/${{ matrix.board }}.dts.pre.tmp" ]; then cat -n build/zephyr/${{ matrix.board }}.dts.pre.tmp; fi
+          if [ -f "build/zephyr/zephyr.dts" ]; then cat -n build/zephyr/zephyr.dts; fi
+      - name: ${{ steps.variables.outputs.display-name }} Kconfig file
+        run: cat build/zephyr/.config | grep -v "^#" | grep -v "^$"
+      - name: Rename artifacts
+        run: |
+          mkdir build/artifacts
+          if [ -f build/zephyr/zmk.uf2 ]
+          then
+            cp build/zephyr/zmk.uf2 "build/artifacts/${{ steps.variables.outputs.artifact-name }}.uf2"
+          elif [ -f build/zephyr/zmk.hex ]
+          then
+            cp build/zephyr/zmk.hex "build/artifacts/${{ steps.variables.outputs.artifact-name }}.hex"
+          fi
+      - name: Archive (${{ steps.variables.outputs.display-name }})
+        uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,96 @@
...
+        env:
+          cache-name: cache-zephyr-modules
+        with:
+          path: |
+            modules/
+            tools/
+            zephyr/
+            bootloader/
+            zmk/
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('manifest-dir/west.yml') }}
```
joelspadin, line 61:
```suggestion
          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('config/west.yml') }}
```
this was an error in the original repo.

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,96 @@
+name: Reusable user config build
+
+on: 
+  workflow_call:
```
joelspadin, line 4: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputs indicates that we can add inputs. A couple of inputs that I think would be useful:

```yaml
on:
  workflow_call:
    inputs:
      build_matrix_path:
        description: Path to the build matrix file
        default: build.yaml
        required: false
        type: string
      config_path:
        description: Path to the config directory
        default: config
        required: false
        type: string
```
This would give people some flexibility as to the organization of their repo. 

For example, in https://github.com/joelspadin/NumBLE/tree/zephyr-3.0 I put both the hardware design files for a keyboard and the ZMK config for it in the same repo. To make it easier to understand, I renamed "config" to "zmk-config" for that repo. It would be great if I could just call this job with

```yaml
with:
  config_path: zmk-config
```

docs/blog/2022-03-08-zephyr-3-0-upgrade-prep.md
```diff
@@ -13,17 +13,17 @@ their repositories now.
 GitHub Actions needs to use our latest Docker image to ensure continued compatibility with the ZMK codebase on Zephyr 3.0 (and beyond). You should:
 
 - Open `.github/workflows/build.yml` in your editor/IDE
-- Change `zmkfirmware/zmk-build-arm:2.5` to `zmkfirmware/zmk-build-arm:stable` wherever it is found
+- Replace its content with the [updated template](https://github.com/zmkfirmware/unified-zmk-config-template/blob/main/.github/workflows/build.yml)
 
-Once the changes are committed and pushed, the build will run as expected.
+The new configuraiton template does not include the pipeline definition itself, but rather references a central configuration file. Thus, you will not have to update this file again in the future, as all updates to the central configuration will automatically be reflected in your new pipeline runs.
```
joelspadin, line 18:
```suggestion
The new configuration template does not include the pipeline definition itself, but rather references a central configuration file. Thus, you will not have to update this file again in the future, as all updates to the central configuration will automatically be reflected in your new pipeline runs.
```

.github/workflows/build-user-config.yml
elagil: Yes, that makes sense. I will include these adjustments in a commit this afternoon.

If we are already adding functionality in this PR, I would also add an input for the chosen artifact file type. For example, I want the `*.bin` file for flashing with `dfu-util`, instead of  `*.hex`. Why is it `*.hex` in the first place?

elagil:
I added some input fields and tested my job again: https://github.com/elagil/ZMK-keyboards/actions/runs/2024498175
Looks like it works as expected.

You may also test with your desired alternative inputs, using my branch [like here](https://github.com/elagil/ZMK-keyboards/blob/test_reusable_user_config/.github/workflows/build.yml).

.github/workflows/build-user-config.yml
elagil: Added changes.

elagil:
Some force-pushing for adhering to conventional commit messages.

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,112 @@
...
+        env:
+          cache-name: cache-zephyr-modules
+        with:
+          path: |
+            modules/
+            tools/
+            zephyr/
+            bootloader/
+            zmk/
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('config/west.yml') }}
```
filterpaper, line 77:
```suggestion
          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/west.yml') }}
```

Does it make sense to use wildcard for the hashed file? `config` folder location is an option for this workflow.

joelspadin: Why not use `${{ inputs.config_path }}` and use the given input?

filterpaper: That's a better idea and should cover all user input cases.

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,112 @@
...
+            DISPLAY_NAME="${{ matrix.board }}"
+            ARTIFACT_NAME="${{ matrix.board }}-zmk"
+          fi
+          echo ::set-output name=extra-cmake-args::${EXTRA_CMAKE_ARGS}
+          echo ::set-output name=artifact-name::${ARTIFACT_NAME}
+          echo ::set-output name=display-name::${DISPLAY_NAME}
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Cache west modules
+        uses: actions/cache@v2
```
filterpaper, line 67:
```suggestion
      - name: Checkout
        uses: actions/checkout@v2

      - name: Cache west modules
        uses: actions/cache@v2
```

Suggestion: empty line between steps to improve readability.

elagil: Sure, why not

.github/workflows/build-user-config.yml
elagil: It says 'This function calculates an individual SHA-256 hash for each matched file', so it does seem to respect file content?

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,124 @@
...
+
+      - name: West Build (${{ steps.variables.outputs.display-name }})
+        run: |
+          west build -s zmk/app -b ${{ matrix.board }} -- -DZMK_CONFIG=${GITHUB_WORKSPACE}/${{ inputs.config_path }} ${{ steps.variables.outputs.extra-cmake-args }} ${{ matrix.cmake-args }}
+      - name: ${{ steps.variables.outputs.display-name }} DTS File
+        if: ${{ always() }}
+        run: |
+          if [ -f "build/zephyr/${{ matrix.board }}.dts.pre.tmp" ]; then cat -n build/zephyr/${{ matrix.board }}.dts.pre.tmp; fi
+          if [ -f "build/zephyr/zephyr.dts" ]; then cat -n build/zephyr/zephyr.dts; fi
+
```
petejohanson, line 105: These lines should be zapped for Zephyr 3.0 and onward.

docs/blog/2022-03-08-zephyr-3-0-upgrade-prep.md
```diff
@@ -13,16 +13,16 @@ their repositories now.
 GitHub Actions needs to use our latest Docker image to ensure continued compatibility with the ZMK codebase on Zephyr 3.0 (and beyond). You should:
 
 - Open `.github/workflows/build.yml` in your editor/IDE
-- Change `zmkfirmware/zmk-build-arm:2.5` to `zmkfirmware/zmk-build-arm:stable` wherever it is found
+- Replace its content with the [updated template](https://github.com/zmkfirmware/unified-zmk-config-template/blob/main/.github/workflows/build.yml)
 
-Once the changes are committed and pushed, the build will run as expected.
+The new configuration template does not include the pipeline definition itself, but rather references a central configuration file. Thus, you will not have to update this file again in the future, as all updates to the central configuration will automatically be reflected in your new pipeline runs.
```
petejohanson, line 18: I'd rather we announce this with an additional new blog post, since this one was already announced a while ago, and read by folks.

petejohanson (changes requested):
Thanks! A few comments.

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,124 @@
...
+          fi
+          echo ::set-output name=extra-cmake-args::${EXTRA_CMAKE_ARGS}
+          echo ::set-output name=artifact-name::${ARTIFACT_NAME}
+          echo ::set-output name=display-name::${DISPLAY_NAME}
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Cache west modules
+        uses: actions/cache@v2
```
filterpaper, line 72:
```suggestion
        uses: actions/cache@v3.0.1
```
Align this change with #1204

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,124 @@
...
+            ARTIFACT_NAME="${{ matrix.shield }}-${{ matrix.board }}-zmk"
+            DISPLAY_NAME="${{ matrix.shield }} - ${{ matrix.board }}"
+          else
+            EXTRA_CMAKE_ARGS=
+            DISPLAY_NAME="${{ matrix.board }}"
+            ARTIFACT_NAME="${{ matrix.board }}-zmk"
+          fi
+          echo ::set-output name=extra-cmake-args::${EXTRA_CMAKE_ARGS}
+          echo ::set-output name=artifact-name::${ARTIFACT_NAME}
+          echo ::set-output name=display-name::${DISPLAY_NAME}
```
filterpaper, line 66:
```suggestion
          echo ::set-output name=display-name::${DISPLAY_NAME}
          echo ::set-output name=zephyr-version::${ZEPHYR_VERSION}
```

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,124 @@
...
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Cache west modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-zephyr-modules
```
filterpaper, line 74:
```suggestion
          cache-name: cache-zephyr-${{ steps.variables.outputs.zephyr-version }}-modules
```

filterpaper (changes requested):
This was discussed with @petejohanson on Discord—GH actions was reusing old caches from Zephyr 2.5 on 3.0 when the hash for user's `west.yml` remained unchanged. These proposed changes will trigger cache-misses with Zephyr version update.

docs/blog/2022-03-08-zephyr-3-0-upgrade-prep.md
elagil: Makes sense, I removed it.

.github/workflows/build-user-config.yml
elagil: Ok!

.github/workflows/build-user-config.yml
elagil: Updated

elagil:
All suggestions were incorporated, please check if I missed something.

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,120 @@
...
+        env:
+          cache-name: cache-zephyr-${{ steps.variables.outputs.zephyr-version }}-modules
+        with:
+          path: |
+            modules/
+            tools/
+            zephyr/
+            bootloader/
+            zmk/
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('${{ inputs.config_path }}/west.yml') }}
```
petejohanson, line 83: The use of the `inputs` here isn't working as expected on pushes. It ends up not finding the `west.yml` file at all in that scenario. See https://github.com/petejohanson/testing-3-0-update-zmk-config/runs/5809977931?check_suite_focus=true

petejohanson (commented):
On a whole, looks god. Just one question/comment on the cache key.

.github/workflows/build-user-config.yml
elagil: Yes, that is a bit strange. I will investigate.

petejohanson: I really like the idea of this change. Wonder if we could skip the input parameters for workflow run for now, to get this in, then iterate?

filterpaper: Perhaps `hashFiles('**/west.yml', '**/build.yaml')` will be sufficient for unique hashes from user's config folder at this point in time?

elagil: Yes, I think that has the same effect, since you hash the same files.

joelspadin: Maybe GH actions don't support interpolating inside single quote strings or nested interpolations? Using globs like that seems like it should work well though.

elagil: I found references of GH actions that do path concatenation this way with success, so I don't think it's a problem. However, it could also be changed to something like `join( <array of  path components>, '/' )`. In any case, globs are probably most general and catch all matching files too.

.github/workflows/build-user-config.yml
```diff
@@ -0,0 +1,120 @@
...
+
+      - name: West Zephyr export
+        run: west zephyr-export
+
+      - name: West Build (${{ steps.variables.outputs.display-name }})
+        run: |
+          west build -s zmk/app -b ${{ matrix.board }} -- -DZMK_CONFIG=${GITHUB_WORKSPACE}/${{ inputs.config_path }} ${{ steps.variables.outputs.extra-cmake-args }} ${{ matrix.cmake-args }}
+
+      - name: ${{ steps.variables.outputs.display-name }} Kconfig file
+        run: cat build/zephyr/.config | grep -v "^#" | grep -v "^$"
```
filterpaper, line 103:
```suggestion
        run: cat build/zephyr/.config | grep -v "^#" | grep -v "^$" | sort
```

Create a sorted output that will be easier to review.

filterpaper, line 103:
```suggestion
        run: cat build/zephyr/.config | grep -v "^#" | grep -v "^$" | sort
```

Create a sorted output that will be easier to review.

elagil: Seems like a reasonable addition.

elagil:
@petejohanson Do you have any further suggestions?

petejohanson (approved):
Ran wonderfully over at https://github.com/petejohanson/testing-3-0-update-zmk-config/runs/6069419685?check_suite_focus=true

No more concerns, this will definitely help us iterate on this, add enhancements like the work from @dxmh in https://github.com/zmkfirmware/unified-zmk-config-template/pull/2 into one spot, and have it propagate to config repos automatically. Thanks!

PR #1184, opened 2022-03-23
New Feature: Allow external power on USB only

infused-kim:
Hey guys,

this PR adds a feature that automatically disables external power when you disconnect the USB cable and automatically enables it again when you re-connect it.

There's already PR https://github.com/zmkfirmware/zmk/pull/1113 for the same feature, but it was not working properly for me. 

The biggest issue is that it was implemented in `ext_power_generic.c` and when turning the board on without a USB cable plugged in, it was cutting the power very early on in the initialization chain and caused OLEDs to not work later on.

In my PR it turns off the power much later and avoids these issues.

I have tested this PR on a sofle choc split keyboard with underglow and OLEDs and everything is working well.


autoferrit:
Has this been updated at all? I am running the sofle choc and would love to see this working in mainline zmk.

PR #1187, opened 2022-03-25
feat(behaviors): Adding global quick tap 

andrewjrae:
Adding an additional option to hold taps to make the quick tap functionality apply to all taps, rather than just the given hold tap. The current implementation uses the key state listener and is perfectly functional. However, I would've liked to have been able to identify any key being tapped (even if it's internal) as well as ignore any time a key was tapped in conjunction with a hold. I looked into this but it seemed more complicated then it was worth since this current implementation covers almost all cases while being much simpler.

So far I've been daily driving this implementation for almost a week and loving it! The fact that it gets rid of almost all input delay when typing is an added bonus I hadn't even considered.

### DIFF
--- a/docs/docs/behaviors/hold-tap.md
+++ b/docs/docs/behaviors/hold-tap.md
@@ -50,7 +50,25 @@ In QMK, unlike ZMK, this functionality is enabled by default, and you turn it of
 
 #### `global-quick-tap`
 
-If global quick tap is enabled, then the quick tap will apply not only when the given hold tap is tapped, but for any key tap before it. This effectively disables the hold tap when typing quickly, which is can be quite useful for homerow mods. It can also have the effect of removing the input delay when typing quickly.
+If global quick tap is enabled, then the quick tap will apply not only when the given hold tap is tapped, but for any key tap before it. This effectively disables the hold tap when typing quickly, which can be quite useful for home row mods. It can also have the effect of removing the input delay when typing quickly.
+
+For example, the following hold tap configuration enables global quick tap with a 125 millisecond term.
+```
+gqt: global-quick-tap {
+	compatible = "zmk,behavior-hold-tap";
+	label = "GLOBAL_QUICK_TAP";
+	#binding-cells = <2>;
+	flavor = "tap-preferred";
+	tapping-term-ms = <200>;
+	quick-tap-ms = <125>;
+	global-quick-tap;
+	bindings = <&kp>, <&kp>;
+};
+```
+
+If you press `&kp A`, and then `&gqt LEFT_SHIFT B` **within** 125 ms after, then `ab` will be output. Importantly, the `b` will be output immediately since it was within the `quick-tap-ms`. This behavior will hold for any key that is output to the computer. The `&gqt LEFT_SHIFT B` binding will only have its underlying hold-tap behavior if it is pressed 125 ms **after** any keycode is sent to the computer.
+
+Note that the higher the `quick-tap-ms`, the harder it will be to use the hold behavior, making this unideal for capitalizing letter while typing normally. However, paired with [mod-tap](mod-tap.md)
 
 #### `retro-tap`
 
--- a/docs/docs/behaviors/hold-tap.md
+++ b/docs/docs/behaviors/hold-tap.md
@@ -50,9 +50,9 @@ In QMK, unlike ZMK, this functionality is enabled by default, and you turn it of
 
 #### `global-quick-tap`
 
-If global quick tap is enabled, then the quick tap will apply not only when the given hold tap is tapped, but for any key tap before it. This effectively disables the hold tap when typing quickly, which can be quite useful for home row mods. It can also have the effect of removing the input delay when typing quickly.
+If global quick tap is enabled, then `quick-tap-ms` will apply not only when the given hold-tap is tapped but for any key tap before it. This effectively disables the hold tap when typing quickly, which can be quite useful for home row mods. It can also have the effect of removing the input delay when typing quickly.
 
-For example, the following hold tap configuration enables global quick tap with a 125 millisecond term.
+For example, the following hold-tap configuration enables global quick tap with a 125 millisecond term.
 ```
 gqt: global-quick-tap {
 	compatible = "zmk,behavior-hold-tap";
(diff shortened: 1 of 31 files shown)


okke-formsma (changes requested):
I think we can merge the 'global' and 'local' quick-tap behaviors a bit;

* keep the `last_tapped` and `store_last_tapped` names, rename  `last_tapped.tap_deadline` to `timestamp` as you did.
* change the signature of `store_last_tapped` to take a position + timestamp
* store the last tapped key timestamp (either a normal keypress or the hold-tap) in `last_tapped.timestamp`

This would make the global & local behaviors use the same data, and fix a bug in the current quick-tap implementation too. In the current implementation, only hold-tap keys are considered for quick-taps, which I think is a bug;

*    tap hold-tap A
*    tap &kp B
*    tap hold-tap A again

this sequence should never trigger local 'quick-tap', but if the keys are mashed fast enough it will in the existing implementation.

okke-formsma (changes requested):
Looks good. 

I don't love the solution on line 634 but I'm not sure if my suggestion is much better. I'd love to hear some arguments for or against. I'm not opposed to merging as-is.

andrewjrae:
I agree @okke-formsma, the solution to avoid hold events updating the last tapped timestamp is very hack. We briefly discussed what I would consider to be the proper solution on Discord the other day (https://discord.com/channels/719497620560543766/719544230497878116/961355224566624286), but that relies on a larger, future, feature.

If we want to use timestamps to identify the holds I'm fine with that as well, it's definitely a bit less hack, though it will require an extra variable / logic to keep track of the last hold timestamp. I'll leave it up to whoever else wants to review ;).

At the very least I think some comments should likely be added to note the preferred implementation down the line.

docs/docs/behaviors/hold-tap.md
```diff
@@ -48,6 +48,10 @@ If you press a tapped hold-tap again within `quick-tap-ms` milliseconds, it will
 
 In QMK, unlike ZMK, this functionality is enabled by default, and you turn it off using `TAPPING_FORCE_HOLD`.
 
+#### `global-quick-tap`
+
+If global quick tap is enabled, then the quick tap will apply not only when the given hold tap is tapped, but for any key tap before it. This effectively disables the hold tap when typing quickly, which is can be quite useful for homerow mods. It can also have the effect of removing the input delay when typing quickly.
+
```
petejohanson, line 54: I don't think regular users will understand what this means. Can we be a bit more verbose, or give some example sequences so this is clear?

petejohanson (commented):
Changes make sense, one comment on the documentation.

docs/docs/behaviors/hold-tap.md
```diff
@@ -48,6 +48,28 @@ If you press a tapped hold-tap again within `quick-tap-ms` milliseconds, it will
 
 In QMK, unlike ZMK, this functionality is enabled by default, and you turn it off using `TAPPING_FORCE_HOLD`.
 
+#### `global-quick-tap`
+
+If global quick tap is enabled, then the quick tap will apply not only when the given hold tap is tapped, but for any key tap before it. This effectively disables the hold tap when typing quickly, which can be quite useful for home row mods. It can also have the effect of removing the input delay when typing quickly.
```
caksoylar, line 53:
```suggestion
If global quick tap is enabled, then `quick-tap-ms` will apply not only when the given hold-tap is tapped but for any key tap before it. This effectively disables the hold behavior when typing quickly, which can be quite useful for home row mods. It can also have the effect of removing the input delay when typing quickly.
```

docs/docs/behaviors/hold-tap.md
```diff
@@ -48,6 +48,28 @@ If you press a tapped hold-tap again within `quick-tap-ms` milliseconds, it will
...
+
+For example, the following hold tap configuration enables global quick tap with a 125 millisecond term.
```
caksoylar, line 55:
```suggestion
For example, the following hold-tap configuration enables global quick tap with a 125 millisecond term.
```

docs/docs/behaviors/hold-tap.md
```diff
@@ -48,6 +48,28 @@ If you press a tapped hold-tap again within `quick-tap-ms` milliseconds, it will
...
+	#binding-cells = <2>;
+	flavor = "tap-preferred";
+	tapping-term-ms = <200>;
+	quick-tap-ms = <125>;
+	global-quick-tap;
+	bindings = <&kp>, <&kp>;
+};
+```
+
+If you press `&kp A`, and then `&gqt LEFT_SHIFT B` **within** 125 ms after, then `ab` will be output. Importantly, the `b` will be output immediately since it was within the `quick-tap-ms`. This behavior will hold for any key that is output to the computer. The `&gqt LEFT_SHIFT B` binding will only have its underlying hold-tap behavior if it is pressed 125 ms **after** any keycode is sent to the computer.
```
caksoylar, line 69:
```suggestion
If you press `&kp A` and then `&gqt LEFT_SHIFT B` **within** 125 ms, then `ab` will be output. Importantly, `b` will be output immediately since it was within the `quick-tap-ms`. This behavior will hold for any key that is output to the computer. The `&gqt LEFT_SHIFT B` binding will only have its underlying hold-tap behavior if it is pressed 125 ms **after** any keycode is sent to the computer.
```

> This behavior will hold for any key that is output to the computer

It's not very clear to me what this means. Does it mean it only applies to `kp` behaviors?

> The `&gqt LEFT_SHIFT B` binding will only have its underlying hold-tap behavior if it is pressed 125 ms **after** any keycode is sent to the computer

I think involving "sent to the computer" might be confusing here, since AFAICT the timing logic is not related to sending HID events. Maybe we can just say "after any key is pressed"  (or released? I am not sure which one when we say "tapped.")

docs/docs/behaviors/hold-tap.md
```diff
@@ -48,6 +48,28 @@ If you press a tapped hold-tap again within `quick-tap-ms` milliseconds, it will
...
+
+Note that the higher the `quick-tap-ms`, the harder it will be to use the hold behavior, making this unideal for capitalizing letter while typing normally. However, paired with [mod-tap](mod-tap.md)
```
caksoylar, line 71: This seems to be missing a part.

caksoylar (commented):
Some docs questions and nits.

docs/docs/behaviors/hold-tap.md
andrewjrae: Yikes, I forgot to save the file, I actually had a different blurb that was supposed to be there.

docs/docs/behaviors/hold-tap.md
andrewjrae: Currently the implementation hooks into the keycode state changed event, which is reflective of what actual HID codes are being sent. But I agree that it's likely not the best way to describe the behavior. I've switched to what I hope is a more clear description.

caksoylar (approved):

dxmh (approved):
Docs look great to me, thanks!

I believe @okke-formsma and @petejohanson approve of these changes based on their [comments][1] [above][2] and [in Discord][3], but it would be good to have at least one explicit approval of the code on this PR before merging – just to be sure!

[1]: https://github.com/zmkfirmware/zmk/pull/1187#pullrequestreview-936886421
[2]: https://github.com/zmkfirmware/zmk/pull/1187#pullrequestreview-945054356
[3]: https://discord.com/channels/719497620560543766/883452966114324550/968341158566719538

petejohanson (approved):

PR #1193, opened 2022-03-29
fix(docs): Fix Windows unicode macro example

caksoylar:
Windows unicode input using alt + number sequence requires using the keypad keycodes, as noted in [the linked doc](https://support.microsoft.com/en-us/office/insert-ascii-or-unicode-latin-based-symbols-and-characters-d13f58d3-7bcb-44a7-a4d5-972ee12e50e0):

>  You must use the numeric keypad to type the numbers, and not the keyboard. Make sure that the NUM LOCK key is on if your keyboard requires it to type numbers on the numeric keypad.

This PR fixes the example sequences and improves the description slightly. Tested before and after the fix and confirmed the keypad keycodes are required.

### DIFF
--- a/docs/docs/behaviors/macros.md
+++ b/docs/docs/behaviors/macros.md
@@ -187,14 +187,14 @@ bindings
 
 ### Unicode Sequences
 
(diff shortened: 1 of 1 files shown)


petejohanson (approved):
This is what I get for putting in an untested example. Thanks for the fix!

PR #1196, opened 2022-03-29
feat(docs): Add aggregated metadata to website

joelspadin:
Added /hardware-metadata.json to the ZMK website. This gives external tools a way to grab all hardware metadata at once.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -44,7 +44,9 @@
   },
   "devDependencies": {
     "@docusaurus/module-type-aliases": "^2.0.0-beta.3",
+    "@docusaurus/types": "^2.0.0-beta.3",
     "@tsconfig/docusaurus": "^1.0.2",
+    "@types/js-yaml": "^4.0.5",
     "@types/react": "^17.0.3",
     "@types/react-helmet": "^6.1.0",
     "@types/react-router-dom": "^5.1.7",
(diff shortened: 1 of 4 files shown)


docs/package.json
```diff
@@ -44,7 +44,9 @@
   },
   "devDependencies": {
     "@docusaurus/module-type-aliases": "^2.0.0-beta.3",
+    "@docusaurus/types": "^2.0.0-beta.18",
     "@tsconfig/docusaurus": "^1.0.2",
+    "@types/js-yaml": "^4.0.5",
```
joelspadin, line 49: Pulled in types for js-yaml and docusaurus for type checking the plugin.

docs/package.json
```diff
@@ -44,7 +44,9 @@
   },
   "devDependencies": {
     "@docusaurus/module-type-aliases": "^2.0.0-beta.3",
+    "@docusaurus/types": "^2.0.0-beta.18",
```
petejohanson, line 47: Can we not pull in a version that matches our other docusaurus deps?

docs/package.json
```diff
@@ -58,7 +60,7 @@
     "prebuild-webpack-plugin": "^1.1.1",
     "prettier": "2.3.1",
     "string-replace-loader": "^3.0.3",
-    "typescript": "^4.2.3",
+    "typescript": "^4.6.3",
```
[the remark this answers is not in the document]
petejohanson: Still true if we stick w/ matching docusaurus dep versions?

petejohanson (changes requested):
A few questions about the deps. Otherwise, looks good.

docs/package.json
joelspadin: Done.

joelspadin:
For some reason I can't respond directly to your other comment. It looks like the older version of the types works with the older version of TypeScript.

petejohanson (approved):
Thanks, LGTM!

PR #1203, opened 2022-03-30
fix(docs): Allow CORS for hardware-metadata.json

joelspadin:
Enabled CORS so external tools can fetch the metadata file.


petejohanson (approved):
Seems reasonable. Thanks.

PR #1208, opened 2022-04-03
refactor(tests): Use GH Actions matrix for tests.

petejohanson:
* To parallelize our tests, generate a dynamic matrix
  of tests to run.

@Nicell Thoughts on this to cut down on our test time? See https://github.com/petejohanson/zmk/actions/runs/2084238754 for an example.

If we do this, we might be to the point we want to get a Teams subscription to GH, since that would bump our concurrent builds up from 20 to 60: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,24 @@ on:
       - "app/src/**"
 
 jobs:
-  integration_test:
+  collect-tests:
+    outputs:
+      test-dirs: ${{ steps.test-dirs.outputs.test-dirs }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Find test directories
+        id: test-dirs
+        run: |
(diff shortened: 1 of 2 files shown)


Nicell (approved):
Looks like there are some big time savings (>60% reduction)! It's a nice short update to achieve parallelization. LGTM!

As for the Team subscription. Would it cost us $48 per person with privileges to the zmkfirmware group per year? Or could we just have one person on the "Team". Curious about what constitutes being a user rather than just a contributor.

petejohanson:
> As for the Team subscription. Would it cost us $48 per person with privileges to the zmkfirmware group per year? Or could we just have one person on the "Team". Curious about what constitutes being a user rather than just a contributor.

@Nicell I expect it would be the headcount for all the users add to our organization, which is 9 currently, but could be dropped to 8 until inoovaker returns. That's $8 * 40 / year = $320 / year.

PR #1211, opened 2022-04-03
feat(battery)!: Make sensor fetch configurable

joelspadin:
Added a ZMK_BATTERY_FETCH Kconfig option to control how battery.c fetches sensor samples. This allows compatibility with more fuel gauge drivers, as some assert if asked to fetch individual channels, while others assert if asked to fetch all channels.

The new default behavior is to use sensor_sample_fetch() to fetch all channels at once. If your driver does not support this, set
CONFIG_ZMK_BATTERY_FETCH_INDIVIDUAL_CHANNELS=y to return to the previous behavior of using sensor_sample_fetch_chan().

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -323,6 +323,23 @@ config ZMK_EXT_POWER
 	bool "Enable support to control external power output"
 	default y
 
+choice ZMK_BATTERY_FETCH
+	prompt "Battery sensor fetch mode"
+
+config ZMK_BATTERY_FETCH_INDIVIDUAL_CHANNELS
+	bool "Use sensor_sample_fetch_chan()"
+	help
+	  Use sensor_sample_fetch_chan() to fetch the individual battery sensor
+	  channels instead of fetching all channels.
+
+config ZMK_BATTERY_FETCH_ALL
+	bool "Use sensor_sample_fetch()"
+	help
+	  Use sensor_sample_fetch() to fetch all battery sensor channels. Use this
+	  if your battery sensor driver requires fetching SENSOR_CHAN_ALL.
+
+endchoice
+
 #Power Management
 endmenu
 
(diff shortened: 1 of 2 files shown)


app/Kconfig
```diff
@@ -389,6 +389,23 @@ config ZMK_EXT_POWER
 	bool "Enable support to control external power output"
 	default y
 
+choice ZMK_BATTERY_FETCH
+	prompt "Battery sensor fetch mode"
+
+config ZMK_BATTERY_FETCH_ALL
+	bool "Use sensor_sample_fetch()"
```
petejohanson, line 396: Do we really need labels on these? Won't these mostly be in `<board>_defconfig` files?

joelspadin: It appears that options in a `choice` require a prompt. I guess I could turn this into a single `bool` and assume that these are the only two options that will ever exist though.

joelspadin: Keeping this as a `choice` as discussed on Discord. Using a single `bool` would let us hide the option from menuconfig, but then it's difficult to find a name that accurately describes the behavior both when it's on and off.

app/src/battery.c
```diff
@@ -27,7 +27,13 @@ uint8_t zmk_battery_state_of_charge() { return last_state_of_charge; }
 static int zmk_battery_update(const struct device *battery) {
     struct sensor_value state_of_charge;
 
+#if defined(CONFIG_ZMK_BATTERY_FETCH_ALL)
+    int rc = sensor_sample_fetch(battery);
+#elif defined(ZMK_BATTERY_FETCH_INDIVIDUAL_CHANNELS)
     int rc = sensor_sample_fetch_chan(battery, SENSOR_CHAN_GAUGE_STATE_OF_CHARGE);
+#else
+#error "No value selected for ZMK_BATTERY_FETCH"
+#endif
```
petejohanson, line 36: Because the default w/ this change is "fetch all", this changes the default fetch style, which potentially risks some breakage for any drivers that don't support that. Thoughts on having "individual channels" be the default choice instead?

joelspadin: Done.

joelspadin:
I think we can just abandon this in favor of switching to the fuel gauge API that Zephyr is adding.

PR #1217, opened 2022-04-05
fix(behaviors): Remove `behavior_hold_tap_data`

kurtis-lew:
Related to discussion during development of tap-dance behavior:
https://github.com/zmkfirmware/zmk/pull/1139#discussion_r810564682

This PR suggests to remove the `struct behavior_hold_tap_data` because it is not used to store data for each hold tap.


petejohanson (approved):
Thanks!

PR #1223, opened 2022-04-06
Stop persisted binding keypress fallthrough

DoctorNefario:
Needed because k_work_reschedule can return positive success codes.

This fixes an issue where bindings such as `RGB_TOG` or `EP_OFF` would also press the key underneath them in the keymap, and never release it.


petejohanson:
@DoctorNefario Looks like you need to run this through `clang-format` to fix the formatting errors.

petejohanson (approved):
LGTM. Thanks!

PR #1227, opened 2022-04-07
fix(hid): Use a full valid range for consumer page

petejohanson:
* Switch to a logical max for the consumer page
  that avoid signed issue, and still allows
  full range of documented consumer page
  values.

### DIFF
--- a/app/include/zmk/hid.h
+++ b/app/include/zmk/hid.h
@@ -78,9 +78,9 @@ static const uint8_t zmk_hid_report_desc[] = {
     HID_REPORT_SIZE(0x08),
 #elif IS_ENABLED(CONFIG_ZMK_HID_CONSUMER_REPORT_USAGES_FULL)
     HID_LOGICAL_MIN8(0x00),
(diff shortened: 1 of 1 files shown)


joelspadin (approved):
Tested on Windows 10: I did need to re-pair before it would work, but I tpot checked C_VOL_DN, C_VOL_UP, and C_AL_CALC and they all worked both over USB and BT.

joelspadin:
>  I did need to re-pair before it would work,

Disregard. I tried a couple more times and it worked without needing to re-pair. Either Windows messed up the first time or it was just taking longer than normal to connect after flashing.

caksoylar:
I also tested on Windows 11 with the usual suspects: `C_VOL_DN`, `C_VOL_UP`, `C_PLAY_PAUSE` and `C_AL_CALC`. They all continued working with USB and BT without re-pairing.

For context, `C_AL_CALC` does not work with `CONFIG_ZMK_HID_CONSUMER_REPORT_USAGES_BASIC=y`.

Issue #1226, opened 2022-04-07
Document new `ignore-modifiers` option for sticky keys


petejohanson:
In #1122, @okke-formsma added a new `ignore-modifiers` property you can set on sticky keys, but the docs got missed for documenting the option. Need to add to `docs/docs/behaviors/sticky-key.md` in the "Configuration" section.

dxmh:
Fixed in #1228

Issue #903, opened 2021-08-09
Sticky shift produces multiple uppercase letters


dxmh:
I use `&sk LSHFT` on a thumb key as my primary method of typing an uppercase letter, so that holding down keys is not necessary (this behaviour is called “one-shot shift” in QMK). 

However, this almost always results in two uppercase letters for me. 

For example:

> HEllo world. THis is an example for the GItHUb issue. 

This feels like it might be exacerbated by a “rolling” typing style, e.g. typing fast and possibly rolling some of the `&sk LSHFT` `&kp H` `&kp E` keys in the `HEllo world` example above. 

I think this may be related to #829.

trankillity:
I am also experiencing this. Initially I thought it might have been lazy typing is definitely more related to ZMK. I believe that the sticky keys code is occasionally applying the stickied key on key UP rather than just on key DOWN. This would explain why this is happening.

You can reproduce this in a simple fashion by holding a `&sk LSHIFT` bind for a second, not pressing anything, then pressing a key before the `release-after-ms` period. It still produces the capital letter, when it should really only produce it on key down, not key up.

Use cases where this causes significant issues are when using modifiers and the mouse (shift clicking, etc).

dxmh:
> I believe that the sticky keys code is occasionally applying the stickied key on key UP rather than just on key DOWN.

Interesting observation. It could be worth trying the `quick-release` config option which is enabled by default for sticky _layer_ keys:

> Sticky layers behave slightly different from sticky keys. They are configured to `quick-release`. This means that the layer is released immediately when another key is pressed. "Normal" sticky keys are not `quick-release`; they are released when the next key is released. This makes it possible to combine sticky layers and sticky keys as such: tap `&sl 1`, tap `&sk LSHIFT` on layer 1, tap `&kp A` on base layer to output shift+A.

https://zmk.dev/docs/behaviors/sticky-layer/#advanced-usage

trankillity:
> Interesting observation. It could be worth trying the `quick-release` config option which is enabled by default for sticky _layer_ keys:

Very interesting! How would I go about adding that in? No docs about it at all unfortunately. Is it as simple as the `hold-tap` configuration of `flavor`?

trankillity:
Have just found the new docs generated in [this PR](https://github.com/zmkfirmware/zmk/pull/722) which link to [these docs](https://deploy-preview-722--zmk.netlify.app/docs/config/behaviors/#sticky-key).

[...]

PR #1228, opened 2022-04-07
Update sticky key documentation with new settings

uosl:
Documentation for [sticky key behavior](https://zmk.dev/docs/behaviors/sticky-key) does not mention `quick-release` like the [sticky layer behavior](https://zmk.dev/docs/behaviors/sticky-layer) page. #903 shows that there is a use case for this setting for sticky keys. This PR adds these findings to the documentation page so people with this problem don't end up having to search through Github issues.

Edit: Included documentation for #1226

### DIFF
--- a/docs/docs/behaviors/sticky-key.md
+++ b/docs/docs/behaviors/sticky-key.md
@@ -7,12 +7,6 @@ sidebar_label: Sticky Key
 
 A sticky key stays pressed until another key is pressed. It is often used for 'sticky shift'. By using a sticky shift, you don't have to hold the shift key to write a capital.
 
-By default, sticky keys stay pressed for a second if you don't press any other key. You can configure this with the `release-after-ms` setting.
-
-Some typists may find that using a sticky shift key interspersed with rapid typing results in two or more capitalized letters instead of one. This happens as the sticky key is active until the next key is released, under which other keys may be pressed and will receive the modifier. You can change this with the `quick-release` setting which will instead deactivate the sticky key on the next key being pressed, as opposed to released.
-
-The sticky behavior is only triggered on release, and if the next key is pressed before the sticky key is released, it will behave like a normal key. This also applies if the next key is another sticky key. If you use *callum-style mods* where you are prone to rolling sticky keys, you can allow for this with the `ignore-modifiers` setting.
-
 ### Behavior Binding
 
 - Reference: `&sk`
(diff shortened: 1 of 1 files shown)


petejohanson:
@uosl Thanks! Since you're already in there, are you up fro adding the `ignore-modifiers` details as well, as noted as missing in #1226 ?

uosl:
Oh no, I was afraid you'd ask of that 😛
But I agree, it wouldn't make sense for anyone else to do it. I'll add it to this PR (maybe tomorrow).

petejohanson:
> Oh no, I was afraid you'd ask of that stuck_out_tongue But I agree, it wouldn't make sense for anyone else to do it. I'll add it to this PR (maybe tomorrow).

Do good work, and you end up getting asked to do even more! :grinning: 

Thanks!

uosl:
I added `ignore-modifiers` to the documentation now as well.

docs/docs/behaviors/sticky-key.md
```diff
@@ -28,11 +32,13 @@ You can use any keycode that works for `&kp` as parameter to `&sk`:
...
 ### Configuration
 
-You can configure a different `release-after-ms` in your keymap:
+You can configure a different `release-after-ms`, enable `quick-release` or enable `ignore-modifiers` in your keymap:
 
 ```
 &sk {
     release-after-ms = <2000>;
+    quick-release;
+    ignore-modifiers;
```
caksoylar, line 41: I believe `ignore-modifiers;` are already set for `&sk` by default: https://github.com/zmkfirmware/zmk/blob/32c8737a2225ce03455d7df0145a96a451672c7b/app/dts/behaviors/sticky_key.dtsi#L15 
We should probably note this in the section where we mention the property.

For this example, we can change it to remove the property with  `/delete-property/` maybe?

docs/docs/behaviors/sticky-key.md
```diff
@@ -7,7 +7,11 @@ sidebar_label: Sticky Key
 
 A sticky key stays pressed until another key is pressed. It is often used for 'sticky shift'. By using a sticky shift, you don't have to hold the shift key to write a capital.
 
-By default, sticky keys stay pressed for a second if you don't press any other key. You can configure this with the `release-after-ms` setting (see below).
```
caksoylar, line 10: Can we move the description of the properties to the configuration section below (before the two example snippets in it)? We can then prepend each paragraph with an header with the property name like 

```
### `release-after-ms`
```

This would make it more consistent with the other behavior doc pages like [`hold-tap`](https://zmk.dev/docs/behaviors/hold-tap/) and [`tap-dance`](https://zmk.dev/docs/behaviors/tap-dance).

docs/docs/behaviors/sticky-key.md
uosl: Can we think of any scenarios for which you'd want to disable `ignore-modifiers`? If not I'll just describe what it does and side-step why you'd want to do it.

docs/docs/behaviors/sticky-key.md
uosl: I'll do this. I wasn't sure how to format the document, so those examples help a lot.

okke-formsma:
lgtm

okke-formsma (approved):

caksoylar (approved):

petejohanson:
@uosl are you able to run `npm run prettier:format` from the docs directory to fix the formatting? If not, I can do that part for you.

dxmh (approved):

PR #1231, opened 2022-04-08
fix(docs): Fix cmake version requirements

caksoylar:
`cmake` now requires version `>3.20.0`, so update Debian/Ubuntu instructions. Note that 22.04 LTS will officially be released on April 21st but it will have [version 3.22.1](https://packages.ubuntu.com/jammy/cmake). Also added PyPI as an easy to install method recommendation, which is also [what ZMK Docker image uses](https://github.com/zmkfirmware/zmk-docker/blob/d69745c8be52487bb47dede1a93456076a778b78/Dockerfile#L28).

### DIFF
--- a/docs/docs/development/setup.md
+++ b/docs/docs/development/setup.md
@@ -70,9 +70,7 @@ sudo apt install -y \
 ```
 
 :::note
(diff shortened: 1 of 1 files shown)


petejohanson (approved):
Thanks!

PR #1232, opened 2022-04-08
feat(behaviors): Support parameterized macros.

petejohanson:
* Add two new compatibles for macros that
      take one or two parameters when bound in
      a keymap.
    * Use `&macro_param_1to1`, `&macro_param_1to2`, `&macro_param_2to1`,
      and `&macro_param_2to2` control entries in the bindings for the macro
      to have the next binding entry have it's values substituted.

Ok, draft for now until I can create updated docs, but hoping for some feedback on this idea. We use more marker bindings to trigger different behavior for the next binding entry, e.g. `&macro_param_1to1`.

So a new macro that surround a given letter with quotes:

```
		quote_letter_macro: quote_letter_macro {
			#binding-cells = <1>;
			label = "ZMK_QLET";
			compatible = "zmk,behavior-macro-one-param";
			wait-ms = <40>;
			tap-ms = <40>;
			bindings = <
				&kp QUOT
				&macro_param_1to1 &kp MACRO_PLACEHOLDER
				&kp QUOT>;
		};
```

Could then be used in your keymap like `&quote_letter A`.

Thoughts?


joelspadin:
The most likely scenario I can think where this could cause an issue is if a behavior packs multiple values into one parameter such that 0xFFFF is a valid and common value, for example `RGBW(255, 255, 255, 255)` for a 4-channel LED color.

joelspadin:
Copying over my suggestion from Discord so it doesn't get lost:

What if instead of having a special parameter value that gets replaced, we add a macro-specific behavior that modifies the following binding? For example, we could have something like

- `&macro_param1_1st` replace 1st param with macro param 1
- `&macro_param1_2nd` replace 2nd param with macro param 1
- `&macro_param2_1st` replace 1st param with macro param 2
- `&macro_param2_2nd` replace 2nd param with macro param 2
- `&macro_param1` alias for `&macro_param1_1st` (for bindings that only take one param)
- `&macro_param2` alias for `&macro_param2_1st`

petejohanson:
> 

I dig this idea. Implementation seemed simple enough. I opted for `&macro_param_1to1`, `&macro_param_2to1` but I'm *really* open to some feedback/ideas on better naming on how to handle this.

I also left a define `MACRO_PLACEHOLDER` in, that could be used just for *clarity* in the bindings for a macro, but you could just as easily just hardcode `0` there if you wanted.

From the test:

```
		quote_letter_macro: quote_letter_macro {
			#binding-cells = <1>;
			label = "ZMK_QLET";
			compatible = "zmk,behavior-macro-one-param";
			wait-ms = <40>;
			tap-ms = <40>;
			bindings = <
				&kp QUOT
				&macro_param_1to1 &kp MACRO_PLACEHOLDER
				&kp QUOT>;
		};
```

Thoughts?

petejohanson:
@joelspadin Rebased onto the new Zephyr 3.2 `main`, still ready for review. TIA.

kvietcong:
Used to make a dynamic CTRL+KEY macro for easy use with a tap_hold key.

No weird behaviors as everything seems to be working just fine, allowing me to hold keys like V (`&tct V V`) to get a CTRL+V *tap* (not hold like with a regular hold_tap). With some documentation, I think the feature would be really easy to use for anyone. I also needed to make sure to `#include <dt-bindings/zmk/macro.h>`, which wasn't mentioned in the PR yet.

<details><summary>Actual Configuration</summary>
<code>

    behaviors: {
        tct: tap_ctrl_tap {
            label = "TAP_CTRL_TAP";
            #binding-cells = <2>;
            tapping_term_ms = <300>;
            flavor = "tap-preferred";
            bindings = <&ctrl_kp>, <&kp>;
        };
    };

    macros {
        ctrl_kp: macro_ctrl_kp {
            #binding-cells = <1>;
            label = "MACRO_CTRL_KP";
            compatible = "zmk,behavior-macro-one-param";
            bindings
                = <&macro_press &kp LCTL>
                , <&macro_param_1to1 &macro_tap &kp MACRO_PLACEHOLDER>
                , <&macro_release &kp LCTL>;
        };
    };

</code>
</details>

app/src/behaviors/behavior_macro.c
```diff
@@ -4,31 +4,36 @@
...
 #include <zephyr/logging/log.h>
 #include <zmk/behavior.h>
 #include <zmk/behavior_queue.h>
 #include <zmk/keymap.h>
+#include <dt-bindings/zmk/macro.h>
 
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
-#if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
+#if DT_HAS_COMPAT_STATUS_OKAY(zmk_behavior_macro) ||                                               \
```
joelspadin, line 17: Could this be simplified using Kconfig + CMake, or would you rather do that in a separate commit?

app/src/behaviors/behavior_macro.c
```diff
@@ -110,21 +137,54 @@ static int behavior_macro_init(const struct device *dev) {
...
+    switch (state->param2_source) {
+    case PARAM_SOURCE_MACRO_1ST:
+        binding->param2 = macro_binding->param1;
+        break;
+    case PARAM_SOURCE_MACRO_2ND:
+        binding->param2 = macro_binding->param2;
+        break;
+    default:
+        break;
+    }
```
joelspadin, line 163: This could be deduplicated with something like

```c
    update_param(&binding->param1, state->param1_source, &macro_binding);
    update_param(&binding->param2, state->param2_source, &macro_binding);
```

Or more functional programming style:

```c
    binding->param1 = update_param(state->param1_source, binding->param1, &macro_binding);
    binding->param2 = update_param(state->param2_source, binding->param2, &macro_binding);
```

joelspadin (commented):
Needs docs updates, but the implementation looks good to me.

app/src/behaviors/behavior_macro.c
petejohanson: Yes, it can be! Rebased onto `main` where we have the `Kconfig.behaviors` and tweaked this to use the DT_HAS defines and simplify the code by only compiling if we have any of the macro DT compatibles enabled.

app/src/behaviors/behavior_macro.c
petejohanson: Good catch. Used the more functional style.

docs/docs/behaviors/macros.md
```diff
@@ -62,6 +78,32 @@ bindings
...
+
+```
+bindings
+    = <&macro_param_1to1>
+    , <&kp MACRO_PLACEHOLDER>
+    ;
+```
+
+:::note
+The `MACRO_PLACEHOLDER` is defined as a convenient way to avoid just putting `0` in that location. It does _not_ control what parameters are substituted; only the macro controls do that.
```
joelspadin, line 96: Might be useful to first explain that you still need to put *something* in for every binding, even if it will just get overwritten due to `&macro_param_XtoY`. Otherwise, it might be confusing why you would need to put `0` there, and therefore why you might use this instead.

docs/docs/behaviors/macros.md
```diff
@@ -62,6 +78,32 @@ bindings
 There are a set of special macro controls that can be included in the `bindings` list to modify the
 way the macro is processed.
 
+### Parameters
+
+When creating a macro that takes parameter(s), there are control when the parameters passed to the macro are used
+within the macro itself. All of the controls are "one shot" and will change how the passed in parameters are used for the very next behavior in the `bindings` list of the macro.
```
joelspadin, line 84: Might be useful to clarify that it applies to the next non-macro control behavior, so `&macro_param1to1 &macro_param2to2 @macro_press &foo 0 0` is a valid thing that works.

petejohanson: Tweaked the wording here.

docs/docs/behaviors/macros.md
petejohanson: Yeah, good call. I expanded on this and pulled it out of the `note` aside it had been hiding in.

app/src/behaviors/behavior_macro.c
```diff
@@ -166,22 +214,20 @@ static const struct behavior_driver_api behavior_macro_driver_api = {
...
+        .count = DT_PROP_LEN(n, bindings),                                                         \
         .bindings = TRANSFORMED_BEHAVIORS(n)};                                                     \
-    DEVICE_DT_INST_DEFINE(n, behavior_macro_init, NULL, &behavior_macro_state_##n,                 \
-                          &behavior_macro_config_##n, APPLICATION,                                 \
-                          CONFIG_KERNEL_INIT_PRIORITY_DEFAULT, &behavior_macro_driver_api);
-
-DT_INST_FOREACH_STATUS_OKAY(MACRO_INST)
+    DEVICE_DT_DEFINE(n, behavior_macro_init, NULL, &behavior_macro_state_##n,                      \
+                     &behavior_macro_config_##n, APPLICATION, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT, \
+                     &behavior_macro_driver_api);
```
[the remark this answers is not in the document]
petejohanson: Renamed.

docs/docs/behaviors/macros.md
```diff
@@ -62,6 +78,30 @@ bindings
 There are a set of special macro controls that can be included in the `bindings` list to modify the
 way the macro is processed.
 
+### Parameters
+
+When creating a macro that takes parameter(s), there are control when the parameters passed to the macro are used
```
caksoylar, line 83: This sentence (specifically "there are control when the parameters [...]") seems to be broken, I couldn't tell the intention.

petejohanson: This was... incomprehensible.... Remind me not to write docs late at night when I'm tired. Reworded, to hopefully actually make sense now. Thanks!

docs/docs/behaviors/macros.md
```diff
@@ -49,6 +49,22 @@ For use cases involving sending a single keycode with modifiers, for instance ct
...
+    raise_layer {
+        bindings = <&my_cool_macro A>
+    };
+```
+
+When defining a parameterized macro, a different `compatible` value will be used depending on how many parameters are passed into it:
+
+- `zmk,behavior-macro` - a parameter that takes no parameters.
+- `zmk,behavior-macro-one-param` - a parameter that takes one parameter when used.
+- `zmk,behavior-macro-one-param` - a parameter that takes two parameters when used.
```
caksoylar, line 66:
```suggestion
- `zmk,behavior-macro-two-param` - a parameter that takes two parameters when used.
```

docs/docs/behaviors/macros.md
```diff
@@ -67,6 +83,30 @@ bindings
...
+    = <&macro_param_1to1>
+    , <&kp MACRO_PLACEHOLDER>
+    ;
+```
+
+Because `kp` takes one parameter, you can't simply make the second entry `<&kp>` in the `bindings` list. Whatever value you do pass in will be replaced when the macro is triggered, so you can put _any_ value there, e.g. `0`, `A` keycode, etc. To make it very obvious that the parameter there is not actually going to be used, you can use `MACRO_PLACEHOLDER` which is simply an alias for `0`.
+
+The available parameter controls are:
+
+- `&macro_param_1to2` - pass the first parameter of the macro into the first parameter of the next behavior in the `bindings` list.
```
caksoylar, line 104:
```suggestion
- `&macro_param_1to1` - pass the first parameter of the macro into the first parameter of the next behavior in the `bindings` list.
```

app/tests/macros/place-holder-parameters/native_posix_64.keymap
caksoylar, line 1: Might be good to have a test using `&macro_param_1to2` and `&macro_param_2to2` as well.

caksoylar (approved):
Docs LGTM, noted a couple things in the logging and about the tests.

app/tests/macros/place-holder-parameters/native_posix_64.keymap
petejohanson: Supplemented that test to include those two param substitutions. Good call.

PR #1239, opened 2022-04-13
fix(usb): Spilt HID from core USB, logging fix.

petejohanson:
* Split core USB init from USB HID init.
* Tweak logging to avoid "log loop" causing
  spurious buffer messages on startup.


app/Kconfig
```diff
@@ -477,6 +477,10 @@ if ZMK_USB_LOGGING
 config ZMK_LOG_LEVEL
 	default 4
 
+# We do this to avoid log loop where logging to USB generates more log messages.
+config USB_CDC_ACM_LOG_LEVEL
+    default 1
```
Nicell, line 482: Consistency with the rest of the file
```suggestion
	default 1
```

Nicell (commented):
Tiny Kconfig format change. Splitting USB/USB_HID makes sense.

app/Kconfig
petejohanson: Fixed.

Nicell (approved):

PR #1242, opened 2022-04-13
Fix battery charge icon out of screen boundary

WSTRN:


petejohanson:
Can you get some pictures of the before and after here? I'd never noticed it truncating.

WSTRN:
> Can you get some pictures of the before and after here? I'd never noticed it truncating.

Sorry, I didn't take the previous photos. The part I circled up outside the screen, about two or three pixels. This problem may occur after the Bluetooth icon shows that the connection is successful.
I just expanded the icon display area, and I don't think it's the best solution.

37257F73-463E-419D-8F31-F2DBAA2F95EF
5CFA65A9-3758-4DAC-B288-6532848B009C
F3A101FA-B3BC-47E1-86BA-509A72662268

WSTRN:
I use this screen.
7799A8DC-8094-4074-B498-5569496A4B6B

petejohanson:
Thanks! Makes sense to me.

petejohanson (approved):

PR #1246, opened 2022-04-15
fix(ble): Restore manual connection params

Nicell:


petejohanson (approved):
Nick, cleaning up my mistakes again. :thanks:

PR #1248, opened 2022-04-18
feat(docs): Provide current status of Displays

dxmh:
This small PR updates the [current "OLED Displays" feature page](https://zmk.dev/docs/features/displays/) to provide the user with information about the current status of display support in ZMK, as discussed in https://github.com/zmkfirmware/zmk/issues/1219.

This is intended to be only an interim placeholder page for Displays, to provide the user with something more than the current TODO, until displays are officially supported and documented.

This change also renames the page from _OLED Displays_ to just _Displays_ because of other display technologies such as ePaper, etc. ("OLED" still appears in the body copy, so that this page is remains discoverable through searches for "OLED").

Any suggestions to adjust the phrasing or expand upon this PR is welcomed! 🙂

PR #911, opened 2021-08-12
fix(docs): Standardize all internal links

Nicell:
Supersedes #729 
Fixes #906 #901

We haven't been using proper internal documentation linking, so I've standardized it all.

Related issue: https://github.com/facebook/docusaurus/issues/2298

PR #1249, opened 2022-04-18
fix(docs): Ensure relative links always resolve

dxmh:
After #1248 was merged, I noticed that the destination of the relative link can sometimes be incorrect depending on the trailing slash of the current URL:

* Access the Displays page _via the sidebar_ then the current URL is `https://zmk.dev/docs/features/displays` and the relative link works: `https://zmk.dev/docs/behaviors/power` ✅

* Access the Displays page _directly_ (or refresh the above) then the current URL becomes `https://zmk.dev/docs/features/displays/` (a trailing slash is added), so of course the relative link results in a 404: `https://zmk.dev/docs/features/behaviors/power` ❌

The fix is to link to the document _file path_ rather than the document _URL_; this ensures that the link resolves regardless of trailing slash config – Docusaurus just takes care of it. More information is at https://docusaurus.io/docs/docs-markdown-features. (Note: Other URLs in the docs already use this convention, but this particular link slipped through in the previous PR.)

We might want to look into changing the wider behaviour so that pages always have a trailing slash (or they do not), but that's probably for a separate thread. (Update: I opened #1250.)

### DIFF
--- a/docs/docs/features/displays.md
+++ b/docs/docs/features/displays.md
@@ -6,5 +6,5 @@ sidebar_label: Displays
 Displays in ZMK are currently a proof of concept and official support is coming soon.
 
 :::info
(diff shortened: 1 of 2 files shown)


dxmh:
Just spotted the same problem with the link to Macros on https://zmk.dev/docs/#features, so fixed that as part of this PR too.

Nicell (approved):
LGTM. As you said most items use .md/.mdx as of #911, but these slipped through. Thanks for fixing this!

PR #1255, opened 2022-04-19
refactor(battery): Automatically enable battery drivers

joelspadin:
`CONFIG_ZMK_BATTERY_NRF_VDDH` and `CONFIG_ZMK_BATTERY_VOLTAGE_DIVIDER` are now automatically enabled if a DT node with the matching "compatible" property is present and enabled, so they no longer need to be manually set for each board.


app/drivers/sensor/battery/Kconfig
```diff
@@ -1,20 +1,25 @@
...
+DT_COMPAT_ZMK_BATTERY_VOLTAGE_DIVIDER := zmk,battery-voltage-divider
+
 config ZMK_BATTERY
 	bool "ZMK battery monitoring"
 	help
 		Enable battery monitoring
 
 config ZMK_BATTERY_NRF_VDDH
-	bool "ZMK nRF VDDH battery monitoring"
+	bool
```
joelspadin, line 13: I followed the convention I saw in Zephyr of not labeling these (you shouldn't ever have to manually set them anymore) and making a `DT_COMPAT_*` constant to use with `${dt_compat_enabled, ...)`.

joelspadin:
Since our workflows don't pick up on the seeduino being changed, I ran another build to test that here: https://github.com/joelspadin/zmk-config-test/runs/6072956468?check_suite_focus=true

petejohanson (approved):
All makes sense. Nice cleanup of the defconfig bits we gain as well. Thanks!

PR #1256, opened 2022-04-19
fix: Exclude `sof` module to avoid build breakage.

petejohanson:
Zephyr upstream seemed to have force-pushed to their `sof` module, causing the sha1 referenced from the Zephyr `west.yml` to be missing. Exclude for now, we don't use it.

### DIFF
--- a/app/west.yml
+++ b/app/west.yml
@@ -31,5 +31,6 @@ manifest:
           - openthread
           - edtt
           - trusted-firmware-m
+          - sof
   self:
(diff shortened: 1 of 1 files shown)


joelspadin (approved):

joelspadin:
https://github.com/zephyrproject-rtos/zephyr/commit/080c64644b35af426a6d6a5b78bad8f804b560d4 appears to be where this was fixed in Zephyr upstream.

PR #1258, opened 2022-04-22
Edits and improvements to the reusable build workflow

filterpaper:
- Add `artifact_name` as a workflow dispatch option. Users can specify a different output artifact name alongside custom build and config path.
- Use `set -x` for verbose output of shell commands. This will display content of all environment variables and aid troubleshooting.
- Simplify `grep` one-liner command for displaying Kconfig file.
- Steps naming consistency.
- Shell `if` `then` statement consistency.


joelspadin (approved):
Seems reasonable to me.

petejohanson (approved):
Thanks!

PR #1259, opened 2022-04-23
feat(behaviors): Add encoder layer switching

drzony:
Add a possibility to switch layers using encoder.


drzony:
Obsoleted by https://github.com/zmkfirmware/zmk/pull/1275

PR #1263, opened 2022-04-24
refactor(split): Seperate peripheral BLE handling.

petejohanson:
* Move foundational BLE code for split
  peripherals to a dedicated file to avoid
  tons of conditionals and awkward code.

As I prepare for some core changes to our advertising/profile code, I'm moving the peripheral BLE code out to a separate file so the two aren's slammed together with awkward conditionals everywhere.


petejohanson:
Oh, and I'd like to move to the auto-updates as part of this, so ideally we'll actually merge in #862 first.

app/src/split/bluetooth/peripheral.c
```diff
@@ -0,0 +1,118 @@
...
+
+static const struct bt_data zmk_ble_ad[] = {
+    BT_DATA_BYTES(BT_DATA_FLAGS, (BT_LE_AD_GENERAL | BT_LE_AD_NO_BREDR)),
+    BT_DATA_BYTES(BT_DATA_UUID16_SOME,
+                  0x0f, 0x18 /* Battery Service */
+                  ),
+    BT_DATA_BYTES(BT_DATA_UUID128_ALL, ZMK_SPLIT_BT_SERVICE_UUID)
+};
+
+int start_advertising() {
```
joelspadin, line 44: Should this be static?

app/src/split/bluetooth/peripheral.c
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
joelspadin, line 2: Should the copyright date be updated since this is a new file?

app/src/split/bluetooth/peripheral.c
```diff
@@ -0,0 +1,118 @@
...
+    settings_subsys_init();
+
+    settings_load_subtree("ble");
+    settings_load_subtree("bt");
+#endif
+
+#if IS_ENABLED(CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START)
+    LOG_WRN("Clearing all existing BLE bond information from the keyboard");
+
+    for (int i = 0; i < 10; i++) {
```
joelspadin, line 106: The 10 here seems kinda arbitrary. Should this be something more like `CONFIG_BT_MAX_PAIRED`? (ble.c has the same issue. Maybe we should have a file of common bluetooth functions and put this in it so it isn't duplicated between the central and peripheral files?)

joelspadin (changes requested):

app/CMakeLists.txt
```diff
@@ -64,26 +64,33 @@ if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL)
...
+
 target_sources_ifdef(CONFIG_ZMK_RGB_UNDERGLOW app PRIVATE src/behaviors/behavior_rgb_underglow.c)
 target_sources_ifdef(CONFIG_ZMK_BACKLIGHT app PRIVATE src/behaviors/behavior_backlight.c)
-target_sources_ifdef(CONFIG_ZMK_BLE app PRIVATE src/behaviors/behavior_bt.c)
-target_sources_ifdef(CONFIG_ZMK_BLE app PRIVATE src/ble.c)
 target_sources_ifdef(CONFIG_ZMK_BLE app PRIVATE src/battery.c)
+
 if (CONFIG_ZMK_SPLIT_BLE AND (NOT CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL))
 	target_sources(app PRIVATE src/split_listener.c)
 	target_sources(app PRIVATE src/split/bluetooth/service.c)
```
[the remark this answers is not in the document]
petejohanson: Fixed

app/src/split/bluetooth/peripheral.c
petejohanson: Indeed, fixing.

app/src/split/bluetooth/peripheral.c
petejohanson: No, this is actually the BT "ids" that can be used by multiple "identities". We used this long ago, but don't any more, so this can be massively simplified.

petejohanson: Switched to just call once w/ the default ID.

app/src/split/bluetooth/peripheral.c
```diff
@@ -0,0 +1,113 @@
...
+
+static void disconnected(struct bt_conn *conn, uint8_t reason) {
+    char addr[BT_ADDR_LE_STR_LEN];
+    struct bt_conn_info info;
+
+    bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
+
+    LOG_DBG("Disconnected from %s (reason 0x%02x)", log_strdup(addr), reason);
+
+    bt_conn_get_info(conn, &info);
```
joelspadin, line 54: Does this have some side effects that we need? `info` doesn't appears to be initialized but never used.

app/src/split/bluetooth/peripheral.c
```diff
@@ -0,0 +1,113 @@
...
+                             uint16_t timeout) {
+    char addr[BT_ADDR_LE_STR_LEN];
+
+    bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
+
+    LOG_DBG("%s: interval %d latency %d timeout %d", log_strdup(addr), interval, latency, timeout);
+}
+
+static struct bt_conn_cb conn_callbacks = {
+    // .connected = connected,
```
joelspadin, line 79: Was there supposed to be a connected callback, or can this comment be removed?

app/src/split/bluetooth/peripheral.c
petejohanson: Nope, hold over from the original combined code. Will remove.

app/src/split/bluetooth/peripheral.c
petejohanson: Removing.

joelspadin (approved):

petejohanson:
Found an issue w/ endpoints not being built for peripherals causing the display widget not to work. Debating how to fix that. That's the build failure right now w/ our tests (thankfully!)

petejohanson:
@joelspadin Ended up doing the extra step to create a dedicacted API for determining connection status on a peripheral to a central, and then a new dedicated status widget that is only for peripherals, removes the profile number for the display, etc. Re-requesting review since the changes have increased in scope significantly.

app/src/display/widgets/peripheral_status.c
```diff
@@ -0,0 +1,61 @@
...
+struct peripheral_status_state {
+    bool connected;
+};
+
+static struct peripheral_status_state get_state(const zmk_event_t *_eh) {
+    return (struct peripheral_status_state){.connected = zmk_split_bt_peripheral_is_connected()};
+}
+
+static void set_status_symbol(lv_obj_t *label, struct peripheral_status_state state) {
+    char *text =
```
joelspadin, line 30: This should probably be a const pointer. Otherwise you'd be allowed to try to modify a constant string.

app/src/display/widgets/peripheral_status.c
```diff
@@ -0,0 +1,61 @@
...
+        state.connected ? (LV_SYMBOL_WIFI " " LV_SYMBOL_OK) : (LV_SYMBOL_WIFI " " LV_SYMBOL_CLOSE);
+
+    LOG_DBG("connected? %s", state.connected ? "true" : "false");
+    lv_label_set_text(label, text);
+}
+
+static void output_status_update_cb(struct peripheral_status_state state) {
+    LOG_DBG("");
```
joelspadin, line 38: Empty log?

app/src/split/bluetooth/peripheral.c
```diff
@@ -39,16 +41,30 @@ static const struct bt_data zmk_ble_ad[] = {
...
     BT_DATA_BYTES(BT_DATA_UUID128_ALL, ZMK_SPLIT_BT_SERVICE_UUID)};
 
+static bool is_connected = false;
+
 static int start_advertising() {
     return bt_le_adv_start(BT_LE_ADV_CONN, zmk_ble_ad, ARRAY_SIZE(zmk_ble_ad), NULL, 0);
 };
 
+static void connected(struct bt_conn *conn, uint8_t err) {
+    is_connected = (err >= 0);
```
joelspadin, line 51: The docs for this parameter say

> HCI error. Zero for success, non-zero otherwise.

Should positive values be treated as connected?

petejohanson: Testing showed it working fine, but let me tweak to only use `0` as connected, and verify it works properly.

petejohanson: Yup, still worked fine. Updated.

app/CMakeLists.txt
```diff
@@ -63,27 +63,37 @@ if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_BLE_ROLE_CENTRAL)
   target_sources(app PRIVATE src/combo.c)
   target_sources(app PRIVATE src/behavior_queue.c)
   target_sources(app PRIVATE src/conditional_layer.c)
+  target_sources(app PRIVATE src/endpoints.c)
+  target_sources(app PRIVATE src/hid_listener.c)
   target_sources(app PRIVATE src/keymap.c)
+
+  target_sources_ifdef(CONFIG_ZMK_BLE app PRIVATE src/behaviors/behavior_bt.c)
```
Katona, line 70: Just out of curiosity: could this line moved inside the `if (CONFIG_ZMK_BLE)` block in line 72? I am not familiar with CMake, but that block seems to have very similar purpose as this `target_sources_ifdef`.

petejohanson: Indeed it should be, good catch.

Took the opportunity to move a few more "central only" files into this conditional block, that should have been moved previously. This should lighten the peripheral side image size a bit, as an added bonus.

Nicell (approved):

PR #1269, opened 2022-04-26
Kscan cleanup

joelspadin:
Switched the matrix kscan driver over to use the GPIO DT spec macros and updated Kconfig to automatically enable each driver based on DT nodes.

I didn't update the direct and demux drivers similarly because those are still defining functions with macros, and I figured it would be a lot easier to review as separate PRs that align them with how the matrix driver now works.


joelspadin:
I've tested this on a board that uses the composite, matrix, and direct GPIO drivers. I don't have anything that uses the demux driver though.

petejohanson (approved):
Perfect. Thanks for the clean up.

PR #1270, opened 2022-04-26
fix(docs): Correct default Mod-Tap behavior

xudongzheng:


kurtis-lew (approved):
LGTM

dxmh (approved):
Yep, this now aligns with what's set here:

https://github.com/zmkfirmware/zmk/blob/11ab52582c0aa66de221d80795af63ea3a9b5428/app/dts/behaviors/mod_tap.dtsi#L13

PR #1271, opened 2022-04-26
feat(usb): Set default HID interval to 1ms (1000Hz)

Nicell:


joelspadin (approved):
Since most boards should be able to support this, I think it makes sense to set it by default, and then it can be overridden for just boards that don't support it. Looks good to me.

petejohanson (approved):
Worked swimmingly on my stm32f072 BDN9 V2, so good by me.

PR #1275, opened 2022-04-27
feat(beahviors): Allow encoders to use any behavior

nickconway:
This PR adds the sensor-rotate behavior, which will allow for encoders to use any behavior in addition to basic keypresses.

In the documentation I have assumed this will deprecate the existing &inc_dec_kp behavior.

This has not been tested extensively but is confirmed to be working with RGB behaviors.


eXsoR65:
Working with rgb behavior:

https://user-images.githubusercontent.com/19521349/165663211-e25bb3d1-510c-4e8e-abd8-4b43eee00fb1.mov

app/dts/bindings/behaviors/zmk,behavior-sensor-rotate.yaml
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2020 The ZMK Contributors
```
kurtis-lew, line 1: Copyright headers should be updated to the current year (2022)

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
kurtis-lew, line 2: Same thing for updating the year of the copyright header as the devicetree binding (`.yaml`)

docs/docs/behaviors/sensor-rotate.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Configuration
+
+An example of how to implement an encoder to change rgb brightness:
+
+```
+/ {
+    behaviors {
+        rgb_encoder: rgb_encoder {
+            compatible = "zmk,behavior-mod-morph";
```
kurtis-lew, line 21: `zmk,behavior-mod-morph` should be replaced with `zmk,behavior-sensor-rotate`

docs/docs/features/encoders.md
```diff
@@ -23,25 +23,17 @@ Keyboards and macropads with encoder support will typically take the two EC11 pi
...
 
 As an example, a complete `sensor-bindings` for a Kyria with two encoders could look like:
 
 ```
-sensor-bindings = <&inc_dec_kp C_VOL_UP C_VOL_DN &inc_dec_kp PG_UP PG_DN>;
+sensor-bindings = <&rgb_encoder &arrow_encoder>;
 ```
 
-Here, the left encoder is configured to control volume up and down while the right encoder sends either Page Up or Page Down.
+Here, the left encoder is configured to control rgb brightness while the right encoder sends either Left or Right.
```
kurtis-lew, line 36: From the `behaviors/sensor-rotate.md` example, users can find the RGB control behavior that `&rgb_encoder` refers to. However, it's not immediately clear that the `&arrow_encoder` behavior is bound to the left/right arrow keys on rotation.
On that note, if the updated sensor-rotate behavior does deprecate the sensor-rotate-kp behavior, would it be more appropriate to combine the main page, `encoders.md`, with the new behavior page you've created in this PR? If so, I'd think a code block with the `&rgb_encoder` and `&arrow_encoder` sensor-rotate behaviors would fit really well here.

kurtis-lew (commented):
Looking good so far! I'm currently unable to get this feature working on my local hardware, but in the meantime I have a few comments.

kurtis-lew:
image

Another thing that I'm picking up when I build firmware locally using this PR is the error in the image above. In my case, when I rotate my encoder clockwise or counterclockwise, I'm unable to get the correct behavior binding to register. Perhaps the incompatible pointer types may be the reason for this issue. I'll continue to debug things from my end and see if I can confirm this.

eXsoR65:
> image
> 
> Another thing that I'm picking up when I build firmware locally using this PR is the error in the image above. In my case, when I rotate my encoder clockwise or counterclockwise, I'm unable to get the correct behavior binding to register. Perhaps the incompatible pointer types may be the reason for this issue. I'll continue to debug things from my end and see if I can confirm this.

I too get the same warring during the build process though Its still working. 
https://github.com/eXsoR65/bdn9_zmk_config/runs/6198384676?check_suite_focus=true#step:9:65


Here my config: https://github.com/eXsoR65/bdn9_zmk_config

docs/docs/features/encoders.md
kurtis-lew: Another thing I just thought of: if the final decision is to keep the separate `docs/behaviors/sensor-rotate.md`, the behaviors section of `docs/sidebars.js` should also be updated to include the new page, enhancing accessibility.

app/dts/bindings/behaviors/zmk,behavior-sensor-rotate.yaml
nickconway: Whoops! Fixed these.

docs/docs/features/encoders.md
nickconway: I've added the other encoder into the example code for now.

kurtis-lew (approved):

docs/docs/features/encoders.md
```diff
@@ -23,25 +23,37 @@ Keyboards and macropads with encoder support will typically take the two EC11 pi
...
+
+    keymap {
+        ...
+        sensor-bindings = <&rgb_encoder &arrow_encoder>;
+    };
+};
 ```
 
-Here, the left encoder is configured to control volume up and down while the right encoder sends either Page Up or Page Down.
+Here, the left encoder is configured to control rgb brightness while the right encoder sends either Left or Right.
```
kurtis-lew, line 56: Similar note as that for `sensor-rotate.md` for capitalization, but also enhancing clarity on the behaviour of the `arrow_encoder` binding.

```suggestion
Here, the left encoder is configured to control RGB brightness while the right encoder sends either the left or right arrow keys.
```

docs/docs/behaviors/sensor-rotate.md
```diff
@@ -0,0 +1,33 @@
...
+## Summary
+
+The Sensor Rotation behavior sends a different keypress, depending on whether the sensor is rotated clockwise or counter-clockwise.
+
+- If rotated counter-clockwise, the first keycode is sent.
+- If rotated clockwise, the second keycode is sent.
+
+### Configuration
+
+An example of how to implement an encoder to change rgb brightness:
```
kurtis-lew, line 15: Another minor detail in passing: suggesting capitalization of "RGB" for consistency across the docs.
```suggestion
An example implementation of an encoder that changes RGB brightness is shown below:
```

kurtis-lew (changes requested):
Did another pass through this PR to make some more minor suggestions. Thanks for the updates based on the first review!

docs/docs/behaviors/sensor-rotate.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Sensor Rotation
+sidebar_label: Sensor Rotation
+---
+
+## Summary
+
+The Sensor Rotation behavior sends a different keypress, depending on whether the sensor is rotated clockwise or counter-clockwise.
```
drzony, line 8: Maybe change "sends a different keypress" into "triggers a configured behavior"

docs/docs/behaviors/sensor-rotate.md
```diff
@@ -0,0 +1,33 @@
...
+
+- If rotated counter-clockwise, the first keycode is sent.
```
drzony, line 10: Instead of "keycode is sent" maybe "behavior is triggered"

docs/docs/behaviors/sensor-rotate.md
nickconway: I think this wording makes more sense and I've added it to the pr, thanks for the suggestion!

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,99 @@
...
+#include <logging/log.h>
+
+#include <drivers/sensor.h>
+#include <zmk/event_manager.h>
+#include <zmk/events/keycode_state_changed.h>
+#include <zmk/keymap.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
```
joelspadin, line 20: There are some recent improvements to Zephyr (documented in https://docs.zephyrproject.org/2.6.0/guides/kconfig/preprocessor-functions.html) that provide a more efficient way to do this.

First add to app/Kconfig something like
```
DT_COMPAT_ZMK_BEHAVIOR_SENSOR_ROTATE := zmk,behavior-sensor-rotate

config ZMK_BEHAVIOR_SENSOR_ROTATE
    bool
    default $(dt_compat_enabled,$(DT_COMPAT_ZMK_BEHAVIOR_SENSOR_ROTATE))
```

and then in CMakeLists.txt you can use
```cmake
target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_SENSOR_ROTATE app PRIVATE src/behaviors/behavior_sensor_rotate.c)
```
so that this whole file just gets excluded from the build if the behavior isn't used anywhere. Then you don't need to conditionally compile the contents of it.

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
+
+    struct zmk_behavior_binding_event event = {.layer = zmk_keymap_highest_layer_active(),
+                                               .timestamp = timestamp};
+
+    behavior_keymap_binding_pressed(data->triggered_binding, event);
+    k_msleep(5);
+    behavior_keymap_binding_released(data->triggered_binding, event);
```
joelspadin, line 65: You should use `zmk_behavior_queue_add()` so you can queue the press and release without blocking the main thread.

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,99 @@
...
+#define _TRANSFORM_ENTRY(idx, node)                                                                \
+    {                                                                                              \
+        .behavior_dev = DT_LABEL(DT_INST_PHANDLE_BY_IDX(node, bindings, idx)),                     \
+        .param1 = COND_CODE_0(DT_INST_PHA_HAS_CELL_AT_IDX(node, bindings, idx, param1), (0),       \
+                              (DT_INST_PHA_BY_IDX(node, bindings, idx, param1))),                  \
+        .param2 = COND_CODE_0(DT_INST_PHA_HAS_CELL_AT_IDX(node, bindings, idx, param2), (0),       \
+                              (DT_INST_PHA_BY_IDX(node, bindings, idx, param2))),                  \
+    }
+
+#define KP_INST(n)                                                                                 \
```
joelspadin, line 86: Minor: I'm guessing you copied this from the key press behavior? The naming of this doesn't make a lot of sense here.

app/src/behaviors/behavior_sensor_rotate.c
nickconway: Should the `k_msleep(5)` be kept in between the press and release with this method?

joelspadin: No. The point is to allow scheduling the release to happen later without forcing the main thread to do nothing for 5 milliseconds. You can do

```c
zmk_behavior_queue_add(position, &binding, true, 0);
zmk_behavior_queue_add(position, &binding, false, 5);
```

I am not sure how to set the key position for a sensor though. The existing code here just uses 0 (by way of not setting the position when initializing `event`).

(We should probably also make a similar change to behavior_sensor_rotate_key_press.c in a separate PR.)

app/src/behaviors/behavior_sensor_rotate.c
nickconway: I think I copied from mod-morph but it seems like most of the behaviors use KP e.g. tap dance.

kurtis-lew: This is something that will be addressed in an upcoming PR. Functionally, it doesn't matter what the `XX_INST(n)` is labelled as, but consistency with the behavior name is nice to have.

nickconway: Changed the name to match the behavior.

app/src/behaviors/behavior_sensor_rotate.c
petejohanson: Fully agreed on this, that API/infrastructure didn't exist when the original sensor behavior was written, which I imagine this was copied from.

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
+
+    zmk_behavior_queue_add(0, *data->triggered_binding, true, 0);
+    zmk_behavior_queue_add(0, *data->triggered_binding, false, 5);
+
+    return ZMK_BEHAVIOR_OPAQUE;
+
+    // return behavior_sensor_keymap_binding_triggered(data->triggered_binding, sensor, timestamp);
```
petejohanson, line 64:
```suggestion
```

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,92 @@
...
+        data->triggered_binding = (struct zmk_behavior_binding *)&cfg->ccw_binding;
+        break;
+    default:
+        return -ENOTSUP;
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
+
+    zmk_behavior_queue_add(0, *data->triggered_binding, true, 0);
+    zmk_behavior_queue_add(0, *data->triggered_binding, false, 5);
```
petejohanson, line 60: The 5 here is arbitrary, and not great. Maybe we should make this an optional value in the DT, that defaults to 5 to keep the existing functionality unchanged?

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,92 @@
...
+    err = sensor_channel_get(sensor, SENSOR_CHAN_ROTATION, &value);
+
+    if (err) {
+        LOG_WRN("Failed to get sensor rotation value: %d", err);
+        return err;
+    }
+
+    switch (value.val1) {
+    case 1:
+        data->triggered_binding = (struct zmk_behavior_binding *)&cfg->cw_binding;
```
petejohanson, line 48: We bother storing this to local data? Seems like this can just be a local variable we set, then pass to the queue. Then we can skip the data altogether for this driver, just use the const config.

docs/docs/features/encoders.md
```diff
@@ -23,25 +23,37 @@ Keyboards and macropads with encoder support will typically take the two EC11 pi
...
+            compatible = "zmk,behavior-sensor-rotate";
+            label = "RGB_ENCODER";
+            #sensor-binding-cells = <0>;
+            bindings = <&rgb_ug RGB_BRD>, <&rgb_ug RGB_BRI>;
+        };
+        arrow_encoder: arrow_encoder {
+            compatible = "zmk,behavior-sensor-rotate";
+            label = "ARROW_ENCODER";
+            #sensor-binding-cells = <0>;
+            bindings = <&kp LEFT>, <&kp RIGHT>;
```
petejohanson, line 45: Looking at this...  I'm not sure I love how much this complicates the simple case. It might be really nice to have an equivalent generic behavior (this one, but w/ tweaks?) that takes a binding cells value of 2, and bindings *without* params, and forwards on the params bound to it.

It would allow us to redefine `inc_dec_kp` like:

```
/ {
	behaviors {
		/* DEPRECATED: `inc_dec_cp` will be removed in the future */
		/omit-if-no-ref/ inc_dec_cp: inc_dec_kp: behavior_sensor_rotate_key_press {
			compatible = "zmk,behavior-sensor-rotate-var";
			label = "ENC_KEY_PRESS";
			#sensor-binding-cells = <2>;
                        bindings = <&kp>, <&kp>;
		};
	};
};
```

Thoughts?

petejohanson (changes requested):
Thanks for working on this! A few minor comments, and one higher level change we might want to roll into this.

docs/docs/features/encoders.md
nickconway: Added this to the pr

app/src/behaviors/behavior_sensor_rotate_var.c
```diff
@@ -0,0 +1,79 @@
...
+        triggered_binding.param1 = binding->param2;
+        break;
+    default:
+        return -ENOTSUP;
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
+
+    zmk_behavior_queue_add(0, triggered_binding, true, 0);
+    zmk_behavior_queue_add(0, triggered_binding, false, cfg->default_wait_ms);
```
petejohanson, line 59: I'm not convinced this is actually doing what you want. It seems like this really should be:

```suggestion
    zmk_behavior_queue_add(0, triggered_binding, true, cfg->tap_ms);
    zmk_behavior_queue_add(0, triggered_binding, false, 0);
```

app/src/behaviors/behavior_sensor_rotate_var.c
```diff
@@ -0,0 +1,79 @@
...
+static const struct behavior_driver_api behavior_sensor_rotate_var_driver_api = {
+    .sensor_binding_triggered = on_sensor_binding_triggered};
+
+static int behavior_sensor_rotate_var_init(const struct device *dev) { return 0; };
+
+#define SENSOR_ROTATE_VAR_INST(n)                                                                  \
+    static struct behavior_sensor_rotate_var_config behavior_sensor_rotate_var_config_##n = {      \
+        .cw_behavior_dev = DT_LABEL(DT_INST_PHANDLE_BY_IDX(n, bindings, 0)),                       \
+        .ccw_behavior_dev = DT_LABEL(DT_INST_PHANDLE_BY_IDX(n, bindings, 1)),                      \
+        .default_wait_ms = DT_INST_PROP(n, wait_ms),                                               \
```
petejohanson, line 73:
```suggestion
        .tap_ms = DT_INST_PROP(n, tap_ms),                                               \
```

app/dts/bindings/behaviors/zmk,behavior-sensor-rotate.yaml
```diff
@@ -0,0 +1,21 @@
...
+    type: string
+    required: true
+  "#sensor-binding-cells":
+    type: int
+    required: true
+    const: 0
+  bindings:
+    type: phandle-array
+    required: true
+  wait-ms:
```
petejohanson, line 19:
```suggestion
  tap-ms:
```

app/dts/bindings/behaviors/zmk,behavior-sensor-rotate-var.yaml
```diff
@@ -0,0 +1,25 @@
...
+    type: string
+    required: true
+  "#sensor-binding-cells":
+    type: int
+    required: true
+    const: 2
+  bindings:
+    type: phandles
+    required: true
+  wait-ms:
```
petejohanson, line 19:
```suggestion
  tap-ms:
```

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,86 @@
...
+        triggered_binding = (struct zmk_behavior_binding *)&cfg->ccw_binding;
+        break;
+    default:
+        return -ENOTSUP;
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
+
+    zmk_behavior_queue_add(0, *triggered_binding, true, 0);
+    zmk_behavior_queue_add(0, *triggered_binding, false, cfg->default_wait_ms);
```
petejohanson, line 57:
```suggestion
    zmk_behavior_queue_add(0, *triggered_binding, true, cfg->tap_ms);
    zmk_behavior_queue_add(0, *triggered_binding, false, 0);
```

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,86 @@
...
+                              (DT_INST_PHA_BY_IDX(node, bindings, idx, param1))),                  \
+        .param2 = COND_CODE_0(DT_INST_PHA_HAS_CELL_AT_IDX(node, bindings, idx, param2), (0),       \
+                              (DT_INST_PHA_BY_IDX(node, bindings, idx, param2))),                  \
+    }
+
+#define SENSOR_ROTATE_INST(n)                                                                      \
+    static struct behavior_sensor_rotate_config behavior_sensor_rotate_config_##n = {              \
+        .cw_binding = _TRANSFORM_ENTRY(0, n),                                                      \
+        .ccw_binding = _TRANSFORM_ENTRY(1, n),                                                     \
+        .default_wait_ms = DT_INST_PROP_OR(n, wait_ms, 5),                                         \
```
petejohanson, line 80:
```suggestion
        .tap_ms = DT_INST_PROP_OR(n, tap_ms, 5),                                         \
```

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,86 @@
...
+#include <zmk/events/keycode_state_changed.h>
+#include <zmk/behavior_queue.h>
+#include <zmk/keymap.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct behavior_sensor_rotate_config {
+    struct zmk_behavior_binding cw_binding;
+    struct zmk_behavior_binding ccw_binding;
+    int default_wait_ms;
```
petejohanson, line 24:
```suggestion
    int tap_ms;
```

app/src/behaviors/behavior_sensor_rotate_var.c
```diff
@@ -0,0 +1,79 @@
...
+#include <zmk/events/keycode_state_changed.h>
+#include <zmk/behavior_queue.h>
+#include <zmk/keymap.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct behavior_sensor_rotate_var_config {
+    char *cw_behavior_dev;
+    char *ccw_behavior_dev;
+    int default_wait_ms;
```
petejohanson, line 24:
```suggestion
    int tap_ms;
```

petejohanson (changes requested):
Comment on the behavior queue usage, naming tweaks.

I don't love the code duplication between the parameterized version and not, but the amount duplicated is so minimal, I'd say it's fine.

nickconway:
> I don't love the code duplication between the parameterized version and not, but the amount duplicated is so minimal, I'd say it's fine.

We could the duplicated code to another file that gets included in each behavior?

app/src/behaviors/behavior_sensor_rotate_var.c
```diff
@@ -0,0 +1,79 @@
...
+        triggered_binding.param1 = binding->param2;
+        break;
+    default:
+        return -ENOTSUP;
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
+
+    zmk_behavior_queue_add(0, triggered_binding, true, cfg->tap_ms);
+    zmk_behavior_queue_add(0, triggered_binding, false, 0);
```
petejohanson, line 59: Instead of simulating this at the zeroth key position, perhaps better to simulate it at "max key position + sensor position" so there are actually unique? May be necessary for some code that tracks state tied to key position (or might actually break them.... hmmm...)

petejohanson (commented):
One more thought...

ReFil:
Excited to see this as it will make the defeault layer shifting stuff i'vebeen working on much neater. Iimplemented a rudimentary sensor behaviour but universall sensor beaviours are even better

app/src/keymap.c
```diff
@@ -269,7 +269,8 @@ int zmk_keymap_sensor_triggered(uint8_t sensor_number, const struct device *sens
                 continue;
             }
 
-            ret = behavior_sensor_keymap_binding_triggered(binding, sensor, timestamp);
+            struct zmk_behavior_binding_event event = {.position = 0, .timestamp = timestamp};
```
petejohanson, line 272: Thoughts on using `sensor_number` for the `position` value?

```suggestion
            struct zmk_behavior_binding_event event = {.position = sensor_number, .timestamp = timestamp};
```

app/src/behaviors/behavior_sensor_rotate_var.c
```diff
@@ -0,0 +1,81 @@
...
+                                       struct zmk_behavior_binding_event event) {
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_sensor_rotate_var_config *cfg = dev->config;
+
+    struct sensor_value value;
+    int err;
+
+    err = sensor_channel_get(sensor, SENSOR_CHAN_ROTATION, &value);
+
+    if (err) {
```
petejohanson, line 39:
```suggestion
    if (err < 0) {
```

We've been bit before by some Zephyr APIs that change to return positive numbers for some success conditions, so check for negative specifically is safer.

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,88 @@
...
+                                       struct zmk_behavior_binding_event event) {
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_sensor_rotate_config *cfg = dev->config;
+
+    struct sensor_value value;
+    int err;
+
+    err = sensor_channel_get(sensor, SENSOR_CHAN_ROTATION, &value);
+
+    if (err) {
```
petejohanson, line 39:
```suggestion
    if (err < 0) {
```

petejohanson (commented):
Couple minor things. Want to give this a test before merging though.

Thanks for the work on this!

app/src/keymap.c
nickconway: I was just thinking to add something like this earlier, we'd need something to differentiate between sensors

petejohanson:
@nickconway can you please rebase to pull in the docs build fixes?

nickconway:
@petejohanson done.

docs/docs/behaviors/sensor-rotate.md
```diff
@@ -0,0 +1,33 @@
...
+
+The Sensor Rotation behavior triggers a different behavior, depending on whether the sensor is rotated clockwise or counter-clockwise.
+
+- If rotated counter-clockwise, the first behavior is triggered.
+- If rotated clockwise, the second behavior is triggered.
+
+### Configuration
+
+An example implementation of an encoder that changes RGB brightness is shown below:
+
```
petejohanson, line 16: We need an example for the -var and non-var variants here.

app/src/behaviors/behavior_sensor_rotate_var.c
petejohanson: This got fixed in the non-var version, but not this one.

petejohanson (changes requested):
A few minor fixes needed.

Also, we probably need to remove `app/dts/bindings/behaviors/zmk,behavior-sensor-rotate-key-press.yaml` in this PR as well as the driver code itself.

thilojaeggi:
A bit off topic but I added this to my keymap file like so:
`
	sensors {
		sensors = <&left_encoder &right_encoder>;
	};
	
	behaviors {
        rgb_encoder: rgb_encoder {
            compatible = "zmk,behavior-sensor-rotate";
            label = "RGB_ENCODER";
            #sensor-binding-cells = <0>;
            bindings = <&rgb_ug RGB_BRD>, <&rgb_ug RGB_BRI>;
        };
    };
`

but it's not working.

What am I missing here?

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,88 @@
...
+    case -1:
+        triggered_binding = (struct zmk_behavior_binding *)&cfg->ccw_binding;
+        break;
+    default:
+        return -ENOTSUP;
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
+
+    zmk_behavior_queue_add(ZMK_KEYMAP_LEN + event.position, *triggered_binding, true, cfg->tap_ms);
```
joelspadin, line 58: I think this will overlap with combo positions?

petejohanson:
Merged in the updated PR #1758

PR #1282, opened 2022-05-02
feat(split): allow multiple peripherals

bezmi:
Similar to #836 using the up-to-date peripheral handling code. I've just added logic to keep scanning until all config defined peripheral slots are filled.

I've tested this code to work on NRF52840 split hardware, with a makerdiary NRF52840-MDK dongle. [See my WIP config files](https://github.com/bezmi/zmk/tree/e73_keebar_wiris/app/boards/shields/wiris).

I noticed that a single peripheral address is stored via `zmk_ble_set_peripheral_addr()` but this doesn't seem to be used anywhere so I've left it as-is for now.


app/src/split/bluetooth/central.c
```diff
@@ -328,6 +328,12 @@ static void split_central_process_connection(struct bt_conn *conn) {
 
     LOG_DBG("New connection params: Interval: %d, Latency: %d, PHY: %d", info.le.interval,
             info.le.latency, info.le.phy->rx_phy);
+
+    // keep scanning until all peripherals are connected
+    for (int i = 0; i < ZMK_BLE_SPLIT_PERIPHERAL_COUNT; i++) {
+        if (peripherals[i].state != PERIPHERAL_SLOT_STATE_CONNECTED)
+            start_scan();
```
petejohanson, line 335: I think this might be fine for now, but I'm curious if we need to at all considering switching to a lower duty scanning if we fail to find a new peripheral in a certain amount of time. Constant scanning will draw more battery if we have a multi-peripheral with something like an "optional numberpad" that a user doesn't bother connecting/charging regularly.

petejohanson (commented):
Just one thought, generally these changes all look good to me.

app/src/split/bluetooth/central.c
bezmi: I agree, would this be a new `start_scan_limited()` function with a longer `bt_le_scan_param::interval` value set? 

I also had a concern regarding security. If we fail to connect one of the peripherals at power-up and we continue scanning indefinitely, then a third party with any ZMK peripheral in range will be able to hijack that peripheral slot and get up to funny business, right? Maybe a better solution is to allow manual triggering of the split pairing from a key-combination after the initial pairing times out.

Ideally I'd prefer authenticated pairing with a pre-shared key as an option, but that is a separate issue.

xudongzheng: The way security is handled in my initial dongle PR is that the central stores the MAC address of the peripherals and will only reconnect to those once all the devices are known.

See https://github.com/slicemk/zmk/blob/a14004531d2ff6935c2c6c8bcaabcb1c0f5c3572/app/src/ble.c#L284-L312 and https://github.com/slicemk/zmk/blob/a14004531d2ff6935c2c6c8bcaabcb1c0f5c3572/app/src/split/bluetooth/central.c#L265-L271

RomboutKarelse: This code works for me until restarting the dongle. It seems like the peripheral addresses are not stored/restored, or am I doing something wrong here? Should I include https://github.com/zmkfirmware/zmk/pull/836 as well?

petejohanson:
Can you please rebase to account for the recent refactors added to help support the upcoming split wired support?

app/src/split/bluetooth/central.c
petejohanson: I'm fine doing the address work as a follow up. 

I do think the scanning tweaks should get in as part of this though.

xudongzheng: I just started testing the latest ZMK today with the address stuff and also an updated https://github.com/slicemk/zmk/blob/a14004531d2ff6935c2c6c8bcaabcb1c0f5c3572/app/src/split/bluetooth/central.c#L323-L353 for scanning. I'll probably update my old PR in the next day or two.

bezmi:
@petejohanson I was just revisiting this. Are you in favour of closing this for #836? Looks like the other PR is more up to date and contains the same functionality.

bezmi:
closing in favour of the OG #836. Which I have been [testing](https://github.com/bezmi/zmk/tree/bezmi-dongle-pr) and it is working well.

PR #1286, opened 2022-05-05
feat(behavior): Add retro hold tap behavior

nickconway:
This PR adds a behavior that is similar to hold-tap with retro-tap enabled but allows the retro-tap behavior to be separate from the normal tap behavior.  Right now, it is essentially a copy of hold-tap with a few changes.


nickconway:
My original plan for this was to add the option to the normal hold-tap but I couldn't figure out how to have an optional binding. If this is possible, I'd want to move everything into the hold-tap code and have the retro-tap behavior be decided with that optional binding.

nickconway:
Closing to reimplement

PR #1289, opened 2022-05-06
feat(behaviors): Custom Retro Tap Behavior

nickconway:
This PR adds the ability to have a custom binding for a retro-tap, giving 3 possible bindings on a hold-tap.


app/src/behaviors/behavior_hold_tap.c
```diff
@@ -699,6 +716,9 @@ static int behavior_hold_tap_init(const struct device *dev) {
         .global_quick_tap = DT_INST_PROP(n, global_quick_tap),                                     \
         .flavor = DT_ENUM_IDX(DT_DRV_INST(n), flavor),                                             \
         .retro_tap = DT_INST_PROP(n, retro_tap),                                                   \
+        .retro_tap_behavior = DT_INST_PROP(n, retro_tap_behavior),                                 \
```
okke-formsma, line 719: instead of this, just use `.retro_tap_behavior_dev = DT_LABEL(DT_INST_PHANDLE_BY_IDX(n, bindings, 2)),` (although we need to figure out how to make it fall back to the tap binding)

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -483,7 +496,11 @@ static void decide_retro_tap(struct active_hold_tap *hold_tap) {
     if (hold_tap->status == STATUS_HOLD_TIMER) {
         release_binding(hold_tap);
         LOG_DBG("%d retro tap", hold_tap->position);
-        hold_tap->status = STATUS_TAP;
+        if (strcmp(hold_tap->config->retro_tap_behavior, "") == 0) {
```
okke-formsma, line 499: I would make the 'retro_tap_behavior' fall back to the tap behavior. Then you don't need this conditional.

okke-formsma (changes requested):
Interesting idea. I think we can make it a little bit cleaner but I like the idea.

voidyourwarranty2:
As my current keymap depends on this patch, I have produced an updated version that works with the current main branch of ZMK (a lot has changed in the hold-tap behavior since): https://github.com/voidyourwarranty2/zmk/tree/devel-retrotap

PR #1290, opened 2022-05-06
refactor(tests): Move remaining tests to posix 64.

petejohanson:
* Update docs and fix up a few lingering tests
  to consistently use native_posix_64 target.


docs/docs/development/posix-board.md
```diff
@@ -21,18 +21,18 @@ apt install -y gcc-multilib
 ## Building
 
 To do this, you can build ZMK targeting the
-`native_posix` board.
+`native_posix_64` board.
 
 ```
-west build --pristine --board native_posix
+west build --pristine --board native_posix_64
```
Nicell, line 27: I've never built native_posix_64 before, but I'm getting a missing keymap error. It looks like a combination of keymap dirs not including the base boards in cmake as well as no default .keymap file (just overlay).

docs/docs/development/tests.md
```diff
@@ -4,7 +4,7 @@ sidebar_label: Tests
 ---
 
 - Running tests requires [native posix support](posix-board.md).
-- Any folder under `/app/tests` containing `native_posix.keymap` will be selected when running `west test`.
+- Any folder under `/app/tests` containing `native_posix_64.keymap` will be selected when running `west test`.
```
Nicell, line 7: Unrelated to this PR, but `west test` has never worked on Windows because it uses a `.sh` for running the test from the `test.py` file.

docs/docs/development/posix-board.md
petejohanson: Yeah, we don't have a default keymap for the board.... Not sure we really want to have one. Perhaps our better example here would show an example that includes how to do that.

PR #1293, opened 2022-05-08
fix(boards): Add flash/settings for XIAO BLE

petejohanson:
* Ensure flash/NVS/settings configs are enabled.


eXsoR65:
Validated fix on Zaphod Lite with Xiao BLE. Bluetooth profile persists when cutting off power and powering back on. Indicating writes persisting to memory.

PR #1294, opened 2022-05-09
fix(boards): Fix battery ADC channel for Mikoto

zhiayang:
Mikoto uses P0.03 for the vbatt divider, which is AIN1 --- not AIN2 as claimed by the board config file. Again, should've caught this earlier, sorry.


petejohanson (approved):
One liners are.the best/worst. Thanks for the fix.

PR #1302, opened 2022-05-16
Fix typo in C_SUBTITLES

stefansturlu:
The Consumer Closed Caption key had a typo. It was called `C_SUBTITILES` but should be called `C_SUBTITLES`.


petejohanson (approved):
Good catch!

Issue #1273, opened 2022-04-27
Keyboard prevents macOS from sleeping


dxmh:
macOS users often report their machines being woken from sleep  by their keyboard (causing laptop battery drain, etc).

@petejohanson mentioned [on Discord][1]:

> It's likely the battery status notifications waking it.
> We probably need some work on our end to honor when a device is going to sleep to stop doing that.
> Alternately, enabling ZMK deep sleep should help.
> Since the keyboard would sleep also eventually, and stop waking your MB.

### Workaround

For users wishing to workaround this, deep sleep can be enabled and configured in your `<keyboard>.conf` file ([example][3]):

```sh
CONFIG_ZMK_SLEEP=y # Enable deep sleep support
CONFIG_ZMK_IDLE_SLEEP_TIMEOUT=900000 # Milliseconds of inactivity before entering deep sleep
```

(Note that deep sleep is not yet documented, though it is covered on the [Power configuration][2] page from the work-in-progress PR #722.)

[1]: https://discord.com/channels/719497620560543766/719909884769992755/811654995761692703
[2]: https://deploy-preview-722--zmk.netlify.app/docs/config/power
[3]: https://github.com/dxmh/zmk-config/blob/3097036876e6682b6915eddcf75643391e77fff4/config/a_dux.conf#L2-L3

kanru:
Is it possible to change the frequency of the battery status notification being sent? For example I don't need to know battery status per second. Probably hourly is good enough for the purpose of knowing that I need to charge it.

caksoylar:
@kanru This seems to be hard-coded to check every minute; you can try to change the timer constant.

petejohanson:
> @kanru This seems to be hard-coded to check every minute; you can try to change the timer constant at: https://github.com/zmkfirmware/zmk/blob/main/app/src/battery.c#L107
> 
> 

I'd also be very open to moving that hardcoded value into a Kconfig setting, PRs welcomed.

filterpaper:
`CONFIG_ZMK_IDLE_SLEEP_TIMEOUT` is millisecond unit while `CONFIG_ZMK_BATTERY_REPORT_INTERVAL` is in seconds. Aligning both might be helpful for consistency and comparison to set `CONFIG_ZMK_BATTERY_REPORT_INTERVAL` > `CONFIG_ZMK_IDLE_SLEEP_TIMEOUT`

dxmh:
[...]

PR #1303, opened 2022-05-17
feat(battery): Add Kconfig setting for battery level report interval

caksoylar:
Makes the interval between BAS GATT reports for battery level configurable using a new `CONFIG_ZMK_BATTERY_REPORT_INTERVAL` setting. Related: https://github.com/zmkfirmware/zmk/issues/1273#issuecomment-1127103347

There are a few decisions I made that I am not 100% sure of, comments welcome:
- Location of Kconfig: right now this is under the battery driver, but it could be under `app/Kconfig` (in a new submenu under "Basic keyboard setup"?) or is there a better spot?
- Name of setting: I didn't mention BAS GATT since it is not very familiar to users. Should it have "level" in the name?
- Should the delay until first report also be configurable, or equal to the interval? I thought there isn't much use in making this configurable and 1 minute seems like a good value.
- Unit of interval: Minutes might be too coarse, milliseconds might be too fine, so I chose seconds.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -518,6 +518,11 @@ config ZMK_SETTINGS_SAVE_DEBOUNCE
 #SETTINGS
 endif
 
+config ZMK_BATTERY_REPORT_INTERVAL
+	depends on ZMK_BLE
+	int "Battery level report interval in seconds"
+	default 60
+
 #Advanced
 endmenu
 
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -104,7 +104,7 @@ static int zmk_battery_init(const struct device *_arg) {
         return rc;
     }
 
-    k_timer_start(&battery_timer, K_MINUTES(1), K_MINUTES(1));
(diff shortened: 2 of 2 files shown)


petejohanson:
Thanks!

> There are a few decisions I made that I am not 100% sure of, comments welcome:
> 
>     * Location of Kconfig: right now this is under the battery driver, but it could be under `app/Kconfig` (in a new submenu under "Basic keyboard setup"?) or is there a better spot?

Should definitely not be with the driver, since is app code *consuming* the driver. `app/Kconfig` somewhere is good. 

> 
>     * Name of setting: I didn't mention BAS GATT since it is not very familiar to users. Should it have "level" in the name?

I think the name you used is fine, IMHO.

> 
>     * Should the delay until first report also be configurable, or equal to the interval? I thought there isn't much use in making this configurable and 1 minute seems like a good value.

I wouldn't complicate it that much. Just one setting is fine.

> 
>     * Unit of interval: Minutes might be too coarse, milliseconds might be too fine, so I chose seconds.

Agreed on seconds being the right unit here.

caksoylar:
Thanks for the comments! I moved the setting to `app/Kconfig` under the `Advanced` submenu, which sounded appropriate.

PR #1173, opened 2022-03-17
feat(kscan_gpio_direct) Add toggle-mode

kurtis-lew:
feat(kscan_gpio_direct) Add toggle switch driver

Addresses #980 by adding a boolean `toggle-mode` to the direct-wire kscan that automatically sets the required pull for the input pin to remove any quiescent current.

Without `toggle-mode`, pins draw about 244.3uA of current. When `toggle-mode` is enabled, `ACTIVE_LOW` pins had a current drain of 0.0uA while `ACTIVE_HIGH` pins were recorded with a quiescent current of 0.2uA.

PR #722, opened 2021-03-07
docs: Add documentation for config options

joelspadin:
This adds a new "config" section to the sidebar with documentation on (most of) the Kconfig and Devicetree options we support.

I documented everything manually, because I didn't want to try to figure out how to write a Kconfig parser in JavaScript, get it reading files from outside the docs directory, and organize it in a reasonable way. As a future improvement, we could try to generate some of this dynamically so it can't get out of sync with the code.

This also adds a section describing how to actually edit configuration, and I replaced some redundant information in the rest of the documentation with links to the new pages.

PR #1305, opened 2022-05-17
feat(kscan): Add toggle-mode to updated direct-wire kscan

kurtis-lew:
An updated version of #1173. Currently being tested by @ebastler without any issues thus far.

Implements a `toggle-mode` boolean to direct-wire kscans to disable pullup/pulldown resistors on active pins, removing the 244.4uA quiescent current that would otherwise exist without this feature.

### DIFF
--- a/app/drivers/kscan/kscan_gpio_direct.c
+++ b/app/drivers/kscan/kscan_gpio_direct.c
@@ -76,6 +76,7 @@ struct kscan_direct_config {
     struct debounce_config debounce_config;
     int32_t debounce_scan_period_ms;
     int32_t poll_period_ms;
+    bool toggle_mode;
 };
 
 #if USE_INTERRUPTS
@@ -123,6 +124,40 @@ static void kscan_direct_irq_callback_handler(const struct device *port, struct 
     k_work_reschedule(&data->work, K_NO_WAIT);
 }
 #endif
+
+static gpio_flags_t kscan_gpio_get_flags(const struct gpio_dt_spec *gpio, bool active) {
+    if (((BIT(0) & gpio->dt_flags) == BIT(0))) { // Devicetree configured input ACTIVE_LOW
+        if (!active) {
+            return GPIO_ACTIVE_LOW | GPIO_PULL_UP;
+        }
+        return GPIO_ACTIVE_LOW;
+    } else { // Devicetree configured input ACTIVE_HIGH
+        if (!active) {
+            return GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN;
+        }
+        return GPIO_ACTIVE_HIGH;
+    }
+    LOG_ERR("Could not determine proper flags to set for pin %d", gpio->pin);
+    return 0;
+}
+
+static int kscan_inputs_set_flags(const struct kscan_gpio_list *inputs,
+                                  const struct gpio_dt_spec *active_gpio) {
+    gpio_flags_t updated_flags;
+    for (int i = 0; i < inputs->len; i++) {
+        updated_flags =
+            GPIO_INPUT | kscan_gpio_get_flags(&inputs->gpios[i], &inputs->gpios[i] == active_gpio);
+        LOG_DBG("Updated flags equal to: %d", updated_flags);
+
+        int err = gpio_pin_configure(inputs->gpios[i].port, inputs->gpios[i].pin, updated_flags);
+        if (err) {
+            LOG_ERR("Unable to configure flags on pin %d on %s", inputs->gpios[i].pin,
+                    inputs->gpios[i].port->name);
+            return err;
+        }
+    }
+    return 0;
+}
 
 static void kscan_direct_read_continue(const struct device *dev) {
     const struct kscan_direct_config *config = dev->config;
@@ -173,6 +208,9 @@ static int kscan_direct_read(const struct device *dev) {
 
             LOG_DBG("Sending event at 0,%i state %s", i, pressed ? "on" : "off");
             data->callback(dev, 0, i, pressed);
+            if (config->toggle_mode && pressed) {
+                kscan_inputs_set_flags(&config->inputs, &config->inputs.gpios[i]);
+            }
         }
 
         continue_scan = continue_scan || debounce_is_active(state);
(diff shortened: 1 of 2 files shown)


app/drivers/kscan/kscan_gpio_direct.c
```diff
@@ -124,6 +125,40 @@ static void kscan_direct_irq_callback_handler(const struct device *port, struct
...
+        }
+        return GPIO_ACTIVE_LOW;
+    } else { // Devicetree configured input ACTIVE_HIGH
+        if (!active) {
+            return GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN;
+        }
+        return GPIO_ACTIVE_HIGH;
+    }
+    LOG_ERR("Could not determine proper flags to set for pin %d", gpio->pin);
+    return 0;
```
petejohanson, line 141: Seems like this could be simplified:

```suggestion
    gpio_flags_t flags = BIT_MASK(0) & gpio->dt_flags;
    if (active) {
      flags |= flags ? GPIO_PULL_DOWN : GPIO_PULL_UP;
    }
    return flags;
```

app/drivers/kscan/kscan_gpio_direct.c
```diff
@@ -124,6 +125,40 @@ static void kscan_direct_irq_callback_handler(const struct device *port, struct
...
+
+static int kscan_inputs_set_flags(const struct kscan_gpio_list *inputs,
+                                  const struct gpio_dt_spec *active_gpio) {
+    gpio_flags_t updated_flags;
+    for (int i = 0; i < inputs->len; i++) {
+        updated_flags =
+            GPIO_INPUT | kscan_gpio_get_flags(&inputs->gpios[i], &inputs->gpios[i] == active_gpio);
+        LOG_DBG("Updated flags equal to: %d", updated_flags);
+
+        int err = gpio_pin_configure(inputs->gpios[i].port, inputs->gpios[i].pin, updated_flags);
```
petejohanson, line 152: Do we need to resort to this instead of the "DT version" since that only takes *extra* flags, and will keep the pull set?

petejohanson (commented):
A couple questions/thoughts.

app/drivers/kscan/kscan_gpio_direct.c
kurtis-lew: That's actually why we need this one instead of the DT version. When we need to disable the pulls, the bitwise OR used in the DT version to apply the extra flags doesn't allow us to set the required bits to 0.

petejohanson: I wonder, since the pull direction for a direct wire can be *directly* inferred from the active high/low setting, should we document/suggest that the pull direction not be set in the DT? That would seem to simplify the DT, and simplify this code quite a bit.

Thoughts?

kurtis-lew: Definitely agree with you there. Just not sure where in the docs would be most appropriate to include this information. Perhaps we can update docs to include `toggle-mode` after #722 is merged?

kurtis-lew: Actually, on second thought, if we leave the pins floating without a pull, I'm not sure if the toggle-switch will initialize with an active pin properly. I will test this on some hardware to verify if this is an error.

petejohanson: I'm saying the init should calculate the appropriate pull as well and add it if not set via the extra_flags to the DT gpio call.

kurtis-lew: Makes a lot more sense. *Currently* I haven't updated the init to do this yet. I'll start working on that part now.

kurtis-lew: Can confirm things are working properly for both `GPIO_ACTIVE_LOW` and `GPIO_ACTIVE_HIGH`: proper pulls are calculated with those variables as the only flags in the DT.

petejohanson: This can now be moved to `gpio_pin_configure_dt` it seems.

app/drivers/kscan/kscan_gpio_direct.c
```diff
@@ -124,6 +125,32 @@ static void kscan_direct_irq_callback_handler(const struct device *port, struct
 }
 #endif
 
+static gpio_flags_t kscan_gpio_get_flags(const struct gpio_dt_spec *gpio, bool active) {
+    gpio_flags_t flags = BIT(0) & gpio->dt_flags;
+    if (!active) {
+        flags |= flags ? GPIO_PULL_UP : GPIO_PULL_DOWN;
+    }
+    return flags;
+}
```
petejohanson, line 134: Perhaps this should just be getting the *extra* flags, e.g.:

```suggestion
static gpio_flags_t kscan_gpio_get_extra_flags(const struct gpio_dt_spec *gpio, bool active) {
    if (active) {
      return 0;
    }
    return (BIT(0) & gpio->dt_flags) ? GPIO_PULL_UP : GPIO_PULL_DOWN;
}
```

This will work fine, once all the calls are using the DT version of the GPIO configure function.

petejohanson (commented):
Seems closer! Tested just fine on my BDN9, but a few tweaks left for the code, IMHO.

app/drivers/kscan/kscan_gpio_direct.c
```diff
@@ -124,6 +125,32 @@ static void kscan_direct_irq_callback_handler(const struct device *port, struct
 }
 #endif
 
+static gpio_flags_t kscan_gpio_get_extra_flags(const struct gpio_dt_spec *gpio, bool active) {
+    gpio_flags_t flags = BIT(0) & gpio->dt_flags;
```
petejohanson, line 129: I don't think we need to return the first bit here at all, since that will be set in the DT still. We need to *check* it, but not include in the "extra" flags we'll use.

Minor thing, but in case we end up using this some other way later, good to really only just return the extra flags we need, not also include the active bit.

petejohanson (changes requested):
One lingering final nitpick. Thanks for the quick turnaround.

petejohanson (approved):
Thanks!

Mithrandir2k18:
The [docs](https://zmk.dev/docs/config/kscan#devicetree-2) on this aren't quite clear to me. If I have a single-row 18 key half, where each switch connects different GPIO pin to ground, is `toggle-mode` a straight up upgrade to `input-gpios`?

ebastler:
Toggle mode would not work with single switches. It is intended for 2, 3, or 4 way slider switches. kurtis originally developed it for the selector switch BT1/BT2/USB on the electronic materials office Altar-I, and I have not seen many other boards that use similar switches, so it's pretty niche.

Those switches always have at least one GPIO shorted to GND which leads to very high battery draw even in idle. Toggle mode circumvents this, by disabling keyscan on a pin as soon as it is switched to ON, then reenabling it the moment another pin goes to ON, and disabling that pin.

On single keyboard pushbuttons wired from GPIO to GND that issue does not exist, as they are only drawing power when actively pressed down by the user, and once disabled they would never be reenabled as the MCU can not know when it is depressed.

Issue #1144, opened 2022-02-22
Shield revision keymap file in ZMK_CONFIG ignored


puradox:
When executing the setup script, `Kyria Rev2` is shown as a keyboard option. After going through the setup script, `kyria_rev2.conf` and `kyria_rev2.keymap` files are created in `${ZMK_CONFIG}/config`

However, when building the separate `_left` and `_right` sides of the board, the created keymap file is not being used.

```
west build -p -d build/left  -b nice_nano_v2 -- -DZMK_CONFIG="/home/sam/Dev/zmk-config/config" -DSHIELD=kyria_rev2_left
west build -p -d build/right -b nice_nano_v2 -- -DZMK_CONFIG="/home/sam/Dev/zmk-config/config" -DSHIELD=kyria_rev2_right
```

For example,
```
$ west build -p -d build/left -b nice_nano_v2 -- -DZMK_CONFIG="/home/sam/Dev/zmk-config/config" -DSHIELD=kyria_rev2_left
-- west build: making build dir /home/sam/Dev/zmk/app/build/left pristine
-- west build: generating a build system
-- ZMK Config directory: /home/sam/Dev/zmk-config/config
-- Board: nice_nano_v2, /home/sam/Dev/zmk/app/boards/arm/nice_nano, kyria_rev2_left, kyria
-- Using keymap file: /home/sam/Dev/zmk/app/boards/shields/kyria/kyria.keymap
```

The expected behavior is to load `${ZMK_CONFIG}/config/${SHIELD_WITHOUT_LEFT_OR_RIGHT}.keymap`.

Marco-Christiani:
Is there a recommended temporary fix in the meantime?

puradox:
Renaming `kyria_rev2.keymap` to `kyria.keymap` will fix the issue. The same applies to the `.conf` file.

chrisgalvan:
Thx @puradox I just ran into this issue today.

PR #1306, opened 2022-05-18
refactor(build): Fixes for config/overlay loading.

petejohanson:
* Support multiple shields properly, e.g. -DSHIELD="foo bar", which
  allows using existing upstream shields like `ssd1306_128x64` along
  with another ZMK shield.
* Properly *add* the "ZMK config" .conf file to the conf file list,
  so that existing board .conf files in `app/boards/<board>.conf` are
  properly honored still. Needed for overrides to usptream boards when
  combined with a local `<shield>.conf` file.

This is needed for properly building with boards like `seeeduino_xiao_ble` to ensure that `app/boards/seeeeduino_xiao_ble.conf` is picked up. It's also necessary for combining *multiple* shields when building with ZMK, and still locating .keymap/.overlay/.conf files properly with a shield list.


petejohanson:
Ok, I've updates this to *also* fix the issue with `kyria_rev2` reported in #1144 

You can see a "stock kyria user config" repo I created then moved to build w/ this branch: https://github.com/petejohanson/kyria-zmk-config/runs/6631523179?check_suite_focus=true

It properly picks up the user config `kyria_rev2.keymap` and `kyria_rev2.conf` files as expected.

app/cmake/ZephyrBuildConfig.cmake
```diff
@@ -74,34 +97,58 @@ foreach(root ${BOARD_ROOT})
...
 
+# Give a shield like `kyria_rev2_left` we want to use `kyria_rev2` and `kyria` as candidate names for
+# overlay/conf/keymap files.
+if(DEFINED SHIELD)
+	foreach(s ${SHIELD_AS_LIST})
+		message(STATUS ${SHIELD_DIR_${s}})
+		get_filename_component(shield_dir_name ${SHIELD_DIR_${s}} NAME)
+		string(REPLACE "_" ";" S_PIECES ${s})
+		list(LENGTH S_PIECES S_PIECES_LEN)
+		foreach (n RANGE ${S_PIECES_LEN-2})
```
joelspadin, line 121: The -2 here was a bit confusing until I looked up that cmake loops to the stop index inclusive.

Since cmake lists are just semicolon-separated strings, would this work and be easier to understand?

```cmake
while (NOT S_PIECES STREQUAL "")
```

(I assume that `list(POP_BACK S_PIECES)` would eventually result in S_PIECES being an empty string, so the loop would be guaranteed to terminate, but I am not very familiar with cmake.)

app/cmake/ZephyrBuildConfig.cmake
```diff
@@ -74,34 +97,58 @@ foreach(root ${BOARD_ROOT})
...
+		if (DEFINED SHIELD)
+			foreach (s ${shield_candidate_names})
+				if (DEFINED $SHIELD_DIR_${s})
+					get_filename_component(shield_dir_name ${SHIELD_DIR_${s}} NAME)
+				endif()
+				list(APPEND overlay_candidates "${ZMK_CONFIG}/${s}_${BOARD}.overlay")
+				list(APPEND overlay_candidates "${ZMK_CONFIG}/${s}.overlay")
+				if (NOT "${shield_dir_name}" STREQUAL "${s}")
+					list(APPEND config_candidates "${ZMK_CONFIG}/${shield_dir_name}.conf")
+					list(APPEND config_candidates "${ZMK_CONFIG}/${shield_dir_name}_${BOARD}.conf")
```
joelspadin, line 147: Are the candidates searched in order? If so, is there any reason the board-specific version is second here but first in the surrounding cases?

joelspadin (commented):
I don't really know what I'm doing with cmake, but I noted a couple things that seemed odd to me.

app/cmake/ZephyrBuildConfig.cmake
petejohanson: Much nicer, indeed.

app/cmake/ZephyrBuildConfig.cmake
petejohanson: No, good catch. Fixed.

petejohanson:
Verified several permutations in https://github.com/petejohanson/kyria-zmk-config/runs/6735971307?check_suite_focus=true including the kyria fix, works as expected now.

PR #1313, opened 2022-05-21
feat(behaviors): Add option for tapdance to output on each key press

nickconway:
Adds the option for a tap-dance to output each bind on key press similar.

### DIFF
--- a/app/dts/bindings/behaviors/zmk,behavior-tap-dance.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-tap-dance.yaml
@@ -14,3 +14,5 @@ properties:
   tapping-term-ms:
     type: int
     default: 200
+  trigger-each-binding:
+    type: boolean
--- a/app/src/behaviors/behavior_tap_dance.c
+++ b/app/src/behaviors/behavior_tap_dance.c
@@ -158,8 +158,8 @@ static int on_tap_dance_binding_pressed(struct zmk_behavior_binding *binding,
     if (tap_dance->counter < cfg->behavior_count) {
         tap_dance->counter++;
         if (cfg->trigger_each_binding) {
-            behavior_keymap_binding_pressed(&tap_dance->config->behaviors[tap_dance->counter - 1],
-                                            event);
+            zmk_behavior_invoke_binding(&tap_dance->config->behaviors[tap_dance->counter - 1],
+                                        event, true);
             reset_timer(tap_dance, event);
             return ZMK_BEHAVIOR_OPAQUE;
         }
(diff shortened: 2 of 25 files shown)


app/dts/bindings/behaviors/zmk,behavior-tap-dance.yaml
```diff
@@ -13,4 +13,6 @@ properties:
     required: true
   tapping-term-ms:
     type: int
-    default: 200
\ No newline at end of file
+    default: 200
+  on-each-press:
```
petejohanson, line 17: Not sure I love this naming... Brainstorming other ideas:

* `trigger-each-binding`
* `tap-every-binding`
* `tap-each`

Thoughts?

petejohanson (changes requested):
One naming suggestion. Also, need to add tests for this configuration and various edge cases around it too, please.

kurtis-lew:
Thought I'd weigh in here. The `each` keyword sounds the most correct for this application since the word,`every`, could be interpreted that it invokes all keybinds at once, or at least that's my personal reaction.

Some other suggestions:
- `activate-each-binding`
- `invoke-each-binding`
- `timerless`/`disable-tapping-term`
  - From what I understand, this boolean value makes tap-dances activate their bindings independent from their - `tapping-term-ms`. We could even just make it so if `tapping-term-ms == -1`, then this feature is enabled.

nickconway:
> Thought I'd weigh in here. The `each` keyword sounds the most correct for this application since the word,`every`, could be interpreted that it invokes all keybinds at once, or at least that's my personal reaction.
> 
> Some other suggestions:
> 
> * `activate-each-binding`
> * `invoke-each-binding`
> * `timerless`/`disable-tapping-term`
>   
>   * From what I understand, this boolean value makes tap-dances activate their bindings independent from their  `tapping-term-ms`. We could even just make it so if `tapping-term-ms == -1`, then this feature is enabled.

I think I like 'activate-each-binding' the best. This still uses the tapping term for resetting the tapdance though, so we couldn't do the negative tapping term.

kurtis-lew:
Good call. Thanks for the clarification 👍

app/src/behaviors/behavior_tap_dance.c
```diff
@@ -156,6 +157,12 @@ static int on_tap_dance_binding_pressed(struct zmk_behavior_binding *binding,
     // value, invoke the last binding available.
     if (tap_dance->counter < cfg->behavior_count) {
         tap_dance->counter++;
+        if (cfg->trigger_each_binding) {
+            behavior_keymap_binding_pressed(&tap_dance->config->behaviors[tap_dance->counter - 1],
```
caksoylar: FYI you should use `zmk_behavior_invoke_binding` instead, see https://github.com/zmkfirmware/zmk/commit/9e36ebd52587c0364562057466a458fdcfcdc685#diff-7eeba6e44a877bcea40b140ff0aaed1559f31fdea34ba7f71f0223b84c68e881

nickconway: Fixed!

PR #1314, opened 2022-05-23
Split Kconfig/CMake refactor

petejohanson:
Clean up our split Kconfig in preperation for the split wired work in #1117 

Cear seperation between split role, and split transport.

Move to ZMK_SPLIT_ROLE_CENTRAL instead of BLE specific role. Existing BLE specific role config left in (for now) for backwards compat.


petejohanson:
Tested the Corne images from this core build, and it all works 100% as expected.

PR #1318, opened 2022-05-26
Feature: shell `tap` command

petejohanson:
Initial work on making the [Zephyr Shell](https://docs.zephyrproject.org/latest/services/shell/index.html) useful for ZMK.

* Fixes for our Kconfig that enables `SENSOR` to not force it on all the time, which helps for `native_posix_64` target builds.
*Minor native_posix_64 tweaks to allow building without any simulated kscan events, since we will use the shell for that instead.
* Shell command! Can do `key tap 0` to generate press then release of the 0th key position.

This is useful for building a native posix executable to test ZMK by hand, generate key events, test changes to USB via the USB IP native driver, etc.

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -91,5 +91,6 @@ target_sources(app PRIVATE src/workqueue.c)
 target_sources(app PRIVATE src/main.c)
 
 add_subdirectory(src/display/)
+add_subdirectory(src/shell/)
 
 zephyr_cc_option(-Wfatal-errors)
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -380,6 +380,8 @@ endmenu
 
 menu "Advanced"
 
+rsource "src/shell/Kconfig"
+
 menu "Initialization Priorities"
 
 if USB_DEVICE_STACK
--- a/app/boards/native_posix_64.overlay
+++ b/app/boards/native_posix_64.overlay
@@ -14,5 +14,6 @@
         rows = <2>;
         columns = <2>;
         exit-after;
+        events = <>;
     };
 };
--- a/app/src/shell/CMakeLists.txt
+++ b/app/src/shell/CMakeLists.txt
@@ -1,0 +1,1 @@
+target_sources_ifdef(CONFIG_ZMK_SHELL_KEY_POSITIONS app PRIVATE key_positions.c)
--- a/app/src/shell/Kconfig
+++ b/app/src/shell/Kconfig
@@ -1,0 +1,8 @@
+menu "Shell"
+
+config ZMK_SHELL_KEY_POSITIONS
+    bool "Key Position commands"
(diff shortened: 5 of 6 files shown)


manna-harbour:
This is great, thanks! I've tested it and would be happy for it to be merged now.

Suggestions for future work:
- Add docs
  - The command is `key`. Sub commands are `tap`, `press`, and `release`. Each take a key position argument. 
  - Using native_posix incuding the shell is covered by @caksoylar at https://github.com/caksoylar/zmk-posix-testbed.
  - How to enable the shell over USB is currently not documented anywhere.
- https://github.com/zmkfirmware/zmk/issues/1444
- This can be used for playing back recorded keystrokes. Timing can be handled externally, but if a `wait` command was added to the shell, recordings with timing could be played back with only file redirection.

manna-harbour (approved):

caksoylar:
> How to enable the shell over USB is currently not documented anywhere

I added some instructions for that here: https://github.com/caksoylar/zmk-posix-testbed/commit/e55dd8504163144212470c3c6beaf97be806b17c However since it didn't work on my system there might be issues. I'd welcome any fixes if there is an issue with the instructions!

manna-harbour:
> > How to enable the shell over USB is currently not documented anywhere
> 
> I added some instructions for that here: [caksoylar/zmk-posix-testbed@e55dd85](https://github.com/caksoylar/zmk-posix-testbed/commit/e55dd8504163144212470c3c6beaf97be806b17c) However since it didn't work on my system there might be issues. I'd welcome any fixes if there is an issue with the instructions!

Thanks for adding those instructions. Accessing the shell over usbip works for me on Linux, apart from https://github.com/zmkfirmware/zmk/issues/1444, although I did have to manually load modules and some commands require root.

But by accessing the shell over USB, I mean enabling the shell on a physical ZMK MCU board directly connected to the host via USB, and accessing the shell via CDC_ACM in the same way as can be done now with USB logging. It may just require adding https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/shell/shell_module/overlay-usb.conf and https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/shell/shell_module/usb.overlay but I haven't tried that yet.

manna-harbour:
I have a working shield with shell over CDC_ACM now, will share after some cleanup.

manna-harbour:
I've made something like KMonad using this PR. See https://github.com/manna-harbour/xmk/discussions/2.

instance-id:
Hello, 
    I was trying to apply this feature to my Kinesis 360Advantage, but I seem to be stuck. I created a detailed outline of the steps taken in the following Gist. If someone happens to have a few minutes and is willing to take a look to see if they can spot where I might have missed something, I would greatly appreciate it.

https://gist.github.com/instance-id/f52b8efbea082f08de06d08cfe7bb504

Also, if this is not the appropriate place to pose this question, please do let me know and I will remove it and relocate it to the specified location.
Thanks!

Issue #1285, opened 2022-05-04
Documentation Improvement: How to reduce build time


infused-kim:
This is regarding this page:
https://zmk.dev/docs/development/build-flash

If you follow the instructions there, west build re-compiles most code (even if it hasn't been changed) on each build leading to a very long build time.

But you can dramatically reduce it by doing an initial build with all the extra parameters like:
```
west build -d build/sofle/left -b nice_nano_v2 -- -DSHIELD=sofle_left -DZMK_CONFIG="/workspaces/zmk-config/config"
```

And then future builds with just:
```
west build -d build/sofle/left
```

The additional settings are cached in the directory.

On my system a pristine build takes exactly 3min, a non pristine build with all the parameters above takes 2min 20seconds and a build with just the `-d` parameter takes just 1m 20seconds (a 44% reduction from the non pristine build!).

This really adds up when developing and testing frequently.

You can check out out this conversation in development for more details:
https://discord.com/channels/719497620560543766/719544230497878116/971091356791476284

I think this would be a worthy addition to the documentation.

dxmh:
Addressed with https://github.com/zmkfirmware/zmk/pull/1324.

PR #1324, opened 2022-05-30
Documentation update - tip for improving build times

abayomi185:
This PR addresses issue #1285.

Adds a tip block with the solution from the referenced issue.
It is currently under the "Building For Split Keyboards" header but could be elsewhere?

Tested and formatted appropriately.

### DIFF
--- a/docs/docs/development/build-flash.md
+++ b/docs/docs/development/build-flash.md
@@ -89,7 +89,7 @@ west build -d build/right -b nice_nano -- -DSHIELD=kyria_right
 This produces `left` and `right` subfolders under the `build` directory and two separate .uf2 files. For future work on a specific half, use the `-d` parameter again to ensure you are building into the correct location.
 
 :::tip
-Build times can be significantly reduced after the initial build by omitting all build arguments except the build directory, e.g `west build -d build/left`. The additional options from your initial build are cached.
+Build times can be significantly reduced after the initial build by omitting all build arguments except the build directory, e.g. `west build -d build/left`. The additional options and intermediate build outputs from your initial build are cached and reused for unchanged files.
 :::
 
 ### Building from `zmk-config` Folder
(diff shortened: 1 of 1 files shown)


docs/docs/development/build-flash.md
```diff
@@ -88,6 +88,10 @@ west build -d build/right -b nice_nano -- -DSHIELD=kyria_right
 
 This produces `left` and `right` subfolders under the `build` directory and two separate .uf2 files. For future work on a specific half, use the `-d` parameter again to ensure you are building into the correct location.
 
+:::tip
+Build times can be significantly reduced after the initial build by omitting all build arguments except the build directory, e.g `west build -d build/left`. The additional options from your initial build are cached.
```
caksoylar, line 92:
```suggestion
Build times can be significantly reduced after the initial build by omitting all build arguments except the build directory, e.g. `west build -d build/left`. The additional options from your initial build are cached.
```

caksoylar (commented):
Thank you for the contribution! Added a suggestion for a tiny fix.

Can we add another half a sentence to describe why it is sped up, like "by reusing intermediate build outputs for unchanged files"?

abayomi185:
Yes. I've added the suggested description in the latest commit.

caksoylar (approved):

dxmh (approved):
I do think this tip would be useful elsewhere on the page as it applies not only to split keyboards. Possibly within the [Pristine Building](https://deploy-preview-1324--zmk.netlify.app/docs/development/build-flash#pristine-building) section with some minor amendments? (e.g. the example pristine build command does not currently include a build directory argument.)

We can rejig later though; this is a worthy addition as-is. Thanks! 👍

caksoylar:
> I do think this tip would be useful elsewhere on the page as it applies not only to split keyboards. Possibly within the Pristine Building section with some minor amendments?

I totally missed that this is under split keyboards and I agree.

PR #1325, opened 2022-06-03
feat(gpio): Add 595 shift register driver.

petejohanson:
* Use SPI for sending serial data.
* Supports 1-4 shift registers chained together by
  setting `ngios` property appropriately.


petejohanson:
You can see this in use in https://github.com/petejohanson/revxlp-config/blob/main/config/boards/shields/revxlp/revxlp.overlay#L12 for setting up the shifter GPIO controller, and then using it for additional outputs.

petejohanson:
Tested working on my revxlp I've been dailying for a couple days now. Merging.

petejohanson:
Using the correct cs-gpios, it works fine, yes. My Zaphod Lite uses shift registers along with the adafruit sharp display breakout, works fine.

PR #1329, opened 2022-06-03
Ergo S-1 OSE

wizarddata:
New board info: https://github.com/wizarddata/Ergo-S-1

## Board/Shield Check-list

### DIFF
--- a/app/boards/shields/ergos1_ose/ergos1_ose.zmk.yml
+++ b/app/boards/shields/ergos1_ose/ergos1_ose.zmk.yml
@@ -1,0 +1,12 @@
+file_format: "1"
+id: redox
+name: Redox
+type: shield
+url: https://github.com/mattdibi/redox-keyboard
+requires: [pro_micro]
+features:
+  - keys
+  - underglow
+siblings:
+  - redox_left
+  - redox_right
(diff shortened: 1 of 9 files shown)


app/boards/shields/ergos1_ose/ergos1_ose.zmk.yml
```diff
@@ -0,0 +1,12 @@
+file_format: "1"
```
caksoylar, line 1: Looks like this file was copied from redox and not modified.

wizarddata: You're right, I'll update that

caksoylar:
Is there a reason you are not using the `pro_micro` interconnect definitions and not setting `requires: [pro_micro]` in the metadata? You don't seem to be using nano-specific pins but I might be wrong.

wizarddata:
This keyboard didn't come from an existing design that used the pro-micro footprint, so referencing the gpio directly seemed to make more sense.

caksoylar:
Changing them to be `pro_micro` would allow your shield to be used with other boards like `nrfmicro` as well, since [it has different internal pin assignments](https://github-wiki-see.page/m/joric/nrfmicro/wiki/Pinout).

wizarddata:
It was brought to my attention that I need to make some changes to board and file names that will effect this pull request. Should I close this one and open a new PR after the changes are made?

wizarddata:
The changes require that the name of the repository and all filenames change as well, I'm going to briefly repoen this PR to see if that change is possible.

wizarddata:
Looks like those changes can't take after the branch name changed, so I'm going to open a new PR.

PR #1330, opened 2022-06-04
feat(behaviors): Continue on mods in caps word.

petejohanson:
* Continue caps word when mods are changed, and
  match the continue list on combination of
    explicit and implicit modifiers.

### DIFF
--- a/docs/docs/behaviors/caps-word.md
+++ b/docs/docs/behaviors/caps-word.md
@@ -23,8 +23,7 @@ Example:
 
 #### Continue List
 
-By default, the caps word will remain active when any alphanumeric character or the underscore (`UNDERSCORE`) characters are pressed. Any other keycode sent,
-will turn off caps word. If you would like to override this, you can set a new array of keys in the `continue-list` property in your keymap:
+By default, the caps word will remain active when any alphanumeric character or underscore (`UNDERSCORE`), backspace (`BACKSPACE`), or delete (`DELETE`) characters are pressed. Any other non-modifier keycode sent will turn off caps word. If you would like to override this, you can set a new array of keys in the `continue-list` property in your keymap:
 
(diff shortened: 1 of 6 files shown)


petejohanson:
Hit this recently when trying out a keymap w/o a dedicated symbol layer, my underscore requires explicit shift, which was stopping caps word.

petejohanson:
Taking the opportunity to add delete and backspace to the default continue list as well. Updated docs accordingly.

docs/docs/behaviors/caps-word.md
```diff
@@ -23,8 +23,7 @@ Example:
 
 #### Continue List
 
-By default, the caps word will remain active when any alphanumeric character or the underscore (`UNDERSCORE`) characters are pressed. Any other keycode sent,
-will turn off caps word. If you would like to override this, you can set a new array of keys in the `continue-list` property in your keymap:
+By default, the caps word will remain active when any alphanumeric character or the underscore (`UNDERSCORE`), backspace (`BACKSPACE`), or delete (`DELETE`) characters are pressed. Any other keycode sent, will turn off caps word. If you would like to override this, you can set a new array of keys in the `continue-list` property in your keymap:
```
caksoylar, line 26:
```suggestion
By default, the caps word will remain active when any alphanumeric character or underscore (`UNDERSCORE`), backspace (`BACKSPACE`), or delete (`DELETE`) characters are pressed. Any other non-modifier keycode sent will turn off caps word. If you would like to override this, you can set a new array of keys in the `continue-list` property in your keymap:
```
Seems to flow better to me when there are multiple non-alphanumeric characters enumerated. Also mentioned modifiers, which seems worth noting since we are updating the docs already.

caksoylar (approved):

PR #1334, opened 2022-06-05
fix: Fixed conf file loading.

petejohanson:
* Properly locate conf files for the shield dir name.


petejohanson (commented):
I've run this with a fork of the reporter's issue, and built locally, and it fixes the loading of some previously loaded config paths. Merging.

PR #1340, opened 2022-06-08
fix(split): Raise release events on disconnect.

### LINKED ISSUE #1339: layer stays active when half disconnects before sending layer unset
on a corne, accessing a layer from the secondary half and turning it off (or sending it to `&bootloader`) will leave the layer activated for the master half until the secondary half re-connects, and accesses and then leaves the layer in question.

Steps to reproduce:
1. Press and hold position for `&lt LAYERX SPACE` on secondary/right side
2. While holding, unplug the half from all power
3. Layer should be stuck on main/left side

petejohanson:
* When a peripheral disconnects from a centraly, raise position events
  to release any active positions from that
  peripheral.

Closes #1339


petejohanson:
Won't be able to test on hardware for a few days, but posting as draft for now if others want to test.

slashfoo:
I tested this and I can no longer reproduce the issue in the manner described. I mean, I could reproduce for a brief moment, but I was expecting that, since I'm assuming it depends on the disconnect timeout.

I'm curious. Are the events raised for release the "currently pressed" from the half that disconnected only, as opposed to ALL?

PR #1349, opened 2022-06-18
refactor(behaviors): Update Behaviors' KConfig Preprocessor Functions

kurtis-lew:
Updates behaviors to compile only those explicitly used in a keymap, similarly to #1269.

The following conversation is linked due to an issue with Zephyr where the momentary layer behavior driver is still compiled even when unused in a keymap.
https://discord.com/channels/719497620560543766/719544230497878116/987568876474736721


kurtis-lew:
Tested for a day now on my Iris with:
- Key-presses
- Hold-taps
- Tap-dances
- Sticky-keys
- Momentary and Toggle Layers
- Transparent behavior

All of which are compiled properly and do not appear to experience any functional changes as a result of this refactor.

Issue #1361, opened 2022-06-23
When `ZMK_BACKLIGHT_AUTO_OFF_USB=y`, disable backlight at start if no USB.


petejohanson:
If the a single color backlight config setting `ZMK_BACKLIGHT_AUTO_OFF_USB=y`, and we start ZMK from battery, currently the backlight is initialized on, and doesn't turn off, even though USB isn't present. Came up when testing the equivalent behavior for RGB Underglow in #1010 

Need to change the backlight init code to default the state based on the USB state in `app/src/backlight.c`.

caksoylar:
This is fixed by #1362 right?

petejohanson:
Should be, yes. Closing, but feel free to re-open if needed.

PR #1362, opened 2022-06-23
fix(backlight): Improve initial power on behaviour

ReFil:
This is just a minor tweak to address #1361 so that the RGB an backlighting behave the same


petejohanson (approved):
Thanks!

PR #1366, opened 2022-06-28
feat(behaviors): Swapper implementation

nickconway:
Implementation of #997. Smart interrupt is probably not the name we should use, but I was out of ideas, help wanted!


caksoylar:
I have been testing the window swapper implementation as in the example for the last day and it is working well so far. It neatly fixes #997 using a combination of first and second approaches proposed in there.

app/dts/bindings/behaviors/zmk,behavior-smart-interrupt.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+compatible: "zmk,behavior-smart-interrupt"
+
+include: zero_param.yaml
+
+properties:
+  bindings:
+    type: phandle-array
+    required: true
+  shared-key-positions:
```
petejohanson, line 14: Not sure I love the term `shared` for this? Maybe `ignored-key-positions` or `continuation-key-positions`?

docs/docs/behaviors/smart-interrupt.md
```diff
@@ -0,0 +1,129 @@
...
+    };
+};
+```
+
+- The sequence `(swap, swap, LS(TAB))` produces `(LA(TAB), LA(TAB), LA(LS(TAB)))`. The LS(TAB) behavior does not fire the interrupt behavior, because it is included in `shared-key-positions`.
+- The sequence `(swap, swap, B)` produces `(LA(TAB), LA(TAB), B)`. The B behavior **does** fire the interrupt behavior, because it is **not** included in `shared-key-positions`.
+
+#### `shared-layers`
+
+- Including `shared-layers` in your smart interrupt definition will let the specified layers NOT trigger the active smart interrupt behavior when they become active (include the layer the behavior is on to accommodate for layer toggling).
```
petejohanson, line 85: Can we document the fallback behavior if this property isn't set? Not clear from reading this if all layers will be ignored, or all layer changes will trigger an interrupt.

app/dts/bindings/behaviors/zmk,behavior-smart-interrupt.yaml
```diff
@@ -0,0 +1,24 @@
...
+    type: array
+    required: false
+    default: []
+  shared-layers:
```
petejohanson, line 18: Ditto on the naming... `continue-layers`/`ignored-layers`?

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#define ZMK_BHV_MAX_ACTIVE_SMART_INTERRUPTS 10
+
+struct behavior_smart_interrupt_config {
+    int32_t shared_key_positions_len;
+    int32_t shared_layers_len;
+    struct zmk_behavior_binding *behaviors;
+    int32_t shared_layers[32];
```
petejohanson, line 28: Why not use a bitmask for the enabled shared layers? Would save a non-trivial amount of storage here.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+    int32_t timeout_ms;
+    int32_t shared_key_positions[];
```
petejohanson, line 30: Ditto, could be a bitmask instead of this "sparse" array.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+#include <zmk/hid.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#define ZMK_BHV_MAX_ACTIVE_SMART_INTERRUPTS 10
+
+struct behavior_smart_interrupt_config {
+    int32_t shared_key_positions_len;
+    int32_t shared_layers_len;
+    struct zmk_behavior_binding *behaviors;
```
petejohanson, line 27: Since we require 3 behavior bindings exactly, I would break them out to explicit fields for each binding, e.g.:

```suggestion
    struct zmk_behavior_binding start_behavior;
    struct zmk_behavior_binding continue_behavior;
    struct zmk_behavior_binding end_behavior;
```

So you're not doing odd access by index work throughout the rest of the code.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+        CONTAINER_OF(item, struct active_smart_interrupt, release_timer);
+    if (!smart_interrupt->is_active) {
+        return;
+    }
+    if (smart_interrupt->timer_cancelled) {
+        return;
+    }
+    if (smart_interrupt->is_pressed) {
+        return;
+    }
```
petejohanson, line 74: This seems a bit *too* verbose, I would simplify:

```suggestion
    if (!smart_interrupt->is_active ||
        smart_interrupt->timer_cancelled ||
        smart_interrupt->is_pressed) {
        return;
    }
```

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+    LOG_DBG("Smart interrupted deactivated due to timer");
+    smart_interrupt->is_active = false;
+    struct zmk_behavior_binding_event event = {.position = smart_interrupt->position,
+                                               .timestamp = k_uptime_get()};
+    behavior_keymap_binding_pressed(&smart_interrupt->config->behaviors[2], event);
+    behavior_keymap_binding_released(&smart_interrupt->config->behaviors[2], event);
```
petejohanson, line 80: Do we need/care to have the time between press and release be configurable? Some hosts might not like the "immediate" press and release, for certain uses.

And if that needs to have an optional delay, probably best to use the behavior queue added for combos to queue up the press and release.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+    if (smart_interrupt == NULL) {
+        if (new_smart_interrupt(event.position, cfg, &smart_interrupt) == -ENOMEM) {
+            LOG_ERR("Unable to create new smart_interrupt. Insufficient space in "
+                    "active_smart_interrupts[].");
+            return ZMK_BEHAVIOR_OPAQUE;
+        }
+        LOG_DBG("%d created new smart_interrupt", event.position);
+    }
+    LOG_DBG("%d smart_interrupt pressed", event.position);
+    stop_timer(smart_interrupt);
```
petejohanson, line 150: Why are we stopping this timer in the press? Smells like missed clean up elsewhere if this is needed.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+    smart_interrupt->is_pressed = true;
+    if (smart_interrupt->first_press) {
+        behavior_keymap_binding_pressed(&cfg->behaviors[0], event);
+        behavior_keymap_binding_released(&cfg->behaviors[0], event);
```
petejohanson, line 154: Ditto about timing between press and release, and the behavior queue.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+static int on_smart_interrupt_binding_released(struct zmk_behavior_binding *binding,
+                                               struct zmk_behavior_binding_event event) {
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_smart_interrupt_config *cfg = dev->config;
+    LOG_DBG("%d smart_interrupt keybind released", event.position);
+    struct active_smart_interrupt *smart_interrupt = find_smart_interrupt(event.position);
+    if (smart_interrupt == NULL) {
+        return ZMK_BEHAVIOR_OPAQUE;
+    }
+    smart_interrupt->is_pressed = false;
```
petejohanson, line 170: Perhaps should be a separate free/release function that captures the intent of this work.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+};
+
+static int smart_interrupt_position_state_changed_listener(const zmk_event_t *eh);
+static int smart_interrupt_layer_state_changed_listener(const zmk_event_t *eh);
+
+ZMK_LISTENER(behavior_smart_interrupt, smart_interrupt_position_state_changed_listener);
+ZMK_SUBSCRIPTION(behavior_smart_interrupt, zmk_position_state_changed);
+
+ZMK_LISTENER(behavior_smart_interrupt2, smart_interrupt_layer_state_changed_listener);
+ZMK_SUBSCRIPTION(behavior_smart_interrupt2, zmk_layer_state_changed);
```
petejohanson, line 201: Why two different listeners? Can have one listener subscribed to multiple events.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+            LOG_DBG("Smart Interrupt interrupted, ending at %d %d", smart_interrupt->position,
+                    ev->position);
+            smart_interrupt->is_active = false;
+            struct zmk_behavior_binding_event event = {.position = smart_interrupt->position,
+                                                       .timestamp = k_uptime_get()};
+            if (smart_interrupt->is_pressed) {
+                behavior_keymap_binding_released(&smart_interrupt->config->behaviors[1], event);
+            }
+            behavior_keymap_binding_pressed(&smart_interrupt->config->behaviors[2], event);
+            behavior_keymap_binding_released(&smart_interrupt->config->behaviors[2], event);
```
petejohanson, line 226: Ditto on timing, and I would wrap all "trigger the interrupt behavior" into a shared function.

app/src/behaviors/behavior_smart_interrupt.c
```diff
@@ -0,0 +1,288 @@
...
+            LOG_DBG("Smart Interrupt layer changed, ending at %d %d", smart_interrupt->position,
+                    ev->layer);
+            smart_interrupt->is_active = false;
+            struct zmk_behavior_binding_event event = {.position = smart_interrupt->position,
+                                                       .timestamp = k_uptime_get()};
+            if (smart_interrupt->is_pressed) {
+                behavior_keymap_binding_released(&smart_interrupt->config->behaviors[1], event);
+            }
+            behavior_keymap_binding_pressed(&smart_interrupt->config->behaviors[2], event);
+            behavior_keymap_binding_released(&smart_interrupt->config->behaviors[2], event);
```
petejohanson, line 261: Here as well.

petejohanson (changes requested):
Thanks for this! Love the idea, and the general approach.

A few thoughts on implementation details/changes added here.

As far as naming.... Let me ponder. "Smart Interrupt" certainly feels better than "Swapper", but I'm curious if anyone has any other ideas for a better name. I also think "smart" is superfluous... A lot of what ZMK does is "smart", so it feels odd to include it in the naming for just this one behavior.

* "Three Phase Behavior"
* *Tri-Phase Behavior"
* "Start/Continue/End Behavior"
* "Tri-State Behavior"

Just a few that come to mind as food for thought.

app/src/behaviors/behavior_smart_interrupt.c
nickconway: From the discussion on discord this is a bit tricky since the keymap can be any length. I've changed it to 8-bit ints for now but will keep working on a better solution.

caksoylar:
About the "tri-state" naming: It might be confusable with the "tri-layer" (`lower+raise = adjust`) [pattern](https://docs.qmk.fm/#/ref_functions?id=olkb-tri-layers) that we support with [conditional layers](https://zmk.dev/docs/features/conditional-layers).

nickconway:
> About the "tri-state" naming: It might be confusable with the "tri-layer" (`lower+raise = adjust`) [pattern](https://docs.qmk.fm/#/ref_functions?id=olkb-tri-layers) that we support with [conditional layers](https://zmk.dev/docs/features/conditional-layers).

Maybe three-phase would be better, or would 'triplex' or 'treble' work here? Start/Continue/End Behavior gets the point across, but I feel like it's a bit too specific.

caksoylar:
I like the "bookend"/"sandwich" suggestions from @petejohanson personally, since it is like the repeatable "continue" behavior is bookended by special start and end events. Three-phase also sounds good.

app/src/behaviors/behavior_tri_state.c
```diff
@@ -0,0 +1,301 @@
...
+            return ZMK_BEHAVIOR_OPAQUE;
+        }
+        LOG_DBG("%d created new tri_state", event.position);
+    }
+    LOG_DBG("%d tri_state pressed", event.position);
+    tri_state->is_pressed = true;
+    if (tri_state->first_press) {
+        behavior_keymap_binding_pressed((struct zmk_behavior_binding *)&cfg->start_behavior, event);
+        behavior_keymap_binding_released((struct zmk_behavior_binding *)&cfg->start_behavior,
+                                         event);
```
petejohanson, line 147: I know we do this for hold taps right now, but any reason not to use the behavior queue here?

app/src/behaviors/behavior_tri_state.c
```diff
@@ -0,0 +1,301 @@
...
+        tri_state->first_press = false;
+    }
+    behavior_keymap_binding_pressed((struct zmk_behavior_binding *)&cfg->continue_behavior, event);
```
petejohanson, line 150: Ditto

app/src/behaviors/behavior_tri_state.c
```diff
@@ -0,0 +1,301 @@
...
+            struct zmk_behavior_binding_event event = {.position = tri_state->position,
+                                                       .timestamp = k_uptime_get()};
+            if (tri_state->is_pressed) {
+                behavior_keymap_binding_released(
+                    (struct zmk_behavior_binding *)&tri_state->config->continue_behavior, event);
+            }
+            behavior_keymap_binding_pressed(
+                (struct zmk_behavior_binding *)&tri_state->config->end_behavior, event);
+            behavior_keymap_binding_released(
+                (struct zmk_behavior_binding *)&tri_state->config->end_behavior, event);
```
petejohanson, line 268: Same.

petejohanson (commented):
An implementation comment... And then a bump again on naming... I'm on the fence on tri-state. @nickconway I didn't see you're thought on "bookend" concept for the name?

nickconway:
@petejohanson Apologies for the delay. I think the bookend name is much better. As for the behavior queue, my understanding was that the keymap_* functions would be used here since we want the action taken immediately?

petejohanson:
> @petejohanson Apologies for the delay. I think the bookend name is much better. As for the behavior queue, my understanding was that the keymap_* functions would be used here since we want the action taken immediately?

Until I figure out an effective way to push items at the *front* of the queue, we will need to use the keymap function approach.

urob:
I have been using this behavior for a while now and love the versatility of it. Two thoughts:

1/ I noted that when binding the behavior to a combo, then one must include the combo keys in `ignore-key-positions`. Otherwise, the exit condition will be triggered immediately. No big deal, but if there's an easy fix that might be cleaner.

2/ There is quite a bit of overlap between this and caps-word (and its generalization in #1451), the main difference being that `tri-state` is position-based whereas `caps-word` is keycode-based. Would it make sense to merge the two approaches? Say, by adding a `ignore-key-events` keyword to `tri-state`.  One could then re-implement `caps-word` with something like
```
bindings = <&kt LSHIFT>, <&none>, <&kt LSHIFT>;
ignore-key-events = <A B C D E F ...>
```
(For convenience, it could make sense to also have `ignore-alphas`, `ignore-numbers` and `ignore-mods` options that set `ignore-key-events` accordingly).

danielo515:
This is fantastic. In QMK I had a user space function for this same functionality, but it was just for alt + tab 
Currently in ZMK I'm doing it with a dedicated layer and two macros. Having this general solution will be Great

kaievns:
@danielo515 can you shed some light onto your magic alt-tab implementation? I looked into your config and I'm a bit lost :) would love just general explanation how you've managed to pull it off. Thanks!

danielo515:
> @danielo515 can you shed some light onto your magic alt-tab implementation? I looked into your config and I'm a bit lost :) would love just general explanation how you've managed to pull it off. Thanks!

It is very simple in reality. I have a key that moves to app switch layer, which has all keys to make me leave that layer, except the ones I want to use (left, right, and Q for quit). Here is the layer:
https://github.com/danielo515/glove80-zmk-config/blob/40d69811521b449946ec341afe3eadedeb2a846e/config/glove80.keymap#L460-L471

And the relevant macros are this ones: 
https://github.com/danielo515/glove80-zmk-config/blob/40d69811521b449946ec341afe3eadedeb2a846e/config/glove80.keymap#L156-L180

kaievns:
> It is very simple in reality. I have a key that moves to app switch layer, which has all keys to make me leave that layer, except the ones I want to use (left, right, and Q for quit).

Ah, so it's not really a one key swapper like QMK, it's just a new layer with different keys that you use to switch between the windows, right?

urob:
You can make it a one-key swapper if you want. Just don't bind the arrows. The docs included with this PR feature an example.

petejohanson:
> @petejohanson Apologies for the delay. I think the bookend name is much better. As for the behavior queue, my understanding was that the keymap_* functions would be used here since we want the action taken immediately?

I say we go w/ bookend term, barring any major disagreement. Can you update this with naming, docs, etc that reflect that?

burgessa23:
I'm trying to use `&swap` on an encoder and can't seem to get it to fire although it compiles fine... any ideas?

bryanforbes:
I have this running on my Glove80 and it's working pretty well so far, but I'm running into an issue. I have the following definition:

```
macswap: macswap {
    label = "macswap";
    compatible = "zmk,behavior-tri-state";
    #binding-cells = <0>;
    bindings = <&kt LGUI>, <&kp TAB>, <&kt LGUI>;
    ignored-key-positions = <38>; // shift
}
```

In my layout, I have a key with `&sk LGUI` and a key with `&macswap` next to each other. The problem I'm running into is if I accidentally hit `&sk LGUI` before `&macswap`, `LGUI` is permanently sticky and I have to turn the keyboard off in order to reset it.

app/src/behaviors/behavior_tri_state.c
```diff
@@ -0,0 +1,301 @@
...
+    }
+}
+
+void trigger_end_behavior(struct active_tri_state *si) {
+    zmk_behavior_queue_add(si->position, si->config->end_behavior, true, si->config->tap_ms);
+    zmk_behavior_queue_add(si->position, si->config->end_behavior, false, 0);
+}
+
+void behavior_tri_state_timer_handler(struct k_work *item) {
+    struct active_tri_state *tri_state = CONTAINER_OF(item, struct active_tri_state, release_timer);
```
lttb, line 73: from [Zephyr v3.5.0 Migration Guide](https://docs.zephyrproject.org/3.5.0/releases/migration-guide-3.5.html)

> [CONTAINER_OF](https://docs.zephyrproject.org/3.5.0/kernel/util/index.html#c.CONTAINER_OF) now performs type checking, this was very commonly misused to obtain user structure from [k_work](https://docs.zephyrproject.org/3.5.0/kernel/services/threads/workqueue.html#c.k_work) pointers without passing from [k_work_delayable](https://docs.zephyrproject.org/3.5.0/kernel/services/threads/workqueue.html#c.k_work_delayable). This would now result in a build error and have to be done correctly using [k_work_delayable_from_work()](https://docs.zephyrproject.org/3.5.0/kernel/services/threads/workqueue.html#c.k_work_delayable_from_work).

```suggestion
    struct k_work_delayable *ditem = k_work_delayable_from_work(item);
    struct active_tri_state *tri_state = CONTAINER_OF(ditem, struct active_tri_state, release_timer);
```

app/src/behaviors/behavior_tri_state.c
```diff
@@ -0,0 +1,301 @@
...
+                (struct zmk_behavior_binding *)&tri_state->config->end_behavior, event);
+            return ZMK_EV_EVENT_BUBBLE;
+        }
+    }
+    return ZMK_EV_EVENT_BUBBLE;
+}
+
+#define _TRANSFORM_ENTRY(idx, node)                                                                \
+    {                                                                                              \
+        .behavior_dev = DT_LABEL(DT_INST_PHANDLE_BY_IDX(node, bindings, idx)),                     \
```
lttb, line 277: - in #1499, `DT_LABEL` was migrated to `DT_PROP`
- and then in #2028, `DT_PROP` was migrated to `DEVICE_DT_NAME`

```suggestion
        .behavior_dev = DEVICE_DT_NAME(DT_INST_PHANDLE_BY_IDX(node, bindings, idx)), \
```

app/src/behaviors/behavior_tri_state.c
```diff
@@ -0,0 +1,301 @@
...
+        .ignored_key_positions = DT_INST_PROP(n, ignored_key_positions),                           \
+        .ignored_key_positions_len = DT_INST_PROP_LEN(n, ignored_key_positions),                   \
+        .ignored_layers = DT_INST_FOREACH_PROP_ELEM(n, ignored_layers, IF_BIT) 0,                  \
+        .ignored_layers_len = DT_INST_PROP_LEN(n, ignored_layers),                                 \
+        .timeout_ms = DT_INST_PROP(n, timeout_ms),                                                 \
+        .tap_ms = DT_INST_PROP(n, tap_ms),                                                         \
+        .start_behavior = _TRANSFORM_ENTRY(0, n),                                                  \
+        .continue_behavior = _TRANSFORM_ENTRY(1, n),                                               \
+        .end_behavior = _TRANSFORM_ENTRY(2, n)};                                                   \
+    DEVICE_DT_INST_DEFINE(n, behavior_tri_state_init, NULL, NULL, &behavior_tri_state_config_##n,  \
```
lttb, line 297: in #2028 `DEVICE_DT_INST_DEFINE` was migrated to `BEHAVIOR_DT_INST_DEFINE` in behaviours

```suggestion
    BEHAVIOR_DT_INST_DEFINE(n, behavior_tri_state_init, NULL, NULL, &behavior_tri_state_config_##n,  \
```

lttb (commented):
I've tried to compile `zmk` with the latest `main` and zephyr 3.5 with this behaviour, and had a few compilation errors.

for anyone interested, there are a few changes that made it work for me (you can find a complete patch in https://github.com/lttb/zmk/commit/2b856c59fefdcab5a208b20accf1291248ac2a3a)

davidsteinberger:
Thank you very much for all the effort. I gave it a try. Works when added directly to the keymap. When I add it via a combo though, it doesn't work. Does anybody have the same problem or a workaround?

nickconway:
Reopened in #3373

PR #1371, opened 2022-07-03
fix(docs): Add missing underglow on/off defines

caksoylar:
These defines were not documented, but they are now.


dxmh (approved):
Good spot, thanks @caksoylar 👍

PR #1379, opened 2022-07-06
Add short SHA to firmware artifact/file names

sethforprivacy:
This commit adds a short SHA tag to the end of artifact and firmware file names so that they are unique per commit/build.

This will output a zip titled like `firmware-015a40d.zip` and `sath65-nice_nano_v2-zmk-015a40d.uf2`, where `015a40d` is the 7-character short SHA ID of the head commit that kicked off the build.

Edit: The Action used to generate the SHA ID is https://github.com/marketplace/actions/short-sha, as using an action instead of a native `echo` command makes it portable across build OSs.

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -18,8 +18,8 @@ on:
         default: "bin"
         required: false
         type: string
-      artifact_name:
-        description: 'Artifact output file name'
+      archive_name:
+        description: 'Archive output file name'
         default: 'firmware'
         required: false
         type: string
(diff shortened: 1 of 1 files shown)


petejohanson:
@sethforprivacy I like this on principle, but the GH Action you used I don't really love. Are there really any advantages to using it over using `git rev-parse --short HEAD`?

sethforprivacy:
> @sethforprivacy I like this on principle, but the GH Action you used I don't really love. Are there really any advantages to using it over using `git rev-parse --short HEAD`?

I went back and forth on using a simple echo or this Action, but settled on the action to ensure it works no matter the build base OS you choose in case a user changes it or it gets changed upstream in the future for some reason.

I'm happy with either approach, just preferred to make it as "fool-proof" as possible by using this dedicated Action. If you prefer the other approach I'll push a commit to do it that way instead 😀

petejohanson:
Let's stick with the call to the git executable.

sethforprivacy:
> Let's stick with the call to the git executable.

Working on the best solution for that approach now, will ping when done and ready for review.

sethforprivacy:
@petejohanson pushed a squashed commit that works well for me, but it did require a simple re-ordering of the first steps of the workflow to make it work properly.

Tested and validated against my own private zmk-config repo without issues on this commit.

sethforprivacy:
Anything else needed from me here, @petejohanson? Should be all ready to merge!

PR #1380, opened 2022-07-09
feat(behaviors): Leader key

nickconway:
Adds leader key functionality


docs/docs/behaviors/leader-key.md
```diff
@@ -0,0 +1,42 @@
...
+/ {
+    keymap {
+        ...
+    };
+};
+```
+
+#### `timerless`
+
+By default, the leader key will have a timeout, and will not wait for a sequence to be completed or another key to be pressed. Specify `timerless` if you want a timeout.
```
AlexCzar, line 42: I struggle to understand this part

* Where and how `timerless` should be specified? Is it a possible value for `timeout-ms` or is it a separate option alongside it?
* Why is the opition called "timerless"? The statement says that we should specify it if we want a timeout, it might be a typo and you meant to write "don't want a timeout"? In the latter case, the naming is still a bit non-intuitive, maybe it should be replaced by `no-timeout = <true>`, or even removed in favour of `timeout-ms = <0>`; or `<timeout-ms = <-1>`, or, if timerless *is* a possible value for `timeout-ms`, maybe a shorter `= <false>` or even `= <no>` could be a better option? (shooting up the sky here).

docs/docs/features/leader.md
```diff
@@ -0,0 +1,39 @@
...
+		};
+	};
+};
+```
+
+- The `compatible` property should always be `"zmk,leader-sequences"` for leader sequences.
+- `key-positions` is an array of key positions. See the info section below about how to figure out the positions on your board.
+- `layers = <0 1...>` will allow limiting a sequence to specific layers. This is an _optional_ parameter, when omitted it defaults to global scope.
+- `bindings` is the behavior that is activated when all the key positions are pressed.
+- (advanced) you can specify `immediate-trigger` if you want the sequence to be triggered as soon as all key positions are pressed. The default is to wait for the leader key's timeout to trigger the sequence if it overlaps another.
```
AlexCzar, line 29: Where and how?

AlexCzar (commented):
Thank you so much, leader key is a very nice feature to have. Just a couple of questions/suggestions regarding the documentation.
Can't wait for this to be merged 👍🏽

app/dts/behaviors/leader_key.dtsi
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+ / {
+	behaviors {
+		/omit-if-no-ref/ leader: leader_key {
```
petejohanson, line 9: It seems to me we shouldn't limit this to just one global `leader` key, as this limits folks to just one leader key, and it's associated sequences.

Is there really no use case for different leader keys, with different possible sequences?

Seems like you could have child props in the yaml binding, and something like:

```
leader_foo: leader_foo {
  compatible = "zmk,behavior-leader";
  ...
  seq1 {
  };
  seq2 {
  };
};
````

That would give us a bit more flexibility.

If we really want to make it easy, we could still define one global leader you could enhance with children, e.g.:

```
&leader {
  seq1 {
  };
};
```

File: app/dts/bindings/behaviors/zmk,behavior-leader-key.yaml [hunk 4]
```diff
@@ -0,0 +1,15 @@
...
+description: Leader key behavior
+
+compatible: "zmk,behavior-leader-key"
+
+include: zero_param.yaml
+
+properties:
+  timerless:
+    type: boolean
+  timeout-ms:
```
petejohanson, line 13: This needs documenting as well. Also see the other reviewers question.

petejohanson (commented):
One important high level question on the design.

Thanks!

app/dts/behaviors/leader_key.dtsi
nickconway: Right now I use multiple leader keys with layers (see [here](https://github.com/nickconway/zmk-config/blob/main/config/include/leader.dtsi)). I use macros to switch layers before pressing the leader key which has the same effect although is a bit more work for the user.

docs/docs/behaviors/leader-key.md
nickconway: - Timerless would be a separate option, defined in the dts for the leader key along with timeout-ms and the other options.
- Yeah, that was a typo, should be "don't want a timeout". Timeout-ms would need to be a positive number for the case of overlaps, unless we had something like overlap-timeout-ms and timeout-ms separate? Timerless is for when no sequence has been completed yet, and once an overlapped sequence is completed, timeout-ms is used.

grezp:
Any update on this branch merging to main?

vladsolokha:
I don't understand the difference of this "leader key" and a sticky layer. Can someone differentiate?

cougarten:
> I don't understand the difference of this "leader key" and a sticky layer. Can someone differentiate?

you can do more than 2 key presses and it can chain arbitrary long sequences as the triggers.

e.g. "super > s > e"  for "leader key" > "shortcuts" > "email" to trigger spelling your email address (or anything you want).

ujl123:
> I don't understand the difference of this "leader key" and a sticky layer. Can someone differentiate?

Other uses I use leader keys in QMK is (all are for windows)

Leader Key + B O O T = qmk reset
Leader Key + R G B = toggle rgb
Leader Key + Q = Alt F4
Leader key + Z Z Z = system shutdown
Leader key + T O P = CTRL + SHIFT + ESC (task manager in windows)
Leader Key + MKDIR = SHIFT + DEL (hard delete in Windows)
Leader KEy + PIN = windows pin (not recommended storing passwords on your keyboard but I'm lazy).


I'm hoping that this gets reviewed and can be merged into the main branch though as this is a major component that makes my 32 key keyboard work in QMK (along with tap mods, tap dances autoshifts, custom autoshift keys)

app/src/leader.c
```diff
@@ -0,0 +1,362 @@
...
+        .immediate_trigger = DT_PROP(n, immediate_trigger),                                        \
+        .is_pressed = false,                                                                       \
+        .key_positions = DT_PROP(n, key_positions),                                                \
+        .key_position_len = DT_PROP_LEN(n, key_positions),                                         \
+        .behavior = ZMK_KEYMAP_EXTRACT_BINDING(0, n),                                              \
+        .layers = DT_PROP(n, layers),                                                              \
+        .layers_len = DT_PROP_LEN(n, layers),                                                      \
+    };
+
+DT_INST_FOREACH_CHILD(0, LEADER_INST)
```
petejohanson, line 352: If we really are only supporting one instance of this node, then the FOREACH mechanics are superfluous and misleading. Just create one static struct instance directly.

app/src/leader.c
```diff
@@ -0,0 +1,362 @@
...
+
+#define INTITIALIAZE_LEADER_SEQUENCES(n) intitialiaze_leader_sequences(&sequence_config_##n);
+
+static int leader_init() {
+    k_work_init_delayable(&release_timer, behavior_leader_key_timer_handler);
+    DT_INST_FOREACH_CHILD(0, INTITIALIAZE_LEADER_SEQUENCES);
```
petejohanson, line 358: Ditto, this adds unnecessary complexity since only a single instance of the node is supported.

app/src/leader.c
```diff
@@ -0,0 +1,362 @@
...
+int32_t release_count;
+int32_t timeout_ms;
+int32_t active_leader_position;
+int8_t layer;
+bool first_release;
+struct k_work_delayable release_timer;
+int64_t release_at;
+bool timer_started;
+bool timer_cancelled;
+bool timerless;
```
petejohanson, line 36: These all need to be `static`.

app/include/zmk/leader.h
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#pragma once
+
+void zmk_leader_activate(int32_t timeout, bool timeout_on_activation, uint32_t position);
+void zmk_leader_deactivate();
+bool zmk_leader_get_status();
```
petejohanson, line 11: `status` I don't love... is active? is enabled? Something a bit more aligned with the bool return value here.

app/src/behaviors/behavior_leader_key.c
```diff
@@ -0,0 +1,54 @@
...
+};
+
+static int behavior_leader_key_init(const struct device *dev) { return 0; }
+
+static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
+                                     struct zmk_behavior_binding_event event) {
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_leader_key_config *cfg = dev->config;
+
+    zmk_leader_activate(cfg->timeout_ms, cfg->timerless, event.position);
```
petejohanson, line 33: If you press this again while it's active, should the behavior be to deactivate to interrupt it?

app/src/leader.c
```diff
@@ -0,0 +1,362 @@
...
+
+    return 0;
+}
+
+ZMK_LISTENER(leader, position_state_changed_listener);
+ZMK_SUBSCRIPTION(leader, zmk_position_state_changed);
+
+#define LEADER_INST(n)                                                                             \
+    static struct leader_seq_cfg sequence_config_##n = {                                           \
+        .virtual_key_position = ZMK_KEYMAP_LEN + __COUNTER__,                                      \
```
petejohanson, line 342: This should be refactored to use the virtual key position work @joelspadin added for combos, etc.

app/src/leader.c
```diff
@@ -0,0 +1,362 @@
...
+
+uint32_t current_sequence[CONFIG_ZMK_LEADER_MAX_KEYS_PER_SEQUENCE] = {-1};
+// the set of candidate leader based on the currently leader_pressed_keys
+int num_candidates;
+struct leader_seq_cfg *sequence_candidates[CONFIG_ZMK_LEADER_MAX_SEQUENCES_PER_KEY];
+int num_comp_candidates;
+struct leader_seq_cfg *completed_sequence_candidates[CONFIG_ZMK_LEADER_MAX_SEQUENCES_PER_KEY];
+// a lookup dict that maps a key position to all sequences on that position
+struct leader_seq_cfg *sequence_lookup[ZMK_KEYMAP_LEN][CONFIG_ZMK_LEADER_MAX_SEQUENCES_PER_KEY] = {
+    NULL};
```
petejohanson, line 66: All should also be `static`

petejohanson (changes requested):
A few more items from my first pass at looking at this again. Thanks again for working on this!

arabshapt:
Please try to finish this feature... I will soon try to merge it on my branch, but it would be great to have it on master. I do not understand how people are not showing more interest for it. If it is easy to add key sequences(similar to vimrc) then it would be even better, but I think that can be done with a custom function... The main thing is having it on master and bugfree:) Thank you all but especially @nickconway and @petejohanson for working on it🙂

arabshapt:
Is there any limitation regarding number of sequences that can be created? For combos there are (I think) 5 combos that can use the same key position. Could one create e.g. 900 sequences: leader + letter_1 + letter_2 = shortcut/macro?
So three key taps, around 900 unique sequences. Would it be possible with this implementation?

tvdab:
I was a happy user of theleader key feature in my previous zmk build. Recently I tried updating my firmware to the latest release of zmk and also update from zephyr sdk 0.15 to 0.16. Everything seems to work, except for the leader key. The leader key acts like a transparent key. Also running the test tests/leader/basic fails:
```
❯ west test tests/leader/basic
Running tests/leader/basic:
--- tests/leader/basic/keycode_events.snapshot	2024-02-25 22:11:10.516535082 +0100
+++ build/tests/leader/basic/keycode_events.log	2024-02-25 22:35:07.778445825 +0100
@@ -1,4 +1,2 @@
-leader: leader key activated
-pressed: usage_page 0x07 keycode 0x04 implicit_mods 0x00 explicit_mods 0x00
-released: usage_page 0x07 keycode 0x04 implicit_mods 0x00 explicit_mods 0x00
-leader: leader key deactivated
+pressed: usage_page 0x07 keycode 0x1E implicit_mods 0x00 explicit_mods 0x00
+released: usage_page 0x07 keycode 0x1E implicit_mods 0x00 explicit_mods 0x00
FAILED: tests/leader/basic
```
Could someone verify if the leader key works with zephyr sdk 0.16? Maybe I'm doing something wrong...

caksoylar:
I am guessing this needs to be rebased for #2028.

tvdab:
I found why it wasn't working in Zephyr 3.5. This is what needs to be changed in order to make it work again:

```
--- a/app/src/behaviors/behavior_leader_key.c
+++ b/app/src/behaviors/behavior_leader_key.c
@@ -47,8 +47,8 @@ static const struct behavior_driver_api behavior_leader_key_driver_api = {
 #define LEAD_INST(n)                                                                               \
     static struct behavior_leader_key_config behavior_leader_key_config_##n = {                    \
         .timerless = DT_INST_PROP(n, timerless), .timeout_ms = DT_INST_PROP(n, timeout_ms)};       \
-    DEVICE_DT_INST_DEFINE(n, behavior_leader_key_init, NULL, NULL,                                 \
-                          &behavior_leader_key_config_##n, APPLICATION,                            \
+    BEHAVIOR_DT_INST_DEFINE(n, behavior_leader_key_init, NULL, NULL,                                 \
+                          &behavior_leader_key_config_##n, POST_KERNEL,                            \
                           CONFIG_KERNEL_INIT_PRIORITY_DEFAULT, &behavior_leader_key_driver_api);

 DT_INST_FOREACH_STATUS_OKAY(LEAD_INST) 
``` 
There are probably more changes to be made (#2028) to update everything to zephyr 3.5.

magoz:
Looking forward to this, great work!

urob:
If anyone is interested using this as a module that works with the current upstream, I just pushed one to https://github.com/urob/zmk-leader-key. For now its a one-for-one adaption of Nick's PR but I am planning some changes going forward.

app/dts/behaviors/leader_key.dtsi
nickconway: Merged in support for multiple instances with child nodes!

nickconway:
I've merged in a bunch of changes just now, support for multiple instances, using child nodes, etc. Using the child nodes depends on a fix that hasn't yet been merged in yet (https://github.com/zephyrproject-rtos/zephyr/pull/62417) but if I use a fork with the fix everything works as expected! (thanks so much @urob for maintaining your module all this time and finding that fix)

docs/docs/behaviors/leader-key.md
nickconway: I've changed this so the default is no timeout since we can't have timeout_ms default to a negative value.

docs/docs/features/leader.md
nickconway: I've added an example in the documentation that should be much clearer on where to implement these options.

jamesgour:
@nickconway what is the latest on this? :)

PR #1278, opened 2022-04-29
feat(behavior): key toggle behavior

cgoates:
Rebase of #731 on latest main.  I also attempted to address the reviewer comments in that previous PR.

Still needs:
- [x] Unit tests
- [x] Documentation
- [x] Test on actual hardware
- [x] Update behavior to compile only if explicitly used in a keymap, as in #1269

PR #731, opened 2021-03-16
Toggle key feature

okke-formsma:
Requested by @mcrosson for use with artsey.

Most of the heavy lifting is done is hid.c, because that's where the current state of the keyboard can be found.

Keeping track of the hid state in the toggle would is tricky because that would mean reimplementing most of hid.c. Also it's not possible to cleanly keep track of when hid_clear methods are called, so just requesting this information from hid seemed the cleanest solution.

* a key is toggled on when the keycode not pressed
* a key is toggled off when the keycode is currently pressed
* something like `LS(A)` will be toggled on when `A` is pressed (the state of added modifiers is ignored)

This PR needs a bit of rework if https://github.com/zmkfirmware/zmk/pull/726 is merged earlier.

PR #1384, opened 2022-07-14
feat(docs): Added behavior queue limit section to macros page

GreenAirplane:
Added information about the `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE` config option to the macros page in the docs.

### DIFF
--- a/docs/docs/behaviors/macros.md
+++ b/docs/docs/behaviors/macros.md
@@ -138,7 +138,7 @@ bindings
 
 Macros use an internal queue to invoke each behavior in the bindings list when triggered, which has a size of 64 by default. Bindings in "press" and "release" modes correspond to one event in the queue, whereas "tap" mode bindings correspond to two (one for press and one for release). As a result, the effective number of actions processed might be less than 64 and this can cause problems for long macros.
 
-You can change the limit by adding `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=n` to your `.conf` file. For example, you can add `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=512` if your macro outputs more than 200 characters.
+To prevent issues with longer macros, you can change the size of this queue via the `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE` setting in your configuration, [typically through your `.conf` file](../config/index.md). For example, `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=512` would allow your macro to type about 256 characters.
 
 ## Common Patterns
 
--- a/docs/docs/behaviors/macros.md
+++ b/docs/docs/behaviors/macros.md
@@ -136,7 +136,7 @@ bindings
 
 ### Behavior Queue Limit
 
-The default queue limit is 64. However, some actions require multiple events (e.g. `&kp` requires `key press` and `key release` events). As a result, the effective number of actions can be less than 64.
+Macros use an internal queue to invoke each behavior in the bindings list when triggered, which has a size of 64 by default. Bindings in "press" and "release" modes correspond to one event in the queue, whereas "tap" mode bindings correspond to two (one for press and one for release). As a result, the effective number of actions processed might be less than 64 and this can cause problems for long macros.
 
 You can change the limit by adding `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=n` to your `.conf` file. For example, you can add `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=512` if your macro outputs more than 200 characters.
 
--- a/docs/docs/behaviors/macros.md
+++ b/docs/docs/behaviors/macros.md
@@ -134,6 +134,12 @@ bindings
     ;
 ```
 
+### Behavior Queue Limit
+
+The default queue limit is 64. However, some actions require multiple events (e.g. `&kp` requires `key press` and `key release` events). As a result, the effective number of actions can be less than 64.
+
+You can change the limit by adding `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=n` to your `.conf` file. For example, you can add `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=512` if your macro outputs more than 200 characters.
+
 ## Common Patterns
 
 Below are some examples of how the macro behavior can be used for various useful functionality.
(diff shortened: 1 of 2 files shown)


dxmh:
Thanks! 

Do you think you could also include this in the recently added [Configuration](https://zmk.dev/docs/config) section too?

GreenAirplane:
So, I've added some information about the `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE` to the [Configuration > Behaviors](https://deploy-preview-1384--zmk.netlify.app/docs/config/behaviors#kconfig) page. I thought about linking [app/Kconfig#L334](https://github.com/zmkfirmware/zmk/blob/b1ce8a0d3346be78301637a9861e4b70587aa2bb/app/Kconfig#L334) but I'm not sure that it is a good idea.

Hopefully, it is the right place for this information.

Also, I don't think that it is worth adding information about `ZMK_BEHAVIOR_KEY_TOGGLE` because based on my understanding, it shouldn't be changed by end users. However, I might be wrong as I'm very new to zmk and just quickly checked PRs related to it: https://github.com/zmkfirmware/zmk/pull/731 and https://github.com/zmkfirmware/zmk/pull/1278

docs/docs/config/behaviors.md
```diff
@@ -9,6 +9,12 @@ See [Configuration Overview](index.md) for instructions on how to change these s
 
 See the [zmk/app/dts/behaviors/](https://github.com/zmkfirmware/zmk/tree/main/app/dts/behaviors) folder for all default behaviors.
 
+## Kconfig
+
+| Config                             | Type | Description                                          | Default |
+| ---------------------------------- | ---- | ---------------------------------------------------- | ------- |
+| `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE`  | int  | The size of the behaviors queue                      | 64      |
```
caksoylar, line 16:
```suggestion
| `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE`  | int  | Maximum number of behaviors to allow queueing from a macro or other complex behavior | 64      |
```

The descriptions in these docs match the description in the Kconfig file. Note that you need to run `prettier` again to format the table properly after this change.

GreenAirplane: done!

docs/docs/behaviors/macros.md
```diff
@@ -134,6 +134,12 @@ bindings
     ;
 ```
 
+### Behavior Queue Limit
+
+The default queue limit is 64. However, some actions require multiple events (e.g. `&kp` requires `key press` and `key release` events). As a result, the effective number of actions can be less than 64.
```
caksoylar, line 139:
```suggestion
Macros use an internal queue to invoke each behavior in the bindings list when triggered, which has a size of 64 by default. Bindings in "press" and "release" modes correspond to one event in the queue, whereas "tap" mode bindings correspond to two (one for press and one for release). As a result, the effective number of actions processed might be less than 64 and this can cause problems for long macros.
```

I think it is good to be precise here if we can, but this might require a review from the contributors to ensure technical correctness.

docs/docs/behaviors/macros.md
```diff
@@ -134,6 +134,12 @@ bindings
...
+
+You can change the limit by adding `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=n` to your `.conf` file. For example, you can add `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=512` if your macro outputs more than 200 characters.
```
caksoylar, line 141:
```suggestion
To prevent issues with longer macros, you can change the size of this queue via the `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE` setting in your configuration, [typically through your `.conf` file](../config/index.md). For example, `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE=512` would allow your macro to type about 256 characters.
```

I added a link to config documentation to keep this generic, but we need to double check the link works correctly.

caksoylar:
> Also, I don't think that it is worth adding information about ZMK_BEHAVIOR_KEY_TOGGLE because based on my understanding, it shouldn't be changed by end users. However, I might be wrong as I'm very new to zmk and just quickly checked PRs related to it: https://github.com/zmkfirmware/zmk/pull/731 and https://github.com/zmkfirmware/zmk/pull/1278

I believe that kind of Kconfig is indeed used only internally so that the feature is automatically enabled when used. We shouldn't be documenting them in the docs.

docs/docs/config/behaviors.md
```diff
@@ -9,6 +9,12 @@ See [Configuration Overview](index.md) for instructions on how to change these s
 
 See the [zmk/app/dts/behaviors/](https://github.com/zmkfirmware/zmk/tree/main/app/dts/behaviors) folder for all default behaviors.
 
+## Kconfig
```
dxmh, line 12: It's tricky finding somewhere for `CONFIG_ZMK_BEHAVIORS_QUEUE_SIZE` to live because it doesn't actually apply to a single behavior specifically. I do think that here at the top of the Behaviors page is most appropriate place for it.

However, I think it might make sense to have it nested beneath a new heading like so:

```suggestion
## General

These are general settings that control behaviors at a higher level.

### Kconfig
```

This would bring it in alignment with other pages, which are typically structured like this (e.g. https://zmk.dev/docs/config/kscan):

> * Group 1
>   * Kconfig
>   * Devicetree
> * Group 2
>   * Kconfig
>   * Devicetree

This would make the hierarchy of content for this page like so:

> - General
> 	- Kconfig
> - Caps Word
> 	- Devicetree
> - Hold-Tap
> 	- Devicetree
> - Key Repeat
> 	- Devicetree
> - Macro
> 	- Devicetree
> - Mod-Morph
> 	- Devicetree
> - Sticky Key
> 	- Devicetree
> - Tap Dance
> 	- Devicetree

What do you think @GreenAirplane, @caksoylar, and @joelspadin?

dxmh: Actually, to align it with the kscan page I linked to as an example ([source](https://github.com/zmkfirmware/zmk/blob/f05a7a13f2e73506eefb2fae06da4d95dbdd1ec0/docs/docs/config/kscan.md)), it could just simply be:

```suggestion
## Common

### Kconfig
```

caksoylar: Good suggestion! "Common" is good I'd say.

GreenAirplane: Yeah, I think `Common` is the best option, especially considering that it is already used in https://zmk.dev/docs/config/kscan.

Committed the suggestion.

dxmh (approved):
Thanks @GreenAirplane!

PR #1385, opened 2022-07-15
feat(kscan): Kconfig for optional scan delay.

petejohanson:
Add optional Kconfig setting to delay scanning after each
output column is set, and inputs are read, to allow inputs
to "settle" after the last column is set back to inactive.

This was found during work on the RP2040 hardware, but this was also encountered by jmnw on Discord for an stm32 build recently.

Without a delay, on some platforms we need to have a small delay after disabling the previous output column to prevent it from causing a "false positive" detected on an input for the next column.


jimmerricks:
I can confirm that this fixes a problem that I was having with an STM32F411 board. I have not tested it enough to know that it doesn't introduce new problems.

petejohanson:
I've been testing this as well on my RP2040 work for quite a while, with no ill effects. Merging!

PR #1387, opened 2022-07-17
feat(behaviors): Adding require-prior-idle-ms for combos and hold-taps

andrewjrae:
Detaching the global-quick-tap functionality from quick-tap term, and implementing it for both combos and hold-taps under the new option `require-prior-idle-ms`. 

~~This constitutes a breaking change for users keymaps in that `global-quick-tap` will no longer be an option. I'm not sure what the protocol is when doing such refactors, but I think it is beneficial in the long run for two reasons:~~ This deprecates the `global-quick-tap` option, however it is backwards compatible in that if it is set it will copy the `quick-tap-ms` to the `require-prior-idle-ms`.

There are two reasons for these changes;

1. This functionality can be added to combos under the new unified name `require-prior-idle-ms`.

2. This allows users to set a lower term for the 'global-quick-tap' (now `require-prior-idle-ms`) functionality (typically ~100ms), and a higher term for the regular quick-tap (typically ~200ms). This means the regular quick-tap is easier to hit, while still allowing the benefits of disabling the hold-taps during fast typing.

PR Progress:

- [x] Refactor hold-taps to support `require-prior-idle-ms` rather than `global-quick-tap`
- [x] Add `require-prior-idle-ms` functionality to combos
- [x] Update docs


andrewjrae:
I've now pushed a change such that he old `global-quick-tap` option is still usable, and will simply copy the value of `quick-tap-ms` to `min-prior-ms`, as per @okke-formsma's suggestion on Discord. A better name for `min-prior-ms` has yet to be chosen however.

caksoylar:
Wanted to drop a note saying this has been working perfectly for me for both combos and hold-taps for the last couple of months of testing. If the idea makes sense to contributors, maybe we can add docs?

andrewjrae:
Yah I'm sorry I've let this go stale, I moved and started a new job hehe. If someone wants to write docs they can feel free to PR my branch, otherwise I should be able to get to it next week.

app/tests/hold-tap/tap-unless-interrupted/6-global-quick-tap/1-basic/native_posix_64.keymap
```diff
@@ -6,17 +6,17 @@
...
+		/* quick tap */
 		ZMK_MOCK_PRESS(0,0,400)
 		ZMK_MOCK_RELEASE(0,0,400)
 		/* hold */
 		ZMK_MOCK_PRESS(0,0,400)
 		ZMK_MOCK_PRESS(1,0,10)
 		ZMK_MOCK_RELEASE(1,0,10)
 		ZMK_MOCK_RELEASE(0,0,400)
-		/* global quick tap */
+		/* min prior term */
```
okke-formsma, line 19: I don't understand this comment - maybe keep the old comment?

andrewjrae: Oops, that's a remnant from when I had renamed `global_quick_tap` to `min_prior_ms`, good catch! I'll clean that up when I write the docs.

urob:
> Wanted to drop a note saying this has been working perfectly for me for both combos and hold-taps for the last couple of months of testing. If the idea makes sense to contributors, maybe we can add docs?

Same here, would be great to see this merged.

infused-kim:
Amazing addition! I have been using it for the past week and it made homerow mods work almost flawlessly. 

But I did notice one area where it could be improved:

Most of the time I don't want to use the mods during fast typing, but sometimes when I have typo, I very quickly hit `GUI+BACKSPACE` to delete the entire misspelled work.

But since I oftentimes hit that shortcut within the `global-quick-tap-ms`, the shortcut fails.

What do you think of adding something like `global-quick-tap-excludes = <BACKSPACE>` that allows you to ignore certain keys for the global-quick-tap?

urob:
> Amazing addition! I have been using it for the past week and it made homerow mods work almost flawlessly. 
> 
> 
> 
> But I did notice one area where it could be improved:
> 
> 
> 
> Most of the time I don't want to use the mods during fast typing, but sometimes when I have typo, I very quickly hit `GUI+BACKSPACE` to delete the entire misspelled work.
> 
> 
> 
> But since I oftentimes hit that shortcut within the `global-quick-tap-ms`, the shortcut fails.
> 
> 
> 
> What do you think of adding something like `global-quick-tap-excludes = <BACKSPACE>` that allows you to ignore certain keys for the global-quick-tap?

I'm not sure this would work as expected. The global-quick-tap triggers the HRM as tap based on the time passed since the previous key. So at the time you press backspace, the Gui-HRM is already resolved as tap (presumably a alpha).

infused-kim:
> > Amazing addition! I have been using it for the past week and it made homerow mods work almost flawlessly.
> > But I did notice one area where it could be improved:
> > Most of the time I don't want to use the mods during fast typing, but sometimes when I have typo, I very quickly hit `GUI+BACKSPACE` to delete the entire misspelled work.
> > But since I oftentimes hit that shortcut within the `global-quick-tap-ms`, the shortcut fails.
> > What do you think of adding something like `global-quick-tap-excludes = <BACKSPACE>` that allows you to ignore certain keys for the global-quick-tap?
> 
> I'm not sure this would work as expected. The global-quick-tap triggers the HRM as tap based on the time passed since the previous key. So at the time you press backspace, the Gui-HRM is already resolved as tap (presumably a alpha).

Hmm, thank you. That’s too bad :(

danielo515:
> Most of the time I don't want to use the mods during fast typing, but sometimes when I have typo, I very quickly hit `GUI+BACKSPACE` to delete the entire misspelled work.

I have a very similar problem, I have several home row modifiers that I don't mind having to hold longer for them to become active, I even do it unconscious. However, I have a bunch of other modifiers that let me access my symbols layer, and those I want to be faster to react. I realised that I have a different behavior defined for those, so probably the timing can be adjusted on so they fire quicker. Not the solution you wanted, but maybe good enough?

andrewjrae:
Okay, I've finally gotten around to rebasing and writing/editing the docs! @petejohanson this should be ready for review, @okke-formsma already reviewed the code changes a while back, and many folks have been running this without issue for some time now.

docs/docs/behaviors/hold-tap.md
```diff
@@ -49,11 +49,11 @@ Defines how long a key must be pressed to trigger Hold behavior.
 
 If you press a tapped hold-tap again within `quick-tap-ms` milliseconds, it will always trigger the tap behavior. This is useful for things like a backspace, where a quick tap+hold holds backspace pressed. Set this to a negative value to disable. The default is -1 (disabled).
 
-#### `global-quick-tap`
+#### `global-quick-tap-ms`
 
-If `global-quick-tap` is enabled, then `quick-tap-ms` will apply not only when the given hold-tap is tapped, but for any key tapped before it. This effectively disables the hold-tap when typing quickly, which can be quite useful for homerow mods. It can also have the effect of removing the input delay when typing quickly.
+If `global-quick-tap-ms` is like `quick-tap-ms` however it will apply for _any_ key tapped before it. This effectively disables the hold-tap when typing quickly, which can be quite useful for homerow mods. It can also have the effect of removing the input delay when typing quickly.
```
caksoylar, line 54:
```suggestion
`global-quick-tap-ms` is like `quick-tap-ms` however it will apply for _any_ non-modifier key tapped before it. This effectively disables the hold-tap when typing quickly, which can be quite useful for homerow mods. It can also have the effect of removing the input delay when typing quickly.
```

caksoylar: My understanding from the code is that this only applies to non-modifier keys, if that is correct then it would be good to be precise here (and in other places).

docs/docs/behaviors/hold-tap.md
```diff
@@ -49,11 +49,11 @@ Defines how long a key must be pressed to trigger Hold behavior.
 
 If you press a tapped hold-tap again within `quick-tap-ms` milliseconds, it will always trigger the tap behavior. This is useful for things like a backspace, where a quick tap+hold holds backspace pressed. Set this to a negative value to disable. The default is -1 (disabled).
```
caksoylar, line 50:
```suggestion
If you press a tapped hold-tap again within `quick-tap-ms` milliseconds of the first press, it will always trigger the tap behavior. This is useful for things like a backspace, where a quick tap+hold holds backspace pressed. Set this to a negative value to disable. The default is -1 (disabled).
```

This also is my understanding from the code, if correct it would be good to be more precise. If not, please let me know.

docs/docs/config/combos.md
```diff
@@ -31,12 +31,13 @@ The `zmk,combos` node itself has no properties. It should have one child node pe
...
+| Property              | Type          | Description                                                                                                               | Default |
+| --------------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `bindings`            | phandle-array | A [behavior](../features/keymaps.md#behaviors) to run when the combo is triggered                                         |         |
+| `key-positions`       | array         | A list of key position indices for the keys which should trigger the combo                                                |         |
+| `timeout-ms`          | int           | All the keys in `key-positions` must be pressed within this time in milliseconds to trigger the combo                     | 50      |
+| `global-quick-tap-ms` | int           | If any key is tapped within `global-quick-tap-ms` before a key in the combo, the key will not be considered for the combo | -1      |
```
caksoylar, line 39:
```suggestion
| `global-quick-tap-ms` | int           | If any non-modifier key is pressed within `global-quick-tap-ms` before a key in the combo, the key will not be considered for the combo | -1      |
```
Related to the clarifications above.

docs/docs/features/combos.md
```diff
@@ -30,6 +30,7 @@ Combos configured in your `.keymap` file, but are separate from the `keymap` nod
 - `layers = <0 1...>` will allow limiting a combo to specific layers. This is an _optional_ parameter, when omitted it defaults to global scope.
 - `bindings` is the behavior that is activated when the behavior is pressed.
 - (advanced) you can specify `slow-release` if you want the combo binding to be released when all key-positions are released. The default is to release the combo as soon as any of the keys in the combo is released.
+- (advanced) you can specify `global-quick-tap-ms` much like in [hold-taps](behaviors/hold-tap.md#global-quick-tap-ms). If any key is tapped within `global-quick-tap-ms` before a key in the combo, the key will not be considered for the combo.
```
caksoylar, line 33:
```suggestion
- (advanced) you can specify a `global-quick-tap-ms` value much like for [hold-taps](behaviors/hold-tap.md#global-quick-tap-ms). If any non-modifier key is pressed within `global-quick-tap-ms` before a key in the combo, the combo will not trigger.
```

caksoylar (commented):
Looks good to me in general, thanks for writing these up!

In addition to the comments I dropped, the devicetree table in `docs/config/behaviors#hold-tap` referring to `global-quick-tap` also needs to be changed.

filterpaper:
Have you thought of using a different name or nomenclature like "instant tap" instead of "quick tap" to distinguish its functionality from the latter?

andrewjrae:
Thanks for the feedback @caksoylar! I'll apply your suggested changes, one quick question though regarding the config table. `global-quick-tap` as an option is still supported for backwards compatibility, it will just use the `quick-tap-ms` for the actual `global-quick-tap-ms` field. Do we want to completely remove `global-quick-tap` from the table, or should we leave it at the bottom and mark at as deprecated somehow?

andrewjrae:
@filterpaper, I did originally hope to use `min-prior-ms` to detach the concept from quick tap since it made a lot less sense for combos. However there was some discussion on this at the time and `global-quick-tap-ms` was chosen to reduce confusion. 

Personally I think having using `instant-tap` would also cause confusion, but I agree that separating the two concepts would be nice. I'm open to other names if folks can come up with something the community likes more (I'm not a very good namer :wink:).

filterpaper:
> @filterpaper, I did originally hope to use `min-prior-ms` to detach the concept from quick tap since it made a lot less sense for combos. However there was some discussion on this at the time and `global-quick-tap-ms` was chosen to reduce confusion.
> 
> Personally I think having using `instant-tap` would also cause confusion, but I agree that separating the two concepts would be nice. I'm open to other names if folks can come up with something the community likes more (I'm not a very good namer 😉).

This PR is splitting both into two distinct timers and is a great opportunity to rename the "global" one into something unique. "Instantly" activating a tap-hold shouldn't be conflated with quickly tapping one to repeat them.

To avoid confusion, it should be treated like any breaking change, by announcing it with a user change guide.

caksoylar:
> Do we want to completely remove global-quick-tap from the table, or should we leave it at the bottom and mark at as deprecated somehow?

I would just remove it given it is kept only for backwards compatibility but maybe contributors might have other preferences.

I also agree with @filterpaper that this would be a good time to change the name since there will be no coupling with `quick-tap-ms` anymore, but I don't have any specific naming suggestions myself 😅 I don't particularly like "instant tap" though, since "instant" is very close to "quick". Instead we could do e.g. `force-tap-ms` but that doesn't translate well to combos.

Ideally the name would reflect the idea of "prevent triggering of this behavior during rapid typing". Maybe something like `prior-idle-ms`?

filterpaper:
You could list these suggestions like `tap-override-ms` etc in a poll to get a consensus of a definition that will make sense to the community.

andrewjrae:
Yah I think a poll would be great, I do like `prior-idle-ms` personally. I did try to poll the #documentation channel for input last time but got nothing :(. Might try in some other keeb servers as well, if anyone has suggestions as to how to best facilitate this let me know :).

megavenik:
hi guys, what is the current status of this merge? Looks like no one here's against it, can we expect it to be approved any time soon?

andrewjrae:
This was on pause for a bit while we decided on better naming. Many ideas were thrown around on discord but I'm leaning towards `require-prior-idle-ms` for both combos and hold taps. I just need to get around to actually renaming and then a final review from @petejohanson and or @caksoylar.

infused-kim:
I have been using this for a few months and it’s astonishing how well it works. 

It makes homerow mods actually usable. 

And I think this will give users a huge reason to pick zmk over qmk. 

I believe zmk state of the firmware blog post is “due” soon. And it would probably be a great opportunity to advertise this feature and “Urob-style homerow mods” in it. 

So, perhaps it would be a good idea to merge this before the state of the firmware post.

andrewjrae:
Okay, renaming done, should be ready for a final review @petejohanson @caksoylar.

caksoylar (commented):
Thanks for the update! A few nitpicks on docs, but otherwise LGTM docs-wise.

(Not doing a "Github approval" since that should come from a non-docs maintainer.)

docs/docs/behaviors/hold-tap.md
```diff
@@ -62,15 +62,15 @@ gqt: global-quick-tap {
...
-    quick-tap-ms = <125>;
-    global-quick-tap;
+    quick-tap-ms = <200>;
+    require-prior-idle-ms = <125>;
     bindings = <&kp>, <&kp>;
 };
 ```
 
-If you press `&kp A` and then `&gqt LEFT_SHIFT B` **within** 125 ms, then `ab` will be output. Importantly, `b` will be output immediately since it was within the `quick-tap-ms`. This quick-tap behavior will work for any key press, whether it is within a behavior like hold-tap, or a simple `&kp`. This means the `&gqt LEFT_SHIFT B` binding will only have its underlying hold-tap behavior if it is pressed 125 ms **after** a key press.
+If you press `&kp A` and then `&gqt LEFT_SHIFT B` **within** 125 ms, then `ab` will be output. Importantly, `b` will be output immediately since it was within the `require-prior-idle-ms`. This "quick-tap" behavior will work for any key press, whether it is within a behavior like hold-tap, or a simple `&kp`. This means the `&gqt LEFT_SHIFT B` binding will only have its underlying hold-tap behavior if it is pressed 125 ms **after** a key press.
```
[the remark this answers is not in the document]
andrewjrae: Nit picking with suggested changes is a-okay with me ;)
Thanks for taking a look!

docs/docs/behaviors/hold-tap.md
```diff
@@ -47,30 +47,30 @@ Defines how long a key must be pressed to trigger Hold behavior.
...
+`require-prior-idle-ms` is like `quick-tap-ms` however it will apply for _any_ non-modifier key pressed before it. This effectively disables the hold-tap when typing quickly, which can be quite useful for homerow mods. It can also have the effect of removing the input delay when typing quickly.
 
-For example, the following hold-tap configuration enables `global-quick-tap` with a 125 millisecond `quick-tap-ms` term.
+For example, the following hold-tap configuration enables `require-prior-idle-ms` with a 125 millisecond term, alongside `quick-tap-ms` with a 200 millisecond term.
 
 ```
-gqt: global-quick-tap {
+rpi: require_prior_idle {
     label = "GLOBAL_QUICK_TAP";
```
caksoylar, line 61:
```suggestion
    label = "REQUIRE_PRIOR_IDLE";
```
This one I noted in the comments was missed, now I can add as a suggestion because it is in the context window of the existing diff :)

app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
```diff
@@ -20,8 +20,11 @@ properties:
     default: -1
   quick_tap_ms: # deprecated
     type: int
-  global-quick-tap:
+  global-quick-tap: # deprecated
```
petejohanson, line 23: I wasn't aware of this previously, but this should be using the dedicated `deprecated` property for bindings to properly warn on this one: https://docs.zephyrproject.org/latest/build/dts/bindings-syntax.html#deprecated

petejohanson (changes requested):
One minor nitpick as well, but otherwise LGTM.

app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
andrewjrae: Oh nice, I went ahead and labeled the other "deprecated" params as well, hopefully that's okay.

andrewjrae:
Okay, I realized my old sweep for global quick tap must have fallen short so I did a better search and caught a few more outliers. Please review the latest commit @petejohanson, I also ended up changing `cradio.keymap` since it was using the now deprecated `global-quick-tap`.

petejohanson (approved):
Thanks for the work on this and your patience. LGTM!

Issue #1393, opened 2022-07-22
Hold-tap feature to hold the hold key until hold-tap decision


Sixzero:
So I read over all the open:issues whether there is already a proposal for this, and I found none.

**Homerows are cool.**

Whynot take it one step further by providing SHIFT homerow one additional functionality?

- Basically we could hold down the SHIFT instantly on holding down our homerow key(F) and not only when some other keys are getting pressed. 

Current hold-tap solutions decide about modifiers when other key was pressed, which works really well, if you use ZMK keyboards, but not when want to use the SHIFT homerow with non-ZMK peripherals like a mouse. 

If the modifier key would get instantly pressed, it wouldn't disturb basically anything, except for GUI mod or maybe ALT mod, but other times there is no trigger usually attached for just holding down SHIFT in any softwares.

Others suggested to create a specific layer for gaming, since the problem came up with playing a game and SHIFT + mouse cordination of units is not that easy with ZMK keyboard, but on this gamer layer the problem is still not solved, since I would sometimes need the F key, and not just an &kp SHIFT key, so we basically lose a key, which is not an option for few key keyboards. 

In theory this this sounds easy to implement, since we have already other hold-tap solutions, nearly the exact solution, of the one requested. One small modification is necessary is that, we hold the mod key until there is another key is getting pressed(simultaneously). Otherwise everything work the same as it works now. 
Would love to know how easy/hard to implement this.

Tl;dr:
We need SHIFT homerow a special hold-tap, with instant mod press and not only decided on other key press.

Sixzero:
It could be not a flavor, but another label, called "zero-hold" or "instant-hold" or anything like that. It could add it this mod behavior for any of the flavors.

Sixzero:
@urob in discord proposed this macro as a solution, which is not far from what I am trying to achieve: 
<&macro_press &kp LEFT_SHIFT>,
<&macro_press &mt LEFT_SHIFT F>`, 
<&macro_pause_for_release>,
<&macro_release  &mt LEFT_SHIFT F>,
<&macro_release &kp LEFT_SHIFT> 

Also another solution is retro-tap with 0 ms tapping term. 

Going to try both, and see what works. :)

Sixzero:
The retro-tap solution seem a little too different from the non-retro-tap holdtap solutions.

The macro one seem working, but not sure, there is definitely delay before it pressing the shift, that is what I feel.

andrewjrae:
[...]

PR #1398, opened 2022-07-23
feat(behaviors): Hold-tap - Hold while undecided

nickconway:
Addresses #1393

### DIFF
--- a/app/src/behaviors/behavior_hold_tap.c
+++ b/app/src/behaviors/behavior_hold_tap.c
@@ -59,6 +59,7 @@ struct behavior_hold_tap_config {
     int quick_tap_ms;
     bool global_quick_tap;
     enum flavor flavor;
+    bool hold_while_undecided;
     bool retro_tap;
     bool hold_trigger_on_release;
     int32_t hold_trigger_key_positions_len;
@@ -67,10 +68,12 @@ struct behavior_hold_tap_config {
 
 // this data is specific for each hold-tap
 struct active_hold_tap {
+    bool first_press;
     int32_t position;
     uint32_t param_hold;
     uint32_t param_tap;
     int64_t timestamp;
+    bool hold_released;
     enum status status;
     const struct behavior_hold_tap_config *config;
     struct k_work_delayable work;
@@ -219,11 +222,13 @@ static struct active_hold_tap *store_hold_tap(uint32_t position, uint32_t param_
         if (active_hold_taps[i].position != ZMK_BHV_HOLD_TAP_POSITION_NOT_USED) {
             continue;
         }
+        active_hold_taps[i].first_press = true;
         active_hold_taps[i].position = position;
         active_hold_taps[i].status = STATUS_UNDECIDED;
         active_hold_taps[i].config = config;
         active_hold_taps[i].param_hold = param_hold;
         active_hold_taps[i].param_tap = param_tap;
+        active_hold_taps[i].hold_released = false;
         active_hold_taps[i].timestamp = timestamp;
         active_hold_taps[i].position_of_first_other_key_pressed = -1;
         return &active_hold_taps[i];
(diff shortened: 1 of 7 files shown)


docs/docs/behaviors/hold-tap.md
```diff
@@ -83,6 +83,10 @@ For example, if you press `&mt LEFT_SHIFT A` and then release it without pressin
 };
 ```
 
+#### `hold-while-undecided`
+
+If enabled, the hold behavior will immediately be held, and will release before the tap behavior is sent. With modifiers (excluding the windows key) this will not affect typing, and is useful for mod + clicking.
```
caksoylar, line 88: Note that `alt` modifiers can also do things like activate menu hotkeys when used on their own. Also worth noting this and `quick-tap` are exclusive (right)?

nickconway: A quick tap will never reach the hold-while-undecided code and vice-versa, so you can use them both on the same behavior.  Good thinking about the alt behavior, I quickly found out that part out.

Sixzero:
Wow, we even can merge is soon? You guys are crazy fast! Loving it!

Sixzero:
I am trying to understand why the modkey getting stuck with this option, but I don't see a problem in your code.

Sixzero:
I am guessing, the problem will be with this condition:
!(hold_tap->status == STATUS_HOLD_TIMER || hold_tap->status == STATUS_HOLD_INTERRUPT)

Sixzero:
If someone would review the pull, I would suggest it to get merged. :) The solution seems to be working perfectly! Also any code that was added, only runs if this flavor is turned on, so it won't have effect on other things.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -472,6 +501,8 @@ static void decide_hold_tap(struct active_hold_tap *hold_tap,
             status_str(hold_tap->status), flavor_str(hold_tap->config->flavor),
             decision_moment_str(decision_moment));
     undecided_hold_tap = NULL;
+
+    release_hold_binding(hold_tap, decision_moment);
```
caksoylar, line 505: Wouldn't it be better to put the if guard `if (hold_tap->config->hold_while_undecided && !hold_tap->hold_released)` from inside the function around every call of the function instead? I guess compiler could optimize it in theory?

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -645,6 +683,13 @@ static int keycode_state_changed_listener(const zmk_event_t *eh) {
         return ZMK_EV_EVENT_BUBBLE;
     }
 
+    if (undecided_hold_tap->first_press) {
```
caksoylar, line 686: The if checks can be combined here as well, since `first_press` isn't used unless `hold_while_undecided` is set isn't it?

Sixzero:
Could someone review and approve this? Would be good if it is out in official! It has been working like a charm since the last update!

Sixzero:
Could we get it reviewed. I have been using it since aug 1, and everything is working perfectly.
It really is and important feature, not just for me but for anyone trying to use SHIFT + mouse.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -433,6 +438,28 @@ static void decide_positional_hold(struct active_hold_tap *hold_tap) {
     hold_tap->status = STATUS_TAP;
 }
 
+static void release_hold_binding(struct active_hold_tap *hold_tap,
+                                 enum decision_moment decision_moment) {
+    bool keyTap =
+        !(hold_tap->status == STATUS_HOLD_TIMER || hold_tap->status == STATUS_HOLD_INTERRUPT);
+    bool holdRelease =
```
petejohanson, line 445: Style issue... we use snake_case variable names.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -433,6 +438,28 @@ static void decide_positional_hold(struct active_hold_tap *hold_tap) {
...
+        (hold_tap->status == STATUS_HOLD_TIMER || hold_tap->status == STATUS_HOLD_INTERRUPT) &&
+        decision_moment == HT_KEY_UP;
```
petejohanson, line 447:
```suggestion
        !key_tap && decision_moment == HT_KEY_UP;
```

petejohanson (changes requested):
@okke-formsma Thoughts on this?

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -645,6 +685,11 @@ static int keycode_state_changed_listener(const zmk_event_t *eh) {
         return ZMK_EV_EVENT_BUBBLE;
     }
 
+    if (undecided_hold_tap->first_press && undecided_hold_tap->config->hold_while_undecided) {
```
okke-formsma, line 688: This seems to be the only place where 'first_press' is used and I do not understand why it is needed at all. What does this code solve? Why should we not capture the events that would be captured on line 697?

okke-formsma:
I do not feel this MR is on the right track implementation-wise. It adds two new state variables (`first_pressed` and `hold_released`) while completely ignoring the existing `state` variable.

I would suggest to implement this in the following way, more in line with the original implementation:

* add a new `status` enum value `STATUS_UNDECIDED_INITIAL_HOLD`
* add a new method `is_undecided` which returns true a hold_tap's status is 'STATUS_UNDECIDED' or `STATUS_UNDECIDED_INITIAL_HOLD`. Replace all tests for `STATUS_UNDECIDED` with this new method.
* add a new `decision_moment` enum value `HT_KEY_DOWN`
* add a new method `decide_initial_hold` which is called at the top of `decide_hold_tap` which sets `STATUS_UNDECIDED_INITIAL_HOLD`
* modify `press_binding` to not press the hold again if `initial_hold` is enabled, and to release the hold when tap is decided.

Sixzero:
@okke-formsma your idea, of trying to solve the functionality without that 2 extra states, seems to be an important request. I also think the functionality should be more deeply integrated, while I think Nicon thought things would be more modular if he used these two states. 

I can't reach @nickconway on discord anymore, I have tried to do so in the last 3 weeks. I don't know what is happening to him, so probably the correction of the pull request is on us.

sporkus:
Just want to add that this PR works perfectly for me. I am using this for shift, gui(Mac) and ctrl(Windows). Without this option, homerow mods aren't very usable with mouse clicks.

Sixzero:
I am using it also, and shift + mouseclicks are insanely good! Loving it so much! I hope some day we will merge it!

theol0403:
Close this in favor of https://github.com/zmkfirmware/zmk/pull/1811

Issue #986, opened 2021-10-17
[BUG] Mod-taps frequently getting stuck


elkowar:
I've noticed more and more that my home-row modifier mod-tap keys keep getting stuck from time to time. So I did some analysis.

Assuming both S and L are configured as mod-taps for LSHFT and RSHFT respectively, and D and K are both mod-taps for LGUI. A tapping term of 150 is being used.

the following inputs consistently result in the shift modifier being stuck (which is then only resolvable by hitting reset)
HOLD(s), WAIT(200ms), TAP(l), RELEASE(s)

additionally, the following inputs:
HOLD(s), WAIT(200ms), TAP(**k**), RELEASE(s)
result in no immediate change, until either d or k (either of the mod-taps for LGUI) are tapped again. As soon as either of them is tapped, the gui modifier starts being stuck as held down.

Looking at the logs, I observed the **error message** `ACTIVE_HOLD_TAP_CLEANED_UP_TOO_EARLY`

Here is some more log output (this is showing the first input pattern I describe):
<summary>
logs
<details>
<pre>
[00:00:51.001,953] <dbg> zmk.zmk_endpoints_send_report: usage page 0x07
[00:00:54.885,833] <dbg> zmk.kscan_matrix_read: Sending event at 1,2 state on
[00:00:54.885,894] <dbg> zmk.zmk_kscan_process_msgq: Row: 1, col: 2, position: 14, pressed: true
[00:00:54.885,925] <dbg> zmk.position_state_changed_listener: 14 bubble (no undecided hold_tap active)
[00:00:54.885,955] <dbg> zmk.position_state_down: combo: capturing position event 14
[00:00:54.885,955] <dbg> zmk.zmk_event_manager_handle_from: Listener captured the event
[00:00:56.227,569] <dbg> zmk.kscan_matrix_read: Sending event at 4,3 state on
[00:00:56.227,630] <dbg> zmk.zmk_kscan_process_msgq: Row: 4, col: 3, position: 21, pressed: true
[00:00:56.227,661] <dbg> zmk.position_state_changed_listener: 21 bubble (no undecided hold_tap active)
[00:00:56.227,691] <dbg> zmk.position_state_down: combo: capturing position event 21
[00:00:56.227,722] <dbg> zmk.release_pressed_keys: combo: releasing position event 14
[00:00:56.227,722] <dbg> zmk.zmk_keymap_apply_position_state: layer: 0 position: 14, binding name: MOD_TAP
[...]

PR #1401, opened 2022-07-25
fix(behaviors): Fixing erroneous combo triggering

andrewjrae:
This is a very simple fix to a rather complicated issue. Essentially,
hold-taps will "release" (raise) their captured keys before actually
telling the event manager they have captured a key. This means the event manager ends up assigning the `last_listener_index` to the hold-tap subscription rather than the combo. So when the combo calls `ZMK_EVENT_RELEASE` it raises after the hold-tap instead of after the
combo as the combo code expects.

The corresponding test (which fails without this change) has also been added.

### DIFF
--- a/app/src/combo.c
+++ b/app/src/combo.c
@@ -253,7 +253,7 @@ static int release_pressed_keys() {
         if (i == 0) {
             LOG_DBG("combo: releasing position event %d",
                     as_zmk_position_state_changed(captured_event)->position);
-            ZMK_EVENT_RELEASE(captured_event)
+            ZMK_EVENT_RAISE_AFTER(captured_event, combo);
         } else {
             // reprocess events (see tests/combo/fully-overlapping-combos-3 for why this is needed)
             LOG_DBG("combo: reraising position event %d",
--- a/app/tests/combo/combos-and-holdtaps-3/events.patterns
+++ b/app/tests/combo/combos-and-holdtaps-3/events.patterns
@@ -1,0 +1,1 @@
+s/.*hid_listener_keycode_//p
--- a/app/tests/combo/combos-and-holdtaps-3/keycode_events.snapshot
+++ b/app/tests/combo/combos-and-holdtaps-3/keycode_events.snapshot
@@ -1,0 +1,5 @@
+pressed: usage_page 0x07 keycode 0xE5 implicit_mods 0x00 explicit_mods 0x00
+pressed: usage_page 0x07 keycode 0x05 implicit_mods 0x00 explicit_mods 0x00
+pressed: usage_page 0x07 keycode 0x06 implicit_mods 0x00 explicit_mods 0x00
+released: usage_page 0x07 keycode 0x05 implicit_mods 0x00 explicit_mods 0x00
+released: usage_page 0x07 keycode 0x06 implicit_mods 0x00 explicit_mods 0x00
--- a/app/tests/combo/combos-and-holdtaps-3/native_posix_64.keymap
+++ b/app/tests/combo/combos-and-holdtaps-3/native_posix_64.keymap
@@ -1,0 +1,40 @@
+#include <dt-bindings/zmk/keys.h>
+#include <behaviors.dtsi>
+#include <dt-bindings/zmk/kscan-mock.h>
+
+&mt {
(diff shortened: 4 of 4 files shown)


jmding8:
I merged in this fix into the keymap where I originally encountered the problem. This seems to be the fix. I have not been able to reproduce the original problem since, and things seem to be working as expected otherwise as well. I'll continue to test tomorrow.

petejohanson:
This workaround may fix combos, but I suspect you *might* have caught the core issue also causing #986 

In particular, this happens:

```
static int position_state_changed_listener(const zmk_event_t *eh) {
    struct zmk_position_state_changed *ev = as_zmk_position_state_changed(eh);

    // blah blah blah, skipped for clarity.

    LOG_DBG("%d capturing %d %s event", undecided_hold_tap->position, ev->position,
            ev->state ? "down" : "up");
    capture_event(eh);
    decide_hold_tap(undecided_hold_tap, ev->state ? HT_OTHER_KEY_DOWN : HT_OTHER_KEY_UP);
    return ZMK_EV_EVENT_CAPTURED;
}
```

So capturing the current event, then possibly rasing it, before it's actually been noted as captured by the return. Perhaps this needs to be fixed to not capture at all based on the decision.

@okke-formsma any thoughts?

okke-formsma:
Let's have a look at what's going on. The position state changed event is potentially handled multiple times by the hold-tap handler, because that's often needed if there are multiple hold-taps that are being decided by the same event (e.g., a key release which causes multiple hold-taps to decide).

```
struct zmk_position_state_changed *ev = as_zmk_position_state_changed(eh);
capture_event(eh);
ZMK_EVENT_RAISE_AT(captured_event, behavior_hold_tap)
```

if we would not capture the position state changed event and raise it again at behavior_hold_tap, we might not release all hold-taps correctly.

Example, multiple tap-preferred hold-taps:
* press `&ht A B` (new undecided hold-tap)
* press `&ht C D` (position event queued)
* press `&kp Z` (position event queued)
* release `&kp Z` (position event queued)

At this point, we have three position events in the queue, and an undecided hold tap. Then, by the `release &kp Z`, the undecided hold-tap is decided, and the three position events are released, causing:

* press `&ht C D` (new undecided hold-tap)
* press `&kp Z` (position event queued)
* release `&kp Z` (position event queued)

Now the `release &kp Z` position event caused the `&ht C D` undecided hold-tap to be decided, and the remaining two position events are released, which cause a normal press and release of `Z`.

TL;DR: we need to queue and release the position events, otherwise we do not properly handle nested hold-taps.

okke-formsma:
There is one more thing!

By re-raising the event, it may actually be fully handled! But the event handler code doesn't know!

```
event_manager.c
        case ZMK_EV_EVENT_CAPTURED:
            LOG_DBG("Listener captured the event");
            event->last_listener_index = i;
            // Listeners are expected to free events they capture
            return 0;
```

the event->last_listener_index is set to hold_tap, and that will cause it to be improperly re-released when the combo releases it!

Instead, the event->last_listener_index should be set before the handler is called!

okke-formsma:
would this be the right fix instead?

```
diff --git a/app/src/event_manager.c b/app/src/event_manager.c
index eef5d839..471432a8 100644
--- a/app/src/event_manager.c
+++ b/app/src/event_manager.c
@@ -25,6 +25,7 @@ int zmk_event_manager_handle_from(zmk_event_t *event, uint8_t start_index) {
         if (ev_sub->event_type != event->event) {
             continue;
         }
+        event->last_listener_index = i;
         ret = ev_sub->listener->callback(event);
         switch (ret) {
         case ZMK_EV_EVENT_BUBBLE:
@@ -35,7 +36,6 @@ int zmk_event_manager_handle_from(zmk_event_t *event, uint8_t start_index) {
             goto release;
         case ZMK_EV_EVENT_CAPTURED:
             LOG_DBG("Listener captured the event");
-            event->last_listener_index = i;
             // Listeners are expected to free events they capture
             return 0;
         default:
```

andrewjrae:
Making the change in the event handler seems like a sensible solution in general to me, it certainly provides a more consistent value of `last_listener_index`.

andrewjrae:
Closing because #1411 encapsulates this fix

PR #1403, opened 2022-07-26
Change wording in RGB_COLOR_HSB description

nickcoutsos:
Some minor spelling/grammar issues were bugging me but when I fixed them the sentence still felt awkward so I went for something more concise since it's already pointing the reader towards the syntax example.

I also removed the second "See below for an example" since the examples are in the very next section with a big heading to advertise them.


caksoylar (approved):
Thanks! There is still a bit of repetition between the parameter description and the note box but this is a strict improvement.

Hello9999901:
Look great :)

Hello9999901 (approved):

petejohanson (approved):

Issue #1281, opened 2022-05-02
Install guide does not align with setup script


dxmh:
On Discord, `goose#9364` [said](https://discord.com/channels/719497620560543766/719909884769992755/970509816718045254):

> in the zmk installation guide, the mcu selection comes first, but when i try it, the shield selection come first. is that a problem?

The “[User Config Setup Script](https://zmk.dev/docs/user-setup#user-config-setup-script)” section of the documentation needs to be reviewed and adjusted where necessary to align with the user journey through the setup script.

PR #1409, opened 2022-07-29
fix: #1281 change user-setup.md order to agree with order in setup sc…

dixls:
I just rearranged these two steps in the docs which is how it appears when running the script.

### DIFF
--- a/docs/docs/user-setup.md
+++ b/docs/docs/user-setup.md
@@ -90,12 +90,6 @@ powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('http
 
 ### Keyboard Selection
 
-:::note
-If you are building firmware for a new keyboard that is not included in the built-in
-list of keyboards, you can choose any keyboard from the list that is similar to yours (e.g. in terms of unibody/split and [onboard controller](hardware.mdx#onboard)/[composite](hardware.mdx#composite)) to generate the repository,
-and edit / add necessary files. You can follow the [new shield guide](development/new-shield.md) if you are adding support for a composite keyboard.
-:::
-
 When prompted, enter the number for the corresponding keyboard you would like to target:
 
 ```
@@ -109,10 +103,15 @@ Keyboard Selection:
 Pick an keyboard:
 ```
 
-If the keyboard selected uses an onboard controller you will skip the MCU Board selection step.
+:::note For a keyboard not in the included list:
+If you are building firmware for a new keyboard that is not included in the built-in
+list of keyboards, you can choose any keyboard from the list that is similar to yours (e.g. in terms of unibody/split and [onboard controller](hardware.mdx#onboard)/[composite](hardware.mdx#composite)) to generate the repository,
+and edit / add necessary files. You can follow the [new shield guide](development/new-shield.md) if you are adding support for a composite keyboard.
+:::
 
 ### MCU Board Selection
 
+If the keyboard selected uses an onboard controller you will skip this step.
 When prompted, enter the number for the corresponding MCU board you would like to target:
 
 ```
--- a/docs/docs/user-setup.md
+++ b/docs/docs/user-setup.md
@@ -91,7 +91,7 @@ powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('http
 ### Keyboard Selection
 
 :::note
-If you are building firmware for a new keyboard or shield that is not included in the built-in
+If you are building firmware for a new keyboard that is not included in the built-in
 list of keyboards, you can choose any keyboard or shield from the list that is similar to yours to generate the repository,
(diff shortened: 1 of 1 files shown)


caksoylar:
Thanks for the PR! Unfortunately there is a bit more of a change than just changed order for the two: You'll notice that it first asks for "keyboard" and not "keyboard shield," since the list now includes keyboards with onboard controllers (in which case the "board" selection menu isn't presented). This would also require rewording the "new shield" note box. Would you like to give a shot to improve the PR given these details?

dixls:
Gotcha, I'll give it a shot with those notes in mind, and try to make it more representative of the actual process!

dixls:
I made some changes, curious to see what you or anyone thinks.

I wonder if, separately from this issue, it might be clearer to make a route in the setup script that allows a user to setup a new shield without choosing an existing shield to build from?

petejohanson:
> I wonder if, separately from this issue, it might be clearer to make a route in the setup script that allows a user to setup a new shield without choosing an existing shield to build from?

That's definitely been on my wish list to add. It would be a welcomed addition.

docs/docs/user-setup.md
```diff
@@ -88,34 +88,40 @@ powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('http
 </TabItem>
 </Tabs>
 
-### MCU Board Selection
+### Keyboard Selection
 
-When prompted, enter the number for the corresponding MCU board you would like to target:
+:::note
```
caksoylar, line 93: How about moving this note box to below the `Keyboard Selection:` code block?

docs/docs/user-setup.md
```diff
@@ -88,34 +88,40 @@ powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('http
...
+If you are building firmware for a new keyboard or shield that is not included in the built-in
+list of keyboards, you can choose any keyboard or shield from the list that is similar to yours to generate the repository,
+and edit / add necessary files according to the [guide for adding new keyboard shield](development/new-shield.md).
```
caksoylar, line 96: It might be good to qualify that the guide is for shields and not for integrated controllers as such:
```suggestion
and edit / add necessary files. You can follow the [new shield guide](development/new-shield.md) if you are adding support for a composite keyboard.
```

docs/docs/user-setup.md
```diff
@@ -88,34 +88,40 @@ powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('http
 </TabItem>
 </Tabs>
 
-### MCU Board Selection
+### Keyboard Selection
 
-When prompted, enter the number for the corresponding MCU board you would like to target:
+:::note
+If you are building firmware for a new keyboard or shield that is not included in the built-in
```
caksoylar, line 94: I think we can simplify to just "keyboards" here:
```suggestion
If you are building firmware for a new keyboard that is not included in the built-in
```

docs/docs/user-setup.md
```diff
@@ -88,34 +88,40 @@ powershell -Command "iex ((New-Object System.Net.WebClient).DownloadString('http
...
+list of keyboards, you can choose any keyboard or shield from the list that is similar to yours to generate the repository,
```
caksoylar, line 95:
```suggestion
list of keyboards, you can choose any keyboard from the list that is similar to yours (e.g. in terms of unibody/split and [onboard controller](hardware.mdx#onboard)/[composite](hardware.mdx#composite)) to generate the repository,
```

caksoylar (commented):
Thanks again! I added a few suggestions on clarifying onboard/composite w.r.t. the custom keyboard note box.

docs/docs/user-setup.md
dixls: I think this is a good idea, but what's the thought on more descriptive titles for the notes? I think once the note gets to be this long it makes it a little harder to parse, and having a more descriptive title might be helpful, but I don't know if this is done elsewhere in the docs.

caksoylar: If I understand you correctly, yes we use titles for note boxes in a few places like here: https://zmk.dev/docs/behaviors/bluetooth#summary. I think we could add a name to this one as well.

caksoylar: @dixls Are you planning to move this to the end of this subsection (and add a note title if you like)? I think that's the only comment I have left, otherwise looks good to me.

dixls: Yes! Sorry, I had made these changes and then got pulled away before checking it over and pushing. Let me know how this version looks, not sure if the Note title feels redundant, but it may more successfully grab peoples attention.

caksoylar (approved):

caksoylar:
Thank you @dixls and sorry it took so long! It fell off my radar.

PR #1114, opened 2022-02-03
feat(behaviors): Allow mod-morph to swallow mods

vrinek:
This PR is based on https://github.com/aumuell/zmk/commit/fa61c6a3c6e7cbfcb5b3d9168b3454bf95c and most, if not all, of the credits should go to @aumuell. I have only contributed some git-fu to get their work rebased on top of main.

# Related issues

Fixes https://github.com/zmkfirmware/zmk/issues/683.

# Feature description

This PR adds a new option to mod-morph: `masked_mods`. This option complements the `mods` option, and allows the user to configure which of the mods will be _swallowed_ when the morph triggers.

## Behavior without `masked_mods`

```
behaviors {
    mmlt: grave_escape {
        compatible = "zmk,behavior-mod-morph";
        label = "LPAR_LT";
        #binding-cells = <0>;
        bindings = <&kp LPAR>, <&kp COMMA>;
        mods = <(MOD_LSFT|MOD_RSFT)>;
    };
};
```

- Pressing `&mmlt` results in `(`
- Pressing `L-shift + &mmlt` results in `L-shift + ,`, which is interpreted as `<`
- Pressing `R-shift + &mmlt` results in `R-shift + ,`, which is interpreted as `<` 

## Behavior with `masked_mods`

```diff
 behaviors {
     mmlt: grave_escape {
         compatible = "zmk,behavior-mod-morph";
         label = "LPAR_LT";
         #binding-cells = <0>;
         bindings = <&kp LPAR>, <&kp COMMA>;
         mods = <(MOD_LSFT|MOD_RSFT)>;
+        masked_mods = <(MOD_LSFT)>;
     };
 };
```

- Pressing `&mmlt` results in `(`
- **Pressing `L-shift + &mmlt` results in `,`** (the `L-shift` is swallowed)
- Pressing `R-shift + &mmlt` results in `R-shift + ,`, which is interpreted as `<` 

# Manual testing

Sadly, the only keyboard I have at hand that runs ZMK is a Corne-ish Zen which (for the time being) depends on a fork of this repo (https://github.com/LOWPROKB/zmk). As such **I have performed no testing of this branch** and would welcome any testers to lend a hand.

What I have tested is https://github.com/vrinek/zmk/tree/masked-mod-morphs-corne-ish-zen, which is the same changes but on top of the branch that works for my keyboard:

1. defined the behavior as described at "Behavior with `masked_mods`"
2. tried pressing the key I bound :heavy_check_mark: 
[...]

Issue #683, opened 2021-02-18
when using mod-morph to make shift-backspace work as delete, the shift key is not 'swallowed'


depadiernos:
I set up mod-morph for the backspace key so that shift-backspace sends the delete keycode. Most operating systems have a keyboard shortcut for delete (e.g. [shift-delete ](https://defkey.com/what-means/shift-del)) which prevents the delete key to work on its own.

As a user, I'd like to be able configure mod-morph where the shift is swallowed so the morphed keycode can be sent by itself. 

Here is an example [keymap](https://github.com/depadiernos/zmk-config/blob/f60ee0bd9023b39624e577a31765c4411822fbd4/config/halfboard.keymap#L37)

okke-formsma:
I think the easiest implementation would be to make a "negative" version of the `implicit_modifiers`, which turns a modifier off for until the next keypress.

trankillity:
Just adding a voice to this. Would love to set up Shift + comma as semicolon, but obviously the shift applies to the semicolon, producing a colon.

alex-popov-tech:
agreed on that, would be very cool to see!

metheon:
Adding my voice to this issue with a couple of examples:

Say I wanna override `shift+space` to be `gui+tab`, I’d then in reality get `shift+gui+tab` where ideally, in my mind, I’d get exactly `gui+tab` as the morphed output.

It’s also a problem in other scenarios, say I wanna override `2` and `3` on my numpad layer to be `,` and `.` together with shift. Then that would output `<` and `>` as that’s the shifted version of `,` and `.`

vrinek:
I am having a similar issue to what others here have mentioned.

As far as I understand, taking [the comment right above me](https://github.com/zmkfirmware/zmk/issues/683#issuecomment-1019086467) as an example (`shift+2` to `,`):

### What we have

1. `shift` down
2. `2` down morphed into `,` down
3. `shift+,` is what the computer sees, interpreting it as `<`
3. `2` up morphed into `,` up
4. `shift` up

### What we would like

_Please correct me if I got this wrong_

1. `shift` down
2. `2` down morphed into `shift` up, followed by `,` down
3. `,` is what the computer sees, interpreting it as `,`
4. `2` up morphed into `,` up, followed by `shift` down
5. `shift` up

[...]

PR #1412, opened 2022-07-31
Fix mod-morph

urob:
This is an updated version of https://github.com/zmkfirmware/zmk/pull/1114. It builds on @aumuell's masked-mods fix 241f82e6cefc3661de4698a6eef4ee511fa8580b. Notable differences with respect to https://github.com/zmkfirmware/zmk/pull/1114 are:

- fixes issue where modifiers in the binding are ignored by allowing `masked_mods` to be overwritten by implicit mods 
- fixes issue where masked mods does not work when the behavior is bind to a hold-tap
- fixes issue where masked mods are unmasked before key release
- added unit tests
- added documentation
- new default: modifiers won't be passed along with morphed binding (fixing https://github.com/zmkfirmware/zmk/issues/683 without the need to specify `masked_mods`)


docs/docs/behaviors/mod-morph.md
```diff
@@ -71,3 +65,30 @@ Example:
 ```
 mods = <(MOD_LGUI|MOD_LSFT|MOD_RGUI|MOD_RSFT)>;
 ```
+
+### Advanced configuration
+
+`masked_mods`
```
caksoylar, line 71: I believe we prefer hyphenated property names by convention, so `masked-mods` would be the right format for the docs (and yaml/dtsi).

docs/docs/behaviors/mod-morph.md
```diff
@@ -71,3 +65,30 @@ Example:
...
+            mods = <(MOD_LSFT|MOD_RSFT)>;
+            masked_mods = <(MOD_LSFT)>;
+        };
+    };
+};
+```
+
+### Limitations
+
+In some circumstance, when the morphed keycode is a hold-tap, the modifier gets sent along with the hold-tap regardless of the specification of `masked_mods`.
```
caksoylar, line 94:
```suggestion
Under certain circumstances where the morphed keycode is a hold-tap, the modifier gets sent along with the hold-tap regardless of the `masked_mods` value.
```

urob: Actually, that part is hopefully obsolete given b903e4f3710d0f331407c6ad26ce466a3a310845. Still waiting for @nickconway to confirm before removing the note.

docs/docs/behaviors/mod-morph.md
urob: done

docs/docs/behaviors/mod-morph.md
urob: It seems the fix worked and mod-morphing into hold-tap no longer ignores masked-mods. So I am dropping the disclaimer

urob:
Quick ping @petejohanson: I think this is ready for review if you get the chance. Below a quick summary of the technical implementation to, hopefully, facilitate the review process.

At its core, the PR does a bit of refactoring of how mods are handled in `hid.c`. In addition to the `explicit_modifiers` mod flag, it now defines two new persistent mod flags: `implicit_modifiers` and `masked_modifiers`. 

Any change in modifiers --- i.e., any call to `zmk_hid_register_mod()`, `zmk_hid_unregister_mod()`, `zmk_hid_implicit_modifiers_press()`, `zmk_hid_implicit_modifiers_release()`, `zmk_hid_masked_modifiers_set()` and `zmk_hid_masked_modifiers_clear()` --- results in a call to `SET_MODIFIERS()`. The `SET_MODIFIERS()` macro in turn sets the modifiers as follows:
```
keyboard_report.body.modifiers = (mods & ~masked_modifiers) | implicit_modifiers;
```

Using the refactored mod behavior, the PR then changes `behavior_mod_morph.c` to first set the `masked_modifiers` mod flag to a configured value when a mod-morph is triggered (by default the mods that trigger it). After the mod-morph is done, it in then clears `masked_modifiers`.

app/src/behaviors/behavior_mod_morph.c
```diff
@@ -88,6 +93,8 @@ static int behavior_mod_morph_init(const struct device *dev) { return 0; }
         .normal_binding = _TRANSFORM_ENTRY(0, n),                                                  \
         .morph_binding = _TRANSFORM_ENTRY(1, n),                                                   \
         .mods = DT_INST_PROP(n, mods),                                                             \
+        .masked_mods = COND_CODE_0(DT_INST_NODE_HAS_PROP(n, masked_mods), (DT_INST_PROP(n, mods)), \
```
petejohanson, line 96: This fallback makes sense, in theory, but changes existing default mod morph behavior....

docs/docs/behaviors/mod-morph.md
```diff
@@ -71,3 +65,26 @@ Example:
 ```
 mods = <(MOD_LGUI|MOD_LSFT|MOD_RGUI|MOD_RSFT)>;
 ```
+
+### Advanced configuration
+
+`masked-mods`
+
+When a modifier specified in `mods` is being held, it won't be sent along with the morphed keycode if it is also part of `masked-mods`. To sent all modifiers along with the morphed keycode, set `masked-mods` to `<0>`. By default, `masked-mods` equals `mods`.
```
petejohanson, line 73: Doesn't this mean the default behavior for mod morph is changed as part of this PR?

Do we think there's any reason to be concerned by that?

petejohanson (commented):
Code changes all seem reasonable, just the one point of discussion around the default behavior being changed by this PR.

docs/docs/behaviors/mod-morph.md
urob: My reading was that most considered the old default a bug (see #683). Personally, I could go either way. But I thought that the proposed default may simplify life for the typical use-case of mod-morph?

petejohanson: Plenty of voices there that would seem to suggest a new default behavior is fine... Wonder how many folks use it today is as and it works well?

urob: Also, one can always enforce that a mod gets sent along by adding it to the binding as implicit mod. Hopefully that'll be intuitive enough so that most people won't even need to adjust the masked-mods property?

petejohanson: Yeah, I am leaning towards this being an "in the greater good" breaking change in behavior, given the fact the original versions sending along the mods was "odd" in the first place.

@joelspadin @Nicell @okke-formsma Thoughts?

eckig: > Plenty of voices there that would seem to suggest a new default behavior is fine... Wonder how many folks use it today is as and it works well?

I am currently using it on my keyboard: Works great (nice work @urob ). And in my opinion you should change the default behavior as it really is somewhat odd (you can always add modifiers via functions but can not remove them later on).

snoyer: > Yeah, I am leaning towards this being an "in the greater good" breaking change in behavior, given the fact the original versions sending along the mods was "odd" in the first place.
> 
> @joelspadin @Nicell @okke-formsma Thoughts?

I am using it and it is working for my use-case of shift-morphing to have `,;`, `.?` and `/\` keys (instead of the regular qwerty `,<`, `.>` and `\|`).
I'm also of the opinion that masking the held mods by default is more sensible than having to mask them explicitly.

To nitpick further, `masked_mods` makes sense at the implementation level (in `hid.c`) but it feels counter-intuitive in the user-facing behavior. Exposing something like `keep-mods` for the behavior would make more sense in my opinion.
The doc could read:
> When a modifier specified in `mods` is being held it won't be sent along with the morphed keycode, unless it is also specified in `keep-mods`.

`keep-mods` would be empty by default which would be equivalent to `masked-mods`=`mods` by default

Nicell: > Yeah, I am leaning towards this being an "in the greater good" breaking change in behavior, given the fact the original versions sending along the mods was "odd" in the first place.

I agree with everyone else here. By default mods should be masked. This was not done originally because Okke didn't have the time to look into it AFAIK. I think this is better behavior, and I can't imagine many people taking advantage of this odd default behavior.

urob: > To nitpick further, `masked_mods` makes sense at the implementation level (in `hid.c`) but it feels counter-intuitive in the user-facing behavior. Exposing something like `keep-mods` for the behavior would make more sense in my opinion.
> 
> The doc could read:
> 
> > When a modifier specified in `mods` is being held it won't be sent along with the morphed keycode, unless it is also specified in `keep-mods`.
> 
> `keep-mods` would be empty by default which would be equivalent to `masked-mods`=`mods` by default

Re: masked-mods vs keep-mods, there's a minor restriction associated with the latter in that one wouldn't be able to mask any mods *other* than the ones triggering the mod-morph.  Not sure how much that's a restriction in practice, but it may limit some creative uses?

okke-formsma: Masking the mods was indeed something that I wanted to do but didn't make it into the original PR.

Masking by default seems to make sense. I don't have an opinion on the `masked-mods` vs `keep-mods` debate.

okke-formsma:
lgtm

docs/docs/behaviors/mod-morph.md
petejohanson: > > To nitpick further, `masked_mods` makes sense at the implementation level (in `hid.c`) but it feels counter-intuitive in the user-facing behavior. Exposing something like `keep-mods` for the behavior would make more sense in my opinion.
> > The doc could read:
> > > When a modifier specified in `mods` is being held it won't be sent along with the morphed keycode, unless it is also specified in `keep-mods`.
> > 
> > 
> > `keep-mods` would be empty by default which would be equivalent to `masked-mods`=`mods` by default
> 
> Re: masked-mods vs keep-mods, there's a minor restriction associated with the latter in that one wouldn't be able to mask any mods _other_ than the ones triggering the mod-morph. Not sure how much that's a restriction in practice, but it may limit some creative uses?

That's a good call out. I think potentially useful to support that for cases we're not imaging yet for this.

snoyer: > > Re: masked-mods vs keep-mods, there's a minor restriction associated with the latter in that one wouldn't be able to mask any mods _other_ than the ones triggering the mod-morph. Not sure how much that's a restriction in practice, but it may limit some creative uses?
> 
> That's a good call out. I think potentially useful to support that for cases we're not imaging yet for this.

I hadn't thought that far. 
My issue was that `masked-mods` looked like an implementation detail leaking into the API for no good reason.
"Future use-cases" could be reason enough to go with it; and with the with the new default I assume it wouldn't show up in the vast majority of cases anyway.

caksoylar: I agree with snoyer, in addition if we are changing the default then it's also fair to break not-yet-imagined use cases that `masked-mods` might allow. Making it more intuitive for users should take priority here IMO.

petejohanson: Thinking about this more... I think I'm now leaning towards `keep-mods` for this. If we really need to filter other mods not specified in the morph later, we can solve that another way.

urob: The main use I could imagine is to set up a "dummy-mod-morph" which doesn't actually do anything other than masking mods. Currently, this is the only way to access the backend hid-functions `zmk_hid_masked_modifiers_set` and  `zmk_hid_masked_modifiers_clear` from the user side. But I agree that if there is demand for that, it would probably better to create a new "negative-mods" behavior that more directly exposes the masked_mods functionality.

I just pushed a new commit that switches from masked-mods to keep-mods. I tried out a couple of edge cases and it all works well on my side.

joelspadin: Having a behavior to set/clear masked mods would also be useful for sending strings, as you could lock out mods before sending the string and restore them after. That said, if the string is too long and it overflows the behavior queue, the restore behavior would never be sent and you wouldn't be able to use those mods anymore, so that wouldn't be great. Maybe instead we could add an optional mask property to specific behaviors, so for example you could add a `&kp` variant that masks mods and use that instead of `&kp` when sending strings.

docs/docs/behaviors/mod-morph.md
```diff
@@ -71,3 +65,26 @@ Example:
 ```
 mods = <(MOD_LGUI|MOD_LSFT|MOD_RGUI|MOD_RSFT)>;
 ```
+
+### Advanced configuration
+
+`keep-mods`
+
+When a modifier specified in `mods` is being held, it won't be sent along with the morphed keycode, unless it is also specified in `keep-mods`. By default, `keep-mods` equals `0`.
```
caksoylar, line 73:
```suggestion
When a modifier specified in `mods` is being held, it won't be sent along with the morphed keycode unless it is also specified in `keep-mods`. By default `keep-mods` equals `0`, which means no modifiers will be preserved.
```

caksoylar (approved):
Looks good to me from a docs perspective.

petejohanson (approved):
Thanks for all the work on this!

PR #1414, opened 2022-08-01
feat(behaviors): Turbo Key

nickconway:
Functionality for continuous repeating of any behavior.


app/src/behaviors/behavior_turbo_key.c
```diff
@@ -0,0 +1,193 @@
...
+
+    // Timer Data
+    bool timer_started;
+    bool timer_cancelled;
+    bool turbo_decided;
+    int64_t release_at;
+    struct k_work_delayable release_timer;
+};
+
+struct active_turbo active_turbos[ZMK_BHV_TURBO_MAX_ACTIVE] = {};
```
petejohanson, line 46: All of this seems a bit overly complex for this... Why not just have a single delayable work that lives in the "data" struct tied to each instance of the behavior?

That would avoid all the tracking of active turbos, etc.

Seems this might have been inspired by the hold-tap logic... I'm not sure this has the same needs/complexity as that.

app/src/behaviors/behavior_turbo_key.c
```diff
@@ -0,0 +1,193 @@
...
+        return;
+    }
+    if (turbo->timer_cancelled) {
+        return;
+    }
+    LOG_DBG("Turbo timer reached.");
+    struct zmk_behavior_binding_event event = {.position = turbo->position,
+                                               .timestamp = k_uptime_get()};
+    zmk_behavior_queue_add(event.position, turbo->config->binding, true, turbo->config->tap_ms);
+    zmk_behavior_queue_add(event.position, turbo->config->binding, false, 0);
```
petejohanson, line 113: Love the use of the behavior queue for this!

petejohanson (commented):
One main question on the state management here. Thanks for the PR!

app/src/behaviors/behavior_turbo_key.c
nickconway: Fixed in recent commits.

docs/docs/behaviors/turbo.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Turbo Behavior
+sidebar_label: Turbo Key
+---
+
+## Summary
+
+The turbo behavior will repeatedly trigger a behavior after a specified amount of time.
```
caksoylar, line 8: I am a bit unclear on how exactly this works and it would be good to clarify further here. I think by default, (no toggle term) it only repeats while the key is held, and cancels repeating when it is released, is it?

But if toggle term is set and tap duration is shorter than that, it stays active and the next press cancels it?

app/src/behaviors/behavior_turbo_key.c
```diff
@@ -0,0 +1,156 @@
...
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_turbo_config *cfg = dev->config;
+    struct behavior_turbo_data *data = dev->data;
+
+    if (!data->is_active) {
+        data->is_active = true;
+
+        LOG_DBG("%d started new turbo", event.position);
+        data->press_time = k_uptime_get();
+        k_work_init_delayable(&data->release_timer, behavior_turbo_timer_handler);
```
petejohanson, line 101: This should be initialized in the init function for the behavior instance, not every time here.

app/src/behaviors/behavior_turbo_key.c
```diff
@@ -0,0 +1,156 @@
...
+        zmk_behavior_queue_add(event.position, cfg->binding, true, cfg->tap_ms);
+        zmk_behavior_queue_add(event.position, cfg->binding, false, 0);
```
petejohanson, line 103: This is a little duplicated from the same code in the timer handler, Can we pull that out to avoid duplication?

app/src/behaviors/behavior_turbo_key.c
```diff
@@ -0,0 +1,156 @@
...
+}
+
+static const struct behavior_driver_api behavior_turbo_key_driver_api = {
+    .binding_pressed = on_keymap_binding_pressed,
+    .binding_released = on_keymap_binding_released,
+};
+
+#define _TRANSFORM_ENTRY(idx, node)                                                                \
+    {                                                                                              \
+        .behavior_dev = DT_LABEL(DT_INST_PHANDLE_BY_IDX(node, bindings, idx)),                     \
```
petejohanson, line 135: Needs to be refactored to:

```suggestion
        .behavior_dev = DT_PROP(DT_INST_PHANDLE_BY_IDX(node, bindings, idx), label),                     \
```

For some Zephyr changes.

app/src/behaviors/behavior_turbo_key.c
```diff
@@ -0,0 +1,156 @@
...
+    uint32_t position;
+    bool is_active;
+    bool is_pressed;
+
+    int32_t press_time;
+
+    int tap_ms;
+    int wait_ms;
+    struct zmk_behavior_binding binding;
+
```
petejohanson, line 38: I don't see any reason this is duplicated in the data, and in the config. Should just be in the config?

app/tests/turbo/behavior_keymap.dtsi
```diff
@@ -0,0 +1,36 @@
...
+            bindings = <&kp C>;
+        };
+        t2: turbo2 {
+            compatible = "zmk,behavior-turbo-key";
+            label = "turbo2";
+            #binding-cells = <0>;
+            tap-ms = <5>;
+            wait-ms = <300>;
+            toggle-term-ms = <50>;
+            bindings = <&kp C>;
```
petejohanson, line 22: Is there any reason we couldn't "parameterize" this the same way we did for macros recently? Seems a shame to force a new turbo for every keycode you want to use like this, for instance.

petejohanson (changes requested):
Sorry for the delay in review. A few thoughts on the implementation and a high level question on parameterizing this.

yuannan:
Is there any reasons why this hasn't been merged? I'd be happy to contribute if there is a clear spec that can be met.

app/tests/turbo/behavior_keymap.dtsi
nickconway: They're parameterized now! These and the macros share a lot of functionality regarding the parameters now, it might make sense to break that all out as well.

0xNul:
Would be great to see this added to the behaviors.

yuannan:
@nickconway do you think it's worth a go reopening this?

PR #1415, opened 2022-08-01
feat(docs): Added a Spelling Error Caution Notice

Hello9999901:
Having helped a number of people in the Discord server, I've noticed many run into errors with writing the wrong keycode name. After discussing with Guest-42 (a person who ran into this error), he discussed the possibility of adding a caution notice as a friendly reminder to make sure the codes are spelled correctly. I am not sure if this will help a lot, so please change as you see fit :)


CB-42458 (approved):
Seems good to me, as I described in the discord server.

caksoylar:
I think this is a pretty trivial warning for most people so I don't particularly like it. By that logic we'd have to add it to all configuration sections, because most typos in DT syntax (maybe except for label/node names etc.) will result in errors. I'll defer to other reviewers if they think it is helpful, though.

petejohanson:
I'm a tad conflicted on this as well, TBH. Seems duplicative to add everywhere, I'm curious if it would actually help all that much as well, given users might read right past it if it's everywhere.

Hello9999901:
> I'm a tad conflicted on this as well, TBH. Seems duplicative to add everywhere, I'm curious if it would actually help all that much as well, given users might read right past it if it's everywhere.

Hmm... I agree adding it everywhere is repetitive and will encourage skipping now that I'm thinking about it. What about only adding it on the full list, since most people would go to there for reference first?

caksoylar:
Sorry, I seem to have missed this. I think that's a good compromise, I'd be good with that.

caksoylar (approved):
Thanks, looks good on the preview as well!

docs/src/components/codes/SpellingCaution.jsx
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
kurtis-lew, line 2:
```suggestion
 * Copyright (c) 2022 The ZMK Contributors
```

petejohanson (approved):

Issue #1410, opened 2022-07-30
Feature request: continue-mod-list for caps-word


urob:
The current behavior is that caps-word continues on all mod. It would be useful to add a configuration option that allows the user to specify for which mods caps-word continues and for which it is deactivated.

Relatedly, would deactivating caps-word on mod-presses be a more sensible default?

urob:
Upon further consideration, having a complete `continue-mod-list` might be overkill. In most use cases, a simple `ignore-modifiers` property that can be turned on or off by the user should be enough. To that effect, I added a PR implementing that feature: https://github.com/zmkfirmware/zmk/pull/1422

urob:
This is fixed by https://github.com/zmkfirmware/zmk/pull/1451. With the PR one can use `/delete-property ignore-modifiers/` and then use `continue-list` to specify for which modifiers caps-word is *not* cancelled. For example, to continue caps-word on shift but cancel it with all other modifiers, one would add the following to the user-config:
```
&caps_word {
    /delete-property/ ignore-modifiers;
    continue-list = <UNDERSCORE  BACKSPACE DELETE LSHFT RSHFT>;
};
```

gregorias:
This feature would be great. My problem is that when I type with caps word, e.g., `FOOBA` and then press ⌃N to traverse an autocompletion list, ZMK types in ⌃⇧N (it’s even worse for deleting a word, ⌃W; ⌃⇧W happens to delete a tab in my terminal). Is there some stop-gap measure to address this that is not “just remember to exit caps word?”

aadcg:
I have used this feature both in QMK and ZMK and I believe that the former has a saner default. 

@gregorias mentioned a pitfall of the current default. Another one is Alt+Tab becoming Alt+Shift+Tab. 

> (QMK docs) Caps Word automatically disables itself at the end of the word. That is, it stops by default once a space or any key other than KC_A--KC_Z, KC_0--KC_9, KC_MINS, KC_UNDS, KC_DELETE, or KC_BACKSPACE is pressed. Caps Word also disables itself if the keyboard is idle for 5 seconds.

PR #1422, opened 2022-08-05
Ignore-modifiers property for capsword

urob:
**EDIT: This PR is now superseded by https://github.com/zmkfirmware/zmk/pull/1451**

This adds an `ignore-modifiers` property to `caps_word`. If set to `true` (the default), caps word will remain active when any modifiers are pressed. This is the current behavior.

This PR allows the user to overwrite the behavior by deleting the (new) `ignore-modifiers` property in their keymap:
```
&caps_word {
    /delete-property/ ignore-modifiers;
};
```

When the property is deleted, pressing a modifier will deactivate caps word. This is useful, for example, for people who activate caps word with somewhat complex key combos or tap dances and who wish to deactivate it by simply pressing `LEFT_SHIFT`.

----
Discussion: for now, the new `ignore-modifiers` property is set to `true` by default. This is in order to ensure backwards-compatibility. My personal feeling is, however, that setting it to `false` would be a more sensible default. But I didn't want to do it without discussion.

### DIFF
--- a/app/dts/behaviors/caps_word.dtsi
+++ b/app/dts/behaviors/caps_word.dtsi
@@ -13,6 +13,7 @@
 			label = "CAPS_WORD";
 			#binding-cells = <0>;
 			continue-list = <UNDERSCORE BACKSPACE DELETE>;
+            ignore-modifiers;
 		};
 	};
 };
--- a/app/dts/bindings/behaviors/zmk,behavior-caps-word.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-caps-word.yaml
@@ -13,3 +13,5 @@ properties:
     required: true
   mods:
     type: int
+  ignore-modifiers:
+    type: boolean
--- a/app/src/behaviors/behavior_caps_word.c
+++ b/app/src/behaviors/behavior_caps_word.c
@@ -32,6 +32,7 @@ struct caps_word_continue_item {
 
 struct behavior_caps_word_config {
     zmk_mod_flags_t mods;
+    bool ignore_modifiers;
     uint8_t index;
     uint8_t continuations_count;
     struct caps_word_continue_item continuations[];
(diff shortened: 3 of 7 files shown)


caksoylar:
This feature caused me some confusion recently, while typing a `CONSTANT_VALUE` in an editor and wanting to accept its completion with `ctrl+n`, which became `ctrl+shift+n` and opened a new terminal window and making me think I got a hold-tap stuck.

If I remember correctly this change was introduced so you could type e.g. `_` (which requires shifting) without deactivating caps lock. Would it make sense to only ignore the modifiers specified in `mods`?

urob:
> This feature caused me some confusion recently, while typing a `CONSTANT_VALUE` in an editor and wanting to accept its completion with `ctrl+n`, which became `ctrl+shift+n` and opened a new terminal window and made me think I got a hold-tap stuck.
> 
> If I remember correctly this change was introduced so you could type e.g. `_` (which requires shifting) without deactivating caps word. Would it make sense to only ignore the modifiers specified in `mods`?

Hm, I have `&kp UNDER` on a combo, so I don't need to use shift to type `_`. But I see how having more flexibility here might be useful. Perhaps having a `continue-mod-list` as suggested in https://github.com/zmkfirmware/zmk/issues/1410 might be the better solution after all.

Leaving this up in the meantime as it still might be useful for others with `&kp UNDER` on combos or dedicated symbol layers.

urob:
> > This feature caused me some confusion recently, while typing a `CONSTANT_VALUE` in an editor and wanting to accept its completion with `ctrl+n`, which became `ctrl+shift+n` and opened a new terminal window and made me think I got a hold-tap stuck.
> > If I remember correctly this change was introduced so you could type e.g. `_` (which requires shifting) without deactivating caps word. Would it make sense to only ignore the modifiers specified in `mods`?
> 
> Hm, I have `&kp UNDER` on a combo, so I don't need to use shift to type `_`. But I see how having more flexibility here might be useful. Perhaps having a `continue-mod-list` as suggested in #1410 might be the better solution after all.
> 
> Leaving this up in the meantime as it still might be useful for others with `&kp UNDER` on combos or dedicated symbol layers.

Actually, one could just combine `continue-list` with `/delete-property/ ignore-modifiers` to achieve the same as a more flexible `continue-mod-list`. E.g., suppose we want to continue caps-word on shift but cancel upon all other modifiers. Using this PR, we can do this using:
```
&caps_word {
    /delete-property/ ignore-modifiers;
    continue-list = <UNDERSCORE  BACKSPACE DELETE LSHFT RSHFT>;
};
```

urob:
Closing this as it is superseded by https://github.com/zmkfirmware/zmk/pull/1451

PR #1424, opened 2022-08-11
fix(docs): added concrete number for Bluetooth advertising name length

Hello9999901:


caksoylar:
That's more specific than what was there, but "~16" isn't exactly concrete, is it? 😃 Is there any way we can establish a definite number? There is an [assert for length 16](https://github.com/zmkfirmware/zmk/blob/4f503246a36319302309b1b6ff78f3a3cdc46cca/app/src/ble.c#L74) but it doesn't seem to be doing anything in practice. I haven't been able dig out anything from Zephyr docs either.

Hello9999901:
I agree. Pete said "16 (iirc)", so I'm not sure whether he rc'ed or not. Do you think that it's better to keep it until we dig something out? I can do some experimentation with it too (just keep adding characters until it doesn't work).

caksoylar:
Experimenting to establish the limit would certainly be useful, if you are willing take it on 😄

Hello9999901:
Great :) I'll shoot some hoops with all 3 OSes and see what I get. (hopefully I get the time to do this).

caksoylar:
Thanks! By the way we should add the constraint to the config documentation as well: https://zmk.dev/docs/config/system#general

Hello9999901:
I have done some testing on a local ZMK build setup with a nRF52840 dev board with a string of certain lengths in alphabetical order: "abcdefghijklm....". When the string reaches over 16 in length, an error arises as expected: 
```ERROR: BLE device name is too long. Math length: "16")```
However, I have tested with a string of length exactly 16 on:
MBP M1 2020 on macOS Ventura Beta 13.5
Razer Blade Stealth Late 2019 with Windows 11 21H2 and Pop_OS! 22.04 LTS
all of the above permutations work as intended.

caksoylar (approved):

petejohanson (approved):

PR #1426, opened 2022-08-12
feat(docs): Add testing related remarks to user setup page

caksoylar:
I added a few improvements to the end of user setup cautioning against common issues that people run into when they are first testing their keyboard. These are related to: (a) encouraging testing over USB before BT, (b) explaining peripherals shouldn't be tested before central for split keyboards, (c) linking to troubleshooting for pairing two halves in case of issues, (d) emphasizing BT profiles' importance a little more.

### DIFF
--- a/docs/docs/user-setup.md
+++ b/docs/docs/user-setup.md
@@ -193,13 +193,14 @@ To flash the firmware, first put your board into bootloader mode by double click
 or the one that is part of your keyboard). The controller should appear in your OS as a new USB storage device.
 
 Once this happens, copy the correct UF2 file (e.g. left or right if working on a split), and paste it onto the root of that USB mass
-storage device. Once the flash is complete, the controller should automatically restart, and load your newly flashed firmware.
+storage device. Once the flash is complete, the controller should automatically restart, and load your newly flashed firmware. It is
+recommended that you test your keyboard works over USB first to rule out hardware issues, before trying to connect to it wirelessly.
 
 ## Wirelessly Connecting Your Keyboard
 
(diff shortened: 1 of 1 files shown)


docs/docs/user-setup.md
```diff
@@ -193,14 +193,31 @@ To flash the firmware, first put your board into bootloader mode by double click
...
+over bluetooth. Peripheral half will only send keystrokes to the central once they are paired and connected. For this reason it is
+recommended to test the left half of a split keyboard first.
+
+:::
 
 ## Wirelessly Connecting Your Keyboard
 
-ZMK will automatically advertise itself as connectable if it is not currently connected to a device. You should be able to see your keyboard from the bluetooth scanning view of your laptop or phone / tablet. It is reported by some users that the connections with Android / iOS devices are generally smoother than with laptops, so if you have trouble connecting, you could try to connect from your phone or tablet first to eliminate any potential hardware issues.
+ZMK will automatically advertise itself as connectable if it is not currently connected to a device. You should be able to see your keyboard from the bluetooth scanning view of your computer or phone / tablet. It is reported by some users that the connections with Android / iOS devices are generally smoother than with laptops, so if you have trouble connecting, you could try to connect from your phone or tablet first to eliminate any potential hardware issues with bluetooth receivers.
```
Hello9999901, line 209: Really detailed stuff! Perhaps "your computer or phone / tablet" can be shortened to just "your device"?

Hello9999901 (changes requested):

docs/docs/user-setup.md
caksoylar: We could probably replace it here but this is the most beginner-friendly page of the docs, so maybe it doesn't hurt to be a bit specific/give examples.

In general I am scared people might think "device" refers to the keyboard, though probably not in this specific spot :)

Hello9999901 (approved):

BenRoe (approved):
Useful and beginner friendly explanation.
Looks good to me.

caksoylar:
Anyone else have any concerns or improvement suggestions?

petejohanson (approved):
LGTM!

PR #1429, opened 2022-08-16
add print to all functions

louiml-net:


louiml-net:
need it for test

app/src/activity.c
```diff
@@ -85,6 +85,7 @@ void activity_expiry_function() { k_work_submit(&activity_work); }
 K_TIMER_DEFINE(activity_timer, activity_expiry_function, NULL);
 
 int activity_init() {
+    printf("activity");
```
petejohanson, line 88: This should be using the logging subsystem, not raw `print` for this.

PR #1430, opened 2022-08-17
Factor out just the mouse key emulation from #778

DelusionalLogic:
This PR includes ONLY enough mouse emulation to enable mouse keys. It's a trimmed down version of #778.

I've left this as 2 commits, the first one is a squashed version of #778, the second removes everything that is not required for mouse keys.

It has been tested to work on a corne in usb mode. bindings on both the left and the right side work.

I would recommend squashing before merging, as having a commit that adds a bunch of stuff and then one that removes it again is sort odd, but it probably makes the review easier for people familiar with #778 already.


codeandr3w:
I tried this and it doesn't work for me: Windows shows the mouse HID device not loading correctly when attached by USB and I get nothing when connected via BLE. I tested with a nice!nano. The mousepr does work for me.

DelusionalLogic:
@codeandr3w That interesting. Do you get any error message or anything in the Windows EventLog? I'm running an expanded version of this PR on my crkbd and it's working in both windows and linux over USB with no issues.

app/include/dt-bindings/zmk/mouse.h
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+#pragma once
+
+/* Mouse press behavior */
+/* Left click */
+#define MB1 (0x01)
```
petejohanson, line 10: Why not alias proper HID usage defines here, like we do for keyboard page?

app/include/dt-bindings/zmk/mouse.h
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
petejohanson, line 2: Please update the copyright dates.

app/include/zmk/endpoints.h
```diff
@@ -13,3 +13,4 @@ int zmk_endpoints_toggle();
 enum zmk_endpoint zmk_endpoints_selected();
 
 int zmk_endpoints_send_report(uint16_t usage_page);
+int zmk_endpoints_send_mouse_report();
```
petejohanson, line 16: Why not just use the existing API and pass the button page id?

app/include/zmk/hid.h
```diff
@@ -89,6 +90,23 @@ static const uint8_t zmk_hid_report_desc[] = {
...
+
+    HID_USAGE_PAGE(HID_USAGE_GD),
+    HID_USAGE(HID_USAGE_GD_MOUSE),
+    HID_COLLECTION(HID_COLLECTION_APPLICATION),
+    HID_REPORT_ID(0x04),
+    HID_USAGE(HID_USAGE_GD_POINTER),
+    HID_COLLECTION(HID_COLLECTION_PHYSICAL),
+    HID_USAGE_PAGE(HID_USAGE_BUTTON),
+    HID_USAGE_MIN8(0x1),
+    HID_USAGE_MAX8(0x10),
```
petejohanson, line 102: Is 0x10 hard coded here arbitrary, or based on some spec?

app/src/behaviors/behavior_mouse_key_press.c
```diff
@@ -0,0 +1,48 @@
...
+                                      struct zmk_behavior_binding_event event) {
+    LOG_DBG("position %d keycode 0x%02X", event.position, binding->param1);
+    return ZMK_EVENT_RAISE(
+        zmk_mouse_button_state_changed_from_encoded(binding->param1, false, event.timestamp));
+}
+
+static const struct behavior_driver_api behavior_mouse_key_press_driver_api = {
+    .binding_pressed = on_keymap_binding_pressed, .binding_released = on_keymap_binding_released};
+
+#define KP_INST(n)                                                                                 \
```
petejohanson, line 41:
```suggestion
#define MKP_INST(n)                                                                                 \
```

petejohanson (changes requested):
A few minor items. Thanks for working on this!

aherriott:
Also was unable to get this pr working on windows 11. nice!nano_v2 connected over USB. Haven't tried the original PR yet, that will be next.

Here's the details of the error from device manager:

Error:
This device cannot start. (Code 10)

Driver:
Microsoft HID-Compliant-Mouse 10.0.22000.653

Event:
Device HID\VID_1D50&PID_615E&Col03\8&222c9c9c&1&0002 had a problem starting.

Driver Name: msmouse.inf
Class Guid: {4d36e96f-e325-11ce-bfc1-08002be10318}
Service: mouhid
Lower Filters: 
Upper Filters: 
Problem: 0xA
Problem Status: 0xC0110004

app/include/zmk/endpoints.h
DelusionalLogic: Unlike keyboards, mice don't have their own usage page in the HID spec. Instead mice is usage 0x02 inside the Generic Desktop (GD) usage page.

Using the generic usage page to signify sending a mouse report seems unintuitive to me, and using the actual usage (0x02) would collide with the "simulation" usage page.

I do agree that adding a separate API surface for mice is also not ideal, but unifying the two would, to my knowledge, require changes to the existing API.

DelusionalLogic: We could do that. I think it's maybe a little odd since we're not telling the api which device we want to send a report for (in case you also had a button subusage inside the keyboard), but there's no technical problem.

app/include/zmk/hid.h
DelusionalLogic: It's sorta arbitrary. It's the number of buttons, and 8 was chosen to fill up a word. There's no real rhyme or reason, and i suspect most systems won't even support all 8 buttons.

DelusionalLogic:
Firstly, Thanks for the review. I'm sorry it too so long to get back to this. I've asked some questions I think we should discuss before merging this, but I'll make the rest of the changes soon.

As for @aherriott. I'm not super familiar with how to debug this on windows, so I'll have to see if I can replicate. I daily a keyboard running this patch on windows without issue, but that keyboard also has additional unreleased changes. I'll try and revert to just this branch and see if i can replicate your issue.

If anyone has any idea of how to debug this, please do share. I'll happily fix it if I know what the issue is.

app/include/zmk/hid.h
petejohanson: That's fine, can we pull out the number of supported buttons into a define value to name it?

app/include/zmk/endpoints.h
petejohanson: Hmmm... Let me ponder that a bit more.

petejohanson: I'm having a hard time thinking of why we would have that subusage any time soon. If you can give me a compelling use case, I'm open to changing toward your proposed approach, otherwise I'd say we stick with the existing API/approach.

app/dts/behaviors/mouse_key_press.dtsi
```diff
@@ -0,0 +1,9 @@
+/ {
+	behaviors {
+		/omit-if-no-ref/ mkp: behavior_mouse_key_press {
```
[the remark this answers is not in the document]
petejohanson: Good point.. I'm more inclined to go with the press version though, since we keep the press/release terminology still.

petejohanson:
So, one thing I'm noticing here, is that the HID report/descriptor here includes the *buttons*, but no X/Y axis report information. I'm *wondering* if perhaps this trips up Windows' HID driver which expects a collection in this usage page to include those.

The original lump sum PR had those items in the report descriptor, so perhaps worth adding that and seeing how the Windows testing goes.

petejohanson:
I've just opened #1991 as a starting point so folks can test the fixes for Windows there, and will be iterating on that to try to address all the review feedback and get this in. Please test there!

petejohanson:
Closing now that #1991 is merged.

PR #1433, opened 2022-08-21
feat(led_strip): Add composite led_strip driver

xingrz:
Some boards use multiple LED strips. This driver composes them into one virtual device.

Tested on Keychron Q1, which I'll submit in another PR.


Anutrix:
@xingrz Can you please rebase?

PR #1438, opened 2022-08-28
RFC: Improve user setup process

joelspadin:
Changed the user setup documentation to use a new setup process where
users create the repo from the template repo, then run one of two new
tools to add a keyboard to it:

- In a web browser, using github.dev and the ZMK Tools extension.
- Locally, using a new Python setup script.

Also made various other improvements to the setup documentation, such
as including instructions for flashing keyboards that use DFU instead of
UF2.

## The documentation currently refers to a package which has not yet been published! 
It can be found at https://github.com/joelspadin/zmk-setup-script. If we decide to go ahead with switching to a Python setup script, it should probably be moved into a ZMK-owned repo.

Instead of running
```
pip3 install --user --upgrade zmk.setup
```
you can clone the repo and run 
```
pip3 install --user .
```
or simply add the repo directory to your `PYTHONPATH` variable, and then the `python3 -m zmk.setup` command will work.


AlexCzar:
It does look nice.
I think the shell script is the stronger option though.
The two main points for me:
* It doesn't require Python to be installed and properly configured (not universally a given, in my experience).
* With the shell script it is easier to make sure nothing harmful is being done to one's system: most people using command line have at least some familiarity with Bash.

joelspadin:
The main problem with a Bash script is that it doesn't work on Windows, so we have to maintain a second script as well.

AlexCzar:
Genuine question as I haven't used Windows for a long time now, is Python easier to install on Windows than git-bash, or wsl, or whatever is currently preferred there to get a minimal bash environment?

caksoylar:
> is Python easier to install on Windows than git-bash, or wsl, or whatever is currently preferred there to get a minimal bash environment?

Python is relatively easy to install on Windows these days because it is on Microsoft Store and it even pops up if you try to invoke `python` on powershell/cmd (though not `python script.py`, last time I tested).

Note that with the proposed approach, beginner users who would have trouble with Python (or command line in general) would be encouraged to use the Github-only workflow with Codespaces and the ZMK plugin.

For others (who are comfortable with the terminal):
1. Nearly all Linux distro users would have `python3` built-in
2. macOS seems to be [planning to remove `python3`](https://macmule.com/2022/01/29/macos-monterey-12-3-will-remove-python-2-7-usr-bin-python/) but most folks seem to be OK using `brew` so it shouldn't be very painful
3. For Windows users, see above (WSL is also available for power users)

Other than being able to maintain a single cross-platform script, Python approach also seems more future-proof to me since shell scripts don't (or shouldn't) scale up further. Joel's current version already has more functionality, such as repeatable calls to add multiple keyboards. [`qmk` CLI](https://pypi.org/project/qmk/) is a good reference here, we could add new shield tooling, wrap local building and flashing commands, integrate linters from the ZMK VSCode plugin, add [Zephyr shell support](https://github.com/caksoylar/zmk-posix-testbed#running-with-interactive-shell) etc.

Nick-Munnich:
Going to close this as obsolete thanks to `zmk-cli`

PR #1446, opened 2022-08-31
First draft Joystick/ADC sensor for comment

codeandr3w:
Added a first-draft PR to support joysticks or analogue sensors. This version as one ADC per ZMK sensor, so you need two "joystick" objects in your .overlay file for one physical joystick (horizontal and vertical).

It works by setting up an ADC channel and then polling it a user-configurable number of times a second. Currently, it works with the ec11 rotate behaviour, although it probably not the behaviour you might want. That's a job for another PR!

I have tested in on a nice!nano and an Edinburgh41 board with a joystick on it. Because the nice!nano has much fewer ADC pins than a pro micro, I had to do some bodge wiring to get the PCB's joystick pins to match the nice!nano's ADC pins.

You should add something like this to your .overlay:
```
    joystickx: joystickx {
        compatible = "joystick";
        label = "JOYSTICKX";
        io-channels = <&adc 0>;
        status = "okay";
        resolution = <512>;
        frequency = <50>;
        min_on = <256>;
        reverse;
    };
```

Add this to your .config file:
`    CONFIG_JOYSTICK=y`

I'm thinking that a "joystick" should probably be optionally 1 or 2 dimensions so that a single poll and a single sensor can map to one joystick. I'm also not sure whether to call it an ADC sensor or Analogue or Analog sensor instead of Joystick. And I'm not sure how to set it up in the best way to integrate with any future mouse feature.

And I don't know how to write tests for it other than to test it on my board.

### DIFF
--- a/app/drivers/sensor/CMakeLists.txt
+++ b/app/drivers/sensor/CMakeLists.txt
@@ -3,3 +3,4 @@
 
 add_subdirectory_ifdef(CONFIG_ZMK_BATTERY battery)
 add_subdirectory_ifdef(CONFIG_EC11 ec11)
+add_subdirectory_ifdef(CONFIG_JOYSTICK joystick)
--- a/app/drivers/sensor/Kconfig
+++ b/app/drivers/sensor/Kconfig
@@ -3,3 +3,4 @@
 
 rsource "battery/Kconfig"
 rsource "ec11/Kconfig"
+rsource "joystick/Kconfig"
--- a/app/drivers/sensor/joystick/CMakeLists.txt
+++ b/app/drivers/sensor/joystick/CMakeLists.txt
@@ -1,0 +1,8 @@
+# Copyright (c) 2020 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+zephyr_include_directories(.)
+
+zephyr_library()
+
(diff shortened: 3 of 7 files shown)


lzpfmh:
Nice job! looking forward to using it soon！

myst729:
I'm trying to make my old USB gamepad (PS2 like, with 2 analog joysticks) wireless. Not quite sure how to get this up and running. Any updates or working example? Thanks.

saitamandl:
Any updates on this PR?
Would be really great to be able to add an PSP style joystick 
Are the PR #1964 and #2176 related to this?

codeandr3w:
> Any updates on this PR?
> Would be really great to be able to add an PSP style joystick 
> Are the PR #1964 and #2176 related to this?

Sorry I haven't worked on this for a while and those two other PRs are by someone else. I got a lot more work and features done in my own repo. The problems I had were:
1. I couldn't integrate with the mouse PR because it wasn't upstream then and it's not designed for analog input
2. I didn't get any community engagement at the time to talk through what was needed to get the PR merged in
3. I couldn't work out how to add tests when there is no hardware available for people to test on

Maybe now there's more interest we should collaborate to get this feature in?

PR #1450, opened 2022-09-10
feat(docs): Fix ambigous tap-preferred documentation

urob:
Having misinterpreted the tap-preferred documentation myself and seeing other people on discord doing the same mistake, I suggest rewording it slightly. The current wording ("It triggers the tap behavior when another key is pressed") may falsely be interpreted as "tap-preferred" being the mirror of "hold-preferred": i.e., that it resolves all interrupts within the tapping-term as tap.

(As an aside, if one wants to implement the mirror of "hold-preferred", one can do so using positional hold taps with `hold-trigger-key-positions` set to a dummy key (setting it all empty is interpreted as all keys)).


caksoylar (approved):
Thank you for the quick update! The old wording was indeed unclear and the new wording is much more informative. I also double-checked [the code](https://github.com/zmkfirmware/zmk/blob/main/app/src/behaviors/behavior_hold_tap.c#L258) and I feel confident that it is a true statement. If any contributors disagree they can drop a note.

petejohanson (approved):

PR #1455, opened 2022-09-13
chore(lighting): fix deprecation

ReFil:
Removed deprecated macros, replaced them with new Zephyr 3 implementation, tested and working on real hardware


petejohanson (approved):
Thanks! That's been near the bottom of my list, appreciate the fix

PR #1457, opened 2022-09-13
feat(Underglow): Momentary effect selection

ReFil:
Adds the ability to have an RGB effect select whilst held down and revert when released. 

Designed to be used in conjunction with #1456  to allow for momentary battery indication through the lighting by holding down a specific key

### DIFF
--- a/app/include/dt-bindings/zmk/rgb.h
+++ b/app/include/dt-bindings/zmk/rgb.h
@@ -39,3 +39,8 @@
 #define RGB_COLOR_HSB_VAL(h, s, v) (((h) << 16) + ((s) << 8) + (v))
 #define RGB_COLOR_HSB(h, s, v) RGB_COLOR_HSB_CMD##(RGB_COLOR_HSB_VAL(h, s, v))
 #define RGB_COLOR_HSV RGB_COLOR_HSB
+
+#define RGB_EFF_SOLID 0
+#define RGB_EFF_BREATHE 1
+#define RGB_EFF_SPECTRUM 2
+#define RGB_EFF_SWIRL 3
--- a/app/include/dt-bindings/zmk/rgb.h
+++ b/app/include/dt-bindings/zmk/rgb.h
@@ -18,7 +18,8 @@
 #define RGB_EFF_CMD 11
 #define RGB_EFR_CMD 12
 #define RGB_EFS_CMD 13
-#define RGB_COLOR_HSB_CMD 14
+#define RGB_MEFS_CMD 14
+#define RGB_COLOR_HSB_CMD 15
 
 #define RGB_TOG RGB_TOG_CMD 0
 #define RGB_ON RGB_ON_CMD 0
@@ -33,6 +34,8 @@
 #define RGB_SPD RGB_SPD_CMD 0
 #define RGB_EFF RGB_EFF_CMD 0
 #define RGB_EFR RGB_EFR_CMD 0
+#define RGB_EFS RGB_EFS_CMD
+#define RGB_MEFS RGB_MEFS_CMD
 #define RGB_COLOR_HSB_VAL(h, s, v) (((h) << 16) + ((s) << 8) + (v))
 #define RGB_COLOR_HSB(h, s, v) RGB_COLOR_HSB_CMD##(RGB_COLOR_HSB_VAL(h, s, v))
 #define RGB_COLOR_HSV RGB_COLOR_HSB
--- a/app/src/behaviors/behavior_rgb_underglow.c
+++ b/app/src/behaviors/behavior_rgb_underglow.c
@@ -18,6 +18,8 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
 
+static uint8_t old_effect;
+
 #if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
 
 static const struct behavior_parameter_value_metadata no_arg_values[] = {
@@ -234,6 +236,9 @@ static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
         return zmk_rgb_underglow_change_spd(-1);
     case RGB_EFS_CMD:
         return zmk_rgb_underglow_select_effect(binding->param2);
+    case RGB_MEFS_CMD:
+        old_effect = zmk_rgb_underglow_calc_effect(0);
+        return zmk_rgb_underglow_select_effect(binding->param2);
     case RGB_EFF_CMD:
         return zmk_rgb_underglow_cycle_effect(1);
     case RGB_EFR_CMD:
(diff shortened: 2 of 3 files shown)


ReFil:
I've edited the documentation to note the new command as well as notes around an edge case when it's powered off with the key held down. Thinking about this I've realised the functionality can now be done using a parameterised macro too, however i think a dedicated behaviour command still has value. A parameterised macro would also present the same issue on the edge case when it's powered off with the key held down, I cant think of a good way to mitigate this (If anyone has ideas i'm happy to implement them)

ReFil:
I'll double check, a variant of this has been in use on the 360 Pro but the 360 pro doesn't use the standard locality implementation

docs/docs/behaviors/underglow.md
caksoylar, line 1: I think we might want to add defines for the [effects](https://github.com/zmkfirmware/zmk/blob/9e9b5f592de35c2b386d7e06c37b5088deff2a07/app/src/rgb_underglow.c#L47) in `dt-bindings/zmk/rgb.h` and document them here to be used with `EFS` and `MEFS`. Otherwise there doesn't seem to be a way for users can't know what index corresponds to what effect, without having to check the source code.

ReFil: Its documented [here](https://zmk.dev/docs/config/underglow) but you're right, it could be clearer, would it be better to do the defines, link to the documentation or both?

caksoylar: Good find, I hadn't noticed it in the config repo. I personally think it would be a good opportunity to add the defines while touching the underglow code, but if not I'd be fine with a link to the page from a docs perspective.

ReFil: I'll sort both out first thing tomorrow :)

ReFil: I've sorted it, i've also added a bit of an explanation as to what the effects actually mean

docs/docs/behaviors/underglow.md
```diff
@@ -61,6 +63,21 @@ They will also override the start values set by [`CONFIG_ZMK_RGB_*_START` settin
...
+:::note Effect Selection
+
+When using the `RGB_EFS` or `RGB_MEFS` definitions you must also include a number as an argument in the keymap corresponding to the effect you want to select e.g. `RGB_EFS 0`. There are 4 RGB effects currently, defined in [`dt-bindings/zmk/rgb.h`](https://github.com/zmkfirmware/zmk/blob/main/app/include/dt-bindings/zmk/rgb.h). They correspond to the options for [`CONFIG_ZMK_RGB_UNDERGLOW_EFF_START`](../config/underglow.md#kconfig).
+
+| Value            | Effect                                     |
+| ---------------- | ------------------------------------------ |
+| RGB_EFF_SOLID    | Solid color (Set by HSV)                   |
+| RGB_EFF_BREATHE  | Breathe a solid colour                     |
+| RGB_EFF_SPECTRUM | Cycle all LEDs through the colour spectrum |
+| RGB_EFF_SWIRL    | Swirl a rainbow around the LEDs            |
```
caksoylar, line 75:
```suggestion
| Value            | Effect                                    |
| ---------------- | ----------------------------------------- |
| RGB_EFF_SOLID    | Solid color (set by HSB)                  |
| RGB_EFF_BREATHE  | Breathe a solid color                     |
| RGB_EFF_SPECTRUM | Cycle all LEDs through the color spectrum |
| RGB_EFF_SWIRL    | Swirl a rainbow around the LEDs           |
```

docs/docs/behaviors/underglow.md
```diff
@@ -61,6 +63,21 @@ They will also override the start values set by [`CONFIG_ZMK_RGB_*_START` settin
...
+
+If the `RGB_MEFS` key is held down for longer than [`CONFIG_ZMK_SETTINGS_SAVE_DEBOUNCE`](../config/system.md#general) milliseconds and the board is reset prior to releasing the key the temporary effect will have been saved to flash memory and will be restored after resetting/power cycling
```
caksoylar, line 77: Is this worth mentioning? Seems unlikely for someone to hit that given the typical config value is 60 seconds.

docs/docs/behaviors/underglow.md
```diff
@@ -75,6 +92,12 @@ However the settings will only be saved after [`CONFIG_ZMK_SETTINGS_SAVE_DEBOUNC
    &rgb_ug RGB_COLOR_HSB(128,100,100)
    ```
 
+1. Select a specific RGB effect (Swirl)
+
+   ```dts
+   &rgb_ug RGB_EFS 3
```
caksoylar, line 98:
```suggestion
   &rgb_ug RGB_EFS RGB_EFF_SWIRL
```

docs/docs/behaviors/underglow.md
ReFil: I've had this issue with 360 Pro users and so figured it was worth mentioning

caksoylar:
```suggestion
If the `RGB_MEFS` key is held down for longer than [`CONFIG_ZMK_SETTINGS_SAVE_DEBOUNCE`](../config/system.md#general) milliseconds and the board is reset prior to releasing the key, the temporary effect will have been saved to flash memory and will be restored after resetting/power cycling.
```
Perhaps put it in a `:::warning` box as well?

ReFil: Good idea, should be done now

docs/docs/behaviors/underglow.md
```diff
@@ -61,6 +63,25 @@ They will also override the start values set by [`CONFIG_ZMK_RGB_*_START` settin
...
+:::note Effect Selection
+
+When using the `RGB_EFS` or `RGB_MEFS` definitions you must also include a parameter corresponding to the effect you want to select, e.g. `&rgb_ug RGB_EFS RGB_EFF_SOLID`. There are currently 4 RGB effects, defined in [`dt-bindings/zmk/rgb.h`](https://github.com/zmkfirmware/zmk/blob/main/app/include/dt-bindings/zmk/rgb.h):
+
+| Value            | Effect                                    |
+| ---------------- | ----------------------------------------- |
+| RGB_EFF_SOLID    | Solid color (set by HSB)                  |
+| RGB_EFF_BREATHE  | Breathe a solid color                     |
+| RGB_EFF_SPECTRUM | Cycle all LEDs through the color spectrum |
+| RGB_EFF_SWIRL    | Swirl a rainbow around the LEDs           |
```
caksoylar, line 75:
```suggestion
| ------------------ | ----------------------------------------- |
| `RGB_EFF_SOLID`    | Solid color (set by HSB)                  |
| `RGB_EFF_BREATHE`  | Breathe a solid color                     |
| `RGB_EFF_SPECTRUM` | Cycle all LEDs through the color spectrum |
| `RGB_EFF_SWIRL`    | Swirl a rainbow around the LEDs           |
```

caksoylar (approved):
LGTM, small nit on the formatting in the table.

ReFil:
Studio metadata has now been added. I think I've done it right but would appreciate someone more knowledgeable on the metadata looking over it

ReFil:
I believe this PR is ready for merging regardless but will rebase it in due course

ReFil:

ReFil:
Is there any appetite for this to be merged in? If so i'll spend some time rebasing it and getting it up to the latest

nmunnich:
Very unlikely atm I think, I don't think any lighting-related PRs will get much attention, combined with this being possible via macro....

ReFil:
Ah I see, fair enough. I think there's a simplicity aspect to doing it in a dedicated behaviour, especially for hardware integrators (it's a lot easier to tell end users to use the momentary effect behaviour than instantiate a funky macro) but if lighting subsystem isn't getting stuff merged lately that's fair

Issue #885, opened 2021-07-28
Underglow overlays for nice_nano don't work on nice_nano_v2


Nicell:
The existing `nice_nano.overlay/conf`s for shield underglow don't apply to the `nice_nano_v2`.

We might want to instead make general spi definitions in the shield DT for `&pro_micro` pins.

malinges:
Zephyr seems to support board revisions, maybe that could help?
- https://docs.zephyrproject.org/latest/application/index.html#building-for-a-board-revision
- https://github.com/zephyrproject-rtos/zephyr/pull/29990

malinges:
I just got it to work, it's pretty easy actually once you figure out the filename conventions: malinges/zmk@b43ada39adf4400fb424d303aabb9539afbb5d40
You can then build for a nice!nano v2 with `west build [-p] -b nice_nano@2 [...]`.
And last but not least, RGB underglow actually works with the resulting firmware (tested on a reviug41).
I'd be happy to submit a PR if you think it's a good idea.

Nicell:
@malinges board revisions seem neat! I'm a bit worried about backwards compatibility however with how we currently have them set up. I'm not sure if it's possible to redirect a board name to another board+revision.

malinges:
@Nicell agreed, the main issue with this approach is that it's a breaking change. Unless it's possible to create some kind of board alias, but I don't have enough experience with ZMK/Zephyr to know if it's even possible (and a quick Google search doesn't come up with anything promising).

But yes, I quickly chatted about this being a breaking change on Discord (#boards-shields) and a few people seemed (legitimately) concerned about it too. Although maybe it's less of an issue for the nice!nano v2 since it's only been supported by ZMK for around 1.5 month? ;-)

As a last resort, do you think it could be possible to duplicate the nano v2's definitions for a given deprecation period, then get rid of the old defs once the deprecation period "expires"? Unfortunately I'm not familiar enough with ZMK's development process to know what would be a satisfying solution here.

malinges:
Interestingly, Zephyr has both a [board alias system and a board deprecation system](https://github.com/zephyrproject-rtos/zephyr/blob/8c31f31143c681045e4fc27e17b9a972c3b82b70/cmake/app/boilerplate.cmake#L228-L241).

[...]

PR #1461, opened 2022-09-21
Update info related to the issue with n!n 2 and underglow

Xoffio:
#885
I think it is important to add this info related to the issue with nice nano 2 and the underglow to avoid confusion.


admilsonmarques (approved):

caksoylar:
Thanks @lesshonor, I believe Zephyr 3.2 update added all the v2 overlays for shields that already had the v1 ones, so we can close this.

PR #1462, opened 2022-09-21
feature(shields): Add nice!view

Nicell:
## Board/Shield Check-list


caksoylar:
Is this definition intended to be used along with other shields using a build definition with multiple shields? If so, does the Github actions workflow support such a definition through `build.yaml`?

Is this also supposed to be compatible with nice!nano only (or with `pro micro` boards, for that matter)? I don't see why that should be, but it seems like the case from the SPI node alias.

Nicell:
> Is this definition intended to be used along with other shields using a build definition with multiple shields? If so, does the Github actions workflow support such a definition through `build.yaml`?

Exactly! The GitHub actions workflow should work fine according to Pete... I need to test it though. Should just need to update the build.yaml shield prop to have `nice_view` after the existing one.

> Is this also supposed to be compatible with nice!nano only (or with `pro micro` boards, for that matter)? I don't see why that should be, but it seems like the case from the SPI node alias.

Should be compatible with all the other Pro Micro boards, I just need to define SPI nodes/aliases for others, which is on my todo.

app/boards/shields/nice_view/nice_view.conf
```diff
@@ -0,0 +1,5 @@
+# Enable nice!view
+CONFIG_ZMK_DISPLAY=y
+CONFIG_SSD1306=n
```
petejohanson, line 3: Why was this necessary? We used to have a bug or two with some random shield having conditionals wrong, causing this to auto-enable, but I don't think we still do?

app/boards/shields/nice_view/nice_view.overlay
```diff
@@ -0,0 +1,29 @@
...
+		compatible = "sharp,ls0xx";
+		label = "DISPLAY";
+		spi-max-frequency = <1000000>;
+		reg = <0>;
+		width = <160>;
+		height = <68>;
+	};
+};
+
+&pro_micro_i2c {
```
petejohanson, line 21: This ties the shield to only working with PM based boards. Given XIAO gaining popularity, I wonder if this should be in `shields/nice_view/boards/*.overlay` files for all the existing PM boards?

app/boards/shields/nice_view/nice_view.zmk.yml
```diff
@@ -0,0 +1,8 @@
+file_format: "1"
+id: nice_view
+name: nice!view
+type: shield
+url: https://nicekeyboards.com/nice-view
+requires: [pro_micro, i2c_oled]
```
petejohanson, line 6: We should consider this, this is an odd requirement, if we want this shield to be a bit more generic.

app/boards/shields/nice_view/nice_view.overlay
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+&nice_view_spi {
+	status = "okay";
+	cs-gpios = <&pro_micro 1 GPIO_ACTIVE_HIGH>;
```
petejohanson, line 9: I think this should be something like:

```suggestion
	cs-gpios = <&nice_view_d 0 GPIO_ACTIVE_HIGH>;
```

And then expect a `nice_view_d` nexus node to be supplied? Possibly with some helper macro to make it not so painful to write every time? Or with an include so you can just write:

````
&nice_view_d {
  gpio-map = <0 0 &pro_micro 1 0>;
};
```

As needed to add the mapping.

app/boards/shields/nice_view/nice_view.overlay
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+&nice_view_spi {
```
petejohanson, line 7: Should we provide this in `shields/nice_view/boards/*.overlay` files for common/in-tree boards?

app/boards/arm/nice_nano/nice_nano.dtsi
```diff
@@ -63,6 +70,7 @@
 	};
 };
 
+nice_view_spi: &spi0 {};
```
petejohanson, line 73: Move this to the boards override directory in the shield itself.

petejohanson (commented):
Some initial thoughts!

app/boards/shields/nice_view/nice_view.conf
Nicell: This was necessary as setting the `default` to `n` in the defconfig did not actually disable it since it's turned on in the other shield when `ZMK_DISPLAY` is enabled.

Nicell:
Our requires/exposes system doesn't really handle this new type of interconnect (built on top of a complete keyboard, not to create one). We'll need to figure out how we handle this with our build system and our hardware docs page. I've set up stubs and done nothing with them for now for both the build system and docs.

app/boards/shields/nice_view/Kconfig.defconfig
```diff
@@ -0,0 +1,26 @@
...
+
+config LS0XX
+	default y
+
+if !ZMK_SPLIT || ZMK_SPLIT_ROLE_CENTRAL
+
+config ZMK_WIDGET_WPM_STATUS
+	default y
+
+endif # !ZMK_SPLIT || ZMK_SPLIT_ROLE_CENTRAL
```
petejohanson, line 22: Can't this be combined to a single default with condition?


```suggestion
config ZMK_WIDGET_WPM_STATUS
	default y if !ZMK_SPLIT || ZMK_SPLIT_ROLE_CENTRAL
```

app/boards/shields/nice_view/nice_view.zmk.yml
```diff
@@ -0,0 +1,8 @@
+file_format: "1"
+id: nice_view
+name: nice!view
+type: shield
+url: https://nicekeyboards.com/nice-view
+requires: [nice_view]
```
petejohanson, line 6:
```suggestion
requires: [nice_view_header]
```

?

app/boards/shields/nice_view_oled/Kconfig.shield
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) 2022 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+config SHIELD_NICE_VIEW_OLED
+	def_bool $(shields_list_contains,nice_view_oled)
```
petejohanson, line 5: As I think about this... Still unsure about the name.

OLED...

- adapter?
- bodge
- converter

Some alternative ideas.

app/boards/shields/nice_view_oled/boards/nice_nano.overlay
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+nice_view_spi: &spi0 {
+	compatible = "nordic,nrf-spim";
+	status = "okay";
+	sck-pin = <20>;
+	mosi-pin = <17>;
+	miso-pin = <25>;
+	cs-gpios = <&pro_micro 1 GPIO_ACTIVE_HIGH>;
```
petejohanson, line 13: I think omit this entirely? Have folks override the `cs-gpios` property in their keymap for whatever pin they bodged.

petejohanson (changes requested):
A few follow up thoughts.

I think each shield deserves a readme file as well to explain their use, lacking tooling for this in some setup script.

app/boards/shields/nice_view_oled/boards/nice_nano.overlay
Nicell: It'd be nice to allow for this to be a default fallback. I'll likely put in the docs for the nice!view to use this pin if available. (it should be for most since it's on the trrs serial pin)

petejohanson: As long as this is clearly documented there, and in the README, that's fine.

petejohanson (changes requested):
I think last thing: can you please add board overrides for the other in-tree nRF52 boards, e.g. nRFMicro, etc?

app/boards/shields/nice_view_adapter/README.md
```diff
@@ -0,0 +1,11 @@
+# nice!view Adapter
+
+This shield is used as an adapter between the nice!view and existing shields/boards that expose an I2C OLED header.
+
+To use this shield, you should add this shield to your list of shields *before* `nice_view`.
+
+The nice!view will use the SDA/SCL pins of the OLED, and then the adapter expects a final pin to be "bodged" from your microcontroller to the nice!view CS pin. This adapter assumes that the CS pin bodged is the `&pro_micro 1 pin` or "D1", which is the top left pin when looking at the front of the board. If you can't use this pin, you'll need to write your own definition for the `&nice_view_spi` bus.
```
caksoylar, line 7:
```suggestion
The nice!view will use the SDA/SCL pins of the header, and then the adapter expects a final pin to be "bodged" from your microcontroller to the nice!view CS pin. This adapter assumes that the CS pin bodged is the `&pro_micro 1 pin` or "D1", which is the top left pin when looking at the front of the board. If you can't use this pin, you'll need to override the `cs-gpios` definition for the `&nice_view_spi` bus.
```
The first sentence might be confusing because there is actually no OLED display involved here. Also thoughts on making it more specific as such for the last sentence?

Nicell: Agreed on the first sentence. Second sentence... I think the override is only possible on zmk-config keymaps? The reason being is we need the override to be the last DT override, and I think currently that's only possible with the zmk-config if I'm not mistaken.

caksoylar: I might be misunderstanding, but I believe you always need it to be defined last, right? Even if you create a new `nice_view_spi` node instead of overriding with `&nice_view_spi`, it still needs to be defined later for it to take effect instead of the node definition coming from board-specific overlays.

caksoylar (approved):

app/boards/shields/nice_view_adapter/nice_view_adapter.zmk.yml
```diff
@@ -0,0 +1,9 @@
+file_format: "1"
+id: nice_view_adapter
+name: nice!view
+type: shield
+url: https://nicekeyboards.com/nice-view
+requires: [i2c_oled]
+exposes: [nice_view_header]
+features:
+  - display
```
petejohanson, line 9: I think this should be removed. The adapter itself doesn't add a display.

petejohanson (changes requested):
Sorry, one last thing I noticed after a re-read of this.

app/boards/shields/nice_view_adapter/nice_view_adapter.zmk.yml
```diff
@@ -0,0 +1,7 @@
+file_format: "1"
+id: nice_view_adapter
+name: nice!view
```
petejohanson, line 3:
```suggestion
name: nice!view adapter
```

petejohanson (changes requested):

petejohanson (approved):
Thanks!

PR #1464, opened 2022-09-22
fix(displays): Adjust alignment and spacing

Nicell:
This fixes and improves a few things. I think a couple of these are subjective.

- Fix right aligned widgets (battery and WPM) by realigning every time the text is updated (maybe `lv_realign` could be used? Didn't work when I tried)
- Add spacing between text elements to make things less cramped
- Add one to the profile index, which is better understood in the non-programming world

### DIFF
--- a/app/src/display/widgets/battery_status.c
+++ b/app/src/display/widgets/battery_status.c
@@ -27,13 +27,13 @@ struct battery_status_state {
 };
 
 static void set_battery_symbol(lv_obj_t *label, struct battery_status_state state) {
-    char text[2] = "  ";
+    char text[3] = {};
 
     uint8_t level = state.level;
 
 #if IS_ENABLED(CONFIG_USB_DEVICE_STACK)
     if (state.usb_present) {
-        strcpy(text, LV_SYMBOL_CHARGE);
+        strcpy(text, LV_SYMBOL_CHARGE " ");
     }
 #endif /* IS_ENABLED(CONFIG_USB_DEVICE_STACK) */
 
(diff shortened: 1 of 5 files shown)


app/src/display/widgets/battery_status.c
```diff
@@ -27,13 +27,13 @@ struct battery_status_state {
 };
 
 static void set_battery_symbol(lv_obj_t *label, struct battery_status_state state) {
-    char text[2] = "  ";
+    char text[3] = {};
```
caksoylar, line 30: Might as well change this to prevent the buffer overflow #1320. This issue [causes segfaults](https://github.com/caksoylar/zmk-posix-testbed#build-and-test) with the `native_posix_64`.

Nicell: Good idea! Fixed this along with adding some space to the layer ones

app/src/display/widgets/output_status.c
```diff
@@ -39,24 +39,24 @@ static struct output_status_state get_state(const zmk_event_t *_eh) {
...
-        strcat(text, LV_SYMBOL_USB "   ");
+        strcat(text, LV_SYMBOL_USB);
         break;
     case ZMK_ENDPOINT_BLE:
         if (state.active_profile_bonded) {
             if (state.active_profile_connected) {
-                snprintf(text, sizeof(text), LV_SYMBOL_WIFI "%i " LV_SYMBOL_OK,
-                         state.active_profile_index);
+                snprintf(text, sizeof(text), LV_SYMBOL_WIFI " %i " LV_SYMBOL_OK,
+                         state.active_profile_index + 1);
```
petejohanson, line 52: I'm all for this... but definitely risks confusion because of the indexing discrepancy between keymaps and the display... We do already not the indexes are zero based in the bluetooth docs... I suppose we just do it?

Once we have studio, we can hide it behind a UI anyways...

petejohanson (commented):
All seems reasonable... Will try to test this on hardware here this week.

app/src/display/widgets/output_status.c
Nicell: That is a fair point, but I think 0-indexing is relatively uncommon and strange in a user-facing application like this. As you said, we note it, so hopefully, we shouldn't have too many issues. Studio will also abstract all that away too!

petejohanson (approved):
Thanks!

PR #1065, opened 2021-12-19
Added nRF52840 Feather Express support and MF68 shield

exentio:
## Board/Shield Check-list
 - [X] This board/shield is tested working on real hardware
   - Partially: only the nRF52840 Feather Express configuration
 - [X] Definitions follow the general style of other shields/boards upstream ([Reference](https://zmk.dev/docs/development/new-shield))
 - [X] `.zmk.yml` metadata file added
 - [X] Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)
 - [X] General consistent formatting of DeviceTree files
 - [X] Keymaps do not use deprecated key defines (Check using the [upgrader tool](https://zmk.dev/docs/codes/keymap-upgrader))
 - [X] `&pro_micro` used in favor of `&pro_micro_d/a` if applicable
 - [ ] If split, no name added for the right/peripheral half
 - [X] Kconfig.defconfig file correctly wraps *all* configuration in conditional on the shield symbol
 - [X] `.conf` file has optional extra features commented out
 - [X] Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)

PR #1465, opened 2022-09-22
feat(board): Add Adafruit Feather NRF52840 board

tmbutcher:
## Board/Shield Check-list

Since #1065 seems to have stalled out, I thought I'd create a new PR to add support for this board. I've been using these files for more than a year now with no problem. This is my first PR, though, so apologies if I'm missing anything that's supposed to be in here.


petejohanson:
Did you see https://github.com/zmkfirmware/zmk/pull/1065#issuecomment-1017121275 where I suggested a better approach to avoid duplicating the existing Zephyr board definition?

tmbutcher:
Ah, yes, I think I saw that, but I probably misunderstood and thought that it related to the issue of splitting the board and shield into different PRs.

In theory, I'm happy to revise this PR as necessary, but in reality I worry that the better approach that you want exceeds my understanding of what should be in the conf and overlay files vs. what doesn't need to be there. I've taken a look at the conf and overlay files that are currently in the boards directory, but that left me with a bunch of questions. I suspect that the issue I'm running into here is that my knowledge of core make files is lacking. I'll take a look at those and see if I can iron out my questions.

caksoylar:
The same comment here https://github.com/zmkfirmware/zmk/pull/1065#issuecomment-2692417001 applies to this one as well. An overlay+config would be a welcome contribution.

Issue #1476, opened 2022-10-01
Update documentation to confirm F13-F19 works with Windows


jeep:
I am able to confirm that F13-F19 work with Windows 11. The application I use to test only recognizes F1-F19, so I cannot test F20-F24.

Please update https://zmk.dev/docs/codes#f-keys to indicate working on Windows.

(I was going to submit a PR, but got lost in the web of imports that generates the table, sorry.)

caksoylar:
If you are still willing to do it yourself, all you need to change are the `os -> windows` values in `hid.js`, like this one here: https://github.com/zmkfirmware/zmk/blob/main/docs/src/data/hid.js#L2735

You can also check out #598 as an example.

PR #1477, opened 2022-10-01
docs(codes): Windows Support: F13-F19

jeep:
F13-F19 are confirmed to work on Windows 11 in VSCode 1.71.2.

Refs: #1476

### DIFF
--- a/docs/src/data/hid.js
+++ b/docs/src/data/hid.js
@@ -2879,7 +2879,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=86",
     os: {
-      windows: null,
+      windows: true,
       linux: true,
       android: false,
       macos: true,
@@ -2900,7 +2900,7 @@ export default [
     ],
     documentation: "https://usb.org/sites/default/files/hut1_2.pdf#page=86",
     os: {
-      windows: null,
+      windows: true,
       linux: true,
       android: false,
       macos: true,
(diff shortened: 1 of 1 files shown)


caksoylar:
Do you mind also testing F20-F24 while you are at it? I just tested all F13-F24 myself and they all get detected on a key tester like https://config.qmk.fm/#/test.

jeep:
VS Code only recognized F1-F19 so I only added those. If a keytester recognizes them, then it seems safe enough, right? If you feel like you need a second person to verify, I suppose I could do the same thing- try a test site.

caksoylar:
Looks like maybe it is a VS code issue? https://github.com/microsoft/vscode/issues/89166

I think seeing it on the key tester is fine, if you could give it a shot and update the docs if it works it would be great.

claviger-pc:
My experience was that when NKRO was enabled, all of the F keys worked. When it was not, then only up to F12 functions. I tested this on windows.

jeep:
> Looks like maybe it is a VS code issue? [microsoft/vscode#89166](https://github.com/microsoft/vscode/issues/89166)
I don't understand... it's not an issue... it's the design. The s/w only supports F1-F19.

I'm happy to change the document...

caksoylar:
> My experience was that when NKRO was enabled, all of the F keys worked. When it was not, then only up to F12 functions. I tested this on windows.

That's interesting, I double checked that I am on default settings (HKRO with report size 6, full consumer usage page) and it works for me on Windows 11 over both USB and BLE. Can you check again on Windows if you get a chance?

claviger-pc:
Yes, sorry, now it does work, including those above F20. I am not sure whether this was my error when I tried it before or whether some other change was made that corrected a problem. Built locally, based on this config, where most is default, including HRKO with 6 keys, like yours: https://github.com/claviger-pc/zmk-config

petejohanson (approved):
Thanks!

PR #1479, opened 2022-10-05
feat(behaviors): caps-word supports alternative layouts

### LINKED ISSUE #1470: Feature request: caps-word can modify some non-alpha characters
Caps-word will currently only capitalize alpha characters. It would be useful to be able to provide it a list of non-alpha characters that it would also apply the shift modifier to. *Why would you ever want this?* I use the Colemak layout, but I do the mapping in my OS rather than the firmware. Therefore the `;` scancode is sent to the host when I type the letter `o`.

I can add `SEMICOLON` to `continue-list`, which prevents caps-lock from deactivating, but since semicolon is not an alpha, this produces a lowercase "o" in the middle of the caps word, for example

**Desired:** "SEMICOLON"

**Actual:** "SEMICoLoN"

I'm not sure what you'd name this option, but something  like the following would solve my problem:

```
&caps_word {
    continue-list = <UNDERSCORE BACKSPACE DELETE SEMI>;
    also-cap = <SEMI>;
};
```

rschenk:
The current Caps-Word feature does not support using alternative layouts (Colemak, Dvorak, etc) that are mapped in the OS, or when using a foreign keyboard. In these cases, there will be alpha characters that the keyboard "thinks" are symbols, and therefore even if you add them to `continue-list`, caps-word will not shift them. This PR will fix #1470

This PR adds the ability to configure caps-word to shift these keycodes so that it works as expected.

It accomplishes this by adding two optional config params:
* `also-mod-list` contains a list of (non-alpha) keycodes that caps-word should also apply the shift modifier to.
* `break-list` contains a list of (alpha) keycodes that should deactivate caps-word.

This is perhaps best illustrated through an example (which is in my docs):

With Colemak mapped in the OS, when you type "o", the keyboard thinks you are pressing `SEMICOLON`. For caps word to produce a capital "O" we need to configure caps word to both remain active _and apply the shift modifier_ when it sees `SEMICOLON`. We can use `continue-list` to tell caps-word to continue when it sees `SEMI` and `also-mod-list` to tell caps word to shift it, even though it's not an alpha.

Likewise in OS-mapped Colemak when you type ";" the keyboard thinks you are pressing `P`. If caps word is active and you type ";" we need to tell caps word to deactivate even though it sees the alpha keycode `P`. Adding `P` to `break-list` accomplishes this.

In my docs, I included sample configurations for Colemak, Dvorak, and Workman (all of which I tested on hardware).

Likewise I wrote unit tests for both cases. I had to do quite a bit of refactoring of the caps-word configuration struct to accomplish this, because it now needs to accept three variable-length arrays instead of just one. Hopefully I didn't churn the code up too much.

I've been using this on my personal os-mapped Colemak board and it's WORKING (see what I did there?)

### DIFF
--- a/app/dts/behaviors/caps_word.dtsi
+++ b/app/dts/behaviors/caps_word.dtsi
@@ -13,6 +13,8 @@
 			label = "CAPS_WORD";
 			#binding-cells = <0>;
 			continue-list = <UNDERSCORE BACKSPACE DELETE>;
+			also-mod-list = <>;
+			break-list = <>;
 		};
 	};
 };
--- a/app/dts/bindings/behaviors/zmk,behavior-caps-word.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-caps-word.yaml
@@ -13,3 +13,11 @@ properties:
     required: true
   mods:
     type: int
+  also-mod-list:
+    type:  array
+    required: false
+    default: []
+  break-list:
+    type:  array
+    required: false
+    default: []
--- a/app/src/behaviors/behavior_caps_word.c
+++ b/app/src/behaviors/behavior_caps_word.c
@@ -30,11 +30,17 @@ struct caps_word_continue_item {
     uint8_t implicit_modifiers;
 };
 
+struct caps_word_item_array {
+    uint8_t length;
+    struct caps_word_continue_item members[];
+};
+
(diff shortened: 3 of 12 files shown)


rschenk:
Closing in favor of #1485 which is a better way of accomplishing the same goal

PR #1480, opened 2022-10-05
docs(power profiling): add nice!view

Nicell:
Adds the nice!view based on my profiling using a Nordic PPK2. Idle is lower than 1uA, but I wrote this originally to use uA as the base unit, so 1 will do.

### DIFF
--- a/docs/src/data/power.js
+++ b/docs/src/data/power.js
@@ -96,4 +96,10 @@ export const displayPower = {
     active: 10000, // Estimated power draw when about half the pixels are on
     sleep: 7, // Deep sleep power draw (display off)
   },
+  // Based on the nice!view using Sharp's LS011B7DH01
+  NICEVIEW: {
+    activePercent: 0.01, // Estimated two refreshes per second taking five milliseconds each
+    active: 225, // Power draw during refresh
+    sleep: 1, // Idle power draw of the display
+  },
 };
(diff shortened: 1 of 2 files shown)


docs/src/data/power.js
```diff
@@ -96,4 +96,10 @@ export const displayPower = {
     active: 10000, // Estimated power draw when about half the pixels are on
     sleep: 7, // Deep sleep power draw (display off)
   },
+  // Based on the nice!view using Sharp's LS011B7DH01
+  NICEVIEW: {
+    activePercent: 0.01, // Estimated two refreshes per second taking five milliseconds each
```
caksoylar, line 101: Why 2 refresh/s?

caksoylar (approved):
One question on the underlying assumptions below.

If you have a way to, maybe you can average out the power usage (with the default widgets so e.g. layer changes refresh the display) over a usage period (15 minutes?) and see if this is consistent with the refresh rate/estimated power usage?

Docs render looks good, adding `nice!view` barely budges the estimate numbers which I guess is expected.

docs/src/data/power.js
Nicell: To be honest, there's not an exact reason this was chosen. The ePaper estimate only had one change per minute taking three seconds, which maybe was a bit low for normal keyboard use. The view can refresh very quickly (5ms), so the active current is only that high for that short period. Because of this, if I had only done one change per minute, the percentage would've been really low. I think an active percent of 1% is applicable for the view and maybe even overestimates how much power it uses.

Nicell:
> If you have a way to, maybe you can average out the power usage (with the default widgets so e.g. layer changes refresh the display) over a usage period (15 minutes?) and see if this is consistent with the refresh rate/estimated power usage?

This would be interesting to try. The biggest hurdle for testing this is controlling other power sucks like typing. I'll have to think if this is possible 🤔

petejohanson (approved):
This WFM!

PR #1481, opened 2022-10-05
feat(ble): Default a few configs for DIS GATT svc.

petejohanson:
Default values for DIS GATT characteristics for:

* Vendor ID
* Product ID
* Manufacturer
* Model

This should help our ZMK keebs show properly in more places on various hosts, and hopefully help with identification for programs like Karabiner Elements on macOS.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -25,6 +25,18 @@ config USB_DEVICE_PID
 config USB_DEVICE_MANUFACTURER
 	default "ZMK Project"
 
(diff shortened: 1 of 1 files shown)


Nicell (commented):
Changes make sense after reading up on the DIS, but I'm wondering where `BT_DIS` is enabled since it defaults to off from what I'm reading.

Nicell (approved):
@petejohanson oops, missed that, LGTM!

Issue #1478, opened 2022-10-02
&caps_word gets deactivated when underscore triggered with right shift and minus


xudongzheng:
Underscore is correctly handled when using left shift with minus, however right shift with minus seems to deactivate &caps_word.

Kickeskov:
Okay I took a quick look at this.

Underlying issue: The code in `src/behaviors/behavior_caps_word.c` compares the pressed modifiers with the configured ones, matching hard on left/right. The definition for `UNDERSCORE` is defined by left shift, thus pressing RS(MINUS) does not satisfy the condition.

Quick local fix:
Change the "Continue List" in the [configuration](https://zmk.dev/docs/behaviors/caps-word) to be `<UNDERSCORE BACKSPACE DELETE (RS(ZMK_HID_USAGE(HID_USAGE_KEY, HID_USAGE_KEY_KEYBOARD_MINUS_AND_UNDERSCORE))) >`. I haven't tested it though.

Possible solutions:
1: Expand the default config to include `(RS(ZMK_HID_USAGE(HID_USAGE_KEY, HID_USAGE_KEY_KEYBOARD_MINUS_AND_UNDERSCORE)))`. (`dts/behaviors/caps_word.dtsi`)
Cons: Not pretty. Hacky. It can be added as a definition in `keys.h` but that's backwards.

2: Make the code not care about modifier sides. This can be done by doing some bitshift magic (<<4 and >>4) on the implicit/explicit modifiers in the `caps_word_is_caps_includelist` function.
Cons: There might be an edge case where this causes undesired behavior, particularly with LALT/RALT.

2b: Make the code not care about modifier sides, but only with SHIFT.
Cons: Edge case fix, where it might be better to make a general case fix by including other modifiers.

I think 2b might be the best approach. It may be possible to write the code in such a way that the bitshift magic does not seem entirely arcane, might need to introduce a helper function for this.

In order to implement the solution I need some insight from someone who knows more than me:
- Is 2b a good approach? Is it better to include all mods?
- In case the bitshift magic gets a bit too hefty and a helper function is needed, would the best place to put this be in `src/behaviors/behavior_caps_word.c`, or is there a designated place for those functions? "unified shift"?

basnijholt:
I just spend an hour scratching my head why I couldn't get this to work. Would it be worth mentioning this bug in the docs to prevent others from running into the same issue?

caksoylar:
[...]

PR #1485, opened 2022-10-07
Capslock-based `caps_word`

snoyer:
According to the documentation:

> The caps word behavior behaves similar to a caps lock, but will automatically deactivate [...]

This implementation of `caps_word` uses the capslock mechanism instead of adding modifiers to key presses.
It will literally send `CAPSLOCK` when activated and send it again at the end of the word (based on configurable word separator key list) to disable it (unless capslock has already been disabled in the meantime).

This should provide a simpler solution to the basic problem of writing words in caps, regardless of the OS layout (see #1470, #1479, #1478).

Some of the current functionality (such as using alternative mods other than shift) does get dropped/broken and more complex use-cases should be covered by a separate, more powerful behavior (`mod-word`? `layer-word`? `smart-word`? ...), see #1451.


caksoylar:
FYI you seemed to have included an extra commit related to LED indicators in the PR branch.

snoyer:
> FYI you seemed to have included an extra commit related to LED indicators in the PR branch.

The LED indicator part deals with handling the capslock/numlock/... HID reports, not with actual LEDs on the board.
It is needed for this PR to know the state of capslock on the host and toggle it accordingly :)

snoyer:
Let's look at a french AZERTY keyboard:
azerty

The numrow is symbols and you need to shift for the numbers, for example the QWERTY `7&` key is `è7` on AZERTY, so pressing the `7` key types `è` and `shift+7` types `7`. Also check out the `ù%` key, that's QWERTY's `'"`.

Even more interesting: depending on the OS layout the capslock key will act as either _shift_-lock or _caps_-lock.
Now that nuance isn't really a thing with QWERTY but let's press `èçàù` (QWERTY `790'`) on the weird french board and see what gets typed:

- no modifiers: `èçàù`, ok
- holding shift: `790%`, as expected
- capslock on with shiftlock OS layout: `790%`, meh
- capslock on with proper capslock OS layout: `ÈÇÀÙ`, _nice!_

Personally, as a user, the last one is what I would expect `caps_word` to do and I don't think that's achievable without using the actual capslock mechanism.

tldr: the capitalization logic already exists in the OS layout and it seems unreasonable (if not impossible, as per previous example) to chase emulating it.

urob:
> - no modifiers: `èçàù`, ok
> 
> - holding shift: `790%`, as expected
> 
> - capslock on with shiftlock OS layout: `790%`, meh
> 
> - capslock on with proper capslock OS layout: `ÈÇÀÙ`, _nice!_
> 
> tldr: the capitalization logic already exists in the OS layout and it seems unreasonable (if not impossible, as per previous example) to chase emulating it by tacking on mods.

What's interesting here is the disconnect between regular shift and caps lock. Based on these examples, it sounds like there might even be a need for a regular modifier that capitalizes `èçàù` without the detour via caps lock, let's call it OSSHIFT. 

Would it make sense to implement such a mod? If we had OSSHIFT, then the regular capsword could deal with these cases by setting the mods to OSSHIFT.

Tangential question: How do French people with regular azerty layouts capitalize these accented characters? Do they always use caps lock?

snoyer:
> What's interesting here is the disconnect between regular shift and caps lock. Based on these examples, it sounds like there might even be a need for a regular modifier that capitalizes `èçàù` without the user having to activate/deactivate caps lock, let's call it OSSHIFT.

Let's not create anything that does "caps" and call it "shift" straight after establishing that "shift" and "caps" are different things ;)

I don't know what you mean by "regular modifier" here.
If the goal is to provide a key that can activate capslock when pressed and deactivate it when released that can be achieved by a  behavior (probably the same implementation as the one I propose here if we add an attribute to enable deactivation on release).


> Would it make sense to implement such a mod? If we had OSSHIFT, then the regular capsword could deal with these cases by setting the mods to OSSHIFT.

I'm not sure what the deal is with applying modifiers to be honest.

Let's forget about the current implementation for a minute, take a step (or a few) back, and think about this feature called "caps word" that we would like to _"behave similar to a caps lock, but will automatically deactivate [at the end of a word, basically]"_ ...
Is there any good reason for it not to be based on, you know, capslock? by having it automatically deactivate at the end of a word?


> Tangential question: How do French people with regular azerty layouts capitalize these accented characters? Do they always use caps lock?

Not sure, if I remember correctly the azerty layout that ships with Windows does shiftlock and not capslock so you'd have to memorize the `alt+digits` codes to do it from the keyboard. I guess the vast majority type them without accents (using shift) and let the spellchecker fix it.

snoyer:
I refactored the `caps_word` behavior into a `capslock` behavior that allows more flexible manipulation of the caps lock state.

In addition to `&caps_word` which still activates caps lock and deactivates it when a word separator is typed or the key is pressed again, this comes with the following bindings for free:

- `&caps_hold` activates caps lock as long as the key is held and deactivates it on release, as per @urob's suggestion.
- `&caps_on` and `&caps_off` respectively simply activate and deactivate caps lock (regardless of the current state, ie. you can mash `&caps_on` and caps lock will stay on instead of toggling on and off as it would with `&kp CAPS`)

petejohanson:
Part of the reason Caps Lock wasn't used in the original implementation was because of this macOS funkiness. macOS requires the caps lock key he held for some minimum time before enabling it on the host, tapping too quickly will *not* enable it.

app/src/behaviors/behavior_capslock.c
```diff
@@ -0,0 +1,214 @@
...
+    uint32_t position;
+    bool active;
+    bool just_activated;
+};
+
+static void toggle_capslock(const struct device *dev) {
+    const struct behavior_capslock_config *config = dev->config;
+    const struct behavior_capslock_data *data = dev->data;
+
+    zmk_behavior_queue_add(data->position, config->capslock_binding, true, 0);
```
rschenk, line 53:
```suggestion
    zmk_behavior_queue_add(data->position, config->capslock_binding, true, 100);
```

@snoyer @petejohanson I dug into the MacOS caps lock delay issue. Indeed, it's a "feature" of MacOS that you  must hold down caps lock briefly to activate it. I started with this wait number at 500ms to confirm that adding such a delay to the capsword code would work, then backed it down until I found the minimum that would reliably work. (The actual threshold seems to be around 75ms.) In my testing this delay does not seem to be noticeable at all.

snoyer: Thanks, @rschenk, I just pushed a similar change before I saw this comment.

snoyer:
> Part of the reason Caps Lock wasn't used in the original implementation was because of this macOS funkiness. macOS requires the caps lock key he held for some minimum time before enabling it on the host, tapping too quickly will _not_ enable it.

@petejohanson, thanks for the details, are there any other reasons for not using caps lock?

rschenk:
@snoyer I just pulled your latest commit. The good news is, it properly activates. The bad news is, it only deactivates by hitting the `&caps_word` key again, and does not deactivate by anything in the disable-on-keys list. I suspect the same was true of the solution I suggested as well (since they're effectively identical). There must be some timing weirdness when trying to disable and send the next keycode via disable-on-keys. I can do some troubleshooting tomorrow, as it seems you don't have a Mac?

snoyer:
> @snoyer I just pulled your latest commit. The good news is, it properly activates. The bad news is, it only deactivates by hitting the `&caps_word` key again, and does not deactivate by anything in the disable-on-keys list. I suspect the same was true of the solution I suggested as well (since they're effectively identical). There must be some timing weirdness when trying to disable and send the next keycode via disable-on-keys. I can do some troubleshooting tomorrow, as it seems you don't have a Mac?

From the little Mac troubleshooting time I managed to get so far (machine isn't mine), it looks like we don't get the updated state of caps lock after we enable it, `disable-on-keys` does trigger but nothing happens because we believe caps lock is already off.
```
[00:05:01.295,104] <dbg> zmk: capslock_find_key_item: found included usage: 0x07 - 0x2C
[00:05:01.295,104] <dbg> zmk: capslock_keycode_state_changed_listener: Deactivating capslock for 0x07 - 0x2C
[00:05:01.295,135] <dbg> zmk: set_capslock_state: capslock state is already 0
```

This would be related to the management of the various locks states from #999 which I have not looked into. It should be investigated independently of this behavior by setting up a listener and verifying if the state is properly tracked when pressing `&kp CAPS`. I'll try and find time to check that later.

petejohanson:
> > Part of the reason Caps Lock wasn't used in the original implementation was because of this macOS funkiness. macOS requires the caps lock key he held for some minimum time before enabling it on the host, tapping too quickly will _not_ enable it.
> 
> @petejohanson, thanks for the details, are there any other reasons for not using caps lock?
> 

The other main concern is properly tracking the caps state from the host, as you've already discovered. 

I'm definitely *not* opposed to merging a change like this, but it might be nice to have it be a DTS property on the behavior to control which approach is used?

snoyer:
> The other main concern is properly tracking the caps state from the host, as you've already discovered.

Yeah, I should have known it couldn't have been that easy when everything seemed fine on my linux machine...

> I'm definitely _not_ opposed to merging a change like this, but it might be nice to have it be a DTS property on the behavior to control which approach is used?

@petejohanson, glad to know you'd be on board with it. Obviously there's nothing to merge until/unless we can get it working properly.

If we did get it working though I believe it would be worth having a discussion about what the current `caps_word` actually _is_ and whether it should be renamed `shift_word` (because again, caps==shift is a sketchy shortcut) or even be refactored as a special case of a broader behavior (see #1451).

snoyer:
@rschenk, I got around to testing on both an Big Sur iMac and a Monterey Macbook, both on bluetooth, and it works on the former but not on the latter... :/
I do get the state change events (`zmk: zmk_led_indicators_process_report: Update LED indicators: endpoint=1, profile=4, flags=2`) from the iMac but not from the Macbook. Maybe the newer MacOS sends different HID reports? We'd need to test #999 alone to make sure I didn't botch something myself with their code.

rschenk:
@snoyer I am using a Macbook Pro running Monterey, I should have mentioned that earlier

snoyer:
@rschenk, it looks like @bortoz found and fixed the underlying issue over at #999. I applied his fix here so if you could try again and let us know that'd be great.

rschenk:
@snoyer I just tested 4c8eb87 and it works! Both USB and Bluetooth behave as expected. 🎉

rschenk:
What is preventing this from getting merged? Are we waiting on #999?

I've been using this on the daily for a month and it's working great

snoyer:
> What is preventing this from getting merged? Are we waiting on #999?

Yeah, I'd like to see the core functionality of #999 merged first, then I'll do a cleaner PR with a capslock behavior that would provides an alternative `caps_word` binding without getting rid of the existing one (I have seen the errors of my ways, replacing the existing one, renaming it `shift_word` maybe, would be a separate discussion).

> I've been using this on the daily for a month and it's working great

That's great to hear! Hopefully we can get moving on it again soon.

infused-kim:
I have been using this for a while and it's been working very well on macOS Ventura. 

Thank you for making this. It solved one of my biggest frustrations when using shortcuts with caps word turned on.

It also rebases without conflicts onto zmk main. But due to the zephyr 3.2 update, it requires a few small changes to compile.

You can look at this commit to see what needs to be changed:
https://github.com/infused-kim/zmk/commit/ea512b980b52669d8df11142242f40b5bfdf4b04

rschenk:
Just to chime in here, I've also been using this full-time since October, and it's working very well, but configuration-wise I think the config options of the existing Caps Word behavior (that is, providing `continue-list`) is much more user friendly than the in this PR which is essentially the inverse of that (providing `disable-on-keys`)

maciej-irl:
I'm using this now and I can also confirm it works very well. It's tricky to merge at first, since "LED Indicators" were renamed to "HID indicators" and the `ZMK_LED_INDICATORS_CAPSLOCK_BIT` constant is now gone. I'd love to see it merged once #999 lands.

nsmaciej:
Hopefully this should be easier to get into shape now that #2041 is merged.

snoyer:
Closed in favor of #2043 which provides the same functionality with an ever so slightly different, and hopefully more favorable, scope and framing.

PR #1491, opened 2022-10-13
chore(build): replace workflow set-output with environment variables

filterpaper:
The use of `set-output` will be [deprecated by GitHub](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). This change replaces it with [environment variables](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files) to share data between jobs and steps.

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -18,8 +18,8 @@ on:
         default: "bin"
         required: false
         type: string
-      artifact_name:
-        description: 'Artifact output file name'
+      archive_name:
+        description: 'Archive output file name'
         default: 'firmware'
         required: false
         type: string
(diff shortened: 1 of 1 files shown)


caksoylar:
This seems to work well in my testing, example build [here](https://github.com/caksoylar/zmk-config/actions/runs/3240134103). The output artifact contents is same as the current workflow.

Two related comments:
1. Style nit: Since we are using environment variables now, would it be better to use snake case variable names like the Github docs examples?
2. There are deprecation warnings related to node 12 coming from actions/checkout as below. It looks like it got updated to node 16 since v3.0.2 (current pinned version). Should we update those to `actions/checkout@v3.1.0` or just `actions/checkout@v3`?
image

filterpaper:
Agree to both comments. Variables are updated to use snake case, and external actions are using v3.

filterpaper:
Matrix output variable name from the first job seem obfuscating: "`needs.matrix.outputs.matrix`". Suggestions for a better self-documenting name?

caksoylar:
> Matrix output variable name from the first job seem obfuscating: "`needs.matrix.outputs.matrix`". Suggestions for a better self-documenting name, like "`needs.fetch.outputs.matrix)`"?

Since the job's display name is "Fetch Build Keyboards" its ID could `fetch_keyboards`? Optionally, the output name could be more specific as well, like `build_matrix`.

filterpaper:
Picked the latter to keep it consistent with the input variable. Everything else build cleanly now with no more warnings.

petejohanson (approved):
Tested as well here, with no issues: https://github.com/petejohanson/test-output-zmk-config/actions/runs/3301976416/jobs/5448210803

Thanks!

PR #1496, opened 2022-10-17
Add NUHS/NUBS alias for Non-US-Backslash/Non-US-Hash

ebastler:
This PR adds additional aliases to NON_US_BACKSLASH (NUBS) and NON_US_HASH (NUHS), as the full length keycode name causes keymaps to look confusing for ISO boards. 

The aliases are still unique and easy to understand in a keymap, while being short enough to be the same length as most regular keycodes.


petejohanson (approved):
Docs at https://deploy-preview-1496--zmk.netlify.app/docs/codes#symbols--punctuation look good, so does the header include changes.

PR #1497, opened 2022-10-18
Update intro.md

TweetyDaBird:
Added footnote VIA/VIAL


petejohanson:
You'll need to re-run the npm task to format the doc with prettier after the changes.

myst729:
What do you mean by "fork"? There is official VIA support in QMK, just turn on ENABLE_VIA.

petejohanson:
I'm not that familiar w/ VIA/VIAL, but it sure seems like it works with stock QMK, just requires actually setting up the keyboard properly with it.

TweetyDaBird:
> What do you mean by "fork"? There is official VIA support in QMK, just turn on ENABLE_VIA.

Yes, for VIA the support is integrated into QMK. for VIial though, it needs to be compiled with the Vial fork (make command) as it compiles in modules etc that simply don't exist in QMK's main repository.

Issue #1299, opened 2022-05-13
Feature Request: New &df layer behaviour


manna-harbour:
I'd like to be able to lock a momentary layer. That is, while holding `&mo` or `&lt`, tap the layer lock key then release the layer hold but stay on the layer.

`&to` can not be used as the layer is deactivated when the layer hold is released.  `&tog` can not be used as it deactivates the layer immediately.

QMK has [`DF`](https://github.com/qmk/qmk_firmware/blob/master/docs/keycodes.md#layer-switching-idlayer-switching), and KMonad has [`layer-switch`](https://github.com/kmonad/kmonad/blob/c43a0180a3b9ca9c4ff4a283b17369fc4e8e46e2/keymap/tutorial.kbd#L583), that can be used for this purpose.

I propose adding a new `&df` layer behaviour to change the default layer.

See the discord comment thread starting [here](https://discord.com/channels/719497620560543766/719565084208398406/907869038091313182) for an earlier discussion on this topic.

dabell-cc:
Could you make use of the example behaviour shown in the "Toggle-on-Tap, Momentary-on-Hold Layers" example found on the [Hold-Tap Behaviour](https://zmk.dev/docs/behaviors/hold-tap#example-use-cases:~:text=Toggle%2Don%2DTap%2C%20Momentary%2Don%2DHold%20Layers) page?

I set my keymap up to allow a 'Fn' key that will toggle the F1..F12 keys layer on when tapped, but allow Fn+F1..F12 to be used without then having to un-toggle the layer. A note of caution though, the documentation currently has the <&tog>, <&mo> bindings reversed so you'll want to switch them. (I put in a PR for it earlier today).

I suppose you might not want to dedicate a key to this behaviour, such as the `&lt` use case, where the tap might be a character entry key.

bmijanovich:
Below is an example of how to achieve a layer lock using [Conditional Layers](https://zmk.dev/docs/features/conditional-layers). By allowing two separate layers to each control the state of the layer you want to lock, you can avoid the layer deactivating when you release `&mo` after pressing `&tog` (or `&to`).

The key points here are:
* You are not limited to a single Conditional Layer condition
* Conditional Layer conditions do not require 2 or more `if-layers`; 1 will do just fine
* Since the `LOWER` layer is controlled by Conditional Layers, you cannot activate it directly; you must instead activate it indirectly using one of the `if-layers`

```
#define DEFAULT 0
#define LOWER 1
[...]

PR #1500, opened 2022-10-20
Base Layer behavior

snoyer:
This base layer behavior (`&base`) works like `&to` but stores the selected layer and automatically restores it when changing endpoint/profile. When switching between profiles the base layer is restored to the last one selected while on that profile.

I think it would make sense to persist the layer selections but I'm not sure how to go about that (settings API? is there guidelines for that?).


snoyer:
Added persistence so the selected base layers are remembered across power cycles.
Can anybody weigh in on whether I'm doing this the right way?

petejohanson:
It's an interesting idea... I wonder if it makes sense to force users to activate this in their keymap, versus adding a configuration to each layer, e.g.

```
  keymap {
    mac_linux_base {
      base-for = <USB_EP BLE_EP(0) BLE_EP(1)>;
      bindings <...>;
    };

    win_base {
      base-for = <BLE_EP(2)>;
      bindings = <...>;
    };
    
  };
```

Thoughts?

snoyer:
> It's an interesting idea... I wonder if it makes sense to force users to activate this in their keymap, versus adding a configuration to each layer, e.g.

@petejohanson, would that suggestion mean hardcoding the base layer to a profile index? If that is the case this can already be achieved by having macros combining `&bt BT_SEL x` and `&to y` for each profile/endpoint.

The version I'm proposing allows the selection of the base layer to be done "at runtime"  without having to plan and stick to (or even know at all) which machine is going to be paired/plugged on which endpoint when compiling the keymap.

snoyer:
Biweekly bump until we get to the bottom of this @petejohanson ? Could you clarify the objection?

I personally can't get my head around wanting to hardcode choices per bluetooth profile at compile time (are users expected to know what device they're going to pair before they flash the firmware?).
I guess I could see having the `base-for` property as an optional way to set the default layer for some profiles but I feels like adding unnecessary complexity.

Can you share some use-cases where compile-time settings are preferable?

petejohanson:
> Biweekly bump until we get to the bottom of this @petejohanson ? Could you clarify the objection?
> 
> I personally can't get my head around wanting to hardcode choices per bluetooth profile at compile time (are users expected to know what device they're going to pair before they flash the firmware?). I guess I could see having the `base-for` property as an optional way to set the default layer for some profiles but I feels like adding unnecessary complexity.
> 
> Can you share some use-cases where compile-time settings are preferable?

Examples:

* As a user, given I have a macOS work laptop and Linux personal laptop, then I want to take the same keymap with me to multiple devices where the 0th slot is always used for my work laptop, and 1st is for my personal laptop. Switching between the two should always pick a different layer with modifier layouts that match.
* As a vendor of keyboards targeting less technical users, I want to off a more constrained set of choices: The only profiles exposed are via a small toggle switch that is for selecting from pairing to a macOS or Windows machine, with stock keymaps for each.

snoyer:
PR branch has been updated. There's been a minor implementation change since the original proposal: the base layers can now be specified through configuration so that the behavior can  change base layers without affecting the state of non-base layers (basically won't kick you out of any higher active layers when changing base, unlike `&to` would)

All of that is, of course, moot if the admin team feels like dynamic base layers are not a valuable use-case. In which case let me know and I'll close the PR.

In any case this behavior is available as a module here: https://github.com/snoyer/zmk-behavior-baselayer

nmunnich:
My opinion on this concept can be read here: https://github.com/zmkfirmware/zmk/issues/1299#issuecomment-2754384387

Id be interested in your thoughts after reading it.

snoyer:
> My opinion on this concept can be read here: [#1299 (comment)](https://github.com/zmkfirmware/zmk/issues/1299#issuecomment-2754384387)
> 
> Id be interested in your thoughts after reading it.

This PR doesn't attempt to modify ZMK's "default" layer logic, it will only save and restore the state of one layer out of a given set of configured base layers. Most likely the base layers would be the first ones in the keymap, one of them being the "default" one but I'm not sure that's really relevant. It does assume the base layers are exclusive of each other so it doesn't save multiple layers.

Regarding your requirements of saving and restoring multiple layers, I think that should be possible but the main question for me would be how to specify a multi-layer state in the keymap? would you use a bit field and write something like `&base_multi 0b1010` (or `&base_multi XXX(1)+XXX(3)` with a bit shifting macro) to save the second and fourth layer for the current endpoint?
Also, in my experience, I don't want to restore the full stack of layers, because if I'm changing endpoints from a different layer I don't want it to be deactivated when the change event is handled. So we'd need to be able to specify which layer to activate, which to deactivate, and which to leave untouched, IMO.

I'm not quite sure I fully grasp your use-case and the layer locking stuff so I can't really tell how much this PR overlaps it or if it's actually a different thing

nmunnich:
My vision would be a behavior, let's call it "&save_layers_to_endpoint". If the behavior is enabled, endpoints are augmented with an additional uint32 to mark which layers are active for them. The behavior simply saves the current active layers to the currently selected endpoint, and when the endpoint is switched the uint32 is read from to determine which layers to activate.

EDIT: The layer locking PR would exist to identify "temporary layers" which shouldn't be touched when switching endpoints, and which should be ignored when setting endpoints.

snoyer:
> My vision would be a behavior, let's call it "&save_layers_to_endpoint". If the behavior is enabled, endpoints are augmented with an additional uint32 to mark which layers are active for them. The behavior simply saves the current active layers to the currently selected endpoint, and when the endpoint is switched the uint32 is read from to determine which layers to activate.

I see, so the key point here is that it saves the _current_ layers when pressed. I don't think that fits what I'm going for here: in my own usage for example I have a "firmware" layer (last one of the keymap) with 3 keys to set the base layer to linux/mac/win (first 3 of the keymap), so the layers to save are independent of the current layer, and wouldn't want the current layer to be included in the save state either.

But what you're describing could be done with a similar implementation.

nmunnich:
Actually, it would be able to solve your use case. The key being that I'm envisioning only the "locked layers" being saved, so if you use a momentary layer to get to the "firmware" layer, or use a tog that has the locking param removed, then the firmware layer wouldn't be saved. You could easily use macros to combine a &to with the &save_layers_to_endpoint, for example.

Both approaches are approximately equivalent, but I personally prefer the idea of saving actively locked layers because it allows for more dynamic reconfiguration on the fly, and easily allows for multiple layers to be activated at once.

snoyer:
> Actually, it would be able to solve your use case. The key being that I'm envisioning only the "locked layers" being saved, so if you use a momentary layer to get to the "firmware" layer, or use a tog that has the locking param removed, then the firmware layer wouldn't be saved. You could easily use macros to combine a &to with the &save_layers_to_endpoint, for example.
> 
> Both approaches are approximately equivalent, but I personally prefer the idea of saving actively locked layers because it allows for more dynamic reconfiguration on the fly, and easily allows for multiple layers to be activated at once.

Ok that makes sense (I think, I would have to actually play with it to make sure I fully get it).

So what's the way forward here? I close this PR and we wait for #2717 to land in main before having another look at the whole save/restore thing?

nmunnich:
I think closing the PR is a bit premature. While I do think my approach is nicer overall, I think more discussion is warranted first. My concept depends on #2717 anyway, so I see no reason to close this until that lands in main and we can do a more direct comparison.

snoyer:
> I think closing the PR is a bit premature.

Yeah it's only been open for almost 3 years, let's not rush anything :D

> While I do think my approach is nicer overall, I think more discussion is warranted first. My concept depends on #2717 anyway, so I see no reason to close this until that lands in main and we can do a more direct comparison.

I don't see a reason to keep it open unless someone from the team confirms it is an interesting use-case and we start talking about what it would take to get it merged.

If Pete, as the BDFL, said this PR doesn't address a legit use-case and the prescribed way is to hardcode everything then it should be closed.

If you, as a maintainer, are saying this PR would be superseded by a more complete/powerful/better feature that is already on the roadmap then it should be closed.

If we're just 2 randos with no vision or authority arguing on whether this PR should stay in limbo because some other PR which is also in limbo may or may not overlap then I'm not convinced we're making any progress but...
Yes, your suggestion sounds more general and more powerful, I'm assuming the backend would be nicer but the keymap side might be more clunky if we need to use macros to `&to` then `&save_current` instead of just `&save X`? I guess a parameterized macro would have an equivalent syntax but wouldn't a `&to` macro kick you out of your current layer before saving?

snoyer:
@nmunnich looks like the bot overruled you and closed the PR even though you said not to because more discussion was warranted :shrug:

PR #1507, opened 2022-10-24
Add a kscan multiplex matrix option

HookyQR:
Adds support for https://kbd.news/Square-or-round-robin-matrix-1400.html with addition of an extra interrupt pin for low power support.

Implementation of this matrix in hardware is left as an exercise for the user. Ensure all lines to the interrupt pin are not able to re-feed the matrix or it will surely fail.

Example:
image

* 6 pin multiplex matrix with interrupt support - for a 6x4 + 3 board*

Shields implementing this scan setup may not be suitable for all boards (see the referenced document above).

It is expected that, with proper design, (compared to a matrix kscan) an additional 2*n diodes will be required, where n is the number of input pins in use. One for the line to the input pin (this may not be required for some controllers), and one for each 'column' to the interrupt pin. Note that not all of the multiplex matrix pins need to be used as input to the controller.

eg. An 8 pin multiplex kscan matrix may use all 8 pins for output, but only use 5 for input.


HookyQR:
Closed in favour of renamed #1694

PR #1510, opened 2022-10-26
feat(docs): Add suggestion to workaround macOS 13 Finder bug

filterpaper:
Copying a UF2 into microcontrollers with macOS 13's Finder will trigger an "error code 100093". This change add a workaround for that error.

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -21,6 +21,10 @@ Variations of the warnings shown below occur when flashing the `<firmware>.uf2` 
 | :--------------------------------------------------------------------------: |
 |                An example of the file transfer error on MacOS                |
 
+### MacOS Ventura error
+
+MacOS 13.0 (Ventura) Finder running on Apple silicon hardware may report an error code 100093 when copying `<firmware>.uf2` files into microcontrollers. This bug is limited to the operating system's Finder. You can work around it by copying on Terminal command line or use a third party file manager.
+
 ### CMake Error
 
 An error along the lines of `CMake Error at (zmk directory)/zephyr/cmake/generic_toolchain.cmake:64 (include): include could not find load file:` during firmware compilation indicates that the Zephyr Environment Variables are not properly defined.


docs/docs/troubleshooting.md
```diff
@@ -110,3 +110,7 @@ For the `nRF52840`, the value `PLUS_8` can be set to any multiple of four betwee
 ### Other notes and warnings
 
 - If you want to test bluetooth output on your keyboard and are powering it through the USB connection rather than a battery, you will be able to pair with a host device but may not see keystrokes sent. In this case you need to use the [output selection behavior](../docs/behaviors/outputs.md) to prefer sending keystrokes over bluetooth rather than USB. This might be necessary even if you are not powering from a device capable of receiving USB inputs, such as a USB charger.
```
petejohanson, line 112: Why remove this?

petejohanson (changes requested):

docs/docs/troubleshooting.md
filterpaper: There is no removal. That floating `-` is a single bullet point from the existing section and it is still there.

petejohanson: Oh! Rendering funkiness in the mobile app. Thanks.

caksoylar:
Thanks! I think it would make sense to move it under the "file transfer error" header as a special case. Technically it fits the description so users might not scroll to the end of the page to see this after noticing that.

filterpaper:
> Thanks! I think it would make sense to move it under the "file transfer error" header as a special case. Technically it fits the description so users might not scroll to the end of the page to see this after noticing that.

Agree with this suggestion and moved it under a sub-header of that section.

docs/docs/troubleshooting.md
```diff
@@ -21,6 +21,10 @@ Variations of the warnings shown below occur when flashing the `<firmware>.uf2`
 | :--------------------------------------------------------------------------: |
 |                An example of the file transfer error on MacOS                |
 
+#### MacOS Ventura error
+
+The Finder app on MacOS 13.0 (Ventura) may report an error code 100093 when copying `<firmware.uf2>` files into microcontrollers. This bug is limited to the operating system's Finder. You can work around it by copying on Terminal command line or use a third party file manager.
```
caksoylar, line 26: Nit, `<firmware>.uf2` with the angle brackets positioned to match above uses.

caksoylar (approved):

petejohanson:
@filterpaper Can you please try rebasing now that #1512 is in? We had netlify build failures due to some outdated deps/build images.

filterpaper:
> @filterpaper Can you please try rebasing now that #1512 is in? We had netlify build failures due to some outdated deps/build images.

Done.

petejohanson:
Looks like just the prettier check failing now.

filterpaper:
> Looks like just the prettier check failing now.

The failed output isn't verbose about why its failing. Fixed the section title depth that was reported by https://prettier.io/playground/ and that site isn't reporting any other format problems.

filterpaper:
All good now—prettier didn't like the extra blank line at the end of the file.

petejohanson (approved):
Thanks!

PR #1514, opened 2022-10-28
Added a board built on the NRF52 DK

lental:
Created a self-contained board meant as a sample project on the NRF52 DK. Allows for a super easy starting point for any developer learning ZMK with the development kit.

Happy to tweak this as necessary to make this more useful. 

## Board/Shield Check-list
 - [ x] This board/shield is tested working on real hardware
 - [ x] Definitions follow the general style of other shields/boards upstream ([Reference](https://zmk.dev/docs/development/new-shield))
 - [ x] `.zmk.yml` metadata file added
 - [ x] Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)
 - [ x] General consistent formatting of DeviceTree files
 - [ x] Keymaps do not use deprecated key defines (Check using the [upgrader tool](https://zmk.dev/docs/codes/keymap-upgrader))
 - [ x] `&pro_micro` used in favor of `&pro_micro_d/a` if applicable
 - [ x] If split, no name added for the right/peripheral half
 - [ x] Kconfig.defconfig file correctly wraps *all* configuration in conditional on the shield symbol
 - [ x] `.conf` file has optional extra features commented out
 - [ x] Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)


nmunnich:
Hey, thanks for the PR. While the motivation is good, this approach is not one that we'll be going with. Considering the structure of the dk, and the fact that we already have a nrf52dk board provided by Zephyr, the approach taken by #2730 is preferable. Apologies that it took so long just to give negative feedback.

PR #1516, opened 2022-10-30
Set bits per pixel and color depth on n!v

HookyQR:
The default bit/pixel is 32, as is the default color depth. This results in LVGL sending all zeros to the nice view, which makes for not happy dev times. 😢

Default it to 1x1 in the nice_view shield so no one else has to go though this. 🎉

### DIFF
--- a/app/boards/shields/nice_view/Kconfig.shield
+++ b/app/boards/shields/nice_view/Kconfig.shield
@@ -3,10 +3,3 @@
 
 config SHIELD_NICE_VIEW
 	def_bool $(shields_list_contains,nice_view)
-
-config LVGL_BITS_PER_PIXEL
-	default 1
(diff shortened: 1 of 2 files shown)


app/boards/shields/nice_view/Kconfig.shield
```diff
@@ -3,3 +3,10 @@
 
 config SHIELD_NICE_VIEW
 	def_bool $(shields_list_contains,nice_view)
+
+config LVGL_BITS_PER_PIXEL
```
petejohanson, line 7: These kinds of changes should go in the .defconfig file.

petejohanson (changes requested):

petejohanson (approved):

PR #1518, opened 2022-10-30
Bugfix: make tap dances work on a combo

okke-formsma:
Tap-dances did not work because they intercepted the keypresses from combos before the combos could trigger.

 By moving the tap-dance behavior lower in the CMakeList, their event handlers are triggered after combos, causing the right sequence of events so tap-dances can be triggered by a combo.

Test included in code. Tested on a board by cryptanon:

image


petejohanson (approved):
Excellent!

PR #1528, opened 2022-11-10
Checkout submodules when building user config

urob:
This modifies the `build-user-config.yml` called by Github Actions to checkout submodules of the user's `zmk-config`. This allows users to build their config in a modular fashion. For instance, instead of adding a shield subdirectory, they can just keep all the hardware related stuff in a separate repo and include that (or someone else's repo) as a submodule.

For anyone who wants to try this out, you can do so by replacing the contents of `.github/workflow/build.yml` in your `zmk-config` repo with the following contents:
```
on: [push, pull_request, workflow_dispatch]

jobs:
  build:
    uses: urob/zmk/.github/workflows/build-user-config.yml@build-with-submodules
```

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -33,6 +33,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+            submodules: recursive
 
       - name: Install yaml2json
         run: python3 -m pip install remarshal
@@ -69,6 +71,8 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+            submodules: recursive
 
       - name: Cache west modules
         uses: actions/cache@v3.0.11


petejohanson:
Why not just add as another module in your west.yml? The whole point of west is to avoid the submodule mechanics.

urob:
> Why not just add as another module in your west.yml? The whole point of west is to avoid the submodule mechanics.

Hm, that wouldn't do anything for building locally, though, no? What I like about the submodules approach is that it works equally when building locally and remotely with Github Actions. In either case, a single `git submodule` command is enough to add a dependency to the zmk-config.

shawnohare:
While I recognize this issue is now older and is not informed by any out of channel discussions,

urob:
Closing this as I came to better appreciate Pete's comment above. I agree that using west submodules integrates better with the rest of the zmk/zephyr ecosystem. And as I learned by now it also works great for setting up a local build environment.

PR #1533, opened 2022-11-11
kscan: Limit maximum amount of key presses

tyalie:
When using a keyboard without diodes in the key matrix, ghosting can become a serious issue. Quite often three keys held at the same time are needed to produce a ghosted fourth key press when using a row, col based keyboard matrix. In these cases it makes sense to allow only three key presses at the same time, ignoring all of them if a fourth is pressed.

The changes made here are a simple work-of-proof how this "limit the amount of key presses" could be done.

Note: There's no configuration options or similar to change this behaviour right now. I'm happy to develop this idea further if I receive some feedback.

Signed-off-by: Sophie Tyalie <dev@flowerpot.me>

### DIFF
--- a/app/drivers/kscan/kscan_gpio_matrix.c
+++ b/app/drivers/kscan/kscan_gpio_matrix.c
@@ -225,6 +225,8 @@ static int kscan_matrix_read(const struct device *dev) {
     struct kscan_matrix_data *data = dev->data;
     const struct kscan_matrix_config *config = dev->config;
 
+    int num_active = 0;
+    int stored_states[3] = {0};
     // Scan the matrix.
     for (int o = 0; o < config->outputs.len; o++) {
         const struct gpio_dt_spec *out_gpio = &config->outputs.gpios[o];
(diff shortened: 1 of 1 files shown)


app/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -225,6 +225,8 @@ static int kscan_matrix_read(const struct device *dev) {
     struct kscan_matrix_data *data = dev->data;
     const struct kscan_matrix_config *config = dev->config;
 
+    int num_active = 0;
```
petejohanson, line 228: You would need to store this in the instance data struct to be used across read passes.

Without this, you could limit/ignore different key presses on each read, if a key press earlier in the scanning comes in after the limit is hit.

petejohanson (changes requested):
I'm definitely open to an enhancement to the matrix kscan driver to add the limit for anti-ghosting in cheap hardware.

The code here seems to be, as you note, a decent PoC, but needs refinement to work well across various sequences, etc.

And yes, the limit would need to be an optional property in the DTS for this driver.

app/drivers/kscan/kscan_gpio_matrix.c
tyalie: > You would need to store this in the instance data struct to be used across read passes.

I think I don't fully get it right now. How would you reset the counter in this case?

JohnAZoidberg:
> Quite often three keys held at the same time are needed to produce a ghosted fourth key press

Once three are pressed the matrix registers all 4, so you don't know which one the third was - that means the limit is two.

PR #1536, opened 2022-11-12
Does this example make more sense?

russelldb:
I was confused in the `tog` example as to why one would `tog` the `NAVI` layer when in the `NAVI` layer.

## Board/Shield Check-list

### DIFF
--- a/docs/docs/behaviors/layers.md
+++ b/docs/docs/behaviors/layers.md
@@ -120,7 +120,7 @@ Example:
 
 		nav_layer {
 			bindings = <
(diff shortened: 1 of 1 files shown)


caksoylar:
`&tog NAVI` in NAVI layer would turn off the layer itself, leaving you back on default layer. Actually `&tog DEFAULT` doesn't do anything since the default layer (first defined one) cannot be turned on or off; it is always on. I think you might be confusing `&tog` with the `&to` behavior in which case this setup would make sense.

russelldb:
That makes sense now, apologies for the time.

PR #1537, opened 2022-11-13
Update link to Zephyr led docs

manthonygfp:
The documentation has a broken link. This change updates the link to a relevant page.

### DIFF
--- a/docs/docs/config/backlight.md
+++ b/docs/docs/config/backlight.md
@@ -34,7 +34,7 @@ Applies to: [`/chosen` node](https://docs.zephyrproject.org/latest/build/dts/int
 
 See the Zephyr devicetree bindings for LED drivers:
 
-- [gpio-leds](https://docs.zephyrproject.org/latest/build/dts/api/bindings/led/gpio-leds.html)
+- [gpio-leds](https://docs.zephyrproject.org/3.0.0/reference/devicetree/bindings/gpio/gpio-leds.html)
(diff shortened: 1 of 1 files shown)


docs/docs/config/backlight.md
```diff
@@ -34,7 +34,7 @@ Applies to: [`/chosen` node](https://docs.zephyrproject.org/latest/build/dts/int
 
 See the Zephyr devicetree bindings for LED drivers:
 
-- [gpio-leds](https://docs.zephyrproject.org/latest/build/dts/api/bindings/gpio/gpio-leds.html)
+- [gpio-leds](https://docs.zephyrproject.org/latest/build/dts/api/bindings/led/gpio-leds.html)
```
petejohanson, line 37: While we are at it, can we instead link to the 3.0 release version docs, so we're not looking to a moving target with the `latest` piece?

petejohanson (changes requested):
Thanks for the fix! Just one suggestion.

docs/docs/config/backlight.md
manthonygfp: I wasn't sure if you wanted to use the 3.0.0 or use whatever the latest is for 3.0.x. If this isn't good, I'll have to figure out a more clever way to always be selecting 3.0.x.

petejohanson: Fixed at 3.0.0 is fine, it should be API stable for 3.0.x.

petejohanson (approved):
Thanks!

PR #1542, opened 2022-11-16
fix(usb): add USB_DC_CLEAR_HALT to supported states

xudongzheng:
See https://manpages.debian.org/testing/linux-manual-4.9/usb_clear_halt.9 for information on halted USB endpoints.

USB_DC_CLEAR_HALT tends to appear occasionally (maybe every day or so) when connected via my USB hub.

Before fix:

```
[24:50:12.712,371] <dbg> zmk: zmk_usb_get_conn_state: state: 9
[24:50:12.712,371] <dbg> zmk: get_selected_endpoint: Only BLE is ready.
```


petejohanson (approved):
Thanks!

PR #1544, opened 2022-11-17
New Behavior: Custom Lock

tyalie:
This pull request introduces a new behavior I've called custom lock (for now).

It behaves similarly to e.g. Caps Lock, Num Lock, ..., except that the behavior transition happens on the side of the keyboard.
The use case I have for it, is to be able to disable or enable a space after typing a dead key, as not every computer I'm using my keyboard with has dead keys on its layout.

Definition in a dts file
```dtsi
behaviors {
  deadlock: dead_lock_prevent {
    compatible = "zmk,behavior-custom-lock";
    label = "dead-key-toggle";
    #binding-cells = <0>;

    deadk: dead_key {
      label = "dead-key";
      #binding-cells = <2>;
      bindings = <&kp>, <&kp>;
    };

    deadm: dead_macro {
      label = "dead-key-w-macro";
      #binding-cells = <2>;
      bindings = <&macro1>, <&macro2>;
    };
  };
};
```

This could be used in a layer like this
```
function_layer {
  bindings = <
    &deadlock  /* toggle this lock on and off */
    &deadk A B  /* will return B if lock is enabled, A otherwise */
    &deadm 0 0
  >;
};
```

EDIT: I've added the ability to store the custom lock state persistently. Using the `persistence` property of `zmk,behavior-custom-lock`, one can choose between no persistence, normal persistence and per profile persistence. For the latter two the zephyr settings infrastructure will be used.


caksoylar:
This looks pretty cool, especially with per-profile behavior and persistence. How about a different kind of naming that emphasizes the binary switching nature of the behavior? One idea (that I don't particularly like a lot) might be "bifunctional key," maybe something to that effect.

tyalie:
Tbqh I'd be happy to have a more expressive name I just didn't want it to hinder my early development

I thing I haven't looked at yet is how to integrate this into e.g. an OLED screen or such. It's currently very encapsulated inside the behavior code, but I don't have an OLED or LEDs to test the changes with. But I think that I might be a good idea to pull the state code out of the behavior in order to have public header interfaces?

caksoylar:
> I thing I haven't looked at yet is how to integrate this into e.g. an OLED screen or such. It's currently very encapsulated inside the behavior code, but I don't have an OLED or LEDs to test the changes with. But I think that I might be a good idea to pull the state code out of the behavior in order to have public header interfaces?

It might be a good idea to define a new event (of the sort declared [here](https://github.com/zmkfirmware/zmk/tree/main/app/include/zmk/events)) then raise it in the behavior. Then a display widget can subscribe to the event and update accordingly.

snoyer:
Would it be worth re-framing the issue in terms an "OS mode" of some sort? I think that came up in the past regarding unicode macros where ZMK would need to be told which OS it's working with to generate the right sequence of keys.

This PR looks like a solution to a binary subset of the wider problem. Could it be extended to an arbitrary number of modes rather than just 2, making it more of a "mode selector" rather than a "lock"?

Alternatively, the other way to go about it is to have separate sets of layers with the different variations of the behaviors. This would be functional in the current state of things except for the persistence aspect which has been proposed in #1500.

tyalie:
> Would it be worth re-framing the issue in terms an "OS mode" of some sort?

I'm not sure if this would be so helpful. Yes I use this PR in order to adapt the keyboard to the Apple layouts. But the dead key prevention that I employ is pretty much vendor (i.e. android on Samsung) or user defined (some people use dead key layouts in Germany, others don't). One could also easily realise fn-lock with this PR. Although there could be better ways for that (e.g. layer toggle), as these are quite a few keys.

> Could it be extended to an arbitrary number of modes rather than just 2, making it more of a "mode selector" rather than a "lock"?

Of course. The code easily allows that. It would need adaptations in the dts definition thoo. But I'm not sure how if an arbitrary amount of states would really be the go to for me here, as it could potentially hinder the easiness of the current solution if you just want to flip between two.

> Alternatively, the other way to go about it is to have separate sets of layers with the different variations of the behaviors.

I think the solution I propose here is a good way if I just wanna switch out a (very) limited amount of keys (e.g. flip ctrl and alt, remove a space after I type ^ or `, ...). If I would introduce new layer sets for every "lock type" the layout could get unmaintainable very quickly (e.g. in my case 6 new layers where 90% are the same keys). 

But the moment I want to e.g. adapt for different Unicode input types across systems the solution proposed here would be a hell to use. In that case I would honestly introduce new layers and maybe just use this feature for the layer activation button. The PR you mentioned could be another solution for this.

PR #1545, opened 2022-11-18
fix(boards): Fix XIAO BLE voltage divider config.

petejohanson:
* Use proper resistor value for the `output-ohms` property.

Found this bug in my own testing, and the fix came from `sctanf` [on Discord](https://discord.com/channels/719497620560543766/719909884769992755/1042977575418474586). Reviewing the schematic confirms that's the proper fix.


sctanf:
Tested, working. Battery percentage reflects my external measured voltage off the battery.

PR #1549, opened 2022-11-20
Bug: Layer-tap from higher-number to lower fails

hunner:
This tests uses `&to` to switch to layer 3 which has a `&lt` for layer 2, but instead the keycodes on layer 3 are still sent. I'm not sure how to fix it, but at least this replicates the bug.

I found this by trying a similar scenario on my own zmk-config.


caksoylar:
Hello, thank you for a thorough description and including a test case! Unless I am the one

hunner:
> In other words, layer priority is tied to definition order, not activation order.

Ah hah! My misunderstanding then. Thanks!

PR #1552, opened 2022-11-25
fix(docs): Correct syntax in capsword mods

nickcoutsos:
Update the example for the `mods` property for `&caps_word` per the devicetree spec

> values may be represented as arithmetic, bitwise, or logical expressions within parenthesis.

This is also in line with the usage in the mod-morph documentation as well.


petejohanson (approved):
Thanks!

PR #1556, opened 2022-11-28
feat(display): Add modifiers widget

caksoylar:
This is a barebones modifiers widget for showing the status of active modifiers on the bottom right of the status screen (thus it should not be enabled together with the WPM widget). It only tracks explicit modifiers (is showing implicit modifiers desirable?) but it seems to work well with sticky keys so it should be useful for tracking their state.

The widget can be enabled with `CONFIG_ZMK_WIDGET_MODS_STATUS=y` in your `.conf` file. You can use `CONFIG_ZMK_WIDGET_MODS_STATUS_CHARACTERS` to change the characters displayed for each modifier (one char each, default is `"CSAG"`).

~~Since Zephyr 3.2 has some LVGL changes affecting widgets I'd happy to revisit once it is merged in as well.~~ This has been completed.

Small demo:
mods

### DIFF
--- a/app/src/display/widgets/mods_status.c
+++ b/app/src/display/widgets/mods_status.c
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2022 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <logging/log.h>
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#include <zmk/display.h>
+#include <zmk/display/widgets/mods_status.h>
+#include <zmk/events/keycode_state_changed.h>
+#include <zmk/event_manager.h>
+#include <zmk/endpoints.h>
+#include <zmk/hid.h>
+
+static sys_slist_t widgets = SYS_SLIST_STATIC_INIT(&widgets);
+
+struct mods_status_state {
+    uint8_t mods;
+};
+
+struct mods_status_state mods_status_get_state(const zmk_event_t *eh) {
+    return (struct mods_status_state){.mods = zmk_hid_get_explicit_mods()};
+};
+
+void set_mods_symbol(lv_obj_t *label, struct mods_status_state state) {
+    char text[5] = {};
+
+    LOG_DBG("mods changed to %i", state.mods);
+    if (state.mods & (MOD_LCTL | MOD_RCTL))
+        strcat(text, "C");
+    if (state.mods & (MOD_LSFT | MOD_RSFT))
+        strcat(text, "S");
+    if (state.mods & (MOD_LALT | MOD_RALT))
+        strcat(text, "A");
+    if (state.mods & (MOD_LGUI | MOD_RGUI))
+        strcat(text, "G");
+
+    lv_label_set_text(label, text);
+    lv_obj_align(label, NULL, LV_ALIGN_IN_BOTTOM_RIGHT, -1, 0);
+}
+
+void mods_status_update_cb(struct mods_status_state state) {
+    struct zmk_widget_mods_status *widget;
+    SYS_SLIST_FOR_EACH_CONTAINER(&widgets, widget, node) { set_mods_symbol(widget->obj, state); }
+}
+
+ZMK_DISPLAY_WIDGET_LISTENER(widget_mods_status, struct mods_status_state, mods_status_update_cb,
+                            mods_status_get_state)
+ZMK_SUBSCRIPTION(widget_mods_status, zmk_keycode_state_changed);
+
+int zmk_widget_mods_status_init(struct zmk_widget_mods_status *widget, lv_obj_t *parent) {
+    widget->obj = lv_label_create(parent, NULL);
+    lv_label_set_align(widget->obj, LV_LABEL_ALIGN_RIGHT);
+
+    lv_obj_set_size(widget->obj, 40, 15);
+
+    sys_slist_append(&widgets, &widget->node);
+
+    widget_mods_status_init();
+    return 0;
+}
+
+lv_obj_t *zmk_widget_mods_status_obj(struct zmk_widget_mods_status *widget) { return widget->obj; }
--- a/app/src/display/widgets/mods_status.c
+++ b/app/src/display/widgets/mods_status.c
@@ -14,6 +14,12 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/endpoints.h>
 #include <zmk/hid.h>
 
+#define MOD_CHARS CONFIG_ZMK_WIDGET_MODS_STATUS_CHARACTERS
+#define MOD_CHARS_LEN (sizeof(MOD_CHARS) - 1)
+
+BUILD_ASSERT(MOD_CHARS_LEN == 4,
+             "ERROR: CONFIG_ZMK_WIDGET_MODS_STATUS_CHARACTERS should have exactly 4 characters");
+
 static sys_slist_t widgets = SYS_SLIST_STATIC_INIT(&widgets);
 
 struct mods_status_state {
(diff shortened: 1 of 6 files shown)


duckyb:
I tested the current version on read hardware and it's working as intended.

app/src/display/widgets/mods_status.c
```diff
@@ -0,0 +1,65 @@
...
+struct mods_status_state mods_status_get_state(const zmk_event_t *eh) {
+    return (struct mods_status_state){.mods = zmk_hid_get_explicit_mods()};
+};
+
+void set_mods_symbol(lv_obj_t *label, struct mods_status_state state) {
+    char text[5] = {};
+
+    LOG_DBG("mods changed to %i", state.mods);
+    if (state.mods & (MOD_LCTL | MOD_RCTL))
+        strcat(text, "C");
```
petejohanson, line 32: Does it make sense to have the display strings for each be Kconfig settings, so folks who are primarily Mac users can use "O" for option, for instance, maybe "CMD" for GUI, etc? Would require some care to determine the string allocation.

app/src/display/widgets/mods_status.c
```diff
@@ -0,0 +1,65 @@
...
+
+ZMK_DISPLAY_WIDGET_LISTENER(widget_mods_status, struct mods_status_state, mods_status_update_cb,
+                            mods_status_get_state)
+ZMK_SUBSCRIPTION(widget_mods_status, zmk_keycode_state_changed);
+
+int zmk_widget_mods_status_init(struct zmk_widget_mods_status *widget, lv_obj_t *parent) {
+    widget->obj = lv_label_create(parent, NULL);
+    lv_label_set_align(widget->obj, LV_LABEL_ALIGN_RIGHT);
+
+    lv_obj_set_size(widget->obj, 40, 15);
```
petejohanson, line 57: We can hopefully ditch this as of the Zephyr 3.2 work.

petejohanson:
Does it make sense to keep each mod in a static location instead of shifting in from the right?

app/src/display/widgets/mods_status.c
caksoylar: Thanks for pointing it out, I hadn't considered that indeed. I think constraining each mod to a single character would make our life easier, otherwise we might want a separator between multiple mods as well. It can also help avoid adding too many (4 separate?) Kconfig options. This way we could have a single Kconfig with a string value of 4 chars to correspond to "ctrl-shift-alt-gui" in order. What do you think?

caksoylar:
Thanks for the review!

> Does it make sense to keep each mod in a static location instead of shifting in from the right?

That was my initial plan but I didn't find a good way to implement it. The font is not monospaced using a single label with placeholders like `" "` for inactive mods doesn't guarantee fixed positioning. We could have 4 separate labels and position them manually but the correct positioning could depend on font and display size which doesn't feel robust. If you have any ideas I'd be happy to hear them.

app/src/display/widgets/mods_status.c
petejohanson: That'd work for me, at least as a first pass at this.

petejohanson:
> Thanks for the review!
> 
> > Does it make sense to keep each mod in a static location instead of shifting in from the right?
> 
> That was my initial plan but I didn't find a good way to implement it. The font is not monospaced so using a single label with placeholders like `" "` for inactive mods doesn't guarantee fixed positioning. We could have 4 separate labels and position them manually but the correct positioning could depend on font and display size which doesn't feel robust. If you have any ideas I'd be happy to hear them.

Hmm.... All painful points. I think to do it right will need to wait on https://docs.lvgl.io/master/layouts/flex.html?highlight=flexbox when we get the Zephyr 3.2 update in. Put each label as a child of a container object using flex layout... and it might just work?

app/src/display/widgets/mods_status.c
caksoylar: Sure, I'll take a shot at that and try to add a preprocessor assert if I can to ensure the value has 4 characters.

caksoylar:
> > Thanks for the review!
> > > Does it make sense to keep each mod in a static location instead of shifting in from the right?
> > 
> > 
> > That was my initial plan but I didn't find a good way to implement it. The font is not monospaced so using a single label with placeholders like `" "` for inactive mods doesn't guarantee fixed positioning. We could have 4 separate labels and position them manually but the correct positioning could depend on font and display size which doesn't feel robust. If you have any ideas I'd be happy to hear them.
> 
> Hmm.... All painful points. I think to do it right will need to wait on https://docs.lvgl.io/master/layouts/flex.html?highlight=flexbox when we get the Zephyr 3.2 update in. Put each label as a child of a container object using flex layout... and it might just work?

Thanks for pointing that out -- I'll definitely experiment with that when I get a chance.

idesignstuff:
> I tested the current version on real hardware and it's working as intended.

Where can I find directions for testing a pr like this on my hardware?

duckyb:
> > I tested the current version on real hardware and it's working as intended.
> 
> Where can I find directions for testing a pr like this on my hardware?

Edit your `west.yml` like so: https://github.com/duckyb/zmk-urchin/blob/b97f8c669eb78be670b0254605888b3322f76639/config/west.yml

idesignstuff:
Awesome! Thank you!

idesignstuff:
> 

So, is it more complicated to bring in more than one of these? I've already got one PR for split encoders. Can this one be added, or is it one at a time?

`FATAL ERROR: Malformed manifest file: /__w/Klein-zmk/Klein-zmk/config/west.yml 
  Schema file: /usr/local/lib/python3.8/dist-packages/west/manifest-schema.yml
  Hint: manifest:
  remotes:
      # zmk official
    - name: zmkfirmware
      # for dual encoders
    - name: infused-kim
      url-base: https://github.com/infused-kim
      # bravekarma
    - name: caksoylar
      url-base: https://github.com/caksoylar
  projects:
   - name: zmk
      remote: infused-kim
      revision: my-changes/split-encoder
      import: app/west.yml
    - name: zmk
      remote: zmkfirmware
      revision: main
      import: app/west.yml
  self:
    path: config`

duckyb:
> > 
> 
> So, is it more complicated to bring in more than one of these? I've already got one PR for split encoders. Can this one be added, or is it one at a time?
> 
> `FATAL ERROR: Malformed manifest file: /__w/Klein-zmk/Klein-zmk/config/west.yml Schema file: /usr/local/lib/python3.8/dist-packages/west/manifest-schema.yml Hint: manifest: remotes: # zmk official - name: zmkfirmware url-base: https://github.com/zmkfirmware # for dual encoders - name: infused-kim url-base: https://github.com/infused-kim # bravekarma - name: caksoylar url-base: https://github.com/caksoylar projects:
> 
>     * name: zmk
>       remote: infused-kim
>       revision: my-changes/split-encoder
>       import: app/west.yml
>       - name: zmk
>       remote: zmkfirmware
>       revision: main
>       import: app/west.yml
>       self:
>       path: config`

Make you own fork with all the changes you want and add it to `west.yml`

infused-kim:
Great feature, thank you. One thing to consider: Users may also want to display caps lock and caps word in the same widget. 

To accommodate that, you might need more than one character and separator after all.

duckyb:
> Great feature, thank you. One thing to consider: Users may also want to display caps lock and caps word in the same widget.
> 
> To accommodate that, you might need more than one character and separator after all.

Is caps word a native modifier? If it isn't I don't think it would make sense to add it, however good point about caps lock, and also num lock could be an interesting addition.

infused-kim:
> Is caps word a native modifier? If it isn't I don't think it would make sense to add it, however good point about caps lock, and also num lock could be an interesting addition.

The current implementation uses shift under the hood, but hopefully https://github.com/zmkfirmware/zmk/pull/1485 will be merged soon which will implement caps word by toggling capslock.

Despite that, I still think there is value in seeing whether capslock or capsword is enabled and I think it's better to have it all in one widget than to force the user to use one for modifiers and one for locks.

caksoylar:
Those two would indeed be very useful to have. Caps lock and other locks would ideally need #999 to get the status of the lock from the computer properly. Caps word is internal to the keyboard so that's easier but it doesn't expose an event to subscribe to or a function to check its status, as far as I can tell? I think it would be good if the basic modifier functionality gets merged first, then can be extended to these over time.

When I get a chance to update this, my plan in order of priority is:
- Add a Kconfig to customize modifier letters
- Adapt to Zephyr 3.2 when it gets merged
- Experiment with flex layouts in the new LVGL to fix positioning

mike1808:
@caksoylar sorry for the ping. Are you planning to bringing this to finish line with the current capabilities?

caksoylar:
Hey, thanks for the reminder. I will rebase this in the next day or two and test locally, I think it should be working OK. I actually implemented the config option and ported to Zephyr 3.2 sometime ago. Still haven't checked out "Experiment with flex layouts in the new LVGL to fix positioning" item, I'll do that if I find some time.

caksoylar:
I'll reopen if I rebase this. Unfortunately I don't have an easy way to test this on hardware anymore.

PR #1563, opened 2022-12-01
feat(display): Add config to show battery level in percentage

caksoylar:
Add a new Kconfig `CONFIG_ZMK_WIDGET_BATTERY_STATUS_SHOW_PERCENTAGE` for battery status widget. When set, show battery level in percentage text instead of using battery icons.

Unfortunately it looks to me there can be issues with widgets overlapping while charging with nice!view resolution and font sizes. This is an emulation, haven't tested on the actual shield yet:
image

### DIFF
--- a/app/src/display/widgets/Kconfig
+++ b/app/src/display/widgets/Kconfig
@@ -15,6 +15,13 @@ config ZMK_WIDGET_BATTERY_STATUS
     default y if BT
     select LVGL_USE_LABEL
 
+if ZMK_WIDGET_BATTERY_STATUS
+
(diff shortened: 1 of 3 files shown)


duckyb:
| charging | wireless |
| --- | --- |
| ZVE02725 (Medium) | ZVE02723 (Medium) |

Works fine 👍🏻

duckyb:
It's good that while charging only the bluetooth logo appears in the top right corner, so overlapping is not likely to occur in real world usage.

petejohanson (approved):
Love it! Thanks for the addition.

PR #1564, opened 2022-12-02
fix(behaviors): Mark sticky key release-after-ms required

nickcoutsos:
While experimenting with custom behaviors based on `zmk,behavior-sticky-key` I noticed that `release-after-ms` is required, and its absence will cause the build to fail.


petejohanson (approved):
Thanks, good catch!

PR #1566, opened 2022-12-02
fix(docs) - updated Windows env setup to windows 11 images

Hello9999901:


joelspadin:
There's now a Zephyr SDK for Windows. Should we update the documentation to refer to that instead of the GNU ARM embedded toolchain first? You don't need to adjust environment variables when using the Zephyr SDK.

Hello9999901:
Ah I see. Definitely. I am not a Windows user, I just had some time and noticed that the documentation images needed a little DLC. I don't think I am experienced enough to write the documentation for that though.

petejohanson:
> There's now a Zephyr SDK for Windows. Should we update the documentation to refer to that instead of the GNU ARM embedded toolchain first? You don't need to adjust environment variables when using the Zephyr SDK.

I definitely feel we need to do this for Windows *and* macOS, now that the SDK is available for those OSes.

Hello9999901:
I agree too. I will have some more time this week, I will try and write one up for macOS. I'm not very experienced on Windows, so I probably will not be attempting that for the time being.

petejohanson:
Superseded by #1621. Thanks for the work on this regardless!

PR #1571, opened 2022-12-06
refactor(kscan): batch GPIO reads by port

joelspadin:
This is an alternative solution to #1567 which does not require any dynamic memory.

Changed the GPIO matrix and direct GPIO key scan drivers to do a single read per port instead of one read per pin. This is much more efficient for some types of GPIO drivers, such as I2C GPIO expanders.

To accomplish this with minimal overhead, we now sort input pins by port at driver init. if we iterate through the pins in the sorted order, all pins on the same port are consecutive, so we only need to read each port once the first time we see it.

### DIFF
--- a/app/drivers/kscan/CMakeLists.txt
+++ b/app/drivers/kscan/CMakeLists.txt
@@ -5,6 +5,7 @@ zephyr_library_named(zmk__drivers__kscan)
 zephyr_library_include_directories(${CMAKE_SOURCE_DIR}/include)
 
 zephyr_library_sources_ifdef(CONFIG_ZMK_KSCAN_GPIO_DRIVER debounce.c)
+zephyr_library_sources_ifdef(CONFIG_ZMK_KSCAN_GPIO_DRIVER kscan_gpio.c)
 zephyr_library_sources_ifdef(CONFIG_ZMK_KSCAN_GPIO_MATRIX kscan_gpio_matrix.c)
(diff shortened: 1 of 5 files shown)


petejohanson (commented):
This all seems reasonable... I'm a bit torn about the *potential* impact on more constrained devices. Can we get some hard numbers on change to ROM/RAM based on these changes? Based on that, I feel like we can make the call on just making this the new default, or putting this behind a DTS boolean prop to trigger the new behavior.

joelspadin:
End of build memory report on my numpad, which has a 4x6 matrix with 6 inputs:

Before change:
```
           FLASH:      208800 B       844 KB     24.16%
            SRAM:       60846 B       256 KB     23.21%
        IDT_LIST:          0 GB         2 KB      0.00%
```

With change:
```
           FLASH:      209224 B       844 KB     24.21%
            SRAM:       60958 B       256 KB     23.25%
        IDT_LIST:          0 GB         2 KB      0.00%
```

joelspadin:
I finally took the time to verify that the direct pin implementation works when using multiple ports.

tahnok:
Seems to be working well to speed up my ergodox (using MCP2038). I have only used it for a few minutes so far, but I'll report back after I've used it for a few days

[config for reference](https://github.com/tahnok/zmk-config)

joelspadin:
Rebased.

petejohanson (approved):
Ok, just tossed this on my daily driver and it's working without issue. Have had other reports on Discord of this working well for folks using IO expanders with inputs on them, where this can be a real win.

PR #1582, opened 2022-12-11
feat: Add global macro timing configs

joelspadin:
Added `CONFIG_ZMK_MACRO_DEFAULT_WAIT_MS` and `CONFIG_ZMK_MACRO_DEFAULT_TAP_MS` to set global defaults for the wait-ms and tap-ms properties of macros.

Also reduced the default timings for macros, since it's been reported many times that 100 ms is too slow.


petejohanson (approved):
Had at least one tester on Discord. Lots of folks already manually dropping these for sane use. Thanks!

PR #1584, opened 2022-12-14
fix(shields): Fix col-offset for splitkb.com corne

petejohanson:
Same column offset for both transforms, since we use the same pins for both, just apply a different transform.

Fixes: 1570


petejohanson:
Got confirmation on the splitkb.com discord server that this fix works: https://discord.com/channels/574598631399751680/574738351836626944/1052414609203675156

Merging.

PR #1585, opened 2022-12-14
feat(boards): Add battery sensor for Makerdiary M60 keyboard

hyx0329:


petejohanson:
@hyx0329 Seems reasonable to me. You've tested on real hardware? Where did you get the resistor values?

hyx0329:
@petejohanson I got those values from the customer service of the store on taobao(presumably the developers). Unfortunately, it seems they are not willing to share the schematics.

petejohanson (approved):
Thanks!

PR #1586, opened 2022-12-14
fix(behaviors): `quick-release` only release one sticky key

### LINKED ISSUE #1365: Sticky layer not released with sticky key and key with modifier
To reproduce:

1. Press and immediately release `&sk LSFT` on layer 0 with `quick-release` enabled for `&sk`.
2. Press and immediately release `&sl 2` to go to layer 2.
3. Press and immediately release `&kp RA(N)` on layer 2 twice - it appears that keymap does not revert to layer 0 after the first press though it does revert after the second press.

A few observations:

- This issue does not occur if `&kp LSFT` is used instead of `&sk LSFT`.
- This issue does not occur if `&kp N` is used instead of `&kp RA(N)`.
- This issue does not occur if `&sk LSFT` is held throughout 2 and 3.
- This issue does not occur without `quick-release` for `&sk`.

This combination is useful for languages such as Spanish where you can have `&kp RA(N)` defined for ñ and use it in conjunction with `&sk LSFT` to get Ñ.

nguyendown:
Fix #1365

`struct zmk_keycode_state_changed *ev` may not keep its values after `ZMK_EVENT_RAISE_AFTER()`. Performing conditions on `*ev` in the next interations may produce unexpected results.

f720a0995e683e1f539d63ca6947dd824243740a
Added `quick-release` tests.
230dad3b31a4408bf700d9df2d550a93d6dffd1a
Added debug logs to several sections of the code.
```
west test tests/sticky-keys/10-callum-mods-no-sl-qr/
less build/tests/sticky-keys/10-callum-mods-no-sl-qr/keycode_events_full.log
```

First iteration. Before `ZMK_EVENT_RAISE_AFTER()`.
`build/tests/sticky-keys/10-callum-mods-no-sl-qr/keycode_events_full.log:75-81`.
```
zmk: sticky_key_keycode_state_changed_listener: sticky_event
zmk: sticky_key_keycode_state_changed_listener: i: 0
zmk: sticky_key_keycode_state_changed_listener: current sk: 700e0
zmk: sticky_key_keycode_state_changed_listener: ev->state: 1
zmk: sticky_key_keycode_state_changed_listener: pressed
zmk: sticky_key_keycode_state_changed_listener: timer_started
zmk: sticky_key_keycode_state_changed_listener: reraise
```

This is the next iteration after `ZMK_EVENT_RAISE_AFTER()`.
`build/tests/sticky-keys/10-callum-mods-no-sl-qr/keycode_events_full.log:109-113`.
```
zmk: sticky_key_keycode_state_changed_listener: i: 1
zmk: sticky_key_keycode_state_changed_listener: current sk: 700e1
zmk: sticky_key_keycode_state_changed_listener: ev->state: 0
zmk: sticky_key_keycode_state_changed_listener: released
zmk: sticky_key_keycode_state_changed_listener: i: 2
```

A temporary fix would be to create a local copy of the struct.
de0b4a5c2bd99d9624792d5068432d9388881b48

However, I do not see any local copies of the struct in the source code so I'm not sure if this is the right approach. Is there a way to tell ZMK event management to keep `ev` values?

The two tests added are the existing tests with `quick-release` enabled.


ldebritto:
Just implemented this fix and now it's working perfectly as intended.

To clarify "just as intended", before it, I was having this problem when I the sequence 
```
&sk LSHFT
&sk ALT 
&kp A
```
The desired output was:
```
Å // ⌥+⇧+a
a // a
```
However, I was getting this:
```
Å // ⌥+⇧+a 
å // ⌥+a
```

Before, my diagnosis software pointed out that shows the first `&kp A` key down with the flags `left_shift` and `left_option` while  the second one would still keep `left_option`, thus producing the special character å. 

In short, quick-release in my `&sk` behavior wasn't releasing both mods as it should per zmk documentation.

Now it's releasing both `&sk LALT` and `&sk LSHFT` on the first `&kp A` down.

ldebritto (approved):

joelspadin:
While working in another area of the code, I made a change which randomly broke a sticky key test. I tracked it down to the same issue, but I wasn't aware of this PR, so I made a nearly identical fix at #1745.

These events don't contain any dynamically allocated data, so I think making copies by value is fine. I would suggest marking the copy as `const` though, since that would prevent anyone from making a future change that causes this problem again. A comment explaining why we need to make a copy of the struct could be helpful too.

nguyendown:
I would revert the changes in `behavior_sticky_key.c` as #1745 describes the issue more accurately. This pull request #1586 will just include the two tests.

petejohanson:
@nguyendown I've just merged #1745, can you please rebase this onto `main` again so we can run these new tests against the updates code, and verify things are all copacetic before merging?

Thanks both for the work on this!

nguyendown:
Rebased. Tested.
```
PASS: tests/sticky-keys/10-callum-mods
PASS: tests/sticky-keys/10-callum-mods-quick-release
PASS: tests/sticky-keys/10-sl-sl-kp
PASS: tests/sticky-keys/1-os-dn-up
PASS: tests/sticky-keys/2-os-dn-up-kcdn-kcup
PASS: tests/sticky-keys/2-os-dn-up-kcdn-kcup-quick-release
PASS: tests/sticky-keys/2-sl-dn-up-kcdn-kcup
PASS: tests/sticky-keys/3a-os-dn-kcdn-kcup-up
PASS: tests/sticky-keys/3b-os-dn-kcdn-up-kcup
PASS: tests/sticky-keys/4-os-dn-up-kcdn-timer-kcup
PASS: tests/sticky-keys/5-os-kcdn-dn-kcup-up
PASS: tests/sticky-keys/7-os-dn-up-kc1dn-kc2dn-kc1up-kc2up
PASS: tests/sticky-keys/8-lsk-osk-combination
PASS: tests/sticky-keys/8-lsk-osk-combination-quick-release
PASS: tests/sticky-keys/9-sk-dn-up-dn-up
```

petejohanson (approved):
Thanks!

PR #1590, opened 2022-12-15
fix(shields): Fix keycodes that deffer from the default keymap

OmriKaplan:
- Asterisk in layer 1 was mapped to 8
- Underscore in layer 1 was mapped to minus
- Right GUI in layer 0 was mapped to right Alt
- Space in layer 1 was mapped to Return
- Plus was mapped with KP_PLUS


petejohanson (approved):
Thanks!

OmriKaplan:
Thanks for the wonderful work on ZMK!

PR #1592, opened 2022-12-16
fix(setup): Fix issues for using Corne-ish Zen with the setup scripts

caksoylar:
- Add the `.conf` file from the [official config repo](https://github.com/LOWPROKB/zmk-config-zen-2/blob/main/config/corneish_zen.conf) so that the ZMK setup script can copy it to the user config folder and the produced repo can match the official config repo
- Make setup scripts also look for `keyboard_dir.{keymap,conf}` files, so that for revisioned board/shields like `corneish_zen_v2` we can retrieve `corneish_zen.keymap` and `corneish_zen.conf` files if the versioned filenames don't exist


petejohanson (approved):
Thanks!

PR #1593, opened 2022-12-17
feat(boards): Add Corne-ish Zen v1

caksoylar:
This PR does some minor refactoring for existing Corne-ish Zen v2 files and adds files for Corne-ish Zen v1 (for GB rounds 1&2).
- Spacing and duplicated Kconfig clean up
- Set a chosen transform in the board files (in case someone writes their keymap from scratch and doesn't do that)
- Tweak `compatible` and `model` values in the DT

v1's differences compared to v2 are:
- Battery sensing with the fuel gauge, with power domain support through `int-gpios` that was necessary for deep sleep in Zephyr 3.x
  - Does not enable or use additional sensor channels from the fuel gauge chip
- Uses `spi2` instead of `spi0` since fuel gauge uses `i2c0`

For v2 refactorings the produced DTS and Kconfig files after a build are identical to `main` except for `model` and `compatible` values that removed v2. Tested right side on real v1 hardware, although I have been using a (non-unified) branch with the same config until my v1 left side died.

## Board/Shield Check-list


app/boards/arm/corneish_zen/Kconfig.board
```diff
@@ -3,10 +3,18 @@
 # SPDX-License-Identifier: MIT
 #
 
+config BOARD_CORNEISH_ZEN_V1_LEFT
```
petejohanson, line 6: To avoid all the annoying `||` in all the left/right checks, it might be worthwhile to create a `BOARD_CORNEISH_ZEN_LEFT` that gets `select`ed by the V1 and V2 actual boards, then use those abstract Kconfig settings in the checks everywhere.

app/boards/arm/corneish_zen/corneish_zen_v1_right.dts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+&i2c0 {
+	status = "ok";
+	compatible = "nordic,nrf-twim";
+	sda-pin = <16>;
+	scl-pin = <13>;
+	clock-frequency = <100000>; 
+
+	fuelgauge: bq274xx@55 {
```
petejohanson, line 62: I thought this fuel gauge sensor driver didn't like the way our battery code probes for the value?

petejohanson (commented):
A few comments.

app/boards/arm/corneish_zen/Kconfig.board
caksoylar: I tried to follow the [Kyria approach](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/Kconfig.defconfig) but it resulted in circular dependencies; I think `Kconfig.board` is processed differently compared to `Kconfig.shield` files. I'll try to find a board example if I can and see if I can refactor some other way 👍

app/boards/arm/corneish_zen/corneish_zen_v1_right.dts
caksoylar: To the best of my understanding the sensor exposes additional channels that could be read by the ZMK `battery.c` code, as added in #897. However just fetching the battery level channel (which is the only one utilized) seems to work OK without that PR and the current `battery.c` code. I have been testing this for 5-6 weeks and the reported battery levels have been pretty sensible.

manna-harbour:
Thanks @caksoylar!

Currently testing with this build (current ZMK main, mousekeys enabled, layer widget disabled): https://github.com/manna-harbour/miryoku_zmk/actions/runs/3727371035.

app/boards/arm/corneish_zen/Kconfig.board
caksoylar: I added left and right Kconfigs following [`efr32_radio` board in Zephyr](https://github.com/zmkfirmware/zephyr/blob/v3.0.0%2Bzmk-fixes/boards/arm/efr32_radio/Kconfig.board), that cleaned it up a little.

manna-harbour:
Haven't noticed any issues since I started testing.

caksoylar:
We had another user confirm on the LPKB Discord so I am publishing the PR for review.

lulzsun:
Confirmed to be working without any issues on my Corne-ish Zen v1, no issues so far either. Thanks!

app/boards/arm/corneish_zen/corneish_zen_v1_left_defconfig
```diff
@@ -0,0 +1,92 @@
...
+#CONFIG_LOG_STRDUP_BUF_COUNT=160
+#CONFIG_I2C_LOG_LEVEL_DBG=y
+#CONFIG_SPI_LOG_LEVEL_DBG=y
+#CONFIG_DISPLAY_LOG_LEVEL_DBG=y
+#CONFIG_LVGL_LOG_LEVEL_DBG=y
+#CONFIG_LVGL_USE_DEBUG=y
+#CONFIG_SENSOR_LOG_LEVEL_DBG=y
+
+# Turn on USB CDC ACM device
+CONFIG_USB_DEVICE_STACK=y
```
petejohanson, line 81: Is this really necessary? Seems odd to enable all of thess by default.

app/boards/arm/corneish_zen/corneish_zen_v1_right.dts
petejohanson: Thanks. Makes sense.

app/boards/arm/corneish_zen/corneish_zen_v1_right_defconfig
```diff
@@ -0,0 +1,86 @@
...
+#CONFIG_SENSOR_LOG_LEVEL_DBG=y
+
+# Enable serial console
+CONFIG_SERIAL=y
+CONFIG_CONSOLE=y
+CONFIG_UART_INTERRUPT_DRIVEN=y
+CONFIG_UART_LINE_CTRL=y
+
+# Enable USB UART
+CONFIG_UART_CONSOLE=y
```
petejohanson, line 86: Same question. Really all necessary?

petejohanson (commented):
Just a couple questions. Thanks.

app/boards/arm/corneish_zen/corneish_zen_v1_left_defconfig
caksoylar: I had just copied the same settings from v2 for v1 but #1690 is relevant here. I will do some testing with the changes there to see if everything works well with or without logging enabled and drop a comment there. After that we can carry over the same changes to v1.

petejohanson: Testing ok on this?

caksoylar: Yes, just tested and dropped a review on #1690, all seems good. So I pushed the same changes to v1 files here.

caksoylar:
Rebased and have been testing (just the right side) OK for a few days now.

petejohanson (approved):
Thanks!

petejohanson:
@caksoylar Looks like the pre-commit check is still not happy with this.

caksoylar:
I hadn't managed to install pre-commit locally properly, hope it's all good now!

petejohanson (approved):

PR #1594, opened 2022-12-17
Add Euro keycode

MvEerd:
A user was looking for the `EURO` keycode so I added this alias when I saw it wasn't available yet.
I realize there are no other Alt-gr keycodes yet, but it seems in a similar vain as shifted keycodes to me so this may be considered useful, perhaps most 'Alt-gr'-ed keycodes could be added later as well but this one seems common enough to warrant adding it by itself, any thoughts on whether this is something you'd like to support?


MvEerd:
It was brought to my attention that 'US Standard' layout definitions are used by default and not 'US International' like I assumed so feel free to close this if the intent is to just stick to standard.

caksoylar:
Closing this one then, note that https://github.com/joelspadin/zmk-locale-generator/releases/download/v1.3.2/keys_us_international.h defines the keycode in the same way and can be used with the US Intl layout.

PR #1600, opened 2022-12-20
fix(docs): added property to disable status of encoder in dtsi for a new shield.

as29:
**Documentation update**: added property to disable status of encoder in dtsi for a new shield. Per my experience of creating a new shield, encoders wouldn't work unless disabled in dtsi and then enabled through config.


caksoylar (approved):
Thank you for the contribution! It seems to be necessary following our discussion [in Discord](https://discord.com/channels/719497620560543766/813882537436905552/1054604544157429821).

Do you mind renaming the PR to be in the [conventional commits format](https://www.conventionalcommits.org/en/v1.0.0/)? You could use something like `fix(docs): Disable encoders in dtsi`.

as29:
renamed PR to follow the [conventional commits format](https://www.conventionalcommits.org/en/v1.0.0/)

dxmh (approved):

PR #1301, opened 2022-05-14
feat(drivers): add driver for MAX17048 fuel gauge

zhiayang:
Adds a driver for MAX17048 I2C fuel gauge. Works on real hardware.

PR #1602, opened 2022-12-21
fix(driver): Fix potentially buggy read/write routines for max7318 driver

zhiayang:
Fix the potentially buggy implementation in I2C read/write routines in the MAX7318 driver. It used the same style of read/write as #1301, which was identified as buggy.

While nobody has yet raised an issue about it, this PR changes the register read/write functions to use the same (correct) style as the fix in #1301.


petejohanson:
Seems reasonable. Is this tested on real hardware?

petejohanson:
Please run prettier though to fix code formatting.

zhiayang:
fixed!

petejohanson (approved):
Thanks!

PR #1606, opened 2022-12-23
feat(docs): Include Zephyr™ SDK Bundle install instructions for macOS

Hello9999901:


docs/docs/development/setup.md
```diff
@@ -331,13 +331,33 @@ Since the Zephyr™ SDK is not available for Windows, we recommending following
...
+- `~/.local`
+- `~/.local/opt`
+- `~/bin`
+- `/opt`
+- `/usr/local`
+
+Note that the install includes the directory `zephyr-sdk-X.XX.X`
+:::
+
+Download the [latest Zephyr SDK™ Bundle](https://github.com/zephyrproject-rtos/sdk-ng/releases) to a directory of your choice. If the host architecture is Apple Silicon based, download the aarch64 version.
```
petejohanson, line 349: For official docs, I'd recommend we link to a static, known-good version like we do for other operating systems.

petejohanson (changes requested):
One comment on the link to the SDK.

Thanks!

docs/docs/development/setup.md
Hello9999901: I think I fixed it, let me know if I can improve upon it!

PR #1610, opened 2022-12-25
fix(behaviors): allow saving sticky keys with the same position

### LINKED ISSUE #1421: bug: ignore-modifiers ignored in macro
The `ignore-modifiers` property of `sl` is ignored when specified as part of a macro. 

Here's a unit-test reproducing the bug: 12c87c459627d8e8ad7969064c631bb348f413d7

Specifically, when `sl 1` and `sk LSHFT` are called via individual key presses (the commented lines in the unit test), both the sticky layer and the sticky shift are applied to the next keypress, resulting in `LSHFT + B`:
```
pressed: usage_page 0x07 keycode 0xE1 implicit_mods 0x00 explicit_mods 0x00
pressed: usage_page 0x07 keycode 0x05 implicit_mods 0x00 explicit_mods 0x00
released: usage_page 0x07 keycode 0xE1 implicit_mods 0x00 explicit_mods 0x00
released: usage_page 0x07 keycode 0x05 implicit_mods 0x00 explicit_mods 0x00
```

However, when the same two behaviors are called via the macro, then only the sticky-shift remains active and we get the output from the default layer (`LSHFT + A`):
```
pressed: usage_page 0x07 keycode 0xE1 implicit_mods 0x00 explicit_mods 0x00
pressed: usage_page 0x07 keycode 0x04 implicit_mods 0x00 explicit_mods 0x00
released: usage_page 0x07 keycode 0xE1 implicit_mods 0x00 explicit_mods 0x00
released: usage_page 0x07 keycode 0x04 implicit_mods 0x00 explicit_mods 0x00
```

nguyendown:
When a sticky key is pressed, a duplicated sticky key will be searched for and released. However, currently only `position` is used to search for a sticky key, which causes issues because keys with different layers may have the same position.

Adding checks on both `position` and `layer` is not enough because key events generated by macro have the same position and layer.
7a4f3a261d0c0f6816553727acb45f6ac05d15ad

I have been considering that two sticky keys `&sk LEFT_SHIFT` at different positions should be treated as the same sticky key. The use of `position` or `layer` is not necessary because `param1` has keycode, usage page, and mods, which are sufficient to identify sticky keys.

Fix #1421

### DIFF
--- a/app/src/behaviors/behavior_sticky_key.c
+++ b/app/src/behaviors/behavior_sticky_key.c
@@ -26,8 +26,6 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #define ZMK_BHV_STICKY_KEY_MAX_HELD 10
 
-#define ZMK_BHV_STICKY_KEY_POSITION_FREE UINT32_MAX
-
 struct behavior_sticky_key_config {
     uint32_t release_after_ms;
     bool quick_release;
@@ -36,6 +34,7 @@ struct behavior_sticky_key_config {
 };
 
 struct active_sticky_key {
+    bool is_active;
     uint32_t position;
     uint32_t param1;
     uint32_t param2;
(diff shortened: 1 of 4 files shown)


petejohanson:
This... Seems like a fairly reasonable fix for this. Let me consider if there's any potential downsides to this approach.

@okke-formsma any concerns?

caksoylar:
I forgot that this PR existed when I developed #2758, sorry! I just realized it after seeing it linked on #1421. The addressed issue should be fixed by the former now.

Issue #1569, opened 2022-12-04
docs: Setup instructions for macOS on Apple silicon


dxmh:
The `gcc-arm-embedded` requirement for ZMK/Zephyr 3.0 on macOS seems to be available for x86-64 host architecture only ([source](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads)); it is not available for AArch64 macOS hosts (i.e. Macs with Apple silicon, such as M1 and M2). 

According to a quick search on the ZMK Discord server, Apple silicon users aren't having much luck running it via Rosetta.

It would be good to document this limitation with the [macOS setup instructions](https://github.com/zmkfirmware/zmk/blob/001105e31844f0f59b82a2e71a4c3ada980b931b/docs/docs/development/setup.md#gnu-arm-embedded-1) and possibly suggest an alternative for AArch64 macOS hosts (such as Docker (https://github.com/zmkfirmware/zmk/pull/879), or the [Zephyr SDK for macOS](https://github.com/zephyrproject-rtos/sdk-ng#releases)).

Hello9999901:
I have a PR for something similar right now. I can add this in with more verbose detail.

caksoylar:
This is now covered by #1621, which links to Zephyr docs that includes specific instructions for Apple Silicon.

PR #1565, opened 2022-12-02
feat(docs) - Added macos specific .zephyrrc configuration

Hello9999901:

PR #1621, opened 2023-01-08
docs: update dev env setup guide

joelspadin:
Renamed the "Basic Setup" page to "Toolchain Setup" to make it clearer that this is for setting up a development environment for local builds.

Linked to the relevant Zephyr setup documentation instead of duplicating that information in our guide. This also switches Windows and macOS setup to using the Zephyr SDK, which is much easier to install.

Updated the command for installing Zephyr's Python dependencies to be OS specific, as Windows and macOS don't need the --user flag.


dxmh (approved):
Preview: https://deploy-preview-1621--zmk.netlify.app/docs/development/setup

This looks good to me. I think the new title of "Toolchain Setup" will help to avoid confusion with the "[Installing ZMK](https://deploy-preview-1621--zmk.netlify.app/docs/user-setup)" page. I also like how the content is now simplified by linking to the Zephyr docs.

I haven't executed the updated setup instructions to confirm them, so will leave this PR open for others to comment.

There are some other open issues/PRs somewhat related to this toolchain setup:

- https://github.com/zmkfirmware/zmk/issues/1569/
- https://github.com/zmkfirmware/zmk/pull/1606/
- https://github.com/zmkfirmware/zmk/pull/1566/
- https://github.com/zmkfirmware/zmk/pull/1565/
- https://github.com/zmkfirmware/zmk/pull/879/

I believe this PR closes/incorporates/supersedes some of the above, so it would be good for us to circle back and update/close these once this is merged.

docs/docs/development/setup.md
```diff
@@ -421,19 +251,13 @@ All subsequent steps must be performed from the VS Code terminal _inside_ the co
 </TabItem>
 </OsTabs>
 
-#### Initialize West
+### Initialize the Application
```
caksoylar, line 254: Might want to reduce the header level of "Step into the repository" above too, to match.

caksoylar: Also we could just make above step two tabs `docker` and `others`, to reduce duplication.

caksoylar (approved):
Thanks Joel, I think this is very helpful as well, and less documentation ZMK hosts and instead refers to Zephyr the better. 

One thing to highlight is I see we are linking to `3.2.0` versions of the Zephyr docs; since the Zephyr update is going to happen relatively soon this is probably fine but we might want to keep in mind to update in the future.

docs/docs/development/setup.md
joelspadin: Unfortunately, if we reduce the tabs to just "docker" and "others", then it can't sync your OS selection with the other tab groups.

joelspadin: > Might want to reduce the header level of "Step into the repository" above too, to match.

Fixed.

joelspadin:
I also updated the IDE integration section, as the old instructions for determining the compiler path no longer work. Sourcing the zephyr-env script no longer sets up the compiler path environment variable, so I have you run a cmake script that prints the path instead.

docs/docs/development/setup.md
caksoylar: > Unfortunately, if we reduce the tabs to just "docker" and "others", then it can't sync your OS selection with the other tab groups.

That's true, although we already have a few other tabs that just divide by Linux/Windows/mac OS etc.

petejohanson (approved):
Thanks @joelspadin !

Issue #1494, opened 2022-10-14
Feature request: preserve locality outside of direct keymap bindings


manna-harbour:
Combos do not preserve locality as noted at https://zmk.dev/docs/features/combos: 

> SOURCE-SPECIFIC BEHAVIORS ON SPLIT KEYBOARDS
Invoking a source-specific behavior such as one of the [reset behaviors](https://zmk.dev/docs/behaviors/reset) using a combo will always trigger it on the central side of the keyboard, regardless of the side that the keys corresponding to key-positions are on.

It seems that this also applies to tap dance and mod morph. I guess it applies to everything other than a direct keymap binding and that the locality system will need to be extended.
  
In the mean time, a note should be added to the reset behavior docs.

bobobo1618:
Could this be resolved by setting global locality for the combo behaviours like it was for RGB underglow and ext power in https://github.com/zmkfirmware/zmk/pull/547/files#diff-fa3059ddb7abee58e50dbc5bb6b0218e6e4948d3742539e4f3a9aa67a1eb7278R149 ?

bobobo1618:
This bug also applied to global behaviours like ext power and RGB underglow for me. Without https://github.com/bobobo1618/miryoku_zmk/commit/6ba964f12a9d8138437cffacc487d4827e30fa8f, the changes only applied to the central half.

manna-harbour:
Updating this issue:

All behaviors using the locality system (https://github.com/zmkfirmware/zmk/pull/547) do not work as intended on split keyboards when the behavior is not bound directly in the keymap. If the behavior is called from any other behavior the locality information is lost and the behavior is run only on central. I.e. source-specific behaviors (https://zmk.dev/docs/behaviors/reset) invoked from indirect bindings on peripheral will run on central instead of peripheral, and global behaviors (https://zmk.dev/docs/behaviors/underglow, https://zmk.dev/docs/behaviors/power) invoked from indirect bindings will run only on central instead of on central and peripheral.

The locality system should be extended to preserve locality information when invoked indirectly.

In the mean time, the docs (https://github.com/zmkfirmware/zmk/pull/1124) should have a warning added for all behaviors with locality (pages linked above), and the current warning at https://zmk.dev/docs/features/combos should be removed. 

Work-arounds have been added to Miryoku ZMK:

- https://github.com/manna-harbour/miryoku_zmk#double-tap-boot
- https://github.com/manna-harbour/miryoku_zmk#global-shift-functions

[...]

Issue #1347, opened 2022-06-16
&rgb_ug RGB_COLOR_HSB on split keyboard is not global when used within macro


ivandaho:
When using using the example code as seen in https://zmk.dev/docs/behaviors/macros#layer-activation--more for **Layer + Underglow Color**, underglow color is only set for the central side of the keyboard during macro execution

- Color for peripheral side is unaffected by the macro
- If i change underglow color directly with `&rgb_ug RGB_COLOR_HSB(86,81,85)` outside of a macro, it changes for both sides as expected

Tested on `main` branch, built via docker method for Corne/NRFMicro
Unsure if any other `&rgb_ug` behavior is affected

reproducable in https://github.com/ivandaho/zmk/tree/corne-ivan, in case you have a corne/NRFMicro
- run `./buildcorne.sh` in `app` directory
- flash firmware
- turn on LEDs by holding these buttons on the central/left side down in this order
  - leftmost button on bottom row
  - rightmost button on bottom row
  - "D" key (4th button on 2nd row)
- press either:
  - leftmost button on bottom row (central/left side)
  - middle button on bottom row (peripheral/right side)

tokazio:
Same issue here https://github.com/tokazio/zmk-config-sofle

caksoylar:
Most likely related to #1494

caksoylar:
@aDogCalledSpot You can test #2409 which should fix this issue.

caksoylar:
This should be fixed by #2409, please report if there's an issue.

PR #1630, opened 2023-01-17
Resolve preserve locality outside of direct keymap bindings

tokazio:
Seems to resolve https://github.com/zmkfirmware/zmk/issues/1494, at least resolve https://github.com/zmkfirmware/zmk/issues/1347

### DIFF
--- a/app/include/zmk/split/bluetooth/central.h
+++ b/app/include/zmk/split/bluetooth/central.h
@@ -6,3 +6,5 @@
 
 int zmk_split_bt_invoke_behavior(uint8_t source, struct zmk_behavior_binding *binding,
                                  struct zmk_behavior_binding_event event, bool state);
+
+int zmk_run_behavior(struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,uint8_t source,bool pressed);
--- a/app/src/keymap.c
+++ b/app/src/keymap.c
@@ -179,17 +179,23 @@ int zmk_keymap_apply_position_state(uint8_t source, int layer, uint32_t position
         .timestamp = timestamp,
     };
 
-    LOG_DBG("layer: %d position: %d, binding name: %s", layer, position,
-            log_strdup(binding.behavior_dev));
+    return zmk_run_behavior(&binding,event,source,pressed);
+}
 
-    behavior = device_get_binding(binding.behavior_dev);
+int zmk_run_behavior(struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,uint8_t source,bool pressed){
+
+    LOG_DBG("layer: %d position: %d, binding name: %s", event.layer, event.position,
+            log_strdup(binding->behavior_dev));
+
+
+    const struct device *behavior = device_get_binding(binding->behavior_dev);
 
     if (!behavior) {
-        LOG_WRN("No behavior assigned to %d on layer %d", position, layer);
+        LOG_WRN("No behavior assigned to %d on layer %d", event.position, event.layer);
         return 1;
     }
 
-    int err = behavior_keymap_binding_convert_central_state_dependent_params(&binding, event);
+    int err = behavior_keymap_binding_convert_central_state_dependent_params(binding, event);
     if (err) {
         LOG_ERR("Failed to convert relative to absolute behavior binding (err %d)", err);
         return err;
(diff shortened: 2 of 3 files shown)


app/include/zmk/split/bluetooth/central.h
```diff
@@ -5,4 +5,6 @@
 #include <zmk/behavior.h>
 
 int zmk_split_bt_invoke_behavior(uint8_t source, struct zmk_behavior_binding *binding,
-                                 struct zmk_behavior_binding_event event, bool state);
\ No newline at end of file
+                                 struct zmk_behavior_binding_event event, bool state);
+
+int zmk_run_behavior(struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,uint8_t source,bool pressed);
```
petejohanson, line 10: I think this general functionality is useful, but a few things:

* This should be in `keymap.h`, not this header.
* I don't love the name. We are triggering the pressed/released callbacks, not "running" something.

app/src/keymap.c
```diff
@@ -179,17 +179,23 @@ int zmk_keymap_apply_position_state(uint8_t source, int layer, uint32_t position
         .timestamp = timestamp,
     };
 
-    LOG_DBG("layer: %d position: %d, binding name: %s", layer, position,
-            log_strdup(binding.behavior_dev));
+    return zmk_run_behavior(&binding,event,source,pressed);
+}
+
+int zmk_run_behavior(struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,uint8_t source,bool pressed){
```
petejohanson, line 185: `binding` probably should be `const` here. Also, be sure to run `clang-format` on any changed files for consistent formatting.

petejohanson (changes requested):
Thanks for working on this! I think this is a good start, but some tweaks needed.

tokazio:
Thanks for the review.

I see some delay when the trigger comes from the left part of the keyboard that i doesn't see when the trigger comes from the right part.

https://user-images.githubusercontent.com/7676865/214581300-7e72af8d-69ae-4c33-8654-1b414f0c1758.mov

xudongzheng:
> I see some delay when the trigger comes from the left part of the keyboard that i doesn't see when the trigger comes from the right part.

This delay is expected. Since the connection interval is 7.5ms and connection latency is 30 (see https://github.com/zmkfirmware/zmk/blob/main/app/src/split/bluetooth/central.c#L412), it can take up to 7.5ms*(30+1)=232.5ms for the peripheral to receive the update from the central.

petejohanson:
Exactly right... We could lower that connection latency, at the expense of increased power consumption on the peripheral side. (side note: it'd be interesting to renegotiate the connection parameters if the power status of the peripheral changes thanks to being plugged in, to improve responsiveness of this exact scenario....)

vladsolokha:
> Thanks for the review.
> 
> I see some delay when the trigger comes from the left part of the keyboard that i doesn't see when the trigger comes from the right part.

Can you share how you make that behavior happen in your keymap? The layer switch and underglow change? Is there a way I  could do the same and have the rgb colors go back to the original effect when the layer is then switched off or released?

I have some rgb mods on &to layer switches in my keymap here, but I was hoping to do the same for my two main layers 1 and 2. I was hoping to leave them as sticky on tap and momentary on hold with the color changes. 

https://github.com/vladsolokha/zmk-config/blob/main/config/splitkb_aurora_sweep.keymap

tokazio:
You Can found m'y config here https://github.com/tokazio/zmk-config-sofle 
Nothing special

tokazio:
> > I see some delay when the trigger comes from the left part of the keyboard that i doesn't see when the trigger comes from the right part.
> 
> This delay is expected. Since the connection interval is 7.5ms and connection latency is 30, it can take up to 7.5ms*(30+1)=232.5ms for the peripheral to receive the update from the central.
> 
> https://github.com/zmkfirmware/zmk/blob/3d2bd017472c57f4b3860033f989a44f094febaa/app/src/split/bluetooth/central.c#L412

Oh ok, good to know it

xudongzheng:
> Is there a possibilité to force the update when this précise Key is down ?

One option is defining a read service/characteristic on the central that you can read from the peripheral. This should have the current LED state.

After a key event such as a press/release on the peripheral happens, you read the state on central from the peripheral (rather than waiting for the central to write to peripheral). You can try to do this immediately or after a few milliseconds to give the central to process and update the state.

I'm not familiar enough with the Bluetooth specification to be certain that this wouldn't be subject to the latency value, but this is the first thing I would try to reduce the latency.

It's not trivial to implement and I would consider just reducing the latency to a lower value at the cost of some battery life.

caksoylar:
@tokazio in case you missed it, @petejohanson noted a couple items here that needed changing: https://github.com/zmkfirmware/zmk/pull/1630#pullrequestreview-1262910398

tokazio:
After some time of use i'm observing that colors, when i push down the upper/lower keys are not the ones i've programmed. Even the effect have changed. I have now the multicolors wave effect. Maybe there is a conflict with the 'persistence' feature of effect/color that is triggered after a given time

seansfkelley:
Thanks for opening this; this resolved an issue I was having with underglow on the peripheral side. I also see noticeable lag between the central and peripheral sides, but for my usages (which are infrequent), that's fine.

@petejohanson it looks like your comments are only stylistic. Would you accept a PR with the same behaviors and your style points addressed, or would you rather prioritize something like https://github.com/zmkfirmware/zmk/pull/2036 instead? (It claims to resolve this issue as well, though I have not tried it.) I can open a cleaned-up PR if so.

petejohanson:
> Thanks for opening this; this resolved an issue I was having with underglow on the peripheral side. I also see noticeable lag between the central and peripheral sides, but for my usages (which are infrequent), that's fine.
> 
> @petejohanson it looks like your comments are only stylistic. Would you accept a PR with the same behaviors and your style points addressed, or would you rather prioritize something like #2036 instead? (It claims to resolve this issue as well, though I have not tried it.) I can open a cleaned-up PR if so.

I'd rather we clean this up and get it merged, and then pursue #2036 as a next step.

seansfkelley:
Sounds good, thanks. It seems like #2409 is currently active and does just that, so I won't add more noisy PRs just yet, but if that stalls out I'll open another PR like this but with focusing on resolving just the stylistic comments.

caksoylar:
My understanding is that this PR is not actually solving the issue besides modifying only macros to invoke behaviors globally. #2409 should solve both global and source locality invocations for every kind of behavior that invokes other behaviors (currently with the exception of sensors not respecting source locality). As such this PR shouldn't be accepted even without the stylistic issues and I don't intend to stall #2409.

caksoylar:
Closing as it is incorporated into #2409.

PR #1632, opened 2023-01-21
Default ext power

donaldsycamore:
Allow ext_power default to be changed (was on by default)
Set it off by default on Glove80


donaldsycamore:
Wrong repo, not quite ready for zmk yet.

PR #1638, opened 2023-01-28
feat(ble): add behavior to disconnect from BLE profile

chrisandreae:
Adds new functionality and a behavior to disconnect an active BLE connection. The motivation for this is that for some devices like phones, the presence of an active BLE connection results in the onscreen keyboard being selected.

This behavior allows the active connection for a given profile to be explicitly disconnected, by calling [`bt_conn_disconnect()`](https://docs.zephyrproject.org/apidoc/latest/group__bt__conn.html#ga14e7c852b0271781594e742ae509c5d3). The connection will be re-established (via `update_advertising()`), when that profile is re-selected.

I believe this behavior is complementary to the proposed automatic disconnection of non-selected devices. Automatic disconnection without a timeout could be frustrating to use when actively switching between devices, as there's a slight but noticeable delay on re-establishing the connnection. On the other hand if a timeout is used, there might be times when the user wishes to disconnect before the timeout has elapsed. This behaviour would permit that.

**Questions:**

In the case that `OUT_USB` is selected, there is still a BLE profile which is considered active and selected, and is being advertised for. If the user tries to disconnect from that profile, it will be immediately re-connected from the advertisement. This would be awkward in the case that the user was wanting to switch between their computer via USB and their phone via BLE. I'm wondering if it would be most consistent to stop advertising for the selected BLE profile in the case that `OUT_USB` is selected and the USB connection is present and active. This would make selecting the USB output more consistent with selecting a different BLE output.

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -280,9 +280,9 @@ int zmk_ble_prof_disconnect(uint8_t index) {
     int result;
 
     if (!bt_addr_le_cmp(addr, BT_ADDR_LE_ANY)) {
-        return -1;
+        return -ENODEV;
     } else if ((conn = bt_conn_lookup_addr_le(BT_ID_DEFAULT, addr)) == NULL) {
-        return -1;
+        return -ENODEV;
     }
 
     result = bt_conn_disconnect(conn, BT_HCI_ERR_REMOTE_USER_TERM_CONN);
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -271,6 +271,27 @@ int zmk_ble_prof_prev() {
                                ZMK_BLE_PROFILE_COUNT);
 };
 
+int zmk_ble_prof_disconnect(uint8_t index) {
+    if (index >= ZMK_BLE_PROFILE_COUNT)
+        return -ERANGE;
+
+    bt_addr_le_t *addr = &profiles[index].peer;
+    struct bt_conn *conn;
+    int result;
+
+    if (!bt_addr_le_cmp(addr, BT_ADDR_LE_ANY)) {
+        return -1;
+    } else if ((conn = bt_conn_lookup_addr_le(BT_ID_DEFAULT, addr)) == NULL) {
+        return -1;
+    }
+
+    result = bt_conn_disconnect(conn, BT_HCI_ERR_REMOTE_USER_TERM_CONN);
+    LOG_DBG("Disconnected from profile %d: %d", index, result);
+
+    bt_conn_unref(conn);
+    return result;
+}
+
 bt_addr_le_t *zmk_ble_active_profile_addr() { return &profiles[active_profile].peer; }
 
 char *zmk_ble_active_profile_name() { return profiles[active_profile].name; }
--- a/docs/docs/behaviors/bluetooth.md
+++ b/docs/docs/behaviors/bluetooth.md
@@ -14,9 +14,11 @@ When pairing to a host device ZMK saves bond information to the selected profile
 
 A ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
 
-An inactive connected profile can be explicitly disconnected using the `BT_DISC` behavior. This can be helpful in
+An _inactive_ connected profile can be explicitly disconnected using the `BT_DISC` behavior. This can be helpful in
 cases when host devices behave differently when a bluetooth keyboard is connected, for example by hiding their on-screen
-keyboard.
+keyboard. Note that at present the active bluetooth profile will immediately reconnect if disconnected. This is true
+even if OUT_USB is selected. To remain disconnected, another bluetooth profile must be first selected using (e.g.)
+`BT_SEL`.
 
 :::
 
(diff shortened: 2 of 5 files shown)


PabloLION:
I want to specify another motivation if I may:
I always put my mac studio into lock-screen mode since it's basically sleeping mode. with this BLE function showing which device is online, the mac wakes up every once in a while. This is especially annoying when I'm taking a nap in front of the screen. It seems this PR will fix the issue.

app/src/ble.c
```diff
@@ -274,6 +274,27 @@ int zmk_ble_prof_prev() {
...
+int zmk_ble_prof_disconnect(uint8_t index) {
+    if (index >= ZMK_BLE_PROFILE_COUNT)
+        return -ERANGE;
+
+    bt_addr_le_t *addr = &profiles[index].peer;
+    struct bt_conn *conn;
+    int result;
+
+    if (!bt_addr_le_cmp(addr, BT_ADDR_LE_ANY)) {
+        return -1;
```
petejohanson, line 286: Perhaps `ENODEV`?

app/src/ble.c
```diff
@@ -274,6 +274,27 @@ int zmk_ble_prof_prev() {
...
+    } else if ((conn = bt_conn_lookup_addr_le(BT_ID_DEFAULT, addr)) == NULL) {
+        return -1;
```
petejohanson, line 288: Ditto

petejohanson (changes requested):
One comment on the code. This also needs to update the BT behavior docs page to document the new command. See https://zmk.dev/docs/behaviors/bluetooth#bluetooth-command-defines

docs/docs/behaviors/bluetooth.md
```diff
@@ -28,12 +33,14 @@ This will allow you to reference the actions defined in this header such as `BT_
...
+| --------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `BT_CLR`        | Clear bond information between the keyboard and host for the selected profile.                                                                            |
+| `BT_NXT`        | Switch to the next profile, cycling through to the first one when the end is reached.                                                                     |
+| `BT_PRV`        | Switch to the previous profile, cycling through to the last one when the beginning is reached.                                                            |
+| `BT_SEL`        | Select the 0-indexed profile by number. Please note: this definition must include a number as an argument in the keymap to work correctly. eg. `BT_SEL 0` |
+| `BT_DISCONNECT` | Disconnect from the 0-indexed profile by number, if it's currently connected and inactive. Please note: this definition must include a number as an       |
```
caksoylar, line 42: The length of this define kind of sticks out from the other ones. What do you think about abbreviating it, e.g. `BT_DIS` or `BT_DSC`?

docs/docs/behaviors/bluetooth.md
```diff
@@ -13,6 +13,11 @@ computer/laptop/keyboard should receive the keyboard input; many of the commands
 When pairing to a host device ZMK saves bond information to the selected profile. It will not replace this when you initiate pairing with another device. To pair with a new device select an unused profile with `BT_SEL`, `BT_NXT` or `BT_PRV` bindings, or by clearing an existing profile using `BT_CLR`.
 
 A ZMK device may show as "connected" on multiple hosts at the same time. This is working as intended, and only the host associated with the active profile will receive keystrokes.
+
+An inactive connected profile can be explicitly disconnected using the `BT_DISCONNECT` behavior. This can be helpful in
+cases when host devices behave differently when a bluetooth keyboard is connected, for example by hiding their on-screen
+keyboard.
```
caksoylar, line 19: I might not have understood #1998 correctly, but does it imply that this shouldn't be used with the currently active profile when USB endpoint is active? If so, can we add a warning note on that here?

docs/docs/behaviors/bluetooth.md
chrisandreae: I'm torn, because those abbreviations aren't especially transparent; you'd have to check the docs to interpret them. NXT, PRV, and SEL all don't have that problem, they're quite self-explanatory.

docs/docs/behaviors/bluetooth.md
chrisandreae: Well, if the disconnected profile is the currently active one, it'll successfully disconnect but then immediately reconnect, because the currently active BT profile always advertises—even if the USB output is selected. #1998 is about a follow-up to make the pause in advertising when switching away from a BT profile consistent, whether that's switching to another BT profile or switching to USB.

I mentioned it to Pete, and he said he'd prefer to leave that to a follow-up because he has a BT refactor in the works that they'd interfere with, and to add an issue to keep track of it.

I don't have a strong opinion about whether the automatic re-connection should be documented explicitly here. If the advertising behavior is changed later, the documentation can always be updated to match.

docs/docs/behaviors/bluetooth.md
petejohanson: I honestly think `BT_DISC` is descriptive enough, at least as descriptive as `BT_SEL` is, and as long as it's clearly documented, should be fine.

chrisandreae: Okay, `BT_DISC` works for me, I'll make the change

docs/docs/behaviors/bluetooth.md
petejohanson: @chrisandreae I would document that expected behavior here for now, and we can update later as needed. Omitting it may just lead to some confusion on how this works *today* for the active profile.

petejohanson (commented):
Code looks good to me. Happy to merge once the docs addition is in.

docs/docs/behaviors/bluetooth.md
chrisandreae: Sure, I've done that. Does this look OK?

caksoylar (approved):

petejohanson (approved):
Thanks!

PR #1640, opened 2023-01-28
feat(underglow): use float for HSB to RGB calculation

xudongzheng:
Most SoC such as nRF52 do not support double-precision natively. Calculations are faster with single-precision, particularly with FPU.

### DIFF
--- a/app/src/rgb_underglow.c
+++ b/app/src/rgb_underglow.c
@@ -75,15 +75,15 @@ static struct zmk_led_hsb hsb_scale_zero_max(struct zmk_led_hsb hsb) {
 }
 
 static struct led_rgb hsb_to_rgb(struct zmk_led_hsb hsb) {
-    double r, g, b;
+    float r, g, b;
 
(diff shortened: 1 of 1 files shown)


xudongzheng:
Testing: nRF52, 40 LEDs, displaying solid color

```
static void zmk_rgb_underglow_effect_solid() {
	uint32_t start = k_cycle_get_32();

	for (int i = 0; i < STRIP_NUM_PIXELS; i++) {
		pixels[i] = hsb_to_rgb(hsb_scale_min_max(state.color));
	}

	uint32_t end = k_cycle_get_32();
	LOG_ERR("delta %d", end-start);
}
```

- double: median 33 cycles (each cycle 1s/32768)
- float: median 24 cycles
- float-fpu: median 4 cycles

Screen Shot 2023-01-28 at 17 21 14

DuMaM (approved):

petejohanson (approved):
Thanks for the tweaks, and for the number crunching on this to back this up w/ hard numbers.

PR #1645, opened 2023-02-01
feat(shields): Add swept keyboard

equiman:
I made this PR using a `swept` keyboard and tested this new shield using a private repository.

front

PCB and components are available here as Open Source:
https://github.com/AYM1607/swept-crkbd/

## Board/Shield Check-list
 - [X] Kconfig.defconfig file correctly wraps *all* configuration in conditional on the shield symbol
 - [X] Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)


app/boards/shields/swept/Kconfig.shield
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2020 Pete Johanson
```
caksoylar, line 1: Need to update the copyright headers to `2023 The ZMK Contributors` across the files.

app/boards/shields/swept/boards/nice_nano.overlay
```diff
@@ -0,0 +1,33 @@
+#include <dt-bindings/led/led.h>
```
caksoylar, line 1: I'd also consider adding an overlay for other common boards, at least for `nice_nano_v2` until a better solution to #885 is implemented.

equiman: Actually, this board only supports v2, so I'll change it. Thanks

app/boards/shields/swept/Kconfig.shield
equiman: Ok. I'll fix that.

Nicell:
I noticed the repo for this shield says to use the corne firmware in ZMK. Is there any notable difference between this firmware and the corne firmware?

equiman:
Yeah, it was using corne because is the shield with the nearest behavior available on ZMK. 

Swept it a more minimalistic keyboard and for example, doesn't have or need support for LEDs. 

Also, I've added this shield because some users are confused while pairing their keyboards, expecting to see 'Swept' instead of 'Corne'

equiman:
stop working with Zephyr 3.2. I've decided rollout changes and use corne that is officially supported and maintained.

PR #1647, opened 2023-02-02
feat(docs): additional note on bug fixed in macOS v13.1

filterpaper:
Add a note to indicate that the Ventura uf2 bug is fixed in version 13.1


caksoylar (approved):

petejohanson (approved):
Thanks!

PR #1627, opened 2023-01-14
feat(shields): Add splitkb.com Kyria Rev3.

petejohanson:
## Board/Shield Check-list
 - [x] This board/shield is tested working on real hardware
 - [x] Definitions follow the general style of other shields/boards upstream ([Reference](https://zmk.dev/docs/development/new-shield))
 - [x] `.zmk.yml` metadata file added
 - [x] Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)
 - [x] General consistent formatting of DeviceTree files
 - [x] Keymaps do not use deprecated key defines (Check using the [upgrader tool](https://zmk.dev/docs/codes/keymap-upgrader))
 - [x] `&pro_micro` used in favor of `&pro_micro_d/a` if applicable
 - [x] If split, no name added for the right/peripheral half
 - [x] Kconfig.defconfig file correctly wraps *all* configuration in conditional on the shield symbol
 - [x] `.conf` file has optional extra features commented out
 - [x] Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)

PR #1651, opened 2023-02-05
refactor: Add more checks to pre-commit

joelspadin:
Updated existing pre-commit hooks and added various new hooks:

- Replace tabs with spaces
- Remove trailing whitespace
- Ensure every non-empty file ends with a new line
- Check YAML file validity
- Prevent adding large files
- Ensure any scripts intended to be executable are executable

Added a GitHub action to run pre-commit on every commit. Removed any existing actions such as Prettier checks which duplicate pre-commit.

Ran pre-commit on the codebase.


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
+      - uses: pre-commit/action@v3.0.0
```
joelspadin, line 15: We could also set up https://pre-commit.ci/lite.html, and then it would provide fixes that people could automatically apply to their PRs.

petejohanson:
So, I generally like this, some thoughts as I ponder:

* By moving this all into the pre-commit workflow, we lose some parallelism (not a huge deal, none of the checks take a long time), but more importantly, easier visibility into the actual failure hit from emails, reviewing PRs, etc.
* I'd love to see some example follow up/toss-away PRs demonstrating the checks actually catching the various failure conditions before merging to be sure they work, and to observe how the failures are reported, etc.

Thanks for working on this!

joelspadin:
Throw-away PR up at #1774. https://pre-commit.ci/lite.html would be able to automatically commit fixes for all of those failures except for the invalid YAML one. I'll also add another commit to this updating the docs to show how to set up pre-commit, since I don't see that anywhere yet.

petejohanson (approved):
Let's pull the bandaid!

Can you rebase, I forgot this was pending, and merged #1627 just now. Soon as this is conflict free, let's get it in!

joelspadin:
Rebased.

PR #1658, opened 2023-02-10
refactor(encoder): use behavior queue to archive key taps for encoder

xqinx:
Remove k_msleep() call used in sensor trigger callback that's meant to simulate a key tap (press and release), instead we can use the existing behavior queue mechanism, where a press and a release events are queued up separately. The trick is to add .binging_pressed and .binding_released callbacks for the sensor driver api, which will get called when the queued events are processed accordingly.

## Board/Shield Check-list

### DIFF
--- a/app/src/behaviors/behavior_sensor_rotate_key_press.c
+++ b/app/src/behaviors/behavior_sensor_rotate_key_press.c
@@ -11,6 +11,7 @@
 #include <logging/log.h>
 
 #include <drivers/sensor.h>
+#include <zmk/behavior_queue.h>
 #include <zmk/event_manager.h>
 #include <zmk/events/keycode_state_changed.h>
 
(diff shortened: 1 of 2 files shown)


app/src/behaviors/behavior_sensor_rotate_key_press.c
```diff
@@ -36,26 +73,22 @@ static int on_sensor_binding_triggered(struct zmk_behavior_binding *binding,
...
 
-    LOG_DBG("SEND %d", keycode);
-
-    ZMK_EVENT_RAISE(zmk_keycode_state_changed_from_encoded(keycode, true, timestamp));
-
-    // TODO: Better way to do this?
-    k_msleep(5);
-
-    return ZMK_EVENT_RAISE(zmk_keycode_state_changed_from_encoded(keycode, false, timestamp));
+    zmk_behavior_queue_add(keycode_position, *binding, true, 20);
```
joelspadin, line 85: This will delay both the press and release by 20 ms instead of immediately pressing and then releasing after 20 ms. Was that intended?

app/src/behaviors/behavior_sensor_rotate_key_press.c
```diff
@@ -36,26 +73,22 @@ static int on_sensor_binding_triggered(struct zmk_behavior_binding *binding,
...
+    return zmk_behavior_queue_add(keycode_position, *binding, false, 20);
```
joelspadin, line 86: Can we make this hold duration configurable?

app/src/behaviors/behavior_sensor_rotate_key_press.c
```diff
@@ -36,26 +73,22 @@ static int on_sensor_binding_triggered(struct zmk_behavior_binding *binding,
 
     switch (value.val1) {
     case 1:
-        keycode = binding->param1;
+        keycode_position = 1;
```
joelspadin, line 76: This would cause every encoder to overlap with the second and third keys in the keymap.

I had a change of my own which was going to be part of a larger PR, but since it would fix this problem for you, I've split it out and created #1659. This gives each sensor a virtual key position which it can use for queueing behaviors.

app/src/behaviors/behavior_sensor_rotate_key_press.c
```diff
@@ -20,11 +21,47 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+    }
+
+    return ZMK_EVENT_RAISE(zmk_keycode_state_changed_from_encoded(keycode, true, event.timestamp));
+}
+
+static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
+                                      struct zmk_behavior_binding_event event) {
+    uint32_t keycode;
+
+    switch (event.position) {
```
joelspadin, line 46: As noted below, using the position like this would make it appear to be the same key position as the second and third keys in the keymap.

Maybe a better way to handle this would be to rebase onto #1275, and then instead of giving the sensor behavior itself press/release functions and needing some way to identify whether a press/release corresponds to clockwise or counterclockwise, you would just queue up a press/release for the wrapped behavior?

joelspadin (changes requested):

app/src/behaviors/behavior_sensor_rotate_key_press.c
xqinx: #1659 looks great! Thank you!

app/src/behaviors/behavior_sensor_rotate_key_press.c
xqinx: Ha, I didn't know about #1275, should have looked more carefully on outstanding PRs. #1275 is a better design and it also uses the behavior queue to achieve the same idea, I feel like this PR is redundant and we should probably just close this PR and wait on #1275. WDYT?

xqinx:
@joelspadin Thank you so much for reviewing and your feedback!

xqinx:
Closing this PR since #1758 (which contains #1275 changes) was merged

PR #1659, opened 2023-02-11
Add virtual key positions for sensors

joelspadin:
This extends the virtual key positions used in combos so that each sensor also gets a virtual key position. This allows sensor behaviors to use the behavior queue API, and it may be useful for future work to improve implicit modifier handling.

Also fixed a couple of compiler warnings in the related sensor code.


app/include/zmk/events/keycode_state_changed.h
```diff
@@ -16,13 +16,15 @@ struct zmk_keycode_state_changed {
     uint8_t implicit_modifiers;
     uint8_t explicit_modifiers;
     bool state;
+    uint32_t position;
```
petejohanson, line 19: I'm a bit confused by the inclusion of this here. I'm not *100*% convinced that every keycode state changed will truly come from a particular position yet in our system.

Can you help me a bit with the *why* of this piece of this refactor?

joelspadin: zmk_behavior_queue_add() requires a position, therefore any sensor code that wants to queue behaviors (for example, #1275) requires a position.

joelspadin: (Alternatively, we could change the behavior queue to make positions optional somehow? I'm not sure how possible that is.)

petejohanson: @joelspadin Maybe I'm missing something really obvious here... but what does that have to do w/ the keycode state changed event? Usually this event is *raised* by behaviors, that already have the position info fed into them by the passed in binding event.

joelspadin: Oh. I see what you mean now. Let me see if I can remove that, or if I did this because something else did require it.

joelspadin: Done.

petejohanson (approved):
Code looks good, tested fine on my ZMK Uno.

PR #1662, opened 2023-02-13
fix keypresses that are not in the transform

purdeaandrei:
Before this change, if a matrix position was not present in the transform, various incorrect behaviors would happen:

1) In some cases out-of-bounds accesses:

Note that the size of the`transform[]` array does not necessarily match the size of the matrix. So for example if key position (ZMK_MATRIX_COLS-1, ZMK_MATRIX_ROWS-1) is not present in the transform, but ends up being pressed, then the array will be accessed beyond its size, and any data could be returned.

2) In other cases the 0th position in the keymap will be used because the `transform[]` array is initialized to all zeros.

### DIFF
--- a/app/src/kscan.c
+++ b/app/src/kscan.c
@@ -47,7 +47,14 @@ void zmk_kscan_process_msgq(struct k_work *item) {
 
     while (k_msgq_get(&zmk_kscan_msgq, &ev, K_NO_WAIT) == 0) {
         bool pressed = (ev.state == ZMK_KSCAN_EVENT_STATE_PRESSED);
-        uint32_t position = zmk_matrix_transform_row_column_to_position(ev.row, ev.column);
+        int32_t position = zmk_matrix_transform_row_column_to_position(ev.row, ev.column);
+
+        if (position < 0) {
+            LOG_WRN("Not found in transform: row: %d, col: %d, pressed: %s", ev.row, ev.column,
+                    (pressed ? "true" : "false"));
+            continue;
+        }
+
         LOG_DBG("Row: %d, col: %d, position: %d, pressed: %s", ev.row, ev.column, position,
                 (pressed ? "true" : "false"));
         ZMK_EVENT_RAISE(new_zmk_position_state_changed(
--- a/app/src/matrix_transform.c
+++ b/app/src/matrix_transform.c
@@ -11,17 +11,32 @@
 
 #ifdef ZMK_KEYMAP_TRANSFORM_NODE
 
-#define _TRANSFORM_ENTRY(i, _)                                                                     \
-    [(KT_ROW(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i)) * ZMK_MATRIX_COLS) +               \
-        KT_COL(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i))] = i
+/* the transform in the device tree is a list of (row,column) pairs that is
+ * indexed by by the keymap position of that key. We want to invert this in
+ * order to be able to quickly determine what keymap position a particular
+ * row,column pair is associated with, using a single lookup.
+ *
+ * We do this by creating the `transform` array at compile time, which is
+ * indexed by (row * ZMK_MATRIX_COLS) + column, and the value contains an
+ * encoded keymap index it is associated with. The keymap index is encoded
+ * by adding INDEX_OFFSET to it, because not all row,column pairs have an
+ * associated keymap index (some matrices are sparse), C globals are
+ * initialized to 0, and the keymap index of 0 is a valid index. We want to
+ * be able to detect the condition when an unassigned matrix position is
+ * pressed and we want to return an error.
+ */
 
-static uint32_t transform[] = {LISTIFY(ZMK_KEYMAP_LEN, _TRANSFORM_ENTRY, (, ), 0)};
+#define INDEX_OFFSET 1
+
+#define TRANSFORM_ENTRY(i, _)                                                                      \
+    [(KT_ROW(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i)) * ZMK_MATRIX_COLS) +               \
+        KT_COL(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i))] = i + INDEX_OFFSET
+
+static uint32_t transform[] = {LISTIFY(ZMK_KEYMAP_LEN, TRANSFORM_ENTRY, (, ), 0)};
 
 #endif
 
-uint32_t zmk_matrix_transform_row_column_to_position(uint32_t row, uint32_t column) {
-    uint32_t matrix_index;
-
+int32_t zmk_matrix_transform_row_column_to_position(uint32_t row, uint32_t column) {
 #if DT_NODE_HAS_PROP(ZMK_KEYMAP_TRANSFORM_NODE, col_offset)
     column += DT_PROP(ZMK_KEYMAP_TRANSFORM_NODE, col_offset);
 #endif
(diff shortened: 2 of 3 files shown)


app/src/matrix_transform.c
```diff
@@ -33,7 +34,12 @@ uint32_t zmk_matrix_transform_row_column_to_position(uint32_t row, uint32_t colu
     matrix_index = (row * ZMK_MATRIX_COLS) + column;
 
 #ifdef ZMK_KEYMAP_TRANSFORM_NODE
-    return transform[matrix_index];
+    if (matrix_index >= ARRAY_SIZE(transform))
+        return -1;
+    value = transform[matrix_index];
+    if (!value)
+        return -1;
+    return value - 1;
```
joelspadin, line 42: This could be simplified to
```c
return transform[matrix_index] - 1;
```

app/src/kscan.c
```diff
@@ -47,14 +47,19 @@ void zmk_kscan_process_msgq(struct k_work *item) {
...
-                (pressed ? "true" : "false"));
-        ZMK_EVENT_RAISE(new_zmk_position_state_changed(
-            (struct zmk_position_state_changed){.source = ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL,
-                                                .state = pressed,
-                                                .position = position,
-                                                .timestamp = k_uptime_get()}));
+        int32_t position = zmk_matrix_transform_row_column_to_position(ev.row, ev.column);
+        if (position < 0) {
+            LOG_WRN("Not found in transform: row: %d, col: %d, pressed: %s", ev.row, ev.column,
+                    (pressed ? "true" : "false"));
```
joelspadin, line 53: Is the pressed/released info relevant here?

app/src/matrix_transform.c
```diff
@@ -13,14 +13,15 @@
 
 #define _TRANSFORM_ENTRY(i, _)                                                                     \
     [(KT_ROW(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i)) * ZMK_MATRIX_COLS) +               \
-        KT_COL(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i))] = i,
+        KT_COL(DT_PROP_BY_IDX(ZMK_KEYMAP_TRANSFORM_NODE, map, i))] = i + 1,
```
joelspadin, line 16: Consider making a `#define` for the `1` here. That would ensure that the same offset is added here as is subtracted later, and it could help make the code easier to understand if given a descriptive name.

app/src/matrix_transform.c
```diff
@@ -13,14 +13,15 @@
 
 #define _TRANSFORM_ENTRY(i, _)                                                                     \
```
joelspadin, line 14: Not an issue with this PR, but since you're editing this code already, this would be nice to fix. All identifiers that begin with an underscore followed by a capital letter are reserved for use by the C compiler.

```suggestion
#define TRANSFORM_ENTRY(i, _)                                                                      \
```

app/src/kscan.c
purdeaandrei: This matches the LOG_DBG for when something is found in the transform (except for the position, which cannot be printed if not found). I think if we assume that this can be triggered only by an incorrect transform, then the user won't care if it's a press/release event. If this is however triggered by a kscan driver that spits out wrong data,  or if it's triggered by a damaged matrix, then the user may find the pressed/released information useful. Imagine someone debugging a short in their keyboard matrix, and they are getting a release event when they put pressure on the area of the keyboard, knowing it was a release could be useful. I'm advocating for keeping it, but if you want I can remove it.

app/src/matrix_transform.c
purdeaandrei: DONE

app/src/matrix_transform.c
purdeaandrei: DONE

app/src/matrix_transform.c
purdeaandrei: I defined macro for the offset, and I changed the return value to -EINVAL when the position is not found, and this now seems more correct, so as it is it can't be simplified anymore.

app/src/matrix_transform.c
```diff
@@ -33,7 +36,12 @@ uint32_t zmk_matrix_transform_row_column_to_position(uint32_t row, uint32_t colu
     matrix_index = (row * ZMK_MATRIX_COLS) + column;
 
 #ifdef ZMK_KEYMAP_TRANSFORM_NODE
-    return transform[matrix_index];
+    if (matrix_index >= ARRAY_SIZE(transform))
+        return -EINVAL;
+    value = transform[matrix_index];
```
joelspadin, line 41: Minor: [Zephyr requires](https://docs.zephyrproject.org/3.1.0/develop/languages/c/index.html#language-standards) a compiler that supports C99 (and some C11 features), so variables don't need to be declared at the top of a scope. Declaring and initializing a variable in the same statement prevents accidentally using it before it's initialized.

```suggestion
    const uint32_t value = transform[matrix_index];
```

(could also do the same for `matrix_index` if you wanted)

app/src/matrix_transform.c
```diff
@@ -33,7 +36,12 @@ uint32_t zmk_matrix_transform_row_column_to_position(uint32_t row, uint32_t colu
     matrix_index = (row * ZMK_MATRIX_COLS) + column;
 
 #ifdef ZMK_KEYMAP_TRANSFORM_NODE
-    return transform[matrix_index];
+    if (matrix_index >= ARRAY_SIZE(transform))
+        return -EINVAL;
```
joelspadin, line 40: I'm not a fan of `if` blocks without braces, but clang-format should automatically change something like

```c
if (error)
    LOG_ERR("Just did a bad thing");
    return -EINVAL;
```
to
```c
if (error)
    LOG_ERR("Just did a bad thing");
return -EINVAL;
```

and make that error more obvious. I see we already do this in several places throughout the codebase too, so if Pete's okay with this, I won't object.

joelspadin (approved):
Looks reasonable to me. @petejohanson should probably look over this too.

app/src/matrix_transform.c
purdeaandrei: DONE

app/src/matrix_transform.c
purdeaandrei: DONE, I added braces.

joelspadin (approved):

app/src/kscan.c
```diff
@@ -47,14 +47,19 @@ void zmk_kscan_process_msgq(struct k_work *item) {
...
-        uint32_t position = zmk_matrix_transform_row_column_to_position(ev.row, ev.column);
-        LOG_DBG("Row: %d, col: %d, position: %d, pressed: %s", ev.row, ev.column, position,
-                (pressed ? "true" : "false"));
-        ZMK_EVENT_RAISE(new_zmk_position_state_changed(
-            (struct zmk_position_state_changed){.source = ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL,
-                                                .state = pressed,
-                                                .position = position,
-                                                .timestamp = k_uptime_get()}));
+        int32_t position = zmk_matrix_transform_row_column_to_position(ev.row, ev.column);
+        if (position < 0) {
```
petejohanson, line 51: I would prefer we structure this as a guard clause to avoid nesting too much:

```
if (position < 0) {
  LOG_WRN("Blah blah");
  continue;
}

Do the real work here
```

app/src/kscan.c
petejohanson: I agree on keeping it.

petejohanson (changes requested):
Just one minor comment. Otherwise looks good. Thanks!

app/src/kscan.c
purdeaandrei: DONE

app/src/matrix_transform.c
```diff
@@ -11,17 +11,17 @@
 
 #ifdef ZMK_KEYMAP_TRANSFORM_NODE
 
-#define _TRANSFORM_ENTRY(i, _)                                                                     \
+#define INDEX_OFFSET 1
```
petejohanson, line 14: This code is great, and clear *based on knowing the context of this PR*, but I do fear that next time this gets looked at in a few months it won't be as clear. I'm not a fan of "code comments for the sake of comments", but can you add a small bit of context/explanation above this define to clarify the offset is used so that only real values are non-zero, so the default 0 value is left as a "undefined transform mapping" to be detected.

Once that's in, this is good to go. Thanks!

purdeaandrei: DONE, added comment.

purdeaandrei:
Rebased to resolve conflicts

petejohanson (approved):
Thanks!

PR #1663, opened 2023-02-13
implement gpio_165 driver

purdeaandrei:
This is an input-only gpio driver for use with 74x165 style parallel-input shift registers, (using SPI peripheral)

The user can choose to either use cs-gpios of spi or load-gpios for connecting to the SH/LD pin of the shift register. The difference is that cs-gpios will keep the input latch enabled while idle, and load-gpios will instead pulse the SH/LD signal before the register content is shifted. In both cases the gpio should be configured GPIO_ACTIVE_HIGH.


app/drivers/zephyr/dts/bindings/gpio/zmk,gpio-165.yaml
```diff
@@ -0,0 +1,35 @@
...
+include: [gpio-controller.yaml, spi-device.yaml]
+
+properties:
+  "#gpio-cells":
+    const: 2
+
+  ngpios:
+    type: int
+    required: true
+    enum:
```
xingrz, line 21: This should not be an enum, I think. 165s can be daisy-chained more than 4.

image

purdeaandrei: You are correct that more than 4 can be daisy-chained, however, zephyr currently only supports up to 32-bit ports, see here:
https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/gpio.h#L525
and here:
https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/gpio.h#L229

I'm not sure how to implement support for more than 4, but I'm think it would have to be in a way that one gpi-165 instance references another, so you can actually refer to the index>32 gpios from your device tree. Or perhaps there could be a common node that doesn't expose a gpio driver, and sub-nodes that each expose up to 32 bits per gpio port. In any case this would be more complicated than just updating the ngpios data type.

Also this PR follows almost 1:1 what the existing gpio_595 driver does, which is also limited to 32 gpios, probably for the same reason.

I think supporting more than 32 gpios is out of scope for this PR, and eventually when support for that is added, it should be added symmetrically for both 165 and 595.

alinelena:
in theory nobody stops you to read in an array of uint8_t. also the symmetry argument for drivers for sipo vs piso needs to be carefully consider. i assume a col2row scanning.
sipo, hc595, in general is used in a matrix like setup, hence 32 columns is more or less sufficient for practical purposes.
piso, hc165 and probably 589, has two uses, 
1. classical matrix, in this case again 32 rows is sufficient for practical purposes. 
2. direct read. in this case 1 shift register pin per key, so easily 32 becomes a limit.

purdeaandrei:
Rebased and made compatible with Zephyr 3.2 changes

purdeaandrei:
@alinelena 
> in theory nobody stops you to read in an array of uint8_t. 

Reading any number of bits from SPI is tivial. That is not what's slowing me down from adding support for more than 32. As I said before what is slowing me down is the zephyr API itself, see `struct gpio_driver_api`, the second argument is of type `gpio_port_value_t`, this is `typedef uint32_t gpio_port_value_t;`.

So to make more than 32 work, I'd have to devise a way to instantiate more than one gpio port.

I'm open to suggestions on that.

For example here's one very simple idea: We could add a value called `offset`, type:int to the dts bindings, and then you could simply define multiple ports with the exact same settings under spi-device, except for one would have an offset of 0, and another would have an offset of 32, and other would have an offset of 64, ... **On a second thought this probably wouldn't work**, since, there would be different semaphores protecting access to the hardware, so the two gpio controllers could possibly collide.

So maybe there would need to be two node types, a parent and a child. The parent would be an spi-device node that isn't a gpio-controller node, that sits under the spi bus, that would handle the semaphore, and possibly the actual shifting. The children of this nodes would be gpio-controller nodes calling into the parent.

So please give suggestions, it is not at all obvious what is the best path to follow for >32-bit support.

> also the symmetry argument for drivers for sipo vs piso needs to be carefully consider. i assume a col2row scanning.
sipo, hc595, in general is used in a matrix like setup, hence 32 columns is more or less sufficient for practical purposes.
piso, hc165 and probably 589, has two uses,
classical matrix, in this case again 32 rows is sufficient for practical purposes.
direct read. in this case 1 shift register pin per key, so easily 32 becomes a limit.

I hear you and agree that having >32 bit support is more important for inputs. However having >32bit support for 595 is still useful, and it would still be cleaner design if symmetry was maintained here. So whatever the design, it should be done so that the same can be applied to 595.

purdeaandrei:
Oh, and one more thing that came to my mind, not really an argument against supporting > 32 bits, but it does weaken the necessity quite a bit:

If you're trying to do direct read of 1 full-size (101-key keyboard), where you configured the 101 keys as 4 separate 32-bit io ports, then with the current kscan_gpio_direct driver it would take **6784 SPI clock cycles** to scan the matrix once. This is because each pin would need at least 32 clock cycles to scan, or a multiple of that depending on position. Here's the code to calculate it:
```python
portsize=32
sum([(i+1+portsize-1)//portsize * portsize for i in range(101)])
```
Okay I hear you saying, but what if PR #1571 is merged, and port reads are grouped nicely so each port is read only once? Then you would need `32 * (1 + 2 + 3 + 4)`  = **320 SPI clock cycles** to scan the matrix once, instead of the ideal 101 clock cycles. I guess it would function, but you really wouldn't want to use it that way, and I suspect you would rather just implement a custom kscan driver that does it efficiently for you.

In fact it would make sense to implement a `kscan165` driver that can support arbitrary number of keys and have it be available for all of zmk, but that is not what this PR is.

Supporting > 32 bits as gpios only really makes sense if the GPIOs are only read sporadically, so not for matrix scanning.

petejohanson:
Can you please just rebase on last time, now that our pre-commit checks are in place, to ensure no issues pop once we merge this? PR is great otherwise. Thanks!

purdeaandrei:
Rebased as requested (no conflicts), not tested since rebase.

alinelena:
tried to use it on one my setups... there is some oddity with it in the sense i share the cs 
https://github.com/alinelena/mlego-zmk/tree/main/config/boards/shields/mlego5x13_rev9

I know the circuit is working since I have a working version of qmk and xiao rp2040.. unfortunately I do not seem to be able to get it working... the mcu does not lock since the encoder works, but not the matrix scanning.

alinelena:
> tried to use it on one my setups... there is some oddity with it in the sense i share the cs https://github.com/alinelena/mlego-zmk/tree/main/config/boards/shields/mlego5x13_rev9
> 
> I know the circuit is working since I have a working version of qmk and xiao rp2040.. unfortunately I do not seem to be able to get it working... the mcu does not lock since the encoder works, but not the matrix scanning.

just to say the pattern for working scanning is like this 

```
    33   writePinLow(SPI_LATCH_PIN);                                                                                                       
    34   spi_transmit(message,2);                                                                                                          
    35   writePinHigh(SPI_LATCH_PIN);                                                                                                      
    36   matrix_output_select_delay(); 
 to select a column then to read the rows

    17     writePinLow(SPI_LATCH_PIN);                                                                                                     
    18     writePinHigh(SPI_LATCH_PIN);                                                                                                    
    19                                                                                                                                     
    20     spi_status_t read_result = spi_read();  
```

PR #1664, opened 2023-02-13
Implement deghost kscan driver

purdeaandrei:
This implements a kscan driver that takes another kscan driver, and dehosts its matrix, for keyboards
such as membrane keyboards that show ghosting effects when in some combinations more than 2 keys
are pressed.

Documentation update included.


purdeaandrei:
This was originally written on top of petejohanson's zephyr upgrade branch, so will probably have to make some changes to make it main-compatible. In particular:
- include headers may need to be changed
- the listify macro has changed

purdeaandrei:
Okay, I pushed a commit to make it compile on main. This commit should be reverted in the future when zephyr is updated. Note that I don't have a way to test this in main other than seeing if it compiles or not. All my testing was done on the zephyr 3.2 branch.

app/drivers/kscan/kscan_deghost.c
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
joelspadin, line 2:
```suggestion
 * Copyright (c) 2023 The ZMK Contributors
```
(or 2022, depending on when you started writing this)

purdeaandrei:
Rebased, and made compatible with zephyr 3.2-related changes (by removing the downgrade commit from the stack)

purdeaandrei:
Rebased as requested (no conflicts), not tested since rebase.

PR #1657, opened 2023-02-10
feat(docs): Implement Animated Hold-Tap Diagrams

kurtis-lew:
Implements a [Motion Canvas](https://motion-canvas.github.io/) project that compiles concurrently with ZMK docs to generate animated timing diagrams for hold-taps. Animations are therefore publicly editable and maintainable, so long as future contributors have knowledge of TypeScript and the Motion Canvas API.

Future expansion of the animationhelpers will target other behaviors, or even other fundamental features of ZMK, as requested in #1284.

PR #1669, opened 2023-02-18
Typo: tap => hold

arialdomartini:
This PR contains a fix to a little typo in the documentation.

In the Hold-Tap section, one of the examples states:

```
tapping-term-ms = <100>;                        // <---[[produces tap if held longer than tapping-term-ms]]
```

I believe this is incorrect: if a key is held longer than `tapping-term-ms`, it has the "hold" behavior, not the tap.

### DIFF
--- a/docs/docs/behaviors/hold-tap.md
+++ b/docs/docs/behaviors/hold-tap.md
@@ -159,7 +159,7 @@ The following are suggested hold-tap configurations that work well with home row
 			label = "LEFT_POSITIONAL_HOLD_TAP";
 			#binding-cells = <2>;
 			flavor = "tap-unless-interrupted";
-			tapping-term-ms = <100>;                        // <---[[produces tap if held longer than tapping-term-ms]]
+			tapping-term-ms = <100>;                        // <---[[produces hold if held longer than tapping-term-ms]]
 			quick-tap-ms = <200>;
 			bindings = <&kp>, <&kp>;
 			hold-trigger-key-positions = <5 6 7 8 9 10>;    // <---[[right-hand keys]]


caksoylar:
I am not sure this is correct, since `tap-unless-interrupted` is an interesting beast: The docs description also mentions this:
> The 'tap-unless-interrupted' flavor triggers a hold behavior only when another key is pressed before tapping-term-ms has expired. It triggers the tap behavior in all other situations.

So I believe in this case tapping term is an upper bound on the timing required to trigger a hold, rather than a lower bound. [This unit test](https://github.com/zmkfirmware/zmk/tree/main/app/tests/hold-tap/tap-unless-interrupted/2-dn-timer-up) also seems to confirm it: It taps the `&ht_tui LEFT_SHIFT F` key for 500 ms (exceeding tapping term) then releases, and `usage_page 0x07 keycode 0x09` (which [corresponds to the `F` keycode](https://github.com/zmkfirmware/zmk/blob/main/app/include/dt-bindings/zmk/hid_usage.h#L299)) is expected to be emitted.

arialdomartini:
I think you are right, and we should close this PR without merging it.

It's clear that my understanding of `tap-unless-interrupted` was not very accurate. 

As a side note, while other modes have helpful time diagrams that aid in comprehension, `tap-unless-interrupted` is the only one that lacks such a diagram. I wonder if adding one would be useful. What do you think?

caksoylar:
> As a side note, while other modes have helpful time diagrams that aid in comprehension, tap-unless-interrupted is the only one that lacks such a diagram. I wonder if adding one would be useful. What do you think?

I'd agree so, I believe @kurtis-lew is already working on updating the diagrams at #1657 so that might be worth checking to not duplicate work.

PR #1670, opened 2023-02-19
fix(shields): Fix extra matrix transform offsets and Kyria five column transforms

caksoylar:
A few split keyboards didn't have `col-offset` defined on the right overlays for the non-default transform which were `jian`, `jiran`, `corne`, `jorne`, `kyria`, `kyria_rev2` shields, while they were properly defined for `corneish_zen` and `splitkb_aurora_corne`. This leads to the right side not behaving correctly when users choose these transforms in their keymap.

I also believe the Kyria five column transform definitions were incorrect; I believe the new version should be, however since I don't have the hardware it would be nice if someone can confirm this from schematics or by testing the hardware.


zakrywilson:
This issue was impacting me with a Chocofi that used the Corne shield. I tested it out and can confirm that it works.

[Here](https://github.com/zakrywilson/zmk-config/blob/chocofi-fix-transform-offsets/config/corne.keymap) is the branch I tested it out with (in case anyone's interested).

petejohanson (approved):
Good catches @caksoylar!

The only downside I see to the matrix transforms like this is it doesn't "free up" the unused pin for the column not needed here, but that's a larger issue to solve another time.

PR #1683, opened 2023-02-24
Various logging improvements

xudongzheng:
- Adds option for RTT logging
- Adds option to disable debug log messages (can be overwhelming when trying to isolate a particular issue)
- Increase log processing frequency


petejohanson (approved):
Thanks, some good minor tweaks!

PR #1687, opened 2023-02-27
fix(hold-tap binding) Mark tapping-term-ms as required

nickcoutsos:
I think this is the last instance of a behavior property being required but not marked as such.

Looking through the other yaml files the only other instances of properties not marked as required and not specifying a default in the definition are:

- caps word's `mods` so the modifiers can be referenced by identifier, and
- macros (where the defaults can come from kconfig)

### DIFF
--- a/app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-hold-tap.yaml
@@ -13,6 +13,7 @@ properties:
     required: true
   tapping-term-ms:
     type: int
+    required: true
   tapping_term_ms: # deprecated
     type: int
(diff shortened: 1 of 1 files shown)


nickcoutsos:
The failing build is for `hummingbird` which [uses the deprecated `tapping_term_ms`](https://github.com/zmkfirmware/zmk/blob/89044079839b156eeb4a3da497d1dab904290f31/app/boards/shields/hummingbird/hummingbird.keymap#L24), so now I'm wondering if omitting the `required` flag is deliberate?

petejohanson:
> 

That's exactly right, why we didn't originally add it as required, was to avoid breaking changes. Let me sleep on this one.

PR #1690, opened 2023-02-28
bugfix(corneish-zen-v2): Disable USB-CDC endpoint by default

chrissicool:
The  corneish-zen v2 board's default configuration enables a USB CDC endpoint by default. This is most probably a debugging left-over. This pull request seeks to disable that endpoint by default. Specifying the `CONFIG_ZMK_USB_LOGGING` config option enables that endpoint automatically.


caksoylar (approved):
I tested these changes on a v2 left and v1 right (from #1593). Keyboard continues to work as expected with or without logging, on both left and right. So it looks good to me.

petejohanson (approved):
Thanks!

PR #1694, opened 2023-03-04
Add charliplex keyscan

HookyKB:
Adds support for https://kbd.news/Square-or-round-robin-matrix-1400.html with addition of an extra interrupt pin for low power support.

Implementation of this matrix in hardware is left as an exercise for the user. Ensure all lines to the interrupt pin are not able to re-feed the matrix or it will surely fail.

Example:
schema

* 6 pin charliplex matrix with interrupt support - for a 6x4 + 3 board*

Shields implementing this scan setup may not be suitable for all boards (see the referenced document above).

It is expected that, with proper design, (compared to a matrix kscan) an additional 2*n diodes will be required, where n is the number of input pins in use. One for the line to the input pin (this may not be required for some controllers), and one for each 'column' to the interrupt pin. Note that not all of the charliplex matrix pins need to be used as input to the controller.

eg. An 8 pin charliplex kscan matrix may use all 8 pins for output, but only use 5 for input.

See https://github.com/HookyKB/zmk/commit/40221a49c8d8300a4f393b6cac7a04000a2e0af8 for example usage.


app/drivers/kscan/Kconfig
```diff
@@ -54,6 +60,30 @@ config ZMK_KSCAN_MATRIX_WAIT_BETWEEN_OUTPUTS
 
 endif # ZMK_KSCAN_GPIO_MATRIX
 
+if ZMK_KSCAN_GPIO_ROUND_ROBIN
+
+config ZMK_KSCAN_MATRIX_WAIT_BEFORE_INPUTS
```
petejohanson, line 65: I'm not sure there's much value in duplicating these values, or using shared config entries for these two properties. Thoughts?

app/drivers/kscan/kscan_gpio_round_robin.c
```diff
@@ -0,0 +1,413 @@
+/*
+ * Copyright (c) 2020-2022 The ZMK Contributors
```
petejohanson, line 2:
```suggestion
 * Copyright (c) 2020-2023 The ZMK Contributors
```

app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-round-robin.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+include: kscan.yaml
+
+properties:
+  gpios:
+    type: phandle-array
+    required: true
+  interrupt-gpios:
+    type: phandle-array
+    required: true
```
petejohanson, line 16: This shouldn't be required, since we should be able to fall back to polling and not use them. (Had a report on Twitter of someone who'd designed a keyboard with RR matrix without the additional hardware, noted they ended up having to put an empty array in place).

Thoughts on removing the extra Kconfig setting in favor of switching to polling mode if the `interrupt-gpios` property isn't set? We don't have the luxury of doing that for the matrix kscan config since that requires knowing if the GPIO inputs themselves support interrupts. In this case, we know unambiguously if we should force polling.

petejohanson (changes requested):
Thanks for this! A few thoughts on the details.

app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-round-robin.yaml
oha-Ohashi: I am that "someone". I don't think it is necessary, either. I was forced to set a fake interrupt pin because of that feature.
https://twitter.com/oha_oha_Ohashi/status/1636855734343450624?s=20

app/drivers/kscan/Kconfig
HookyKB: I thought that initially as well, but since multiple `kscan`s can be used, I thought it best to keep them seperate. I note that this one is a copy paste error though, it should be `CONFIG_ZMK_KSCAN_ROUND_ROBIN_WAIT_BEFORE_INPUTS`.

I'm easy either way. Can fix the typo, or rely on the MATRIX version. Your call.

app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-round-robin.yaml
petejohanson: Yes indeed!

HookyKB: Sounds good. Will update the docs to match. @oha-Ohashi sorry for the mess around, glad you got it sorted.

oha-Ohashi:
Also, I'd love to have a feature that allows me to reverse the columns and rows. Currently, I'm working with a circuit that looks like this:

image

(source: https://jpskenn.github.io/blog/2020/12/05/first-year-of-my-self-made-keyboard-activity)

I find this circuit quite beautiful and it would be amazing if it could be supported by default.

I'm not certain how it can be achieved, though.

app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-round-robin.yaml
HookyKB: I won't be able to omit the code if relying on the `interrupt-gpios` setting as it requires the node to access that value. It's small difference anyway though, so shouldn't be a problem.

app/drivers/kscan/kscan_gpio_round_robin.c
```diff
@@ -0,0 +1,413 @@
...
+
+#if CONFIG_ZMK_KSCAN_DEBOUNCE_RELEASE_MS >= 0
+#define INST_DEBOUNCE_RELEASE_MS(n) CONFIG_ZMK_KSCAN_DEBOUNCE_RELEASE_MS
+#else
+#define INST_DEBOUNCE_RELEASE_MS(n)                                                                \
+    DT_INST_PROP_OR(n, debounce_period, DT_INST_PROP(n, debounce_release_ms))
+#endif
+
+#define USE_POLLING IS_ENABLED(CONFIG_ZMK_KSCAN_ROUND_ROBIN_POLLING)
+#define USE_INTERRUPT (!USE_POLLING)
```
petejohanson: You should be able to define a couple defines like

```
#define ANY_POLLING DT_INST_FOREACH(ADDS_ONE_IF_NO_INTERRUPT_GPIO) > 0
#define ANY_INTERRUPT DT_INST_FOREACH(ADDS_ONE_IF_HAS_INTERRUPT_GPIO) > 0
```

To use for the conditional *code* needed for each approach, and then similar "per instance" checks to decide what to do for a given instance.

app/drivers/kscan/Kconfig
HookyKB: @petejohanson waiting for your opinion on this one. Have done the typo fix for now.

app/drivers/kscan/kscan_gpio_round_robin.c
HookyKB: Didn't see this early enough, have made a change including both tracks, which would give the 'per instance' code anyway.

I'll take another look later.

HookyKB:
@oha-Ohashi 

The column/row don't actually matter with the round robin. If you consider `RC` as `send, receive` or `out, in` you can build you settings the same way. Having the diodes before or after the switch doesn't matter either.

In your transform, you would just have:

```
RC(2,1) RC(3,1)
RC(1,2) RC(3,2)
RC(0,3) RC(2,3)
```
It's not really rows and columns, just out/in pairings.

Also note that flipping all the diode directions (but no position) gives you the same capability, but with the out/in reversed, and you keep the same picture.

oha-Ohashi:
@HookyKB Thank you for your response. it seems that my understanding was not enough.

HookyKB:
@petejohanson auto polling config added. See https://github.com/zmkfirmware/zmk/pull/1694#discussion_r1144131612 regarding the wait config.

nguyendown:
I find that "charlieplex" would be a proper name for this type of key scan.

HookyKB:
> I find that "charlieplex" would be a proper name for this type of key scan.

It is, but Round Robin is equally proper, and less ambiguous IMO.

nguyendown:
> It is, but Round Robin is equally proper, and less ambiguous IMO.

While some people may be familiar with the term "round-robin tournament," it still does not accurately reflect the underlying technique used in the keyscan, which relies on diodes being forward biased.

What's more important is that the keyscan is entirely based on charlieplexing, so there's no reason to come up with a new term for it. Let's just use the already established term. It's less confusing and way more straightforward.

HookyKB:
@nguyendown I don't see a problem. No-one is 'confused', you would just prefer a different name. I didn't come up with the name, I used the title of the reference document, and in there you'll see that the author was also unaware of the name you're suggesting.

Also, with the additional interrupt handler, it is now its own entity, so, I'm pretty happy calling it _not_ charliplexing. You'll may also like to note that charlieplexing was a term assigned to a long pre-existing circuit structure, so 🤷. (And thus named for someone who didn't come up with it, and I like that even less.)

petejohanson:
Code looks good. Can you please rebase one more time to pick up the pre-commit check work, to be sure the MD files all pass?

Thanks!

nguyendown:
Can we put this pull request on hold? Because of the naming convention? I am currently messaging the author of the article about the proper name.

petejohanson:
I'm happy to discuss more. Regardless of technicalities, the term "round robin matrix" has become pervasive. At a minimum, if we do want to merge this with a different name, then the docs should clearly identify this as being the same design known as "round robin matrix" by many.

nguyendown:
I'll bring it here so that we can have an open discussion.

<blockquote>
There are three names in the article: Charlieplexing, Round Robin Matrix, and Square Matrix. I immediately knew that "Square Matrix" would not work well. The term is not unique, and people may ask, "square matrix? Is it linear algebra?" You'd have to explain "square means N-square keys" every time you mention it. You'd also need to tell them "it's not exactly N-square, but like N-square minus N." Not to mention that "square matrix" could also refer to a normal matrix with four cols, and four rows.

"Round Robin" is interesting, but even if someone is familiar with the term "round robin tournament," they may only have a vague idea of how it works. That's not enough to make a keyboard because the technique relies on diodes being forward biased, which is not covered by "round robin tournament".

Also, the term "round robin" is not unique.
https://en.wikipedia.org/wiki/Round-robin

Charlieplexing does not describe itself at all. Why is this a more appropriate name? Let me explain. Multiplexing is hard to describe in just a few words, which is why a typical name for multiplexing usually is something + "plexing". Just like learning a language, people will have to look it up to understand a word. Because the term is unique, a web search for "Charlieplexing" leads to a handful of material that describes it well. And because the term gives us zero presumptions, it gives zero confusion. Plus, it is a well established term. If someone already knows about Charlieplexing, they'll understand how the keyscan works.

I get that using a simpler term makes it more accessible to a wider range of people. However, I would argue that it's not necessary because only enthusiasts, who are technical audiences, would learn about the keyscan anyway, regardless how complex the term is. End users typically do not care about it. If they want to understand it, "Charlieplexing" is still a better term for web searching.

One more thing I'd like to point out is that adaptations (or discussions) of Charlieplexing in keyboards do exist. It is rare, but that doesn't mean we should replace the term.
https://github.com/thanks4opensource/markoplexing
https://github.com/crides/zmk/blob/wren/app/drivers/kscan/kscan_gpio_charlieplex.c
You can find discussions about "charlieplex" in the Discord servers of ZMK and QMK as well. By keeping the term, we can maintain consistency in communication.
</blockquote>

### Zephyr

> Also, the term "round robin" is not unique.

Zephyr also refers to "round robin" as "round robin scheduling" `grep -rl "round.robin" zmk/zephyr`

### Keyscan

I'll explain why I insist on using the term "Charlieplex" for the keyscan.

Here's the original Charlieplex (1). `SW0-1` indicates that the key press will be detected when `pin0` outputs and `pin1` inputs.



(2) and (3) are the same as the original, but the switches are arranged in a grid.



(4) and (5) have an additional diode per pin to prevent ghosting.



(6) and (7) use an interrupt pin to detect key press from sleep.



All the variants use the same scanning technique as the original, which is why the keyscan should be named "Charlieplex."

### Variants

Trying to name every possible variation or arrangement is not practical. Not only is it not feasible, but it also creates even more confusion. My suggestion is to simply describe any additional features. We can refer to (1), (2), and (3) as the Charlieplex. (4) and (5) can be called the Anti-Ghosting Charlieplex, while (6) and (7) can be referred to as the Low-Power Anti-Ghosting Charlieplex.

michaelrommel:
Hmmmm, is it really now the case, that because of differences in the name of a feature, it is not being merged and users are not able to make use of the feature?

I honestly couldn't care less if the feature is called "Low-Power Anti-Ghosting Charlieplex in Round-Robin, Square-Matrix Fashion with superpowers" or just "Matrix Connection Variant #7"...

Please: who can make a decision to merge a feature here?

Thank you for considering your users and preventing further forks just to implement already submitted feature branches...

HookyKB:
@petejohanson Yep, no problems. I'll do the rename & rebase this weekend. Two devs in disagreement meant nothing IMO, but once the repo owner rolls in, I'm generally happy with what they decide. 👍

HookyKB:
@petejohanson all yours. Tested with my `intro` board, all working as expected.

docs/docs/config/kscan.md
```diff
@@ -152,6 +152,43 @@ The output pins (e.g. columns for `col2row`) should have the flag `GPIO_ACTIVE_H
...
 ```
 
+## Charliplex Driver
+
+Keyboard scan driver where keys are arranged on a matrix with each GPIO used as both input and output.
+
+- With `interrupt-gpios` unset, this allows n pins to drive n\*(n-1) keys.
+- With `interrupt-gpios` set, n pins will drive (n-1)\*(n-2) keys, but provide much improved power handling.
+
+Definition file: [zmk/app/drivers/kscan/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/drivers/kscan/Kconfig)
```
HookyKB, line 162: FYI, all of the links in this file are wrong, they haven't been updated to point to the `/module` directory.

I leave this one in the same state deliberately incase the fix is done with a find/replace which may then break this one.

docs/docs/config/kscan.md
```diff
@@ -152,6 +152,43 @@ The output pins (e.g. columns for `col2row`) should have the flag `GPIO_ACTIVE_H
...
+
+| Config                                             | Type        | Description                                                               | Default |
+| -------------------------------------------------- | ----------- | ------------------------------------------------------------------------- | ------- |
+| `CONFIG_ZMK_KSCAN_CHARLIPLEX_WAIT_BEFORE_INPUTS`   | int (ticks) | How long to wait before reading input pins after setting output active    | 0       |
+| `CONFIG_ZMK_KSCAN_CHARLIPLEX_WAIT_BETWEEN_OUTPUTS` | int (ticks) | How long to wait between each output to allow previous output to "settle" | 0       |
+
+- With `CONFIG_ZMK_KSCAN_CHARLIPLEX_POLLING` enabled this allows n pins to drive n\*(n-1) keys.
+- With `CONFIG_ZMK_KSCAN_CHARLIPLEX_POLLING` disabled n pins will drive (n-1)\*(n-2) keys, but provide much improved power handling.
```
caksoylar, line 170: Can we combine this with the `interrupt-gpios` bullet points above? I assume when you have one unset the other one needs to be set, and vice versa.

docs/docs/config/kscan.md
```diff
@@ -400,12 +437,54 @@ Consider a keyboard with a [duplex matrix](https://wiki.ai03.com/books/pcb-desig
...
+        label = "KSCAN";
+
+        interrupt-gpios = <&pro_micro 21 (GPIO_ACTIVE_HIGH|GPIO_PULL_DOWN) >;
+        gpios
+          = <&pro_micro 16 (GPIO_ACTIVE_HIGH|GPIO_PULL_DOWN) >
+          , <&pro_micro 17 (GPIO_ACTIVE_HIGH|GPIO_PULL_DOWN) >
+          , <&pro_micro 18 (GPIO_ACTIVE_HIGH|GPIO_PULL_DOWN) >
+          , <&pro_micro 19 (GPIO_ACTIVE_HIGH|GPIO_PULL_DOWN) >
+          , <&pro_micro 20 (GPIO_ACTIVE_HIGH|GPIO_PULL_DOWN) >
+          ; // addressable space is 5x5, (minus paired values)
```
caksoylar, line 474:
```suggestion
        interrupt-gpios = <&pro_micro 21 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN) >;
        gpios
          = <&pro_micro 16 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN) >
          , <&pro_micro 17 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN) >
          , <&pro_micro 18 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN) >
          , <&pro_micro 19 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN) >
          , <&pro_micro 20 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN) >
          ; // addressable space is 5x5, (minus paired values)
```

docs/docs/config/kscan.md
HookyKB: Actually, this is left over from before a previous change and has come back in during the rebase. The code determines if it needs to use polling based on the presence/absence of the `interrupt-gpios`.

Will cut it out.

docs/docs/config/kscan.md
HookyKB: Sure

caksoylar (approved):
LGTM from docs POV.

caksoylar:
I also don't want to drag the naming issue any longer but I had to point out: All prior references use "charli**e**plexing" while this PR seems to be dropping the "e" everywhere.

petejohanson:
> I also don't want to drag the naming issue any longer but I had to point out: All prior references use "charli**e**plexing" while this PR seems to be dropping the "e" everywhere.

I definitely *screwed this up* in my response, and I'm sure that didn't help things. My apologies!

HookyKB:
> I also don't want to drag the naming issue any longer but I had to point out: All prior references use "charli**e**plexing" while this PR seems to be dropping the "e" everywhere.

Oops. Seems like those with merge capabilities could have just fixed and merged that instead of waiting for me. I'll get on it though. 😉

docs/docs/config/kscan.md
caksoylar:
```suggestion
Definition file: [zmk/app/module/drivers/kscan/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/module/drivers/kscan/Kconfig)
```

docs/docs/config/kscan.md
```diff
@@ -152,6 +152,40 @@ The output pins (e.g. columns for `col2row`) should have the flag `GPIO_ACTIVE_H
...
+| Config                                              | Type        | Description                                                               | Default |
+| --------------------------------------------------- | ----------- | ------------------------------------------------------------------------- | ------- |
+| `CONFIG_ZMK_KSCAN_CHARLIEPLEX_WAIT_BEFORE_INPUTS`   | int (ticks) | How long to wait before reading input pins after setting output active    | 0       |
+| `CONFIG_ZMK_KSCAN_CHARLIEPLEX_WAIT_BETWEEN_OUTPUTS` | int (ticks) | How long to wait between each output to allow previous output to "settle" | 0       |
+
+### Devicetree
+
+Applies to: `compatible = "zmk,kscan-gpio-charlieplex"`
+
+Definition file: [zmk/app/drivers/zephyr/dts/bindings/kscan/zmk,kscan-gpio-charlieplex.yaml](https://github.com/zmkfirmware/zmk/blob/main/app/drivers/zephyr/dts/bindings/kscan/zmk%2Ckscan-gpio-charlieplex.yaml)
```
caksoylar, line 173:
```suggestion
Definition file: [zmk/app/module/dts/bindings/kscan/zmk,kscan-gpio-charlieplex.yaml](https://github.com/zmkfirmware/zmk/blob/main/app/module/dts/bindings/kscan/zmk%2Ckscan-gpio-charlieplex.yaml)
```

caksoylar (commented):
Now that #1975 that fixes the docs links is merged, we should fix the broken links here.

HookyQR:
> Now that #1975 that fixes the docs links is merged, we should fix the broken links here.

Done. Please note, going forward, if you want any changes to this PR, please add them. I don't care how it gets done. I'm not watching this, and won't be making any further changes. Wish I could say it has been fun...

caksoylar:
Noted 👍

ReSummit:
Hello everyone, I tried the Charlieplex code for a Seeeduino Xiao with the testamatta 011. I had some issues trying to implement this because I got errors regarding the gpio output and input pins not being able to be set. While the code compiles correctly, the logs on the microcontroller seem to throw an error that doesn't allow the charlieplex code to run.

image

This basically stops the code from running as any error stops charlieplex from being setup. After looking through the code, I found out that Zephyr does not play well with using the GPIO flags `GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN` when assigning a pin to output. I don't think this satisfies the ZMK clean code guidelines, but how I fixed this was by creating a temp variable for the `gpio_dt_spec` parameter and cloning the parameters except for the flag, which I used `GPIO_ACTIVE_HIGH` for. This fixed the issue, but I feel that a better solution may exist.

HookyKB:
@ReSummit, which version of the Xiao are you using?
Can you share your repo so I can check out the settings. Have you tried just setting the pins without the `| GPIO_PULL_DOWN`?

Also, I doubt you'll get this board to work properly without ghosting as the board is a duplex matrix, not charlieplex.

ReSummit:
> @ReSummit, which version of the Xiao are you using? It has been tested with the Xiao BLE. You can see the config here: [7a9a74e](https://github.com/zmkfirmware/zmk/commit/7a9a74e68734fbde3a17a24260f54acf798b5814)
> 
> Can you share your repo so I can check out the settings. Have you tried just setting the pins without the `| GPIO_PULL_DOWN`?
> 
> Also, I doubt you'll get this board to work properly without ghosting as the board is a duplex matrix, not charlieplex.

I am currently using the Seeeduino Xiao SAMD21. I tried it separately in a Arduino sketch to see if the GPIO input / output switching was a problem but it did fine.

Here is a fork of your repo with the changes I made. Checkout kscan/round_robin and refer to board/shield/011 contents to see the config I have for the board. It does work, though I will note what you mentioned about ghosting:
https://github.com/ReSummit/zmk-charlieplex/tree/kscan/round_robin

I did try doing it without `| GPIO_PULL_DOWN` but this resulted in the microcontroller believing that I pressed all the keys at once (or at the very least, spamming random keyboard inputs to the computer). This is with all the pins without `| GPIO_PULL_DOWN`; if >1 of the pins has `| GPIO_PULL_DOWN`, then it will fail to initialize.

HookyKB:
@ReSummit You might need to disable the console for it to work. I had a problem with console enabled as it was trying to set a pin high when something else was using it.

app/boards/shields/intro/intro.dtsi L14-15:
```
        /delete-property/ zephyr,console;
        /delete-property/ zephyr,shell-uart;
```
And set serial to disable at the bottom.

petejohanson:
> > @ReSummit, which version of the Xiao are you using? It has been tested with the Xiao BLE. You can see the config here: [7a9a74e](https://github.com/zmkfirmware/zmk/commit/7a9a74e68734fbde3a17a24260f54acf798b5814)
> > Can you share your repo so I can check out the settings. Have you tried just setting the pins without the `| GPIO_PULL_DOWN`?
> > Also, I doubt you'll get this board to work properly without ghosting as the board is a duplex matrix, not charlieplex.
> 
> @HookyKB I am currently using the Seeeduino Xiao SAMD21. I tried it separately in a Arduino sketch to see if the GPIO input / output switching was a problem but it did fine.
> 
> Here is a fork of your repo with the changes I made. Checkout kscan/round_robin and refer to board/shield/011 contents to see the config I have for the board. It does work, though I will note what you mentioned about ghosting: https://github.com/ReSummit/zmk-charlieplex/tree/kscan/round_robin
> 
> I did try doing it without `| GPIO_PULL_DOWN` but this resulted in the microcontroller believing that I pressed all the keys at once (or at the very least, spamming random keyboard inputs to the computer). This is with all the pins without `| GPIO_PULL_DOWN`; if >1 of the pins has `| GPIO_PULL_DOWN`, then it will fail to initialize.
> 
> (edit, forgot mention)

This is probably because the SAM0 GPIO driver in Zephyr enforces some checks that the nrfx one doesn't:

* SAM0: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/gpio/gpio_sam0.c#L72
* nrfx: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/gpio/gpio_nrfx.c#L85

Your fix seems so-so, I wonder if perhaps we should only be specifying the active high/low in the DTS, and then inferring the pull flag to add whenever setting a pin as in input.

ReSummit:
> @ReSummit Nice to know it _works_. Hopefully this can be a solution for duplex as well.
> 
> You might need to disable the console for it to work. I had a problem with console enabled as it was trying to set a pin high when something else was using it.
> 
> app/boards/shields/intro/intro.dtsi L14-15:
> 
> ```
>         /delete-property/ zephyr,console;
>         /delete-property/ zephyr,shell-uart;
> ```
> 
> And set serial to disable at the bottom.

@HookyKB Unfortunately this doesn't change anything. Just recompiled and no keys are detected being pressed. I actually needed those properties enabled to even get a USB serial debug and get the error messages.

> This is probably because the SAM0 GPIO driver in Zephyr enforces some checks that the nrfx one doesn't:
> 
> * SAM0: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/gpio/gpio_sam0.c#L72
> * nrfx: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/gpio/gpio_nrfx.c#L85
> 
> Your fix seems so-so, I wonder if perhaps we should only be specifying the active high/low in the DTS, and then inferring the pull flag to add whenever setting a pin as in input.

@petejohanson This was actually what I was looking for when I saw the error code that was returned from `gpio_pin_configure_dt` as I got an error code `-ENOTSUP` which happened to be -134. Especially in the `gpio_sam0_config` function:
image

Inferring the pull flag sounds like a good idea, but part of me wonders if there should be another parameter for that. Probably a later problem though...?

petejohanson:
@ReSummit I just push an additional commit:

```
commit a0cc94e70bfb6662920b7f8730bb8f5048c40926 (HEAD -> pr/HookyKB/1694, HookyKB/kscan/round_robin)
Author: Peter Johanson <peter@peterjohanson.com>
Date:   Fri Nov 24 05:45:42 2023 +0000

    refactor(kscan): Infer pull flags, use LISTIFY.
    
    * To avoid issues with platforms that enforce no pull flags when
      pins are used as outputs, infer the pull flags in code instead of
      setting them in the DTS for charlieplex kscan driver.
    * Use `LISTIFY` macro instead of deprecated `UTIL_LISTIFY`.
```

Can you please test?

ReSummit:
> @ReSummit I just push an additional commit:
> 
> ```
> commit a0cc94e70bfb6662920b7f8730bb8f5048c40926 (HEAD -> pr/HookyKB/1694, HookyKB/kscan/round_robin)
> Author: Peter Johanson <peter@peterjohanson.com>
> Date:   Fri Nov 24 05:45:42 2023 +0000
> 
>     refactor(kscan): Infer pull flags, use LISTIFY.
>     
>     * To avoid issues with platforms that enforce no pull flags when
>       pins are used as outputs, infer the pull flags in code instead of
>       setting them in the DTS for charlieplex kscan driver.
>     * Use `LISTIFY` macro instead of deprecated `UTIL_LISTIFY`.
> ```
> 
> Can you please test?

Testing at the moment but uh, it doesn't compile... This error shows up:

```bash
<zmk-path>/app/module/drivers/kscan/kscan_gpio_charlieplex.c:398:52: error: expected expression before ',' token
  398 |         LISTIFY(INST_LEN(n), KSCAN_GPIO_CFG_INIT, (, ), n)};                                       \
      |
```

I changed the GPIO flags to now include `| GPIO_PULL_DOWN` but the same error appears.

petejohanson:
@ReSummit Seems to compile here fine, but I've identified a small fix that *likely* was the issue. Please try again?

ReSummit:
That fixed the compile error. Uploaded the code but the error from before persists
image

ReSummit:
Looking back through the code that was added and I think it has to do with the set_as_output function as I think the flags are still there (because in my code modifications, I filtered out all PULLUP / PULLDOWN flags that were causing problems).

petejohanson:
> Looking back through the code that was added and I think it has to do with the set_as_output function as I think the flags are still there (because in my code modifications, I filtered out all PULLUP / PULLDOWN flags that were causing problems).

It doesn't work after removing the pull flag from your DTS/overlay?

ReSummit:
> > Looking back through the code that was added and I think it has to do with the set_as_output function as I think the flags are still there (because in my code modifications, I filtered out all PULLUP / PULLDOWN flags that were causing problems).
> 
> It doesn't work after removing the pull flag from your DTS/overlay?

Whoops, forgot to take out the flags, uploaded again and am now seeing a different logged output, but none of the keys work. The logs seem to show a debug log about sending events:
```
.
..
...
[00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,4 state on
[00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,5 state on
[00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,6 state on
[00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,7 state on
[00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,8 state on
[00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 1, pressed: true
[00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 2, pressed: true
[00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 3, pressed: true
[00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 4, pressed: true
```

petejohanson:
> > > Looking back through the code that was added and I think it has to do with the set_as_output function as I think the flags are still there (because in my code modifications, I filtered out all PULLUP / PULLDOWN flags that were causing problems).
> > 
> > 
> > It doesn't work after removing the pull flag from your DTS/overlay?
> 
> Whoops, forgot to take out the flags, uploaded again and am now seeing a different logged output, but none of the keys work. The logs seem to show a debug log about sending events:
> 
> ```
> .
> ..
> ...
> [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,4 state on
> [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,5 state on
> [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,6 state on
> [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,7 state on
> [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,8 state on
> [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 1, pressed: true
> [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 2, pressed: true
> [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 3, pressed: true
> [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 4, pressed: true
> ```

Looks like my bit checking logic was errant. Just pushed another fix for testing. Thanks!

ReSummit:
> > > > Looking back through the code that was added and I think it has to do with the set_as_output function as I think the flags are still there (because in my code modifications, I filtered out all PULLUP / PULLDOWN flags that were causing problems).
> > > 
> > > 
> > > It doesn't work after removing the pull flag from your DTS/overlay?
> > 
> > 
> > Whoops, forgot to take out the flags, uploaded again and am now seeing a different logged output, but none of the keys work. The logs seem to show a debug log about sending events:
> > ```
> > .
> > ..
> > ...
> > [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,4 state on
> > [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,5 state on
> > [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,6 state on
> > [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,7 state on
> > [00:00:00.013,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 9,8 state on
> > [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 1, pressed: true
> > [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 2, pressed: true
> > [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 3, pressed: true
> > [00:00:00.014,000] <wrn> zmk: Not found in transform: row: 0, col: 4, pressed: true
> > ```
> 
> Looks like my bit checking logic was errant. Just pushed another fix for testing. Thanks!

Just compiled and uploaded. The keys are working now! Just in case anything looks weird, I attached a keypress debug but it looks good!

```
[00:00:00.000,000] <dbg> zmk: zmk_usb_get_conn_state: state: 11
[00:00:00.000,000] <dbg> zmk: get_selected_transport: No endpoint transports are ready.
[00:00:00.000,000] <inf> zmk: Welcome to ZMK!

[00:00:00.197,000] <inf> usb_hid: Device reset detected
[00:00:00.198,000] <dbg> zmk: zmk_usb_get_conn_state: state: 1
[00:00:00.198,000] <dbg> zmk: zmk_usb_get_conn_state: state: 1
[00:00:00.198,000] <dbg> zmk: get_selected_transport: No endpoint transports are ready.                                  
[00:00:00.325,000] <inf> usb_hid: Device reset detected                                                                  
[00:00:00.325,000] <dbg> zmk: zmk_usb_get_conn_state: state: 1                                                           
[00:00:00.325,000] <dbg> zmk: zmk_usb_get_conn_state: state: 1                                                           
[00:00:00.325,000] <dbg> zmk: get_selected_transport: No endpoint transports are ready.                                  
[00:00:00.431,000] <inf> usb_hid: Device configured
[00:00:00.431,000] <dbg> zmk: zmk_usb_get_conn_state: state: 3
[00:00:00.431,000] <dbg> zmk: zmk_usb_get_conn_state: state: 3
[00:00:00.431,000] <dbg> zmk: get_selected_transport: Only USB is ready.
[00:00:06.213,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 6,0 state on
[00:00:06.214,000] <dbg> zmk: zmk_kscan_process_msgq: Row: 6, col: 0, position: 6, pressed: true
[00:00:06.214,000] <dbg> zmk: position_state_changed_listener: 6 bubble (no undecided hold_tap active)
[00:00:06.214,000] <dbg> zmk: zmk_keymap_apply_position_state: layer: 0 position: 6, binding name: KEY_PRESS
[00:00:06.214,000] <dbg> zmk: on_keymap_binding_pressed: position 6 keycode 0x7001C
[00:00:06.214,000] <dbg> zmk: hid_listener_keycode_pressed: usage_page 0x07 keycode 0x1C implicit_mods 0x00 explicit_mods 0x00
[00:00:06.214,000] <dbg> zmk: zmk_hid_implicit_modifiers_press: Modifiers set to 0x00
[00:00:06.214,000] <dbg> zmk: zmk_endpoints_send_report: usage page 0x07
[00:00:06.264,000] <dbg> zmk: kscan_charlieplex_read: Sending event at 6,0 state off
[00:00:06.264,000] <dbg> zmk: zmk_kscan_process_msgq: Row: 6, col: 0, position: 6, pressed: false
[00:00:06.264,000] <dbg> zmk: position_state_changed_listener: 6 bubble (no undecided hold_tap active)
[00:00:06.264,000] <dbg> zmk: zmk_keymap_apply_position_state: layer: 0 position: 6, binding name: KEY_PRESS
[00:00:06.264,000] <dbg> zmk: on_keymap_binding_released: position 6 keycode 0x7001C
[00:00:06.264,000] <dbg> zmk: hid_listener_keycode_released: usage_page 0x07 keycode 0x1C implicit_mods 0x00 explicit_mods 0x00
[00:00:06.264,000] <dbg> zmk: zmk_hid_implicit_modifiers_release: Modifiers set to 0x00
[00:00:06.264,000] <dbg> zmk: zmk_endpoints_send_report: usage page 0x07
```

petejohanson (approved):
Rebased this and removed the deprecated `label` property. LGTM, will merge once the tests pass. Thanks @HookyKB !

PR #1714, opened 2023-03-17
feat(shields): Add MechWild BB65

kylemccreery:
## Board/Shield Check-list


app/boards/shields/bb65/bb65.yml
caksoylar, line 1: I think this file is supposed to be named `bb65.zmk.yml` according to [metadata specs](https://zmk.dev/docs/development/hardware-metadata-files) -- same for #1713.

app/boards/shields/bb65/boards/pillbug.conf
caksoylar, line 1: Is there a reason for these settings to be in the board-specific file (and always enabled unless disabled explicitly) rather than in `bb65.conf`? Also applies to #1713.

kylemccreery: Yes. The logic here is that if you use an alternate board for this shield (such as the blackpill, which is supported) the pin defs wont be the same, so it will not have the pillbug.overlay file to reference and will throw errors without the feature being broken out.

app/boards/shields/bb65/bb65.yml
kylemccreery: Noted. Making changes. This was first written some time ago and I just copied it over.

kylemccreery: Fixed.

app/boards/shields/bb65/boards/pillbug.conf
caksoylar: I see, that makes sense. My only concern would be that it is undiscoverable for users if they want to disable it, since it doesn't appear in the files that get copied to the user config repo. If that is something you'd want, you could add a commented out `CONFIG_ZMK_RGB_UNDERGLOW=n` so at least it is more discoverable. (If you do that, you might also want to test the build to make sure it will actually override the setting if uncommented.)

(I'd also tend towards not enabling optional features by default but I guess that's a vendor/maintainer decision.)

kylemccreery: I see the point you are making. I based this off a fairly old way of approaching this. I will make the changes to put the conf into the main file and not keep it restricted here. This will proc issues in the case where someone wants to build with a blackpill and tries to enable it, but I suppose I can just do the thing I wanted to avoid and define those boards as well.

kylemccreery: Wanted to avoid because lazy and I don't have one built right now to use it so I can't test the light defs. I can do it later though and add to the PR.

caksoylar: I think it'd be good for more blackpill support in ZMK, but I don't necessarily want to force you one way or the other!

Also worth noting that we have the longstanding issue #885 related to board vs. shield-specific settings for underglow and it affects more common board/shield combinations (e.g. `nice_nano_v2` and `corne`).

kylemccreery: Oh the change to mapped pins would be perfect here too. I wanted to add the blackpill support for these boards when I did this PR, but got lazy since I can't find my screwdriver. Hell, they are the reason I added the blackpill defs to the repo in the first place.

caksoylar (approved):

Nick-Munnich:

PR #1722, opened 2023-03-24
Refactor events API to avoid heap/dynamic allocations

petejohanson:
Ok, this is a semi-major core refactor to address one of the major nits I had with the initial design for the event manager API I added to ZMK, mainly that it depending on dynamic memory allocation for alloc/free of the event structs. As a result, we need to use some large, arbitrary heap pool in ZMK to be sure we have enough allocated for whatever odd combination of events get raised/captured in the system.

This PR moves to stack allocated event structures, and change the expectation that anyone wanting to capture an even should copy it into a locally allocated storage for it's use as needed. See updates to hold-tap, combos, etc that demonstrate this kind of change.

This potentially disrupts any open behavior PRs, so need to consider the *timing* of the merge of this, but want to get this reviewed for the general approach now, since this I believe is ready, after I found and fixed the final bug in combos I'd introduced.

### DIFF
--- a/app/include/zmk/event_manager.h
+++ b/app/include/zmk/event_manager.h
@@ -38,7 +38,8 @@ struct zmk_event_subscription {
         zmk_event_t header;                                                                        \
         struct event_type data;                                                                    \
     };                                                                                             \
-    struct event_type##_event *new_##event_type(struct event_type);                                \
+    struct event_type##_event copy_raised_##event_type(const struct event_type *ev);               \
+    int raise_##event_type(struct event_type);                                                     \
     struct event_type *as_##event_type(const zmk_event_t *eh);                                     \
     extern const struct zmk_event_type zmk_event_##event_type;
 
@@ -46,12 +47,14 @@ struct zmk_event_subscription {
     const struct zmk_event_type zmk_event_##event_type = {.name = STRINGIFY(event_type)};          \
     const struct zmk_event_type *zmk_event_ref_##event_type __used                                 \
         __attribute__((__section__(".event_type"))) = &zmk_event_##event_type;                     \
-    struct event_type##_event *new_##event_type(struct event_type data) {                          \
-        struct event_type##_event *ev =                                                            \
-            (struct event_type##_event *)k_malloc(sizeof(struct event_type##_event));              \
-        ev->header.event = &zmk_event_##event_type;                                                \
-        ev->data = data;                                                                           \
-        return ev;                                                                                 \
+    struct event_type##_event copy_raised_##event_type(const struct event_type *ev) {              \
+        struct event_type##_event *outer = CONTAINER_OF(ev, struct event_type##_event, data);      \
+        return *outer;                                                                             \
+    };                                                                                             \
+    int raise_##event_type(struct event_type data) {                                               \
+        struct event_type##_event ev = {.data = data};                                             \
+        ev.header.event = &zmk_event_##event_type;                                                 \
+        return ZMK_EVENT_RAISE(ev);                                                                \
     };                                                                                             \
     struct event_type *as_##event_type(const zmk_event_t *eh) {                                    \
         return (eh->event == &zmk_event_##event_type) ? &((struct event_type##_event *)eh)->data   \
@@ -68,17 +71,15 @@ struct zmk_event_subscription {
             .listener = &zmk_listener_##mod,                                                       \
     };
 
-#define ZMK_EVENT_RAISE(ev) zmk_event_manager_raise((zmk_event_t *)ev);
+#define ZMK_EVENT_RAISE(ev) zmk_event_manager_raise((zmk_event_t *)&ev);
 
 #define ZMK_EVENT_RAISE_AFTER(ev, mod)                                                             \
-    zmk_event_manager_raise_after((zmk_event_t *)ev, &zmk_listener_##mod);
+    zmk_event_manager_raise_after((zmk_event_t *)&ev, &zmk_listener_##mod);
 
 #define ZMK_EVENT_RAISE_AT(ev, mod)                                                                \
-    zmk_event_manager_raise_at((zmk_event_t *)ev, &zmk_listener_##mod);
+    zmk_event_manager_raise_at((zmk_event_t *)&ev, &zmk_listener_##mod);
 
-#define ZMK_EVENT_RELEASE(ev) zmk_event_manager_release((zmk_event_t *)ev);
-
-#define ZMK_EVENT_FREE(ev) k_free((void *)ev);
+#define ZMK_EVENT_RELEASE(ev) zmk_event_manager_release((zmk_event_t *)&ev);
 
 int zmk_event_manager_raise(zmk_event_t *event);
 int zmk_event_manager_raise_after(zmk_event_t *event, const struct zmk_listener *listener);
--- a/app/include/zmk/event_manager.h
+++ b/app/include/zmk/event_manager.h
@@ -52,8 +52,8 @@ struct zmk_event_subscription {
         return *outer;                                                                             \
     };                                                                                             \
     int raise_##event_type(struct event_type data) {                                               \
-        struct event_type##_event ev = {.data = data};                                             \
-        ev.header.event = &zmk_event_##event_type;                                                 \
+        struct event_type##_event ev = {.data = data,                                              \
+                                        .header = {.event = &zmk_event_##event_type}};             \
         return ZMK_EVENT_RAISE(ev);                                                                \
     };                                                                                             \
     struct event_type *as_##event_type(const zmk_event_t *eh) {                                    \
@@ -71,15 +71,19 @@ struct zmk_event_subscription {
             .listener = &zmk_listener_##mod,                                                       \
     };
 
-#define ZMK_EVENT_RAISE(ev) zmk_event_manager_raise((zmk_event_t *)&ev);
+#define ZMK_ASSERT_EVENT_LIKE(ev)                                                                  \
+    (__ASSERT((uint8_t *)&(ev).header - (uint8_t *)&ev == 0,                                       \
+              "header must be first element of event"))
+
+#define ZMK_EVENT_RAISE(ev) (ZMK_ASSERT_EVENT_LIKE(ev), zmk_event_manager_raise(&(ev).header))
 
 #define ZMK_EVENT_RAISE_AFTER(ev, mod)                                                             \
-    zmk_event_manager_raise_after((zmk_event_t *)&ev, &zmk_listener_##mod);
+    (ZMK_ASSERT_EVENT_LIKE(ev), zmk_event_manager_raise_after(&(ev).header, &zmk_listener_##mod))
 
 #define ZMK_EVENT_RAISE_AT(ev, mod)                                                                \
-    zmk_event_manager_raise_at((zmk_event_t *)&ev, &zmk_listener_##mod);
+    (ZMK_ASSERT_EVENT_LIKE(ev), zmk_event_manager_raise_at(&(ev).header, &zmk_listener_##mod))
 
-#define ZMK_EVENT_RELEASE(ev) zmk_event_manager_release((zmk_event_t *)&ev);
+#define ZMK_EVENT_RELEASE(ev) (ZMK_ASSERT_EVENT_LIKE(ev), zmk_event_manager_release(&(ev).header))
 
 int zmk_event_manager_raise(zmk_event_t *event);
 int zmk_event_manager_raise_after(zmk_event_t *event, const struct zmk_listener *listener);
--- a/app/include/zmk/events/mouse_button_state_changed.h
+++ b/app/include/zmk/events/mouse_button_state_changed.h
@@ -19,8 +19,8 @@ struct zmk_mouse_button_state_changed {
 
 ZMK_EVENT_DECLARE(zmk_mouse_button_state_changed);
 
-static inline struct zmk_mouse_button_state_changed_event *
-zmk_mouse_button_state_changed_from_encoded(uint32_t encoded, bool pressed, int64_t timestamp) {
-    return new_zmk_mouse_button_state_changed((struct zmk_mouse_button_state_changed){
+static inline int raise_zmk_mouse_button_state_changed_from_encoded(uint32_t encoded, bool pressed,
+                                                                    int64_t timestamp) {
+    return raise_zmk_mouse_button_state_changed((struct zmk_mouse_button_state_changed){
         .buttons = ZMK_HID_USAGE_ID(encoded), .state = pressed, .timestamp = timestamp});
 }
--- a/app/src/behaviors/behavior_hold_tap.c
+++ b/app/src/behaviors/behavior_hold_tap.c
@@ -88,7 +88,24 @@ struct active_hold_tap {
 struct active_hold_tap *undecided_hold_tap = NULL;
 struct active_hold_tap active_hold_taps[ZMK_BHV_HOLD_TAP_MAX_HELD] = {};
 // We capture most position_state_changed events and some modifiers_state_changed events.
-const zmk_event_t *captured_events[ZMK_BHV_HOLD_TAP_MAX_CAPTURED_EVENTS] = {};
+
+enum captured_event_tag {
+    ET_NONE,
+    ET_POS_CHANGED,
+    ET_CODE_CHANGED,
+};
+
+union captured_event_data {
+    struct zmk_position_state_changed_event position;
+    struct zmk_keycode_state_changed_event keycode;
+};
+
+struct captured_event {
+    enum captured_event_tag tag;
+    union captured_event_data data;
+};
+
+struct captured_event captured_events[ZMK_BHV_HOLD_TAP_MAX_CAPTURED_EVENTS] = {};
 
 // Keep track of which key was tapped most recently for the standard, if it is a hold-tap
 // a position, will be given, if not it will just be INT32_MIN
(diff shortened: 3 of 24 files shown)


caksoylar:
I have been testing this full time for the last week on a split with keymap including hold-taps, combos and BLE profile change events with no issues. (Also mouse keys as a bonus, with [this patch](https://github.com/caksoylar/zmk/commit/d1a3069) to get it working on this PR.)

Among changes covered here, I haven't tested with sticky keys, key toggles, key repeat, layer state and USB status change events, WPM events, sensor/encoders.

petejohanson:
> I have been testing this full time for the last week on a split with keymap including hold-taps, combos and BLE profile change events with no issues. (Also mouse keys as a bonus, with [this patch](https://github.com/caksoylar/zmk/commit/d1a3069) to get it working on this PR.)
> 
> Among changes covered here, I haven't tested with sticky keys, key toggles, key repeat, layer state and USB status change events, WPM events, sensor/encoders.

I've just flashed to my revxlp RP2040 last night with a keymap that includes combos, sticky keys, caps word, etc. and will daily that with this change the next week as well. First day of use was uneventful.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -184,25 +200,33 @@ static void release_captured_events() {
     // [k1_down, k1_up, null, null, null, ...]
     // now mt2 will start releasing it's own captured positions.
     for (int i = 0; i < ZMK_BHV_HOLD_TAP_MAX_CAPTURED_EVENTS; i++) {
-        const zmk_event_t *captured_event = captured_events[i];
-        if (captured_event == NULL) {
+        struct captured_event captured_event = captured_events[i];
```
joelspadin, line 203: Same as above.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -632,7 +656,10 @@ static int position_state_changed_listener(const zmk_event_t *eh) {
 
     LOG_DBG("%d capturing %d %s event", undecided_hold_tap->position, ev->position,
             ev->state ? "down" : "up");
-    capture_event(eh);
+    struct captured_event capture;
+    capture.tag = ET_POS_CHANGED;
+    memcpy(&capture.data.position, eh, sizeof(struct zmk_position_state_changed_event));
```
joelspadin, line 661: It seems weird to have used `ev = as_zmk_position_state_changed(eh)` earlier, but then have to go back to using `eh` in order to copy the whole event. I feel like that could be error-prone, since there's no obvious way to copy an event in a type-safe way.

Maybe we could generate a function per event that uses CONTAINER_OF to get a `struct <event_type>_event*` from a `struct <event_type>*`, so then you could use assignment instead of memcpy? Then the compiler would be able to ensure that the types match.

```suggestion
    struct captured_event capture = {
      .tag = ET_POS_CHANGED,
      .data = { .position = *zmk_position_state_changed_container(ev) },
    };
```

(Not really sure what to call said function.) The main downside to this is that there's no guarantee that a `struct <event_type>` is actually contained within a `struct <event_type>_event`. 

Alteratively, if we only ever need to use that to copy a whole event (which is the case as far as I've gotten on this review so far), it could be called "copy" and return the whole event by value instead of as a pointer to the event's container. This has the same problem with if the data isn't part of an event, but it makes the purpose of the function clearer.

Alternatively (and this would require changing a *lot* of other code, but it might we worth it for making copying events simpler and safer), change the `as_<event_type>()` functions to return a pointer to the `_event` struct instead of just the data, and then no extra function is needed to copy the event, but you do have to write a bunch of `.data` everywhere.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -659,7 +686,10 @@ static int keycode_state_changed_listener(const zmk_event_t *eh) {
     // if a undecided_hold_tap is active.
     LOG_DBG("%d capturing 0x%02X %s event", undecided_hold_tap->position, ev->keycode,
             ev->state ? "down" : "up");
-    capture_event(eh);
+    struct captured_event capture;
+    capture.tag = ET_CODE_CHANGED;
+    memcpy(&capture.data.keycode, eh, sizeof(struct zmk_keycode_state_changed_event));
```
joelspadin, line 691: Same as above

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -234,7 +234,9 @@ static int sticky_key_keycode_state_changed_listener(const zmk_event_t *eh) {
                 if (sticky_key->config->quick_release) {
                     // immediately release the sticky key after the key press is handled.
                     if (!event_reraised) {
-                        ZMK_EVENT_RAISE_AFTER(eh, behavior_sticky_key);
+                        struct zmk_keycode_state_changed_event dupe_ev;
+                        memcpy(&dupe_ev, eh, sizeof(struct zmk_keycode_state_changed_event));
```
joelspadin, line 238: See earlier comment about finding a better way to copy events.

app/src/combo.c
```diff
@@ -262,37 +260,34 @@ static int clear_candidates() {
 }
 
 static int capture_pressed_key(const zmk_event_t *ev) {
```
joelspadin, line 262: Would be nice if this took a `const struct zmk_position_state_changed_event*` so it couldn't easily be called with the wrong type of event. Then line 267 could just assign instead of using memcpy as well.

See also my earlier comment about finding a better way to copy events. It could alternatively take a `const struct zmk_position_state_changed*` which is assumed to be part of an event, then use the container/copy function to get the data to copy.

app/src/combo.c
```diff
@@ -262,37 +260,34 @@ static int clear_candidates() {
...
-            continue;
-        }
-        pressed_keys[i] = ev;
-        return ZMK_EV_EVENT_CAPTURED;
+    if (pressed_keys_count == CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY) {
+        return ZMK_EV_EVENT_BUBBLE;
     }
-    return ZMK_EV_EVENT_BUBBLE;
+
+    memcpy(&pressed_keys[pressed_keys_count++], ev,
```
joelspadin, line 267: See earlier comment about finding a better way to copy events

app/src/combo.c
```diff
@@ -262,37 +260,34 @@ static int clear_candidates() {
...
-            LOG_DBG("combo: reraising position event %d",
-                    as_zmk_position_state_changed(captured_event)->position);
-            ZMK_EVENT_RAISE(captured_event);
+            LOG_DBG("combo: reraising position event %d", ev.data.position);
+            ZMK_EVENT_RAISE(ev);
         }
     }
-    return CONFIG_ZMK_COMBO_MAX_KEYS_PER_COMBO;
+
+    // pressed_keys_count = 0;
```
joelspadin, line 289: Leftover code from testing? Or, the old code seemed to clear out the pressed_keys[] array. Is this actually needed here?

app/src/combo.c
```diff
@@ -262,37 +260,34 @@ static int clear_candidates() {
...
-    for (int i = 0; i < CONFIG_ZMK_COMBO_MAX_KEYS_PER_COMBO; i++) {
-        const zmk_event_t *captured_event = pressed_keys[i];
-        if (pressed_keys[i] == NULL) {
-            return i;
-        }
-        pressed_keys[i] = NULL;
+    uint32_t count = pressed_keys_count;
+    pressed_keys_count = 0;
+    for (int i = 0; i < count; i++) {
+        struct zmk_position_state_changed_event ev = pressed_keys[i];
```
joelspadin, line 278: Make variable a pointer to avoid copy.

app/src/combo.c
```diff
@@ -469,7 +463,9 @@ static int position_state_up(const zmk_event_t *ev, struct zmk_position_state_ch
     if (released_keys > 1) {
         // The second and further key down events are re-raised. To preserve
         // correct order for e.g. hold-taps, reraise the key up event too.
-        ZMK_EVENT_RAISE(ev);
+        struct zmk_position_state_changed_event dupe_ev;
+        memcpy(&dupe_ev, ev, sizeof(struct zmk_position_state_changed_event));
```
joelspadin, line 467: See earlier comment about finding a better way to copy events

app/include/zmk/event_manager.h
```diff
@@ -38,21 +38,24 @@ struct zmk_event_subscription {
         zmk_event_t header;                                                                        \
         struct event_type data;                                                                    \
     };                                                                                             \
-    struct event_type##_event *new_##event_type(struct event_type);                                \
+    struct event_type##_event new_##event_type(struct event_type);                                 \
+    int raise_new_##event_type(struct event_type);                                                 \
```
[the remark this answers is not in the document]
petejohanson: So.... Part of the reason for creating the `RAISE_` macros was because we *can't* raise the event by passing the pointer to it if there isn't actually a local created to create an owner that has a proper scope so it'll be cleared from the stack when the local function exits.

So we can't just pass the result *directly* to the raise function. *But* I'm fine if the `create_` terminology reads better.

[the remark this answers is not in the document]
joelspadin: To clarify this and my previous comment, it looks like the only other thing using the `new_<event>` function is `create_layer_state_changed()`. If `create_layer_state_changed()` were to return a `struct zmk_layer_state_changed` instead of a `struct zmk_layer_state_changed_event`, then you could just pass the result to `raise_new_zmk_layer_state_changed_event()` and have it wrap the data in an event for you, which would simplify the code a bit:

```diff
- struct zmk_layer_state_changed_event ev = create_layer_state_changed(layer, state);
- ZMK_EVENT_RAISE(ev);
+ zmk_raise_layer_state_changed_event(create_layer_state_changed(layer, state));
```

Then, there would be nothing left using `new_<event>` aside from `raise_new_<event>`. Is there any reason to create an event object but *not* immediately raise it? If not, we could remove the `new_<event>` function entirely (and roll its code into the raise function). If so, it might be nice to rename it to avoid the association with C++'s `new` operator, but that's probably not a big deal.

And then for this comment, I meant specifically the "new" in `raise_new_<event>`. There's no "raise *old* event" function to make a distinction between, so I think the name could be simplified to just `raise_<event>`.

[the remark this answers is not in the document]
petejohanson: Indeed. Ok, so renamed the `raise_` function to skip the `new` and removed the `new_` generated functions entirely in favor of the `raise_` being used uniformly. Thanks!

app/src/combo.c
petejohanson: Nuked.

app/src/combo.c
petejohanson: Yup, Changed this to do just that.

app/src/behaviors/behavior_hold_tap.c
petejohanson: Ok, I took your idea of a the `CONTAINER_OF` approach and am now generating `copy_raised_foo` functions. I included the `raised_` portion to try to somewhat address the "won't work if not actually contained in a <event_type>_event" issue at least in the naming to make it clear.

Updated the few spots (hold taps and combos) to remove the use of `memcpy` and instead call the generated functions as you suggested. 

Looks a *lot* cleaner to me, but would love your take on the updated version. Thanks!

app/include/zmk/events/mouse_button_state_changed.h
```diff
@@ -19,8 +19,8 @@ struct zmk_mouse_button_state_changed {
 
 ZMK_EVENT_DECLARE(zmk_mouse_button_state_changed);
 
-static inline struct zmk_mouse_button_state_changed_event *
-zmk_mouse_button_state_changed_from_encoded(uint32_t encoded, bool pressed, int64_t timestamp) {
-    return new_zmk_mouse_button_state_changed((struct zmk_mouse_button_state_changed){
+static inline int raise_zmk_mouse_button_state_changed_from_encoded(uint32_t encoded, bool pressed,
```
joelspadin, line 22: Should we add a `raise_zmk_keycode_state_changed_from_encoded()` function too?

app/src/behaviors/behavior_sticky_key.c
joelspadin: Should this use `copy_raised_zmk_keycode_state_changed()` now?

app/src/combo.c
joelspadin: Not yet changed to a pointer.

joelspadin (changes requested):

app/src/combo.c
petejohanson: So not sure about this, since the only use is to pass this to `ZMK_EVENT_RELEASE` or `ZMK_EVENT_RAISE` which expect an item on the stack.

app/include/zmk/events/mouse_button_state_changed.h
petejohanson: Indeed. Added and refactored to use.

app/src/combo.c
petejohanson: j/k. Fixed.

app/include/zmk/event_manager.h
```diff
@@ -68,17 +71,15 @@ struct zmk_event_subscription {
             .listener = &zmk_listener_##mod,                                                       \
     };
 
-#define ZMK_EVENT_RAISE(ev) zmk_event_manager_raise((zmk_event_t *)ev);
+#define ZMK_EVENT_RAISE(ev) zmk_event_manager_raise((zmk_event_t *)&ev);
```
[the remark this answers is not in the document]
petejohanson: Indeed. Fixed.

app/include/zmk/event_manager.h
```diff
@@ -68,17 +71,22 @@ struct zmk_event_subscription {
             .listener = &zmk_listener_##mod,                                                       \
     };
 
-#define ZMK_EVENT_RAISE(ev) zmk_event_manager_raise((zmk_event_t *)ev);
+#define ZMK_EVENT_RAISE(ev)                                                                        \
+    ((__ASSERT((uint8_t *)&(ev).header - (uint8_t *)&ev == 0,                                      \
+               "header must be first element of event")),                                          \
+     zmk_event_manager_raise(&(ev).header));
```
joelspadin, line 77:
```suggestion
     zmk_event_manager_raise(&(ev).header))
```
If these should be usable in expressions like `return ZMK_EVENT_RAISE(ev)`, then they shouldn't end in a semicolon. That will prevent other things that should work like
```c
result = ZMK_EVENT_RAISE(ev) ? OK : ERROR;
```
and it also allows 
```c
int test(void) {
  return ZMK_EVENT_RAISE(ev)  // no semicolon
}
```
to compile, which just feels weird.

app/include/zmk/event_manager.h
```diff
@@ -68,17 +71,22 @@ struct zmk_event_subscription {
...
 
 #define ZMK_EVENT_RAISE_AFTER(ev, mod)                                                             \
-    zmk_event_manager_raise_after((zmk_event_t *)ev, &zmk_listener_##mod);
+    ((__ASSERT((uint8_t *)&(ev).header - (uint8_t *)&ev == 0,                                      \
+               "header must be first element of event")),                                          \
+     zmk_event_manager_raise_after(&(ev).header, &zmk_listener_##mod));
```
joelspadin, line 82:
```suggestion
     zmk_event_manager_raise_after(&(ev).header, &zmk_listener_##mod))
```

app/include/zmk/event_manager.h
```diff
@@ -68,17 +71,22 @@ struct zmk_event_subscription {
...
 
 #define ZMK_EVENT_RAISE_AT(ev, mod)                                                                \
-    zmk_event_manager_raise_at((zmk_event_t *)ev, &zmk_listener_##mod);
-
-#define ZMK_EVENT_RELEASE(ev) zmk_event_manager_release((zmk_event_t *)ev);
+    ((__ASSERT((uint8_t *)&(ev).header - (uint8_t *)&ev == 0,                                      \
+               "header must be first element of event")),                                          \
+     zmk_event_manager_raise_at(&(ev).header, &zmk_listener_##mod));
```
joelspadin, line 87:
```suggestion
     zmk_event_manager_raise_at(&(ev).header, &zmk_listener_##mod))
```

app/include/zmk/event_manager.h
```diff
@@ -68,17 +71,22 @@ struct zmk_event_subscription {
...
 
-#define ZMK_EVENT_FREE(ev) k_free((void *)ev);
+#define ZMK_EVENT_RELEASE(ev) zmk_event_manager_release((zmk_event_t *)&ev);
```
joelspadin, line 89:
```suggestion
#define ZMK_EVENT_RELEASE(ev) zmk_event_manager_release((zmk_event_t *)&ev)
```

Also, I think we want to add the assert and use `&ev.header` instead of a cast here too?

app/include/zmk/event_manager.h
```diff
@@ -68,17 +71,22 @@ struct zmk_event_subscription {
...
-#define ZMK_EVENT_RAISE(ev) zmk_event_manager_raise((zmk_event_t *)ev);
+#define ZMK_EVENT_RAISE(ev)                                                                        \
+    ((__ASSERT((uint8_t *)&(ev).header - (uint8_t *)&ev == 0,                                      \
+               "header must be first element of event")),                                          \
+     zmk_event_manager_raise(&(ev).header));
 
 #define ZMK_EVENT_RAISE_AFTER(ev, mod)                                                             \
-    zmk_event_manager_raise_after((zmk_event_t *)ev, &zmk_listener_##mod);
+    ((__ASSERT((uint8_t *)&(ev).header - (uint8_t *)&ev == 0,                                      \
+               "header must be first element of event")),                                          \
```
joelspadin, line 81: I recommend creating a macro like
```c
#define ZMK_ASSERT_IS_EVENT(ev) \
    (__ASSERT((uint8_t *)&(ev).header - (uint8_t *)&ev == 0,                                      \
               "header must be first element of event"))
```
to simplify this. (Feel free to come up with a better name for this, as "is event" isn't strictly accurate since it only checks the header member, not the type.)

joelspadin (changes requested):

joelspadin (approved):

PR #1730, opened 2023-04-02
fix(ci): Refactor prepare variable step

filterpaper:
* Refactor preparation step using shell parameter expansion to insert shield variable if `matrix.shield` is not null.
* Use quotes as simple guards against input injection.

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -54,18 +54,13 @@ jobs:
     steps:
       - name: Prepare variables
         shell: sh -x {0}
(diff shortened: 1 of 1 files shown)


caksoylar (approved):
Looks good to me! Tested at https://github.com/caksoylar/zmk-config/actions/runs/4648672180 (board/shield combinations, with multiple shields) and https://github.com/caksoylar/zmk-config/actions/runs/4649024079 (board-only), works as expected.

petejohanson (approved):
Thanks!

PR #1736, opened 2023-04-06
fix(docs): ignore-modifiers clarification

quelcom:
Clarify the potential need of adding the `ignore-modifiers;` setting for new sticky key behaviors when combining multiple modifiers into a single key.

### DIFF
--- a/docs/docs/behaviors/sticky-key.md
+++ b/docs/docs/behaviors/sticky-key.md
@@ -75,7 +75,7 @@ This configuration would apply to all sticky keys. This may not be appropriate i
 };
 ```
 
-If you combine multiple modifiers in a single key, like for example `&skq LS(LALT)`, you might need to add the `ignore-modifiers;` setting because it is not inherited from the default `&sq` behaviour.
+If you combine multiple modifiers in a single key, like for example `&skq LS(LALT)`, you might need to add the `ignore-modifiers;` setting because it is not inherited from the default `&sq` behavior.
 
 ### Advanced usage
 
--- a/docs/docs/behaviors/sticky-key.md
+++ b/docs/docs/behaviors/sticky-key.md
@@ -36,7 +36,7 @@ Some typists may find that using a sticky shift key interspersed with rapid typi
 
 #### `ignore-modifiers`
 
-This setting is enabled by default. It ensures that if a sticky key modifier is pressed before a previously pressed sticky key is released, the modifiers will get combined so you can add more sticky keys or press a regular key to apply the modifiers. This is to accommodate _callum-style mods_ where you are prone to rolling sticky keys. If you want sticky key modifiers to only chain after release, you can disable this setting.
+This setting is enabled by default. It ensures that if a sticky key modifier is pressed before a previously pressed sticky key is released, the modifiers will get combined so you can add more sticky keys or press a regular key to apply the modifiers. This is to accommodate _callum-style mods_ where you are prone to rolling sticky keys. If you want sticky key modifiers to only chain after release, you can disable this setting. Please note that [modifier functions](https://zmk.dev/docs/codes/modifiers#modifier-functions), such as `&sk LS(LALT)`, require `ignore-modifiers` enabled in order to function properly.
 
 #### Example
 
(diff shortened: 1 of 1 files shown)


docs/docs/behaviors/sticky-key.md
```diff
@@ -75,6 +75,8 @@ This configuration would apply to all sticky keys. This may not be appropriate i
 };
 ```
 
+If you combine multiple modifiers in a single key, like for example `&skq LS(LALT)`, you might need to add the `ignore-modifiers;` setting because it is not inherited from the default `&sq` behavior.
```
caksoylar, line 78:
```suggestion
If you combine multiple modifiers in a single key, like for example `&skq LS(LALT)`, you might need to add the `ignore-modifiers;` setting because it is not inherited from the default `&sk` behavior.
```

caksoylar (approved):

nguyendown:
The behavior `skq` is created to disable `ignore-modifiers`, so it is not relevant to add `ignore-modifiers` back to `skq`. It is better to add a notice in the `ignore-modifiers` section, something like: "[Modifier functions](https://zmk.dev/docs/codes/modifiers#modifier-functions), such as `&sk LS(LALT)`, require `ignore-modifiers` enabled in order to function properly."

caksoylar:
> The behavior skq is created to disable ignore-modifiers, so it is not relevant to add ignore-modifiers back to skq.

I don't think that is necessarily true, `quick-release` in general is useful for reducing misfires when typing fast, like `HEllo` after a sticky shift, isn't it?

> It is better to add a notice in the ignore-modifiers section, something like: "[Modifier functions](https://zmk.dev/docs/codes/modifiers#modifier-functions), such as &sk LS(LALT), require ignore-modifiers enabled in order to function properly."

This is a good idea too, but I don't prefer one over the other.

After going over the rest of the page, it iooks like it might be beneficial to mention `ignore-modifiers` in the "Advanced usage" section as well.

quelcom:
Thank you both for the suggestions.

If the `ignore-modifiers;` was part of the `&skq` behavior, this pull request wouldn't even exist. It makes sense that the setting is added in the example because it is enabled by default and the `&skq` behavior is mostly related to the `quick-release` setting.

Will also add a comment regarding combining modifier functions in the `ignore-modifiers` section.

docs/docs/behaviors/sticky-key.md
```diff
@@ -36,7 +36,7 @@ Some typists may find that using a sticky shift key interspersed with rapid typi
 
 #### `ignore-modifiers`
 
-This setting is enabled by default. It ensures that if a sticky key modifier is pressed before a previously pressed sticky key is released, the modifiers will get combined so you can add more sticky keys or press a regular key to apply the modifiers. This is to accommodate _callum-style mods_ where you are prone to rolling sticky keys. If you want sticky key modifiers to only chain after release, you can disable this setting.
+This setting is enabled by default. It ensures that if a sticky key modifier is pressed before a previously pressed sticky key is released, the modifiers will get combined so you can add more sticky keys or press a regular key to apply the modifiers. This is to accommodate _callum-style mods_ where you are prone to rolling sticky keys. If you want sticky key modifiers to only chain after release, you can disable this setting. Please note that [modifier functions](https://zmk.dev/docs/codes/modifiers#modifier-functions), such as `&sk LS(LALT)`, require `ignore-modifiers` enabled in order to function properly.
```
caksoylar, line 39:
```suggestion
This setting is enabled by default. It ensures that if a sticky key modifier is pressed before a previously pressed sticky key is released, the modifiers will get combined so you can add more sticky keys or press a regular key to apply the modifiers. This is to accommodate _callum-style mods_ where you are prone to rolling sticky keys. If you want sticky key modifiers to only chain after release, you can disable this setting. Please note that activating multiple modifiers via [modifier functions](https://zmk.dev/docs/codes/modifiers#modifier-functions) such as `&sk LS(LALT)`, require `ignore-modifiers` enabled in order to function properly.
```

This reads a little better to me this way.

caksoylar (approved):

petejohanson (approved):
Thanks!

PR #1739, opened 2023-04-08
fix(shields): Fix nice!view for deep sleep

caksoylar:
Currently building with `nice_view_adapter` shields isn't working if deep sleep is enabled, e.g. the following fails to build: 
```sh
west build -p -d build/corne_left_nv -b nice_nano -- -DSHIELD="corne_left nice_view_adapter nice_view" -DCONFIG_ZMK_SLEEP=y
```

This seems to be because `spi0_sleep` pinctrl nodes aren't defined and adding those seems to fix the issue. Corne-ish Zen definitions [defines and uses them ](https://github.com/zmkfirmware/zmk/blob/main/app/boards/arm/corneish_zen/corneish_zen_v2_left.dts#L65)in a similar way.


petejohanson (approved):
Good catch, thank you!

Issue #1706, opened 2023-03-14
Feature request: Option to make Caps Word behave like QMK


chalop:
On QMK's caps word, when you press the minus key it automatically converts it to an underscore (https://docs.qmk.fm/#/feature_caps_word). I was happy to see this key on ZMK too, but it doesn't behave the same. 

How feasible would be to make the minus key is pressed it "shifts" it to underscore and caps word remains active? I posted on the discord if there was a way to make it but the [post](https://discord.com/channels/719497620560543766/1082001859121782856) didn't gain any traction.

jjangga0214:
@chalop For now, as a workaround, you can duplicate the layer which has all the same settings but the MINUS key is UNDERSCORE. And you replace `&caps_word` with a new macro, like the below example.

```
        new_caps_word: new_caps_word {
            label = "new_caps_word";
            compatible = "zmk,behavior-macro";
            #binding-cells = <0>;
            bindings
                = <&macro_tap &caps_word &tog DUPLICATED_LAYER>;
        };
```

And on the DUPLICATED_LAYER, you activate macro instead of space, esc, and etc.
```
        new_space: new_space {
            label = "new_space";
            compatible = "zmk,behavior-macro";
            #binding-cells = <0>;
            bindings
                = <&macro_tap &kp SPACE &tog DUPLICATED_LAYER>;
        };
```

chalop:
It's not an ideal solution, but it works really great! It also has the added benefit of turning on a status led on the adv 360 (since its technically just another layer)

As for what the ideal solution would look like, I would need to get more familiar with zmk's syntax in order to suggest an api for it. I'll leave the issue open in case someone has a need to implement the qmk behavior on zmk.

Thank you!

jjangga0214:
In my humble opinion, instead of just letting minus and underscore the only special exception, it'd be better to let users configure this. For example, while `&caps_word` I want to input ':' over ';'. 

(For those who wonder why: That's because I use [Colemak](https://colemak.com) layout. For input source language layout reasons, I can't but let zmk layout be QWERTY and os layout be Colemak, not the opposite. So the OS does remapping under the hood. So OS remaps ';' to 'o' and ':' to 'O'. That is why ':', not ';', should be inputted when using `&caps_word`.)

joelspadin:
I'm working on a PR to improve caps word which will add

* `shift-list`: a list of additional key codes to be shifted (defaults to `MINUS`)
[...]

PR #1742, opened 2023-04-08
Allow customization of which keys caps_word modifies

joelspadin:
Added new options to &caps_word to bring it closer to feature parity with QMK:

- Added `shift-list` to include extra keys to have modifiers applied, for example to change `MINUS` to `UNDERSCORE`. Anything specified in this list is implicitly part of `continue-list`.
- Added `no-default-keys`, which removes the implicit alphanumeric keys from `continue-list` and `shift-list` when set. This allows fully customizing the keys, for example for non-US layouts.


Kikangh:
WoW ! Thanks for working on the no-default-keys part, as a foreign (exotic) layout user, this should make it a wee bit easier for me to setup caps_word for my use case 🙏

theol0403:
How does this compare to the features added to https://github.com/zmkfirmware/zmk/pull/1451?

joelspadin:
> How does this compare to the features added to #1451?

From my understanding of that PR, they are almost completely different. That PR does the following:

- Adds a way to activate a layer instead of/in addition to activating modifiers.
- Adds a way to select whether alphas, numbers, and/or modifiers continue/break a word.
- Any custom `zmk,behavior-caps-word` behaviors will change as follows unless they are updated:
  - No modifiers are applied unless `mods` is set.
  - Word break occurs on any key not in `continue-list` (previously alphanumeric and modifier keys would always continue a word)

To restore the original behavior of a custom caps word behavior, add:
```
mods = <MOD_LSFT>;
ignore-alphas;
ignore-numbers;
ignore-modifiers;
```

This PR does:

- Adds a way to select the specific keys that continue/break a word (except modifiers, which still always continue, but that could be changed if necessary).
- Adds a way to select the specific keys that have modifiers applied to them.
- Adds a timeout where caps word will deactivate if no key is pressed for some time.
- `&caps_word` now defaults to shifting '-' to underscore `_` and having a 5 second idle timeout.
- Any custom `zmk,behavior-caps-word` behaviors will change as follows unless they are updated:
  - A 5 second idle timeout is added.

To restore the original behavior of a custom caps word behavior, add:
```
idle-timeout-ms = <0>;
```

theol0403:
Thanks for the detailed explanation! Looking forward to seeing both of these merged.

caksoylar:
Would you consider adding explicit modifier changes (discussed around [here](https://discord.com/channels/719497620560543766/719565084208398406/1092557842575261776)) to this PR, or a separate one?

joelspadin:
I was thinking to do that in a separate change to keep this easier to review.

joelspadin:
Fixes #1706

theol0403:
> Defaults to MINUS so hyphens get turned into underscores.

I would say wanting to type in UPPER-DASH-CASE is common enough that having it shift to underscore automatically might be too opinionated for a default. In my personal config, I have hyphen and underscore on separate keys in a symbol layer (pretty common), so would need to explicitly disable this default because now I have two underscores on my symbol layer during caps word. 

Perhaps I am misunderstanding, but wouldn't having hyphens in the default shift-list also require hyphens to be in the default continue-list (which is only A-Z)?

joelspadin:
Added more tests, updated the documentation, and changed so that anything in `shift-list` will also be treated as a continue key, so you don't need to add keys to both lists for them to work.

joelspadin:
Are you advocating for changing the default so that `-` continues a word but does not get shifted to `_`? Or, do you just mean that `-` should end a word, so to type out `UPPER-DASH-CASE` you'd press caps word again at the start of each word? 

If you do think the default should change, in what situations would that be used? I can't think of any, but I'm also a primarily C/C++ programmer, so I'm much more familiar with use cases for `SCREAMING_SNAKE_CASE` than for `SCREAMING-KEBAB-CASE`.

infused-kim:
> Are you advocating for changing the default so that `-` continues a word but does not get shifted to `_`? Or, do you just mean that `-` should end a word, so to type out `UPPER-DASH-CASE` you'd press caps word again at the start of each word?
> 
> If you do think the default should change, in what situations would that be used? I can't think of any, but I'm also a primarily C/C++ programmer, so I'm much more familiar with use cases for `SCREAMING_SNAKE_CASE` than for `SCREAMING-KEBAB-CASE`.

I see where you are coming from with these defaults. I think I am approaching it from a slightly different point of view.

## 1. Shifting "-" or not by default
I think caps word should behave as closely to "normal" caps lock as possible. Normal caps lock (at least on macOS) doesn't seem to shift any symbols or numbers. It seems to only affect letters.

And I think caps word should behave exactly the same way in the default config. That's the behavior a user would expect and so that's what they should get.

On top of that, only shifting one symbol, but not any of the other symbols by default is extra confusing and unexpected.

So I think it would be best to avoid that.

I do think that your use-case makes perfect sense and many programmers would probably prefer that behavior. But I think it should be an option and not a default.

And even though a portion of users would prefer that behavior, it's not like it is impossible to type a "_" without the default. Users can just type shift+- and that's what most users would expect to have to do anyways (even with normal caps lock).

## 2. Should "-" deactivate caps word or not by default
This one is a bit more difficult to answer, because we don't have any pre-existing behavior that users might be used to.

I think I would lean towards keeping capsword enabled on both "-" and "_". Simply, because I can't think of many instances where a user wouldn't type another deactivating character like ` `, `.`, or `;` after the underscore or dash.

infused-kim:
I just thought of two more reason against shifting "-“ by default. 

1. 
Doing so also makes it impossible to actually type “-“ during caps word should the user desire to do so. 

For many users that’s fine, but it should be a conscious decision and not unexpected behavior. 

2.
The beauty of caps word is that it makes your life easier and then seamlessly gets out of the way.

But users won’t use it just to type programming definitions like “FOO-BAR” or “FOO_BAR”. 

Sometimes people may type a sentence like “[…] and that’s why this is BAD-- I won’t allow this.”

The expectation here is that “I can just type and don’t need to worry about turning off caps word or weird exceptions”.

theol0403:
I agree with @infused-kim. Personally, the biggest issue for me with having hyphen shifted by default is the inability to type "-" during caps word, which doesn't feel like a natural default. This is extra relevant to me because I already have a dedicated "_" key on a layer and don't use shift+"-". 

Another very minor point is that looking through the source code and docs, the need to make several references to "MINUS" to explain its behaviour reflects its unintuitive nature.

While it is very easy to remove the default behaviour, it is equally easy to add behaviour, and I think more users would expect the default to be unopinionated. I think it would make more sense to have "For example, to add `MINUS` to the shift list so it is automatically changed to `UNDERSCORE`, add:" in the docs (as a nice way to guide the reader towards this possibly desired feature). 

Finally, I have fewer opinions on whether MINUS should be in continue-list by default, but I vote to keep it.

joelspadin:
The latest version removes `MINUS` so the default behavior is unchanged.

joelspadin:
Rebased and fixed whitespace issues for pre-commit.

app/include/zmk/keys.h
```diff
@@ -20,6 +22,21 @@ struct zmk_key_event {
     bool pressed;
 };
 
+struct zmk_key_decoded {
```
petejohanson, line 25: Not sure I love the decoded terminology... Complete key? Key details? Full key?

app/src/behaviors/behavior_caps_word.c
```diff
@@ -24,33 +24,66 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
 
+static const struct device *devs[DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT)];
+static size_t dev_count = 0;
+static int pressed_key_count = 0;
+
+static void restart_caps_word_idle_timer(const struct device *dev) {
+    const struct behavior_caps_word_config *config = dev->config;
+    struct behavior_caps_word_data *data = dev->data;
+
+    if (config->idle_timeout_ms) {
```
petejohanson, line 53: Any reason not to use a delayable work for this, instead of a timer?

petejohanson (changes requested):
Couple questions on the implementation. Thanks!

app/include/zmk/keys.h
joelspadin: Since this is all the data from a DT parameter, how about zmk_key_param?

app/src/behaviors/behavior_caps_word.c
joelspadin: Nope. Fixed.

SebastianBehrens:
Hi @petejohanson, are you hesistant to merging this; or can one expect this to clear soon?

ctranstrum:
+1 on merging this one

joelspadin:
Rebased, cleaned up and simplified the code some more, and dropped the idle timeout feature for now. Macros can cause keys to be pressed a different number of times than they are released, so counting the number of presses and releases to determine when no keys are pressed isn't reliable. Doing that correctly will require some changes to core code, so I'll handle that separately.

pltanton:
Is there any chance this can be merged?

BlaiseRideout:
@joelspadin you mind rebasing and resolving merge conflicts?

drewlwhitney:
I would also love to see this merged, I'm currently stuck using #2731 to make `caps-word` useable

Strix-Vyxlor:
any progress on this? I'm currently stuck being unable to have my m key capitalized when using caps_word.

tuxflo:
I would also like to see this merged. Is there something we (as the community) could help to get this into main?

PR #1745, opened 2023-04-08
fix(behaviors): Fix use after free in sticky key

joelspadin:
Fixed an issue where the sticky key behavior would call ZMK_EVENT_RAISE_AFTER(), which would free the provided event, but then it would keep using that now-freed event data.


joelspadin:
I just discovered #1586, which this is nearly identical to. This PR marks the copy as `const` and adds a comment explaining why it's needed, while the other one adds extra tests. Maybe we should take the behavior_sticky_key.c from this PR and the tests from the other?

petejohanson (approved):
Good catch! Thanks.

PR #1746, opened 2023-04-09
fix(underglow): Move to spi3 for underglow bus.

petejohanson:
* Workaround Zephyr bug for Nordic SPI(M) driver after the
  pinctrl refactor by using spi3 peripheral for the SPI bus for
  the WS2812 led_strip driver.

### DIFF
--- a/app/boards/shields/chalice/boards/nice_nano.overlay
+++ b/app/boards/shields/chalice/boards/nice_nano.overlay
@@ -1,13 +1,13 @@
 #include <dt-bindings/led/led.h>
 
 &pinctrl {
-	spi1_default: spi1_default {
+	spi3_default: spi3_default {
 		group1 {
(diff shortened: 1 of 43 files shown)


app/boards/shields/softie/Kconfig.defconfig
caksoylar, line 1: `git add` mishap?

petejohanson: Indeed. Good catch. Removed.

petejohanson:
A bit more research, if you check out https://infocenter.nordicsemi.com/topic/ps_nrf52840/spim.html?cp=5_0_0_5_24#topic__table.instances you see that SPI 1-3 all have a lot of limitations not listed for SPI3; I suspect one of those limitations is causing problems when configured via pinctrl somehow.

PR #1747, opened 2023-04-10
fix(display): Do not override font size in status screen

caksoylar:
This commit fixes the normal font styling in stock status screen so that the selected font size is used. It has resolved the widget sizing issues in my testing with [the native posix setup](https://github.com/caksoylar/zmk-posix-testbed/tree/nice-view).

### DIFF
--- a/app/src/display/status_screen.c
+++ b/app/src/display/status_screen.c
@@ -34,16 +34,9 @@ static struct zmk_widget_layer_status layer_status_widget;
 static struct zmk_widget_wpm_status wpm_status_widget;
 #endif
 
-lv_style_t global_style;
-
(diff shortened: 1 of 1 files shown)


petejohanson:
@caksoylar Thanks for the fixes! Can we split out the two items into separate PRs though, for clarify/future reference? TIA.

caksoylar:
Sure thing, did that now.

petejohanson (approved):
Thanks!

caksoylar:
I'll drop the other changes for now. I believe it comes up only for `native_posix_64` which must be due to the different sizing of integers and I am guessing the better solution is to use a more appropriate format string specifier in `sprintf` so it works in both architectures.

PR #1753, opened 2023-04-14
fix(boards): Fix Corne-ish Zen logo alignment

caksoylar:
The Corne-ish Zen logo on the peripheral was slightly misaligned, which I am guessing causes the black bar at the bottom to appear. Moving it slightly so it doesn't go off-screen fixes this issue, see pictures below:

Before:
image

After:
image


petejohanson (approved):
Thanks!

PR #1754, opened 2023-04-14
feat(display): Set mono theme and add invert config

caksoylar:
# Update

This PR sets `CONFIG_LV_USE_THEME_MONO` by default when `ZMK_DISPLAY` is enabled, so that small font selection used in the built-in status screen widgets is no longer ineffective.

It also adds a new Kconfig `ZMK_DISPLAY_INVERT` to invert the colors used in the display from black-on-white to white-on-black, by setting the appropriate value in mono theme initialization.

# Original
LVGL mono theme was getting selected by `ZMK_DISPLAY` but not actually used outside of the `corneish_zen` configuration. This change sets it by default if OLEDs or ls0xx memory displays like nice!view are used. This change could use some testing to make sure it doesn't break on OLEDs, nice!views, Adafruit memory displays etc. (Unfortunately I don't have any at the moment, tested on native_posix+SDL setup.)

Given that the theme is set correctly, we can also add an invert option by setting the `dark_bg` parameter of `lv_theme_mono_init` accordingly. The config is currently named `CONFIG_ZMK_DISPLAY_INVERT` but maybe it should be `CONFIG_ZMK_MONO_THEME_INVERT` or something else? Also it is possible that there is a better way to make this change.

### DIFF
--- a/app/src/display/Kconfig
+++ b/app/src/display/Kconfig
@@ -9,6 +9,7 @@ menuconfig ZMK_DISPLAY
     select LV_THEMES
     select LV_THEME_MONO
     select LV_CONF_MINIMAL
+    imply LV_USE_THEME_MONO
 
 if ZMK_DISPLAY
 
--- a/app/src/display/Kconfig
+++ b/app/src/display/Kconfig
@@ -15,6 +15,13 @@ config ZMK_DISPLAY_BLANK_ON_IDLE
     bool "Blank display on idle"
     default y if SSD1306
 
+if LV_USE_THEME_MONO
+
+config ZMK_DISPLAY_INVERT
+    bool "Invert display colors"
+
+endif
+
 choice LV_TXT_ENC
     default LV_TXT_ENC_UTF8
 
--- a/app/src/display/Kconfig
+++ b/app/src/display/Kconfig
@@ -6,8 +6,6 @@ menuconfig ZMK_DISPLAY
     default n
     select DISPLAY
     select LVGL
-    select LV_THEMES
-    select LV_THEME_MONO
     select LV_CONF_MINIMAL
     imply LV_USE_THEME_MONO
 
--- a/app/src/display/main.c
+++ b/app/src/display/main.c
@@ -91,7 +91,8 @@ int zmk_display_is_initialized() { return initialized; }
 static void initialize_theme() {
 #if IS_ENABLED(CONFIG_LV_USE_THEME_MONO)
     lv_disp_t *disp = lv_disp_get_default();
-    lv_theme_t *theme = lv_theme_mono_init(disp, false, CONFIG_LV_FONT_DEFAULT);
+    lv_theme_t *theme =
+        lv_theme_mono_init(disp, IS_ENABLED(CONFIG_ZMK_DISPLAY_INVERT), CONFIG_LV_FONT_DEFAULT);
     theme->font_small = CONFIG_ZMK_LV_FONT_DEFAULT_SMALL;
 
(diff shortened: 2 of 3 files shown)


caksoylar:
One question: Should we default to `LV_USE_THEME_MONO` unconditionally? E.g. it can be added as `select LV_USE_THEME_MONO` under `config ZMK_DISPLAY`.

Right now it seems to be necessary for setting the small font via `CONFIG_ZMK_LV_FONT_DEFAULT_SMALL`, since it is set the  `ifdef` block in `main.c` line 97. I think there might be other places where the monochrome display is assumed inherently.

sfcgeorge (approved):
Reviewing just to say I've been using this for a week and it's working as expected. Is there anything left to do to take this off Draft and get wider review?

caksoylar:
On my end, the question on the previous comment is the main thing I want to clear up to figure out the best approach.

joelspadin (changes requested):
Working for me.

> The config is currently named CONFIG_ZMK_DISPLAY_INVERT but maybe it should be CONFIG_ZMK_MONO_THEME_INVERT or something else?

I think what you have is fine unless mono is the only theme that will ever support being inverted.

> Should we default to LV_USE_THEME_MONO unconditionally?

This seems like it would be an improvement to me, so long as it remains possible to override it to another theme. All of the most common displays to use with keyboards are monochrome.

app/src/display/Kconfig
```diff
@@ -6,16 +6,22 @@ menuconfig ZMK_DISPLAY
     default n
     select DISPLAY
     select LVGL
-    select LV_THEMES
-    select LV_THEME_MONO
```
caksoylar, line 10: I haven't been able to find where these two settings are defined in Zephyr/LVGL or through `menuconfig`; everything seems to work fine for me without them so I added a commit removing these. If anyone thinks this is necessary I can drop that specific commit from the PR.

app/src/display/main.c
```diff
@@ -91,7 +91,8 @@ int zmk_display_is_initialized() { return initialized; }
 static void initialize_theme() {
 #if IS_ENABLED(CONFIG_LV_USE_THEME_MONO)
     lv_disp_t *disp = lv_disp_get_default();
-    lv_theme_t *theme = lv_theme_mono_init(disp, false, CONFIG_LV_FONT_DEFAULT);
+    const bool invert = IS_ENABLED(CONFIG_ZMK_DISPLAY_INVERT);
+    lv_theme_t *theme = lv_theme_mono_init(disp, invert, CONFIG_LV_FONT_DEFAULT);
```
petejohanson, line 95: Do we really need to pull this out to a local?

```suggestion
    lv_theme_t *theme = lv_theme_mono_init(disp, IS_ENABLED(CONFIG_ZMK_DISPLAY_INVERT), CONFIG_LV_FONT_DEFAULT);
```

app/src/display/Kconfig
petejohanson: I can't either... Odd it works. Fine to remove, IMHO.

petejohanson (changes requested):
Just one minor code style comment. Otherwise LGTM. Thanks!

app/src/display/main.c
caksoylar: Yeah, that works just as well, committed the change.

caksoylar:
Looks like some conditional layers tests are failing but I couldn't reproduce locally 😕

petejohanson (approved):
Thanks!

PR #1755, opened 2023-04-16
fix doc typo for development/boards-shields-keymaps

sporkus:
Current: A `keymap/keymap.overlay` file that includes the default keymap for that keyboard. Users will be able to override this keymap in their user configs. 

Fix:
A `${board_name}.keymap` file that includes the default keymap for that keyboard. Users will be able to override this keymap in their user configs.


caksoylar (approved):
Thanks!

petejohanson (approved):
Thanks!

PR #1757, opened 2023-04-16
Add a way to jump to bootloader and/or reset settings on startup

joelspadin:
This adds an optional feature to reboot into the bootloader if a specific key is held when the keyboard is powered on.

This is similar to QMK's [bootmagic lite](https://docs.qmk.fm/#/feature_bootmagic) feature, and it is primarily intended as a way to recover a keyboard which doesn't have a physical reset button in case it is flashed with firmware that doesn't have a &bootloader key in its keymap.


joelspadin:
This theoretically works on the peripheral (right) side of a split, but I don't own a split keyboard, so if someone wants to test it, that would be appreciated!

app/src/startup_bootloader_key.c
```diff
@@ -0,0 +1,54 @@
...
+SYS_INIT(timeout_init, APPLICATION, CONFIG_APPLICATION_INIT_PRIORITY);
+
+static int event_listener(const zmk_event_t *eh) {
+    const int64_t uptime = k_uptime_get();
+    if (uptime > timeout_uptime) {
+        return ZMK_EV_EVENT_BUBBLE;
+    }
+
+    const struct zmk_position_state_changed *ev = as_zmk_position_state_changed(eh);
+    if (ev && ev->state) {
```
caksoylar, line 42: I could be reading this wrong, but this is looking for a key press rather than release event, right? If so, does it work when the keyboard is powered up/reset while the key is already in the down position?

caksoylar: It works, so I guess the initial scan/interrupt raises a key down event

caksoylar:
I tested it on a split keyboard with https://github.com/caksoylar/zmk/commit/5bc195be9d7a503cab5d77617c6c30603191b088 and it works as expected. I'll add a suggestion to docs emphasizing that the key positions are affected by matrix transforms.

docs/docs/features/startup-bootloader-key.md
```diff
@@ -0,0 +1,94 @@
...
+        compatible = "zmk,startup-bootloader-key";
+        key-position = <0>;
+    };
+    ...
+};
+```
+
+:::info
+
+Key positions are numbered like the keys in your keymap, starting at 0. So, if the first key in your keymap is `Q`, this key is in position `0`. The next key (possibly `W`) will have position 1, etcetera.
```
caksoylar, line 31:
```suggestion
Key positions are numbered like the keys in your keymap, starting at 0. So, if the first key in your keymap is `Q`, this key is in position `0`. The next key (possibly `W`) will have position 1, etcetera. Note that the key index is affected by the chosen [matrix transform](../config/kscan#matrix-transform), so make sure to select an index that would be available in all of them if you define multiple matrix transforms.
```

Might be good to add a note like this one?

joelspadin: This was a direct copy from the combos documentation. Do you think this should be added there too?

joelspadin:
As discussed starting at https://discord.com/channels/719497620560543766/719565084208398406/1097298953650847824, I'm going to tweak this to be a "boot magic key" which performs one or more actions (jump to bootloader, reset settings, etc.) instead of only supporting jump to bootloader.

docs/docs/features/startup-bootloader-key.md
caksoylar: That's a good point, the first part applies there as well.

For the second part, combos are different because they are defined by the user in the keymap, where the transform is known. Here this is directed towards users writing keyboard definitions, where they don't know what matrix transform will be chosen by the user.

joelspadin:
Added options to jump to bootloader and/or reset settings (currently just runs the same function to clear BLE bonds as the settings_reset firmware does, but we can improve that later). 

Renamed the feature to "boot magic key" since it is now, just like QMK's feature, a way to perform some magic function if a key is held at boot.

app/src/ble.c
```diff
@@ -38,6 +38,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #if IS_ENABLED(CONFIG_ZMK_BLE_PASSKEY_ENTRY)
 #include <zmk/events/keycode_state_changed.h>
+#include "ble.h"
```
caksoylar, line 41: Is this included by mistake?

caksoylar:
FYI this now has some minor merge conflicts, I rebased it (and seems to be functional) at https://github.com/caksoylar/zmk/compare/f72433e2cdf51df1810e11570869834bacd3001f...caksoylar:zmk:caksoylar/experimental.

petejohanson (changes requested):
I truly love this. A couple minor notes on implementation. Would love to get this rebased and in.

docs/docs/features/boot-magic-key.md
```diff
@@ -0,0 +1,209 @@
...
+
+**shield_right.overlay**
+
+```c
+#include "shield.dtsi"
+
+&bootloader_key_right {
+    status = "okay";
+};
+```
```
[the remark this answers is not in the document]
caksoylar: I should certainly try it again, but that didn't work the same for me the first time I tried it: https://discord.com/channels/719497620560543766/719544230497878116/1097338556655075499

caksoylar:
For posterity, I ran this for a few months, until I recently switched to a Zephyr 3.5 based branch. I haven't tried the settings reset functionality but the bootloader has been working well ([rebased commits here](https://github.com/caksoylar/zmk/compare/ebda7dd0cdbc92b3d1abbc534829a429825700a0...df5bd406ca7b0340d00680b4fd7069d1738f6ee8)).

joelspadin:
I plan to first finish a PR that makes settings reset actually reset all settings, which will add a function to call for that, and then I'll update this PR to use that.

rianadon:
I'm interested in using this for my keyboards, so I [rebased onto the latest commits](https://github.com/rianadon/zmk/tree/bootloader-key). It would be amazing to have this merged even if the settings reset functionality is cut out.

docs/docs/features/boot-magic-key.md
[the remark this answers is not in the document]
rianadon: It did work for me when I defined them locally in the left and right overlays.

PR #1758, opened 2023-04-17
feat(behaviors): Add reusable sensor behaviors.

petejohanson:
* Add new sensor behaviors that either take full bindings add definition, or accept parameters when bound in the keymap.
* Remove existing hard-coded key press sensor behavior and instead leverage new generic sensor behaviors to achieve the same functionality.

I've taken the amazing work from @nickconway in #1275 and done the following:

* Rebased onto latest `main` with Zephyr 3.2 fixes.
* Tweaked the behavior positions used thanks to the work from @joelspadin in #1659 

~~Still on my list is the docs updates I'd noted~~, but this should be nearly there, and is ready for review.


app/src/keymap.c
```diff
@@ -270,8 +270,9 @@ int zmk_keymap_sensor_triggered(uint8_t sensor_number, const struct device *sens
                 continue;
             }
 
-            const uint32_t position = ZMK_VIRTUAL_KEY_POSITION_SENSOR(sensor_number);
-            ret = behavior_sensor_keymap_binding_triggered(binding, sensor, position, timestamp);
+            struct zmk_behavior_binding_event event = {
+                .position = ZMK_VIRTUAL_KEY_POSITION_SENSOR(sensor_number), .timestamp = timestamp};
+            ret = behavior_sensor_keymap_binding_triggered(binding, sensor, event);
```
petejohanson, line 275: @joelspadin  Thoughts on if we should pass behaviors the actual sensor position, then let them generate the virtual key position if they need to queue behaviors?

app/Kconfig
```diff
@@ -522,6 +522,18 @@ choice CBPRINTF_IMPLEMENTATION
 
 endchoice
 
+DT_COMPAT_ZMK_BEHAVIOR_SENSOR_ROTATE := zmk,behavior-sensor-rotate
```
joelspadin, line 525: This is already a pretty large file. Since we're likely to do this for all of the behaviors at some point, any thoughts on adding a src/behaviors/Kconfig (and maybe doing the same for cmake stuff)?

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,88 @@
...
+        triggered_binding = (struct zmk_behavior_binding *)&cfg->cw_binding;
+        break;
+    case -1:
+        triggered_binding = (struct zmk_behavior_binding *)&cfg->ccw_binding;
+        break;
+    default:
+        return -ENOTSUP;
+    }
+
+    LOG_DBG("Sensor binding: %s", log_strdup(binding->behavior_dev));
```
joelspadin, line 56: I don't think log_strdup is needed anymore?

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,88 @@
...
+static int on_sensor_binding_triggered(struct zmk_behavior_binding *binding,
+                                       const struct device *sensor,
+                                       struct zmk_behavior_binding_event event) {
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_sensor_rotate_config *cfg = dev->config;
+
+    struct sensor_value value;
+    int err;
+
+    err = sensor_channel_get(sensor, SENSOR_CHAN_ROTATION, &value);
```
joelspadin, line 37: Minor: 

```suggestion
    int err = sensor_channel_get(sensor, SENSOR_CHAN_ROTATION, &value);
```

or

```suggestion
    const int err = sensor_channel_get(sensor, SENSOR_CHAN_ROTATION, &value);
```
since it isn't ever changed.

app/src/behaviors/behavior_sensor_rotate_var.c
```diff
@@ -0,0 +1,81 @@
...
+static int on_sensor_binding_triggered(struct zmk_behavior_binding *binding,
+                                       const struct device *sensor,
+                                       struct zmk_behavior_binding_event event) {
+    const struct device *dev = device_get_binding(binding->behavior_dev);
+    const struct behavior_sensor_rotate_var_config *cfg = dev->config;
+
+    struct sensor_value value;
+    int err;
+
+    err = sensor_channel_get(sensor, SENSOR_CHAN_ROTATION, &value);
```
joelspadin, line 37: Same as the previous file.

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,88 @@
...
+
+    if (err < 0) {
+        LOG_WRN("Failed to get sensor rotation value: %d", err);
+        return err;
+    }
+
+    struct zmk_behavior_binding *triggered_binding;
```
joelspadin, line 44: Make this a `const struct zmk_behavior_binding*` and then you shouldn't need to cast anything on lines 47 and 50.

app/src/behaviors/behavior_sensor_rotate.c
```diff
@@ -0,0 +1,88 @@
...
+    switch (value.val1) {
+    case 1:
+        triggered_binding = (struct zmk_behavior_binding *)&cfg->cw_binding;
```
joelspadin, line 47: Casting away const is dangerous.

app/src/behaviors/behavior_sensor_rotate_var.c
```diff
@@ -0,0 +1,81 @@
...
+#include <zmk/event_manager.h>
+#include <zmk/events/keycode_state_changed.h>
+#include <zmk/behavior_queue.h>
+#include <zmk/keymap.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct behavior_sensor_rotate_var_config {
+    char *cw_behavior_dev;
+    char *ccw_behavior_dev;
```
joelspadin, line 24: You might be able to share a lot of code between this and the other version if you use the previous file's config struct for both version, but just initialize the params to 0 in this version. Then inside the "binding triggered" function, everything before and after the `switch` block could be identical.

joelspadin (changes requested):

app/src/keymap.c
joelspadin: As mentioned on Discord, my main concern with that is that it would be very easy to just pass the sensor position along when queueing behaviors, which could cause very subtle bugs.

app/src/behaviors/behavior_sensor_rotate.c
petejohanson: Good catch. Removed.

app/src/behaviors/behavior_sensor_rotate_var.c
petejohanson: Yeah, so moved the whole callback to a "common" code, and added an `bool override_params` field to the config struct to determine if the parameters get updated or not.

app/Kconfig
petejohanson: Moved to `app/Kconfig.behaviors` for now. I think in the future, we'll:

* Move to `app/src/behaviors/Kconfig`
* Add `app/src/behaviors/CMakeLists.txt`
* Tweak the current conditions for split/central state for the behavior compilation in `app/CMakeLists.txt` into individual Kconfig symbols for each behavior that use Kconfig logic for conditionally enabling them or not.

Don't really want to cram all of that into this PR though.

Thoughts?

joelspadin: Seems reasonable to me.

docs/docs/behaviors/sensor-rotate.md
```diff
@@ -0,0 +1,78 @@
...
+
+    keymap {
+        compatible = "zmk,keymap";
+
+        base {
+            ...
+            sensor-bindings = <&rot_kp PG_DN PG_UP>;
+        }
+    }
+};
```
joelspadin, line 77: Inconsistent indentation (mixed tabs and spaces)

app/src/behaviors/behavior_sensor_rotate_common.c
```diff
@@ -0,0 +1,53 @@
...
+        LOG_WRN("Failed to get sensor rotation value: %d", err);
+        return err;
+    }
+
+    struct zmk_behavior_binding triggered_binding;
+    switch (value.val1) {
+    case 1:
+        triggered_binding = cfg->cw_binding;
+        if (cfg->override_params) {
+
```
joelspadin, line 33: Super minor: extra blank line here

docs/docs/behaviors/sensor-rotate.md
petejohanson: Indeed. Fixed.

app/src/behaviors/behavior_sensor_rotate_common.c
petejohanson: Good catch. Fixed.

joelspadin (approved):

PR #1764, opened 2023-04-20
docs: add llcp setting to troubleshooting guide

npx:
I added a section to the troubleshooting guide regarding the frequent disconnects caused by the new Zephyr default llcp.
I think it is worthwhile mentioning it there.

I am in no position to phrase the explanation correctly, so any help is appreciated. 

I just prepared this PR to make this quick and easy :)


petejohanson:
Closing for now, since we've defaulted the legacy LLCP project wide already.

PR #1767, opened 2023-04-21
fix(docs): Update references to QMK settings

filterpaper:
* `TAPPING_FORCE_HOLD` was replaced with an equivalent QUICK_TAP_TERM.
* `IGNORE_MOD_TAP_INTERRUPT` will be a default setting.

### DIFF
--- a/docs/docs/behaviors/hold-tap.md
+++ b/docs/docs/behaviors/hold-tap.md
@@ -317,4 +317,4 @@ This hold-tap example implements a [momentary-layer](layers.md/#momentary-layer)
 
 ### Comparison to QMK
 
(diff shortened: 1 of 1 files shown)


filterpaper:
The `HOLD_ON_OTHER_KEY_PRESS` reference is correct. `IGNORE_MOD_TAP_INTERRUPT` will be a default behavior. It is removed in develop branch and will be merged in the next cycle. I guess it will make sense to remove it now as well.

caksoylar:
Thank you! Looks like next merge will be end of May, so reasonably soon that we could omit mentioning `IGNORE_MOD_TAP_INTERRUPT`. Anyone have any objections?

caksoylar (approved):

PR #1768, opened 2023-04-21
Add custom nice!view widgets

Nicell:
Scrreen example

Looking for feedback on the structure of this. It's a bit annoying since a single canvas can take on multiple events (top one up to three on the left side), so this is an option. Maybe isn't the best.

Also working on making a better image for the right side than just a nice!view logo.


app/boards/shields/nice_view/Kconfig.defconfig
```diff
@@ -26,4 +25,24 @@ endchoice
 
 endif # ZMK_DISPLAY
 
+config CUSTOM_WIDGET_STATUS
```
petejohanson, line 28:
```suggestion
config NICE_VIEW_WIDGET_STATUS
```

app/boards/shields/nice_view/Kconfig.defconfig
```diff
@@ -26,4 +25,24 @@ endchoice
...
+    bool "Custom status widget"
+    default y if ZMK_DISPLAY_STATUS_SCREEN_CUSTOM
```
petejohanson, line 30: I think maybe instead enhance the existing custom screen config to `select` this symbol?

app/boards/shields/nice_view/Kconfig.defconfig
```diff
@@ -26,4 +25,24 @@ endchoice
...
+
+config CUSTOM_WIDGET_STATUS
+    select LV_FONT_MONTSERRAT_18
+    select LV_FONT_MONTSERRAT_14
+    select LV_FONT_UNSCII_8
+    select ZMK_WPM
+
+endif # !ZMK_SPLIT || ZMK_SPLIT_ROLE_CENTRAL
+
+config CUSTOM_WIDGET_INVERTED
```
petejohanson, line 45: Colors or orientation?

petejohanson (commented):
A few initial thoughts

mike1808:
I tried running the widgets and they work as expected. However, the widgets add a significant delay to the main thread and every 5-6th keypress gets ignored. 
Adding `CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED` helped and with dedicated queue I'm not noticing any delay.

davenull:
I have been testing on nicell's fork the last week, and the widgets all work great, and I have not noticed any issues around delay at all.

mike1808:
Hey. I rebased this branch with the main and now the display looks a bit stange. 
PXL_20230531_163000523

mike1808:
@roblim thanks. I figured that out, just giving heads up that the current version doesn't work properly with upstream.

roblim:
I've been testing this PR rebased on `main` over the past few days on a Kyria V3+nice!nano, but am having to revert because of inconsistent behavior of the peripheral side. Things seemed great for the first day, but eventually, the peripheral side started lagging and dropping keypresses, getting progressively worse until unrecoverable. At first, I could manage to type what I wanted if I slowed down and was deliberate, but eventually, lag would be like 5+ seconds. Resetting or re-flashing didn't help. I had the following added in config:

```
CONFIG_ZMK_DISPLAY_WORK_QUEUE_DEDICATED=y
CONFIG_LV_USE_THEME_MONO=n
```

I tried doubling the thread stack size to see if things would change, and things seemed to get better, but eventually performance would degrade. Here's my config if it helps: https://github.com/roblim/zmk-config-kyria/blob/29a4b30386374d487b099fb11ec25168dee4fba1/config/kyria_rev3.conf.

app/boards/shields/nice_view/README.md
```diff
@@ -1,5 +1,15 @@
...
+This shield requires that an `&nice_view_spi` labeled SPI bus is provided with _at least_ MOSI, SCK, and CS pins defined.
+
+## Disable custom widget
+
+The nice!view shield includes a custom vertical widget. To use the built-in ZMK one, add the following items to your `.conf` file:
+
+```
+CONFIG_ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN=y
+CONFIG_ZMK_LV_FONT_DEFAULT_SMALL_MONTSERRAT_26=y
+CONFIG_LV_FONT_DEFAULT_MONTSERRAT_26=y
```
caksoylar, line 14: Can these two settings be selected for the built-in screen in Kconfig instead?

Nicell: I tried adding them as a default choice under the built-in screen, but that didn't work. You can't select choices directly, so not sure how to "force" a choice.

caksoylar: I did some experimenting and this compiles since we use named `choice`s:
```diff
diff --git a/app/boards/shields/nice_view/Kconfig.defconfig b/app/boards/shields/nice_view/Kconfig.defconfig
index 2656a2cd..323eafe6 100644
--- a/app/boards/shields/nice_view/Kconfig.defconfig
+++ b/app/boards/shields/nice_view/Kconfig.defconfig
@@ -53,7 +53,21 @@ config NICE_VIEW_WIDGET_INVERTED
 
 endif # ZMK_DISPLAY_STATUS_SCREEN_CUSTOM
 
-config ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
-    select LV_FONT_MONTSERRAT_26
+if ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
+
+config LV_FONT_MONTSERRAT_26
+    default y
+
+choice ZMK_LV_FONT_DEFAULT_SMALL
+    default ZMK_LV_FONT_DEFAULT_SMALL_MONTSERRAT_26
+
+endchoice
+
+choice LV_FONT_DEFAULT
+    default LV_FONT_DEFAULT_MONTSERRAT_26
+
+endchoice
+
+endif
 
 endif # SHIELD_NICE_VIEW
```
But it still doesn't set the right choice, so I guess [this comment](https://github.com/zephyrproject-rtos/zephyr/issues/6948#issuecomment-379210116) still applies:
>  I just looked into defaults on choices btw, and the C tools in Zephyr haven't been patched to prefer later defaults there either, so it's a bit inconsistent: Later defaults override earlier defaults for symbols, but earlier defaults override later defaults for choices.
>
> That means that the "base" choice definition (where the symbols are defined) can't have any defaults (with satisfied conditions). Otherwise, those will take precedence over the defaults in the Kconfig.* files, which are included later.

Nicell: Yep, I tried exactly this with the same results. Agreed that right now this isn't possible.

app/boards/shields/nice_view/Kconfig.defconfig
petejohanson: So, you could instead do something like:

```
config ZMK_DISPLAY_STATUS_SCREEN_CUSTOM
  imply CUSTOM_WIDGET_STATUS
  imply CUSTOM_WIDGET_BLAH
```

Doing so aligns the "hierarchy" of the Kconfig matches the code. The custom screen "consumes" or "uses" the widgets, so I think it works better to have the Kconfig for that custom screen implying/enabling the widgets it uses.

app/boards/shields/nice_view/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+if(CONFIG_ZMK_DISPLAY AND CONFIG_NICE_VIEW_WIDGET_STATUS)
+  zephyr_library_include_directories(${CMAKE_SOURCE_DIR}/include)
+  zephyr_library_sources(custom_status_screen.c)
+  target_sources(app PRIVATE widgets/bolt.c)
```
petejohanson, line 4: Why not use `zephyr_library_sources` consistently here?

app/src/display/widgets/Kconfig
```diff
@@ -3,6 +3,8 @@
 
 menu "ZMK Display Widgets"
 
+if ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
+
```
petejohanson, line 7: Instead of this, I think we should change the defaults or each widget so they're not defaulted on, and instead `imply` them all or the build in screen choice. That feels aligned with the same idea of "higher level aggregating in code *and* config" like for your custom screen.

petejohanson (changes requested):
A few others items from a closer review.

app/boards/shields/nice_view/CMakeLists.txt
Nicell: Not sure why, fixed.

app/src/display/widgets/Kconfig
Nicell: I removed the defaults and implied then all ***on*** the built-in screen (that or was confusing me 😅). It seems the `if` of the `default y if...` was unnecessary given the `depend on` lines were already there. I can build with or without the built-in screen option and they work properly on the left and right sides of my corne!

app/src/display/Kconfig
```diff
@@ -42,6 +41,11 @@ choice ZMK_DISPLAY_STATUS_SCREEN
 config ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
     bool "Built in status screen"
     select LV_OBJ_LABEL
+    imply LV_USE_THEME_MONO
+    imply ZMK_WIDGET_LAYER_STATUS
+    imply ZMK_WIDGET_BATTERY_STATUS
+    imply ZMK_WIDGET_OUTPUT_STATUS
+    imply ZMK_WIDGET_PERIPHERAL_STATUS
```
caksoylar, line 48: Is the conditional logic we had before (BT enabled, central vs, peripheral) not necessary?

Nicell: As far as I can tell, the `depends on` was all we needed. Things seem to function equivalently before and after making this change. Tested both left and right with built-in and custom widgets.

caksoylar: Oh right, we still have the `depends on` with the same conditions, I somehow missed that 😅

petejohanson (approved):
Thanks!

PR #1769, opened 2023-04-21
fix(bluetooth): Revert to Legacy LLCP

petejohanson:
* Reports of constant/frequent disconnects, with HCI err 0x28, "instant passed", seem linked to newer LLCP that became default in Zephyr 3.2, so revert to the Legacy LLCP for now until a proper fix for new LLCP can be found.

This needs a decent amount of testing, since it switches a core BT building block that hasn't been as well tested on the Zephyr 3.2 work.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -120,6 +120,11 @@ menuconfig ZMK_BLE
 
 if ZMK_BLE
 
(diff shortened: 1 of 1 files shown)


petejohanson:
Several folks reporting this fixes their stability issues. Merging.

PR #1773, opened 2023-04-23
fix(docs): switch sensor rotate parameter order

nickcoutsos:
This PR brings the new _Sensor Rotation_ behavior documentation in line with the parameter ordering used in the behavior implementation and the existing encoder documentation.


petejohanson (approved):
Thanks!

PR #1780, opened 2023-04-25
fix(docs): Add key to interconnect tabs

joelspadin:
Added a `key` property to interconnect tabs list items to fix a React warning.


petejohanson (approved):
Thanks!

PR #1783, opened 2023-04-26
feat(boards): Add Chocofi

tommi-l:
Add Chocofi with very basic Colemak-dh default keymap. Tested but not sure how it will behave with a display

## Board/Shield Check-list

  - Note: tested without display
  - Available at [beekeeb.com](https://shop.beekeeb.com/product/presoldered-chocofi-split-keyboard/) ,  [kriscables.com](https://kriscables.com/chocofi/) and others

### DIFF
--- a/app/boards/shields/chocofi/Kconfig.defconfig
+++ b/app/boards/shields/chocofi/Kconfig.defconfig
@@ -1,0 +1,46 @@
+if SHIELD_CHOCOFI_LEFT
+
+config ZMK_KEYBOARD_NAME
+    default "Chocofi"
+
+config ZMK_SPLIT_ROLE_CENTRAL
+    default y
+
+endif
+
+if SHIELD_CHOCOFI_LEFT || SHIELD_CHOCOFI_RIGHT
+
+config ZMK_SPLIT
+    default y
+
+if ZMK_DISPLAY
+
+config I2C
+    default y
+
+config SSD1306
+    default y
+
+config SSD1306_REVERSE_MODE
+    default y
+
+endif # ZMK_DISPLAY
+
+if LVGL
+
+config LV_Z_VDB_SIZE
+    default 64
+
+config LV_Z_DPI
+    default 148
+
+config LV_Z_BITS_PER_PIXEL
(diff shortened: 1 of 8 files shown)


caksoylar:
Hi, I believe Chocofi is pin-compatible with the Corne, isn't it? In this case we might not want to duplicate the config.

For reference, you can set the 5 column matrix transform for the `corne` shield to be able to omit the outer columns in the keymap and use the correct key indexing for combo key positions by setting it in the `chosen` node of the keymap like:
```dts
...

/ {
    chosen {
        zmk,matrix_transform = &five_column_transform;
    };
...
[rest of the keymap]
```

You can also change the keyboard name by putting `CONFIG_ZMK_KEYBOARD_NAME="Chocofi"` in `corne.conf`.

infused-kim:
> Hi, I believe Chocofi is pin-compatible with the Corne, isn't it? In this case we might not want to duplicate the config.

Yes, that works. But the Chocofi is getting quite popular and from a usability perspective it would be easier for new zmk users if they could just select the keyboard instead of having to figure out which other keyboard is pin compatible. 

It comes down to what zmk maintainers value more, usability or ease of maintaining / duplication. 

Perhaps a compromise could be to create a Chocofi shield, but import the dtsi of the corne.

petejohanson:
Why not add a new field that indicates compatible keyboards to our metadata? We already could benefit from this with the cradio shield as well.

caksoylar:
I think it would be nice to incorporate matrix transforms into the metadata as well. This way user setup could let users select the matrix transform (maybe in the future) and compatible keyboards could indicate a subset of the main shield's transforms, etc.

manna-harbour:
> Why not add a new field that indicates compatible keyboards to our metadata? We already could benefit from this with the cradio shield as well.

That would be nice. You can see some compatible keyboards (including chocofi mentioned with corne and also another chocofi shield) in the [defacto ZMK out-of-tree keyboards directory](https://github.com/manna-harbour/miryoku/discussions/81).

Nick-Munnich:

PR #1786, opened 2023-04-27
refactor(shields): MurphPad

lesshonor:
- **Added matrix transforms and physical layouts for three configurations.** This is a numpad, but notably the first three positions of column 0 are actually tact switches on the back of the PCB. I believe the new transforms make this clearer (the keymap appearance matches QMK Configurator now, too)
- Tweaked default keymap:
  - more closely resembles the QMK Configurator default map (reduces confusion for users who are new to ZMK; also generally less niche)
  - more useful OOTB for ZMK (immediate access to output selection, easier BT profile switching).
- Cleaned up formatting a bit.
- Updated encoder configuration to current ZMK standards.
- Avoid forcing RGB on for `nice_nano`.
- Corrected the number of default LEDs. While some users may opt for less or more, there are exactly eight footprints on the PCB for 5050 SMD LEDs.

## Board/Shield Check-list

### DIFF
--- a/app/boards/shields/murphpad/murphpad.conf
+++ b/app/boards/shields/murphpad/murphpad.conf
@@ -1,9 +1,15 @@
-# Uncomment to turn on logging, and set ZMK logging to debug output
-# CONFIG_ZMK_USB_LOGGING=y
+# To enable logging, use the snippet per the documentation.
+# Delaying log output may help catch issues close to startup. Uncomment and adjust if necessary (default is 1000ms)
+#CONFIG_LOG_PROCESS_THREAD_STARTUP_DELAY_MS=3000
 
-# Uncomment both to enable encoder
+# Uncomment both to enable encoder(s)
 CONFIG_EC11=y
 CONFIG_EC11_TRIGGER_GLOBAL_THREAD=y
 
-# Uncomment to enable OLED
-CONFIG_ZMK_DISPLAY=y
+# Uncomment to disable OLED
+#CONFIG_ZMK_DISPLAY=n
+
+# Uncomment both to enable underglow
+# Note that this will only work if an implementation exists for your board; check under the shield folder for board-specific overlays.
+#CONFIG_ZMK_RGB_UNDERGLOW=y
+#CONFIG_WS2812_STRIP=y
--- a/app/boards/shields/murphpad/Kconfig.defconfig
+++ b/app/boards/shields/murphpad/Kconfig.defconfig
@@ -6,6 +6,9 @@ if SHIELD_MURPHPAD
 config ZMK_KEYBOARD_NAME
     default "MurphPad"
 
+config ZMK_DISPLAY
+    default y
+
 if ZMK_DISPLAY
 
 config I2C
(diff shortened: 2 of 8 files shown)


caksoylar (approved):
I had noticed the discrepancies when comparing vs. the configurator, looks good to me.

Nick-Munnich:
Looking at this the only point I want to raise is that it has OLED defaulting to `y`. Imo OLED/RGB/Display etc. should always be opt-in, rather than opt-out.

Nick-Munnich:
> I would agree with you on most shields. On this one: yes, you can opt out of the display, but the plate will have an unsightly hole in it unless you modify the plate file and get that alternate version cut. I've probably assisted with some fifty MurphPad builds in one way or another; plenty didn't bother with RGB or encoders, but no one has ever _not_ installed the display, even if they couldn't get it working. ...That said, I still wanted to facilitate the option _just in case_—hence "enabling the OLED in a less-authoritarian way". I'd do the same for the encoder if that was actually possible.

That's a fair argument. I think I'd still prefer explicit for consistency with other devices, though. We need to improve our related documentation on it though, imo. I didn't actually notice that the encoder stuff wasn't commented out, I'd obviously be in favor of doing that too.

All that said I am perfectly happy to be overruled in this case.

app/boards/shields/murphpad/murphpad.conf
```diff
@@ -1,9 +1,15 @@
-# Uncomment to turn on logging, and set ZMK logging to debug output
-# CONFIG_ZMK_USB_LOGGING=y
+# To enable logging, use the snippet per the documentation.
+# Delaying log output may help catch issues close to startup. Uncomment and adjust if necessary (default is 1000ms)
+#CONFIG_LOG_PROCESS_THREAD_STARTUP_DELAY_MS=3000
```
Nick-Munnich, line 3:
```suggestion
# CONFIG_LOG_PROCESS_THREAD_STARTUP_DELAY_MS=3000
```
Minor thing while I'm investigating to see what makes sense, the lack of spaces after the # is annoying me.

app/boards/shields/murphpad/murphpad.conf
```diff
@@ -1,9 +1,15 @@
...
-# Uncomment both to enable encoder
+# Uncomment both to enable encoder(s)
 CONFIG_EC11=y
 CONFIG_EC11_TRIGGER_GLOBAL_THREAD=y
 
-# Uncomment to enable OLED
-CONFIG_ZMK_DISPLAY=y
\ No newline at end of file
+# Uncomment to disable OLED
+#CONFIG_ZMK_DISPLAY=n
```
Nick-Munnich, line 10:
```suggestion
# CONFIG_ZMK_DISPLAY=n
```

app/boards/shields/murphpad/murphpad.conf
```diff
@@ -1,9 +1,15 @@
...
+
+# Uncomment both to enable underglow
+# Note that this will only work if an implementation exists for your board; check under the shield folder for board-specific overlays.
+#CONFIG_ZMK_RGB_UNDERGLOW=y
+#CONFIG_WS2812_STRIP=y
```
Nick-Munnich, line 15:
```suggestion
# CONFIG_ZMK_RGB_UNDERGLOW=y
# CONFIG_WS2812_STRIP=y
```

petejohanson (approved):
LGTM!

PR #1790, opened 2023-05-03
fix(boards): unify board I2C configuration, use TWI driver

krikun98:
https://github.com/petejohanson/zmk/commit/b74830bec6dab11c4cbf59b16eecb9a62b5ff429 (the commit moving boards to pinctrl) used the `nrf-twim` driver for all `nrfmicro` revisions and the `bluemicro840`.
That driver does not seem to work with ZMK OLED screens for some reason. Both the issue and the fix were observed and tested with `nrfmicro_13_52833` and `nrfmicro_13` with various shields.

Note: the Corne-ish Zen V1 uses the same `nrf-twim` driver for its fuel gauge, that seems to work fine (tested on my Zen). The issue only occurs with the screens.

### DIFF
--- a/app/boards/arm/bluemicro840/bluemicro840_v1.dts
+++ b/app/boards/arm/bluemicro840/bluemicro840_v1.dts
@@ -63,7 +63,7 @@
 };
 
 &i2c0 {
-    compatible = "nordic,nrf-twim";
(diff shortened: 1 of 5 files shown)


petejohanson (approved):
Good catch. Thanks!

PR #1793, opened 2023-05-08
Boot protocol support (based on #1140, with some refactoring)

chrisandreae:
This PR adding boot protocol support is based on @rbitanga's https://github.com/zmkfirmware/zmk/pull/1140, with some refactoring and other changes that we think are improvements, and is rebased and presented for review as @joelspadin requested [here](https://github.com/zmkfirmware/zmk/pull/1140#issuecomment-1317924785) (sorry about the huge delay!). The work was largely done by @thefloweringash.

The changes from #1140 are as follows:
* Minimize the api changes for Bluetooth by keeping `zmk_hid_get_keyboard_report()` and returning the .body like before.
* Keeping the logic about "full" vs "body" entirely within the usb code path.
* The endpoint now calls either `zmk_usb_hid_send_keyboard_report()` and   `zmk_usb_hid_send_consumer_report()` instead of `zmk_usb_hid_send_report(...)`.
    - These functions now internally dispatch on protocol to either       `zmk_hid_get_keyboard_report()` or a new function for boot reports      `zmk_hid_get_boot_report()`.
    - There's a change here from the PR version in the behavior of `get_report()`: when in boot protocol, don't include the report id. I believe this is correct, in that implicit boot protocol report descriptor does not use       multiple reports, so any boot protocol report should not include the report       id field.
* Use a single definition of a boot report, used for regular reports in
  non-6KRO, and for rollover in all branches.
* Handle gaps in the zmk report when producing a boot report in HKRO mode. For example, if it was 8KRO, it would be possible to have the state `0 0 0 0 0 0 0 17` (by pressing 8 keys, and letting go of the first 7). Copying the first 6 bytes would not show up the single pressed key.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -108,16 +108,6 @@ config ZMK_USB_BOOT
     default y
     select USB_HID_BOOT_PROTOCOL
     select USB_DEVICE_SOF
-
-if ZMK_USB_BOOT
-
-# Declare boot keyboard interface
-config USB_HID_PROTOCOL_CODE
-    default 1
-
-#ZMK_USB_BOOT
-endif
-
 #ZMK_USB
 endif
 
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -103,6 +103,21 @@ config USB_NUMOF_EP_WRITE_RETRIES
 config USB_HID_POLL_INTERVAL_MS
     default 1
 
+config ZMK_USB_BOOT
+    bool "USB Boot Protocol Support"
+    default y
+    select USB_HID_BOOT_PROTOCOL
+    select USB_DEVICE_SOF
+
+if ZMK_USB_BOOT
+
+# Declare boot keyboard interface
+config USB_HID_PROTOCOL_CODE
+    default 1
+
+#ZMK_USB_BOOT
+endif
+
 #ZMK_USB
 endif
 
@@ -575,4 +590,3 @@ osource "$(ZMK_CONFIG)/boards/shields/*/Kconfig.shield"
 
 
 source "Kconfig.zephyr"
-
--- a/app/include/zmk/hid.h
+++ b/app/include/zmk/hid.h
@@ -128,7 +128,7 @@ typedef struct zmk_hid_keyboard_report_body zmk_hid_boot_report_t;
 struct zmk_hid_boot_report {
     zmk_mod_flags_t modifiers;
     uint8_t _reserved;
-    uint8_t keys[6];
+    uint8_t keys[HID_BOOT_KEY_LEN];
 } __packed;
 
 typedef struct zmk_hid_boot_report zmk_hid_boot_report_t;
--- a/app/include/zmk/hid.h
+++ b/app/include/zmk/hid.h
@@ -116,12 +116,24 @@ static const uint8_t zmk_hid_report_desc[] = {
     HID_END_COLLECTION,
 };
 
-// struct zmk_hid_boot_report
-// {
-//     uint8_t modifiers;
-//     uint8_t _unused;
-//     uint8_t keys[6];
-// } __packed;
+#if IS_ENABLED(CONFIG_ZMK_USB_BOOT)
+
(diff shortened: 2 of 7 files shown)


typorian:
Any chance to get this looked at again? It'd be really neat to use my kyria to type in my bitlocker password.

app/Kconfig
```diff
@@ -103,6 +103,21 @@ config USB_NUMOF_EP_WRITE_RETRIES
...
+    bool "USB Boot Protocol Support"
+    default y
+    select USB_HID_BOOT_PROTOCOL
+    select USB_DEVICE_SOF
+
+if ZMK_USB_BOOT
+
+# Declare boot keyboard interface
+config USB_HID_PROTOCOL_CODE
+    default 1
```
petejohanson, line 116: menuconfig mentions this Kconfig as deprecated:

```
Name: USB_HID_PROTOCOL_CODE
Prompt: HID Boot Interface protocol code (DEPRECATED)
Type: int
Value:

Help:

  This option is deprecated.
  Please use usb_hid_set_proto_code() instead.
```

so we should probably switch to using that API instead.

app/include/zmk/hid.h
```diff
@@ -91,12 +91,24 @@ static const uint8_t zmk_hid_report_desc[] = {
...
+#define HID_BOOT_KEY_LEN 6
+
+#if IS_ENABLED(CONFIG_ZMK_HID_REPORT_TYPE_HKRO) &&                                                 \
+    CONFIG_ZMK_HID_KEYBOARD_REPORT_SIZE == HID_BOOT_KEY_LEN
+typedef struct zmk_hid_keyboard_report_body zmk_hid_boot_report_t;
+#else
+struct zmk_hid_boot_report {
+    zmk_mod_flags_t modifiers;
+    uint8_t _reserved;
+    uint8_t keys[6];
```
petejohanson, line 106:
```suggestion
    uint8_t keys[HID_BOOT_KEY_LEN];
```

app/include/zmk/hid.h
```diff
@@ -91,12 +91,24 @@ static const uint8_t zmk_hid_report_desc[] = {
...
-// } __packed;
+#if IS_ENABLED(CONFIG_ZMK_USB_BOOT)
+
+#define HID_ERROR_ROLLOVER 0x1
+#define HID_BOOT_KEY_LEN 6
+
+#if IS_ENABLED(CONFIG_ZMK_HID_REPORT_TYPE_HKRO) &&                                                 \
+    CONFIG_ZMK_HID_KEYBOARD_REPORT_SIZE == HID_BOOT_KEY_LEN
+typedef struct zmk_hid_keyboard_report_body zmk_hid_boot_report_t;
+#else
```
petejohanson, line 102: Nice

app/src/usb_hid.c
```diff
@@ -23,11 +23,78 @@ static K_SEM_DEFINE(hid_sem, 1, 1);
 
 static void in_ready_cb(const struct device *dev) { k_sem_give(&hid_sem); }
 
+#define HID_GET_REPORT_TYPE_MASK 0xff00
+#define HID_GET_REPORT_ID_MASK 0x00ff
+
+#define HID_REPORT_ID_KEYBOARD 1
+#define HID_REPORT_ID_CONSUMER 2
+
```
petejohanson, line 31: These are defined now in `hid.h` now that #1993 is in. Please rebase and leverage those.

petejohanson (changes requested):
Just a few minor things. On a whole this looks good.

petejohanson (approved):
LGTM. Tested here on several BIOSes and on Linux, Android and Windows here and no impact to existing USB functionality. Thanks!

typorian:
@petejohanson Thank you very much for getting this moved along. I'm not sure how to diagnose this, but after entering my bitlocker pin (which works now, yay!) I have to reconnect the keyboard at the login screen, because now it doesn't immediately work there anymore. Unfortunately I only have this one thinkpad with bitlocker where I can test this.

PR #1802, opened 2023-05-14
fix: Enable BT_TINYCRYPT_ECC when using HCI

joelspadin:
`BT_TINYCRYPT_ECC` is required for `BT_SMP_SC_PAIR_ONLY` to work on setups that use BT HCI like the nRF5340.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -120,6 +120,10 @@ menuconfig ZMK_BLE
 
 if ZMK_BLE
 
+# BT_TINYCRYPT_ECC is required for BT_SMP_SC_PAIR_ONLY when using HCI
+config BT_TINYCRYPT_ECC
+    default y if BT_HCI && !BT_CTLR
+
 choice BT_LL_SW_LLCP_IMPL
     default BT_LL_SW_LLCP_LEGACY
 


app/Kconfig
```diff
@@ -118,6 +118,12 @@ menuconfig ZMK_BLE
     select BT_SETTINGS
     select SETTINGS
 
+# BT_TINYCRYPT_ECC is required for BT_SMP_SC_PAIR_ONLY when using HCI
+if BT && !BT_CTLR
+config BT_TINYCRYPT_ECC
+    default y
+endif
```
petejohanson, line 125: Any reason this can't be simplified?

```suggestion
config BT_TINYCRYPT_ECC
    default y if !BT_CTLR
```

joelspadin: Because I don't know how Kconfig works? 😄 I'll try that.

joelspadin: I tried that and it didn't work. Then I went back to what's in this commit, and that doesn't work either.

joelspadin: Seems to just be build system weirdness. It works with a pristine build.

petejohanson (approved):
Thanks!

PR #1803, opened 2023-05-14
refactor: Split endpoint to transport and instance

joelspadin:
Changed the endpoints code to rename the existing endpoint types to "endpoint transport" and add the concept of "endpoint instances". A transport is the method by which data is sent, while instances allow describing multiple endpoints that use the same transport (e.g. bluetooth profiles).

Also added new APIs to get the total number of possible endpoint instances and assign each instance a unique index, which can be used for tracking separate state for each endpoint in other code files.


app/include/zmk/endpoints_types.h
```diff
@@ -6,7 +6,21 @@
...
+};
+
+/**
+ * A specific endpoint to which data may be sent.
+ */
+struct zmk_endpoint_instance {
+    enum zmk_transport transport;
+    union {
+        // ZMK_TRANSPORT_USB: no data
+        int ble_profile_index; // ZMK_TRANSPORT_BLE
```
petejohanson, line 24: I wonder, since we might have someone add a custom 2.4Ghz transport later, for instance, if we would want to do something like:


```suggestion
        struct zmk_transport_ble_details ble; // ZMK_TRANSPORT_BLE
```

And

```
struct zmk_transport_ble_details {
    int profile_index;
};
```

So you'd access `endpoint.ble.profile_index` to be *really* explicit.

Thoughts?

app/boards/arm/corneish_zen/widgets/output_status.c
```diff
@@ -113,14 +113,7 @@ static void output_status_update_cb(struct output_status_state state) {
...
                             output_status_update_cb, get_state)
-ZMK_SUBSCRIPTION(widget_output_status, zmk_endpoint_selection_changed);
-
-#if IS_ENABLED(CONFIG_USB_DEVICE_STACK)
-ZMK_SUBSCRIPTION(widget_output_status, zmk_usb_conn_state_changed);
-#endif
-#if defined(CONFIG_ZMK_BLE)
-ZMK_SUBSCRIPTION(widget_output_status, zmk_ble_active_profile_changed);
-#endif
+ZMK_SUBSCRIPTION(widget_output_status, zmk_endpoint_changed);
```
petejohanson, line 116: Does the endpoint change event actually fire when the active profile becomes cleared? That's one edge case that is handled because the "active profile changed" actually fires when the active profile goes from paired to open/cleared.

petejohanson (commented):
On a whole, I really like it. A couple minor questions/suggestions.

app/include/zmk/endpoints_types.h
joelspadin: I used an anonymous union here so that things unlikely to have any extra data wouldn't need that extra level, but you could still add them for things that needed them. That said, there really isn't much of a difference between
```c
endpoint.ble_profile_index
```
and
```c
endpoint.ble.profile_index
```
so I think I'll go with what you suggested. It would also help avoid needing to update a lot of code if we ever needed to add a new field for an existing transport.

app/boards/arm/corneish_zen/widgets/output_status.c
joelspadin: Unsure. It does check for changes based on active profile changed events (https://github.com/zmkfirmware/zmk/pull/1803/files#diff-5f5f09e0a8d09090e60302308d29a5906d8a41a643992a1d5c4a047033c65688R331), but I haven't checked whether that results in a change to the selected endpoint and therefore fires this event. Will check that.

app/include/zmk/endpoints_types.h
petejohanson: Yeah, avoiding destructive refactors later is definitely an added bonus.

app/src/endpoints.c
```diff
@@ -16,44 +16,90 @@
...
+    switch (endpoint.transport) {
+    case ZMK_TRANSPORT_USB:
+        return INSTANCE_INDEX_OFFSET_USB;
+
+    case ZMK_TRANSPORT_BLE:
+        return INSTANCE_INDEX_OFFSET_BLE + endpoint.ble_profile_index;
+    }
+
+    LOG_ERR("Invalid transport %d", endpoint.transport);
+    return 0;
```
snoyer, line 92: Does that make any invalid case return the same `0` value as the valid USB case?

joelspadin: Yes. I didn't want to pick a value that wasn't a valid array index, because then everything that uses this would have to add error checking for a case that shouldn't ever occur.

snoyer: > Yes. I didn't want to pick a value that wasn't a valid array index, because then everything that uses this would have to add error checking 

I assumed that was the reason

> for a case that shouldn't ever occur.

My concern was silently overriding existing data when passing an invalid instance but if that invalid case is practically unreachable then nevermind

app/src/endpoints.c
```diff
@@ -233,18 +300,21 @@ static void disconnect_current_endpoint() {
...
-        /* Cancel all current keypresses so keys don't stay held on the old endpoint. */
+    if (!zmk_endpoint_instance_equals(new_instance, current_instance)) {
+        // Cancel all current keypresses so keys don't stay held on the old endpoint.
         disconnect_current_endpoint();
 
-        current_endpoint = new_endpoint;
-        LOG_INF("Endpoint changed: %d", current_endpoint);
+        current_instance = new_instance;
+
+        char endpoint_str[10];
```
snoyer, line 312: Can we have a `#define` with the nominal length `zmk_endpoint_instance_print` needs (similar to [`BT_ADDR_LE_STR_LEN` for `bt_addr_to_str`](https://github.com/zephyrproject-rtos/zephyr/blob/706cfbbb76151f4223c225c98295a44c8ecec598/include/zephyr/bluetooth/addr.h#L206-L225))?

app/src/endpoints.c
snoyer: Should the return type be unsigned if there's no error codes?

snoyer: I see, I was under the impression array indices were better as `uint` or `size_t` but it's apparently not that simple https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es107-dont-use-unsigned-for-subscripts-prefer-gslindex. Thanks for the detailed answer.

app/src/endpoints.c
joelspadin: Done.

app/boards/arm/corneish_zen/widgets/output_status.c
joelspadin: Looks like it doesn't update if you aren't plugged in to USB and the active profile disconnects. Will fix.

joelspadin: Added the BLE event handler back in with a note as to why it's still needed.

app/include/zmk/endpoints_types.h
joelspadin: Done.

app/boards/arm/corneish_zen/widgets/output_status.c
```diff
@@ -39,31 +38,31 @@ LV_IMG_DECLARE(USB_connected);
 static sys_slist_t widgets = SYS_SLIST_STATIC_INIT(&widgets);
 
 struct output_status_state {
-    enum zmk_endpoint selected_endpoint;
+    struct zmk_endpoint_instance selected_endpoint;
     bool active_profile_connected;
     bool active_profile_bonded;
     uint8_t active_profile_index;
```
petejohanson, line 44:
```suggestion
```

Looks like this is no longer used here?

app/include/zmk/endpoints.h
```diff
@@ -6,10 +6,66 @@
...
+/**
+ * The total number of different (struct zmk_endpoint_instance) values that can
+ * be selected.
+ *
+ * Note that this value may change between firmware versions, so it should not
+ * be used in any persistent storage.
+ */
+#define ZMK_ENDPOINT_COUNT (ZMK_ENDPOINT_USB_COUNT + ZMK_ENDPOINT_BLE_COUNT)
+
+bool zmk_endpoint_instance_equals(struct zmk_endpoint_instance a, struct zmk_endpoint_instance b);
```
petejohanson, line 38: So, Zephyr uses `eq` shorthand pretty consistently, any reason not to use that convention as well here?

petejohanson (commented):
Just a couple minor things. On a whole this looks great.

app/include/zmk/endpoints.h
joelspadin: Nope. Wasn't aware of that convention. I can update this.

joelspadin: Done. Also changed the `_print()` function to `_to_str()` and reordered the arguments to match what looks like Zephyr conventions for that type of function as well.

petejohanson:
Tested here on my ZMK Uno, and toggling the 3-way switch to toggle the selected endpoint all works as expected, including the output display, which is one of the main consumers of the change to the events. Will review the code, but functionality wise it looks good.

app/boards/arm/corneish_zen/widgets/output_status.c
petejohanson: @joelspadin Did you see this review comment?

petejohanson (changes requested):
Code and behavior all looks great, just my one comment about the extra struct field that looks like it can be removed as well.

app/src/events/endpoint_selection_changed.c
caksoylar, line 1: Does it make sense to rename this file as well?

joelspadin: Yep. Done.

app/boards/arm/corneish_zen/widgets/output_status.c
joelspadin: Whoops. I missed that. Done.

joelspadin:
Addressed the review comments and fixed a missing include. I put the fixes up as a separate commit so you can easily review the difference. (You'll probably want to squash them together when merging this.)

joelspadin:
Guess I missed the nice!view code since that was probably added after I started this.

joelspadin:
This last commit is untested aside from that it builds. It would be good to have someone verify this on a nice!view. (I do have one, but I haven't hooked it up to anything yet, so I'd need to do a bit of work before I could test it myself.)

I also took the opportunity to optimize a few functions by passing the large status struct by pointer instead of value.

petejohanson (approved):
This all looks good to me. I can *build* with a nice!view here, but don't easily have hardware to test on right now either. Will put out call for testers on Discord for some validation before merging this.

petejohanson:
Ok, was able to test this on my ZMK Uno built with a nice!view, and they all build *and* run perfectly fine. Merging, thanks @joelspadin!

PR #1807, opened 2023-05-19
refactor(split): allow central to define connection parameters

### LINKED ISSUE #1614: Define split connection parameters on central instead of peripheral
The split connection parameters are currently defined on both the central (https://github.com/zmkfirmware/zmk/blob/main/app/src/split/bluetooth/central.c#L412) and peripheral (https://github.com/zmkfirmware/zmk/blob/main/app/src/split/bluetooth/Kconfig#L57-L58).

I propose defining the parameters only on the central since it is responsible for initiating the connection and must approve changes. This makes it easier to change the parameters either via Kconfig or dynamically (to reduce power consumption, for instance).

- Add `CONFIG_BT_GAP_AUTO_UPDATE_CONN_PARAMS=n` to peripheral.
- Define options `ZMK_BLE_PERIPHERAL_PREF_INT`, `ZMK_BLE_PERIPHERAL_PREF_LATENCY`, and `ZMK_BLE_PERIPHERAL_PREF_TIMEOUT` on central and using those to initiate connection to peripheral.

xudongzheng:
Fixes #1614

### DIFF
--- a/app/src/split/bluetooth/Kconfig
+++ b/app/src/split/bluetooth/Kconfig
@@ -33,6 +33,18 @@ config ZMK_SPLIT_BLE_CENTRAL_SPLIT_RUN_QUEUE_SIZE
     int "Max number of behavior run events to queue to send to the peripheral(s)"
     default 5
 
+config ZMK_SPLIT_BLE_PREF_INT
+    int "Connection interval to use for split central/peripheral connection"
+    default 6
+
+config ZMK_SPLIT_BLE_PREF_LATENCY
+    int "Latency to use for split central/peripheral connection"
+    default 30
+
+config ZMK_SPLIT_BLE_PREF_TIMEOUT
+    int "Supervision timeout to use for split central/peripheral connection"
+    default 400
+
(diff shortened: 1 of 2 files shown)


petejohanson (commented):
Code seems reasonable. Has this tested well without any issues, in particular if you fail to flash both sides, and the peripheral still requests different params?

xudongzheng:
In the event that the central is updated to include these options (and kept on the defaults) but peripheral continues running the old firmware, nothing really changes. It's the same as before except central is now getting the initial connection parameters from `ZMK_SPLIT_BLE_PREF_*` instead of having it hardcoded in `central.c`.

If the parameters are changed from the defaults on central but peripheral continues running the old firmware: central will initiate a connection with the new custom parameters but peripheral will request the old defaults. In theory central has the ability to reject that request but that seems like unnecessary code that will eventually serve no purpose.

petejohanson:
This will need rebasing now that #836 has been merged. There's also some CI failures that hopefully will be addressed when you do so. Thanks again.

petejohanson (approved):
Thanks!

PR #1811, opened 2023-05-23
feat(behaviors): hold-tap hold while undecided

theol0403:
This PR builds on the changes from #1398 and attempts to implement @okke-formsma 's suggestions. Addresses https://github.com/zmkfirmware/zmk/issues/1393.

#### `hold-while-undecided`

If enabled, the hold behavior will immediately be held on hold-tap press, and will release before the tap behavior is sent.

#### `hold-while-undecided-linger`

If your tap behavior activates the same modifier as the hold behavior, and you want to avoid a double tap when transitioning from the hold to the tap, you can use `hold-while-undecided-linger`. When enabled, the hold behavior will continue to be held until _after_ the tap behavior is released. For example, if the hold is `&kp LGUI` and the tap is `&sk LGUI`, then with `hold-while-undecided-linger` enabled, the host will see `LGUI` held down continuously until the sticky key is finished, instead of seeing a release and press when transitioning from hold to sticky key.

## Discussion
I did my best to implement the suggestions in https://github.com/zmkfirmware/zmk/pull/1398#issuecomment-1279818462. However, some of them seemed unnecessary, and my implementation ended up looking quite like @nickconway 's. There seemed to be no benefit to adding `STATUS_UNDECIDED_INITIAL_HOLD`, as it wasn't used in the logic to decide whether to hold, and it was just treated identically to `STATUS_UNDECIDED`. 

I am not sure about the name of `hold-while-undecided-linger`, any suggestions? Maybe `hold-while-undecided-release-deferred` or `hold-while-undecided-around-tap`?

PR #1812, opened 2023-05-23
feat(behaviors): lazy sticky keys

theol0403:
This PR adds a new feature to sticky keys. 

#### `lazy`

By default, sticky keys are activated on press until another key is pressed. You can enable the `lazy` setting to instead activate the sticky key right before the other key is pressed. This is useful for mouse interaction or situations where you don't want the host to see anything during a sticky-key timeout, for example `&sk LGUI`, which can trigger a menu if pressed alone.

## Discussion
This PR is fairly straightforward. I use it along with `hold-while-undecided` (https://github.com/zmkfirmware/zmk/pull/1811) to make nice callum-style sticky mods that are invisible to the OS and don't interfere with mouse interactions. I can also use it without `hold-while-undecided` to make a sticky `LGUI` that doesn't trigger the OS menu upon timeout.


theol0403:
I force-pushed a V2 of this feature. This time, all the tests work, and a lot of edge cases were solved, especially with compound behaviors.

This PR now introduces two breaking changes to the regular sticky key behavior. These changes make more sense to me, and play nicer with the way lazy sticky keys are implemented, but can be reverted if necessary. 

### Sticky Release Order
Previously, the order of events emitted for a sticky mod (sticky tapped -> other key tapped) would be:
 `mod pressed -> other key pressed -> mods released -> other key released`. 
However, my PR changes this to:
 `mod pressed -> other key pressed -> other key released -> mods released`.

Specifically, when the other key is released, my PR ensures the other key is released *before* the sticky key is released, not the other way around. This means that the sticky key now properly sends its keycodes "around" the other key.

### Timeout while other key held
Previously, when the sticky key is tapped, and then the other key is held indefinitely, the sticky key will release on timeout even if the other key is held. My PR changes this to release if and only if the other key is released, so the sticky key is again always "around" the other key. 

### Discussion
This PR is now fully ready for review. I'd love feedback if these changes are reasonable, or if there are any situations where they would be a regression.

theol0403:
Note that tapping a lazy sticky key will not trigger other behaviors such as the release of other sticky keys or layers. If you want to use a lazy sticky key to activate the release of a sticky layer, potential solutions include wrappping the sticky key in a simple macro which presses the sticky behavior around the sticky key press, doing the same with `&mo LAYER`, or triggering a tap of some key like `K_CANCEL` on sticky key press.

theol0403:
Bump - this is a nice improvement to sticky keys that help with layouts using sticky keys on OSs like windows.

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -266,7 +303,12 @@ void behavior_sticky_key_timer_handler(struct k_work *item) {
     if (sticky_key->timer_cancelled) {
         sticky_key->timer_cancelled = false;
     } else {
-        release_sticky_key_behavior(sticky_key, sticky_key->release_at);
+        // If the key is lazy, a release is not needed on timeout
+        if (sticky_key->config->lazy) {
+            clear_sticky_key(sticky_key);
+        } else {
+            release_sticky_key_behavior(sticky_key, sticky_key->release_at);
+        }
```
petejohanson, line 311: This same conditional logic is added in two different place. Can we pull it out into a dedicated function, or move the logic directly into `release_sticky_key_behavior`?

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -247,14 +265,33 @@ static int sticky_key_keycode_state_changed_listener(const zmk_event_t *eh) {
...
+    bool event_reraised = false;
+    for (int i = 0; i < ZMK_BHV_STICKY_KEY_MAX_HELD; i++) {
+        struct active_sticky_key *sticky_key = sticky_keys_to_release_after_reraise[i];
+        if (sticky_key) {
+            if (!event_reraised) {
+                ZMK_EVENT_RAISE_AFTER(eh, behavior_sticky_key);
+                event_reraised = true;
+            }
+            release_sticky_key_behavior(sticky_key, ev_copy.timestamp);
+        }
```
petejohanson, line 291: I favor early continue to avoid deeper nesting:

```suggestion
        if (!sticky_key) {
            continue;
        }
        
        if (!event_reraised) {
            ZMK_EVENT_RAISE_AFTER(eh, behavior_sticky_key);
            event_reraised = true;
        }
        release_sticky_key_behavior(sticky_key, ev_copy.timestamp);
```

app/src/behaviors/behavior_sticky_key.c
```diff
@@ -247,14 +265,33 @@ static int sticky_key_keycode_state_changed_listener(const zmk_event_t *eh) {
...
         }
     }
-    if (event_reraised) {
-        return ZMK_EV_EVENT_CAPTURED;
+
+    // give each sticky key a chance to press their behavior before the event is reraised
+    for (int i = 0; i < ZMK_BHV_STICKY_KEY_MAX_HELD; i++) {
+        struct active_sticky_key *sticky_key = sticky_keys_to_press_before_reraise[i];
+        if (sticky_key) {
+            press_sticky_key_behavior(sticky_key, ev_copy.timestamp);
```
petejohanson, line 277: Do we need to clear if from `sticky_keys_to_press_before_reraise` now that it's pressed?

petejohanson (commented):
I like it! A few questions on the implementation. Thanks!

app/src/behaviors/behavior_sticky_key.c
theol0403: It will only loop once through the array before discarding it (and it is null-initialized). The purpose is simply to collect all the out-of-order presses (of just-in-time mods) and releases (of expired mods) and execute them in order "around" the other key.

app/src/behaviors/behavior_sticky_key.c
theol0403: I'll pull it out. The logic doesn't really make sense in `release_sticky_key_behavior`, because the logic reads as "if we've timed out, but we are lazy, then we don't need to release the behavior because we never pressed it. Otherwise, release the behavior."

app/tests/sticky-keys/11-lazy-timeout-during/native_posix_64.keymap
```diff
@@ -0,0 +1,37 @@
...
+
+&sk {
+    lazy;
+    release-after-ms = <50>;
+};
+
+/ {
+    keymap {
+        compatible = "zmk,keymap";
+        label ="Default keymap";
```
theol0403, line 15: I read somewhere about labels being removed... just wanted to bring this to your attention if that's something you don't want to merge in.

joelspadin: Yep. The label property is now deprecated. It won't hurt to include it (aside from a compiler warning that will probably just get ignored), but ideally it should be removed.

caksoylar (commented):
Docs LGTM (not doing an approve to not confuse the rest of the review situation)

petejohanson (approved):
Thanks!

PR #1816, opened 2023-05-30
feat(docs): Add note on modifier functions to macros

caksoylar:
Add a shost note for a common use case that is easier solved by modifier functions than macros. Unfortunately the modifier functions section isn't very discoverable since it only shows up when you click on [the modifiers section](https://zmk.dev/docs/codes/modifiers) and not on the full codes page.


petejohanson (approved):
Good call, this comes up pretty frequently. Thanks!

PR #1817, opened 2023-05-30
feat(docs): Add behavior types section for hold-tap

caksoylar:
Add a subsection under advanced usage that details use cases involving different behavior types, especially relevant for behaviors that take zero or two parameters. The parameter issue is a frequent question on the Discord and this would also help to make more explicit that hold-tap can work with (pretty much) any behavior.


petejohanson (approved):
Another great call. Thanks!

PR #1818, opened 2023-05-30
docs: Move BT troubleshooting items and note known issue for Windows

caksoylar:
Move BT-related items from under troubleshooting page to under the troubleshooting section under new BT feature page, then link to that from the original page.

In addition:
- note the known issue with Windows sometimes not receiving keystrokes and provide a workaround[^1]
- note that TX power config also improves split keyboard inter-half connectivity

[^1]: Recently (after the Zephyr 3.2 upgrade) this issue popped up, where some Windows devices can be properly connected but not receive key strokes from the keyboard. A simple un-pair/re-pair doesn't seem to solve it and other actions detailed here is required. 
I have seen this issue reported a few times and I experienced it myself once, and they were solved with these suggestions.
I think investigating the root cause would be good but meanwhile, we should have it noted since it has a simple but non-obvious fix.

### DIFF
--- a/docs/docs/features/bluetooth.md
+++ b/docs/docs/features/bluetooth.md
@@ -38,6 +38,20 @@ Management of the bluetooth in ZMK is accomplished using the [`&bt` behavior](..
 
 ## Troubleshooting
 
+### Connectivity Issues
+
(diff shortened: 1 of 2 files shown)


petejohanson (approved):
:+1:

Issue #34, opened 2020-06-24
ZMK CLI


innovaker:
- A CLI tool that serves as the main point of entry for most users.
- Probably written in Python and published as `zmk`.
- Wraps up git, west and other tooling.
  - Similar to how west wraps around git, cmake, ninja, etc.
- Encapsulates multiplatform support.
- Provides sanity checks on parameters.
  - e.g. checks the board, shield, keymap etc. and invalid combinations of them.
- Provides rudimentary error handling.
  - Provides useful feedback for common errors.
  - Points users to ZMK help for unexpected errors.
- Perhaps provides an interactive prompt/menu for commands and parameters.
- Perhaps caches parameters.
- Perhaps wrap flashing as well?
- Perhaps setting up dev environments?

From:
- https://discord.com/channels/719497620560543766/719565084208398406/720010131613417493
- https://discord.com/channels/719497620560543766/719565084208398406/720016130554462290
- https://discord.com/channels/719497620560543766/719565084208398406/721331753393848451
- https://discord.com/channels/719497620560543766/719565084208398406/721462710356672602

innovaker:
Should build upon multiplatform exploration started by #12 and #26.

innovaker:
@petejohanson, would we still like to proceed with this idea - not as a replacement for `setup.sh` - but as a development tool?

If so, then let's add it to Hacktoberfest.

petejohanson:
I'm kind of on the fence here. To a certain degree, this "hides" west, but only so well. I would rather consider what we really want to get from such a tool before we suggest it fora newcomer. Having someone spend a bunch of time on something, that we then say "Nah, that's not what we want!" really would stink.

okke-formsma:
I have set up one `west` command (`west test`) as a proof of concept; https://github.com/zmkfirmware/zmk/pull/385

caksoylar:
For reference, @joelspadin developed https://github.com/joelspadin/zmk-setup-script that currently encapsulates the user setup script in a Python-based CLI that uses an extensible `zmk` module.

joelspadin:
After using QMK again recently, I was thinking of borrowing some ideas from that, such as having a command such as `zmk install` which creates a config repo and remembers where it is, and then all further commands run from anywhere will operate on that repo unless otherwise specified.

That setup script might then be split across `zmk install` and `zmk add-keyboard` commands, with `zmk` being a script you can pip install.

joelspadin:
[...]

PR #1820, opened 2023-05-30
feat(shields): Add dactyl shields

dunk2k:
## Board/Shield Check-list


caksoylar:
I am not sure if it makes sense to include handwired keyboards (which these are, right?) in the main ZMK repo, since there is no standard on how a user is supposed to wire them (or is there?). If it is to be of any use, there needs to be extensive documentation on how everything is wired for these shields so the user can follow it. But I don't like that idea very much either.

Ideally our new shield documentation would be straightforward enough for users to follow for bespoke handwired keyboards. Another thing that would help is a ZMK CLI #34 with a `new-keyboard` command to output a boilerplate, or prompt the user for pin assignments and optional features.

dunk2k:
> I am not sure if it makes sense to include handwired keyboards (which these are, right?) 

That's correct. If there was documentation to declare that handwired boards are not excepted, it would've spared me this effort.

> in the main ZMK repo, since there is no standard on how a user is supposed to wire them (or is there?). If it is to be of any use, there needs to be extensive documentation on how everything is wired for these shields so the user can follow it. But I don't like that idea very much either.

I can create wiring documentation for each shield, though it appears the effort might be wasted due to a clear indication of aversion. Were I to create a wiring diagram for each shield, and inadvertently set a presedence for handwired boards in ZMK, would this PR be approved?
i.e. Am I wasting my time putting any more effort into this?

dunk2k:
> OSH/personal projects without general availability should create a zmk-config repo instead

I would ask for clarification of 'general availability'.  
If there was a clause that exclusively mentions, not infers, that handwired shields are not permitted due to specific reasons**, I'd of genuinely been averted from cresting this PR.

** - i deduce is because a handwired shield isn't strictly defined as a PCB design is; which is understandable, if it is a principle of the 'shield' concept/terminology.

> (Just to clarify, I am presenting my own opinions as a contributor here and not commenting in behalf of the project/maintainers.)

I appreciate the heads up, of what may to come 🙂

dunk2k:
Closing based on review comments

PR #1821, opened 2023-05-31
docs: Correction for Japanese charactor

kadoyau:
I found a typo. 
There is no Japanese word "カソマ". "カンマ" (or "コンマ") is correct. 
"ソ" and "ン" are different characters :)


caksoylar (approved):
Translators agree with the latter being comma properly, LGTM.

PR #1822, opened 2023-06-03
fix(bluetooth): Passkey pairing improvements.

petejohanson:
* Capture the last 6 entered digits, and then require pressing
   Enter/Return to submit the entered digits. This matches the
   messaging shown on hosts regarding how to complete pairing.
* Fix the wording on the Kconfig menu item to accurately describe
  the feature.

Tested on Linux w/ GNOME and on Android, both properly show the expected dialog upon pairing, and pairing only succeeds when a valid value is typed in on the keyboard, followed by pressing Enter/Return.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -160,8 +160,9 @@ config BT_DEVICE_APPEARANCE
     default 961
 
 config ZMK_BLE_PASSKEY_ENTRY
-    bool "Experimental: Requiring typing passkey from host to pair BLE connection"
+    bool "Require passkey entry on the keyboard to complete pairing"
     default n
+    select RING_BUFFER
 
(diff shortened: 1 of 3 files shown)


caksoylar:
Looks good to me and working as expected so far, tested on Windows 11, Android, iOS and iPad OS.

One thing that might be good to note (maybe as a sentence in the config table item?): Existing paired profiles require re-pairing after enabling this option.

petejohanson:
> Looks good to me and working as expected so far, tested on Windows 11, Android, iOS and iPad OS.
> 
> One thing that might be good to note (maybe as a sentence in the config table item?): Existing paired profiles require re-pairing after enabling this option.

Good call. Added a note in parentheses there.

caksoylar (approved):

PR #1823, opened 2023-06-03
fix(docs): Fix broken link in BT troubleshooting

caksoylar:
Missed updating a link target while moving BT troubleshooting items in #1818.


petejohanson (approved):
Good catch!

PR #1824, opened 2023-06-05
fix(boards): Proper i2c pinctrl for BlueMicro840

petejohanson:
* Use the proper pin assignmets after the move to pinctrl for the
  Zephyr 3.2 migration.


petejohanson:
Tested by `hainguyen` on Discord and confirmed working.

PR #1826, opened 2023-06-06
feat(docs): Add an example for combining just modifiers

filterpaper:
An an explicit documentation example of combining just modifiers.


docs/docs/codes/modifiers.mdx
```diff
@@ -38,6 +38,8 @@ These functions take the form: `XX(code)`
   - `&kp LS(A)` = `LEFT_SHIFT`+`A` (a capitalized **A**).
 - They can be combined:
   - `&kp LC(RA(B))` = `LEFT_CONTROL`+`RIGHT_ALT`+`B`
+- Modifiers can be combined to activate two or more of them at the same time:
```
caksoylar, line 41:
```suggestion
- They can be applied to a modifier keycode to create combined modifier keys:
```
I think this is a bit clearer and distinguishes it more from the previous item, but let me know what you think.

caksoylar (commented):
Thanks for taking the time to add the item, it is frequently asked.

caksoylar (approved):

Issue #1207, opened 2022-04-03
Key repeat cannot be rolled


andrewjrae:
Currently key repeat cannot be rolled with another key. For example: `G-down -> key_repeat-down -> G-up -> key_repeat-up` results in `g`, when it should result in `gg`. This is important functionality for those trying to use key repeat in everyday typing to reduce same finger bigrams.

This is also somewhat related to #1076, since the key code needs to be released before being tapped again.

It may be possible to fix this at the core by enforcing that a release is sent before pressing a code that is already pressed (this should also solve the problem in #1076). However, I'm not sure this is always desirable behavior (and could  potentially end up in erroneous releases being sent), and so for now I'll implement a patch just for key repeat. If others think that this standalone patch is preferred, I'll put up a PR.

gaoDean:
Used it for a few minutes, 10/10. Thank you so much!

ze-or:
since I replaced  ZMK_EV_EVENT_CAPTURED with ZMK_EV_EVENT_HANDLED crashing has stopped.

The rolling issue for triple presses of the same key is the only remaining issue I can tell.

andrewjrae:
FYI @ze-or I've implemented a new patch that should fix your triple roll case, by patching at the core. I have yet to test this for very long, but so far it seems to work well.

@petejohanson, I think this fix is more along the lines of what we'd want, and I _think_ I put it in a good spot.

If anyone else has success with this let me know and I can put up a PR!

caksoylar:
It works fine for me, e.g. to roll `+=` or `-_-` which weren't possible before, thanks @andrewjrae!

Issue #759, opened 2021-04-10
Modifiers are not working well with Windows RDP app


caksoylar:
There are issues using sending key presses with modifier functions and overall issues with held modifiers when using the RDP app ("Remote Desktop Connection") on Windows 10. I use it for connecting to a local VM with Hyper-V using `vmconnect.exe`, so there is no network latency. For instance when sending a key press like `LALT(TAB)` the modifier is almost always ignored. When using regular modifiers like Shift, they are also not detected some of the time.

Note that there is also a different RDP app ("Remote Desktop Classic" on MS Store) that doesn't seem to have the issue, so it is likely related to the peculiar handling of keyboard inputs of the former app. However this app doesn't allow connections to Hyper-V VMs.

Here is me sending `&kp UNDER` in the host (not using RDP) on https://config.qmk.fm/#/test:
![image](https://user-images.githubusercontent.com/7876996/114259576-c2a08f80-9983-11eb-8326-51997be30ccc.png)

Here it is on the remote VM through the classic RDP app:
![image](https://user-images.githubusercontent.com/7876996/114259584-d2b86f00-9983-11eb-8b08-4792873d73e3.png)

Worth noting that the `minus` keycode is also reported differently, but I don't know if that has any relevance.

okke-formsma:
Could you show the timings when you press a keycode like that in QMK? The solution for #756 does not apply here, as that is for hold-taps and does not apply to shifted keycodes.

caksoylar:
This is a bit interesting, here is QMK underscore keycode on local (no RDP), one short tap and one long tap:
![image](https://user-images.githubusercontent.com/7876996/114295893-0ca97400-9a5d-11eb-9eef-9d0fb639948d.png)

This is the same on remote:
![image](https://user-images.githubusercontent.com/7876996/114295909-277be880-9a5d-11eb-90b1-cc67a1cb2dd5.png)

First of all the order of key ups are different, QMK seems to release shift ~16ms after the minus, but ZMK releases shift before the minus (they seem to be held for the exact same duration).

I also have issues getting manual shift holds recognized with ZMK if I am not pressing it very deliberately which doesn't happen with QMK. Let me know if there is anything I can share to help debug that.

okke-formsma:
Are you using ZMK while connected through bluetooth or usb? Could you verify the issue is the same using both connection methods?

[...]

PR #1828, opened 2023-06-07
feat(core): Adding pre-release for keys that were already pressed.

andrewjrae:
There are a couple of cases where a user may end up rolling two physical keys that happen to map to the same keycode. When this happens, we register a second key press, but this does nothing at the OS level, meaning only one key press will be registered by the computer.

The two cases I'm aware of are:
 1. Rolling the key-repeat behavior (#1207)
 2. Rolling symbols that map to the same keycode (#1076)

This PR solves both of these issues by simply releasing the keycode before it is pressed again.

I've also added some test suites to cover this behavior, though they rely on the new print out when pre-releasing keys in order to be effective.

This requires further testing, currently I'm running this without issue, but more users testing on different setups is needed. I'm curious if we're going to need larger delay between the releases and presses for some setups (remote desktops typically are sensitive to this kind of thing).

### DIFF
--- a/app/src/hid_listener.c
+++ b/app/src/hid_listener.c
@@ -19,7 +19,8 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 static int hid_listener_keycode_pressed(const struct zmk_keycode_state_changed *ev) {
     int err, explicit_mods_changed, implicit_mods_changed;
 
-    if (zmk_hid_is_pressed(ZMK_HID_USAGE(ev->usage_page, ev->keycode))) {
+    if (!is_mod(ev->usage_page, ev->keycode) &&
+        zmk_hid_is_pressed(ZMK_HID_USAGE(ev->usage_page, ev->keycode))) {
         LOG_DBG("unregistering usage_page 0x%02X keycode 0x%02X since it was already pressed",
                 ev->usage_page, ev->keycode);
         err = zmk_hid_release(ZMK_HID_USAGE(ev->usage_page, ev->keycode));
--- a/app/src/hid_listener.c
+++ b/app/src/hid_listener.c
@@ -19,6 +19,20 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 static int hid_listener_keycode_pressed(const struct zmk_keycode_state_changed *ev) {
     int err, explicit_mods_changed, implicit_mods_changed;
 
+    if (zmk_hid_is_pressed(ZMK_HID_USAGE(ev->usage_page, ev->keycode))) {
+        LOG_DBG("unregistering usage_page 0x%02X keycode 0x%02X since it was already pressed",
+                ev->usage_page, ev->keycode);
+        err = zmk_hid_release(ZMK_HID_USAGE(ev->usage_page, ev->keycode));
+        if (err < 0) {
+            LOG_DBG("Unable to pre-release keycode (%d)", err);
+            return err;
+        }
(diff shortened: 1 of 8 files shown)


caksoylar:
I have been testing these changes this week and everything is working as expected. It seems to fix #1076 but I haven't tried #1207. It works well with remote desktop too (at least with clients that didn't have the issues in #759).

@andrewjrae heads up that one test `tests/modifiers/explicit/kp-lctl-dn-lctl-dn-lctl-up-lctl-up` is failing which I reproduced locally too, so it might need a look.

dlip:
I just tested this and it works well for the case the key is a regular `&kp`.
This may be a separate issue, but I had issues with combos, so if you have a combo for the same letter on both halves and you rolled or pressed them together, it would only output 1 letter.
To give you some more context as to what I'm trying to achieve, I'm implementing the [Taipo](https://inkeys.wiki/en/keymaps/taipo) layout which has 2 mirrored halves that you are meant to alternate between for each keypress. Here is my [repo](https://github.com/dlip/zmk-taipo). The only work around I've found so far is to pair both halves via bluetooth separately, but then layers/sticky keys don't work correctly.

dlip:
I opened #1899 as a separate issue for the combo problem

petejohanson:
@andrewjrae Did you see the comment from @caksoylar regarding the test failures?

andrewjrae:
I did @petejohanson, I haven't dug into it though, I was going to cover that when I did the renaming. I can try to get to this week after next, but if you're interested in pushing this through feel free to take the reigns.

vvv-ca:
@andrewjrae it looks like this pre-releases modifiers. Is this intended?
I believe the kp-lctl-dn-lctl-dn-lctl-up-lctl-up test fails because it tests rolling modifiers and errors out with "<err> zmk: Tried to unregister modifier 0 too often".

andrewjrae:
Thanks for looking @vvv-ca.
It was initially intended to pre-release any key which had been hit more than once, but looking at the code this will mess with the mod tracking that happens and we reach this issue.

Likely mods should be excluded from the pre-release code. I added another commit to prevent this from happening, and that test is now passing. It would be prettied up by a `IS_MODIFIER_KEYCODE` macro though, any thoughts on adding that somewhere @petejohanson, and if so where?

app/src/hid_listener.c
```diff
@@ -19,6 +19,22 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 static int hid_listener_keycode_pressed(const struct zmk_keycode_state_changed *ev) {
     int err, explicit_mods_changed, implicit_mods_changed;
 
+    if (!(ev->usage_page == HID_USAGE_KEY && ev->keycode >= HID_USAGE_KEY_KEYBOARD_LEFTCONTROL &&
+          ev->keycode <= HID_USAGE_KEY_KEYBOARD_RIGHT_GUI) &&
```
petejohanson, line 23: Check out `app/include/zmk/keys.h` for an `is_mod` function that should be used here (and our `hid.c` should probably be moved to using)

petejohanson (changes requested):
Just one comment on the code itself, thanks for the test on this!

app/src/hid_listener.c
andrewjrae: Ah, this is what I was looking for, thanks! I just changed this instance since `hid.c` seems to be done with the usage page by the time it checks the mod, still, would be good to refactor those instances into their own is mod func. Out of scope for this PR though.

petejohanson (approved):
LGTM!

PR #1837, opened 2023-06-13
feat(shields): Added Atreus 42

ralphie02:
## Board/Shield Check-list

### DIFF
--- a/app/boards/shields/atreus42/Kconfig.defconfig
+++ b/app/boards/shields/atreus42/Kconfig.defconfig
@@ -1,0 +1,6 @@
+if SHIELD_ATREUS42
+
+config ZMK_KEYBOARD_NAME
+	default "RAH-Atreus42-SS"
+
+endif
--- a/app/boards/shields/atreus42/Kconfig.defconfig
+++ b/app/boards/shields/atreus42/Kconfig.defconfig
@@ -4,6 +4,6 @@
 if SHIELD_ATREUS42
 
 config ZMK_KEYBOARD_NAME
-	default "Atreus42"
+    default "RAH-Atreus42-SS"
 
 endif
--- a/app/boards/shields/atreus42/Kconfig.defconfig
+++ b/app/boards/shields/atreus42/Kconfig.defconfig
@@ -1,6 +1,9 @@
+# Copyright (c) 2023 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
 if SHIELD_ATREUS42
 
 config ZMK_KEYBOARD_NAME
-	default "RAH-Atreus42-SS"
+	default "Atreus42"
 
 endif
--- a/app/boards/shields/atreus42/Kconfig.shield
+++ b/app/boards/shields/atreus42/Kconfig.shield
@@ -1,0 +1,2 @@
+config SHIELD_ATREUS42
+  def_bool $(shields_list_contains,atreus42)
--- a/app/boards/shields/atreus42/Kconfig.shield
+++ b/app/boards/shields/atreus42/Kconfig.shield
@@ -1,2 +1,5 @@
+# Copyright (c) 2023 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
 config SHIELD_ATREUS42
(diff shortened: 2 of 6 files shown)


caksoylar:
Is this a PCB generally available for sale?

A few other notes:
- while you checked the item `&pro_micro used in favor of &pro_micro_d/a if applicable` you seem to be using the `_d/_a` pins
- your url in the metadata file would ideally link to a resource that can give information about the keyboard
- tabs will need to be replaced by 4 spaces in Kconfig and keymap/overlay files

ralphie02:
@caksoylar 
Yes, there is a PCB here:  https://shop.profetkeyboards.com/product/atreus-pcb, however, I mostly just use the sockets for the switches, not the chip itself. 

Regarding your other notes:
- `&pro_micro`, I DID remove `d/a` but realized what it actually did so reverted it but forgot to untick the box. Will do that now
- Would something like this url work? https://atreus.technomancy.us/
- Will update tabs to spaces right now

caksoylar:
> Yes, there is a PCB here: https://shop.profetkeyboards.com/product/atreus-pcb, however, I mostly just use the sockets for the switches, not the chip itself.

Does this mean that the pin assignments correspond to the PCB sold in the linked shop, or what you are handwiring with the sockets (if I understand you correctly)? 

Note the item in the checklist, which means that handwires are not a good fit to be included in the ZMK repo:
> Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)

> &pro_micro, I DID remove d/a but realized what it actually did so reverted it but forgot to untick the box. Will do that now

Note that you need to be able to tick the box, i.e. use only `&pro_micro` since the `_a/_d` variants are deprecated.

> Would something like this url work? https://atreus.technomancy.us/

Sure, that looks good to me.

ralphie02:
> > Yes, there is a PCB here: https://shop.profetkeyboards.com/product/atreus-pcb, however, I mostly just use the sockets for the switches, not the chip itself.
> 
> Does this mean that the pin assignments correspond to the PCB sold in the linked shop, or what you are handwiring with the sockets (if I understand you correctly)?
> 
> Note the item in the checklist, which means that handwires are not a good fit to be included in the ZMK repo:

I must have missed this. I AM handwiring a few things for this shield so I suppose this PR would never get merged. 
> 
> > Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)
> 
> > &pro_micro, I DID remove d/a but realized what it actually did so reverted it but forgot to untick the box. Will do that now
> 
> Note that you need to be able to tick the box, i.e. use only `&pro_micro` since the `_a/_d` variants are deprecated. Was there a reason you had to revert to using them?

I didn't realize it's deprecated, the checklist didn't mention this. The build simply didn't work when I removed `_a/_d`, I'm using a nice!nano if that helps clarify things. It would be problematic for me in the future once support is removed....Should I create an issue (feature request) for this?

> 
> > Would something like this url work? https://atreus.technomancy.us/
> 
> Sure, that looks good to me.

caksoylar:
> I didn't realize it's deprecated, the checklist didn't mention this. The build simply didn't work when I removed _a/_d, I'm using a nice!nano if that helps clarify things. It would be problematic for me in the future once support is removed....Should I create an issue (feature request) for this?

`&pro_micro` is equivalent to the old `&pro_micro_d` definition, and you should also be able to replace where you used `&pro_micro_a` with the equivalent digital pin number. Below is one diagram shows both digital (Arduino) and analog pin numbers: 
image

Following above, for instance instead of `&pro_micro_a 0` you would use `&pro_micro 18` since that is the digital pin number for that. `&pro_micro_a 7` would be `&pro_micro 6`, etc.

ralphie02:
Cool. That works! Too bad this can't get merged but at least the `_a/_d` is updated. 

Will close the PR since it won't get merged anyway.

caksoylar:
Glad it worked out. Thank you for the contribution anyway, the closed PR might come in handy if anyone else ends up handwiring the same way you did.

PR #1840, opened 2023-06-19
feat: Add soft lock/off support.

petejohanson:
Initial work on a soft lock/off support for ZMK. Triggering soft off puts the device into deep sleep with only a specific GPIO pin configured to wake the device, avoiding waking from other key presses in the matrix like the normal deep sleep.


docs/docs/features/soft-on-off.md
```diff
@@ -0,0 +1,55 @@
...
+## Adding Soft On/Off to a Board
+
+### Hardware Design
+
+ZMK's soft on/off requires a dedicated GPIO pin to be used to trigger powering off, and to wake the core from the
+soft off state when it goes active again later.
+
+The simplest way to achieve this is with a push button between a GPIO pin and ground.
+
+Another, more complicate option, if you want to use a key combination in your existing keyboard matrix, is to tie two of the switch outputs in the matrix together through an AND gate and connect that to the dedicated GPIO pin.
```
caksoylar, line 29:
```suggestion
Another, more complicated option is to tie two of the switch outputs in the matrix together through an AND gate and connect that to the dedicated GPIO pin. This way you can use a key combination in your existing keyboard matrix to trigger soft on/off.
```

docs/docs/features/soft-on-off.md
```diff
@@ -0,0 +1,55 @@
...
+/ {
+  ...
+
+  soft_on_off {
+    compatible = "zmk,soft-on-off-gpio";
+
+    input-gpios = <&gpio0 6 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+        wakeup-sources = <&kscan0>;
+  };
+};
```
caksoylar, line 45:
```suggestion
/ {
    ...

    soft_on_off {
        compatible = "zmk,soft-on-off-gpio";
        input-gpios = <&gpio0 6 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
        wakeup-sources = <&kscan0>;
    };
};
```
Consistent with our new formatting guidelines

docs/docs/features/soft-on-off.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Here are the requirements for the node:
+
+- The `compatible` property for the node must be `zmk,soft-on-off-gpio`.
+- The `input-gpios` property is a phandle to the GPIO pin, with the correct pull up/down and active high/low flag set.
```
caksoylar, line 51:
```suggestion
- The `input-gpios` property is a phandle to the GPIO pin, with the correct pull up/down and active high/low flag set. In the above example the soft on/off would be triggered by pulling the pin low.
```
Might be good to elaborate like this shortly? We don't tend to document what the flags mean in general though, so we can leave it out as well.

docs/docs/features/soft-on-off.md
```diff
@@ -0,0 +1,55 @@
...
+- The `wakeup-sources` is a list of devices that needs to be shutdown before going into power off state, to ensure
+  that _only_ our soft on/off device will wake up the keyboard later. Typically this points to the `kscan` node that
+  is configured for the keyboard, which can also wake the keyboard from a "deep sleep from idle" state.
```
caksoylar, line 54:
```suggestion
  is configured for the keyboard, which can also wake the keyboard from a ["deep sleep from idle"](../config/power.md#idlesleep) state.
```

joelspadin:
Would it be possible to split this up into general soft off functions and then the driver that controls them based on a GPIO? I can think of a couple use cases which would want to trigger the same soft off behavior, but which couldn't use this GPIO code:

1. Connect a toggle switch and use code like the direct GPIO kscan driver to control on/off. This would let you add a power switch to a board with a built-in charger but without the switch being in line with the battery, so it can still be charged even when the keyboard is "off". (
2. Trigger soft off with a behavior and then use the reset switch to wake the keyboard by way of rebooting it. This would let you add soft off behavior to any keyboard that has an accessible reset switch without adding any special hardware.

petejohanson:
Closing in favor of #1942 which supersedes this.

PR #728, opened 2021-03-16
feature(split): add split-side encoder support

stephen:
This PR adds support for encoders on the peripheral side. It adds a new characteristic on the GATT service that the central subscribes to.

I only tested this on my own split Kyria.

PR #1841, opened 2023-06-19
feature(split): add support for sensors from peripheral

petejohanson:
This commit adds a new GATT characteristics on the peripheral side and wires it up to read sensor values. The central side subscribes to this new characteristics and replays sensor values on its side.

This is building off the awesome work in #728 but updated to leverage the sensor refactors recently merged in #1039


burgessa23:
working fine on Hillside 52

caksoylar:
Should we update the docs in this PR or do it in another one after merging?

Places to update:
- Intro table footnote: https://github.com/zmkfirmware/zmk/blob/main/docs/docs/intro.md?plain=1#L29, https://github.com/zmkfirmware/zmk/blob/main/docs/docs/intro.md?plain=1#L47
- Encoders feature page: https://github.com/zmkfirmware/zmk/blob/main/docs/docs/features/encoders.md?plain=1#L8

petejohanson:
> Should we update the docs in this PR or do it in another one after merging?
> 
> Places to update:
> 
>     * Intro table footnote: https://github.com/zmkfirmware/zmk/blob/main/docs/docs/intro.md?plain=1#L29, https://github.com/zmkfirmware/zmk/blob/main/docs/docs/intro.md?plain=1#L47
> 
>     * Encoders feature page: https://github.com/zmkfirmware/zmk/blob/main/docs/docs/features/encoders.md?plain=1#L8

Good catch. Updated here in a new commit.

p3likan0:
Not fully working on sofle keyboard.

My sensor binding is not getting executed on my right encoder (peripheral one) , you can check it here: https://github.com/p3likan0/zmk-firmware/blob/master/config/sofle.keymap#L33

I do have logs when moving the encoder:
```
[00:02:52.702,087] <dbg> zmk: split_central_sensor_notify_func: [SENSOR NOTIFICATION] data 0x2000f4b3 length 12
[00:02:52.702,148] <dbg> zmk: peripheral_sensor_event_work_callback: Trigger sensor change for 1
[00:02:52.702,209] <dbg> zmk: zmk_keymap_sensor_event: layer: 2 sensor_index: 1, binding name: ENC_KEY_PRESS
[00:02:52.702,270] <dbg> zmk: zmk_behavior_sensor_rotate_common_accept_data: val1: 1, val2: 1, remainder: 1/1 triggers: 0 inc keycode 0x7004E dec keycode 0x7004B
[00:02:52.702,331] <dbg> zmk: zmk_keymap_sensor_event: layer: 1 sensor_index: 1, binding name: ENC_KEY_PRESS
[00:02:52.702,392] <dbg> zmk: zmk_behavior_sensor_rotate_common_accept_data: val1: 1, val2: 1, remainder: 2/2 triggers: 0 inc keycode 0x7004E dec keycode 0x7004B
[00:02:52.702,453] <dbg> zmk: zmk_keymap_sensor_event: layer: 0 sensor_index: 1, binding name: ENC_KEY_PRESS
[00:02:52.702,514] <dbg> zmk: zmk_behavior_sensor_rotate_common_accept_data: val1: 1, val2: 1, remainder: 3/3 triggers: 0 inc keycode 0x70052 dec keycode 0x70051
[00:02:52.859,558] <dbg> zmk: split_central_sensor_notify_func: [SENSOR NOTIFICATION] data 0x2000f4b3 length 12
[00:02:52.859,649] <dbg> zmk: peripheral_sensor_event_work_callback: Trigger sensor change for 1
[00:02:52.859,710] <dbg> zmk: zmk_keymap_sensor_event: layer: 2 sensor_index: 1, binding name: ENC_KEY_PRESS
[00:02:52.859,771] <dbg> zmk: zmk_behavior_sensor_rotate_common_accept_data: val1: -1, val2: -1, remainder: 2/2 triggers: 0 inc keycode 0x7004E dec keycode 0x7004B
[00:02:52.859,832] <dbg> zmk: zmk_keymap_sensor_event: layer: 1 sensor_index: 1, binding name: ENC_KEY_PRESS
[00:02:52.859,863] <dbg> zmk: zmk_behavior_sensor_rotate_common_accept_data: val1: -1, val2: -1, remainder: 1/1 triggers: 0 inc keycode 0x7004E dec keycode 0x7004B
[00:02:52.859,924] <dbg> zmk: zmk_keymap_sensor_event: layer: 0 sensor_index: 1, binding name: ENC_KEY_PRESS
[00:02:52.859,985] <dbg> zmk: zmk_behavior_sensor_rotate_common_accept_data: val1: -1, val2: -1, remainder: 0/0 triggers: 0 inc keycode 0x70052 dec keycode 0x70051
[00:03:00.386,627] <dbg> zmk: vddh_sample_fetch: ADC raw 1220 ~ 4465 mV => 100%
```

I have zero experience with ZMK but it might even be a configuration mistake on my configuration?

p3likan0:
Working on sofle v2!

Thanks for this amazing job. I hope it arrives to master :)

mike1808:
Tested on my keyboard and it works as expected.

mike1808:
I tested on my keyboard and there is one big issue: whenever I on a layer the keyboard registers both mappings: both for the current layer and for the default layer.

PhanKhang:
Tested on SofleV2.1 with nano v2 wired (I don't have any batteries yet). Everything works. Both encoders are working.

petejohanson:
> I tested on my keyboard and there is one big issue: whenever I on a layer the keyboard registers both mappings: both for the current layer and for the default layer. 
> 
> Upd: after flashing the build from the CI it works perfectly fine. Something is wrong in my local tooling. 

Fixed, with some help from Mike on Discord.

DiogoDoreto:
Just tested it and can confirm this is working with Waterfowl after cherry-picking my config update from https://github.com/DiogoDoreto/zmk/commit/fb7c3741ea1906c7743eb6ede9e32b05ff5d23e8

Thanks for your work!

narze:
Tested on Splitkb Aurora Lily58 and it works, thank you!

larryare:
Works on my custom split keyboard flawlessly.

mike1808:
I'm curious to know what is this pr blocked on?  Does it need a review?

petejohanson:
> I'm curious to know what is this pr blocked on? Does it need a review?

Yes, the main thing remaining is review, which will happen when others on the team have bandwidth to do so.

D0hnut5:
> e: After another look this morning, applying the changes in [72ca2fa](https://github.com/zmkfirmware/zmk/commit/72ca2fa431ab3bca980a9dcfa7563078d1ea4a30) alone seems to have solved the problem. I am not sure if extracting that commit from the feature PR will break anything. Will have to see if I can rope anyone with a fully-assembled keyboard into testing the bugfix.

Per lesshonor's roping I tested this on a fully assembled Pillbug Waka. It fixed the issues I was experiencing: every third encoder detent would not send a code and the encoder would not work on layers.

nobiliana:
> I can report that this PR at least partially fixes a bug introduced in [295ed83](https://github.com/zmkfirmware/zmk/commit/295ed83409ab0bede761eb2e51f4e4670e4dbf08), after which encoder events [in this config](https://github.com/MechWild/murphpad-zmk-config/tree/da2204d7eaca7903f6af774b27b8dbb9769a679d) began triggering too often.
> 
> I still get 4 calls to [volume](https://github.com/MechWild/murphpad-zmk-config/blob/da2204d7eaca7903f6af774b27b8dbb9769a679d/config/murphpad.keymap#L72) [up or down](https://github.com/MechWild/murphpad-zmk-config/blob/da2204d7eaca7903f6af774b27b8dbb9769a679d/config/murphpad.keymap#L86) per-step if the configuration uses `resolution` ([as this shield unfortunately still does in `main`](https://github.com/zmkfirmware/zmk/blob/d17c473d45b3bd01d80b8a5eab417986be0281da/app/boards/shields/murphpad/murphpad.overlay#L37-L53)), but switching to `steps`/`triggers-per-rotation` works as expected.
> 
> I currently have no way to test the split part of this.
> 
> e: After another look this morning, applying the changes in [72ca2fa](https://github.com/zmkfirmware/zmk/commit/72ca2fa431ab3bca980a9dcfa7563078d1ea4a30) alone seems to have solved the problem. I am not sure if extracting that commit from the feature PR will break anything. Will have to see if I can rope anyone with a fully-assembled keyboard into testing the bugfix.

Also as per Lesshonor, this PR was tested on a Pillbug with bb40 shield. Encoder performance was showing 2-5 activations per detent; after reflashing with this PR, I am getting the expected 1 activation per detent (1 in 15 seem to be missed, but that could be hardware) using Steps and Trigger. 

Split functionality was not tested.

app/include/zmk/split/bluetooth/service.h
```diff
@@ -21,4 +31,7 @@ struct zmk_split_run_behavior_payload {
 } __packed;
 
 int zmk_split_bt_position_pressed(uint8_t position);
-int zmk_split_bt_position_released(uint8_t position);
\ No newline at end of file
+int zmk_split_bt_position_released(uint8_t position);
+int zmk_split_bt_sensor_triggered(uint8_t sensor_number,
```
joelspadin, line 35: Minor: `sensor_event` uses `sensor_index`, while this uses `sensor_number`. Are these the same thing? If so, making the name consistent would be good.

This also appears to be all the same data as a `sensor_event`. Could you just pass a pointer to that struct instead?

app/src/split/bluetooth/central.c
```diff
@@ -165,6 +168,46 @@ int confirm_peripheral_slot_conn(struct bt_conn *conn) {
...
+    if (!data) {
+        LOG_DBG("[UNSUBSCRIBED]");
+        params->value_handle = 0U;
+        return BT_GATT_ITER_STOP;
+    }
+
+    LOG_DBG("[SENSOR NOTIFICATION] data %p length %u", data, length);
+
+    struct sensor_event sensor_event;
+    memcpy(&sensor_event, data, MIN(length, sizeof(sensor_event)));
```
joelspadin, line 197: It would be a good idea to verify that `length >= offsetof(struct sensor_event, channel_data)`. Otherwise, it would be possible to not write anything to `channel_data_size`, and then the array size will be a garbage value which may cause other code to read past the end of the array.

Also, we are truncating the array data if given more channels than we support, but we aren't clamping `channel_data_size` accordingly. This means we're protecting against writing past the end of the array, but not reading past it.

app/src/split/bluetooth/central.c
```diff
@@ -165,6 +168,46 @@ int confirm_peripheral_slot_conn(struct bt_conn *conn) {
...
+    struct zmk_sensor_event ev = {.sensor_index = sensor_event.sensor_index,
+                                  .channel_data_size = sensor_event.channel_data_size,
+                                  .timestamp = k_uptime_get()};
+
+    memcpy(&ev.channel_data, sensor_event.channel_data,
```
joelspadin, line 202: Minor: any reason you used `&` on one array parameter but not the other? (I believe it is not necessary in either case.)

app/src/split/bluetooth/central.c
```diff
@@ -263,14 +302,34 @@ static uint8_t split_central_chrc_discovery_func(struct bt_conn *conn,
         slot->subscribe_params.value_handle = bt_gatt_attr_value_handle(attr);
         slot->subscribe_params.notify = split_central_notify_func;
         slot->subscribe_params.value = BT_GATT_CCC_NOTIFY;
-        split_central_subscribe(conn);
+        split_central_subscribe(conn, &slot->subscribe_params);
+#if ZMK_KEYMAP_HAS_SENSORS
+    } else if (bt_uuid_cmp(((struct bt_gatt_chrc *)attr->user_data)->uuid,
```
joelspadin, line 307: Since this is used repeatedly here, it would improve readability a bit to do something like

```c
const struct bt_uuid* chrc_uuid = ((const struct bt_gatt_chrc*)attr->user_data)->uuid;

if (bt_uuid_cmp(chrc_uuid, ...) == 0) {
  ...
} else if (bt_uuid_cmp(chrc_uuid, ...) == 0) {
  ...
} else if ...
```

app/src/split/bluetooth/central.c
```diff
@@ -263,14 +302,34 @@ static uint8_t split_central_chrc_discovery_func(struct bt_conn *conn,
...
+        slot->discover_params.start_handle = attr->handle + 2;
         slot->run_behavior_handle = bt_gatt_attr_value_handle(attr);
     }
 
-    bool subscribed = (slot->run_behavior_handle && slot->subscribe_params.value_handle);
+    bool subscribed = (slot->run_behavior_handle && slot->subscribe_params.value_handle
+#if ZMK_KEYMAP_HAS_SENSORS
+                       && slot->sensor_subscribe_params.value_handle
+#endif /* ZMK_KEYMAP_HAS_SENSORS */
+    );
```
joelspadin, line 332:
```suggestion
    bool subscribed = slot->run_behavior_handle && slot->subscribe_params.value_handle;
#if ZMK_KEYMAP_HAS_SENSORS
    subscribed = subscribed && slot->sensor_subscribe_params.value_handle;
#endif
```

Slightly less ugly, maybe?

joelspadin (changes requested):
This looks pretty good to me, except I think we should have better error checking when reading data into the sensor_event structs.

app/include/zmk/split/bluetooth/service.h
petejohanson: So.... `struct sensor_event` was intended as a *data exchange* format, which is why it is packed. I'm *hesitant* to have a data exchange struct also be used for public API like this. Thoughts?

app/src/split/bluetooth/central.c
petejohanson: Yeah, I like that approach better for sure. Thanks.

app/src/split/bluetooth/central.c
petejohanson: Added a check there.

app/src/split/bluetooth/central.c
petejohanson: Good catch. Fixed.

app/src/split/bluetooth/central.c
petejohanson: Much cleaner. Moved to that.

app/src/split/bluetooth/central.c
petejohanson: Done. Thanks.

joelspadin (approved):

PR #1844, opened 2023-06-20
fix(sensors): Sensor rotate behavior fix for layers.

petejohanson:
* Properly return transparent value when processing mode
  for higher disabled layers means the processing is ignored.


ReFil:
Can confirm this fixes the issue I was encountering with encoders on a BT75 PCB

petejohanson:
Minor two line fix, with confirmation. Merging.

PR #1849, opened 2023-06-25
feat(boards): M60 keyboard use back button as power switch

hyx0329:
M60 doesn't have a physical power switch. This pr makes it possible to fully power off the keyboard.

This _may_ also improve keyboard power stability.

I've been using it for a month, so far so good. Pick it if anyone need this.


hyx0329:
rebased and updated, would like to merge and requesting review

caksoylar:
It looks like this is specific to m60 since it uses the specific button on the back, not the `nrf52840_m2` board? If that's the case it looks like it should live in the shield folder, not the board.

caksoylar:
Also potential overlap with #2085.

hyx0329:
> It looks like this is specific to m60

Yes it's m60 specific. Now I've moved everything m60-specific to the shields folder and updated the code to use zephyr's gpio api.

> Also potential overlap with 2085

I don't think it's overlap with soft off as this is literally a built-in hardware switch to control SYSOFF on the PMU to disconnect battery.

From BQ24075 datasheet:
> SYSOFF: System Enable Input. Connect SYSOFF high to turn off the FET connecting the battery to the system
output. When an adapter is connected, charging is also disabled. Connect SYSOFF low for normal operation.
SYSOFF is internally pulled up to VBAT through a large resistor (approximately 5 MΩ). Do not leave SYSOFF
unconnected to ensure proper operation.

PR #1852, opened 2023-06-26
feat(docs): Clarify use of MACRO_PLACEHOLDER with modifiers

nickcoutsos:
Added a note to explain (as I understand it) that `MACRO_PLACEHOLDER` can't be combined with modifier functions due to the entire binding cell being substituted.

Would it be worth making a more full example? Maybe a `&meh` or `&hyp` macro?

### DIFF
--- a/docs/docs/behaviors/macros.md
+++ b/docs/docs/behaviors/macros.md
@@ -100,7 +100,7 @@ bindings
 Because `kp` takes one parameter, you can't simply make the second entry `<&kp>` in the `bindings` list. Whatever value you do pass in will be replaced when the macro is triggered, so you can put _any_ value there, e.g. `0`, `A` keycode, etc. To make it very obvious that the parameter there is not actually going to be used, you can use `MACRO_PLACEHOLDER` which is simply an alias for `0`.
 
 :::note
-This isn't a find-and-replace, the behavior is being triggered with the parameter supplied to the macro. To apply a modifier, for example, instead of using `&kp LS(MACRO_PLACEHOLDER)` you would need additional bindings like `<&kt LSHFT>, <&kp MACRO_PLACEHOLDER>, <&kt LSHFT>`.
+This isn't a find-and-replace, the behavior is being triggered with the parameter supplied to the macro. To apply a modifier, for example, instead of using `&kp LS(MACRO_PLACEHOLDER)` you would need additional bindings like `<&sk LSHFT>, <&macro_param_1to1>, <&kp MACRO_PLACEHOLDER>`.
 :::
 
 The available parameter controls are:
--- a/docs/docs/behaviors/macros.md
+++ b/docs/docs/behaviors/macros.md
@@ -99,6 +99,10 @@ bindings
 
 Because `kp` takes one parameter, you can't simply make the second entry `<&kp>` in the `bindings` list. Whatever value you do pass in will be replaced when the macro is triggered, so you can put _any_ value there, e.g. `0`, `A` keycode, etc. To make it very obvious that the parameter there is not actually going to be used, you can use `MACRO_PLACEHOLDER` which is simply an alias for `0`.
 
+:::note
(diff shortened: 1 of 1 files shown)


docs/docs/behaviors/macros.md
```diff
@@ -99,6 +99,10 @@ bindings
 
 Because `kp` takes one parameter, you can't simply make the second entry `<&kp>` in the `bindings` list. Whatever value you do pass in will be replaced when the macro is triggered, so you can put _any_ value there, e.g. `0`, `A` keycode, etc. To make it very obvious that the parameter there is not actually going to be used, you can use `MACRO_PLACEHOLDER` which is simply an alias for `0`.
 
+:::note
+This isn't a find-and-replace, the behavior is being triggered with the parameter supplied to the macro. To apply a modifier, for example, instead of using `&kp LS(MACRO_PLACEHOLDER)` you would need additional bindings like `<&kt LSHFT>, <&kp MACRO_PLACEHOLDER>, <&kt LSHFT>`.
```
petejohanson, line 103: I don't see how this would work w/ `&kt` here. Either show the press and release around this, or use a sticky key to achieve this, IMHO.

petejohanson (changes requested):

docs/docs/behaviors/macros.md
nickcoutsos: Ok, sticky key seems like a much simpler implementation (also forgot the control binding to select the parameter). Regarding `&kt` is the issue that the key may already be active and thus become toggled _off_ or is there something else I'm misunderstanding?

docs/docs/behaviors/macros.md
```diff
@@ -99,6 +99,10 @@ bindings
 
 Because `kp` takes one parameter, you can't simply make the second entry `<&kp>` in the `bindings` list. Whatever value you do pass in will be replaced when the macro is triggered, so you can put _any_ value there, e.g. `0`, `A` keycode, etc. To make it very obvious that the parameter there is not actually going to be used, you can use `MACRO_PLACEHOLDER` which is simply an alias for `0`.
 
+:::note
+This isn't a find-and-replace, the behavior is being triggered with the parameter supplied to the macro. To apply a modifier, for example, instead of using `&kp LS(MACRO_PLACEHOLDER)` you would need additional bindings like `<&sk LSHFT>, <&macro_param_1to1>, <&kp MACRO_PLACEHOLDER>`.
+:::
```
caksoylar, line 104:
```suggestion
:::warning
This functionality isn't implemented as a text replacement; the parameter supplied to the macro has to be exactly `MACRO_PLACEHOLDER` (or `0`) for the substitution to work as expected. For example, to apply a modifier to the parameter you cannot use `&kp LS(MACRO_PLACEHOLDER)` in the `bindings` list: this could be worked around using additional bindings like `<&sk LSHFT>, <&macro_param_1to1>, <&kp MACRO_PLACEHOLDER>`.
:::
```
I think this reads better but let me know what you think.

nickcoutsos: Wouldn't mind some confirmation, and maybe this detail would be good to add, my understanding was basically the opposite. As I see it you could have any value because the macro behavior will just ignore it based on the use of the param control bindings.

https://github.com/zmkfirmware/zmk/blob/b259d5a22e423458b94bd1aba70fd62f327528c0/app/src/behaviors/behavior_macro.c#L147-L155

caksoylar: Ah, I think you are right, nothing checks whether the parameter is `0` or not before replacing (but it would be good to have confirmation!). So it the parameter is consumed by the next binding after e.g. `&macro_param_1to1` no matter its value. 

Here is another iteration:

```suggestion
:::warning
This functionality isn't implemented like a text replacement; the corresponding parameter of the following behavior will be substituted completely by the macro parameter. For example, to apply a modifier to the parameter you cannot use `&kp LS(MACRO_PLACEHOLDER)` in the `bindings` list: this could be worked around using additional bindings like `<&sk LSHFT>, <&macro_param_1to1>, <&kp MACRO_PLACEHOLDER>`.
:::
```

PR #1859, opened 2023-07-06
feat(docs): A note about unmounting UF2 drive

filterpaper:
Add a note about the UF2 drive that automatically unmounts itself to restart, which may surprise some new users.


caksoylar (approved):
Good clarification, thanks!

PR #1867, opened 2023-07-14
fix: Disable battery sensors if reporting is off

caksoylar:
Currenly when `ZMK_BATTERY_REPORTING` is turned off, builds with battery sensors fail due to a kernel init priority missing (ex. [1](https://github.com/KLingO-MS/zmk-config-KgO-Sweep/actions/runs/5553387397/jobs/10141875080#step:9:417), [2](https://github.com/tsndr/zmk-config/actions/runs/5536781382/jobs/10104891126#step:9:400)), which I think is due to `SENSOR` Kconfig not getting selected.

Ideally the sensors wouldn't be enabled at all if battery level reporting is disabled but having them in the DT automatically enables them. This change disables them (in a bit of a roundabout way) through disabling `ZMK_BATTERY`: https://github.com/zmkfirmware/zmk/blob/bbb27ac02769c7bf5f20b76605f2b4e0f811e6e0/app/drivers/sensor/CMakeLists.txt#L4

Another option to directly disable the symbols would be to add `depends on ZMK_BATTERY_REPORTING` on each sensor, but this seemed slightly better to me. Let me know if that makes more sense.

Alternatively, maybe there is a more direct way to fix the kernel init priority warning, in which case the sensors would get compiled but not have an effect if reporting is disabled.

### DIFF
--- a/app/drivers/sensor/battery/Kconfig
+++ b/app/drivers/sensor/battery/Kconfig
@@ -6,6 +6,7 @@ DT_COMPAT_ZMK_BATTERY_VOLTAGE_DIVIDER := zmk,battery-voltage-divider
 
 config ZMK_BATTERY
     bool "ZMK battery monitoring"
+    depends on ZMK_BATTERY_REPORTING
     help
         Enable battery monitoring
 
@@ -13,7 +14,7 @@ config ZMK_BATTERY_NRF_VDDH
     bool
     default $(dt_compat_enabled,$(DT_COMPAT_ZMK_BATTERY_NRF_VDDH))
     select ADC
-    select ZMK_BATTERY
+    imply ZMK_BATTERY
     help
         Enable ZMK nRF VDDH voltage driver for battery monitoring.
 
@@ -21,6 +22,6 @@ config ZMK_BATTERY_VOLTAGE_DIVIDER
     bool
     default $(dt_compat_enabled,$(DT_COMPAT_ZMK_BATTERY_VOLTAGE_DIVIDER))
     select ADC
-    select ZMK_BATTERY
+    imply ZMK_BATTERY
     help
         Enable ZMK battery voltage divider driver for battery monitoring.
--- a/app/drivers/sensor/battery/Kconfig
+++ b/app/drivers/sensor/battery/Kconfig
@@ -6,7 +6,6 @@ DT_COMPAT_ZMK_BATTERY_VOLTAGE_DIVIDER := zmk,battery-voltage-divider
 
 config ZMK_BATTERY
     bool "ZMK battery monitoring"
-    depends on ZMK_BATTERY_REPORTING
     help
         Enable battery monitoring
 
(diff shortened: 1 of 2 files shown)


app/drivers/sensor/battery/Kconfig
```diff
@@ -6,21 +6,22 @@ DT_COMPAT_ZMK_BATTERY_VOLTAGE_DIVIDER := zmk,battery-voltage-divider
...
+    depends on ZMK_BATTERY_REPORTING
     help
         Enable battery monitoring
 
 config ZMK_BATTERY_NRF_VDDH
     bool
     default $(dt_compat_enabled,$(DT_COMPAT_ZMK_BATTERY_NRF_VDDH))
     select ADC
-    select ZMK_BATTERY
+    imply ZMK_BATTERY
```
petejohanson, line 17: This seems reasonable, but why not also make this `depends on SENSOR` as an alternative?

petejohanson (commented):
One thought on the approach chosen here.

app/drivers/sensor/battery/Kconfig
caksoylar: If I understand you correctly that sounds like one of the other alternatives I mentioned:
> Another option to directly disable the symbols would be to add depends on ZMK_BATTERY_REPORTING on each sensor, but this seemed slightly better to me. Let me know if that makes more sense.

But `depends on SENSOR` is probably better that that 😄

caksoylar: Although, in that case the battery sensors would still get compiled in if `SENSOR` is enabled through some other symbol like `ZMK_KEYMAP_SENSORS`.

caksoylar: I pushed a fixup commit that converts to that; everything builds OK but enabling sensors gets the battery sensors (but not reporting) enabled as well:
```sh
> west build -d build/kl2 -b nice_nano_v2 -- -DSHIELD=kyria_left -DCONFIG_ZMK_BATTERY_REPORTING=n -DCONFIG_EC11=y
[...]
> rg BATTERY build/kl2/zephyr/.config
87:# CONFIG_ZMK_BATTERY_REPORTING is not set
238:CONFIG_DT_HAS_ZMK_BATTERY_NRF_VDDH_ENABLED=y
481:CONFIG_ZMK_BATTERY=y
482:CONFIG_ZMK_BATTERY_NRF_VDDH=y
```

petejohanson: I think that's expected behavior, if the nodes are enabled, and the necessary support Kconfig symbols are enabled, we auto-enable the driver. You'd either disable the node in the DTS, or disable it explicitly in the Kconfig as well.

caksoylar: Got it, does the current state look good in that case?

petejohanson: Yes, I believe this is correct now. Thanks!

petejohanson (approved):
Thanks!

PR #1870, opened 2023-07-15
refactor: use low priority workqueue for underglow and battery reporting

xudongzheng:
Blocking operations on the high priority system workqueue may result in deadlocks, particularly when Bluetooth is in use.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -504,6 +504,14 @@ config ZMK_BATTERY_REPORT_INTERVAL
     int "Battery level report interval in seconds"
     default 60
 
+config ZMK_LOW_PRIORITY_THREAD_STACK_SIZE
+    int "Low priority thread stack size"
+    default 768
+
+config ZMK_LOW_PRIORITY_THREAD_PRIORITY
+    int "Low priority thread priority"
+    default 10
+
 #Advanced
 endmenu
 
--- a/app/include/zmk/workqueue.h
+++ b/app/include/zmk/workqueue.h
@@ -1,0 +1,1 @@
+struct k_work_q *zmk_workqueue_lowprio_work_q();
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -18,6 +18,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/event_manager.h>
 #include <zmk/battery.h>
 #include <zmk/events/battery_state_changed.h>
+#include <zmk/workqueue.h>
 
 static uint8_t last_state_of_charge = 0;
 
@@ -77,7 +78,9 @@ static void zmk_battery_work(struct k_work *work) {
 
 K_WORK_DEFINE(battery_work, zmk_battery_work);
 
-static void zmk_battery_timer(struct k_timer *timer) { k_work_submit(&battery_work); }
+static void zmk_battery_timer(struct k_timer *timer) {
+    k_work_submit_to_queue(zmk_workqueue_lowprio_work_q(), &battery_work);
+}
 
 K_TIMER_DEFINE(battery_timer, zmk_battery_timer, NULL);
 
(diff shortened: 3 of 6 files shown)


app/Kconfig
```diff
@@ -133,6 +133,14 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
     default 4096 if SOC_RP2040
     default 2048
 
+config ZMK_LOW_PRIORITY_THREAD_STACK_SIZE
+    int "Low priority thread stack size"
+    default 768
+
+config ZMK_LOW_PRIORITY_THREAD_PRIORITY
+    int "BLE notify thread priority"
```
caksoylar, line 141:
```suggestion
    int "Low priority thread priority"
```
A more generic description perhaps?

app/src/rgb_underglow.c
```diff
@@ -203,7 +203,7 @@ static void zmk_rgb_underglow_tick_handler(struct k_timer *timer) {
         return;
     }
 
-    k_work_submit(&underglow_work);
+    k_work_submit_to_queue(&lowprio_work_q, &underglow_work);
```
caksoylar, line 206: I think this needs a `#include <zmk/workqueue.h>` as well, doesn't it?

xudongzheng:
@caksoylar Thanks for the suggestions, both have been addressed.

I've also confirmed that the underglow part works on RP2040.

app/Kconfig
```diff
@@ -118,6 +118,14 @@ menuconfig ZMK_BLE
...
     imply ZMK_BATTERY_REPORTING
 
+config ZMK_LOW_PRIORITY_THREAD_STACK_SIZE
+    int "Low priority thread stack size"
+    default 768
+
+config ZMK_LOW_PRIORITY_THREAD_PRIORITY
+    int "Low priority thread priority"
+    default 10
+
```
petejohanson, line 128: This config is all reasonable, but I don't love *where* you're dropping it in this file. I would nest this under the "Advanced" menu somewhere.

app/include/zmk/workqueue.h
```diff
@@ -0,0 +1 @@
+extern struct k_work_q lowprio_work_q;
```
petejohanson, line 1: I don't *love* exposing this as an `extern` here. I'd vote for either a function (that would get inlined probably) or a C-macro/define at least.

In particular, if we have a space constrained device (RAM wise) we might want to just-reuse the system workqueue instead of adding a second queue, and using the define/function will let use swap those out w/o the consumer caring as much.

Thoughts?

app/src/battery.c
```diff
@@ -97,14 +100,7 @@ static int zmk_battery_init(const struct device *_arg) {
...
 
-    int rc = zmk_battery_update(battery);
-
-    if (rc != 0) {
-        LOG_DBG("Failed to update battery value: %d.", rc);
-        return rc;
-    }
-
-    k_timer_start(&battery_timer, K_MINUTES(1), K_SECONDS(CONFIG_ZMK_BATTERY_REPORT_INTERVAL));
+    k_timer_start(&battery_timer, K_NO_WAIT, K_SECONDS(CONFIG_ZMK_BATTERY_REPORT_INTERVAL));
```
petejohanson, line 103: :+1:

petejohanson (changes requested):
A couple thoughts on the implementation. Thanks for working on this!

app/include/zmk/workqueue.h
xudongzheng: I've introduced `zmk_lowprio_work_q()` but generally I'm hesitant to make things like this configurable unless if there's a good concrete reason - ie 1KB is the difference between fitting and not fitting on a chip.

Furthermore currently the purpose of the low priority workqueue is not well defined. It may end up being used for other tasks that must run with a lower priority than the system workqueue on low resource MCUs, similar to how battery notification is for BLE.

petejohanson:
Code looks reasonable, will test locally soon.

xudongzheng:
A separate commit has been added for turning underglow off using the low priority work queue as well.

Interestingly I observed quite a bit of LED flickering on RP2040 when `led_strip_update_rgb()` is called from the system workqueue. It's practically non-existent while running on the low priority workqueue.

app/include/zmk/workqueue.h
```diff
@@ -0,0 +1 @@
+struct k_work_q *zmk_lowprio_work_q();
```
petejohanson, line 1: Functionality tested w/o issue, but as I peek at this code one more time, it seems this naming should be adjusted to be "namespaced" to match the header, e.g.:

```suggestion
struct k_work_q *zmk_workqueue_lowprio_work_q();
```

petejohanson (approved):
LGTM. @Nicell Any concerns with the underglow bits?

Nicell (approved):
No concerns about the underglow parts, looks good!

PR #1874, opened 2023-07-17
feat(boards): Add the Pico 87 Keyboard

flimberger:
Add the Bolt Industries Pico 87 keyboard, a DIY tenkeyless keyboard using the Raspberry Pi Pico.

Since there is no common connector defined for the board, the GPIO bank is referenced directly as &gpio0 in the device tree overlay.

The board is available here: https://www.boltind.com/product/bolt-industries-pico-87-mechanical-keyboard-pcbs-and-hardware-kit/

## Board/Shield Check-list

### DIFF
--- a/app/boards/shields/pico87/pico87.zmk.yml
+++ b/app/boards/shields/pico87/pico87.zmk.yml
@@ -1,0 +1,8 @@
+file_format: "1"
+id: pico87
+name: Pico 87
+type: shield
+url: https://github.com/mountainmannm/Bolt-Industries/
+requires: [rpi_pico]
+features:
+  - keys
--- a/app/boards/arm/rpi_pico/rpi_pico.zmk.yml
+++ b/app/boards/arm/rpi_pico/rpi_pico.zmk.yml
@@ -1,0 +1,9 @@
+file_format: "1"
+id: rpi_pico
+name: Raspberry Pi Pico
+type: board
+arch: arm
+outputs:
+  - usb
+url: https://www.raspberrypi.com/products/raspberry-pi-pico
+exposes: [rpi_pico]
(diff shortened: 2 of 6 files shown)


app/boards/shields/pico87/pico87.zmk.yml
```diff
@@ -0,0 +1,8 @@
+file_format: "1"
+id: pico87
+name: Pico 87
+type: shield
+url: https://github.com/mountainmannm/Bolt-Industries/
+requires: [rpi_pico]
```
caksoylar, line 6: We currently don't have metadata for the `rpi_pico` board that would have `exposes: [rpi_pico]` that this would be compatible with. Not sure what's the exact solution here, do we add a `app/boards/arm/rpi_pico/rpi_pico.zmk.yml`? cc: @petejohanson

petejohanson: > We currently don't have metadata for the `rpi_pico` board that would have `exposes: [rpi_pico]` that this would be compatible with. Not sure what's the exact solution here, do we add a `app/boards/arm/rpi_pico/rpi_pico.zmk.yml`? cc: @petejohanson

Yes, we probably do need to add metadata file there. I need to double check, the *exposes* was intended for the generic "interconnect" concept. Originally the `depends` was consider to work with either a board ID or an interconnect ID, I'm not sure we fully fleshed that out or not though.

Ideally we should be able to just depend on the *board ID* to start, and not create some interconnect until we actually have/need it. No one has *yet* created enough RPi Pico drop in replacements it's necessary at this point.

flimberger: I added `app/boards/arm/rpi_pico/rpi_pico.zmk.yml`, which basically looks like `app/boards/arm/adafruit_kb2040/adafruit_kb2040.zmk.yml`.

Nick-Munnich:

PR #1875, opened 2023-07-17
Add BSIM based BLE tests

petejohanson:
In preparation for some major refactors of our BLE code to use [State Machine Framework](https://docs.zephyrproject.org/latest/services/smf/index.html), this PR adds some core integration tests for our BLE code.

* Like our behavior tests, we build for a posix target, but this one using the [BabbleSim board](https://docs.zephyrproject.org/3.4.0/boards/posix/nrf52_bsim/doc/index.html#nrf52-bsim) that uses [BabbleSim](https://babblesim.github.io/) to provide a simulated 2.4GHz PHY.
* We have a simple posix/BSIM BLE/HoG client, with a few command line flags that can alter it's behavior a bit (e.g. `-halt_after_bonding` or `-wait_on_start=1300`) so we can simulate centrals/hosts connecting to our ZMK keyboard and performing disconnects, bonding, etc.
* Like behavior tests, our ZMK posix peripheral executable uses mock KSCAN to trigger behaviors, e.g. press/release of keycodes, or switching/clearing profiles.
* The output of the central(s) is captured using pattern matching and then the filtered output is used as a snapshot to compare with the known good snapshot. That allows us to verify *what clients see* when interacting with ZMK, and verify that doesn't regress/change when we refactor the ZMK side BLE code.
* Add a new GHA job to run these tests.

I've layers my WIP BLE changes for SMF onto this for testing, and it properly caught a regression there, and then I have them *still passing* with my changes at this point.

This is a baseline set of 5 tests, but would like to supplement with more as we go.

Thoughts?


app/west.yml
```diff
@@ -30,5 +38,88 @@ manifest:
           - edtt
           - trusted-firmware-m
           - sof
+    - name: bsim
```
Nicell, line 41: Will all of these need to be installed now when people build locally or in the cloud? I can imagine this might slow things down.

Nicell (commented):
Basic cleanup and a question. This is really cool! I'll need to look closer at how the `main.c` works, but I'm really impressed with the test run on this PR.

I noticed in the logs from the test run that the start, test logs, and pass/fail logging to GitHub actions are all jumbled up between different tests. It'd be nice if they could be grouped for better readability.

app/tests/ble/central/src/main.c
```diff
@@ -0,0 +1,357 @@
+/* main.c - Application main entry point */
+
+/*
+ * Copyright (c) 2015-2016 Intel Corporation
```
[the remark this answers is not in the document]
petejohanson: So, this was a *sample* from the Zephyr project, that I copied in and modified heavily. I think this should be retained, and we'll add our standard header on the next line.

app/tests/ble/central/prj.conf
```diff
@@ -0,0 +1,8 @@
+CONFIG_BT=y
+CONFIG_LOG=y
+CONFIG_BOOT_BANNER=n
+# CONFIG_BT_DEBUG_LOG=y
```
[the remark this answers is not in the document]
petejohanson: Yup. Removed.

app/tests/ble/central/src/main.c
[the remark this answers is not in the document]
petejohanson: Added.

app/west.yml
petejohanson: If you look, the `group` for all of these is `babblesim`, and near the top of the file:

```
group-filter: [-babblesim]
```

Which disables it by default, and you can use a command line to enable it so `west update` pulls it in.

petejohanson:
> Basic cleanup and a question. This is really cool! I'll need to look closer at how the `main.c` works, but I'm really impressed with the test run on this PR.
> 
> I noticed in the logs from the test run that the start, test logs, and pass/fail logging to GitHub actions are all jumbled up between different tests. It'd be nice if they could be grouped for better readability.

@Nicell Ok, I did some clean up so the verbose output will only show when running a single test during development, otherwise it's quieter. See https://github.com/zmkfirmware/zmk/actions/runs/5988986557/job/16244737224?pr=1875#step:10:1 for example.

Thoughts?

app/run-ble-test.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
joelspadin, line 1: I usually use Python or PowerShell for this type of stuff, so Bash is a bit of a foreign language to me... I'm not going to review this in detail unless you want me to try.

app/tests/ble/central/src/main.c
```diff
@@ -0,0 +1,370 @@
...
+         .option = "halt_after_bonding",
+         .type = 'b',
+         .dest = (void *)&halt_after_bonding,
+         .descript = "Halt any further logic after bonding the first time"},
+        {.is_switch = true,
+         .option = "clear_bond_on_disconnect",
+         .type = 'b',
+         .dest = (void *)&clear_bond_on_disconnect,
+         .descript = "Clear bonds on disconnect and reconnect"},
+        {// .is_switch = true,
```
joelspadin, line 62: Is this supposed to be commented out?

app/tests/ble/central/src/main.c
```diff
@@ -0,0 +1,370 @@
...
+         .type = 'u',
+         .dest = (void *)&wait_on_start,
+         .descript = "Time in milliseconds to wait before starting the test process"},
+        ARG_TABLE_ENDMARKER};
+
+    native_add_command_line_opts(options);
+}
+
+NATIVE_TASK(ble_central_native_posix_options, PRE_BOOT_1, 1);
+// NATIVE_TASK(flash_native_posix_cleanup, ON_EXIT, 1);
```
joelspadin, line 74: Unused code?

app/tests/ble/central/src/main.c
```diff
@@ -0,0 +1,370 @@
...
+
+    return BT_GATT_ITER_CONTINUE;
+}
+
+static uint8_t discover_func(struct bt_conn *conn, const struct bt_gatt_attr *attr,
+                             struct bt_gatt_discover_params *params) {
+    int err;
+
+    if (!attr) {
+        LOG_DBG("Discover complete");
```
joelspadin, line 104: Many of the other messages are surrounded with `[]`, which I assume is to make them easier to pick out of logs. However, some of them like this one aren't. Should all of them be formatted like that, or is there a difference between the two types of messages?

app/tests/ble/central/src/main.c
```diff
@@ -0,0 +1,370 @@
...
+
+    switch (data->type) {
+    case BT_DATA_UUID16_SOME:
+    case BT_DATA_UUID16_ALL:
+        if (data->data_len % sizeof(uint16_t) != 0U) {
+            LOG_DBG("AD malformed");
+            return true;
+        }
+
+        for (i = 0; i < data->data_len; i += sizeof(uint16_t)) {
```
joelspadin, line 179: This can read past the end of the data unless we verify that data_len is a multiple of sizeof(uint16_t). Consider `data_len == 1`. On the first iteration, `0 < 1` so we enter the loop, and then we memcpy 2 bytes out of an array with only 1 byte.

app/tests/ble/central/src/main.c
```diff
@@ -0,0 +1,370 @@
...
+    if (clear_bond_on_disconnect) {
+        bt_unpair(BT_ID_DEFAULT, bt_conn_get_dst(default_conn));
+    }
+}
+
+static K_WORK_DELAYABLE_DEFINE(disconnect_work, do_disconnect_of_active);
+
+static void security_changed(struct bt_conn *conn, bt_security_t level, enum bt_security_err err) {
+    if (err > BT_SECURITY_ERR_SUCCESS) {
+        exit(1);
```
joelspadin, line 306: Should we log something on security failure?

app/tests/ble/profiles/bond-clear-then-bond-second-client/snapshot.log
```diff
@@ -0,0 +1,33 @@
+profile 0 [1;33m<wrn> bt_id: No static addresses stored in controller[0m
```
joelspadin, line 1: Are there any log settings we can use to disable colorization so we don't have escape sequences in the logs?

petejohanson: Yeah, finally found it, the new central app didn't have the log color setting turned off. Fixed.

app/tests/ble/central/src/main.c
petejohanson: No difference... I'll update them for consistency, but it's not really of significance.

app/tests/ble/central/src/main.c
petejohanson: To be honest... Given we're only using this for tests, I'm fine with the simplicity this affords us and if we crash later, I'll bang my head over it then.

app/tests/ble/central/src/main.c
petejohanson: Modified to be more consistent.

app/tests/ble/central/src/main.c
petejohanson: Added

.github/workflows/test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         id: test-dirs
         run: |
           cd app/tests/
-          export TESTS=$(ls -d * | jq -R -s -c 'split("\n")[:-1]')
+          export TESTS=$(ls -d * | grep -v ble | jq -R -s -c 'split("\n")[:-1]')
```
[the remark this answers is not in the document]
petejohanson: Yeah... That's what I'd like to avoid. Not sure I want to pull that trigger *now*. I think we do it whenever we actually have one more added that gives us issues. Any concern?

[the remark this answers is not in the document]
joelspadin: Nope. If we start adding more folders that need to be excluded later, then we can come up with something.

app/tests/ble/central/src/main.c
joelspadin: I see now that there is a check for the data length. If that was always there and you didn't just add that, then I must have just skimmed over it.

caksoylar (approved):
LGTM for the parts I reviewed. It doesn't look like it needs any docs updates, does it?

petejohanson:
> LGTM for the parts I reviewed. It doesn't look like it needs any docs updates, does it?

It could use some docs on how to add the tests, but nothing that's end user affecting that I think is a must have for this PR.

PR #1876, opened 2023-07-17
docs(readme): improve writing

trueNAHO:
- docs(readme): add missing punctuations
- docs(readme): add link to MIT license


trueNAHO:
Should all three commits be merged into a single one to remove unnecessarily small commits?

caksoylar:
Can you actually drop the last commit? In my opinion it doesn't make a material improvement and reads worse in some cases. Don't worry about squashing, we can squash it while merging.

trueNAHO:
> Don't worry about squashing, we can squash it while merging.

In that case, I simply reverted the last commit.

caksoylar (approved):

PR #1879, opened 2023-07-19
Various settings reset improvements.

petejohanson:
* When the clear bonds Kconfig is set, also clear peripheral
  address slots addresses from settings as well.

fix(shields): Make settings_reset more flexible.
    
    * Don't reference `pro_micro` nexus node in settings_reset
      so it can be used with other controllers.
    * Use mock kscan node instead.

fix(bluetooth): Split improvements
    
    * Proper usage of bt_uuid_cmp.
    * Central's don't start scanning for peripherals if
      `ZMK_BLE_CLEAR_BONDS_ON_START` is enabled.
    * Split peripherals don't advertize if
      `ZMK_BLE_CLEAR_BONDS_ON_START` is enabled.

fix(bluetooth): Corrected use of `bt_addr_le_cmp`
    
    * Properly compare to zero when comparing LE addresses.

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -606,7 +606,7 @@ static int zmk_ble_init(const struct device *_arg) {
 
     bt_unpair(BT_ID_DEFAULT, NULL);
 
-    for (int i = 0; i < ZMK_BLE_PROFILE_COUNT; i++) {
+    for (int i = 0; i < 8; i++) {
         char setting_name[15];
         sprintf(setting_name, "ble/profiles/%d", i);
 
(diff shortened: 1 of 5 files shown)


app/src/ble.c
```diff
@@ -615,7 +615,21 @@ static int zmk_ble_init(const struct device *_arg) {
             LOG_ERR("Failed to delete setting: %d", err);
         }
     }
-#endif
+
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
```
xudongzheng, line 619: The preceding `ble/profiles/%d` should also only exist on central if I'm not mistaken.

For both, I'd also consider looping to some fixed integer (such as 16) and removing this `#if IS_ENABLED` entirely. This way the same reset file can be used for both halves and there's no orphaned settings if user decides to swap the controllers for central and peripheral.

petejohanson: Oh, the peripherals only use `peripheral.c` which already has different bond clearing on the start, so really the checks are *right* since `ble.c` is only run on non-splits or split centrals.

petejohanson: I'm not sure it's worth disrupting that much just to clear unused/dangling settings in the swap scenario.

xudongzheng: Ah yes, forgot about the fact that there are non-split keyboards.

petejohanson: Actually, moved to your suggestion, since the settings_reset shield isn't built as a split at all.

xudongzheng (approved):

PR #1881, opened 2023-07-20
fix(battery): change nRF52 ADC gain

xudongzheng:
With ADC pin, maximum 3.6V input needs gain 1/6 to be less than the 0.6V reference.

With VDDHDIV5, maximum 6V input corresponds to 1.2V so gain 1/2 will be less than the 0.6V reference and be slightly more precise.


petejohanson (approved):
LGTM, but I'd like @Nicell to weigh in as well since he's contributed most of the battery sensor work.

Nicell (approved):
I set the gain to 1/5th based on the nano's divider outputting 3V max, but it's fair that others may try to use the full 3.6V range available to them, so that makes sense to change. Good catch on the VDDHDIV5 gain change, that was just duplicated from the original ADC code probably. I haven't gotten the chance to test this on real hardware, but I don't see why this should cause any issues.

PR #1885, opened 2023-07-24
refactor(underglow): allow full range for initial brightness

xudongzheng:
The brightness is already scaled within zmk_rgb_underglow_tick(). With the current range condition, setting max brightness to 20% results in starting with 4% (20% of 20%) brightness.


caksoylar:
From your description this is a bug fix rather than a refactor, right?

xudongzheng:
> From your description this is a bug fix rather than a refactor, right?

Yes, I suppose that would be more accurate.

PR #1886, opened 2023-07-26
refactor: rearrange Kconfig inclusion order

xudongzheng:
This makes it possible to override ZMK Kconfig options from shields.

This is useful for options such as `ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS`, `BT_MAX_PAIRED`, `BT_MAX_CONN`, among others. Defining these in the shield `Kconfig.defconfig` works better than in the `.conf` since the they can be overridden from the user ZMK config repository.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -1,6 +1,13 @@
 # Copyright (c) 2020 The ZMK Contributors
 # SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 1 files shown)


caksoylar:
I believe this is obsoleted by #2537.

Nick-Munnich:
Agree, should be obsolete.

PR #1890, opened 2023-08-03
feat(shields): Add Bluetooth bindings to kyria keymaps

filterpaper:
Bluetooth bindings are useful for handling pairings with hosts. This change adds the header and a few default commands as template for new users to work with.


caksoylar:
I think it might make sense to include profile 0 in the mappings, if not all 5 of them.

filterpaper:
You are right, my mistake for not starting the profile at zero. All three keymaps are now updated.

caksoylar (approved):

PR #1892, opened 2023-08-05
Glove80 Board Definition

chrisandreae:
Adds board definition for MoErgo Glove80.

## Board/Shield Check-list
 - [X] Kconfig.defconfig file correctly wraps *all* configuration in conditional on the shield symbol
 - [X] Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)


app/boards/arm/glove80/glove80_rh.dts
```diff
@@ -0,0 +1,116 @@
...
+        full-ohms = <(1500000 + 820000)>;
+    };
+
+    vbatt-vddh {
+        compatible = "zmk,battery-nrf-vddh";
+        label = "BATTERY-VDDH";
+    };
+};
+
+&spi1 {
```
petejohanson, line 53: This should be moved to SPI3, see #1746 for the details and why-for.

app/boards/arm/glove80/glove80_rh.dts
```diff
@@ -0,0 +1,116 @@
...
+        label = "BATTERY";
+        io-channels = <&adc 5>;
+        output-ohms = <820000>;
+        full-ohms = <(1500000 + 820000)>;
+    };
+
+    vbatt-vddh {
+        compatible = "zmk,battery-nrf-vddh";
+        label = "BATTERY-VDDH";
+    };
```
petejohanson, line 50: I'm really confused... Are both actually present on the hardware? One or the other really should be disabled if not actually used for anything, or removed entirely.

app/boards/arm/glove80/glove80_rh-pinctrl.dtsi
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+&pinctrl {
+    spi1_default: spi1_default {
+        group1 {
+            psels = <NRF_PSEL(SPIM_MOSI, 0, 13)>, // WS2812_VEXT_DATA
+            <NRF_PSEL(SPIM_SCK, 0, 27)>,          // no connection
+            <NRF_PSEL(SPIM_MISO, 0, 14)>;         // no connection
+        };
+    };
```
petejohanson, line 13: See my comment about moving to SPI3 and removing the bogus pin selections here.

app/boards/arm/glove80/glove80_lh.dts
```diff
@@ -0,0 +1,109 @@
...
+        full-ohms = <(1500000 + 820000)>;
+    };
+
+    vbatt-vddh {
+        compatible = "zmk,battery-nrf-vddh";
+        label = "BATTERY-VDDH";
+    };
+};
+
+&spi1 {
```
petejohanson, line 52: See comment about SPI3

app/boards/arm/glove80/glove80_lh.dts
```diff
@@ -0,0 +1,109 @@
...
+        label = "BATTERY";
+        io-channels = <&adc 5>;
+        output-ohms = <820000>;
+        full-ohms = <(1500000 + 820000)>;
+    };
+
+    vbatt-vddh {
+        compatible = "zmk,battery-nrf-vddh";
+        label = "BATTERY-VDDH";
+    };
```
petejohanson, line 49: Same comment.

app/boards/arm/glove80/Kconfig.defconfig
```diff
@@ -0,0 +1,80 @@
...
+
+config ZMK_USB
+    default y
+
+config ZMK_BATTERY_VOLTAGE_DIVIDER
+    default y
+
+config ZMK_BATTERY_NRF_VDDH
+    default y
+
```
petejohanson, line 43: Again, not sure why both are needed.

app/boards/arm/glove80/Kconfig.defconfig
```diff
@@ -0,0 +1,80 @@
...
+if BOARD_GLOVE80_LH
+
+config BOARD
+    default "glove80 lh"
+
+config ZMK_SPLIT_BLE_ROLE_CENTRAL
+    default y
+
+config ZMK_KEYBOARD_NAME
+    default "Glove80"
```
petejohanson, line 13: This gets overridden in the left/right _defconfig files.. Why bother here?

app/boards/arm/glove80/glove80_rh_defconfig
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) 2021 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+CONFIG_SOC_SERIES_NRF52X=y
+CONFIG_SOC_NRF52840_QIAA=y
+CONFIG_BOARD_GLOVE80_RH=y
+
+# Keyboard IDs
+CONFIG_ZMK_KEYBOARD_NAME="Glove80 Right"
```
petejohanson, line 9: Doesn't this accidentally cause the right hand to advertise in a way that's visible to hosts? Seems like this would invite accidental disconnects. If you just need this for USB, set just the USB Kconfig you need?

petejohanson (commented):
A few comments/questions on the shield.

app/boards/arm/glove80/glove80_rh_defconfig
chrisandreae: Empirically, it doesn't show up for either MacOS or Windows. But also that's definitely something we'd want to avoid. Do you recommend removing it and using only `CONFIG_USB_DEVICE_PRODUCT` for future-compatibility purposes?

The intent was for the name to be defined when scanning for Bluetooth devices using lower level tools like nRF Connect.

app/boards/arm/glove80/Kconfig.defconfig
chrisandreae: Fair, I'll remove it.

app/boards/arm/glove80/glove80_rh.dts
chrisandreae: This was an oversight; both are present on the hardware, so we left both defined and only one selected with `chosen`. I'll experiment a little and then go with whichever is more accurate.

app/boards/arm/glove80/glove80_rh.dts
chrisandreae: Thanks, I'll make that change.

app/boards/arm/glove80/glove80_rh_defconfig
chrisandreae: Out of band:
> As long as the lack of other GAP attributes or advertised services is enough for hosts not to show it still, fine to leave it.

petejohanson (approved):
Code looks good to me. Please 100% confirm the underglow was tested after the SPI3 refactor, then good to me for merge.

app/boards/arm/glove80/Kconfig.defconfig
```diff
@@ -0,0 +1,71 @@
...
+
+config BT_CTLR
+    default BT
+
+config ZMK_BLE
+    default y
+
+config ZMK_USB
+    default y
+
```
petejohanson, line 34: These will need to be moved to the `<board>_defconfig` files after the #1140 merge. See #2006 for the details.

petejohanson (changes requested):

petejohanson (approved):
LGTM. Thanks.

piotr-dobrogost:
Out of curiosity; why isn't the board definition maintained here by the producer itself?

petejohanson:
> Out of curiosity; why isn't the board definition maintained here by the producer itself?

@chrisandreae is working on behalf of MoErgo... So, it is maintained by the producer? Unless I don't understand your question.

piotr-dobrogost:
@petejohanson Thanks, I didn't know that.

manna-harbour:
> Definitions follow the general style of other shields/boards upstream

Why is this using lh/rh instead of left/right?

chrisandreae:
That's not a convention I was aware of. I have no particular objection to it being changed, but it would impact any existing users.

PR #1895, opened 2023-08-08
fix(docs): detail overriding the `led_strip` `chain-length` property

SethMilliken:
Quite a few people have struggled with setting the correct `chain-length` for their `led_strip`. For some, this is their first time needing to change a pre-defined devicetree property, and so they aren't familiar with the technique. I commonly see folks adding a duplicate of the entire `*.overlay` file with only the `chain-length` value changed. Having clear documentation for this specific application of the property override technique could help forestall these problems, and give those of us helping out in Discord something to easily reference.

### DIFF
--- a/docs/docs/features/underglow.md
+++ b/docs/docs/features/underglow.md
@@ -39,6 +39,20 @@ use Kconfig.
 
 If your board or shield does not have RGB underglow configured, refer to [Adding RGB Underglow to a Board](#adding-rgb-underglow-to-a-board).
 
+### Modifying the number of LEDs
+
+A common issue when enabling underglow is that some of the installed LEDs do not illuminate. This can happen when a board's default underglow configuration accounts only for either the downward facing LEDs or the upward facing LEDs under each key. On a split keyboard, a good sign that this may be the problem is that the unilluminated LEDs on each half are symmetrical.
+
+The number of underglow LEDs is controlled by the `chain-length` property in the `led_strip` node. You can [change the value of this property](../config/index.md#changing-devicetree-properties) in the `<keyboard>.keymap` file by adding a stanza like this one outside of any other node (i.e. above or below the `/` node):
+
+```
+&led_strip {
+    chain-length = <21>;
+};
+```
+
+where the value is the total count of LEDs (per half, for split keyboards).
+
 ## Configuring RGB Underglow
 
 See [RGB underglow configuration](/docs/config/underglow).


SethMilliken:
Edit: removed trailing whitespace `prettier` complained about.

docs/docs/features/underglow.md
```diff
@@ -162,3 +162,15 @@ CONFIG_ZMK_RGB_UNDERGLOW=y
 # Use the STRIP config specific to the LEDs you're using
 CONFIG_WS2812_STRIP=y
 ```
+
+## Change `chain-length` Only
+
+A common need is to change only the `chain-length` value for a `led_strip` node that is otherwise fully configured in the board definition, often because the default value accounts only for either the downward facing LEDs or the upward facing LEDs under each key. This can be done by [changing that property](https://zmk.dev/docs/config#changing-devicetree-properties) in the `*.keymap` file by adding a stanza like this one outside of any other node (i.e. above or below the `/` node):
```
caksoylar, line 168:
```suggestion
A common need is to change only the value of the `chain-length` property for a `led_strip` node that is otherwise fully configured in the board definition, often because the default value accounts only for either the downward facing LEDs or the upward facing LEDs under each key. This can be done by [changing that property](../config/index.md#changing-devicetree-properties) in the `<keyboard>.keymap` file by adding a stanza like this one outside of any other node (i.e. above or below the `/` node):
```
(We standardized on `.md` links, but please double check the link works after the change as I don't have a dev environment right now)

docs/docs/features/underglow.md
```diff
@@ -162,3 +162,15 @@ CONFIG_ZMK_RGB_UNDERGLOW=y
 # Use the STRIP config specific to the LEDs you're using
 CONFIG_WS2812_STRIP=y
 ```
+
+## Change `chain-length` Only
```
caksoylar, line 166:
```suggestion
## Modifying the number of LEDs
```

caksoylar (commented):
Overall looks good, thank you! It makes more sense to me to have this under https://zmk.dev/docs/features/underglow#enabling-rgb-underglow, since this is geared towards the end-user rather than the keyboard config author. What do you think?

docs/docs/features/underglow.md
SethMilliken: Your suggestions for changing the name and location of the section prompted me to reconsider the language of the body as well. It seems like it ought to lead either with the objective we're attempting to reach or the symptoms of the problem we're trying to solve, rather than a summary of the solution. How about this:

## Modifying the number of LEDs

A common issue when enabling underglow is that some of the installed LEDs do not illuminate. This can happen when a board's default underglow configuation accounts only for either the downward facing LEDs or the upward facing LEDs under each key. On a split keyboard, a good sign that this may be the problem is that the unilluminated LEDs on each half are symmetrical.

The number of underglow LEDs is controlled by the `chain-length` property in the `led_strip` node. You can [change the value of this property](../config/index.md#changing-devicetree-properties) in the `<keyboard>.keymap` file by adding a stanza like this one outside of any other node (i.e. above or below the `/` node):
...

caksoylar: Sure, looks good to me!

SethMilliken:
Updated language. Incorporated suggestions. Successfully tested relative link. Moved to be subsection of https://zmk.dev/docs/features/underglow#enabling-rgb-underglow

caksoylar (approved):
Thanks!

PR #1908, opened 2023-08-29
Update leeloo_common.dtsi

mstralenya:
Added status "disabled" by default for Leeloo to avoid triggering of all encoder actions


petejohanson (approved):
LGTM. Good catch.

PR #1912, opened 2023-09-02
fix(split): reserve peripheral slot before stopping scanning

xudongzheng:
In the event that the peripheral MAC address does not match, this allows scanning to continue.

### DIFF
--- a/app/src/split/bluetooth/central.c
+++ b/app/src/split/bluetooth/central.c
@@ -433,19 +433,21 @@ static int stop_scanning() {
 static bool split_central_eir_found(const bt_addr_le_t *addr) {
     LOG_DBG("Found the split service");
 
+    // Reserve peripheral slot. Once the central has bonded to its peripherals,
+    // the peripheral MAC addresses will be validated internally and the slot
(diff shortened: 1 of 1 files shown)


app/src/split/bluetooth/central.c
```diff
@@ -433,20 +433,22 @@ static int stop_scanning() {
...
-    if (err < 0) {
-        return false;
-    }
-
+    // Reserve peripheral slot. Once the central has bonded to its peripherals,
+    // the peripheral MAC addresses will be validated internally and the slot
+    // reservation will fail if there is a mismatch.
     int slot_idx = reserve_peripheral_slot(addr);
     if (slot_idx < 0) {
         LOG_ERR("Failed to reserve peripheral slot (err %d)", slot_idx);
```
xudongzheng, line 441: In the context of central being in the vicinity of multiple ZMK peripherals, this may not be really an error. Maybe `LOG_INF("unable to reserve peripheral slot ...")` should be considered.

petejohanson (approved):
Good catch. Thanks.

PR #1913, opened 2023-09-02
feat(split): Use directed advertising.

petejohanson:
* Split centrals to scan with their identity so they receive direct
  advertising packets.
* Split peripherals to use direct advertising if they have an existing
  bond to a split central.

This should *drastically* improve the experience when using splits in environments where multiple splits are present.


app/src/split/bluetooth/peripheral.c
```diff
@@ -43,15 +43,55 @@ static const struct bt_data zmk_ble_ad[] = {
...
+static void each_bond(const struct bt_bond_info *info, void *user_data) {
+    bt_addr_le_t *addr = (bt_addr_le_t *)user_data;
+
+    char addr_str[BT_ADDR_LE_STR_LEN];
+
+    bt_addr_le_to_str(&info->addr, addr_str, sizeof(addr_str));
+    LOG_DBG("EACH BOND: %s", addr_str);
+
+    if (bt_addr_le_cmp(&info->addr, BT_ADDR_LE_NONE) != 0) {
+        LOG_DBG("Capturing it!");
```
joelspadin, line 55: I'm guessing some of these log messages were for debugging? If these are useful to leave in, I think more descriptive messages would be nice.

app/src/split/bluetooth/peripheral.c
```diff
@@ -43,15 +43,55 @@ static const struct bt_data zmk_ble_ad[] = {
...
+                                                    : *BT_LE_ADV_CONN_DIR(&central_addr);
+        return bt_le_adv_start(&adv_param, NULL, 0, NULL, 0);
+    } else {
+        return bt_le_adv_start(BT_LE_ADV_CONN, zmk_ble_ad, ARRAY_SIZE(zmk_ble_ad), NULL, 0);
+    }
 };
 
+static bool low_duty_advertising = false;
+
+static void advertising_cb(struct k_work *work) { start_advertising(low_duty_advertising); }
```
joelspadin, line 81: Having a return value on `start_advertising` doesn't seem useful if we're just discarding it.

app/src/split/bluetooth/peripheral.c
petejohanson: Good catch. Removed since they're not really needed.

app/src/split/bluetooth/peripheral.c
petejohanson: Good call. Added actual checking w/ an error log message if we fail to advertise.

app/src/split/bluetooth/peripheral.c
```diff
@@ -43,15 +43,57 @@ static const struct bt_data zmk_ble_ad[] = {
...
 static bool is_connected = false;
 
-static int start_advertising() {
-    return bt_le_adv_start(BT_LE_ADV_CONN, zmk_ble_ad, ARRAY_SIZE(zmk_ble_ad), NULL, 0);
+static void each_bond(const struct bt_bond_info *info, void *user_data) {
+    bt_addr_le_t *addr = (bt_addr_le_t *)user_data;
+
+    char addr_str[BT_ADDR_LE_STR_LEN];
+
+    bt_addr_le_to_str(&info->addr, addr_str, sizeof(addr_str));
```
joelspadin, line 51: This appears to be unused now that the logging is removed.

app/src/split/bluetooth/peripheral.c
```diff
@@ -43,15 +43,57 @@ static const struct bt_data zmk_ble_ad[] = {
...
+static int start_advertising(bool low_duty) {
+    bt_addr_le_t central_addr = bt_addr_le_none;
+
+    bt_foreach_bond(BT_ID_DEFAULT, each_bond, &central_addr);
+
+    if (bt_addr_le_cmp(&central_addr, BT_ADDR_LE_NONE) != 0) {
+
+        char addr_str[BT_ADDR_LE_STR_LEN];
+
+        bt_addr_le_to_str(&central_addr, addr_str, sizeof(addr_str));
```
joelspadin, line 67: This appears to be unused now that the logging is removed.

joelspadin (changes requested):

app/src/split/bluetooth/peripheral.c
petejohanson: Indeed. Removed.

joelspadin (approved):
Code looks good to me. I can't say I'm super familiar with the bluetooth stuff, but it looks reasonable to me.

jdart:
I just tested this out with 2 splits each with their own dongle. I flashed all 6 mcu's and unplugged the dongles, then plugged them in and all 4 halves connected to their dongle successfully. Looks to have solved the issues I was having. Ty!

PR #1914, opened 2023-09-02
docs: Correct default macro timings, note DT limits

caksoylar:
Two simple changes: update the wait/tap-ms default values in the macros behavior page which were missed, and add a remark on the limit of number of bindings you can pass (which came up from user questions before, and might come up until #1893 is merged). I tested the bindings limit and saw that it succeeded at 256 items and failed at 257.


petejohanson (approved):
Good catch! Thanks.

PR #1915, opened 2023-09-02
feat(docs): Note how to use GPIO outside interconnect definition

caksoylar:
This is to answer a question that comes up often, usually in the context of using the three middle pins in nice!nanos. I put it right after the interconnects section so it appears below everything, but any suggestions on where else this could be mentioned is welcome.

### DIFF
--- a/docs/docs/development/new-shield.md
+++ b/docs/docs/development/new-shield.md
@@ -120,6 +120,10 @@ endif
 
 <InterconnectTabs items={Metadata}/>
 
+To use GPIO pins that are not part of the interconnects as described above, you can use the GPIO labels that are specific to each controller type.
+For instance, pins numbered `PX.Y` in nRF52840-based boards can be referred to via `&gpioX Y` labels.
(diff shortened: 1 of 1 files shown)


petejohanson:
@caksoylar This comes up often in the context of the NFC pins on the XIAO BLE as well.... I wonder if it should be interconnect specific notes so the can be a bit more specifically tailored for those actual scenarios?

caksoylar:
I thought about augmenting the interconnect docs, however I thought this was orthogonal to the interconnect concept as it is dependent on the SoC. For instance both nice!nanos and Xiao BLE use the same `&gpio0/1` but have different interconnects. And nice!nano and Proton C have the same interconnect but use different gpio labels (I think it was `&gpioa/b` for the latter).

We can still point out the specific examples for nano/Xiao in the interconnect section, instead of this. But one downside would be that other interconnects won't show the remarks.

Alternatively, I can add the Xiao BLE NFC pins as an additional example in the current write-up.

petejohanson (approved):

PR #1916, opened 2023-09-02
docs: Document necessary flags for kscan GPIOs

caksoylar:
I noticed that the GPIO flags aren't really explained anywhere and they are frequently missed by people when they e.g. swap diode directions. This adds prescriptive paragraphs for matrix and direct pin kscans. It is not actually explaining what the flags do, but this doesn't seem to be the right place for it and in practice it should help.

~~Perhaps we should add a link to Zephyr docs for the explanations of the flags, I'll see if I can find one.~~ Added a link.

Also includes a fix for the examples, where `rows/cols` are incorrectly implied to be supported by the kscan nodes.


petejohanson (approved):
Definite improvement. I would eventually like to refactor this *anyways* since the diode direction thing is a bit... wonky, but that captures current state of things nicely.

PR #1917, opened 2023-09-02
fix(docs): Remove diode-direction from direct GPIO driver

caksoylar:
Fixes what is probably a copy paste mistake of including `diode-direction` property in direct scan nodes.


petejohanson (approved):
Good catch!

PR #1921, opened 2023-09-03
fix(driver): Fix broken compilation for MAX7318 driver

zhiayang:
The driver got neglected and didn't compile. Now it does (:

Tested to work on real hardware.

EDIT: what's the best way to make a CI test for this so that it keeps being compiled?

### DIFF
--- a/app/module/drivers/gpio/gpio_max7318.c
+++ b/app/module/drivers/gpio/gpio_max7318.c
@@ -12,16 +12,16 @@
 
 #include <errno.h>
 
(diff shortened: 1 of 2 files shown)


petejohanson:
> 
> EDIT: what's the best way to make a CI test for this so that it keeps being compiled?

We would need a board or shield in-tree that has the driver enabled, and add that to the `core-coverage.yml` file, which drives our regression suite during CI for PRs.

petejohanson (approved):
Thanks!

Issue #1901, opened 2023-08-14
Getting a seeeduino xiao ble to deep sleep


rich-westarete:
I built a one-off keyboard for personal use, using the seeeduino xiao ble. Immediately I noticed it was consuming battery much faster than I would have expected. 

I turned on USB logging and it reported that it could not go into deep sleep because of the GD25Q16 chip.  I added `&qspi { status = "disabled";}`; to the overlay file. I also added the following lines to the .config file, which I don’t fully understand, but my intention was to allow ZMK to save settings to the internal flash memory.
```
CONFIG_MPU_ALLOW_FLASH_WRITE=y
CONFIG_NVS=y
CONFIG_SETTINGS_NVS=y
CONFIG_FLASH=y
CONFIG_FLASH_PAGE_LAYOUT=y
CONFIG_FLASH_MAP=y
```

Over the weekend battery usage was close to inline with what I would expect. I'm not sure which of these changes made the difference, but something surely did.

PR #1927, opened 2023-09-06
fix(boards): Disable QSPI for Xiao BLE

caksoylar:
The GD25Q16 flash connected via QSPI seems to be causing issues with excessive battery use and inability to sleep. Since ZMK doesn't use it, we can disable it as an easy fix. Credit goes to @rich-westarete for figuring out the issue in #1901.

I have been testing this setting in the last few days. Before the change my daily battery use on a split central varied between 3%-12% (with a 350 mAh battery). Deep sleep would not work properly, where about half the times I left it idle it went into sleep and the other half it was awake when I came back. After this change deep sleep is consistent and the battery usage has gone down to 1-2% per day.

### DIFF
--- a/app/boards/seeeduino_xiao_ble.overlay
+++ b/app/boards/seeeduino_xiao_ble.overlay
@@ -32,3 +32,6 @@
     };
 };
 
+&qspi {
+    status = "disabled";
+};


petejohanson (approved):
Right on. Good catch and fix.

PR #1934, opened 2023-09-12
Microdox V2 shield definition

alexkang:
## Board/Shield Check-list

- [y] This board/shield is tested working on real hardware
- [y] Definitions follow the general style of other shields/boards upstream ([Reference](https://zmk.dev/docs/development/new-shield))
- [y] `.zmk.yml` metadata file added
- [y] Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)
- [y] General consistent formatting of DeviceTree files
- [y] Keymaps do not use deprecated key defines (Check using the [upgrader tool](https://zmk.dev/docs/codes/keymap-upgrader))
- [n/a] `&pro_micro` used in favor of `&pro_micro_d/a` if applicable
- [y] If split, no name added for the right/peripheral half
- [y] Kconfig.defconfig file correctly wraps _all_ configuration in conditional on the shield symbol
- [y] `.conf` file has optional extra features commented out
- [y] Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead)


app/boards/shields/microdox/microdox.keymap
caksoylar, line 1: These shouldn't be refactored into a common dtsi since the setup script will copy the keymap file (and not the dtsi's) into user configs if the user requests it.

On the other hand, you shouldn't need `microdox_v2.keymap` to build and the build should be able to pick up `microdox.keymap`, but you should test it. If the keymap isn't meant to change between the two (e.g. if they don't have a different number of keys etc.) then I'd go with that.

app/boards/shields/microdox/microdox.zmk.yml
caksoylar, line 1: `v1` and `v2` aren't all siblings, so this is not right. You need to create a separate `microdox_v2.zmk.yml` instead, e.g. see https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/kyria_rev3.zmk.yml.

app/boards/shields/microdox/microdox.dtsi
```diff
@@ -4,62 +4,22 @@
  * SPDX-License-Identifier: MIT
  */
 
-#include <dt-bindings/zmk/matrix_transform.h>
+#include "microdox_common.dtsi"
 
 / {
     chosen {
-        zephyr,display = &oled;
         zmk,kscan = &kscan0;
```
caksoylar, line 11: You can move this line into `microdox_common.dtsi` as well and get rid of the chosen node. Same for `microdox_v2.dtsi`.

app/boards/shields/microdox/microdox.keymap
alexkang: Ah you're right! The build works without microdox_v2.keymap. Reverted the keymap changes.

app/boards/shields/microdox/microdox_v2.zmk.yml
```diff
@@ -0,0 +1,13 @@
+file_format: "1"
+id: microdox
+name: Microdox
```
caksoylar, line 3:
```suggestion
file_format: "1"
id: microdox_v2
name: Microdox v2
```

caksoylar:
LGTM in general. One comment is that it is not very clear to me what's the exact difference in v1 vs. v2 as a hypothetical user, especially since the URLs for both boards link to the same page. From what I can tell with a little research, Microdox sold by Boardsource switched from v1 to v2 at a certain date?

While we don't have a good mechanism to expose board/shield-specific docs, a README in the definition folder is the most frequently used approach, currently. Would you like to add a README to provide some details, such as the v1/v2 difference?

alexkang:
Good call. Added a README explaining the differences. You're right, the PCB just changed around April 2022 which had different traces, but kept the same layout.

caksoylar (approved):
Thanks for adding the README 👍

kdb424:
Checked this on my personal microdox, and it looks good to me from a functionality perspective. Thanks for the addition! There are no plans to make the V1 anymore, so glad to see V2 ported over.

petejohanson (approved):
Thanks!

PR #1941, opened 2023-09-22
app: cmake: use $ENV{ZEPHYR_BASE} as hint

M1cha:
That's the common way to do this and has the advantage that it works when the zephyr repo is an a different location and was not registered in the users cmake registry - which is a very common usecase supported by zephyr. You end up with that by `west init`-ing the git url of either the zmk repo directly or another manifest repo which imports the zmk repo. This is more convenient than the whole manual `west init -l` stuff.

This works because west automatically exports the `ZEPHYR_BASE` environment variable to all extensions - including `west build`. For that exact reason it should not break anyones scripts unless they're using cmake directly and didn't set the $ZEPHYR_BASE environment variable.

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -12,7 +12,7 @@ list(APPEND ZEPHYR_EXTRA_MODULES
 )
 
 # Find Zephyr. This also loads Zephyr's build system.
-find_package(Zephyr REQUIRED HINTS ../zephyr)
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project(zmk)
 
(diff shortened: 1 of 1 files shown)


petejohanson:
I'm a tad wary to do this, as it *forces* setting that, which IIRC, we had issues with previously, which is why we moved to the relative directory version as found here.

M1cha:
I can't find any documentation of these issues since it's basically been in git forever. The `HINT` is just a hint though. Meaning it doesn't have to exist and it will look in the global cmake package registry as a fallback. Setting `ZEPHYR_ENV` is only necessary when not using `west` because west sets it automatically. Not using west is also something that I would not recommend and that doesn't seem to be used by this projects CI (which is why it passed).

PR #1942, opened 2023-09-25
Feature: Soft Off

petejohanson:
This PR supersedes #1840 after some useful feedback from @joelspadin in https://github.com/zmkfirmware/zmk/pull/1840#issuecomment-1694718252

This new branch includes all the previous features, and also:

* Includes a `&soft_off` behavior that you can trigger from your keymap, and use a keyboard/controllers reset button to wake from.
* Supports controlling the `hold-time-ms` of the behavior, so you can require the behavior be pressed for at least that long in order for soft off to be triggered on release.
* Adds support for multiple "waker" devices. Not really needed *yet*, but trying to avoid refactors later.

See the docs deploy for this PR for details on the feature and behavior.

Only item that is a nice-to-have for the future would be some circuit diagrams for the possible hardware designs to leverage this, to complement the text descriptions of them. Don't think that should hold up this work though today.


docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,164 @@
...
+### Hardware Design
+
+ZMK's soft on/off requires a dedicated GPIO pin to be used to trigger powering off, and to wake the core from the
+soft off state when it goes active again later.
+
+#### Simple Direct Pin
+
+The simplest way to achieve this is with a push button between a GPIO pin and ground.
+
+#### Martix Intergated Hardware Combo
```
kurtis-lew, line 38:
```suggestion
#### Matrix Integrated Hardware Combo
```

Just browsing through; small typo here

petejohanson: Indeed. Fixed. Thanks!

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,164 @@
...
+        compatible = "zmk,soft-off-wakeup-sources";
+        wakeup-sources = <&wakeup_source>;
+    };
+};
+```
+
+Here are the properties for the node:
+
+- The `compatible` property for the node must be `zmk,soft-off-wakeup-sources`.
+- The `wakeup-sources` property is a phandle array pointing to all the devices that should be enabled during the shutdown process to be sure they can later wake the keyboard.
```
caksoylar, line 164:
```suggestion
- The `wakeup-sources` property is a [phandle array](../config/index.md#devicetree-property-types) pointing to all the devices that should be enabled during the shutdown process to be sure they can later wake the keyboard.
```

caksoylar (commented):
Note that the descriptions of the nodes (like the properties etc.) can go in a table format in a dedicated page under "Configuration" and linked to from here. It is not crucial but it might have been good for consistency.

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,164 @@
...
+};
+```
+
+The only difference is using a `compatible` value of `zmk,behavior-key-scanned`.
+
+#### Wakeup Sources
+
+Zephyr has general support for the concept of a device as a "wakeup source", which ZMK has not previously used. Adding soft off requires properly updating the existing `kscan` devices with the `wakeup-source` property, e.g.:
+
+```
```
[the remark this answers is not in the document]
petejohanson: I believe it's necessary to be sure that it *doesn't* get shut down when doing an automatic deep sleep, now that device PM is enabled. I will double check this all to be sure though.

docs/docs/behaviors/soft-off.md
```diff
@@ -0,0 +1,38 @@
...
+
+```
+&soft_off
+```
+
+### Configuration
+
+#### Hold Time
+
+By default, the keyboard will be turned off as soon as the key bound to the behavior is released, even it the key is only tapped briefly. If you would prefer that the key need be held a certain amount of time before releasing, you can set the `hold-time-ms` to a non-zero value in your keymap:
```
kdb424, line 26:
```suggestion
By default, the keyboard will be turned off as soon as the key bound to the behavior is released, even in the key is only tapped briefly. If you would prefer that the key need be held a certain amount of time before releasing, you can set the `hold-time-ms` to a non-zero value in your keymap:
```
Typo

caksoylar: Should be "if", right?

kdb424: Typo to fix a typo. Good catch.

petejohanson: Thanks for the double catch. Fixed.

docs/docs/features/soft-off.md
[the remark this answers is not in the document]
petejohanson: Yeah, exactly this. There's added code to `src/kscan.c` that will enable the device as a wakeup (tagging in DTS just marks it as being wakeup source *capable*, not that it's enabled as a wakeup source), so that suspend with device PM enabled will keep the kscan device enabled, so it'll be able to wake the device from "normal" deep sleep. This piece is all about keeping normal deep sleep happy.

wslfung:
Testing this on a nice!nanov2 sofle, I haven't been able to turn the right side off.

petejohanson:
> another issue I found is the idle sleep now doesn't wake up with key presses anymore.  

Did you tag your kscan node with `wakeup-source` properly for your shield?

wslfung:
> > another issue I found is the idle sleep now doesn't wake up with key presses anymore.  
> 
> Did you tag your kscan node with `wakeup-source` properly for your shield?

Does that affect both the soft_off and CONFIG_ZMK_IDLE_SLEEP_TIMEOUT?

petejohanson:
> > > another issue I found is the idle sleep now doesn't wake up with key presses anymore.  
> > 
> > Did you tag your kscan node with `wakeup-source` properly for your shield?
> 
> Does that affect both the soft_off and CONFIG_ZMK_IDLE_SLEEP_TIMEOUT?
> 
> And if I were to add it, would it be the sofle. dtsi since that's the shield I'm using? 

It's needed for wake from deep sleep to work when soft off is enabled.

wslfung:
> > > > another issue I found is the idle sleep now doesn't wake up with key presses anymore.
> > > 
> > > 
> > > Did you tag your kscan node with `wakeup-source` properly for your shield?
> > 
> > 
> > Does that affect both the soft_off and CONFIG_ZMK_IDLE_SLEEP_TIMEOUT?
> > And if I were to add it, would it be the sofle. dtsi since that's the shield I'm using?
> 
> It's needed for wake from deep sleep to work when soft off is enabled.

happy to report that worked as expected after I added wakeup-source in the sofle shield's kscan

petejohanson:
> Testing this on a nice!nanov2 sofle, I haven't been able to turn the right side off. I have mapped a &soft_off on the right half on the default layer, but that only turns off the left side.
> 
> I added .locality in behavior_soft_off_driver_api and moved the target_sources to above the if statement in CMakeLists to make the peripheral power off

Ok, I've just pushed tweaks for this to do this hopefully correctly, including making it soft off the peripheral side ASAP on press, not on release, so it can be invoked sooner before the central goes away. Can you please test this branch as is?

wslfung:
> > Testing this on a nice!nanov2 sofle, I haven't been able to turn the right side off. I have mapped a &soft_off on the right half on the default layer, but that only turns off the left side.
> > I added .locality in behavior_soft_off_driver_api and moved the target_sources to above the if statement in CMakeLists to make the peripheral power off
> 
> Ok, I've just pushed tweaks for this to do this hopefully correctly, including making it soft off the peripheral side ASAP on press, not on release, so it can be invoked sooner before the central goes away. Can you please test this branch as is?

just discarded my changes and only added wakeup-source; to kscan for sofle shield.  It works.  soft off on the peripheral turns off the peripheral side properly.

docs/docs/config/power.md
```diff
@@ -24,6 +24,18 @@ Definition file: [zmk/app/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/
 | `CONFIG_ZMK_SLEEP`              | bool | Enable deep sleep support                             | n       |
 | `CONFIG_ZMK_IDLE_SLEEP_TIMEOUT` | int  | Milliseconds of inactivity before entering deep sleep | 900000  |
 
+## Soft Off
+
+The [soft off feature](../features/soft-off.md) allows turning the keyboard on/off from either dedicated hardware of using the [`&soft_off` behavior](../behaviors/soft-off.md) to turn off and a reset button to turn back on again.
```
caksoylar, line 29:
```suggestion
The [soft off feature](../features/soft-off.md) allows turning the keyboard on/off from either dedicated hardware, or using the [`&soft_off` behavior](../behaviors/soft-off.md) to turn off and a reset button to turn back on again.
```

docs/docs/config/power.md
```diff
@@ -24,6 +24,18 @@ Definition file: [zmk/app/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/
...
+
+### Kconfig
+
+Definition file: [zmk/app/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/Kconfig)
+
+| Config                   | Type | Description                                                         | Default |
+| ------------------------ | ---- | ------------------------------------------------------------------- | ------- |
+| `CONFIG_ZMK_PM_SOFT_OFF` | bool | Enable soft off functionality from the keymap of dedicated hardware | n       |
```
caksoylar, line 37:
```suggestion
| `CONFIG_ZMK_PM_SOFT_OFF` | bool | Enable soft off functionality from the keymap or dedicated hardware | n       |
```
Guessing a bit here

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,164 @@
...
+:::
+
+Once powered off, the keyboard will only wake up when:
+
+- You press the same button/sequence that you pressed to power off the keyboard, or
+- You press a reset button found on the keyboard.
+
+## Soft Off With Existing Designs
+
+For existing designs, using soft off is as simple as placing the [Soft Off Behavior](../behaviors/soft-off.md) in your keymap and then invoking it. For splits, at least for now, you'll need to place it somewhere on each side of your keymap and trigger on both sides, starting from the peripheral side first.
```
caksoylar, line 23: Is this still correct for splits? I saw a global locality change in the updates but haven't looked deep into it.

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,164 @@
...
+
+You can then wake up the keyboard by pressing the reset button once, and repeating this for each side for split keyboards.
+
+## Adding Soft On/Off To New Designs
```
caksoylar, line 27:
```suggestion
## Adding A Dedicated Soft On/Off Button To New Designs
```
I think updating this header would clarify that this is talking about the first approach mentioned earlier in:
> Depending on the keyboard, this may be through a dedicated on/off push button, or merely through an additional binding in the keymap to turn the device off and the existing reset button to turn the device back on.


This is technically not an individual button in case of hardware combo, but do we have a better name for this "dedicated" approach?

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Hardware Design
+
+ZMK's soft on/off requires a dedicated GPIO pin to be used to trigger powering off, and to wake the core from the
```
caksoylar, line 31:
```suggestion
ZMK's dedicated soft on/off button feature requires a dedicated GPIO pin to be used to trigger powering off, and to wake the core from the
```
Tying with the previous comment

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,164 @@
...
+
+```
+/ {
+    behaviors {
+        hw_soft_off: behavior_hw_soft_off {
+            compatible = "zmk,behavior-soft-off";
+            #binding-cells = <0>;
+            label = "HW_SO";
+            hold-time-ms = <5000>;
+        };
```
caksoylar, line 79:
```suggestion
        hw_soft_off: hw_soft_off {
            compatible = "zmk,behavior-soft-off";
            #binding-cells = <0>;
            hold-time-ms = <5000>;
        };
```

caksoylar (commented):
A few further notes on a second review.

Also a remark, tying back to https://github.com/zmkfirmware/zmk/pull/1942/files#r1385777170: `wakeup-source;` seems now required for deep sleep as well, so getting the board/shield changes will follow soon I am guessing. Might be worth an announcement as well?

petejohanson:
> A few further notes on a second review.
> 
> Also a remark, tying back to https://github.com/zmkfirmware/zmk/pull/1942/files#r1385777170: `wakeup-source;` seems now required for deep sleep as well, so getting the board/shield changes will follow soon I am guessing. Might be worth an announcement as well?

It's required *only* if you turn on `ZMK_PM_SOFT_OFF`, as that enables `PM_DEVICE`. Which is part of the reason I don't see it as critical to update existing shields in this PR.

> 
> Edit: Can we document `wakeup-source;` in `config/kscan.md` and add it to the examples there as well? Either here or in the next PR?

I will add to that in this PR.

caksoylar:
> > A few further notes on a second review.
> > Also a remark, tying back to https://github.com/zmkfirmware/zmk/pull/1942/files#r1385777170: `wakeup-source;` seems now required for deep sleep as well, so getting the board/shield changes will follow soon I am guessing. Might be worth an announcement as well?
> 
> It's required _only_ if you turn on `ZMK_PM_SOFT_OFF`, as that enables `PM_DEVICE`. Which is part of the reason I don't see it as critical to update existing shields in this PR.

Ah OK then, shouldn't be a breaking change in that case.

docs/docs/features/soft-off.md
petejohanson: No, it's not! Good catch.

docs/docs/features/soft-off.md
petejohanson: Perhaps we should focus on "Dedicated On/Off GPIO Pin" as a concept?

caksoylar: Yeah, that sounds clear and precise.

petejohanson: Updated the header.

docs/docs/config/kscan.md
```diff
@@ -450,7 +455,8 @@ Note that the entire addressable space does not need to be mapped.
     };
 
     kscan0: kscan {
-        compatible = "zmk,kscan-gpio-charlieplex";
+        compatible = "zmk,kscan-gpio-charlieplex";k
```
caksoylar, line 458:
```suggestion
        compatible = "zmk,kscan-gpio-charlieplex";
```

caksoylar (approved):
Fixed a typo on the last update, otherwise docs look good to me.

app/boards/shields/zmk_uno/boards/nrf52840dk_nrf52840.overlay
```diff
@@ -20,5 +29,38 @@ encoder: &qdec0 {
...
+
+    soft_off_wakers {
+        compatible = "zmk,soft-off-wakeup-sources";
+        status = "okay";
+
+        wakeup-sources = <&wakeup_source>;
+    };
+
+    soft_off_behavior_key {
+        compatible = "zmk,behavior-key";
```
joelspadin, line 61: Now that I see what this does, I think this naming could be very confusing. It fits the pattern of `zmk,behavior-*` for things that can be bound to a keymap, but it is fundamentally a different thing. It's more akin to a combination of kscan driver and keymap than a behavior.

We could also make it even clearer that this isn't a traditional behavior by avoiding the `zmk,behavior-*` pattern and/or using `gpio-key` instead of `key`. How about something like `zmk,gpio-key-behavior-trigger`?

app/dts/bindings/zmk,behavior-key-scanned.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+properties:
+  key:
+    type: phandle
+    required: true
+    description: The GPIO key that triggers wake via interrupt
+  bindings:
+    type: phandle
+    required: true
+    description: The GPIO key that triggers wake via interrupt
```
joelspadin, line 19: Copy/paste of the property above?

app/dts/bindings/zmk,behavior-key.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+properties:
+  key:
+    type: phandle
+    required: true
+    description: The GPIO key that triggers wake via interrupt
+  bindings:
+    type: phandle
+    required: true
+    description: The GPIO key that triggers wake via interrupt
```
joelspadin, line 19: Copy/paste of the above property?

app/dts/bindings/zmk,soft-off-wakeup-sources.yaml
```diff
@@ -0,0 +1,14 @@
...
+  Description of all possible wakeup-sources from a forces
+  soft-off state.
+
+compatible: "zmk,soft-off-wakeup-sources"
+
+properties:
+  wakeup-sources:
+    type: phandles
+    required: true
+    description: List of wakeup-sources that should be enabled to wake the system from forces soft-off state.
```
joelspadin, line 14:
```suggestion
    description: List of wakeup-sources that should be enabled to wake the system from soft-off state.
```
extra word? Or was that supposed to be "forced"?

app/dts/bindings/zmk,soft-off-wakeup-sources.yaml
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2023 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+description: |
+  Description of all possible wakeup-sources from a forces
```
joelspadin, line 5:
```suggestion
  Description of all possible wakeup-sources from a
```
Extra word? Or was that supposed to be "forced"?

app/dts/bindings/zmk,behavior-key-scanned.yaml
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2023 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+description: |
+  Driver for a dedicated key triggered by matrix scanning for invoking a connected behavior.
+
+compatible: "zmk,behavior-key-scanned"
```
joelspadin, line 7: If renaming `zmk,behavior-key`, this should be renamed similarly.

app/dts/bindings/zmk,wakeup-trigger-key.yaml
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2023 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+description: |
+  Driver for a dedicated key for waking the device from sleep
+
+compatible: "zmk,wakeup-trigger-key"
```
joelspadin, line 7: If renaming `zmk,behavior-key` to something that includes `gpio-key`, consider renaming this as well.

e.g. if we like `zmk,gpio-key-behavior-trigger`, then this could be `zmk,gpio-key-wakeup-trigger` for consistency.

app/module/drivers/kscan/kscan_gpio_direct.c
```diff
@@ -318,6 +319,28 @@ static int kscan_direct_init(const struct device *dev) {
...
+        break;
+    case PM_DEVICE_ACTION_RESUME:
+        kscan_direct_enable(dev);
+        break;
+    default:
+        ret = -ENOTSUP;
+        break;
+    }
+
+    return ret;
```
joelspadin, line 339: Minor: this is ignoring the return code for enable/disable. Could fix that and simplify a bit by just using `return` in the switch?

```suggestion
    switch (action) {
    case PM_DEVICE_ACTION_SUSPEND:
        return kscan_direct_disable(dev);

    case PM_DEVICE_ACTION_RESUME:
        return kscan_direct_enable(dev);

    default:
        return -ENOTSUP;
    }
```

app/module/drivers/kscan/kscan_gpio_matrix.c
```diff
@@ -421,6 +422,28 @@ static int kscan_matrix_init(const struct device *dev) {
     return 0;
 }
 
+#if IS_ENABLED(CONFIG_PM_DEVICE)
+
+static int kscan_matrix_pm_action(const struct device *dev, enum pm_device_action action) {
```
joelspadin, line 427: Same as above.

app/src/behavior_key.c
```diff
@@ -0,0 +1,159 @@
...
+    data->read_time = k_uptime_get();
+    k_work_reschedule(&data->update_work, K_NO_WAIT);
+}
+
+static int behavior_key_init(const struct device *dev) {
+    const struct behavior_key_config *config = dev->config;
+    struct behavior_key_data *data = dev->data;
+
+    if (!device_is_ready(config->key.port)) {
+        LOG_ERR("GPIO port is not ready");
```
joelspadin, line 99: Could include the device name in the error for easier debugging.

app/src/behavior_key.c
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    k_work_init_delayable(&data->update_work, bk_update_work);
+    data->dev = dev;
+
+    gpio_pin_configure_dt(&config->key, GPIO_INPUT);
+    gpio_init_callback(&data->key_callback, bk_gpio_irq_callback, BIT(config->key.pin));
+    gpio_add_callback(config->key.port, &data->key_callback);
+
+    while (gpio_pin_get_dt(&config->key)) {
```
joelspadin, line 110: This probably deserves a comment to explain why this is being done. I'm guessing this is to avoid the case where the keyboard just woke up due to pressing this key, so you don't want to immediately read that as a trigger to go back to sleep? (Does this need debouncing too? If so, moving this out to a "wait for release" function would be helpful to keep the init function simple.)

app/src/behavior_key_scanned.c
```diff
@@ -0,0 +1,194 @@
...
+
+    gpio_pin_interrupt_configure_dt(&config->key, GPIO_INT_DISABLE);
+}
+
+static void bks_read(const struct device *dev) {
+    const struct behavior_key_scanned_config *config = dev->config;
+    struct behavior_key_scanned_data *data = dev->data;
+
+    if (k_sem_take(&data->sem, K_NO_WAIT) < 0) {
+        // k_work_reschedule(&data->update_work, K_NO_WAIT);
```
joelspadin, line 58: Leftover testing code?

app/src/behavior_key_scanned.c
```diff
@@ -0,0 +1,194 @@
...
+        return;
+    }
+
+    data->active_scan_detected = true;
+    data->read_time = time;
+
+    if (!zmk_debounce_is_active(&data->debounce_state)) {
+        // When we get that very first interrupt, we need to schedule the update checks to fall in
+        // between each of the real scans, so we can do our checks for state *after* each scan has
+        // occurred.
```
joelspadin, line 117: This is trying to schedule this device's reads to happen in between the kscan driver's scans so they don't conflict? If so, this seems really fragile to me. Off the top of my head, here are some things that could break it:

- The scan periods are set to different things on the two devices.
- You're using a bunch of GPIO expanders or something, and doing one scan takes longer than half the scan period.

app/src/behavior_key_scanned.c
```diff
@@ -0,0 +1,194 @@
...
+
+static void bks_gpio_irq_callback(const struct device *port, struct gpio_callback *cb,
+                                  const gpio_port_pins_t pin) {
+    struct behavior_key_scanned_data *data =
+        CONTAINER_OF(cb, struct behavior_key_scanned_data, key_callback);
+    const struct behavior_key_scanned_config *config = data->dev->config;
+
+    uint32_t time = k_uptime_get();
+
+    if (k_sem_take(&data->sem, K_MSEC(10)) < 0) {
```
joelspadin, line 105: I don't think ISRs are allowed to take a semaphore? The documentation implies that ISRs can *give* a semaphore, but only threads can take one.

app/src/behavior_key_scanned.c
joelspadin, line 1: I'm not really sure I understand what this doing or why it exists, so I'm not sure if my comments on this file are useful.

app/src/behaviors/behavior_soft_off.c
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
joelspadin, line 2:
```suggestion
 * Copyright (c) 2023 The ZMK Contributors
```
unless you started writing this file in 2020?

app/src/pm.c
```diff
@@ -0,0 +1,65 @@
...
+    device_count = z_device_get_all_static(&devs);
+
+    // There may be some matrix/direct kscan devices that would be used for wakeup
+    // from normal "inactive goes to sleep" behavior, so disable them as wakeup devices
+    // and then suspend them so we're ready to take over setting up our system
+    // and then putting it into an off state.
+    for (int i = 0; i < device_count; i++) {
+        const struct device *dev = &devs[i];
+
+        LOG_DBG("soft-on-off pressed cb: suspend device");
```
joelspadin, line 47: This is going to spam a *lot* of the same line to the log when entering soft off. Maybe move it out of the loop?

app/src/pm.c
```diff
@@ -0,0 +1,65 @@
...
+    DT_FOREACH_PROP_ELEM(DT_INST(0, zmk_soft_off_wakeup_sources), wakeup_sources, DEVICE_WITH_SEP)};
+
+#endif
+
+int zmk_pm_soft_off(void) {
+#if IS_ENABLED(CONFIG_PM_DEVICE)
+    size_t device_count;
+    const struct device *devs;
+
+#if !IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
```
joelspadin, line 34: Probably something for another PR, but there are a lot of these checks for "if split and peripheral" or "if not split or central" in the codebase. Might be useful to create some macros to simplify those.

app/src/pm.c
```diff
@@ -0,0 +1,65 @@
...
+
+#if HAS_WAKERS
+    for (int i = 0; i < ARRAY_SIZE(soft_off_wakeup_sources); i++) {
+        const struct device *dev = soft_off_wakeup_sources[i];
+        pm_device_wakeup_enable(dev, true);
+        pm_device_action_run(dev, PM_DEVICE_ACTION_RESUME);
+    }
+#endif // HAS_WAKERS
+
+    LOG_DBG("soft-on-off interrupt: go to sleep");
```
joelspadin, line 63: This isn't necessarily called from an interrupt handler?

app/src/wakeup_trigger_key.c
```diff
@@ -0,0 +1,87 @@
...
+static int wakeup_trigger_key_pm_action(const struct device *dev, enum pm_device_action action) {
+    const struct wakeup_trigger_key_config *config = dev->config;
+    int ret = 0;
+
+    switch (action) {
+    case PM_DEVICE_ACTION_RESUME:
+        ret = gpio_pin_interrupt_configure_dt(&config->trigger, GPIO_INT_LEVEL_ACTIVE);
+        if (ret < 0) {
+            LOG_ERR("Failed to configure wakeup trigger key GPIO pin interrupt (%d)", ret);
+            return ret;
```
joelspadin, line 45: Minor: mixing early returns with allowing the case to run until the break and `return ret` at the end seems a bit weird.

Maybe pull each of these cases out into a separate function, so then you can just do this?

```c
switch (action) {
case PM_DEVICE_ACTION_RESUME:
    return wakeup_trigger_key_pm_resume(dev);

case PM_DEVICE_ACTION_SUSPEND:
    return wakeup_trigger_key_pm_suspend(dev);

default:
    return -ENOTSUP;
}
```

joelspadin (changes requested):

app/boards/shields/zmk_uno/boards/nrf52840dk_nrf52840.conf
```diff
@@ -0,0 +1 @@
+CONFIG_ZMK_PM_SOFT_OFF=y
```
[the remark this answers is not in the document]
petejohanson: I specifically *did* want to force this just for the Uno, so we could easily test the on/off with Button 1 on the DK.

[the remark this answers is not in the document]
joelspadin: This makes it impossible to disable this config though, in case someone wants to override the default config and use the buttons for something else.

app/src/behavior_key_scanned.c
petejohanson: Good catch indeed. Will be pushing a rework that avoids this.

app/dts/bindings/zmk,behavior-key-scanned.yaml
petejohanson: Fixed

app/boards/shields/zmk_uno/boards/nrf52840dk_nrf52840.overlay
petejohanson: *grumble grumble* Yes, you're absolutely right. Renamed for clarity. Thanks!

app/src/wakeup_trigger_key.c
petejohanson: Much better. Refactored.

app/Kconfig.behaviors
```diff
@@ -1,6 +1,16 @@
 # Copyright (c) 2023 The ZMK Contributors
 # SPDX-License-Identifier: MIT
 
+config ZMK_BEHAVIOR_KEY
```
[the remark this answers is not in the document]
petejohanson: Renamed per the other review comment. I believe that *mostly addresses it.

app/dts/bindings/zmk,soft-off-wakeup-sources.yaml
petejohanson: Was intended to be "forced". Fixed.

app/src/behavior_key.c
petejohanson: Pulled this out into a separate function and added a comment. Thanks.

app/src/behavior_key_scanned.c
petejohanson: Yeah, so this is the best I can imagine for handling this, which is the best way I can think of to detect this.

The scan periods need to be at least the matrix scan rate *or larger* and it should be ok. Definitely a risk, but not one I can easily see a workaround for.

As for slower scanning... I tweaked the scheduling to be just 1ms before the next debounce scanning period, to reduce the risk of the second scenario happening.

app/boards/shields/zmk_uno/boards/nrf52840dk_nrf52840.conf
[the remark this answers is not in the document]
petejohanson: Fair. Let me try to use a different imply/default then instead.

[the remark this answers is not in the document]
petejohanson: Yup. Confirmed. Tweaked the `Kconfig.defconfig` for the shield instead.

app/src/gpio_key_wakeup_trigger.c
```diff
@@ -0,0 +1,96 @@
...
+    const struct gpio_key_wakeup_trigger_config *config = dev->config;
+
+    int ret = gpio_pin_interrupt_configure_dt(&config->trigger, GPIO_INT_LEVEL_ACTIVE);
+    if (ret < 0) {
+        LOG_ERR("Failed to configure wakeup trigger key GPIO pin interrupt (%d)", ret);
+        goto exit;
+    }
+
+    for (int i = 0; i < config->extra_gpios_count; i++) {
+        ret = gpio_pin_configure_dt(&config->extra_gpios[i], GPIO_OUTPUT_ACTIVE);
```
joelspadin, line 46: These are being set active here, but they are deactivated when the interrupt is deactivated. Is that intentional?

app/src/gpio_key_wakeup_trigger.c
```diff
@@ -0,0 +1,96 @@
...
+
+    return 0;
+}
+
+#if IS_ENABLED(CONFIG_PM_DEVICE)
+
+static int gpio_key_wakeup_trigger_pm_resume(const struct device *dev) {
+    const struct gpio_key_wakeup_trigger_config *config = dev->config;
+
+    int ret = gpio_pin_interrupt_configure_dt(&config->trigger, GPIO_INT_LEVEL_ACTIVE);
```
joelspadin, line 39: The logic for resume/suspend feels backwards to me, because I was thinking of this in terms of the power state of the system until I read pm.c. This might warrant to comment somewhere to explain that this device is normally suspended and then gets resumed just prior to going into soft off.

app/CMakeLists.txt
```diff
@@ -22,7 +22,11 @@ target_sources(app PRIVATE src/matrix_transform.c)
 target_sources(app PRIVATE src/sensors.c)
 target_sources_ifdef(CONFIG_ZMK_WPM app PRIVATE src/wpm.c)
 target_sources(app PRIVATE src/event_manager.c)
+target_sources_ifdef(CONFIG_ZMK_GPIO_KEY_BEHAVIOR_TRIGGER app PRIVATE src/gpio_key_behavior_trigger.c)
+target_sources_ifdef(CONFIG_ZMK_GPIO_SCANNED_KEY_BEHAVIOR_TRIGGER app PRIVATE src/gpio_scanned_key_behavior_trigger.c)
+target_sources_ifdef(CONFIG_ZMK_PM_SOFT_OFF app PRIVATE src/pm.c)
```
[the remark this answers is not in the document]
petejohanson: We might add other PM specific bits here eventually. I don't mind adding the extra Kconfig and doing that little bit of future-proofing.

app/src/gpio_key_wakeup_trigger.c
petejohanson: Added a comment there

app/src/gpio_key_wakeup_trigger.c
petejohanson: Nope. Just oversight. We don't actually ever use the suspend code right now... But added for completeness.

app/src/gpio_key_wakeup_trigger.c
```diff
@@ -0,0 +1,96 @@
...
+    if (ret < 0) {
+        LOG_ERR("Failed to configure wakeup trigger key GPIO pin interrupt (%d)", ret);
+        goto exit;
```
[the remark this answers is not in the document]
petejohanson: No. Just was too tired last time I was fixing up review items. Reverted to saner return form. Thanks.

Luerl21:
Will you do a soft off for charliepex

Issue #1944, opened 2023-09-28
[Bug] Overlapping combos with different timeouts


HelloThisIsFlo:
Hi,

I'm having an issue with overlapping combo with different timeouts. What I want to achieve is to have different timeouts for combos that are using keys on one hand only vs combos that requires two hands to activate. That way I can have 2-hands combos that are easier to trigger, while still keeping a short timeout for 1-hand combos so fast typing doesn't activate the combo.

Unfortunately, I think there is a bug in how the combo timeout are processed (see Context & The Problem). I did try to put a lot of logging try figure out what is going on, but I am not a C/C++ developer, and after spending 4hours trying to figure it out I didn't get very far. I really struggled with logging struct, or arrays. So the progress was excruciatingly slow 😬

That being said, I'd be more than happy to run local experiments if you need more info to isolate the issue

### Context
- I set up 2 combos, they have 1 key in common `20`
  - Combo 1 `OneHandLeft` is: `20 21` with a timeout of `2000` (large for debug purposes)
  - Combo 2 `TwoHands` is: `20 29` with a timeout of `4000` (large for debug purposes)
  - Here are the key positions
    ```
    ╭────────────────────╮ ╭────────────────────╮
    │ __  __  __  __  __ │ │ __  __  __  __  __ │
    │ __  __  __  __  __ │ │ __  __  __  __  __ |
    | 20  21  __  23  24 │ │ __  __  __  28  29 |
    ╰───────╮ __  __  __ | | __  __  __ ╭───────╯
            ╰────────────╯ ╰────────────╯
    
    ```

#### Expected
- When pressing `20`, I would expect the key to wait for all associated combo timeouts to expire before deciding that no combo is a match.
- In other words I would expect it to trigger the regular keypress _only_ after `4000ms` have passed

#### Actual Behavior
- The behavior associated with the key is triggered as soon as the combo with the shortest timeout times out
- It triggers `2000ms` after the key was pressed.
- This is consistent with other situations (see Real World Example below), the key will trigger its behavior as soon a **one combo expires**. It does not wait for all other combos with longer timeouts.

---
#### Real World Example
I'm using this example keymap on this setup:
- **Actual Keyboard:** Chocofi
- **Board:** Nice!Nano V2
- **Shield:** Corne (w/ `five_column_transform`)
- **Build command:**
    ```
[...]

PR #1945, opened 2023-09-30
Fix bug with overlapping combos timeouts

HelloThisIsFlo:
> ### This is a follow-up to #1944 . See the issue for more context.
> This PR fixes the bug mentioned in the issue

## Context

When a timeout fires, the code will filter the remaining combos. Eliminating the one the just timed out, and reordering the remaining ones.

The code _does_ filter the timed out combos correctly, and it does re-order the remaining ones correctly. But! It didn’t clean up properly the position of the very last candidate.

They all move one spot, so when they move, they override the one that was previously there. But the very last combo doesn’t get overridden by anything.

```
Timestamp = 60ms

Before filtering
Combo expires_at=60ms
Combo expires_at=70ms
Combo expires_at=80ms
Combo expires_at=90ms

After filtering
Combo expires_at=70ms
Combo expires_at=80ms
Combo expires_at=90ms
Combo expires_at=90ms <- This extra one is the problem
```

After filtering it returns the number of active candidates. The first time it does it, the number returned is correct, even though this extra entry in `candidates` is not cleaned up.

But then second time it looks at the list, it will return an incorrect number of active candidates. Because of the duplication that happens each time the last candidate is moved up one spot.

## The Bug

When a combo is timing out, the `filter_timed_out_candidates` function is called by `combo_timeout_handler`. 

When a combo times out, we know it’s not going to be activated. The only times when a combo can activate is when a key is pressed (which also calls `filter_timed_out_candidates`).

Since a combo can not be triggered by a timeout, we should only `cleanup()` once **no candidates are remaining.**

However the current code will `cleanup()` if `0 or 1` candidates are remaining. I’m not quite sure why that was implemented this way, maybe it was a quick fix for the issue described above in the context section. Or maybe the idea was that when we have 1 candidate remaining we should clean up. Either way, this was the source of the problem described in #1944 

## Solution

1. Make sure we properly clear out the unused spots in the `candidates` list
2. On timeout, only `cleanup()` if exactly `0 candidates` are left

I also added a test to showcase the proper timeout behavior, and added some debug logging to make this test possible.

I ran all the tests, they all pass ✅

It’s my first PR on this repo, so let me know if if I should update something.

Thanks for this amazing piece of software 😍

### DIFF
--- a/app/src/combo.c
+++ b/app/src/combo.c
@@ -204,22 +204,33 @@ static inline bool candidate_is_completely_pressed(struct combo_cfg *candidate) 
 static int cleanup();
 
 static int filter_timed_out_candidates(int64_t timestamp) {
-    int num_candidates = 0;
+    int remaining_candidates = 0;
     for (int i = 0; i < CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY; i++) {
         struct combo_candidate *candidate = &candidates[i];
         if (candidate->combo == NULL) {
             break;
         }
         if (candidate->timeout_at > timestamp) {
-            // reorder candidates so they're contiguous
-            candidates[num_candidates].combo = candidate->combo;
-            candidates[num_candidates].timeout_at = candidate->timeout_at;
-            num_candidates++;
+            bool need_to_bubble_up = remaining_candidates != i;
+            if (need_to_bubble_up) {
+                // bubble up => reorder candidates so they're contiguous
+                candidates[remaining_candidates].combo = candidate->combo;
+                candidates[remaining_candidates].timeout_at = candidate->timeout_at;
+                // clear the previous location
+                candidates[i].combo = NULL;
+                candidates[i].timeout_at = 0;
+            }
+
+            remaining_candidates++;
         } else {
             candidate->combo = NULL;
         }
     }
-    return num_candidates;
+
+    LOG_DBG("after filtering out timed out combo candidates: remaining_candidates=%d timestamp=%d",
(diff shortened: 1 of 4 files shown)


HelloThisIsFlo:
On a sidenote, this PR should conflict with #1387, but maybe it's safer to merge #1387 first. If there's a conflict I'll make sure to update my PR accordingly 👍

HelloThisIsFlo:
Apologies for the failing pre-commit hook. I'll fix that now

petejohanson:
> Apologies for the failing pre-commit hook. I'll fix that now

No worries. For security reasons, PRs from new contributors need approval, so the feedback there can sometimes come "late".

See the docs for setting up pre-commit locally if you are inclined to continue contributing. Thanks for the PR!

HelloThisIsFlo:
That's fair enough. For now it's running on my personal fork, so there's no rush on my side. While it's a very useful fix for me (it allows different timeouts for 1 hand combo vs 2 hand combos), I can see how it's also not affecting a lot of people. So it's not a critical bug at all. Take the time to review it whenever it's approved 👍

I fixed the `pre-commit` issue

petejohanson (approved):
What a great PR! Thoroughly documented description of both the bug, and the fix, and clean code for the fix, along with accompanying test to boot!

Will merge soon as I see all the tests go green.

Thanks for the fix!

HelloThisIsFlo:
Thank you for your comment 🙏  
I'm happy to contribute to this amazing piece of software 😃

PR #1946, opened 2023-10-01
feat(boards): Update for mikoto board definition

zhiayang:
This PR updates the mikoto board definition to use Zephyr revisions, while adding support for new revisions (6.1+, 7.1+) that have incompatibilities/new hardware.

Unfortunately this breaks existing configs that use `mikoto_520` as the board name. Is that a major blocker?

## Board/Shield Check-list

- x ] Definitions follow the general style of other shields/boards upstream ([Reference](https://zmk.dev/docs/development/new-shield))


app/boards/arm/mikoto/mikoto.zmk.yml
caksoylar, line 1: How do you specify the right version when building with Zephyr? Note that if you only have this metadata file, the ZMK setup script will output `board: mikoto` when the board is selected. Not sure what version that would correspond to, or how you would specify a different revision with the setup script in this case.

zhiayang: I referenced the Zephyr docs for this: https://docs.zephyrproject.org/latest/hardware/porting/board_porting.html#multiple-board-revisions

I set the default revision as `5.20.0` in the `revision.cmake`, so if you just specify the board as `mikoto` then you'll get that revision.

To get other revisions you use `mikoto@7.1.0`, for example

caksoylar: Do you have an idea if the number of users that use the older version is negligible? Because with this change anyone that currently uses the `mikoto` board will automatically be using the new revision. I suspect the changes between versions are breaking, are they not?

caksoylar: I guess the point I am trying to make is that the Zephyr board revisioning system is great, but I don't know if the ZMK tooling (i.e. the setup scripts and user config repos) can really account for them at the moment.

zhiayang: I set the default revision as `5.20`, which is what the previous `mikoto_520` corresponds to.

I'm actually worried about that part, because the old board name was `mikoto_520`, whereas the new one is just `mikoto`...

caksoylar: Yeah sorry, I edited my comment above after noticing your point in the original post.

zhiayang: Ah, I see. Yea, not sure about the implications of this change; Pete pointed me towards the Zephyr revisioning system.

I don't actually know how the ZMK scripts work, since I always build my firmwares locally xD

app/boards/arm/mikoto/mikoto.yaml
```diff
@@ -1,5 +1,5 @@
-identifier: mikoto_520
-name: mikoto_520
+identifier: mikoto
+name: mikoto
```
caksoylar, line 2:
```suggestion
name: Mikoto 5.20
```
In theory, you could change this so the display name in the script is "Mikoto 5.20". Then you could add copies of this file with `identifier: mikoto@6.1.0` `identifier: mikoto@7.1.0` with the corresponding `name`s, for instance. However I am not sure if everything in ZMK scripts would work well with the `@` syntax or something would break, I guess we could try and see.

zhiayang: I'll give that a go

app/boards/arm/mikoto/mikoto.zmk.yml
petejohanson: I'd say breaking existing build would definitely be a failure. Can we preserve the existing board ID but update metadata to reference the new versioned board(s)?

As far as the setup script, you can build docs locally to generate a new setup.sh and test locally, to see the behavior after your metadata changes. I expect you'd need to add a metadata file per version, or enhance our metadata to handle board versions and tweak the setup scripts appropriately.

zhiayang:
a bunch of stuff came up, haven't had time to move this forward atm. though i will say that my changes work locally (ignoring any breaking changes).

app/boards/arm/mikoto/mikoto.zmk.yml
zhiayang: i remember seeing on the discord there was some talk about moving to something like python for the setup script -- if that's on the horizon, should I wait for that to be ready/merged before continuing?

i've had a brief look at the setup.sh script, and it looks a *little* fragile if i were to change it to support board revisions.

zhiayang:
had some time to pick this up again; so what i've changed now is:

1. teach the schema about the new fields `revisions` and `default_revision`. in case you're curious about the regex, it comes from the [zephyr docs](https://docs.zephyrproject.org/latest/hardware/porting/board_porting.html#multiple-board-revisions); it handles LETTER, NUMBER, and MAJOR.MINOR.PATCH revisions.
2. put back `mikoto_520`; it can still be used as in `-DBOARD=mikoto_520`, and i've tried my best to deduplicate the definitions and stuff.
3. edit the setup script (only the bash one for now -- if this approach is OK-ed i'll update the powershell one too) to teach it about board revisions. Mikoto 5.20 is still in there, but i've marked it deprecated -- and the zmk metadata makes it alias `mikoto@5.20.0`.

basically, now we should handle the new cases (`board@revision`) for all boards that opt-in to this mechanism, and we keep compat with people using `mikoto_520` in their repos.

I've set up a test repo that builds both `mikoto_520` and `mikoto@7.1.0`, and both work: https://github.com/zhiayang/zmk-test

app/build.sh
caksoylar, line 1: I think this commit efa9134e included these three files (build.sh, uf2s) by accident.

zhiayang: oops, you're right. I'll clean this up after #2157 gets merged, thanks!

zhiayang:
aight, #2157 got merged so i'll pick this up in a bit

medoix:
@zhiayang the display adapter overlay also needs to be copied to `mikoto.overlay` for the nice!view and similar displays to work.

https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/nice_view_adapter/boards/mikoto_520.overlay

zhiayang:
Okay, finally had time to catch up with this. Also realised that at least one store is selling v7.2 so I figured I had to get on this somewhat quickly.

Did testing locally, and was able to compile boards with both v5.20 and v7.2

app/boards/arm/mikoto/mikoto.zmk.yml
```diff
@@ -1,10 +1,14 @@
...
 type: board
 arch: arm
 outputs:
   - usb
   - ble
 url: https://github.com/zhiayang/mikoto
 exposes: [pro_micro]
+revisions:
+  - "5.20"
+  - "7.2"
```
Nick-Munnich, line 13: Missing 6.1 here

petejohanson:
>     2. put back `mikoto_520`; it can still be used as in `-DBOARD=mikoto_520`, and i've tried my best to deduplicate the definitions and stuff.

It seems this got lost. I'm very happy to switch to using board versioning, but until we add ZMK versioning and want to have a breaking release, I'd request we not drop the `mikoto_520` board ID completely.

zhiayang:
Oops, guess I accidentally yeeted that. Will get on it.

sebkinne:
@zhiayang - thank you for making the Mikoto, and for this PR!

I've tried building ZMK for my 7.2 with this branch, but it looks like there might be an issue with the MAX17048 config - ZMK battery reporting shows 0% (screen) or 100% (host OS) at all times.

I'm not familiar enough with ZMK to offer any suggestions, but figured I'd flag it before merging. I'll try and dig into it this weekend though!

zhiayang:
@sebkinne are you building with the board `mikoto@7.2` ? Just using `mikoto` will get you 5.20 which doesn't have the fuel gauge IC enabled.

I'll do some testing this week, but my daily driver isn't the mikoto any more due to work wireless policies.

sebkinne:
@zhiayang yep, building with mikoto@7.2 (for the corne): https://github.com/sebkinne/zmk-config/tree/mikoto

zhiayang:
Alright, I'll see about investigating it this week.

zhiayang:
@petejohanson seems like there's a nice `ZEPHYR_BOARD_ALIASES` feature that we can use to just alias `mikoto_520` to `mikoto`, which just gets rid of the duplicated definitions and stuff entirely. Updated the PR to use that, lmk what you think?

A (probably non-issue) caveat: seems like you can't specify board revisions as the alias target (eg. I can't alias it to `mikoto@5.20` directly), but it uses the default revision -- which is still 5.20 for now. We can probably change the default if/when we drop `mikoto_520` entirely.

zhiayang:
@sebkinne It took a while, but I managed to get around to looking at this. Turns out I just forgot to enable the gpio pull-ups. It works on my side, could you test on your end? Thanks!

sebkinne:
Thanks @zhiayang - that worked like a charm!

petejohanson (approved):
Thanks!

PR #1947, opened 2023-10-02
boards: nrf52840dk_nrf52840: default kscan

hlord2000:
Adds a default kscan and keymap for the nRF52840DK

Signed-off-by: Kelly Helmut Lord <helmut@helmutlord.com>

### DIFF
--- a/app/boards/nrf52840dk_nrf52840.keymap
+++ b/app/boards/nrf52840dk_nrf52840.keymap
@@ -1,0 +1,16 @@
+#include <behaviors.dtsi>
+#include <dt-bindings/zmk/keys.h>
+
+/ {
+    keymap {
+        compatible = "zmk,keymap";
+
+        layer_1 {
+            label = "base";
+            bindings = <
+                &kp A &kp B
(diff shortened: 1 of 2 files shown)


petejohanson (changes requested):
Thinking about this more... I wonder if all of this shouldn't go in an `nrf52840_nrf52840dk.keymap` file instead... so it *only* gets picked up at all if you don't build with a shield with it's own keymap. Can you give that a try?

hlord2000:
> Thinking about this more... I wonder if all of this shouldn't go in an `nrf52840_nrf52840dk.keymap` file instead... so it _only_ gets picked up at all if you don't build with a shield with it's own keymap. Can you give that a try?

I attempted to add a .keymap file but it will require changes to ZephyrBuildConfig.cmake. As it is now, the build will only look for a .keymap file if compiling with a shield option. This worked, however:
`west build -b nrf52840dk_nrf52840 -p -- -DKEYMAP_FILE=./boards/nrf52840dk_nrf52840.keymap`

PR #1954, opened 2023-10-05
Split serial

M1cha:
This picks up on #1117. I created this as a draft PR so I can collect feedback of the maintainers before investing the work to finish things up. Specifically:
- Are you okay with COBS?
- Do you agree with the general way the code works?

known TODOs:
- [ ] Add support for sending with an interrupt-based UART driver.
- [ ] Select transport via device tree.
- [ ] Implement and test sensors.

It's also very apparent, that I had to copy quite some code from the BLE central. In future we might want to refactor both implementations with the goal of changing the abstraction layer to something that reduces code duplication even more. Right now we should focus on getting this working though so we can see how much overlap we really end up with.

### DIFF
--- a/app/src/split/CMakeLists.txt
+++ b/app/src/split/CMakeLists.txt
@@ -4,3 +4,7 @@
 if (CONFIG_ZMK_SPLIT_BLE)
     add_subdirectory(bluetooth)
 endif()
+
+if (CONFIG_ZMK_SPLIT_SERIAL)
+    add_subdirectory(serial)
+endif()
--- a/app/src/split/Kconfig
+++ b/app/src/split/Kconfig
@@ -18,9 +18,13 @@ config ZMK_SPLIT_BLE
     select BT_USER_PHY_UPDATE
     select BT_AUTO_PHY_UPDATE
 
+config ZMK_SPLIT_SERIAL
+    bool "Serial"
+
 endchoice
 
 #ZMK_SPLIT
 endif
 
 rsource "bluetooth/Kconfig"
+rsource "serial/Kconfig"
--- a/app/src/split/serial/CMakeLists.txt
+++ b/app/src/split/serial/CMakeLists.txt
@@ -1,0 +1,13 @@
+# Copyright (c) 2022 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+target_link_libraries(app PRIVATE
+    COBS
+)
+
+if (CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
+  target_sources(app PRIVATE central.c)
+  target_sources(app PRIVATE interrupt.c)
+else ()
+  target_sources(app PRIVATE peripheral.c)
+  target_sources(app PRIVATE polling.c)
+endif()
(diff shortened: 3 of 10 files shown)


petejohanson:
I'll review COBS a bit closer, but generally I don't see a huge problem using something like this for this. We don't really need advanced features like protobuf for things like this, e.g. backwards/forwards compat with encoded data, etc.

The general approach: I agree, there's definitely some duplication. I'd love our final solution to abstract out just the "transport" portion into an API that allows swapping out the implementation of just that piece, instead of duplicating the whole thing, but I understand the hesitancy to do this "too early". I think as long as we have an *eye* on this eventual de-dup, I'm fine with it.

Now that upstream Zephyr in recent versions has added a PIO based UART driver, I think sticking with "UART device as the abstraction level" is reasonable and will offer enough flexibility I don't think we need our own "wired transport" abstraction over the UART driver one.

Thanks for picking this up!

kdb424:
I've got great interest in this, and will be tracking and testing, and porting to Lulu and another unnamed board as soon as PIO is working. I have a backlog of 2 wire splits that I'd love to see get full ZMK support. Not sure how much I can contribute on a code front, but more than willing to give feedback and stress test.

M1cha:
@kdb424 do you need PIO on both sides? For now I only implemented PIO on the peripheral side because:
- My hw is wired in a way that the receiver supports hardware UART(rx only)
- The PIO driver is (currently) polling only which means that it can impact the whole system because it prevents any other threads from running. That's okay for the peripheral side because there we're just sending and can thus control when we want to send data. Also, the peripheral doesn't have anything else to do anyway because BLE/USB/... are all handled by the central.

kdb424:
I've got one and two wire RP2040 boards of various pins used. I'm sure I have something around that I can test with, from dev boards, to full keyboards. I'll find something that works with it to test.

eiis1000:
I'm not super familiar with the way QMK/ZMK and these communications work, so this is a pretty layman's question, but to what extent are the remaining TODO items core functionality (which to me means "I can put a nice nano v2 on one half and a sea picro on the other half and the keypresses all register") vs nice-to-haves? More specifically, is this ready for beta testing with that core functionality?

M1cha:
I still intend to clean this up and merge this, I simply got distracted because I'm currently in my keyboard exploration phase to find the best one :sweat_smile: 

Assuming you configure your devicetree correctly, this should work for you as is. I've tested this on two (custom) keyboards so far.

Timoyoungster:
> I still intend to clean this up and make this mergeable.
Do you have some sort of estimate when you will probably get to it?
I'd also gladly help, if I can do something (do not have much experience with open-source yet but would love to get into it).

M1cha:
> Not sure how this and #2080 interact, but worth noting.

Thanks for mentioning this. It sounds like this needs to be coordinated better (by the ZMK maintainers?) The we'll work together instead of re-implementing the same thing over and over again and dumping WIP PRs in here.

That being said, from reading the description(I didn't read the code, yet) it looks like it supports more usecases.

jesseleite:
Excited about wired split support! For the layman who doesn't fully understand UART, what types of cables could be used to connect peripheral to central? TRRS? USB-C? RJ45?

silvergasp:
@M1cha I'd like to give this PR a go. Obviously as it's in it's early stages it's lacking some docs. Is your corne-keyboard repo the best example of this in action?

nmunnich:
Made obsolete by #2766

PR #1957, opened 2023-10-07
fix(nice_view): Fix incorrect union access

joelspadin:
Fixed an error in a previous commit where a member of the selected endpoint was used without checking if it was the correct transport type. The nice!view status screen displays the active BLE profile regardless of whether BLE is active, so we have to get that data directly from the BLE code instead of from the selected endpoint.


petejohanson (approved):
Code looks good, tester on Discord confirmed behavior.

PR #1958, opened 2023-10-07
docs: Add devicetree and kconfig syntax highlighting

joelspadin:
Added syntax highlighting for devicetree and kconfig files.

The PrismJS project is not accepting contributions right now as they work on a version 2 of the library, so the new language files are added directly here.

Also enabled syntax highlighting for various languages that are used in the docs but aren't enabled in Docusaurus by default.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,13 +8,13 @@ The ZMK Documentation is licensed [CC-BY-NC-SA](http://creativecommons.org/licen
 
 ### Installation
 
-```
+```sh
 $ npm ci
 ```
 
 ### Local Development
 
-```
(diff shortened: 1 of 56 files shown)


joelspadin:
Tweaked the syntax highlighting to more closely match the scopes used by highlight.js, and ported the github (with adjusted background color) and github-dark-dimmed themes from highlight.js. Note that the sources for these themes are under the BSD license.

caksoylar (approved):
LGTM! Unless anyone has concerns on the technical side re: adding our own syntax highlighters and colorschemes, I am down to merge.

PR #1961, opened 2023-10-09
feat(shields): add right encoders support to Waterfowl

DiogoDoreto:
Updated Waterfowl configs to comply with encoder sensors changes from #1841 

@JW2586 FYI :)


DiogoDoreto:
@petejohanson I confirm! I have been using it today and it's all good

PR #1965, opened 2023-10-13
feat(boards): Add nRF52 high voltage DC/DC config

joelspadin:
Added a Kconfig option to enable SOC_DCDC_NRF52X_HV for nice_nano_v2 and mikoto. According to Nordic's documentation, the DC/DC regulator is more efficient than the LDO regulator, so this is enabled by default.

The following boards do not support this mode and were not changed:

- nice_nano
- nice60
- nrfmicro_11, nrfmicro_13
- nrf52840_m2
- bluemicro840

I could not find schematics to confirm whether the following boards support this mode:

- bt60_v1, bt60_v2
- bt65_v1
- bt75_v1
- corneish_zen_v1, corneish_zen_v2
- pillbug
- puchi_ble_v1
- s40nc

### DIFF
--- a/app/boards/arm/nice_nano/Kconfig
+++ b/app/boards/arm/nice_nano/Kconfig
@@ -5,3 +5,9 @@ config BOARD_ENABLE_DCDC
     select SOC_DCDC_NRF52X
     default y
     depends on (BOARD_NICE_NANO || BOARD_NICE_NANO_V2)
+
+config BOARD_ENABLE_DCDC_HV
+    bool "High voltage DCDC converter"
+    select SOC_DCDC_NRF52X_HV
+    default y
+    depends on (BOARD_NICE_NANO_V2)
--- a/app/boards/arm/mikoto/Kconfig
+++ b/app/boards/arm/mikoto/Kconfig
@@ -4,6 +4,12 @@ config BOARD_ENABLE_DCDC
     default y
     depends on (BOARD_MIKOTO_520)
 
+config BOARD_ENABLE_DCDC_HV
+    bool "High voltage DCDC converter"
+    select SOC_DCDC_NRF52X_HV
+    default y
+    depends on (BOARD_MIKOTO_520)
+
 choice BOARD_MIKOTO_CHARGER_CURRENT
     prompt "Charge current to supply to attached batteries"
     depends on (BOARD_MIKOTO_520)


app/boards/arm/nice_nano/Kconfig
```diff
@@ -5,3 +5,9 @@ config BOARD_ENABLE_DCDC
     select SOC_DCDC_NRF52X
     default y
     depends on (BOARD_NICE_NANO || BOARD_NICE_NANO_V2)
+
+config BOARD_ENABLE_DCDC_HV
+    bool "High voltage DCDC converter"
+    select SOC_DCDC_NRF52X_HV
+    default y
+    depends on (BOARD_NICE_NANO_V2)
```
caksoylar, line 13: Does this cause a Kconfig warning when building for v1? Another option would be `default y if ...` or `imply ...` under `BOARD_...` Kconfig.

joelspadin: I don't see any warnings related to this in https://github.com/zmkfirmware/zmk/actions/runs/6512077486/job/17689044001?pr=1965

The goal is to only define this config for V2. V1 doesn't use high voltage mode, so there's no point in giving the option to enable this.

caksoylar: Right, I wondered if it gives a warning like "you tried to enable `SOC_DCDC_NRF52X_HV` but its dependency `BOARD_NICE_NANO_V2` was not satisfied" because of `default y`, even if you didn't force it specifically. But that doesn't seem to be the case.

joelspadin:
Verified on a nice!nano v2. I don't have a Mikoto to test.

zhiayang:
I can give this a try on mikoto — whats the expected result other than “it compiles and flashes”?

joelspadin:
Expected result is it compiles, flashes, doesn't brick it, and you may be get a few extra minutes of battery life?

Based on something Pete said in discord, if you try to enable this mode on something that doesn't have the necessary inductor, then it will effectively brick it until you apply 3.3v to VDD to bypass the now non-functional VDDH regulator. From looking at the schematic though, there should be no risk of that happening.

petejohanson:
> I can give this a try on mikoto — whats the expected result other than “it compiles and flashes”?

Have you had a chance to test this yet?

zhiayang:
> > I can give this a try on mikoto — whats the expected result other than “it compiles and flashes”?
> 
> Have you had a chance to test this yet?

Ah sorry, I did! It worked (nothing bad happened) but I forgot to update you guys x_x

petejohanson (approved):
Thanks!

Issue #1860, opened 2023-07-07
Encoder docs are out of date


joelspadin:
https://zmk.dev/docs/features/encoders has no information about adjusting properties for different types of encoders.

https://zmk.dev/docs/development/new-shield#encoders is missing the change from `resolution` to `steps` and doesn't show anything about `triggers-per-rotation` on the sensor node.

https://zmk.dev/docs/config/encoders is also missing those changes. We should probably add the `sensors` node configuration to that page too.

caksoylar:
> https://zmk.dev/docs/features/encoders has no information about adjusting properties for different types of encoders.

Can you elaborate a bit further on what properties?

> https://zmk.dev/docs/config/encoders is also missing those changes. We should probably add the sensors node configuration to that page too.

Would it make sense to convert that page to a "Sensors" page instead that has a "Encoders" section?

joelspadin:
> Can you elaborate a bit further on what properties?

`steps` for the alps,ec11 node and `triggers-per-rotation` for the sensors node and child nodes.

> Would it make sense to convert that page to a "Sensors" page instead that has a "Encoders" section?

Likely yes, though we don't yet support any other types of sensors.

PR #1970, opened 2023-10-18
Updated encoder config docs & standardization of existing shields

petejohanson:
* Update new shield guide for new sensor/encoder settings.
* Add DTS section to encoder config docs.
* Update existing in-tree shields/boards to use the new encoder configuration consistently.

### DIFF
--- a/app/boards/arm/bdn9/bdn9_rev2.dts
+++ b/app/boards/arm/bdn9/bdn9_rev2.dts
@@ -43,7 +43,7 @@
         label = "LEFT_ENCODER";
         a-gpios = <&gpioa 4 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
         b-gpios = <&gpioa 8 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
-        resolution = <4>;
+        steps = <80>;
         status = "disabled";
     };
     mid_encoder: encoder_mid {
@@ -51,7 +51,7 @@
         label = "MID_ENCODER";
         a-gpios = <&gpioa 9 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
         b-gpios = <&gpioa 10 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
-        resolution = <4>;
+        steps = <80>;
         status = "disabled";
     };
     right_encoder: encoder_right {
@@ -59,7 +59,7 @@
         label = "RIGHT_ENCODER";
         a-gpios = <&gpioa 15 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
         b-gpios = <&gpiob 3 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
-        resolution = <4>;
+        steps = <80>;
         status = "disabled";
     };
 
@@ -67,6 +67,7 @@
         compatible = "zmk,keymap-sensors";
         status = "disabled";
         sensors = <>;
+        triggers-per-rotation = <20>;
     };
 };
 
--- a/app/boards/shields/murphpad/murphpad.overlay
+++ b/app/boards/shields/murphpad/murphpad.overlay
@@ -39,7 +39,7 @@
         label = "Encoder 1";
         a-gpios = <&pro_micro 8 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
         b-gpios = <&pro_micro 7 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
-        resolution = <4>;
+        steps = <80>;
         status = "disabled";
     };
 
@@ -48,7 +48,7 @@
         label = "Encoder 2";
         a-gpios = <&pro_micro 1 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
         b-gpios = <&pro_micro 0 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
-        resolution = <4>;
+        steps = <80>;
         status = "disabled";
     };
 
--- a/docs/docs/config/encoders.md
+++ b/docs/docs/config/encoders.md
@@ -29,13 +29,55 @@ If `CONFIG_EC11` is enabled, exactly one of the following options must be set to
 
 ### Devicetree
 
+#### Keymap Sensor Config
+
+For shields/boards that export a `sensors` node configuration label, both global and per-sensor settings can be set by overriding the properties there.
+
+To override the general settings, update them on the exported `sensors` node, e.g.:
+
+```
+&sensors {
+    triggers-per-rotation = <18>;
+};
+```
+
+Per sensor overrides can be added with ordered nested nodes with the correct overrides, e.g.:
+
+```
+&sensors {
+    left_config {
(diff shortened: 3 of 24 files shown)


docs/docs/development/new-shield.md
```diff
@@ -435,26 +435,31 @@ left_encoder: encoder_left {
         label = "LEFT_ENCODER";
         a-gpios = <PIN_A (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
         b-gpios = <PIN_B (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
-        resolution = <4>;
+        steps = <20>;
```
caksoylar, line 438:
```suggestion
        steps = <80>;
```
How about changing the example here to a common default?

docs/docs/development/new-shield.md
```diff
@@ -435,26 +435,31 @@ left_encoder: encoder_left {
...
         status = "disabled";
     };
 ```
 
 Here you will have to replace PIN_A and PIN_B with the appropriate pins that your PCB utilizes for the encoder(s). For keyboards that use the Pro Micro or any of the Pro Micro replacements, Sparkfun's [Pro Micro Hookup Guide](https://learn.sparkfun.com/tutorials/pro-micro--fio-v3-hookup-guide/hardware-overview-pro-micro) has a pinout diagram that can be useful to determine the right pins. Reference either the blue numbers labeled "Arduino" (digital pins) or the green numbers labeled "Analog" (analog pins). For pins that are labeled as both digital and analog, refer to your specific board's .dtsi file to determine how you should refer to that pin.
```
caksoylar, line 443:
```suggestion
Here you need to replace `PIN_A` and `PIN_B` with the appropriate pins that your PCB utilizes for the encoder(s). See [shield overlays section above](#shield-overlays) on the appropriate node label and pin number to use for GPIOs.
```
This seems a bit outdated, what do you think of this? (Also need to double check the section link works.)

docs/docs/development/new-shield.md
```diff
@@ -435,26 +435,31 @@ left_encoder: encoder_left {
...
 
+The `steps` property should corresponded to the documented pulses per rotation for the encoders used on the keyboard. If users use different encoders when they build, the value can be overridden in their keymap.
```
caksoylar, line 445:
```suggestion
The `steps` property should corresponded to the documented pulses per rotation for the encoders used on the keyboard, typically found on the datasheet of the component. If users use different encoders when they build, the value can be overridden in their keymap.
```
Would it be fair and/or useful to point this out specifically?

docs/docs/development/new-shield.md
caksoylar: Also, can you fix the indentation so the first line is consistent with the other lines?

caksoylar (commented):
Can we also add a pointer to docs/features/encoders.md to the config page, for users that need to adjust the encoder properties in the keymap? Right now it says
> See the [New Keyboard Shield](https://zmk.dev/docs/development/new-shield#encoders) documentation for how to add or modify additional encoders to your shield.

but I feel like config page is better for adjusting already configured encoders.

caksoylar:
This would close #1860.

caksoylar:
Also a bit unrelated but since we are touching encoder definitions in all shields: Should we get rid of the recommendation to set `CONFIG_EC11_TRIGGER_GLOBAL_THREAD=y` by making it default? I don't know the nuances of when you wouldn't want that, but I don't think I have seen a config with it not used.

petejohanson:
> 1. ...I was surprised to see a PR which was ostensibly for docs incorporate changes to shields and boards.

Yeah, the PR title/summary got auto-populated by my first commit, and then I got to the follow up work. I will update accordingly.

> 
>     2. The suggested changes for MurphPad are at odds with [my outstanding PR](https://github.com/zmkfirmware/zmk/pull/1786), which has been tested on an actual PCB and stock encoder provided by the vendor/creator. (I am also unsure if there is some reason its keymap was not modified—does the `sensors` node actually belong in the `$shieldname.overlay` file?)

So, we have a *mix* of approaches, probably in part because we weren't very diligent about how to structure things when encoder work got added.

IMHO, the "correct" way to do this is to add the sensors node in the overlay, and give it a label, e.g. `sensors: sensors {` so that it can be *referred to* in keymaps and updated as necessary.

So if you only installed encoders in certain positions, you might do:

```
&sensors {
  sensors = <&position_one_encoder>;
};
``` 

for instance. I didn't want to "rock the boat" too much as part of this refactor, but I'm certainly open to it if we want to be consistent so that future shields/boards that might copy others get it "right".

app/boards/arm/bdn9/bdn9_rev2.dts
```diff
@@ -43,30 +43,31 @@
...
-        resolution = <4>;
+        steps = <20>;
         status = "disabled";
     };
 
     sensors: sensors {
         compatible = "zmk,keymap-sensors";
         status = "disabled";
         sensors = <>;
+        triggers-per-rotation = <20>;
```
joelspadin, line 70: 20 steps and 20 triggers per rotation doesn't seem like it would be a correct value for many encoders. Did you mean to set steps to 80 everywhere?

petejohanson: I did. Updated.

app/boards/shields/murphpad/murphpad.overlay
```diff
@@ -48,7 +48,7 @@
         label = "Encoder 2";
         a-gpios = <&pro_micro 1 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
         b-gpios = <&pro_micro 0 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
-        resolution = <4>;
+        steps = <20>;
```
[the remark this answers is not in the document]
petejohanson: Indeed. Fixed.

docs/docs/development/new-shield.md
petejohanson: I like it. Merged this in, and confirmed in the deploy previous that the section link does work as expected.

app/boards/shields/murphpad/murphpad.overlay
caksoylar, line 1: This one doesn't seem to have `triggers-per-rotation` defined (`sensors` node is in the keymap)

caksoylar (approved):
LGTM, just noting the missing tweak for `murphpad`.

docs/docs/config/encoders.md
```diff
@@ -29,13 +29,49 @@ If `CONFIG_EC11` is enabled, exactly one of the following options must be set to
...
+
+To override the general settings, update them on the exported `sensors` node, e.g.:
+
+```
+&sensors {
+    triggers-per-rotation = <18>;
+};
+```
+
+Per sensor overrides can be added with ordered nested nodes with the correct overrides, e.g.:
```
joelspadin, line 44: Might be useful to note that these are listed in the order that the sensors are defined in the board/shield, and the names don't matter.

joelspadin (commented):
Looks good to me aside from the issue with murphpad missing triggers-per-rotation.

docs/docs/config/encoders.md
petejohanson: Added a note about this.

PR #1971, opened 2023-10-18
feat(docs): Document ZMK_BATTERY_REPORTING config

ReFil:
Whilst working on other battery reporting stuff i noticed `CONFIG_ZMK_BATTERY_REPORTING` is currently undocumented, A new KConfig heading for battery.md has been added in line with the other sections in the configuration section of the docs, `CONFIG_ZMK_BATTERY_REPORT_INTERVAL` has been moved from system to battery for consistency.

### DIFF
--- a/docs/docs/config/battery.md
+++ b/docs/docs/config/battery.md
@@ -7,6 +7,21 @@ See the [battery level feature page](../features/battery.md) for more details on
 
 See [Configuration Overview](index.md) for instructions on how to change these settings.
 
+### Kconfig
+
+Definition file: [zmk/app/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/Kconfig)
+
+| Config                               | Type | Description                                            | Default |
+| ------------------------------------ | ---- | ------------------------------------------------------ | ------- |
+| `CONFIG_ZMK_BATTERY_REPORTING`       | bool | Enables/disables all battery level detection/reporting | n       |
+| `CONFIG_ZMK_BATTERY_REPORT_INTERVAL` | int  | Battery level report interval in seconds               | 60      |
+
+:::note Default setting
+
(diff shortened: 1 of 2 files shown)


docs/docs/config/battery.md
```diff
@@ -7,6 +7,15 @@ See the [battery level feature page](../features/battery.md) for more details on
 
 See [Configuration Overview](index.md) for instructions on how to change these settings.
 
+### Kconfig
+
+Definition file: [zmk/app/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/Kconfig)
+
+| Config                               | Type | Description                                            | Default |
+| ------------------------------------ | ---- | ------------------------------------------------------ | ------- |
+| `CONFIG_ZMK_BATTERY_REPORTING`       | bool | Enables/disables all battery level detection/reporting | n       |
```
caksoylar, line 16: While it is technically true that this is disabled by default, it is `imply`d by `ZMK_BLE` so in practice this will typically be true. In order not to confuse users (because they check these pages and think "oh I should enable that") can you add a note below the table that this will be set to true if BLE is enabled?

docs/docs/config/battery.md
```diff
@@ -7,6 +7,21 @@ See the [battery level feature page](../features/battery.md) for more details on
...
+| `CONFIG_ZMK_BATTERY_REPORT_INTERVAL` | int  | Battery level report interval in seconds               | 60      |
+
+:::note Default setting
+
+Whilst this is disabled by default it is implied by `CONFIG_ZMK_BLE`, thus any board with BLE enabled will have this automatically enabled unless explicitly overriden.
```
caksoylar, line 21:
```suggestion
While `CONFIG_ZMK_BATTERY_REPORTING` is disabled by default it is implied by `CONFIG_ZMK_BLE`, thus any board with BLE enabled will have this automatically enabled unless explicitly overriden.
```

caksoylar (changes requested):

caksoylar (approved):

PR #1026, opened 2021-11-15
Add nix-shell

bromanko:
I prefer using [nix](https://nixos.org/manual/nix/stable/) to manage my development environments and dependencies. This PR adds a `nix-shell` config with development dependencies, and a `direnv` configuration to automatically source it when entering the directory. 

I am curious if this is a helpful addition. If so, perhaps someone else can test it and we can add some reference to it in the documentation for local development setup.

PR #1974, opened 2023-10-21
Add Nix flake

adisbladis:
This is another take on #1026 that aims to minimally implement a Nix development environment.

Notable differences are:

- Uses `pyproject.nix` for auto-packaging of the environment
  - Not only is mach-nix unmaintained, but it's also very heavy weight.

  - Uses nixpkgs Python packages
  Which means better binary cache performance than mach-nix offers.

- No direnv additions 
  I use direnv myself but expect users to be able to `echo 'use flake' > .envrc` if they want to use direnv.

- Uses zephyr sources from a flake input 
  This should make maintenance simpler as there is no manual updating required.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,52 @@
+{
+  description = "# Zephyr™ Mechanical Keyboard (ZMK) Firmware";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+    flake-utils.url = "github:numtide/flake-utils";
+
+    # Customize the version of Zephyr used by the flake here
+    zephyr = {
+      type = "github";
+      owner = "zmkfirmware";
+      repo = "zephyr";
+      ref = "v3.5.0+zmk-fixes";
+      flake = false;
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,49 @@
+{
+  description = "# Zephyr™ Mechanical Keyboard (ZMK) Firmware";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+    flake-utils.url = "github:numtide/flake-utils";
+
+    # Customize the version of Zephyr used by the flake here
+    zephyr.url = "github:zephyrproject-rtos/zephyr/v3.2.0";
```
adamcstephens, line 9: Should this be the zmk fork instead? It can also be updated to 3.5 now too, https://github.com/zmkfirmware/zmk/pull/1995

adisbladis: Yes, it should! Fixed.

kilipan:
Would love for a nix flake to be added, too!

PR #1975, opened 2023-10-22
chore(docs): Fix links to relocated files

nickcoutsos:
Some of the driver code directory structure was reorganized recently to match Zephyr better, but the existing links in the documentation hadn't been updated.

I _think_ this is everything, after some manual checking I looked at the PR where the files were moved and used that to look for affected files.

### DIFF
--- a/docs/docs/config/battery.md
+++ b/docs/docs/config/battery.md
@@ -48,7 +48,7 @@ Driver for reading the voltage of a battery using a Nordic nRF52's VDDH pin. Thi
 
 Applies to: `compatible = "zmk,battery-nrf-vddh"`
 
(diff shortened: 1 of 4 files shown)


caksoylar (changes requested):
Thank you for the fixes, Nick! Can you also fix the link to the header at https://github.com/zmkfirmware/zmk/blob/main/docs/docs/config/kscan.md?plain=1#L287? It looks like it should now go to https://github.com/zmkfirmware/zmk/blob/main/app/module/include/dt-bindings/zmk/kscan_mock.h instead.

nickcoutsos:
Got it, thanks

caksoylar (approved):

PR #1990, opened 2023-11-07
fix: save the build log when running tests

ecstatic-morse:
...and direct the user there if something goes wrong.

### DIFF
--- a/app/run-test.sh
+++ b/app/run-test.sh
@@ -30,7 +30,7 @@ mkdir -p build/$testcase
 west build -d build/$testcase -b native_posix_64 -- -DZMK_CONFIG="$(pwd)/$testcase" >build/$testcase/build.log 2>&1
 if [ $? -gt 0 ]; then
     echo "FAILED: $testcase did not build" | tee -a ./build/tests/pass-fail.log
-    echo $'\t' "see ./build/$testcase/build.log"
+    printf '\t%s\n' "see ./build/$testcase/build.log"
     exit 1
 fi
 
--- a/app/run-test.sh
+++ b/app/run-test.sh
@@ -26,9 +26,11 @@ fi
 testcase="$path"
 echo "Running $testcase:"
 
-west build -d build/$testcase -b native_posix_64 -- -DZMK_CONFIG="$(pwd)/$testcase" > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


app/run-test.sh
```diff
@@ -26,9 +26,11 @@ fi
 testcase="$path"
 echo "Running $testcase:"
 
-west build -d build/$testcase -b native_posix_64 -- -DZMK_CONFIG="$(pwd)/$testcase" > /dev/null 2>&1
+mkdir -p build/$testcase
+west build -d build/$testcase -b native_posix_64 -- -DZMK_CONFIG="$(pwd)/$testcase" >build/$testcase/build.log 2>&1
 if [ $? -gt 0 ]; then
     echo "FAILED: $testcase did not build" | tee -a ./build/tests/pass-fail.log
+    echo $'\t' "see ./build/$testcase/build.log"
```
caksoylar, line 33:
```suggestion
    echo '\t'"see ./build/$testcase/build.log"
```
`$'..'` seems to be a bash-ism, doesn't work with `dash` in Ubuntu here.

ecstatic-morse: Can't say I write much crossplatform shell script :smiley: . Your version prints `\t` in bash. How about

```suggestion
    printf '\t%s' "see ./build/$testcase/build.log"
```

caksoylar: That's probably the way to go, yeah.

caksoylar: Thanks for updating! To give some context to my initial comment: the script has shebang `#!/bin/sh`, which is only guaranteed to be a POSIX-compatible shell. That is *usually* not bash, for instance in Debian-based systems it will be [dash](https://en.wikipedia.org/wiki/Almquist_shell#dash). Not all POSIX-compatible shells support the same features, but they are all supposed to support the POSIX spec. So `#!/bin/sh` scripts should avoid bash-isms (or dash-isms, as it turns out). (https://github.com/dylanaraps/pure-sh-bible is a great resource if you are interested)

petejohanson:
Can you please rebase this to address the conflict? Thanks!

petejohanson:
@ecstatic-morse Did you see my request to rebase?

PR #1991, opened 2023-11-07
Pointer Core: HID, mouse keys behavior, etc.

petejohanson:
Taking the amazing starting points in #778 which was then updated in #1430 and trying to get this across the finish line!

Still needs squashing, but getting this up as a draft for others to focus testing on.

In particular, I have addressed the issues with Windows compatibility by:

* Limiting to 5 buttons, as someone on Discord noted might help,
* Re-adding X/Y/Wheel data/descriptors; without this Windows is grumpy still.

In addition, this is rebased on latest ZMK `main` to avoid any conflicts. Now that I've gottem basics working locally building on others' great work, I will start reviewing more deeply with the intent of marking this non-draft, then getting this starting point in.

Testers welcome!


ReFil:
Would changing this to allow the x, y and wheel reports to be easily set and sent be within the scope of this PR? I think that would be really handy for pointer device implementation

petejohanson:
> Would changing this to allow the x, y and wheel reports to be easily set and sent be within the scope of this PR? I think that would be really handy for pointer device implementation

Given how long the pointer work has stalled, I'm hesitant to increase scope here much more. I'd rather we get the working baseline with mouse keys in, then iterate on the API for x/y/wheel. My main concern is whether we need to handle absolute reporting any time soon, and if so, designing this flexibly from the get go for it.

app/include/zmk/endpoints.h
```diff
@@ -69,3 +69,4 @@ int zmk_endpoints_toggle_transport(void);
 struct zmk_endpoint_instance zmk_endpoints_selected(void);
 
 int zmk_endpoints_send_report(uint16_t usage_page);
+int zmk_endpoints_send_mouse_report();
```
bryanforbes, line 72: Should this be guarded with `#if IS_ENABLED(CONFIG_ZMK_MOUSE)`?

app/src/endpoints.c
```diff
@@ -193,6 +193,35 @@ int zmk_endpoints_send_report(uint16_t usage_page) {
     return -ENOTSUP;
 }
 
+int zmk_endpoints_send_mouse_report() {
```
bryanforbes, line 196: Should this be guarded with `#if IS_ENABLED(CONFIG_ZMK_MOUSE)`?

app/include/zmk/hid.h
```diff
@@ -10,6 +10,7 @@
 #include <zephyr/usb/class/usb_hid.h>
 
 #include <zmk/keys.h>
+#include <zmk/mouse.h>
```
bryanforbes, line 13: Along the same lines as my other comments, should the additions here be guarded?

app/include/zmk/hid.h
```diff
@@ -89,6 +90,37 @@ static const uint8_t zmk_hid_report_desc[] = {
     /* INPUT (Data,Ary,Abs) */
     HID_INPUT(0x00),
     HID_END_COLLECTION,
+
+    HID_USAGE_PAGE(HID_USAGE_GD),
+    HID_USAGE(HID_USAGE_GD_MOUSE),
+    HID_COLLECTION(HID_COLLECTION_APPLICATION),
+    HID_REPORT_ID(0x04),
```
bryanforbes, line 97: Should `0x04` (and the other report ID constants) be made into `#define`s and then used in `hid.c` in the report structs?

app/src/hog.c
```diff
@@ -56,6 +56,11 @@ static struct hids_report consumer_input = {
     .type = HIDS_INPUT,
 };
 
+static struct hids_report mouse_input = {
+    .id = 0x04,
```
bryanforbes, line 60: Here is another place where the report constants could be used

app/include/zmk/hid.h
petejohanson: Agreed. Also took the opportunity to refactor our main item values to use bit flag constants as well for clarity, which has been on my wishlist for this code.

app/src/hog.c
petejohanson: Indeed. Updated.

caksoylar:
@petejohanson Are you intending to update the docs before merging this? I think it would be good to either (a) hide the docs for now, until full mouse keys support is merged (move/scroll included), or (b) note more specifically that this is currently WIP and only supports button presses. 

I also don't like linking to the header file for the actually useful information, but we can fix that later.

petejohanson:
> @petejohanson Are you intending to update the docs before merging this? I think it would be good to either (a) hide the docs for now, until full mouse keys support is merged (move/scroll included), or (b) note more specifically that this is currently WIP and only supports button presses.
> 
> I also don't like linking to the header file for the actually useful information, but we can fix that later.

I haven't had a chance to even look at the docs yet after grabbing the existing work as a starting point. Definitely will review and adjust/remove/update as necessary. Thanks for pointing it out.

petejohanson:
> @petejohanson Are you intending to update the docs before merging this? I think it would be good to either (a) hide the docs for now, until full mouse keys support is merged (move/scroll included), or (b) note more specifically that this is currently WIP and only supports button presses.
> 
> I also don't like linking to the header file for the actually useful information, but we can fix that later.

@caksoylar Took a look and cleaned up the added page a ton. Thoughts?

caksoylar:
> > @petejohanson Are you intending to update the docs before merging this? I think it would be good to either (a) hide the docs for now, until full mouse keys support is merged (move/scroll included), or (b) note more specifically that this is currently WIP and only supports button presses.
> > I also don't like linking to the header file for the actually useful information, but we can fix that later.
> 
> @caksoylar Took a look and cleaned up the added page a ton. Thoughts?

The wording looks great, but I think we should include a table like below under "Behavior Binding," what do you think?

| Define | Action |
| - | - |
| `MB1`, `LCLK` | Left click |
| `MB2`, `RCLK` | Right click |
| `MB3`, `RCLK` | Middle click |
| `MB4` | Mouse button 4 |
| `MB5` | Mouse button 5 |

Mouse buttons 4 and 5 typically map to "back" and "forward" mappings in most applications.

petejohanson:
> > > @petejohanson Are you intending to update the docs before merging this? I think it would be good to either (a) hide the docs for now, until full mouse keys support is merged (move/scroll included), or (b) note more specifically that this is currently WIP and only supports button presses.
> > > I also don't like linking to the header file for the actually useful information, but we can fix that later.
> > 
> > 
> > @caksoylar Took a look and cleaned up the added page a ton. Thoughts?
> 
> The wording looks great, but I think we should include a table like below under "Behavior Binding," what do you think?
> Define 	Action
> `MB1`, `LCLK` 	Left click
> `MB2`, `RCLK` 	Right click
> `MB3`, `MCLK` 	Middle click
> `MB4` 	Mouse button 4
> `MB5` 	Mouse button 5
> 
> Mouse buttons 4 and 5 typically map to "back" and "forward" actions in most applications.

Great call. Updated that section of that page to include that table and the note about mouse buttons 4 and 5.

docs/docs/behaviors/mouse-emulation.md
```diff
@@ -0,0 +1,64 @@
...
+the [`dt-bindings/zmk/mouse.h`](https://github.com/zmkfirmware/zmk/blob/main/app/include/dt-bindings/zmk/mouse.h) header
+provided by ZMK near the top:
+
+```
+#include <dt-bindings/zmk/mouse.h>
+```
+
+The following defines are available:
+
+| Define        | Action         |
```
caksoylar, line 33: I would put this table under "Behavior Binding" because they apply to `&mkp` behavior, and we can remove the "The following defines are available:" text in that case as well. What do you think?

petejohanson: Moved.

docs/docs/behaviors/mouse-emulation.md
```diff
@@ -0,0 +1,64 @@
...
+## Configuration Option
+
+This feature can be enabled via a config option:
+
+```
+CONFIG_ZMK_MOUSE=y
+```
+
+If you use the mouse key press behavior in your keymap, the feature will automatically be enabled for you.
+
```
caksoylar, line 20:
```suggestion
```
Do you think we should remove this section altogether? It doesn't sound relevant to the end users if it gets automatically enabled.

docs/docs/behaviors/mouse-emulation.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Mouse Emulation Behaviors
+sidebar_label: Mouse Emulation
+---
+
+## Summary
+
+Mouse emulation behaviors send mouse events. Currently, only mouse button presses are supported, but movement
+and scroll action support is planned for the future.
+
```
caksoylar, line 10:
```suggestion

:::caution
This feature will be automatically enabled when you use any of the following behaviors in your keymap and might require re-pairing your keyboard to your devices to work over bluetooth. If mouse behaviors are still not recognized by your device after doing that, you can try [these troubleshooting steps](../features/bluetooth.md#windows-connected-but-not-working).
:::
```
I think this is something that folks are likely to hit, to refresh the HID descriptors (should we mention that explicitly?)

caksoylar (commented):
A couple more comments on the docs side, sorry about the piecemeal review!

As an aside, I haven't had a chance to test it yet but I can try to make some time tonight.

docs/docs/behaviors/mouse-emulation.md
petejohanson: I wonder if we might want to call that out explicitly near the top of this doc page?

caksoylar: You mean the HID descriptor change? Might be a good idea, though I am not confident enough on the details myself. Maybe worth expanding my caution note.

petejohanson: @caksoylar Pushed some tweaks. Thoughts?

caksoylar: LGTM, thanks.

caksoylar (approved):

caksoylar:
Tested and working over USB and BLE on a nice!nano split on Windows. I did have to do the https://deploy-preview-1991--zmk.netlify.app/docs/features/bluetooth#windows-connected-but-not-working trick twice to get the keyboard working over BLE after the mouse keys flash; I think I might not have waited enough for BT to turn off the first time.

app/src/behaviors/behavior_mouse_key_press.c
```diff
@@ -0,0 +1,48 @@
...
+    LOG_DBG("position %d keycode 0x%02X", event.position, binding->param1);
+    return ZMK_EVENT_RAISE(
+        zmk_mouse_button_state_changed_from_encoded(binding->param1, false, event.timestamp));
+}
+
+static const struct behavior_driver_api behavior_mouse_key_press_driver_api = {
+    .binding_pressed = on_keymap_binding_pressed, .binding_released = on_keymap_binding_released};
+
+#define KP_INST(n)                                                                                 \
+    DEVICE_DT_INST_DEFINE(n, behavior_mouse_key_press_init, device_pm_control_nop, NULL, NULL,     \
```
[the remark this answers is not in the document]
petejohanson: Good catch. That changed as part of the Zephyr upgrade recently. Fixing.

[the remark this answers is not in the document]
petejohanson: Fixed. Confirmed compiled fine here after turning on deep sleep.

PR #1994, opened 2023-11-09
refactor: address transport switch enumeration warning

xudongzheng:
When building without USB or Bluetooth, the compiler would emit a warning due to ZMK_TRANSPORT_USB or ZMK_TRANSPORT_BLE not being handled.

### DIFF
--- a/app/src/endpoints.c
+++ b/app/src/endpoints.c
@@ -122,57 +122,69 @@ struct zmk_endpoint_instance zmk_endpoints_selected(void) {
 
 static int send_keyboard_report(void) {
     switch (current_instance.transport) {
-#if IS_ENABLED(CONFIG_ZMK_USB)
     case ZMK_TRANSPORT_USB: {
+#if IS_ENABLED(CONFIG_ZMK_USB)
         int err = zmk_usb_hid_send_keyboard_report();
         if (err) {
             LOG_ERR("FAILED TO SEND OVER USB: %d", err);
         }
         return err;
-    }
+#else
+        LOG_ERR("USB endpoint is not supported");
+        return -ENOTSUP;
 #endif /* IS_ENABLED(CONFIG_ZMK_USB) */
+    }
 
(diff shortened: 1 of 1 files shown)


app/src/endpoints.c
```diff
@@ -153,26 +153,26 @@ static int send_consumer_report(void) {
     struct zmk_hid_consumer_report *consumer_report = zmk_hid_get_consumer_report();
 
     switch (current_instance.transport) {
-#if IS_ENABLED(CONFIG_ZMK_USB)
     case ZMK_TRANSPORT_USB: {
+#if IS_ENABLED(CONFIG_ZMK_USB)
```
petejohanson, line 157: Doesn't this cause this to fall through to the next label with this change? Maybe better to only define the given transport in the enum if that feature is enabled, instead?

joelspadin: `enum zmk_transport` is written to settings, so conditionally defining values could cause the meaning of a saved value to change from one build to another. Sort of a theoretical problem when we only have two transports, and disabling one means you can't switch at all, but that could become a problem if we ever added another RF protocol as a transport for example.

I'd recommend changing these to something more like
```c
case ZMK_TRANSPORT_USB:
#if IS_ENABLED(CONFIG_ZMK_USB)
    return ...
#else
    return -ENOTSUP;
#endif
```
to avoid the fallthrough.

Alternatively, you could add a default case, but then you wouldn't get a warning if a new transport is added and someone forgot to add handling for it somewhere.

petejohanson: Good call on the settings storage issue. I agree with the `-ENOTSUP` version you proposed.

joelspadin (approved):
Looks good to me, Some existing error messages could be improved a bit if you want though.

petejohanson:
Can you please rebase this so we can get it merged? Updated version LGTM.

xudongzheng:
Rebased as requested, and the mouse report should be handled as well.

PR #2002, opened 2023-11-11
fix(docs): Fix debouncing driver support note

caksoylar:
Small fix on an outdated note.


petejohanson (approved):
Thanks! Noticed this the other day too.

PR #2004, opened 2023-11-11
feat(build): Add support for artifact-name in build.yaml

caksoylar:
This change adds a new optional property `artifact-name` to `build.yaml`, so that if provided, the base of the output filename will be equal to that for an entry.

While it isn't discoverable because there isn't a way to document it (besides perhaps expanding the comment section in the template yaml), this field would allow users to build the same board+shield combination with different keymaps or different CMake args (like adding Kconfig settings).

Some context: https://discord.com/channels/719497620560543766/719909884769992755/1171174369045774406

[Example build.yaml](https://github.com/caksoylar/zmk-config/blob/test-artifact-name/build.yaml): 
```yaml
---
include:
  - board: seeeduino_xiao_ble
    shield: rommana_left rgbled_widget
    artifact-name: my_rommana_left
  - board: seeeduino_xiao_ble
    shield: rommana_right rgbled_widget
    artifact-name: my_rommana_right
  - board: nice_nano
    shield: corne_left
    artifact-name: stock_corne_left
  - board: nice_nano
    shield: corne_left
    cmake-args: -DKEYMAP_FILE=/__w/zmk-config/zmk-config/config/corneish_zen.keymap
```
Output from example: https://github.com/caksoylar/zmk-config/actions/runs/6837199541
image

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -55,12 +55,14 @@ jobs:
       - name: Prepare variables
         shell: sh -x {0}
         env:
+          board: ${{ matrix.board }}
           shield: ${{ matrix.shield }}
+          artifact_name: ${{ matrix.artifact-name }}
         run: |
           echo "zephyr_version=${ZEPHYR_VERSION}" >> $GITHUB_ENV
           echo "extra_cmake_args=${shield:+-DSHIELD=\"$shield\"}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


petejohanson:
> A tangentially-related idea: #1379

Good catch. Of the two, this feels more immediately useful, so merging this first, then we can revisit #1379`

petejohanson (approved):
LGTM. Thanks for the link to an example build using this. We should update the examples in the comment in the unified template repo now that this is in.

caksoylar:
Thanks, I can look into updating the example as well.

PR #2006, opened 2023-11-13
fix(usb): Tweak how ZMK_USB gets enabled.

petejohanson:
* Previous version of multiple overrides of the default value of ZMK_USB were problematic. Move to using board _defconfig files for the defaults for those, along with proper `depends on` for ZMK_USB that accounts for split and split roles.

This addresses a current build failure in ZMK main when building for a controller like proton-c with a split right side. Although not important now, this should get cleaned up in preparation for eventual wired split support. See an example failure here: https://github.com/zmkfirmware/zmk/actions/runs/6853902894/job/18635806769


petejohanson:
Many of the tests are green, and this is hitting production split builds, so merging this to stem the pain.

PR #2014, opened 2023-11-14
Add a basic shell to ZMK and use it for testing

ecstatic-morse:
Includes #1318.

Running tests using the shell allows testing multiple input sequences for a single configuration without rebuilding the configuration each time. As an example, I've converted the `toggle-layer` test to use ZMK shell commands.


ecstatic-morse:
This works but the configuration is still a bit wonky.

Obviously low priority, but this should speed up writing and running tests quite a bit.

PR #1973, opened 2023-10-20
feat(ble): Allow disabling BLE BAS reporting

ReFil:
The battery reporting has been known to cause macOS computers to wakeup repeatedly. In some cases (e.g. display or custom lighting implementation) one might want to collect battery SOC without broadcasting over BLE

PR #2022, opened 2023-11-15
feat(ble): Only update BAS when active

ReFil:
As a further mitigation against spurious wakeups (in addition to #1973) without having to disable the battery service entirely, the battery percentage can only be updated when the board is active i.e. when in use, which infers the computer is also in use

### DIFF
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -18,6 +18,8 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/event_manager.h>
 #include <zmk/battery.h>
 #include <zmk/events/battery_state_changed.h>
+#include <zmk/events/activity_state_changed.h>
+#include <zmk/activity.h>
 #include <zmk/workqueue.h>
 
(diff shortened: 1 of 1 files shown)


petejohanson (changes requested):
One high level comment on the implementation of this. Thanks!

app/src/battery.c
```diff
@@ -105,4 +107,26 @@ static int zmk_battery_init(const struct device *_arg) {
     return 0;
 }
 
+static int battery_event_listener(const zmk_event_t *eh) {
+
+    if (as_zmk_activity_state_changed(eh)) {
+        switch (zmk_activity_get_state()) {
+        case ZMK_ACTIVITY_ACTIVE:
+            k_timer_start(&battery_timer, K_NO_WAIT, K_SECONDS(CONFIG_ZMK_BATTERY_REPORT_INTERVAL));
```
petejohanson, line 115: You should pull this and the duplicates line from the init function into a small local `start_reporting` function.

petejohanson (changes requested):
One code organization comment.

petejohanson (approved):
Thanks!

Issue #1460, opened 2022-09-20
F13-F24 not functioning when NKRO enabled


ReFil:
With CONFIG_ZMK_HID_REPORT_TYPE_NKRO=y set F13-F24 don't register on the host. Commenting out that line makes the keys work with no other changes

Hello9999901:
What operating system? It is stated in the documentation that NKRO has different effects on different OSes. Personally, NKRO does not work at all on macOS.

ai212983:
@Hello9999901 Can you please point me to where exactly in the docs it is stated? [Codes page](https://zmk.dev/docs/codes#f-keys) lists F13-F24 as not supported on Android only. System configuration page [says](https://zmk.dev/docs/config/system#hid) 

> CONFIG_ZMK_HID_REPORT_TYPE_NKRO | Enable full N-key roll over. This may prevent the keyboard from working with some BIOS/UEFI versions.

Can't find anything about NKRO having "different effects on different OSes".

sr2r:
thanks for logging this issue, got it fixed after setting CONFIG_ZMK_HID_REPORT_TYPE_NKRO from `y` to `n`

ReFil:
This is the same as issue #1236, It was fixed by changing this line to use a larger HID usage https://github.com/zmkfirmware/zmk/blob/82e85699edd7261500ea2feb0697f8a6a5bbd42c/app/include/zmk/hid.h#L16

caksoylar:
@ReFil just to confirm, you are saying it works with `HID_USAGE_KEY_KEYPAD_EQUAL`, or that it works when changed to a different value? That line seems to be changed two years ago: https://github.com/zmkfirmware/zmk/commit/91ba034896c3b1668b88286e18d8f38314ec39c3

ReFil:
I had to change it to a different value, I used `HID_USAGE_KEY_KEYPAD_000` because it's greater than the f13-f24 and international Lang keys I wanted to make work

caksoylar:
Would you like to keep this open in that case, since (my understanding is) it isn't resolved?

ReFil:
Closing as this was solved in #2024

Issue #1236, opened 2022-04-11
International keys & Launguage keys not working on macOS


jdack41:
INT_RO, INT_YEN, LANG1, LANG2 not working on macOS, but worked on Windows(both of them set on Japanese environment).
These keys used on JIS layout mac keyboards.

yemartin:
With a Kinesis Advantage360 Pro, even with the fix introduced by #1521, `LANG1` and `LANG2` did not work out of the box on macOS. `Key Codes.app` would not show any output when pressing these keys.

The Kinesis firmware builder tool has [NKRO enabled by default](https://github.com/KinesisCorporation/Adv360-Pro-ZMK/blob/09ff934f7d2a42e3b0b8e3f65595732536baf72d/config/boards/arm/adv360/adv360_left_defconfig#L53). But I had to [turn NKRO off](https://github.com/yemartin/Adv360-Pro-ZMK/commit/05840e690a687ea27ff743488837c48d50b59978) to get `LANG1` and `LANG2` to work (thanks to @caksoylar invaluable help and [spot-on suggestion](https://discord.com/channels/719497620560543766/1078162827983990835/1078201526096498768) on the Discord forum).

Does this mean NKRO needs a fix similar to HKRO's #1521 ?

yemartin:
Pinging @jhorology and @petejohanson (from #1521) for visibility. Do you have any idea about this?

petejohanson:
> With a Kinesis Advantage360 Pro, even with the fix introduced by #1521, `LANG1` and `LANG2` did not work out of the box on macOS. `Key Codes.app` would not show any output when pressing these keys.
> 
> The Kinesis firmware builder tool has [NKRO enabled by default](https://github.com/KinesisCorporation/Adv360-Pro-ZMK/blob/09ff934f7d2a42e3b0b8e3f65595732536baf72d/config/boards/arm/adv360/adv360_left_defconfig#L53). But I had to [turn NKRO off](https://github.com/yemartin/Adv360-Pro-ZMK/commit/05840e690a687ea27ff743488837c48d50b59978) to get `LANG1` and `LANG2` to work (thanks to @caksoylar invaluable help and [spot-on suggestion](https://discord.com/channels/719497620560543766/1078162827983990835/1078201526096498768) on the Discord forum).
> 
> Does this mean NKRO needs a fix similar to HKRO's #1521 ?

[...]

PR #2024, opened 2023-11-15
feat(hid): Add KConfig option for higher NKRO usages

ReFil:
By default, the maximum NKRO usage is set to maximize compatibility, but some keys don't work. This adds the ability to use those extended keys, at the cost of compatibility (android doesn't like the higher usage range, it seems). The chosen max was largely arbitrary, although it needs to be a multiple of 8 - 1 (otherwise windows will refuse to enumerate it properly, reporting "driver error"), For this reason I made it a configurable bool rather than a variable value. There's also no reason to have it variable across a wide range, a low and a high value seem like a good compromise

This fix addresses #1460 and is linked to #1236

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -53,9 +53,18 @@ config ZMK_HID_REPORT_TYPE_NKRO
     help
       Enable full N-Key Roll Over for HID output. This selection will prevent the keyboard
       from working with some BIOS/UEFI versions that only support "boot keyboard" support.
-      This option also prevents using some infrequently used higher range HID usages.
+      This option also prevents using some infrequently used higher range HID usages (notably F13-F24 and INTL1-9)
+      These usages can be re enabled with ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT.
 
 endchoice
+
+if ZMK_HID_REPORT_TYPE_NKRO
+
+config ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT
+    bool "Enable extended NKRO reporting"
+    default n
+
+endif
 
 if ZMK_HID_REPORT_TYPE_HKRO
 
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -60,8 +60,10 @@ endchoice
 
 config ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT
     bool "Enable extended NKRO reporting"
-    default n
     depends on ZMK_HID_REPORT_TYPE_NKRO
+    help
+      Enables higher usage range for NKRO (F13-F24 and INTL1-9).
+      Please note this is not compatible with Android currently and you will get no input
 
 
 if ZMK_HID_REPORT_TYPE_HKRO
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -58,13 +58,11 @@ config ZMK_HID_REPORT_TYPE_NKRO
 
 endchoice
 
-if ZMK_HID_REPORT_TYPE_NKRO
-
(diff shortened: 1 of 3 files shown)


docs/docs/config/system.md
```diff
@@ -33,12 +33,24 @@ Exactly zero or one of the following options may be set to `y`. The first is use
 | `CONFIG_ZMK_HID_REPORT_TYPE_HKRO` | Enable `CONFIG_ZMK_HID_KEYBOARD_REPORT_SIZE` key roll over.                                           |
 | `CONFIG_ZMK_HID_REPORT_TYPE_NKRO` | Enable full N-key roll over. This may prevent the keyboard from working with some BIOS/UEFI versions. |
 
+:::note NKRO usages
+
+By default the NKRO max usage is set so as to maximise compatibility, however certain less frequently used keys (F13-F24 and INTL1-8) will not work. The solution is to set 'CONFIG_ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT=y'. This is known to break compatibility with Android.
```
caksoylar, line 38:
```suggestion
By default the NKRO max usage is set so as to maximize compatibility, however certain less frequently used keys (F13-F24 and INTL1-8) will not work with it. One solution is to set `CONFIG_ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT=y`, however this is known to break compatibility with Android and thus not enabled by default.
```

caksoylar (commented):
The docs LGTM other than one improvement suggestion.

app/Kconfig
```diff
@@ -53,10 +53,19 @@ config ZMK_HID_REPORT_TYPE_NKRO
...
 
 endchoice
 
+if ZMK_HID_REPORT_TYPE_NKRO
+
+config ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT
+    bool "Enable extended NKRO reporting"
+    default n
+
+endif
```
petejohanson, line 67: This should probably depend on the NKRO option:

```suggestion
config ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT
    bool "Enable extended NKRO reporting"
    default n
    depends on ZMK_HID_REPORT_TYPE_NKRO

```

petejohanson (changes requested):
One comment on the Kconfig details.

app/Kconfig
```diff
@@ -53,10 +53,17 @@ config ZMK_HID_REPORT_TYPE_NKRO
...
       from working with some BIOS/UEFI versions that only support "boot keyboard" support.
-      This option also prevents using some infrequently used higher range HID usages.
+      This option also prevents using some infrequently used higher range HID usages (notably F13-F24 and INTL1-9)
+      These usages can be re enabled with ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT.
 
 endchoice
 
+config ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT
+    bool "Enable extended NKRO reporting"
+    default n
```
petejohanson, line 63: Sorry, I don't always remember to do this properly, but the default for booleans is already no, so this is redundant, and can even cause issues sometimes.


```suggestion
```

app/Kconfig
```diff
@@ -53,10 +53,17 @@ config ZMK_HID_REPORT_TYPE_NKRO
...
+    depends on ZMK_HID_REPORT_TYPE_NKRO
```
petejohanson, line 64: It would be nice to add a `help` section here that also warns of the Android compatibility risk

petejohanson (changes requested):
Couple minor things I missed the first time.

petejohanson (approved):

PR #2027, opened 2023-11-17
Feature: pointer movement/scrolling

petejohanson:
Continuation of the incremental integration of the great work from @krikun98 in #778 with the movement and scrolling pieces next. Based on the rebased work from @caksoylar and and the split acceleration from @bryanforbes

## TODO

- [x] Docs update
- [ ] Review acceleration curves/functions configurability
- [ ] Review @manna-harbour Discord discussion on per-device scaling factors.
- [ ] Review @manna-harbour Discord note on storing float values, and on send of the report discard whole values, but retain the fractional remainder to be included in the follow up calculations, to avoid jerkiness/jumpiness.


app/src/mouse/key_listener.c
```diff
@@ -0,0 +1,201 @@
...
+}
+
+K_WORK_DEFINE(mouse_tick, &mouse_tick_timer_handler);
+
+void mouse_timer_cb(struct k_timer *dummy) {
+    LOG_DBG("Submitting mouse work to queue");
+    k_work_submit_to_queue(zmk_mouse_work_q(), &mouse_tick);
+}
+
+K_TIMER_DEFINE(mouse_timer, mouse_timer_cb, mouse_clear_cb);
```
krikun98, line 65: The part that initially made me bang my head against the wall with all this is the fact that Zephyr sometimes seems to de-prioritize these timers if the system is under load.
This is (as far as I was able to understand) the primary cause of the instability with mouse movement - reports can be sent out at random intervals or dropped entirely.
I was exploring re-writing this system to use a dedicated thread, but I wasn't skilled enough at C at the time.
I'm not sure this basis for the tick system is robust enough for general use.

petejohanson: So... timers should be using the hardware timers to trigger, and happen in ISR context. The *queue* has a certain priority, can be preempted, etc.  Tweaking that thread priority might help, but I'll play a bit to see what I can determine. Thanks for the insight!

krikun98: This is most certainly not definitive - it's a half-remembered conclusion from an investigation I wasn't skilled enough to perform two years ago, just food for thought. I distinctly remember reports being delayed or dropped and that causing instability. `hog.c:335` might also be worth taking a look at.

app/include/dt-bindings/zmk/mouse.h
```diff
@@ -22,3 +22,37 @@
...
+#define ZMK_MOUSE_DEFAULT_MOVE_VAL 600
+#endif
+
+#ifndef ZMK_MOUSE_DEFAULT_SCRL_VAL
+#define ZMK_MOUSE_DEFAULT_SCRL_VAL 10
+#endif
+
+/* Mouse move behavior */
+#define MOVE_Y(vert) ((vert)&0xFFFF)
+#define MOVE_Y_DECODE(encoded) (int16_t)((encoded)&0x0000FFFF)
```
caksoylar, line 36: I think it might make sense to move these decode defines outside the dt-bindings header, and to either plain mouse.h or the corresponding event headers they are used in.

krikun98:
Could you please change the target to https://github.com/petejohanson/zmk/tree/core/zephyr-3.5-update here? The review is unnecessarily cluttered.

app/src/behaviors/behavior_input_two_axis.c
```diff
@@ -0,0 +1,270 @@
...
+    struct k_work_delayable *d_work = k_work_delayable_from_work(work);
+    struct behavior_input_two_axis_data *data =
+        CONTAINER_OF(d_work, struct behavior_input_two_axis_data, tick_work);
+    const struct device *dev = data->dev;
+    const struct behavior_input_two_axis_config *cfg = dev->config;
+
+    uint32_t timestamp = k_uptime_get();
+
+    LOG_INF("tick start times: %lld %lld %lld", data->state.x.start_time, data->state.y.start_time,
+            timestamp);
```
krikun98, line 151: Build gives this warning here:
```
zmk/app/src/behaviors/behavior_input_two_axis.c:150:13: warning: format '%lld' expects argument of type 'long long int', but argument 4 has type 'uint32_t' {aka 'unsigned int'} [-Wformat=]
  150 |     LOG_INF("tick start times: %lld %lld %lld", data->state.x.start_time, data->state.y.start_time,
```

krikun98: In general, the logging isn't very clear unless you have the code in front of you. Maybe add some info? I.e. "tick times: x start XXX, y start XXX, current XXX".

app/src/mouse/main.c
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <zephyr/kernel.h>
+#include <zmk/mouse.h>
+
+#if IS_ENABLED(CONFIG_ZMK_MOUSE_WORK_QUEUE_DEDICATED)
+K_THREAD_STACK_DEFINE(mouse_work_stack_area, CONFIG_ZMK_MOUSE_DEDICATED_THREAD_STACK_SIZE);
+static struct k_work_q mouse_work_q;
```
krikun98, line 12: Is this work queue even used anywhere anymore?

app/src/behaviors/behavior_mouse_key_press.c
```diff
@@ -11,35 +11,48 @@
...
+    }
+}
+
 static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
                                      struct zmk_behavior_binding_event event) {
     LOG_DBG("position %d keycode 0x%02X", event.position, binding->param1);
 
-    return ZMK_EVENT_RAISE(
-        zmk_mouse_button_state_changed_from_encoded(binding->param1, true, event.timestamp));
+    process_key_state(device_get_binding(binding->behavior_dev), binding->param1, true);
```
caksoylar, line 37:
```suggestion
    process_key_state(zmk_behavior_get_binding(binding->behavior_dev), binding->param1, true);
```

app/src/behaviors/behavior_mouse_key_press.c
```diff
@@ -11,35 +11,48 @@
...
+    return 0;
 }
 
 static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
                                       struct zmk_behavior_binding_event event) {
     LOG_DBG("position %d keycode 0x%02X", event.position, binding->param1);
-    return ZMK_EVENT_RAISE(
-        zmk_mouse_button_state_changed_from_encoded(binding->param1, false, event.timestamp));
+
+    process_key_state(device_get_binding(binding->behavior_dev), binding->param1, false);
```
caksoylar, line 46:
```suggestion
    process_key_state(zmk_behavior_get_binding(binding->behavior_dev), binding->param1, false);
```

app/src/behaviors/behavior_input_two_axis.c
```diff
@@ -0,0 +1,272 @@
...
+    data->dev = dev;
+    k_work_init_delayable(&data->tick_work, tick_work_cb);
+
+    return 0;
+};
+
+static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
+                                     struct zmk_behavior_binding_event event) {
+
+    const struct device *behavior_dev = device_get_binding(binding->behavior_dev);
```
caksoylar, line 231:
```suggestion
    const struct device *behavior_dev = zmk_behavior_get_binding(binding->behavior_dev);
```

app/src/behaviors/behavior_input_two_axis.c
```diff
@@ -0,0 +1,272 @@
...
+    int16_t x = MOVE_X_DECODE(binding->param1);
+    int16_t y = MOVE_Y_DECODE(binding->param1);
+
+    zmk_input_synth_pointer_adjust_speed(behavior_dev, x, y);
+    return 0;
+}
+
+static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
+                                      struct zmk_behavior_binding_event event) {
+    const struct device *behavior_dev = device_get_binding(binding->behavior_dev);
```
caksoylar, line 244:
```suggestion
    const struct device *behavior_dev = zmk_behavior_get_binding(binding->behavior_dev);
```

caksoylar (commented):
I think these are some missed updates for the rebase to account for https://github.com/zmkfirmware/zmk/commit/36eda571b

petejohanson:
> I think these are some missed updates for the rebase to account for [36eda571b](https://github.com/zmkfirmware/zmk/commit/36eda571b)

Indeed. Hopefully all good now. Will test further later.

app/src/behaviors/behavior_input_two_axis.c
petejohanson: Indeed. Cleaned up and fixed the build warning.

app/src/mouse/main.c
petejohanson: No. Removed for now.

app/src/mouse/hid_input_listener.c
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    if (evt->sync) {
+        zmk_endpoints_send_mouse_report();
+        zmk_hid_mouse_scroll_set(0, 0);
+        zmk_hid_mouse_movement_set(0, 0);
+    }
+}
+
+INPUT_CALLBACK_DEFINE(NULL, input_handler);
```
fabiobaltieri, line 110: If the handler requries syncronization between multiple event types you really want to bind the callback to a specific device. If you trigger the callback unfiltered, any other event from other devices with the sync bit set is going to to trigger the handler (the report in this case), potentially with inconsistent status.

For this use case I'd recommend splitting the callback, the key code can be handled unflitered just fine so it can be its own thing. For the movement and scroller you should have a callback per device and specific the handler as the first argument of `INPUT_CALLBACK_DEFINE`. I like to have a binding to instantiate as a child of the device generating the events and then use `DT_INST_PARENT` to find its handler here (see https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/input/input_keymap.c#L105C46-L105C60 for an example), but you may as just have a known nodelabel and makes things even simpler.

petejohanson: @fabiobaltieri Yeah, that makes a lot of sense! We probably should be staging the in-progress movement data per-device as well so that the sync flag can then be used to submit only that data to our HID layer.

That does mean we'll need to explicitly list all devices we want to have considered for input, instead of a non-filtered subscription as a catch all.

fabiobaltieri: Yeah that's the proper way of doing it, the Linux parallel of that would be opening a specific `/dev/input/event*` device node.

petejohanson:
I've pushed the Cirque Pinnacle driver into a standalone module to make it easier to test that driver along with this PR: https://github.com/petejohanson/cirque-input-module

arturo182:
I found a bug-ish where mouse movements don't update the activity state. So (among other things) moving the mouse does not turn the backlight back on.

I'm not 100% sure if there even is a way to make a PR to a draft PR on Github, so I will just link the commit here: https://github.com/arturo182/zmk/commit/4c4c105f4eca9ece164bb05560e67b3758ff367c

I think you should be able to cherry-pick it to your branch and it will appear here. Assuming my commit is the correct way to fix the issue, I am still learning ZMK :)

petejohanson:
> I found a bug-ish where mouse movements don't update the activity state. So (among other things) moving the mouse does not turn the backlight back on.
> 
> I'm not 100% sure if there even is a way to make a PR to a draft PR on Github, so I will just link the commit here: [arturo182@4c4c105](https://github.com/arturo182/zmk/commit/4c4c105f4eca9ece164bb05560e67b3758ff367c)
> 
> I think you should be able to cherry-pick it to your branch and it will appear here. Assuming my commit is the correct way to fix the issue, I am still learning ZMK :)
> 
> PS: This is verified on HW.

That's probably the simplest way to solve it. After the recent change I just pushed to make input subscriptions *explicit*, your approach does have the unwanted side effect of updating activity state even for input devices we're not attached to, so I may approach it a slightly different way for this PR.

Thanks for raising the issue!

app/src/mouse/hid_input_listener.c
petejohanson: @fabiobaltieri Ok, so I pushed a change here so that any input device you want to use requires you add an explicit listener node, e.g.:

```
    glidepoint_input {
        compatible = "zmk,input-listener";
        device = <&glidepoint>;

        xy-swap;
        y-invert;
        x-invert;
    };
```

That then creates a dedicated listener that buffers the event data scoped to that device until it gets a sync event, which causes the locally buffered data to be pushed into our HID layer.

tokyo2006:
Hi, @petejohanson I used your PR branch petejohanson:feat/pointers-move-scroll to build the firmware for my keyboard, but recently the builds have been failing. So I found a commit (7e45c4f) from your branch that allowed me to successfully build, but it changed the up-and-down movement of the trackball to scrolling. Can you please tell me how to resolve this issue? Thank you very much. I think the simplest solution would be to use one of the commits from your branch, but I'm not sure which one to use.

caksoylar:
The report descriptor changed, so you just need to re pair over BT to update it in most cases: https://deploy-preview-2178--zmk.netlify.app/docs/features/bluetooth#refreshing-the-hid-descriptor

tokyo2006:
> The report descriptor changed, so you just need to re pair over BT to update it in most cases: https://deploy-preview-2178--zmk.netlify.app/docs/features/bluetooth#refreshing-the-hid-descriptor

Hi @caksoylar I can't open your link

docs/docs/behaviors/mouse-emulation.md
```diff
@@ -69,3 +66,67 @@ This example will send press of the fourth mouse button when the binding is trig
...
+The following will send a scroll down event to the host when pressed/held:
+
+```
+&msc MOVE_DOWN
+```
+
+The following will send a scroll left event to the host when pressed/held:
+
+```
+&msc MOVE_LEFT
```
eteq, line 131: I think this is what's actually meant, right?  At any rate this seems to work where MOVE_* did not...
```suggestion
| `SCRL_UP`    | Scroll up    |
| `SCRL_DOWN`  | Scroll down  |
| `SCRL_LEFT`  | Scroll left  |
| `SCRL_RIGHT` | Scroll right |

### Examples

The following will send a scroll down event to the host when pressed/held:

```
&msc SCRL_DOWN
```

The following will send a scroll left event to the host when pressed/held:

```
&msc SCRL_LEFT
```

**Comment by ryanwwest** [comment]
Very excited by the recent work on this. I can barely wait for the day we can have a single ZMK primary 'dongle' talking to 3 zmk peripherals - split keyboard left, split keyboard right, and wireless trackball (e.g. variant of Ploopy Adept).

**Comment by myst729** [comment]
Mouse scroll keys cannot be used by encoders. I defined several sensor behaviors like this:

```c
  behaviors {
    volume_encoder: volume_encoder {
      compatible = "zmk,behavior-sensor-rotate";
      #sensor-binding-cells = <0>;
      bindings = <&kp C_VOL_UP>, <&kp C_VOL_DN>;
    };

    scroll_encoder: scroll_encoder {
      compatible = "zmk,behavior-sensor-rotate";
      #sensor-binding-cells = <0>;
      bindings = <&msc SCRL_UP>, <&msc SCRL_DOWN>;
    };

    rgb_encoder: rgb_encoder {
      compatible = "zmk,behavior-sensor-rotate";
      #sensor-binding-cells = <0>;
      bindings = <&rgb_ug RGB_BRI>, <&rgb_ug RGB_BRD>;
    };
  };
```

Both `sensor-bindings = <&volume_encoder>;` and `sensor-bindings = <&rgb_encoder>;` works, but `sensor-bindings = <&scroll_encoder>;` doesn't.

caksoylar:
> Mouse scroll keys work with key presses, but cannot be used by encoders. I defined several sensor behaviors like this:

This is because the duration × move speed of the scroll behavior isn't enough for the OS to trigger a scroll. You need to either
- bump `tap-ms` of the [sensor-rotate behavior](https://zmk.dev/docs/config/behaviors#sensor-rotation) to a larger value, like `100`, or
- increase max speed of the scroll behavior (either via setting the `ZMK_MOUSE_DEFAULT_SCRL_VAL` define before importing the mouse header, or directly using `&msc SCRL_Y(-50)` etc.). The default speed is 10.

I'd probably try a combination of the two; e.g. try a tap-ms of 50 and see if scroll speed needs increasing.

myst729:
> I'd probably try a combination of the two; e.g. try a tap-ms of 50 and see if scroll speed needs increasing.

Adding `tap-ms = <100>;` does the trick. Thanks.

fiplox:
> > I'd probably try a combination of the two; e.g. try a tap-ms of 50 and see if scroll speed needs increasing.
> 
> Adding `tap-ms = <100>;` does the trick. Thanks.

while adding `tap-ms = <100>;` makes it work, I find a delay being too big. Playing with parameters, found that `tap-ms` of 30 with `#define ZMK_MOUSE_DEFAULT_SCRL_VAL 35` works almost perfect.

While using this branch I found out that if I use scroll with an encoder and I scroll too fast, the scroll is never released (infinite scroll) and I have to disconnect my keyboard.

eteq:
Has @petejohanson abandoned this PR? Or is it awaiting review? (If abandoned I might try to pick it up in a follow-on PR but not if @petejohanson is still working on it)

caksoylar:
This isn't abandoned, I am guessing Studio work is getting more priority right now. Pete mentioned on Discord that he is planning to refactor the endpoints a bit (the current split one has some issues on some OS, like Ubuntu 22.04).

Meanwhile if anyone needs a rebased branch in the very near future, you can cherry-pick from https://github.com/caksoylar/zmk/commits/caksoylar/experimental.

app/include/dt-bindings/zmk/mouse.h
petejohanson: IMHO, they are complementary and should live together.... I think keeping them here is appropriate.

proostas:
I built a new firmware for my keyboard after the recent changes had been pushed and I noticed a very strange behavior with certain keys. 

I have the following settings for mouse behaviors:

```
&mmv {
    acceleration-exponent = <1>;  // 1
    time-to-max-speed-ms = <300>; // 300
    delay-ms = <0>;
    trigger-period-ms = <8>;      // 16
};
&msc {
    acceleration-exponent = <1>;  // 0
    time-to-max-speed-ms = <40>;  // 300
    delay-ms = <0>;
    trigger-period-ms = <8>;      // 16
};
```

In both nodes the delay-ms is set to 0. But I clearly have a delay with (and only with) **&msc SCRL_DOWN** and **&mmv MOVE_LEFT** now. Could it be caused by some regression?

docs/docs/behaviors/mouse-emulation.md
```diff
@@ -69,3 +66,67 @@ This example will send press of the fourth mouse button when the binding is trig
...
+| Define       | Action     |
+| :----------- | :--------- |
+| `MOVE_UP`    | Move up    |
+| `MOVE_DOWN`  | Move down  |
+| `MOVE_LEFT`  | Move left  |
+| `MOVE_RIGHT` | Move right |
+
+### Examples
+
+The following will send a scroll down event to the host when pressed/held:
```
woodies11, line 90: Did you mean `mouse movement down` event?

caksoylar:
Closing since #2477 is merged.

hailee0710:
@caksoylar does it mean that I can use the main zmk for pointing instead of this PR?

screaming-service:
I have been using main since, it is working.

petejohanson:
> @caksoylar does it mean that I can use the main zmk for pointing instead of this PR? 

Correct.  See the documentation on zmk.dev for the details.

PR #2028, opened 2023-11-19
refactor!: Remove deprecated label property

joelspadin:
Zephyr has deprecated the `label` property in favor of using the devicetree node name as a node identifier. This PR contains several changes to remove all instances of `label` from the codebase. The property is still supported, however, so most user keymaps will not need to be changed.

The most significant changes are:

- Layers have a new `name` property to replace `label` when determining the layer name shown on displays.
- Node names for all predefined behaviors have been shortened.
- Behaviors now use `BEHAVIOR_*_DEFINE()` macros instead of `DEVICE_*_DEFINE()`. These define the device as before, but also add it to a new list of behavior devices.
- Looking up a behavior by name is now done with `zmk_behavior_get_binding()` instead of `device_get_binding()`. This searches only the list of devices, so it is faster, and there is no possibility of returning the wrong device if one of the shortened behavior names collides with another node defined by the SoC.
- If logging is enabled, an additional check will be run at system init, which will log an error if any two behaviors have the same name.

For the most part, these changes are non-breaking. However, because the names of predefined behavior nodes have been changed, if any keymap changes the settings on a behavior like

```dts
/ {
  behaviors {
    behavior_mod_tap {
      tapping-term-ms = <150>;
    };
  };
};
```
instead of the more common
```dts
&mt {
  tapping-term-ms = <150>;
};
```
then it will need to be updated to use the new name.

### DIFF
--- a/docs/blog/2020-10-03-bootloader-fix.md
+++ b/docs/blog/2020-10-03-bootloader-fix.md
@@ -176,7 +176,6 @@ this to all of the `.dts` files for the boards that were affected by this issue.
 
 ```diff
         code_partition: partition@26000 {
-            label = "code_partition";
 -           reg = <0x00026000 0x000d2000>;
 +           reg = <0x00026000 0x000c6000>;
         };
@@ -184,7 +183,6 @@ this to all of the `.dts` files for the boards that were affected by this issue.
 
 -       storage_partition: partition@f8000 {
 +       storage_partition: partition@ec000 {
-            label = "storage";
 -           reg = <0x000f8000 0x00008000>;
 +           reg = <0x000ec000 0x00008000>;
         };
@@ -193,3 +191,7 @@ this to all of the `.dts` files for the boards that were affected by this issue.
 And with those changes, we should no longer run into this issue! In the process
 of these changes, we lost 48KB of space for application code, but we're only
 using around 20% of it anyways. 🎉
+
+## Article Updates
+
+- 12/2023: Removed the deprecated `label` property from code snippets.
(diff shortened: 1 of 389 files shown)


docs/docs/config/keymap.md
```diff
@@ -19,7 +19,7 @@ Each child node can have the following properties:
 
 | Property          | Type          | Description                                                            |
 | ----------------- | ------------- | ---------------------------------------------------------------------- |
-| `label`           | string        | Unique label for the node                                              |
+| `name`            | string        | Name for the layer on displays                                         |
```
petejohanson, line 22:
```suggestion
| `display-name`            | string        | Name for the layer on displays                                         |
```

docs/blog/2020-10-03-bootloader-fix.md
```diff
@@ -176,15 +176,13 @@ this to all of the `.dts` files for the boards that were affected by this issue.
 
 ```diff
         code_partition: partition@26000 {
-            label = "code_partition";
```
petejohanson, line 179: As discussed in Discord, adding a small "Changes" or "Edits" list to any changed blog posts would be good for posterity whenever changing old posts.

petejohanson (changes requested):
Just a couple super minor things. On a whole this looks great! Thanks for tackling it.

docs/docs/config/keymap.md
joelspadin: Fixed

docs/blog/2020-10-03-bootloader-fix.md
joelspadin: Added. Not sure I picked the best heading for it, so feel free to suggest a better one.

petejohanson:
@joelspadin Needs a small base to fix a conflict, then we can merge right away. Thanks!

PR #2032, opened 2023-11-22
feat(split): Add is_bonded function to peripherals

ReFil:
There is already a function to see if the peripheral is connected, a matching one for if it's bonded is a useful addition for displays/indicators. `is_bonded` gets reset to false in the advertising function in preparation for runtime peripheral bond clearing (#1757)

Tested on a 360 Pro with some custom indicators stuff and seems to work as expected

PR #2033, opened 2023-11-22
feat(split): Add peripheral is bonded/connected to central

ReFil:
For easy use by displays and indicators it would be helpful to be able to see if any peripheral at an index is connected and bonded, follow on from #2032

### DIFF
--- a/app/include/zmk/split/bluetooth/central.h
+++ b/app/include/zmk/split/bluetooth/central.h
@@ -6,3 +6,6 @@
 
 int zmk_split_bt_invoke_behavior(uint8_t source, struct zmk_behavior_binding *binding,
                                  struct zmk_behavior_binding_event event, bool state);
+
+bool zmk_split_bt_central_peripheral_is_connected(uint8_t index);
+bool zmk_split_bt_central_peripheral_is_bonded(uint8_t index);
--- a/app/include/zmk/ble.h
+++ b/app/include/zmk/ble.h
@@ -36,4 +36,5 @@ int zmk_ble_unpair_all();
 
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
 int zmk_ble_put_peripheral_addr(const bt_addr_le_t *addr);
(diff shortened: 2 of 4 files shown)


caksoylar:
Nice, it might be useful for a single display (on central) setup on splits.

caksoylar:
Is there an existing event to hook up to in order to listen to peripheral changes? If not, might be a good opportunity to add one here.

ReFil:
There is but it doesn't work quite right. This gets addressed in #2036 as it's needed for that functionality

app/include/zmk/split/bluetooth/central.h
```diff
@@ -5,4 +5,7 @@
 #include <zmk/behavior.h>
 
 int zmk_split_bt_invoke_behavior(uint8_t source, struct zmk_behavior_binding *binding,
-                                 struct zmk_behavior_binding_event event, bool state);
\ No newline at end of file
+                                 struct zmk_behavior_binding_event event, bool state);
+
+bool zmk_split_bt_central_peripheral_is_connected(uint8_t index);
+bool zmk_split_bt_central_peripheral_is_bonded(uint8_t index);
```
petejohanson, line 11: Thoughts on a single API that returns an enum with the state? The "is connected" makes no sense if we don't even have an address bonded to that index, for example, so a single API to just query the "peripheral state" makes more sense, IMHO.

ReFil: I think that's a good idea in general, my only concern would be that's a divergence from how it's currently handled for central->host connections and peripheral->central connections which both use a pair of bool return functions. Would it be better to change all of them over to a consistent API with one `BLE_CONN_STATE` enum with connected, bonded but not connected and unbounded values?

ReFil: Thinking about this more i'm concerned changing the other functions would leak to breakage of module and config repo custom display implementations. A stopgap would be implement enum return, then change the bool return functions to reference that enum return function for some period of time then remove them, similar to how the encoder changes were handled

PR #2037, opened 2023-11-26
bt: add BT_CLR_ALL behaviour

chrisandreae:
Defines a new behavior to clear all paired Bluetooth profiles.

Implementation question: is it appropriate for this to subsume the (commented out) `BT_FULL_RESET`?

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -234,6 +234,7 @@ void zmk_ble_clear_all_bonds(void) {
 
     // Automatically switch to profile 0
     zmk_ble_prof_select(0);
+    update_advertising();
 };
 
 int zmk_ble_active_profile_index(void) { return active_profile; }
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -82,7 +82,7 @@ static bt_addr_le_t peripheral_addrs[ZMK_SPLIT_BLE_PERIPHERAL_COUNT];
 
 #endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL) */
 
-static void raise_profile_changed_event() {
+static void raise_profile_changed_event(void) {
     ZMK_EVENT_RAISE(new_zmk_ble_active_profile_changed((struct zmk_ble_active_profile_changed){
         .index = active_profile, .profile = &profiles[active_profile]}));
 }
@@ -93,7 +93,7 @@ static void raise_profile_changed_event_callback(struct k_work *work) {
 
 K_WORK_DEFINE(raise_profile_changed_event_work, raise_profile_changed_event_callback);
 
-bool zmk_ble_active_profile_is_open() {
+bool zmk_ble_active_profile_is_open(void) {
     return !bt_addr_le_cmp(&profiles[active_profile].peer, BT_ADDR_LE_ANY);
 }
 
@@ -112,7 +112,7 @@ void set_profile_address(uint8_t index, const bt_addr_le_t *addr) {
     k_work_submit(&raise_profile_changed_event_work);
 }
 
-bool zmk_ble_active_profile_is_connected() {
+bool zmk_ble_active_profile_is_connected(void) {
     struct bt_conn *conn;
     struct bt_conn_info info;
     bt_addr_le_t *addr = zmk_ble_active_profile_addr();
@@ -161,7 +161,7 @@ bool zmk_ble_active_profile_is_connected() {
     }                                                                                              \
     advertising_status = ZMK_ADV_CONN;
 
-int update_advertising() {
+int update_advertising(void) {
     int err = 0;
     bt_addr_le_t *addr;
     struct bt_conn *conn;
@@ -210,7 +210,7 @@ static void update_advertising_callback(struct k_work *work) { update_advertisin
 
 K_WORK_DEFINE(update_advertising_work, update_advertising_callback);
 
-int zmk_ble_clear_bonds() {
+int zmk_ble_clear_bonds(void) {
     LOG_DBG("");
 
     if (bt_addr_le_cmp(&profiles[active_profile].peer, BT_ADDR_LE_ANY)) {
@@ -224,7 +224,7 @@ int zmk_ble_clear_bonds() {
     return 0;
 };
 
-int zmk_ble_clear_all_bonds() {
+int zmk_ble_clear_all_bonds(void) {
     LOG_DBG("zmk_ble_clear_all_bonds()");
 
     // Unpair all profiles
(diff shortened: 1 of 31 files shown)


app/src/ble.c
```diff
@@ -222,6 +222,23 @@ int zmk_ble_clear_bonds() {
     return 0;
 };
 
+int zmk_ble_clear_all_bonds() {
```
joelspadin, line 225: This is an issue with the existing functions too, but leaving the parameter list blank in C allows the function to accept any number of parameters of arbitrary types. The parameter list needs to be `(void)` for it to not take any parameters.
```suggestion
int zmk_ble_clear_all_bonds(void) {
```

app/src/ble.c
```diff
@@ -222,6 +222,23 @@ int zmk_ble_clear_bonds() {
...
+    LOG_DBG("zmk_ble_clear_all_bonds()");
+
+    // Unpair all profiles
+    for (uint8_t i = 0; i < ZMK_BLE_PROFILE_COUNT; i++) {
```
joelspadin, line 229:
```suggestion
    for (int i = 0; i < ZMK_BLE_PROFILE_COUNT; i++) {
```
Using integers of small types as loop counters is generally a bad idea. In this specific case, it doesn't matter because ZMK_BLE_PROFILE_COUNT is a compile-time constant which *shouldn't* be bigger than 255, so the compiler will optimize this to the same code no matter what type is used here, but see https://stackoverflow.com/questions/34891181/for-array-loop-use-int-or-uint8-t

app/src/ble.c
```diff
@@ -222,6 +222,23 @@ int zmk_ble_clear_bonds() {
...
+        if (bt_addr_le_cmp(&profiles[i].peer, BT_ADDR_LE_ANY)) {
+            bt_unpair(BT_ID_DEFAULT, &profiles[i].peer);
+            set_profile_address(i, BT_ADDR_LE_ANY);
+        }
+    }
+
+    // Automatically switch to profile 0
+    zmk_ble_prof_select(0);
+
+    return 0;
```
joelspadin, line 239: If this function doesn't have any errors to report, then it shouldn't return anything. Having a return code means that callers of this function will likely be written like
```c
int rc = zmk_ble_clear_all_bonds();
if (rc) {
  // This is dead code
  LOG_ERR("Failed to clear all bonds");
}
```

Alternatively, `bt_unpair()` *does* have a return code which we are currently ignoring. If we changed this function to report errors from the functions it calls, then the return code would be meaningful.

(This appears to also be an issue with multiple other functions in this file too.)

app/src/ble.c
```diff
@@ -222,6 +222,23 @@ int zmk_ble_clear_bonds() {
     return 0;
 };
 
+int zmk_ble_clear_all_bonds() {
+    LOG_DBG("zmk_ble_clear_all_bonds()");
+
+    // Unpair all profiles
+    for (uint8_t i = 0; i < ZMK_BLE_PROFILE_COUNT; i++) {
+        if (bt_addr_le_cmp(&profiles[i].peer, BT_ADDR_LE_ANY)) {
```
joelspadin, line 230: This duplicates code from `zmk_ble_clear_bonds()`.

@petejohanson what do you think about this?
1. Rename `zmk_ble_clear_bonds()` to `zmk_ble_clear_active_profile_bonds()`
2. Pull out this `if` block into a static function for clearing bonds for a specific profile given its index.
3. Change `zmk_ble_clear_bonds()` to call that function on `active_profile`, and change this function to call it on every profile.

joelspadin (changes requested):

chrisandreae:
I've no objections to the suggested changes beyond consistency with the rest of the file - if the new code is updated to follow those style guidelines, should the existing code be updated as well?

petejohanson:
> I've no objections to the suggested changes beyond consistency with the rest of the file - if the new code is updated to follow those style guidelines, should the existing code be updated as well?

Might as well while you're there. Thanks.

chrisandreae:
OK, I've taken a linting pass over the project and made sure that the everything now passes `-Wstrict-prototypes`.

I've also updated the functions that can't meaningfully fail to not pretend to return an error.

caksoylar:
Can you update the docs to add a reference to the new define?

app/src/ble.c
petejohanson: Agreed. We should try to avoid duplicating this, especially if more comes up later that needs to happen when a  given profile is cleared.

petejohanson (changes requested):
Just one comment on the code in agreement w/ Joel. Also see Cem's note about the docs updates needed. Thanks!

app/src/ble.c
chrisandreae: I'm quite happy to change this if it better matches the project style (and have done so), but the concern I have is that this isn't indexing an arbitrary array as described in your link, it's calling a function with a `uint8_t` argument type. It doesn't ring quite as true to me to iterate with a wider type that will need a narrowing cast to pass to the called function. Probably just a nit-pick, so it's your call.

chrisandreae:
> Can you update the docs to add a reference to the new define?

Done!

> Agreed. We should try to avoid duplicating this

Makes sense, done!

docs/docs/behaviors/bluetooth.md
```diff
@@ -10,7 +10,10 @@ between the keyboard and the host. By default, ZMK supports five "profiles" for
 computer/laptop/keyboard should receive the keyboard input; many of the commands here operate on those profiles.
 
 :::note Connection Management
-When pairing to a host device ZMK saves bond information to the selected profile. It will not replace this when you initiate pairing with another device. To pair with a new device select an unused profile with `BT_SEL`, `BT_NXT` or `BT_PRV` bindings, or by clearing an existing profile using `BT_CLR`.
+When pairing to a host device ZMK saves bond information to the selected
+profile. It will not replace this when you initiate pairing with another device.
+To pair with a new device select an unused profile with `BT_SEL`, `BT_NXT` or
+`BT_PRV` bindings, or by clearing an existing profile using `BT_CLR` or `BT_CLR_ALL`.
```
caksoylar, line 16:
```suggestion
To pair with a new device, select a profile that doesn't have a pairing with `BT_SEL`, `BT_NXT` or
`BT_PRV` bindings, or clear an already paired profile using `BT_CLR` or `BT_CLR_ALL`.
```
Existing issues but would be good to correct while touching here.

caksoylar (commented):
Thanks for adding docs!

petejohanson (approved):
Thanks!

PR #2043, opened 2023-11-29
Capslock behavior

snoyer:
This Caps Lock behavior provides ways to create improved, configurable Caps Lock keys.
Unlike sending a regular `&kp CAPSLOCK` which blindly _toggles_ caps lock, this behavior uses the HID indicator state to predictably _set_ caps lock regardless of the current status on the connected device (basically it will send `&kp CAPSLOCK` if toggling is needed but do nothing if the current state already matches the target state).

The main motivation is to provide an alternative `&capsword` that uses caps lock instead of shift modifiers but there are a few additional configuration that could be sensible:

- `&capslock_on` enables caps lock
- `&capslock_off` disables caps lock
- `&capslock_hold` enables caps lock and disables it when released
- `&capslock_word` enables caps lock, and disables it when a word separator is typed or the key is pressed again

This PR supersedes #1485 where I naively proposed to replace the current shift-based `&capsword`, which turned out to be misguided. The shift-based version does have issues (layout compatibility) but this capslock-based version does have its own too (OS compatibility).
It seems like we can't have the one true way and we should provide both and let user pick between "good and wrong" `&capsword` (using shift modifiers works transparently across OS, but not across layouts because shift isn't actually the same thing as capitalization) and "right and bad" `&capslock_word` (caps lock does proper capitalization across layouts, but its activation/deactivation does not work transparently across all OS due to their various implementation quirks) depending on their use-case.

### DIFF
--- a/app/dts/behaviors/capslock.dtsi
+++ b/app/dts/behaviors/capslock.dtsi
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2023 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <dt-bindings/zmk/keys.h>
+
+/ {
+	behaviors {
+		/omit-if-no-ref/ capslock_on: behavior_capslock_on {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_ON";
+			#binding-cells = <0>;
+			capslock-press-duration = <5>;
+			enable-on-press;
+		};
+		
+		/omit-if-no-ref/ capslock_off: behavior_capslock_off {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_OFF";
+			#binding-cells = <0>;
+			capslock-press-duration = <5>;
+			disable-on-release;
+		};
+		
+		/omit-if-no-ref/ capslock_hold: behavior_capslock_hold {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_HOLD";
+			#binding-cells = <0>;
+			capslock-press-duration = <5>;
+			enable-on-press;
+			disable-on-release;
+		};
+		
+		/omit-if-no-ref/ capslock_word: behavior_capslock_word {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_WORD";
+			#binding-cells = <0>;
+			capslock-press-duration = <5>;
+			enable-on-press;
+			disable-on-next-release;
+			disable-on-keys = <SPACE TAB ENTER ESC>;
+		};
+		
+		/* MacOS compatibility */
+		
+		/omit-if-no-ref/ capslock_on_mac: behavior_capslock_on_mac {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_ON_MAC";
+			#binding-cells = <0>;
+			capslock-press-duration = <95>;
+			enable-on-press;
+		};
+		
+		/omit-if-no-ref/ capslock_off_mac: behavior_capslock_off_mac {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_OFF_MAC";
+			#binding-cells = <0>;
+			capslock-press-duration = <95>;
+			disable-on-release;
+		};
+		
+		/omit-if-no-ref/ capslock_hold_mac: behavior_capslock_hold_mac {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_HOLD_MAC";
+			#binding-cells = <0>;
+			capslock-press-duration = <95>;
+			enable-on-press;
+			disable-on-release;
+		};
+		
+		/omit-if-no-ref/ capslock_word_mac: behavior_capslock_word_mac {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_WORD_MAC";
+			#binding-cells = <0>;
+			capslock-press-duration = <95>;
+			enable-on-press;
+			disable-on-next-release;
+			disable-on-keys = <SPACE TAB ENTER ESC>;
+		};
+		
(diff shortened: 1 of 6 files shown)


joelspadin (commented):
I have two reservations with this:
1. It feels like this one behavior driver does a lot of related, but very different things. Is this intended to allow for different combinations of configuration that don't exist in the predefined behaviors?
2. Having it also handle "capslock word" means that it could easily get out of sync with `&caps_word`. It would be nice to have feature parity, especially if PRs like https://github.com/zmkfirmware/zmk/pull/1742 get accepted and they end up only applying to one of the two implementations.

Would it be better to pull some of the caps lock handling out into core functions that can be called from anywhere? 

Then I think it would be possible to use the existing `&caps_word` behavior, add a new property to it that tells it to hold caps lock, and set `mods = <0>` (or have that done automatically when caps lock mod is enabled) to get the existing behavior to use caps lock instead of shift.

app/dts/behaviors/capslock.dtsi
```diff
@@ -0,0 +1,84 @@
...
+		};
+		
+		/omit-if-no-ref/ capslock_word: behavior_capslock_word {
+			compatible = "zmk,behavior-capslock";
+			label = "CAPSLOCK_WORD";
+			#binding-cells = <0>;
+			capslock-press-duration = <5>;
+			enable-on-press;
+			disable-on-next-release;
+			disable-on-keys = <SPACE TAB ENTER ESC>;
```
joelspadin, line 43: This is inconsistent with `&caps_word`, which uses a list of keys that _continue_ a word rather than end it. There are many other keys (such as punctuation) that I would probably expect to end a word.

As I noted on the PR as a whole, I think it would be confusing to have this and `&caps_word` behave differently. It would likely be easier to keep the two in sync if they were literally the same behavior, but with a property to switch to using caps lock instead of modifiers.

snoyer:
> 1. It feels like this one behavior driver does a lot of related, but very different things. Is this intended to allow for different combinations of configuration that don't exist in the predefined behaviors?

Yes, this is a _capslock_ behavior driver that just happens to allow for _some_ implementation of "caps word" (by automatically deactivating capslock after a configurable word separator has been typed). But other use cases are (include? I don't know maybe there's more):

- you want 2 separate keys for turning capslock on and off, instead of the single `&kp CAPSLOCK` that toggles it
- you want a capslock key that doesn't actually lock and is only active when held, kindof how a modifier works

Maybe I went about it wrong with them implementation, it seemed straightforward to me to have the various activation/deactivation conditions exposed in the behavior and combine them in the behavior instances (not sure about the terminology here, sorry). I'm all ears is there is a more proper way to do it.

> Would it be better to pull some of the caps lock handling out into core functions that can be called from anywhere?

That would also be great but from the few times it's been mentioned (previous PR, Discord), capslock is quirky enough across various OS (Mac has a minimum keypress duration, some user may have remapped/disabled the key altogether, ...) that that may be a hard/risky thing to do. I'm all for it if you have a solution there.


> 2. Having it also handle "capslock word" means that it could easily get out of sync with `&caps_word`. It would be nice to have feature parity, especially if PRs like [Bring caps word to feature parity with QMK #1742](https://github.com/zmkfirmware/zmk/pull/1742) get accepted and they end up only applying to one of the two implementations.
> 
> Then I think it would be possible to use the existing `&caps_word` behavior, add a new property to it that tells it to hold caps lock, and set `mods = <0>` (or have that done automatically when caps lock mod is enabled) to get the existing behavior to use caps lock instead of shift.

I believe looking at the two versions as alternative implementations of the same feature is a red herring. Even though there is obviously a huge apparent overlap they are actually two diametrically opposite features:

- shift-based version _emulates_ capitalization on the keyboard. This is great because the shifting mechanism is standard across OS, but it requires to be layout-aware to implement the capitalization logic (configuring `continue-list` and possibly other proposed exceptions and workarounds, see #1470, #1479, #1478)
- capslock-based version _delegates_ capitalization to the OS. This is great because the OS already knows about capitalization logic across layouts, but it requires to be OS/device-aware because activating/deactivating capslock is surprisingly not standard (for Mac you need to set `capslock-press-duration = <95>`, if you swapped the capslock and escape key you'd need to configure `capslock-press-keycode` accordingly)

I'm also not convinced you can achieve feature parity between the two versions in the first place.
For example on an AZERTY keyboard `èçàù` shifts to `790%`, with capslock it capitalizes to `ÈÇÀÙ` (see [this comment on previous PR](https://github.com/zmkfirmware/zmk/pull/1485#issuecomment-1272222159)) and I don't think you can emulate that with modifiers.

joelspadin:
I didn't mean feature parity with regards to making sure that `&caps_word` produces the same output with modifiers as it does with caps lock (if that were possible, then you wouldn't be creating a version of it that uses caps lock to begin with). I meant with regards to how it determines when it's hit the end of a word, idle timeouts, etc.

snoyer:
> I didn't mean feature parity with regards to making sure that `&caps_word` produces the same output with modifiers as it does with caps lock (if that were possible, then you wouldn't be creating a version of it that uses caps lock to begin with). I meant with regards to how it determines when it's hit the end of a word, idle timeouts, etc.

Ok, I see, that make sense, I did not need to type this whole novel in the end :D

I think we're going to have to disagree on the "end of word" configuration though. `&caps_word` gets away with having a partial `continue-list` because of the hardcoded QWERTY assumption; while the whole point of using capslock is not having to know or care about the layout: pressing the `&capslock_word` key ensures capslock is on (toggles it only if needed), and then does nothing until one of the given stop keys is pressed, at which point it ensures capslock is off (toggles it only if needed).
The fact the capslock version sits and waits for the time to deactivate, as opposed to the shift version which has to actively modify most keys is enough of a conceptual difference to warrant different configuration modes IMO

snoyer:
This behavior is available as a module: https://github.com/snoyer/zmk-behavior-capslock

PR #1243, opened 2022-04-13
feat(battery): Split battery reporting over BLE GATT

Katona:
This PR implements #764.

The changes are relatively simple:

- the central registers on battery level changes of peripheral over BLE GATT.
- then it 'relays' the peripheral battery level changes to the host using BLE GATT again. :)
- for this to work, the central no longer relies on the Battery Service of Zephyr since that assumes one battery only. Instead, it defines its own BAS with two battery level characteristics.

**What I tested**

I tested only on my wireless Corne. For the test I used the  [Web Bluetooth API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Bluetooth_API) since that was the easiest to setup for me. Here is a [link](https://www.youtube.com/watch?v=_aqyuBeeCSs) to a video demonstrating it.

**What I didn't test**

- I didn't test non split BLE keyboards since I don't have any.
- Tested only using macOS. I don't know how other OSes display battery level when multiple one are present in one device.

As indicated in #764 the OSes doesn't seem to be prepared to devices with more than one battery (macOS certainly not), but not all is lost in my opinion. I still believe this PR is useful since:

- it is actually standard (the BLE standard allows it).
- it is not breaking anything: OSes (macOS) ignore the second battery, so battery level reporting still works for central and, eventually OSes might support it (not likely, though).
- we can develop utilities which can display both batteries in the system tray or menu.

PR #2045, opened 2023-11-29
feat(ble): Support peripheral battery levels.

petejohanson:
* Add ability to fetch and report peripheral battery levels on split centrals.
* Add additional support for adding a new Battery Level service to split centrals that exposes fetched peripheral battery levels to connected hosts.

This is a continuation of the great work from @Katona in #1243 with some important changes:

* Move both central fetching and reporting of peripheral battery levels behind new (documented) Kconfig flags
* After reviewing the Battery [Level Service specification](https://www.bluetooth.com/specifications/specs/battery-service/), it became clear to me there could be an easier structure for this. In particular, that spec lays out the requirements if you're going to have more that one battery level service, how to tag a particular battery level characteristic as the main one, etc. Given that, this PR:
  * Simplifies the *core* battery reporting by simply leveraging the normal Zephyr BAS directly still, basically everywhere, including splits of both roles. I have an open PR to Zephyr to properly tag the Zephyr BAS as the main one: https://github.com/zephyrproject-rtos/zephyr/pull/65944 I am monitoring the response there, assuming no major concerns I will be cherry picking that one small change into our Zephyr fork.
  * Behind the extra Kconfig flag there is now  a dedicated new "central BAS proxy" that adds *another* Battery Level service where the Battery Level characteristics are all marked with a description of `auxiliary` to disambiguate them from the main one. The new service properly handles multiple peripherals now as well.
* Tweaks one default Zephyr L2CAP buffer value on splits to be sure we don't get `-ENOMEM` errors on reconnects to peripherals. 

Many thanks to @Katona for all the work they put into this!

# Testing

I've performed the following testing so far:

* Confirmed the battery reporting for the split central `main` battery is the one displayed on the follow OSes:
  * Linux (GNOME, using upower, on latest Fedora release)
  * macOS
  * Windows 11
  * Android 14
* Confirmed that using `bluetoothctl` I can subscribe to notifications on the *other* Battery Level service on a split central, and properly receive notifications it has "proxied" from the split peripheral.

The following testing would be nice:

* iOS device
* The custom macOS app/service that @Katona wrote.

To test, you need to enable the following off-by-default Kconfig settings:

```
CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING=y
CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_PROXY=y
```

### DIFF
--- a/app/src/split/bluetooth/central.c
+++ b/app/src/split/bluetooth/central.c
@@ -27,6 +27,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/event_manager.h>
 #include <zmk/events/position_state_changed.h>
 #include <zmk/events/sensor_event.h>
+#include <zmk/events/battery_state_changed.h>
 #include <zmk/hid_indicators_types.h>
 
 static int start_scanning(void);
@@ -47,6 +48,10 @@ struct peripheral_slot {
     struct bt_gatt_subscribe_params sensor_subscribe_params;
     struct bt_gatt_discover_params sub_discover_params;
     uint16_t run_behavior_handle;
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
+    struct bt_gatt_subscribe_params batt_lvl_subscribe_params;
+    struct bt_gatt_read_params batt_lvl_read_params;
+#endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING) */
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
     uint16_t update_hid_indicators;
 #endif // IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
@@ -265,6 +270,110 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
     return BT_GATT_ITER_CONTINUE;
 }
 
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
+
+static uint8_t peripheral_battery_levels[ZMK_SPLIT_BLE_PERIPHERAL_COUNT] = {0};
+
+int zmk_split_get_peripheral_battery_level(uint8_t source, uint8_t *level) {
+    if (source >= ARRAY_SIZE(peripheral_battery_levels)) {
+        return -EINVAL;
+    }
+
+    if (peripherals[source].state != PERIPHERAL_SLOT_STATE_CONNECTED) {
+        return -ENOTCONN;
+    }
+
+    *level = peripheral_battery_levels[source];
+    return 0;
+}
+
+K_MSGQ_DEFINE(peripheral_batt_lvl_msgq, sizeof(struct zmk_peripheral_battery_state_changed),
+              CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_QUEUE_SIZE, 4);
+
+void peripheral_batt_lvl_change_callback(struct k_work *work) {
+    struct zmk_peripheral_battery_state_changed ev;
+    while (k_msgq_get(&peripheral_batt_lvl_msgq, &ev, K_NO_WAIT) == 0) {
+        LOG_DBG("Triggering peripheral battery level change %u", ev.state_of_charge);
+        peripheral_battery_levels[ev.source] = ev.state_of_charge;
+        ZMK_EVENT_RAISE(new_zmk_peripheral_battery_state_changed(ev));
+    }
+}
+
+K_WORK_DEFINE(peripheral_batt_lvl_work, peripheral_batt_lvl_change_callback);
+
+static uint8_t split_central_battery_level_notify_func(struct bt_conn *conn,
+                                                       struct bt_gatt_subscribe_params *params,
+                                                       const void *data, uint16_t length) {
+    struct peripheral_slot *slot = peripheral_slot_for_conn(conn);
+
+    if (!slot) {
+        LOG_ERR("No peripheral state found for connection");
+        return BT_GATT_ITER_CONTINUE;
+    }
+
+    if (!data) {
+        LOG_DBG("[UNSUBSCRIBED]");
+        params->value_handle = 0U;
+        return BT_GATT_ITER_STOP;
+    }
+
+    if (length == 0) {
+        LOG_ERR("Zero length battery notification received");
+        return BT_GATT_ITER_CONTINUE;
+    }
+
+    LOG_DBG("[BATTERY LEVEL NOTIFICATION] data %p length %u", data, length);
+    uint8_t battery_level = ((uint8_t *)data)[0];
+    LOG_DBG("Battery level: %u", battery_level);
+    struct zmk_peripheral_battery_state_changed ev = {
+        .source = peripheral_slot_index_for_conn(conn), .state_of_charge = battery_level};
+    k_msgq_put(&peripheral_batt_lvl_msgq, &ev, K_NO_WAIT);
+    k_work_submit(&peripheral_batt_lvl_work);
+
+    return BT_GATT_ITER_CONTINUE;
+}
+
+static uint8_t split_central_battery_level_read_func(struct bt_conn *conn, uint8_t err,
+                                                     struct bt_gatt_read_params *params,
+                                                     const void *data, uint16_t length) {
+    if (err > 0) {
+        LOG_ERR("Error during reading peripheral battery level: %u", err);
+        return BT_GATT_ITER_STOP;
+    }
+
+    struct peripheral_slot *slot = peripheral_slot_for_conn(conn);
+
+    if (!slot) {
+        LOG_ERR("No peripheral state found for connection");
+        return BT_GATT_ITER_CONTINUE;
+    }
+
+    if (!data) {
+        LOG_DBG("[READ COMPLETED]");
+        return BT_GATT_ITER_STOP;
+    }
+
+    LOG_DBG("[BATTERY LEVEL READ] data %p length %u", data, length);
+
+    if (length == 0) {
+        LOG_ERR("Zero length battery notification received");
+        return BT_GATT_ITER_CONTINUE;
+    }
+
+    uint8_t battery_level = ((uint8_t *)data)[0];
+
+    LOG_DBG("Battery level: %u", battery_level);
+
+    struct zmk_peripheral_battery_state_changed ev = {
+        .source = peripheral_slot_index_for_conn(conn), .state_of_charge = battery_level};
+    k_msgq_put(&peripheral_batt_lvl_msgq, &ev, K_NO_WAIT);
+    k_work_submit(&peripheral_batt_lvl_work);
+
+    return BT_GATT_ITER_CONTINUE;
+}
+
+#endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING) */
+
 static int split_central_subscribe(struct bt_conn *conn, struct bt_gatt_subscribe_params *params) {
     int err = bt_gatt_subscribe(conn, params);
     switch (err) {
@@ -306,10 +415,6 @@ static uint8_t split_central_chrc_discovery_func(struct bt_conn *conn,
 
     if (bt_uuid_cmp(chrc_uuid, BT_UUID_DECLARE_128(ZMK_SPLIT_BT_CHAR_POSITION_STATE_UUID)) == 0) {
         LOG_DBG("Found position state characteristic");
-        slot->discover_params.uuid = NULL;
-        slot->discover_params.start_handle = attr->handle + 2;
-        slot->discover_params.type = BT_GATT_DISCOVER_CHARACTERISTIC;
-
         slot->subscribe_params.disc_params = &slot->sub_discover_params;
         slot->subscribe_params.end_handle = slot->discover_params.end_handle;
         slot->subscribe_params.value_handle = bt_gatt_attr_value_handle(attr);
@@ -342,9 +447,27 @@ static uint8_t split_central_chrc_discovery_func(struct bt_conn *conn,
         LOG_DBG("Found update HID indicators handle");
         slot->update_hid_indicators = bt_gatt_attr_value_handle(attr);
 #endif // IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
+    } else if (!bt_uuid_cmp(((struct bt_gatt_chrc *)attr->user_data)->uuid,
+                            BT_UUID_BAS_BATTERY_LEVEL)) {
+        LOG_DBG("Found battery level characteristics");
+        slot->batt_lvl_subscribe_params.disc_params = &slot->sub_discover_params;
+        slot->batt_lvl_subscribe_params.end_handle = slot->discover_params.end_handle;
+        slot->batt_lvl_subscribe_params.value_handle = bt_gatt_attr_value_handle(attr);
+        slot->batt_lvl_subscribe_params.notify = split_central_battery_level_notify_func;
+        slot->batt_lvl_subscribe_params.value = BT_GATT_CCC_NOTIFY;
+        split_central_subscribe(conn, &slot->batt_lvl_subscribe_params);
+
+        slot->batt_lvl_read_params.func = split_central_battery_level_read_func;
+        slot->batt_lvl_read_params.handle_count = 1;
+        slot->batt_lvl_read_params.single.handle = bt_gatt_attr_value_handle(attr);
+        slot->batt_lvl_read_params.single.offset = 0;
+        bt_gatt_read(conn, &slot->batt_lvl_read_params);
+#endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING) */
     }
 
-    bool subscribed = (slot->run_behavior_handle && slot->subscribe_params.value_handle);
+    bool subscribed = slot->run_behavior_handle && slot->subscribe_params.value_handle;
+
 #if ZMK_KEYMAP_HAS_SENSORS
     subscribed = subscribed && slot->sensor_subscribe_params.value_handle;
 #endif /* ZMK_KEYMAP_HAS_SENSORS */
@@ -352,6 +475,9 @@ static uint8_t split_central_chrc_discovery_func(struct bt_conn *conn,
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
     subscribed = subscribed && slot->update_hid_indicators;
 #endif // IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
+    subscribed = subscribed && slot->batt_lvl_subscribe_params.value_handle;
+#endif /* IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING) */
 
     return subscribed ? BT_GATT_ITER_STOP : BT_GATT_ITER_CONTINUE;
 }
@@ -382,7 +508,6 @@ static uint8_t split_central_service_discovery_func(struct bt_conn *conn,
     LOG_DBG("Found split service");
     slot->discover_params.uuid = NULL;
     slot->discover_params.func = split_central_chrc_discovery_func;
-    slot->discover_params.start_handle = attr->handle + 1;
     slot->discover_params.type = BT_GATT_DISCOVER_CHARACTERISTIC;
 
     int err = bt_gatt_discover(conn, &slot->discover_params);
@@ -604,6 +729,13 @@ static void split_central_disconnected(struct bt_conn *conn, uint8_t reason) {
     bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
 
     LOG_DBG("Disconnected: %s (reason %d)", addr, reason);
+
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
+    struct zmk_peripheral_battery_state_changed ev = {
+        .source = peripheral_slot_index_for_conn(conn), .state_of_charge = 0};
+    k_msgq_put(&peripheral_batt_lvl_msgq, &ev, K_NO_WAIT);
+    k_work_submit(&peripheral_batt_lvl_work);
+#endif // IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
 
     err = release_peripheral_slot_for_conn(conn);
 
--- a/app/src/split/bluetooth/central_bas_proxy.c
+++ b/app/src/split/bluetooth/central_bas_proxy.c
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/init.h>
+#include <sys/types.h>
+#include <zephyr/kernel.h>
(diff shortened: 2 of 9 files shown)


equiman:
I've merged these changes into a branch and config the `west.yml` file to use it:

```
manifest:
  remotes:
    - name: deinsoftware
      url-base: https://github.com/deinsoftware
  projects:
    - name: zmk
      remote: deinsoftware
      revision: main
      import: app/west.yml
  self:
    path: config
 ```

And tested it with the @Maksim-Isakau [zmk-split-battery](https://github.com/Maksim-Isakau/zmk-split-battery) app for windows. But it only shows one battery level.

image

Will it need some extra configuration on my [corne-zmk-config](https://github.com/deintech/corne-zmk-config)?

**Comment by petejohanson** [comment]
> I've merged these changes into a branch and config the `west.yml` file to use it:
> 
> ```
> manifest:
>   remotes:
>     - name: deinsoftware
>       url-base: https://github.com/deinsoftware
>   projects:
>     - name: zmk
>       remote: deinsoftware
>       revision: main
>       import: app/west.yml
>   self:
>     path: config
>  ```
> 
> And tested it with the @Maksim-Isakau [zmk-split-battery](https://github.com/Maksim-Isakau/zmk-split-battery) app for windows. But it only shows one battery level.
> 
> image

It seems that the app from @Katona assumes it just needs to use the first Battery Service, I stead of aggregating all of them from the device: https://github.com/Katona/zmk-ble/blob/main/zmk-ble/ZmkPeripheral.swift#L53

It will need to get updated to work properly.

**Comment by Maksim-Isakau** [comment]
> And tested it with the @Maksim-Isakau [zmk-split-battery](https://github.com/Maksim-Isakau/zmk-split-battery) app for windows. But it only shows one battery level.

@equiman fixed. Please note that you have to specify
```
CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING=y
CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_PROXY=y
```
in your KConfig file of zmk-config, split battery reporting is off by default. I stumbled upon this at first

**Comment by Maksim-Isakau** [comment]
> iOS device

@petejohanson iOS 17 seems to work fine:
ios-zmk-split-batt-lvl

15% is peripheral, 49 main.
On a Lock Screen i see the one that has minimal battery level (tested after charging peripheral).

**Comment by petejohanson** [comment]
> > iOS device
> 
> @petejohanson iOS 17 seems to work fine: ios-zmk-split-batt-lvl
> 
> 15% is peripheral, 49 main. On a Lock Screen i see the one that has minimal battery level (tested after charging peripheral).

It's a shame they don't use our display string at all or otherwise disambiguate the two battery levels, but at least they show them both.

**Comment by petejohanson** [comment]
> Thanks @petejohanson  and @Maksim-Isakau. Now it's working again.
> 
> image
> 
> It's possible to add some config labels to rename each one?
> 
> I think `Peripheral #: ` is descriptive but not a cool name,
> I would like to rename it as `Left` and `Right` for example.
> 

I wonder if perhaps I can enhance this to add a CUD based on the keyboard name? Or we offer an additional Kconfig for the battery  label, and we proxy the CUD from the connected peripherals once they are connected....

**Comment by petejohanson** [comment]
Ok, so the Zephyr PR was accepted and merged, so I've cherry picked that into our Zephyr fork in https://github.com/zmkfirmware/zephyr/pull/27

Waiting to see if @Katona is able to work on his macOS app bits, but I don't think that's a blocker for merging this in the meantime.

**Comment by petejohanson** [comment]
> Well, today something strange is happening. It's showing all battery levels at -1
> 
> image
> 
> Or sometimes only, show the central level:
> 
> image

Is the main battery display in Windows working?

**Comment by equiman** [comment]
Yes, it is: 
image

image

**Comment by bryanforbes** [comment]
It seems like there should be an API for getting the peripheral battery level that's similar to `zmk_battery_state_of_charge()`. This would be useful for displaying peripheral battery level in the RGB underglow like the Glove80 does. Right now, the only place where the levels are stored is in the proxy, but it seems wrong to force users to enable the proxy just to get the levels for use elsewhere. Which means the simplest solution would be to store the levels in `central.c` when fetching is enabled. (cc @chrisandreae)

**Comment by petejohanson** [comment]
> It seems like there should be an API for getting the peripheral battery level that's similar to `zmk_battery_state_of_charge()`. This would be useful for displaying peripheral battery level in the RGB underglow like the Glove80 does. Right now, the only place where the levels are stored is in the proxy, but it seems wrong to force users to enable the proxy just to get the levels for use elsewhere. Which means the simplest solution would be to store the levels in `central.c` when fetching is enabled. (cc @chrisandreae)

Changed the storage to the central code, added a new API for fetching the levels there, and then leveraged that in the BAS proxy code instead.

**Comment by equiman** [comment]
Few days ago, the keyboard stop reporting correctly the battery level, even on windows. 

Always shown the level in 50%. Today it was run on battery, and report it has 50%. Once I've fully charged it, still reporting 50%.

image

And the battery status won't connect any more:

image

**Comment by petejohanson** [comment]
> Few days ago, the keyboard stop reporting correctly the battery level, even on windows.
> 
> Always shown the level in 50%. Today it was run on battery, and report it has 50%. Once I've fully charged it, still reporting 50%.
> 
> image
> 
> And the battery status won't connect any more:
> 
> image

Interesting... Can you please try the latest code I just pushed? Also, getting the ZMK logs when this happens would be useful. Thanks for testing!

File: docs/docs/config/system.md [hunk 1]
```diff
@@ -104,15 +104,18 @@ Note that `CONFIG_BT_MAX_CONN` and `CONFIG_BT_MAX_PAIRED` should be set to the s
...
+| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_POSITION_QUEUE_SIZE`      | int  | Max number of key state events to queue when received from peripherals     | 5                                          |
+| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_SPLIT_RUN_STACK_SIZE`     | int  | Stack size of the BLE split central write thread                           | 512                                        |
+| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_SPLIT_RUN_QUEUE_SIZE`     | int  | Max number of behavior run events to queue to send to the peripheral(s)    | 5                                          |
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_STACK_SIZE`            | int  | Stack size of the BLE split peripheral notify thread                       | 650                                        |
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_PRIORITY`              | int  | Priority of the BLE split peripheral notify thread                         | 5                                          |
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_POSITION_QUEUE_SIZE`   | int  | Max number of key state events to queue to send to the central             | 10                                         |
```
caksoylar, line 121:
```suggestion
| Config                                                  | Type | Description                                                                | Default                                    |
| ------------------------------------------------------- | ---- | -------------------------------------------------------------------------- | ------------------------------------------ |
| `CONFIG_ZMK_SPLIT`                                      | bool | Enable split keyboard support                                              | n                                          |
| `CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS`            | bool | Enable split keyboard support for passing indicator state to peripherals   | n                                          |
| `CONFIG_ZMK_SPLIT_ROLE_CENTRAL`                         | bool | `y` for central device, `n` for peripheral                                 |                                            |
| `CONFIG_ZMK_SPLIT_BLE`                                  | bool | Use BLE to communicate between split keyboard halves                       | y                                          |
| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING`   | bool | Enable fetching split peripheral battery levels to the central side        | n                                          |
| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_PROXY`      | bool | Enable central reporting of split battery levels to hosts                  | n                                          |
| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_QUEUE_SIZE` | int  | Max number of battery level events to queue when received from peripherals | `CONFIG_ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS` |
| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_POSITION_QUEUE_SIZE`      | int  | Max number of key state events to queue when received from peripherals     | 5                                          |
| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_SPLIT_RUN_STACK_SIZE`     | int  | Stack size of the BLE split central write thread                           | 512                                        |
| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_SPLIT_RUN_QUEUE_SIZE`     | int  | Max number of behavior run events to queue to send to the peripheral(s)    | 5                                          |
| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_STACK_SIZE`            | int  | Stack size of the BLE split peripheral notify thread                       | 650                                        |
| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_PRIORITY`              | int  | Priority of the BLE split peripheral notify thread                         | 5                                          |
| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_POSITION_QUEUE_SIZE`   | int  | Max number of key state events to queue to send to the central             | 10                                         |
```
This ordering (moving `CONFIG_ZMK_SPLIT_ROLE_CENTRAL` up and then `CONFIG_ZMK_SPLIT_BLE` before BLE-specific options) makes more sense to me, but not a big deal.

caksoylar (approved):
Docs look good to me, noted one nit re: table ordering.

equiman:
> Interesting... Can you please try the latest code I just pushed? Also, getting the ZMK logs when this happens would be useful. Thanks for testing!

Sure, I'll try to do it once I finish something.
About the logs, is there some link or documentation about how to enable it and where the logs files are stored?

petejohanson:
> 
> Sure, I'll try to do it once I finish something.
> About the logs, is there some link or documentation about how to enable it and where the logs files are stored?  

https://zmk.dev/docs/development/usb-logging

bryanforbes:
I just tested this with my Glove80 on my MacBook Pro running Sonoma with the battery widget and both halves report their levels properly:

image

I was also able to update the Glove80's underglow status to use the new API for fetching the peripheral levels successfully.

equiman:
> Interesting... Can you please try the latest code I just pushed?

I've merged the last changes and it's working again 🎉🥳

image

app/src/split/bluetooth/central.c
```diff
@@ -265,6 +270,95 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
     return BT_GATT_ITER_CONTINUE;
 }
 
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
+
+static uint8_t peripheral_battery_levels[ZMK_SPLIT_BLE_PERIPHERAL_COUNT] = {0};
+
+int zmk_split_get_peripheral_battery_level(uint8_t source, uint8_t *level) {
+    if (source < 0 || source >= ZMK_SPLIT_BLE_PERIPHERAL_COUNT) {
```
joelspadin, line 278: `source` is unsigned, so it can't be less than 0.

Also, consider using `ARRAY_SIZE(peripheral_battery_levels)` so the bounds check stays in sync with the size of the array it's checking, even if that size is changed to use a different symbol later.

app/src/split/bluetooth/central.c
```diff
@@ -265,6 +270,95 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
...
+}
+
+K_WORK_DEFINE(peripheral_batt_lvl_work, peripheral_batt_lvl_change_callback);
+
+static uint8_t split_central_battery_level_notify_func(struct bt_conn *conn,
+                                                       struct bt_gatt_subscribe_params *params,
+                                                       const void *data, uint16_t length) {
+    struct peripheral_slot *slot = peripheral_slot_for_conn(conn);
+
+    if (slot == NULL) {
```
joelspadin, line 305: Minor: seems a bit inconsistent to use `slot == NULL` for one null check and `!data` for another.

app/src/split/bluetooth/central.c
```diff
@@ -265,6 +270,95 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
...
+    }
+
+    if (!data) {
+        LOG_DBG("[UNSUBSCRIBED]");
+        params->value_handle = 0U;
+        return BT_GATT_ITER_STOP;
+    }
+
+    LOG_DBG("[BATTERY LEVEL NOTIFICATION] data %p length %u", data, length);
+    uint8_t battery_level = ((uint8_t *)data)[0];
```
joelspadin, line 317: Do we need to verify that `length >= 1` or is that ensured elsewhere?

app/src/split/bluetooth/central.c
```diff
@@ -265,6 +270,95 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
...
+    LOG_DBG("Battery level: %u", battery_level);
+    struct zmk_peripheral_battery_state_changed ev = {
+        .source = peripheral_slot_index_for_conn(conn), .state_of_charge = battery_level};
+    k_msgq_put(&peripheral_batt_lvl_msgq, &ev, K_NO_WAIT);
+    k_work_submit(&peripheral_batt_lvl_work);
+
+    return BT_GATT_ITER_CONTINUE;
+}
+
+static uint8_t split_central_battery_level_read_func(struct bt_conn *conn, uint8_t err,
```
joelspadin, line 327: Comments from the above function apply here too.

app/src/split/bluetooth/central_bas_proxy.c
```diff
@@ -0,0 +1,92 @@
...
+    if (ev->source >= CONFIG_ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS) {
+        LOG_WRN("Got battery level event for an out of range peripheral index");
+        return ZMK_EV_EVENT_BUBBLE;
+    }
+
+    LOG_DBG("Peripheral battery level event: %u", ev->state_of_charge);
+
+    // 1 offsets for the service attribute, then offset 5 for each battery because that's how
+    // many attributes are added per battery
+    uint8_t index = 1 + (5 * ev->source);
```
joelspadin, line 81: You might as well just use `int` here. Not that anyone will likely try to set up something with 51 peripherals and cause this to overflow, but using an int that matches the register size can be more efficient, and the extra stack space is either zero or negligible depending on what the compiler decides to do.

app/src/split/bluetooth/central_bas_proxy.c
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    if (ev->source >= CONFIG_ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS) {
+        LOG_WRN("Got battery level event for an out of range peripheral index");
+        return ZMK_EV_EVENT_BUBBLE;
+    }
+
+    LOG_DBG("Peripheral battery level event: %u", ev->state_of_charge);
+
+    // 1 offsets for the service attribute, then offset 5 for each battery because that's how
```
joelspadin, line 79: >  then offset 5 for each battery because that's how many attributes are added per battery

Is there a `#define` somewhere for this, or is this just something baked into the spec and unlikely to change?

petejohanson: Tweaked this for defines to hopefully makes this clearer.

app/src/split/bluetooth/central_bas_proxy.c
petejohanson: Fair. Done.

app/src/split/bluetooth/central.c
petejohanson: Indeed. Fixed.

app/src/split/bluetooth/central.c
petejohanson: Good catch. Added checks for this in case we have an empty notification payload.

app/src/split/bluetooth/central.c
petejohanson: Made these consistent.

app/src/split/bluetooth/central.c
petejohanson: Good call. changed.

docs/docs/config/system.md
petejohanson: Fair. Tweaked the ordering.

caksoylar (approved):

app/src/split/bluetooth/central.c
```diff
@@ -265,6 +270,105 @@ static uint8_t split_central_notify_func(struct bt_conn *conn,
...
+}
+
+K_MSGQ_DEFINE(peripheral_batt_lvl_msgq, sizeof(struct zmk_peripheral_battery_state_changed),
+              CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_QUEUE_SIZE, 4);
+
+void peripheral_batt_lvl_change_callback(struct k_work *work) {
+    struct zmk_peripheral_battery_state_changed ev;
+    while (k_msgq_get(&peripheral_batt_lvl_msgq, &ev, K_NO_WAIT) == 0) {
+        LOG_DBG("Triggering peripheral battery level change %u", ev.state_of_charge);
+        peripheral_battery_levels[ev.source] = ev.state_of_charge;
```
chrisandreae, line 293: Should the `peripheral_battery_levels` value also be reset to 0 (or a sentinel value?) when the peripheral disconnects?

Additionally, should there be a matching API to permit consumers to ask about the peripheral's presence without subscribing to `zmk_split_peripheral_status_changed` and maintaining an independent local view on the state?

A consumer of the peripheral battery level such as a display would almost certainly want to display something different when the peripheral is disconnected.

caksoylar: Worth noting that https://github.com/zmkfirmware/zmk/pull/2033 adds an API for checking peripheral bonded/connected status.

chrisandreae: Oh, excellent. Well, then those would play nicely, then. 👍 

In addition to that though, I think it might still still be worth changing `zmk_split_get_peripheral_battery_level` here to return an error code if the specified peripheral is not connected.

petejohanson: > Oh, excellent. Well, those would play nicely, then. 👍
> 
> In addition to that though, I think it might still still be worth changing `zmk_split_get_peripheral_battery_level` here to return an error code if the specified peripheral is not connected.

Good call. Added that check and the API now returns `-ENOTCONN` in that scenario. Pushed the changes.

petejohanson: Ok, I've also added code to trigger an event on disconnect to set the level back to 0 for the newly disconnected peripheral as well. I believe that rounds this out enough.

equiman:
I've merged the branch with the last changes, and still working on windows:

image

bryanforbes:
I've been using the latest for roughly a day and I noticed that it seems to be reporting the right side (peripheral) battery level of my Glove80 as the main bluetooth battery level in macOS:

image
image

The battery widget (first image) is correctly ordering the halves: left on top (89%), right below it (100%). But the BT menu is reporting the battery level as 100% (right half).

joelspadin (approved):

petejohanson:
> I've been using the latest for roughly a day and I noticed that it seems to be reporting the right side (peripheral) battery level of my Glove80 as the main bluetooth battery level in macOS:
> image image
> 
> The battery widget (first image) is correctly ordering the halves: left on top (89%), right below it (100%). But the BT menu is reporting the battery level as 100% (right half).
> 
> Update: After a while (I'm not sure how long), the BT menu started reporting the left half percentage and the battery widget switched the order of the two halves from the images I uploaded earlier (left on bottom, right on top). Could this be some sort of race condition?

Not sure it's a race condition, as much as indeterminate behavior on how macOS handles multiple battery level characteristics split over two different battery GATT services... Initial testing seemed to show that they properly picked the characteristic marked as the `main`.

Are you perhaps on a different underlying Zephyr branch, testing the Zephyr 3.5 work? Not sure I cherry picked a needed Zephyr fix there (will be doing that momentarily)

bryanforbes:
> Are you perhaps on a different underlying Zephyr branch, testing the Zephyr 3.5 work?

No, I’m using MoErgo’s main (Zephyr 3.2) as my base and replaced their split battery reporting commit with the work here and updated the underglow status to use the new API. I’ll test this against a stock ZMK later this afternoon to try to narrow down the issue.

petejohanson:
> > Are you perhaps on a different underlying Zephyr branch, testing the Zephyr 3.5 work?
> 
> No, I’m using MoErgo’s main (Zephyr 3.2) as my base and replaced their split battery reporting commit with the work here and updated the underglow status to use the new API. I’ll test this against a stock ZMK later this afternoon to try to narrow down the issue.

I *suspect* you'll find the same thing occurring, and it's simply that the macOS selection of the battery level to display is non-deterministic.

petejohanson:
> I've been using the latest for roughly a day and I noticed that it seems to be reporting the right side (peripheral) battery level of my Glove80 as the main bluetooth battery level in macOS:
> image image
> 
> The battery widget (first image) is correctly ordering the halves: left on top (89%), right below it (100%). But the BT menu is reporting the battery level as 100% (right half).
> 
> Update: After a while (I'm not sure how long), the BT menu started reporting the left half percentage and the battery widget switched the order of the two halves from the images I uploaded earlier (left on bottom, right on top). Could this be some sort of race condition?

I don't think we're getting any help any time soon...

image

My gut is to not hold up our merge of this just because macOS doesn't do sane battery level characteristic selection. The new Sonoma battery widget actually shows the multiple levels. *Hopefully* it will do so with labels when we add the GATT CUD to each battery level.

Also, the proxy portion is optional, and not enabled by default so there's less risk of this being an issue while we continue to iterate.

@bryanforbes Any concerns?

bryanforbes:
No concerns from me

dudicoco:
I've added the parameters to my config but I do not see peripheral battery status on macOs:
 
Screenshot 2024-01-05 at 17 11 29

https://github.com/dudicoco/corne-wireless-zmk-config/blob/dacb95898f9254b86278ddb7112ca28b014f9655/config/corne.conf#L20-L23

madushan1000:
here is a python script to see split battery levels on linux. https://gist.github.com/madushan1000/9744eb6350a5dd9685fb6bfbb25fbb8a

dstudzinski:
> I've been using the latest for roughly a day and I noticed that it seems to be reporting the right side (peripheral) battery level of my Glove80 as the main bluetooth battery level in macOS:
> 
> 
> 
> image
> 
> image
> 
> 
> 
> The battery widget (first image) is correctly ordering the halves: left on top (89%), right below it (100%). But the BT menu is reporting the battery level as 100% (right half).
> 
> 
> 
> Update: After a while (I'm not sure how long), the BT menu started reporting the left half percentage and the battery widget switched the order of the two halves from the images I uploaded earlier (left on bottom, right on top). Could this be some sort of race condition?

What's your widget/app for macos? I can't find any.

bryanforbes:
> What's your widget/app for macos? I can't find any.

It’s just the battery widget in Sonoma in the notification center.

bjornborg:
> here is a python script to see split battery levels on linux. https://gist.github.com/madushan1000/9744eb6350a5dd9685fb6bfbb25fbb8a

Great script. I have fused this to another gnome extension, so it is graphically shown in gnome.
image
The repo is located here: https://github.com/bjornborg/gnome-bluetooth-battery-indicator/tree/gnome.42

mh4x0f:
I created an app inspired by the @Maksim-Isakau  project just for Linux on Dbus Bluez.

image


https://github.com/mh4x0f/zmkBATx

dudicoco:
> I've been using the latest for roughly a day and I noticed that it seems to be reporting the right side (peripheral) battery level of my Glove80 as the main bluetooth battery level in macOS:
> 
> image image
> The battery widget (first image) is correctly ordering the halves: left on top (89%), right below it (100%). But the BT menu is reporting the battery level as 100% (right half).
> 
> Update: After a while (I'm not sure how long), the BT menu started reporting the left half percentage and the battery widget switched the order of the two halves from the images I uploaded earlier (left on bottom, right on top). Could this be some sort of race condition?

@bryanforbes can you please share what widget are you using for mac?

dudicoco:
Figured out how to display the split battery status on mac:
1. Follow the instructions on how to add a widget - https://support.apple.com/en-il/guide/mac-help/mchl52be5da5/mac.
2. Add the battery widget.

mike1808:
> here is a python script to see split battery levels on linux. https://gist.github.com/madushan1000/9744eb6350a5dd9685fb6bfbb25fbb8a

thanks! created a polybar module if anyone uses it with i3 or other wms

https://gist.github.com/mike1808/13af77145a292c627fa6508f5fb6730b
image

rottencandy:
> here is a python script to see split battery levels on linux. https://gist.github.com/madushan1000/9744eb6350a5dd9685fb6bfbb25fbb8a

Thank you, I made a bash version of this script
https://gist.github.com/rottencandy/901189a3cfac94bcbdefdf40236b042e

linkarzu:
@bryanforbes or any macOS user out there, has someone been able to get the battery values via CLI?
The reason I'm asking, is because I want to get the values, and show them in SketchyBar and probably set a notification when battery is around 5% so that I charge the keyboard

As seen on image2, I do see the values in the macOS battery widget, but I'd like to get it via a terminal command

<div align="left">
    
</div>

<div align="left">
    
</div>

PR #2052, opened 2023-12-04
refactor: Improve keymap alignment

filterpaper:
Refactor Corne and Lily58 keymaps to reduce editing errors:
* Align codes with consistent indentation
* Add grid lines for guidance


nmunnich:
Rather than these sorts of changes I'd prefer to see work be put into integrating a devicetree formatter. This approach just causes what I would consider unnecessary churn.

PR #1989, opened 2023-11-06
refactor: Hook into CMake loading in a better spot.

petejohanson:
* Shift to using an extra Zephyr module to do keymap location
  work after all board roots are resolved. This avoids duplicate work
  and allows us to load custom boards from Zephyr modules as well as
  user config setups.

PR #2054, opened 2023-12-04
fix(docs): Fix text blurb about split Kconfig.defconfig

caksoylar:
This fixes some outdated text in the description of `Kconfig.defconfig` for split keyboards. Since #1989 is touching new shield guide as well, I can rebase once that is merged -- just wanted to have this here so I don't forget about it.

### DIFF
--- a/docs/docs/development/new-shield.md
+++ b/docs/docs/development/new-shield.md
@@ -113,10 +113,11 @@ config ZMK_KEYBOARD_NAME
 endif
 ```
 
-Similarly to defining the halves of a split board in `Kconfig.shield` it is important to set the `ZMK_KEYBOARD_NAME` for each half of a split keyboard.
-You'll also want to set which half is the central side. Most boards set it to the left.
-Then on the peripheral half, you'll want to turn USB on so that it shows USB status on displays properly.
-Finally, you'll want to turn on the split option for both sides. This can all be seen below.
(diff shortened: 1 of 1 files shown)


docs/docs/development/new-shield.md
```diff
@@ -92,10 +92,10 @@ config ZMK_KEYBOARD_NAME
 endif
 ```
 
-Similarly to defining the halves of a split board in `Kconfig.shield` it is important to set the `ZMK_KEYBOARD_NAME` for each half of a split keyboard.
-You'll also want to set which half is the central side. Most boards set it to the left.
-Then on the peripheral half, you'll want to turn USB on so that it shows USB status on displays properly.
-Finally, you'll want to turn on the split option for both sides. This can all be seen below.
+For split keyboards, `Kconfig.defconfig` needs to specify a few more options.
+First, it configures one side as the central, which is usually the left side.
```
joelspadin, line 96: Pedantic: the code sample below sets the name and central side in the opposite order as is described here.

caksoylar: Good point. I reworded it, please have a look again.

joelspadin (approved):

docs/docs/development/new-shield.md
joelspadin: Looks reasonable to me.

PR #2056, opened 2023-12-04
fix(docs): Update outdated information in the hardware page

caksoylar:
It looks like the text in this page hasn't been touched in some time. I updated it the way I thought best to remove the outdated info, but I am open to feedback!

### DIFF
--- a/docs/docs/hardware.mdx
+++ b/docs/docs/hardware.mdx
@@ -41,7 +41,7 @@ export const toc = [
 ];
 
 With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets.
-That being said, there are currently only a few specific [boards](/docs/faq#what-is-a-board)/[shields](faq.md#what-is-a-shield) that have been implemented and tested by the ZMK contributors.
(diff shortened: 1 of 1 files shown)


docs/docs/hardware.mdx
```diff
@@ -41,25 +41,19 @@ export const toc = [
 ];
 
 With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets.
-That being said, there are currently only a few specific [boards](/docs/faq#what-is-a-board)/[shields](faq.md#what-is-a-shield) that have been implemented and tested by the ZMK contributors.
+That being said, there are specific [boards](/docs/faq#what-is-a-board)/[shields](faq.md#what-is-a-shield) that have been implemented and tested by the ZMK contributors, listed below.
```
joelspadin, line 44: These two links go to the same page, but one is absolute and the other is relative. Would be nice to make that consistent.

caksoylar: Good catch, thanks.

petejohanson (approved):
LGTM.

PR #2057, opened 2023-12-04
feat(bt): Add new beta testing Kconfig setting.

petejohanson:
* In order to have an easy way to test and phase in BT changes, add a
  new `ZMK_BLE_BETA_TESTING` Kconfig setting.


docs/docs/config/bluetooth.md
```diff
@@ -9,7 +9,8 @@ See [Configuration Overview](index.md) for instructions on how to change these s
...
-| `CONFIG_BT_GATT_ENFORCE_SUBSCRIPTION` | bool | Low level setting for GATT subscriptions. Set to `n` to work around an annoying Windows bug with battery notifications.                      | y       |
+| Option                                | Type | Description                                                                                                                                                                                                                                            | Default |
+| ------------------------------------- | ---- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------- |
+| `CONFIG_ZMK_BLE_BETA_TESTING`         | bool | Enables a combination of settings that are planned to be default in future versions of ZMK. This includes changes to timing on BLE pairing initation, BT Secure Connection passkey entry, and allowing overwrite of keys from previously paired hosts. | n       |
```
caksoylar, line 14: Are we planning to do this one time, for these specific changes, or keep it in the future for other changes? The reason for the question is if it is the latter then we might want to not note the specifics here and perhaps link to the Kconfig file instead?

petejohanson: I imagine us keeping it and tweaking/evolving it as we eventually migrate the ZMK defaults... Not sure it hurts to duplicate it here so it's just a tad more accessible, and we update both spots as needed.

Thoughts?

caksoylar: If we are OK keeping this up to date as things change then that is likely the best solution.

joelspadin:
By calling this "beta testing", I worry it might be confused with https://zmk.dev/docs/features/beta-testing. What do you think about something more like `CONFIG_ZMK_BLE_EXPERIMENTAL_FEATURES`? Alternatively, we could add a section to the beta testing page that shows how to enable this flag.

petejohanson:
> By calling this "beta testing", I worry it might be confused with https://zmk.dev/docs/features/beta-testing. What do you think about something more like `CONFIG_ZMK_BLE_EXPERIMENTAL_FEATURES`? Alternatively, we could add a section to the beta testing page that shows how to enable this flag.

Not a bad idea to clearly differentiate these two things. Not sure I want to add to the beta-testing page... That's nicely focused right now.

caksoylar (approved):

Issue #2059, opened 2023-12-06
build-user-config action no longer copy parent directories


TristanCacqueray:
It seems like since https://github.com/zmkfirmware/zmk/pull/1989 , config using https://github.com/urob/zmk-nodefree-config no longer build because of such errors:

```
 /tmp/tmp.vVFsbR0CcL/config/corne.keymap:17:10: fatal error: ../zmk-nodefree-config/keypos_def/keypos_42keys.h: No such file or directory
   17 | #include "../zmk-nodefree-config/keypos_def/keypos_42keys.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

vvision:
Reproduced on my side today too.
```
/tmp/tmp.UUyIFrP4xQ/config/splitkb_aurora_corne.keymap:13:10: fatal error: ../src/corne.dtsi: No such file or directory
   13 | #include "../src/corne.dtsi"
      |          ^~~~~~~~~~~~~~~~~~~
```
I'm also gonna pin the version of `zmkfirmware/zmk/.github/workflows/build-user-config.yml`.

petejohanson:
Can you give this a try?

https://github.com/petejohanson/revxlp-config/actions/runs/7132623640/workflow

Doing so will *only* copy if we detect the new Zephyr module bits, so we shouldn't break existing configs with that change. If it works for you I'll get this into a PR.

ObiWanKeoni:
@petejohanson that seems to [work](https://github.com/ObiWanKeoni/zmk-config/actions/runs/7135073855) for my config!

lou-bi:
Hi, got this exact issue while building my firmware with Urob (not as a submodule).
I've follow your link @petejohanson , but Im not sure I did what I should. https://github.com/lou-bi/zmk-config/commit/4b95d82d875f61b568f7a87b878df674533644f1#annotation_16243079431

What are we supposed to do with this build.yaml you linked ?

And will this be fixed in a later released ? Im more comfortable using default build params. For no reasons propably.

ObiWanKeoni:
@lou-bi it looks like you've modified your `build.yaml` instead of your `.github/build.yml`, I think reverting your first change and applying Pete's fix will help move you along

> And will this be fixed in a later released ? Im more comfortable using default build params. For no reasons propably.

[...]

PR #2064, opened 2023-12-08
fix(build): Only copy if zephyr module detected

petejohanson:
* To fix issues with breakage with existing user config repos, only do our copy strategy if we detect a zephyr module present in the repo as well.


caksoylar (approved):
LGTM, does it need additional testing?

petejohanson:
> LGTM, does it need additional testing?

Tested by a couple folks in #2059 and I tested in one of my config repos as well, no issue. Merging.

PR #2066, opened 2023-12-11
Fix: Use a lookup table & interpolation to better estimate remaining battery capacity

MickiusMousius:
The current battery capacity estimate uses a linear approximation to estimate the remaining battery capacity. This is computationally very efficient, however, the estimate is very inaccurate once below 60%.

This PR uses a small table of bounded functions to provide a more accurate estimate of remaining capacity.

The original data to create the estimate was taken from here: https://blog.ampow.com/lipo-voltage-chart/

The accuracy of the functions (in relation to the lookup chart least) can be seen below:

Screenshot 2023-12-11 at 17 23 23

In terms of testing I've used a Python version of the algorithm to validate the behaviour. I am also running the change on my daily driver keyboard.

Adding a little to the data/numbers on this: If a search is done for "lipo discharge lookup chart" in Goole/DuckDuckGo image search these numbers keep coming up or numbers that are within a few millivolts of those that I've used. So long as the discharge rate is substantially below C (most charts show 0.2C) then the shape and values are very consistent.

### DIFF
--- a/app/module/drivers/sensor/battery/battery_common.c
+++ b/app/module/drivers/sensor/battery/battery_common.c
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
     return 0;
 }
 
-uint8_t lithium_ion_mv_to_pct(int16_t bat_mv) {
-    // Simple linear approximation of a battery based off adafruit's discharge graph:
-    // https://learn.adafruit.com/li-ion-and-lipoly-batteries/voltages
+uint8_t lithium_ion_mv_to_pct(int16_t batt_mv) {
+    // Lookup table of slope formulas for calculating remaining battery capacity.
+    //
+    // The current set of values acheives the same as: bat_mv * 2 / 15 - 459;
+    //
+    // Future updates sould include more complete lookup tables for different
+    // battery types slectable through config flags.
+    struct lookup_point {
+        int16_t millivolts;
+        int16_t percent;
+    };
 
-    if (bat_mv >= 4200) {
-        return 100;
-    } else if (bat_mv <= 3450) {
-        return 0;
+    static const struct lookup_point battery_lookup[] = {
+        {.millivolts = 4200, .percent = 100},
+        {.millivolts = 3450, .percent = 0},
+    };
+
+    if (batt_mv > battery_lookup[0].millivolts) {
+        return battery_lookup[0].percent;
     }
 
-    return bat_mv * 2 / 15 - 459;
+    for (int i = 1; i < ARRAY_SIZE(battery_lookup); i++) {
+        struct lookup_point one = battery_lookup[i - 1];
+        struct lookup_point two = battery_lookup[i];
+        if (batt_mv >= two.millivolts) {
+            const int t = batt_mv - one.millivolts;
+            const int dx = two.millivolts - one.millivolts;
+            const int dy = two.percent - one.percent;
+            return one.percent + dy * t / dx;
+        }
+    }
+
+    return battery_lookup[ARRAY_SIZE(battery_lookup) - 1].percent;
 }


joelspadin (changes requested):

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,36 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
+
 uint8_t lithium_ion_mv_to_pct(int16_t bat_mv) {
-    // Simple linear approximation of a battery based off adafruit's discharge graph:
-    // https://learn.adafruit.com/li-ion-and-lipoly-batteries/voltages
+    // Lookup table of slope formulas for calculating remaining battery capacity.
+    // The original values used to calculate slopes come from:
+    //
+    //       https://blog.ampow.com/lipo-voltage-chart/
+
+    const batt_est_formula lipo_formula_lookup[9] = {
```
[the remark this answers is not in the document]
MickiusMousius: I'll do It this way, it is more work for the processor, but is way easier to update the lookup table if someone feels so inclined.

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,36 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
     return 0;
 }
 
+typedef struct {
```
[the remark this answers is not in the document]
MickiusMousius: I'll fix this.

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,36 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
 
-    if (bat_mv >= 4200) {
-        return 100;
-    } else if (bat_mv <= 3450) {
-        return 0;
+    for (int i = 0; i < 9; i++) {
+        batt_est_formula current_formula = lipo_formula_lookup[i];
+        if (bat_mv >= current_formula.millivolts) {
+            float capacity = bat_mv * current_formula.slope + current_formula.offset;
+            return (uint8_t)capacity;
```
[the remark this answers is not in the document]
MickiusMousius: Yes, I figure it will read 100% for a little bit, it will also read 0% for a little bit before dying (depending on the battery protection IC at least). This is the kind of behaviour I'd want, so hopefully this holds true for others.

app/module/drivers/sensor/battery/battery_common.c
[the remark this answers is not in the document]
MickiusMousius: I've updated it now, here's the plotted graph and Python code:

`
battery_lookup = [
  {"millivolts": 4200, "percent": 100},
  {"millivolts": 3870, "percent": 60},
  {"millivolts": 3690, "percent": 10},
  {"millivolts": 3610, "percent": 5},
  {"millivolts": 3270, "percent": 0},
]


def batt_capacity(batt_mv):
  if batt_mv > 4200:
    return 100

  for i in range(1, len(battery_lookup)):
    one = battery_lookup[i - 1]
    two = battery_lookup[i]
    if batt_mv >= two["millivolts"]:
      slope = (one["percent"] - two["percent"]) / (one["millivolts"] - two["millivolts"])
      offset = one["percent"] - slope * one["millivolts"]
      return (slope * batt_mv) + offset

  return 0


x = list (range(3000, 4300))
y = [batt_capacity(mv) for mv in x]
plt.plot(y, x)
plt.ylabel('Millivolts')
plt.xlabel('State of charge')
plt.show()
`
Figure_1

And compared to the source data:

Screenshot 2023-12-12 at 12 52 35

(Also, thanks for the feedback 😁)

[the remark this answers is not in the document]
MickiusMousius: There’s an error in the C code, the counter should start at “1”, it is starting at “0”.

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
+    };
 
-    if (bat_mv >= 4200) {
+    static const struct lookup_point battery_lookup[] = {
+        {.millivolts = 4200, .percent = 100}, {.millivolts = 3870, .percent = 60},
+        {.millivolts = 3690, .percent = 10},  {.millivolts = 3610, .percent = 5},
+        {.millivolts = 3270, .percent = 0},
+    };
+
+    if (batt_mv > 4200) {
```
joelspadin, line 49: This should be

```c
if (batt_mv > battery_lookup[0].millivolts)
```

to ensure it can't get out of sync with the lookup table.

If this were a general linear interpolation routine, it should also use

```c
  return battery_lookup[0].percent;
```

but since the first entry in the table should always be 100%, returning 100 is fine.

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
-    // Simple linear approximation of a battery based off adafruit's discharge graph:
-    // https://learn.adafruit.com/li-ion-and-lipoly-batteries/voltages
+uint8_t lithium_ion_mv_to_pct(int16_t batt_mv) {
+    // Lookup table of slope formulas for calculating remaining battery capacity.
+    // The original values used to calculate slopes come from:
+    //
+    //       https://blog.ampow.com/lipo-voltage-chart/
+    //
+    struct lookup_point {
+        float millivolts;
```
joelspadin, line 39: The constants in the table are all integers, so we can save a few bytes by storing these fields as `int16_t` instead of 32-bit floats.

Then we can either cast to float when doing the math that uses these numbers, or we can change the algorithm so it works with integers as I suggest below.

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
-}
\ No newline at end of file
+    for (int i = 1; i < ARRAY_SIZE(battery_lookup); i++) {
+        struct lookup_point one = battery_lookup[i - 1];
+        struct lookup_point two = battery_lookup[i];
+        if (batt_mv >= two.millivolts) {
+            float slope = (one.percent - two.percent) / (one.millivolts - two.millivolts);
+            float offset = one.percent - slope * one.millivolts;
+            float batt_percent = (slope * batt_mv) + offset;
+            return (uint8_t)batt_percent;
```
joelspadin, line 60: Minor: Cast to `uint8_t` is redundant. That's the return type, so it automatically gets cast to that.

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
-    return bat_mv * 2 / 15 - 459;
-}
\ No newline at end of file
+    for (int i = 1; i < ARRAY_SIZE(battery_lookup); i++) {
+        struct lookup_point one = battery_lookup[i - 1];
+        struct lookup_point two = battery_lookup[i];
+        if (batt_mv >= two.millivolts) {
+            float slope = (one.percent - two.percent) / (one.millivolts - two.millivolts);
+            float offset = one.percent - slope * one.millivolts;
+            float batt_percent = (slope * batt_mv) + offset;
```
joelspadin, line 59: This can be simplified a bit to use only one multiply and one divide. If the divide is placed last, then it should also work using only integer math, so there's no need to cast anything to a float anywhere.

The formula for linear interpolation between two points (x0, y0) and (x1, y1) is

```c
y = y0 + (y1 - y0) * (x - x0) / (x1 - x0)
```

which using the variable names here is

```c
return one.percent + (two.percent - one.percent) * (batt_mv - one.millivolts) / (two.millivolts - one.millivolts);
```

or pulling some of that out into separate variables to make it easier to read

```c
const int t = batt_mv - one.millivolts;
const int dx = two.millivolts - one.millivolts;
const int dy = two.percent - one.percent;

return one.percent + dy * t / dx;
```

(note that because you ordered the table in descending order, many of the values here would end up negative, but they will all cancel out)

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
+            return (uint8_t)batt_percent;
+        }
+    }
+
+    return 0;
```
joelspadin, line 64: Optional: If this were a general linear interpolation routine, this should be

```c
return battery_lookup[ARRAY_SIZE(battery_lookup) - 1].percent;
```

but since the last entry should always be 0%, this is fine.

joelspadin (changes requested):

app/module/drivers/sensor/battery/battery_common.c
MickiusMousius: Excellent suggestion!

app/module/drivers/sensor/battery/battery_common.c
MickiusMousius: Huh, I didn't know C++ did that, nice :)

app/module/drivers/sensor/battery/battery_common.c
MickiusMousius: Hmmm, I should have looked up the standard form for linear interpolation!

I just did what I see the numbers doing, this is much much nicer!

More generally you've spent more time nursing this PR than it would have taken you to do it yourself, I really appreciate that. I've learned a little about C++ as a result.

app/module/drivers/sensor/battery/battery_common.c
joelspadin: Yep. C/C++ will do that any time you assign a variable of one type to something of another, for instance
```c
void do_stuff(float y) { ... }

float f = 1.5;
int x = f;  // automatically casts from float to int

do_stuff(x);  // automatically casts from int to float
```
(either that or it will error because the types can't be converted)

joelspadin (commented):
Probably need to get a few more people to test this and make sure it works as expected, but the code looks good to me now.

MickiusMousius:
Wonderful, thanks for that.

Is the testing something that usually handles itself or should I try to recruit people to test or…?

joelspadin:
I put a messsage in the testing channel on Discord, but not sure if anyone has seen it.

@MickiusMousius Could you rebase this onto the tip of main or merge the tip of main into it? I think there were some issues with the GitHub actions that caused the checks to fail, which aren't related to your change, and the recent changes on main should fix that.

MickiusMousius:
Done

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
+    };
+
+    if (batt_mv > battery_lookup[0].millivolts) {
+        return battery_lookup[0].percent;
     }
 
-    return bat_mv * 2 / 15 - 459;
-}
\ No newline at end of file
+    for (int i = 0; i < ARRAY_SIZE(battery_lookup); i++) {
```
joelspadin, line 53: Either I missed this on the last review, or you changed it since I looked at it. This loop should start at 1 since you're looking at `i` and `i - 1`

MickiusMousius: Yep, somehow I changed it, though the change was definitely not deliberate. Either way I've fixed this.

I've had a poke around in repo but I can not seem to find any tests for sensor code, admittedly C++ is definitely not my best language, but when I used to code for a living I'd typically implement unit tests for something like this (partly for my own sanity). Is there anything I can use as a reference for implementing some tests?

joelspadin: We have test for keymap features, but those work by logging output and diffing it against the expected output. I'm not sure how difficult it would be to bring in a unit testing framework for general code.

MickiusMousius: Merged again, I noticed there are a few more changes since last time :)

joelspadin:
If anyone would like to help validate whether this does improve things, I wrote a Windows program to record and graph battery levels over time: https://github.com/joelspadin/BleBatteryRecorder

To test this PR, first load software that does *not* include this PR and record the battery going from a full charge to empty. Then, load software that does include this PR and repeat. Graph the results. If we can confirm that the second line is more linear than the first, then we can merge this.

sctanf:
It may be not a good idea to take random data from a site for this, since the battery type and discharge method can be different.
A while ago I graphed data from discharging a battery with constant current and constant power (like a buck regulator), with a more common li-ion batterie rather than lipo, which can make a significant difference.
Discharge curve using linear regulator|Discharge curve using buck regulator
---|---
%_V Constant Current|%_V Constant Power

MickiusMousius:
@sctanf Sorry for the slow reply, suffice to say it's been a busy few months. I do agree that the *optimal* solution here would be to log the energy consumption from a typical set of batteries with the same power profile as a typical ZMK keyboard to find the most accurate "common curve".

That said, I am also a firm believer in the idea of aiming to make things better with a "perfect" end goal being easier to reach as a result.

Personally I do not have the tools to record this profile (or the time), however, I decided that finding the a common profile for the most common chemistry and then implementing a solution that allowed for a better curve later is a step along the path to "perfect".

If you're able to supply a more representative discharge curve using the most common parts (controller, battery size & chemistry etc) that would get things much closer to perfect IMO. I imagine most people would welcome that (I certainly would).

joelspadin:
My biggest concern is that we don't yet have any data that says this is better than the linear approximation. I do think using a lookup table is an improvement since it will let us use any piecewise linear function we want, but I want to make sure this new function isn't *worse* than what we have already.

MickiusMousius:
Anecdotally I suspect that it is “equivalent”. Overall it felt better to start off, but the end of the curve is still quite poor.

The final 10-15% of charge lasts more than a week on my work keyboard.

The first 10% lasts a few days at most (better than before, but not great).

MickiusMousius:
@joelspadin I've tested with a nice nano controller, the difference compared to a supermini is not significant in my opinion. I'm happy to purchase a tool to record the battery over time.

However, I'd like the following features:
- Able to log without being attached to a computer
- Able to record both voltage & current (I could log the voltage drop across a small resistor too I guess)

Do you have any recommendations?

@sctanf If you've got any suggestions that would be welcome too.

joelspadin:
https://www.nordicsemi.com/Products/Development-hardware/Power-Profiler-Kit-2 is probably the best way to monitor power usage, but I think it does still require the profiler be connected to a computer. (It's also a bit expensive. I have one myself, but I'd need to get a new battery to test this well, since all my batteries under 500 mAh are very dead. Probably also would want to test on a variety of batteries.)

MickiusMousius:
I have a few batteries that are all new and unused, at least the more common sizes used by splits like the Corne and Sofle (e.g 301230, 801230).

I looked at the PPK2 but could not tell if it also recorded voltage when in amp meter mode. If it can I’ll buy one as it’ll be handy for a few upcoming projects anyway. Do you know if it also logs voltage in amp meter mode?

I’ll need to get my hands on a computer I can leave logging. I suspect my work may have some NUCs collecting dust on a shelf I can borrow for a few weeks.

nmunnich:
Couple thoughts:

1. I'd like to see this new approach be locked behind a Kconfig flag. We can then leave the current function as the default for now, and switch once we've gotten more validation that the lookup table is equivalent or an improvement, as Joel said.
2. This PR needs its commits adjusted: https://zmk.dev/docs/development/contributing/pull-requests (recent development)

After those points are covered (with appropriate docs changes to cover the Kconfig flag) I would be strongly in favor of merging, as the rest looks solid to me and the concept is good.

joelspadin:
Rather than locking it behind a Kconfig flag and needing to maintain two versions of the function, could we instead switch over to the lookup table fully, but make the table match the original function? Then we could test adjustments to the table.

nmunnich:
Sure, I'd be happy with that. Push through the lookup table first and then pushing through improvements to the lookup table in a followup would be a solid idea.

MickiusMousius:
I’ll try and update it later this week to replicate the existing function.

— over the horizon —

I’ve also recently purchased a PPK2 so I can make some “real life” discharge curves for later on.

I suspect I’ll need to produce 2 such curves, one against a linear regulator and one using a switching regulator based on what scantf said. Hopefully the overlap between the two will be very close. When I get to that I’ll just submit a separate PR with all the information. It’s a test that will likely take months to conclude given the very impressive battery life these devices offer.

caksoylar:
Separately, like This sort of extension would allow us to also specify e.g. a curve for li-ion batteries, for which current approximation is not accurate.

sctanf:
The PPK2 can only supply a voltage, not measure it. It might still be interesting to see how current changes as the battery voltage drops, along with profiling the power usage in general.

You will likely want to track the battery voltage reading off of the device itself anyway, and those readings could be used to create a discharge profile. I've done a similar process to create profiles before, although now I realize that there is a significant variation between different batteries. A generic profile might be a little difficult.

I'm not sure the best way to take voltage readings from the device, especially over a few months. For short tests with a higher discharge rate, I attached a debugger and logged the voltage to a file.

MickiusMousius:
Change made as requested. I'm running it on my secondary keyboard now, my main uses a max17048.

As an aside, looking at the conversation here I feel as though there is not agreement on what an acceptable margin of error should be.

Personally I think an estimate that is within a few percent would be fine. More importantly from a UX perspective I'd want to _feel_ like the battery is draining at a consistent rate, this will give me a sense of confidence in the numbers, even if they're not 100% accurate.

I am unsure what the community would deem to be best here however.

nmunnich:
The points which I personally care about the most are the 80% and 20% points of the battery (not measured values, but physical). I'd like it if ZMK could attempt to limit the battery charge state within those points for maximum battery health in the future. That makes a lower margin of error much more significant, though.

Such discussion are better left for the desired followup issue though, imo. This PR looks good to me now (though it would be appreciated if you could squash the PRs with a suitable commit message, this could also be done by whoever merges it.) I for one can't picture us adding a lookup table with enough elements that it would make sense to use binary search over the linear search.

joelspadin:
I think having ZMK limit the charge to 80% is something completely out of scope for this PR. ZMK can't control the battery charger unless the hardware is specifically designed to allow it. Many controllers, like the nice!nano, have the charger operating completely independently with no way to disable it.

If your hardware _does_ support disabling the charger, it would already be possible to write custom code that watches the battery charge level and disables the charger when it gets to a certain level. You would need to disable the charger prior to taking a voltage reading as well, or you will be reading the charge voltage rather than the battery voltage.

nmunnich:
> I think having ZMK limit the charge to 80% is something completely out of scope for this PR. 

Oh absolutely, but I think in general any further discussion in this direction is out of scope of this PR now. Topics like the 80% stuff and better estimation tables and all that should be in other issues/PRs.

MickiusMousius:
> If your hardware _does_ support disabling the charger, it would already be possible to write custom code that watches the battery charge level and disables the charger when it gets to a certain level.

Funnily enough I have a design I'm working on that does just this (hopefully), the prototype PCBs arrive this week :)

I've squashed the commit history per nmunnich's request.

When doing the next PR I'll quantify measurements & errors. Hopefully it'll be possible to produce one curve that will be good enough for LiPo and another for LiIon, but we'll see. As before that is a little way off, I've a few other projects to complete yet & running these tests will take time.

nmunnich (approved):

app/module/drivers/sensor/battery/battery_common.c
```diff
@@ -29,15 +29,37 @@ int battery_channel_get(const struct battery_value *value, enum sensor_channel c
...
-        return 100;
-    } else if (bat_mv <= 3450) {
-        return 0;
+uint8_t lithium_ion_mv_to_pct(int16_t batt_mv) {
+    // Lookup table of slope formulas for calculating remaining battery capacity.
+    //
+    // The current set of values acheives the same as: bat_mv * 2 / 15 - 459;
+    //
+    // Future updates sould include more complete lookup tables for different
+    // battery types slectable through config flags.
```
joelspadin, line 38: A few typos. Also, a bit pedantic, but this is a lookup table of `{voltage, state_of_charge}` points, not a lookup table of formulas.

```suggestion
    // Lookup table for calculating remaining battery capacity.
    //
    // The current set of values achieves the same as: bat_mv * 2 / 15 - 459;
    //
    // Future updates should include more complete lookup tables for different
    // battery types selectable through config flags.
```

joelspadin (commented):
I think Pete should approve this since it is a design change, but the code looks good to me.

MickiusMousius:
Sorry for the slow reply.

The more I think about how something like this would be used in practice the more I feel that it should be a module & not a core part of ZMK. I fully anticipate that I will be building lookup tables for specific batteries for some of my projects, this will vary from board to board and battery to battery. I expect others will be in the same boat.

I expect that some pre-baked lookup tables for common battery types would be helpful, though right now the module seems like it makes this approach accessible to the largest amount of users.

I'm happy to fix my typos etc, but can't help but wonder if this is the wrong approach entirely. Admittedly I have no idea how to make a battery reporting module, but there's really only one way to learn 🤷‍♂️

nmunnich:
Hmmm, interesting thought. This PR is only for turning voltage measurements into percentages. I think a module might be *too* flexible for this purpose. What I *could* see would be, rather than having the lookup table built in, having the lookup table be read from devicetree. Then we could predefine some lookup tables, like how we predefine behaviors, but anyone could define their own lookup table if they have a weird battery. That wouldn't allow for more advanced algorithms that take the battery's behavior over time into account, but I would argue for that you *would* want it in a module. Strictly speaking since you can pick your battery measurement tool (MAX17048, voltage measurement, ...) you can already make that stuff in a module.

PR #2067, opened 2023-12-12
feat(build): Add release workflow steps

filterpaper:
* Add additional steps to publish firmware artifacts as a release in the user repository.
* User can disable the feature and also customise the name and tag.

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -23,6 +23,21 @@ on:
         default: "firmware"
         required: false
         type: string
+      release_enable:
+        description: "Enable publishing of artifacts as a release"
+        default: true
+        required: false
+        type: boolean
+      release_name:
+        description: "Release name"
+        default: "ZMK Firmware"
+        required: false
+        type: string
+      release_tag:
+        description: "Release tag"
+        default: "latest"
+        required: false
+        type: string
 
 jobs:
   matrix:
@@ -161,3 +176,25 @@ jobs:
         with:
           name: ${{ inputs.archive_name }}
           path: ${{ env.build_dir }}/artifacts
+
+  publish:
+    runs-on: ubuntu-latest
+    if: ${{ inputs.release_enable && !cancelled() }}
+    needs: build
+    name: Publish firmware
+    steps:
+      - name: Delete previous release
+        uses: dev-drprasad/delete-tag-and-release@v1.0.1
+        with:
+          tag_name: ${{ inputs.release_tag }}
+          github_token: ${{ github.token }}
+
+      - name: Download artifact
+        uses: actions/download-artifact@v3.0.2
+
+      - name: Publish ${{ inputs.release_tag }} release
+        uses: softprops/action-gh-release@v0.1.15
+        with:
+          name: ${{ inputs.release_name }}
+          tag_name: ${{ inputs.release_tag }}
+          files: "**/*.*"


.github/workflows/build-user-config.yml
```diff
@@ -161,3 +176,25 @@ jobs:
...
           name: ${{ inputs.archive_name }}
           path: ${{ env.build_dir }}/artifacts
+
+  publish:
+    runs-on: ubuntu-latest
+    if: ${{ inputs.release_enable && !cancelled() }}
+    needs: build
+    name: Publish firmware
+    steps:
+      - name: Delete previous release
```
joelspadin, line 186: Why would you want to delete previous releases? As noted above, I don't think we want to put in any behaviors that rewrite Git history, especially not by default.

.github/workflows/build-user-config.yml
```diff
@@ -23,6 +23,21 @@ on:
...
+        required: false
+        type: boolean
+      release_name:
+        description: "Release name"
+        default: "ZMK Firmware"
+        required: false
+        type: string
+      release_tag:
+        description: "Release tag"
+        default: "latest"
```
joelspadin, line 38: It seems like you're just constantly re-tagging whatever commit is getting built as "latest" every time the workflow is run? Rewriting Git history is not great, and probably not something we want to do by default (or even support at all). What happens if you push two commits in quick succession, and "latest" gets moved in the middle of one of the builds?

Typically, you would use action-gh-release by setting up a separate workflow that triggers when a tag is added, or by adding it to the existing workflow and gating the release steps with
```yaml
  if: startsWith(github.ref, 'refs/tags/')
```

so that it creates a new release for every tag you add. Then, a release is only created when you explicitly ask for one by creating a new tag, and since each new tag has a different name, all your releases stick around and reference the specific commits from which they were built.

.github/workflows/build-user-config.yml
```diff
@@ -161,3 +176,25 @@ jobs:
...
+        uses: dev-drprasad/delete-tag-and-release@v1.0.1
+        with:
+          tag_name: ${{ inputs.release_tag }}
+          github_token: ${{ github.token }}
+
+      - name: Download artifact
+        uses: actions/download-artifact@v3.0.2
+
+      - name: Publish ${{ inputs.release_tag }} release
+        uses: softprops/action-gh-release@v0.1.15
```
joelspadin, line 196: v1 appears to be the current version

.github/workflows/build-user-config.yml
```diff
@@ -161,3 +176,25 @@ jobs:
...
+    name: Publish firmware
+    steps:
+      - name: Delete previous release
+        uses: dev-drprasad/delete-tag-and-release@v1.0.1
+        with:
+          tag_name: ${{ inputs.release_tag }}
+          github_token: ${{ github.token }}
+
+      - name: Download artifact
+        uses: actions/download-artifact@v3.0.2
```
joelspadin, line 193: Any reason you're using this specific patch instead of just `@v3`?

joelspadin (changes requested):

.github/workflows/build-user-config.yml
filterpaper: The purpose of this addition is to simply publish _only_ the latest artifact as a release with a single tag. Users can continue to find older artifacts in the Actions tab.

It is _not_ meant to publish new tags/versions in the conventional manner. Deleting the previous release ensures that latest release time matches the artifact content upload time to avoid ambiguity.

> Rewriting Git history is not great, and probably not something we want to do by default (or even support at all). 

If these workflow addition are not desired from a maintainer point of view, PR will be withdrawn.

joelspadin: I think having a way to automatically generate releases could be useful, but if we added that, I would prefer that we use GitHub releases as they're intended to be used.

filterpaper: > I think having a way to automatically generate releases could be useful, but if we added that, I would prefer that we use GitHub releases as they're intended to be used.

Thanks for the feedback.

PR #2070, opened 2023-12-13
fix(ci): Upgrade to github-script@v7

petejohanson:
Latest test run had odd github-script related errors, so trying upgrading to the latest version.


petejohanson (commented):
Ok, can confirm the build now succeeds, and artifacts are properly uploaded as expected with the tweaked API usage.

PR #2072, opened 2023-12-14
Make settings_reset builds reset all settings

joelspadin:
- Added `zmk_settings_erase()`, which directly clears the backend data for settings.
- Changed the `settings_reset` shield to run this function on startup instead of using `CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START`, so it now resets all settings instead of just Bluetooth ones.
- Removed `CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START`, as it is no longer needed.

Future PRs can add additional ways to trigger `zmk_settings_erase()`.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -565,6 +565,9 @@ endmenu
 
 if SETTINGS
 
+config ZMK_SETTINGS_RESET_ON_START
+    bool "Delete all persistent settings when the keyboard boots"
+
 config ZMK_SETTINGS_SAVE_DEBOUNCE
     int "Milliseconds to debounce settings saves"
     default 60000
--- a/app/boards/shields/settings_reset/settings_reset.conf
+++ b/app/boards/shields/settings_reset/settings_reset.conf
@@ -1,1 +1,4 @@
-CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START=y
+CONFIG_SETTINGS=y
+CONFIG_ZMK_SETTINGS_RESET_ON_START=y
+# Disable BLE so splits don't try to re-pair until normal firmware is flashed.
+CONFIG_ZMK_BLE=n
--- a/app/src/settings/reset_settings_on_start.c
+++ b/app/src/settings/reset_settings_on_start.c
@@ -4,16 +4,10 @@
  * SPDX-License-Identifier: MIT
  */
 
-#include <zephyr/device.h>
 #include <zephyr/init.h>
 
 #include <zmk/settings.h>
 
-static int reset_settings_init(const struct device *dev) {
-    ARG_UNUSED(dev);
-    return zmk_settings_erase();
-}
-
 // Reset after the kernel is initialized but before any application code to
 // ensure settings are cleared before anything tries to use them.
-SYS_INIT(reset_settings_init, POST_KERNEL, CONFIG_APPLICATION_INIT_PRIORITY);
+SYS_INIT(zmk_settings_erase, POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT);
--- a/app/src/settings/reset_settings_on_start.c
+++ b/app/src/settings/reset_settings_on_start.c
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2023 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/init.h>
+
+#include <zmk/settings.h>
+
+static int reset_settings_init(const struct device *dev) {
+    ARG_UNUSED(dev);
+    return zmk_settings_erase();
+}
+
+// Reset after the kernel is initialized but before any application code to
+// ensure settings are cleared before anything tries to use them.
+SYS_INIT(reset_settings_init, POST_KERNEL, CONFIG_APPLICATION_INIT_PRIORITY);
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -99,5 +99,6 @@ target_sources_ifdef(CONFIG_ZMK_LOW_PRIORITY_WORK_QUEUE app PRIVATE src/workqueu
 target_sources(app PRIVATE src/main.c)
 
 add_subdirectory(src/display/)
+add_subdirectory_ifdef(CONFIG_SETTINGS src/settings/)
 
 zephyr_cc_option(-Wfatal-errors)
--- a/app/include/zmk/settings.h
+++ b/app/include/zmk/settings.h
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) 2023 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#pragma once
+
+/**
+ * Erases all saved settings.
+ *
+ * @note This does not automatically update any code using Zephyr's settings
+ * subsystem. This should typically be followed by a call to sys_reboot().
+ */
+int zmk_settings_erase(void);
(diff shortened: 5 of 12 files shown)


petejohanson (commented):
A couple high level questions after a first pass at this.

app/Kconfig
```diff
@@ -194,10 +194,6 @@ config ZMK_BLE_MOUSE_REPORT_QUEUE_SIZE
     int "Max number of mouse HID reports to queue for sending over BLE"
     default 20
 
-config ZMK_BLE_CLEAR_BONDS_ON_START
```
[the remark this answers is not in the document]
joelspadin: Do you think it needs to retain the behavior of only clearing BLE bonds, or could we make it just select the other KConfig option to clear settings on start that you suggested in your other comment?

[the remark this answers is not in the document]
petejohanson: I am a bit on the fence. Other areas we use settings subsystem right now are things like RGB selection, ext power on/off, etc. In the future we also plan to use it for runtime keymap updates.


So... I really think there may be value in both to be supported long term. I get that CircuitPython can muck with the flash region, and having a blunt force "clear all settings" is useful, but I'm not convinced it negates the need for the existing one, as I consider it more.

Thoughts?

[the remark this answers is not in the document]
joelspadin: I was thinking that having a special firmware to clear just bluetooth settings would be unnecessary with something like #2037. The only case that wouldn't cover is fixing pairing of splits, but do we really need to support a special firmware build for that which is separate from the full settings reset?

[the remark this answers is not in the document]
petejohanson: > I was thinking that having a special firmware to clear just bluetooth settings would be unnecessary with something like #2037. The only case that wouldn't cover is fixing pairing of splits, but do we really need to support a special firmware build for that which is separate from the full settings reset?

Repairing splits is *the reason* we originally created the settings reset in the first place. And I do think long term we should have a way to continue to do so that doesn't *also* force a user to lose their runtime keymap they've possible spent a long time updating/creating.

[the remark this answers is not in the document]
joelspadin: I put this back in for now. I'm thinking it should probably still be removed, but only after it's been replaced by a more convenient way to fix pairing for splits--ideally one that doesn't require flashing special firmware.

app/boards/shields/settings_reset/settings_reset.c
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/init.h>
+
+#include <zmk/settings.h>
+
+static int reset_settings_init(const struct device *dev) {
```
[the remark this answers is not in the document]
joelspadin: Done. (I originally didn't do this because you could just add this shield to any build to get its special behavior, but I don't really have a strong opinion either way, and having the new config be able to `select SETTINGS` does seem a bit cleaner.)

joelspadin:
Added the new KConfig option to the docs.

app/Kconfig
```diff
@@ -560,6 +560,11 @@ endif
 #Logging
 endmenu
 
+config ZMK_SETTINGS_RESET_ON_START
+    bool "Delete all persistent settings when the keyboard boots"
+    default n
+    select SETTINGS
```
petejohanson, line 566: I wonder if instead this should be:

```suggestion
    depends on SETTINGS
```

I don't think it makes much sense to turn this on unless you've already got other settings turned on elsewhere?

app/boards/shields/settings_reset/settings_reset.conf
```diff
@@ -1 +1,3 @@
-CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START=y
+CONFIG_ZMK_SETTINGS_RESET_ON_START=y
```
petejohanson, line 1: So, this would change this shield a bit from what it used to do... I'm not sure that's a bad thing, but probably warrants a blog post or at least announcement on Discord when merged, so folks are aware of the updated functionality.

petejohanson (commented):
A few thoughts.

app/boards/shields/settings_reset/settings_reset.conf
petejohanson: So... the other areas that would be reset by this:

* Output selection
* External power on/off
* Backlight on/off
* RGB on/off and selected effect

Of those... the output selection and external power are maybe the *potentially* most troublesome, but I don't think some docs/info wouldn't help address.

Let's move forward with this.

app/boards/shields/settings_reset/settings_reset.conf
```diff
@@ -1 +1,4 @@
-CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START=y
+CONFIG_SETTINGS=y
+CONFIG_ZMK_SETTINGS_RESET_ON_START=y
+# Disable BLE so splits don't try to re-pair until normal firmware is flashed.
+CONFIG_ZMK_BLE=n
```
petejohanson, line 4: I do wonder if this might lead to folks not being clear if the updated firmware flashed or not... Maybe a docs update to note that when you flash settings reset, the device won't be visible/connectable at all again until flashed with regular firmware?

petejohanson (commented):
Just one comment, but this all seems reasonable. Thoughts on tweaks to the settings reset docs to make it clear what this file now does?

app/boards/shields/settings_reset/settings_reset.conf
joelspadin: Yeah, that seems like it would be good to clarify. Will do.

joelspadin: Done. How does that look?

petejohanson: LGTM. Has this gotten thorough testing locally there?

joelspadin: I've tested it with the NVS backend. The file backend is untested. I'm not sure what currently supported build (if any) could be used for that.

app/Kconfig
```diff
@@ -561,6 +561,10 @@ endmenu
 
 if SETTINGS
 
+config ZMK_SETTINGS_RESET_ON_START
+    bool "Delete all persistent settings when the keyboard boots"
+    default n
```
petejohanson, line 566: As we fixed up recently, this line is unnecessary, and even can cause issues (see https://docs.zephyrproject.org/latest/build/kconfig/tips.html#redundant-defaults):

```suggestion
```

app/src/settings/reset_settings_on_start.c
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/init.h>
+
+#include <zmk/settings.h>
+
+static int reset_settings_init(const struct device *dev) {
```
petejohanson, line 12: This changed in Zephyr 3.5:

```suggestion
static int reset_settings_init(void) {
```

Probably should rebase to be sure this is 100% happy.

app/src/settings/reset_settings_on_start.c
```diff
@@ -0,0 +1,19 @@
...
+    ARG_UNUSED(dev);
+    return zmk_settings_erase();
+}
+
+// Reset after the kernel is initialized but before any application code to
+// ensure settings are cleared before anything tries to use them.
+SYS_INIT(reset_settings_init, POST_KERNEL, CONFIG_APPLICATION_INIT_PRIORITY);
```
petejohanson, line 19: Instead of application init priority, probably should be:

```suggestion
SYS_INIT(reset_settings_init, POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT);
```

petejohanson (changes requested):
Going over this with a fine-toothed comb one last time, found a couple lingering items.

joelspadin:
> Going over this with a fine-toothed comb one last time, found a couple lingering items.

Rebased and updated everything.

PR #2077, opened 2023-12-16
refactor(core): Make low priority queue optional.

petejohanson:
* Add a new Kconfig symbol to enable the low priority queue, and make the two features that depend on it `select` the symbol to turn it on. This helps ensure no wasted RAM/ROM on devices that don't need it.

This came up while working on more resource constrained devices like SAMD21 and stm32f072.


petejohanson:
Tagging @xudongzheng who did the original low priority queue work.

joelspadin (approved):

PR #2081, opened 2023-12-21
Disconnect old BLE profile when selecting new one

jridgewell:
This disconnects the current bluetooth profile when switching to a new one. I noticed that after switching profiles, my old computer would occasionally wake from sleep. This seems to be caused by battery notifications being sent to every connected profile.

I think this might help a bit with battery life? I imagine maintaining active connections to multiple computers must be using some energy.


joelspadin:
ZMK intentionally stays connected to each profile to allow for immediate switching between them. An _option_ to only be connected to the active profile might be useful, but I don't think we want to change the default behavior here.

If we were to add that as an option, I expect we'd want it to not connect to the other profiles to begin with as well. I can't say I'm too well versed in the BLE code, so I could be wrong about this, but I believe with just this change it will disconnect from a profile when switching away from it, but when the keyboard boots it will still connect to every profile right away.

jridgewell:
> An option to only be connected to the active profile might be useful, but I don't think we want to change the default behavior here.

I'd be happy with a compile time config value that disabled fast-switching, if that's what you're suggesting.

> If we were to add that as an option, I expect we'd want it to not connect to the other profiles to begin with as well… when the keyboard boots it will still connect to every profile right away

You're correct, and I didn't check this. If we do a config value, I think we could skip the connect step during boot.

jridgewell:
Actually, this may be possible already by setting `CONFIG_BT_MAX_CONN=1`. That will hopefully fix my problem and reconnect on boot.

jridgewell:
> this may be possible already by setting `CONFIG_BT_MAX_CONN=1`

This doesn't work as intended. For a split keyboard, the peripheral never connects to the central, and it doesn't disconnect the old profile when activating the new one. I think this needs to be on ZMK's side.

docs/docs/config/bluetooth.md
```diff
@@ -16,3 +16,4 @@ See [Configuration Overview](index.md) for instructions on how to change these s
...
 | `CONFIG_ZMK_BLE_PASSKEY_ENTRY`         | bool | Enable passkey entry during pairing for enhanced security. (Note: After enabling this, you will need to re-pair all previously paired hosts.)                                                                                                                           | n       |
 | `CONFIG_BT_GATT_ENFORCE_SUBSCRIPTION`  | bool | Low level setting for GATT subscriptions. Set to `n` to work around an annoying Windows bug with battery notifications.                                                                                                                                                 | y       |
+| `CONFIG_ZMK_BLE_FAST_SWITCHING`        | bool | Enables faster switching between profiles by maintaining the bluetooth connection even after switching to a different profile. Disabling this will disconnect (but remain paired) from the old profile when selecting a new one.                                        | y       |
```
caksoylar, line 19:
```suggestion
| `CONFIG_ZMK_BLE_FAST_SWITCHING`        | bool | Enable faster switching between profiles by maintaining the bluetooth connection even after switching to a different profile. Disabling this will disconnect (but remain paired) from the previous profile when selecting a different one.                                        | y       |
```
Note that this table needs reformatting (with prettier, would be easiest, e.g. via [`npm run prettier:format`](https://zmk.dev/docs/development/documentation#formatting-and-linting-your-changes) or through the [pre-commit hooks](https://zmk.dev/docs/development/pre-commit).

jridgewell:
@joelspadin I've updated this to use a `CONFIG_ZMK_BLE_FAST_SWITCHING` config value (enabled by default).

There's a bit of complication with our bluetooth startup. We use an [open advertisement](https://github.com/zmkfirmware/zmk/blob/a171ff1f614d503bc93eb75957c70d510bad86e2/app/src/ble.c#L202) when the active_profile [hasn't yet connected](https://github.com/zmkfirmware/zmk/blob/a171ff1f614d503bc93eb75957c70d510bad86e2/app/src/ble.c#L173) due to a [bug with direct advertisements](https://github.com/zmkfirmware/zmk/blob/a171ff1f614d503bc93eb75957c70d510bad86e2/app/src/ble.c#L174-L175) and privacy bluetooth connections. So we'll end up connecting to all paired profiles within range when we startup (and we'll keep open advertising until the active connects). The only way to really fix this is to disconnect all non-active profiles after the active one connects. It's a bit hacky, and will still wake my non-active computers, but at least the connection won't be maintained.

docs/docs/config/bluetooth.md
jridgewell: Done.

petejohanson:
It would be nice to add a new BLE test to cover this scenario and verify the behavior against future refactors.

huber-th:
This will be a nice option to have. I'm looking forward to this being merged. I was looking for existing work before I start digging into the code myself as I was considering implementing this feature myself had it not already existed as a PR/assigned issue.

This will allow people to trade the fast switching feature for the convenience of ensuring the on-screen keyboard appears when switching away from devices like phones and tablets without actively having to trigger a disconnect via `BT_DISC #` all the time.

Nick-Munnich:
It may be an idea to be able to set this on a per-profile basis rather than globally. The issues seem to be more on the end of the target profile and won't necessarily apply to all, so it would make sense to tie it to profiles I think.

PR #2084, opened 2023-12-27
Add some python things to `.gitignore`

waffle87:
As per title. (Zephyr docs recommend installing Python packages in a virtual environment)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 /build
 *.DS_Store
 __pycache__
+.python-version
+.venv


petejohanson (approved):
Seems reasonable if you go that route. Thanks!

petejohanson:
@waffle87 Actually... Soon as I pushed the button I considered.. Would it make sense to pin a Python version in the `.python-version` and add to the repo, to ensure folks get a compatible one in their virtualenv? (I'm no Python expert, so correct me if this doesn't work like that, my experience is the equivalent from Ruby.)

waffle87:
As far as I'm aware, the `.python-version` file is (a bit annoyingly) not related to official Python at all. It's used by the [pyenv utility](https://github.com/pyenv/pyenv) to select which Python version a user would like within a certain project, not so much a minimum requirement. E.g. if `3.8` (the minimum Zephyr Python version) was to be put in the file, Python 3.8 would be attempted to be used, which most users probably don't actually have. Let me know what you think.

petejohanson:
> As far as I'm aware, the `.python-version` file is (a bit annoyingly) not related to official Python at all. It's used by the [pyenv utility](https://github.com/pyenv/pyenv) to select which Python version a user would like within a certain project, not so much a minimum requirement. E.g. if `3.8` (the minimum Zephyr Python version) was to be put in the file, Python 3.8 would be attempted to be used, which most users probably don't actually have. Let me know what you think.

Ok, fair enough. I'm good with this as is then. Thanks.

PR #2085, opened 2023-12-30
Soft Off (take 3)

petejohanson:
Ok, after some good discussion with @joelspadin around the cursed "scanned behavior trigger" we've decided to pursue another route. This PR replaces #1942 and removes the scanned trigger, as well as fixing the last few review items there. Hoping this will be the one that finally wins. A follow up PR will add "side band kscan behavior triggers" so that the "matrix integrated soft off" is supported sanely.


joelspadin (approved):

app/boards/shields/zmk_uno/zmk_uno_split.keymap
```diff
@@ -16,9 +16,7 @@
 
 /* :REMOVE ME
```
joelspadin, line 17: Not sure if this is common in other IDEs (no need to change this if not), but Visual Studio and VS Code have a shortcut to comment/uncomment a block of code. At least in DTS files, it only seems to work with `//` comments though.

e.g, if this were formatted like

```dts
// &kscan_direct { status = "okay"; };
// &kscan_matrix { status = "disabled"; };
```

then you could just select both lines and press <kbd>Ctrl+K Ctrl+U</kbd> to uncomment and <kbd>Ctrl+K Ctrl+C</kbd> to comment it out again.

app/dts/bindings/kscan/zmk,kscan-sideband-behaviors.yaml
```diff
@@ -0,0 +1,29 @@
...
+    type: phandle
+    required: true
+
+child-binding:
+  description: "A sideband behavior tied to a row/column pair"
+
+  properties:
+    row:
+      type: int
+      required: true
```
joelspadin, line 23: Could `row` be made optional (defaulting to 0) to simplify use with the direct GPIO kscan driver (and anything else that creates a flat list instead of a matrix)?

app/dts/bindings/kscan/zmk,kscan-sideband-behaviors.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2023, The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+description: |
+  kscan sideband behavior runner. Only basic system behavior should be used,
```
joelspadin, line 5: Typo?
```suggestion
  kscan sideband behavior runner. Only basic system behaviors should be used,
```

Also, I'm not sure how better to describe this, but I think it may be unclear what constitutes "basic system behaviors".

app/module/drivers/kscan/kscan_gpio_direct.c
```diff
@@ -331,7 +352,9 @@ static const struct kscan_driver_api kscan_direct_api = {
                  "ZMK_KSCAN_DEBOUNCE_RELEASE_MS or debounce-release-ms is too large");             \
                                                                                                    \
     static struct kscan_gpio kscan_direct_inputs_##n[] = {                                         \
-        LISTIFY(INST_INPUTS_LEN(n), KSCAN_DIRECT_INPUT_CFG_INIT, (, ), n)};                        \
+        COND_CODE_1(DT_INST_NODE_HAS_PROP(n, input_gpios),                                         \
+                    (LISTIFY(INST_INPUTS_LEN(n), KSCAN_DIRECT_INPUT_CFG_INIT, (, ), n)),           \
```
joelspadin, line 356: Renaming this to better differentiate it from the input-keys version could make this easier to understand.
```suggestion
                    (LISTIFY(INST_INPUTS_LEN(n), KSCAN_DIRECT_GPIO_INPUT_CFG_INIT, (, ), n)),           \
```

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,167 @@
...
+#include <zmk/event_manager.h>
+#include <zmk/behavior.h>
+#include <zmk/keymap.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct ksbb_entry {
+    uint8_t row;
+    uint8_t column;
+    struct zmk_behavior_binding binding;
```
joelspadin, line 24: Putting this first might result in a smaller struct.

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,167 @@
...
+struct ksbb_data {
+    kscan_callback_t callback;
+    bool enabled;
+};
+
+#define GET_KSBB_DEV(n) DEVICE_DT_GET(DT_DRV_INST(n)),
+
+// The kscan callback has no context with it, so we keep a static array of all possible
+// KSBBs to check when a kscan callback from the "wrapped" inner kscan fires.
+static const struct device *ksbbs[DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT)] = {
```
joelspadin, line 42: Specifying the array size shouldn't be necessary?
```suggestion
static const struct device *ksbbs[] = {
```

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,167 @@
...
+            data->callback(ksbb, row, column, pressed);
+        }
+    }
+}
+
+static int ksbb_configure(const struct device *dev, kscan_callback_t callback) {
+    const struct ksbb_config *cfg = dev->config;
+    struct ksbb_data *data = dev->data;
+
+    if (!callback) {
```
joelspadin, line 81: Based on the check on line 71, callback being NULL appears to be a valid state, so I assume it is okay if you don't set a callback to begin with. Is it not allowed to set the callback back to NULL after previously setting it to something?

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,167 @@
...
+
+static int ksbb_init(const struct device *dev) {
+    const struct ksbb_config *config = dev->config;
+
+    if (!device_is_ready(config->kscan)) {
+        LOG_ERR("kscan %s is not ready", config->kscan->name);
+        return -ENODEV;
+    }
+
+    kscan_config(config->kscan, &ksbb_inner_kscan_callback);
```
joelspadin, line 118: Is it okay to call a kscan device's config function before it is initialized? If not, is there anything here that ensures that the kscan we're attaching to is initialized first? (I'm not seeing anything, but I don't know if Zephyr is automatically determining device's dependencies or anything.)

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,167 @@
...
+#define GET_KSBB_DEV(n) DEVICE_DT_GET(DT_DRV_INST(n)),
+
+// The kscan callback has no context with it, so we keep a static array of all possible
+// KSBBs to check when a kscan callback from the "wrapped" inner kscan fires.
+static const struct device *ksbbs[DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT)] = {
+    DT_INST_FOREACH_STATUS_OKAY(GET_KSBB_DEV)};
+
+void ksbb_inner_kscan_callback(const struct device *dev, uint32_t row, uint32_t column,
+                               bool pressed) {
+    for (int i = 0; i < ARRAY_SIZE(ksbbs); i++) {
```
joelspadin, line 47: If you pulled these loops out into separate functions, I think it would be easier to read the code, e.g.

```c
const struct device* ksbb = find_sideband_behaviors(dev);
if (!ksbb) {
  LOG_ERR("Got an event from kscan driver \"%s\" which has no sideband behaviors", dev->name);
  return;
}

... cfg = ...
... data = ...
const struct ksbb_entry entry = find_entry(cfg, row, column);
if (entry) {
  ... event = ...
  if (pressed) { ... } else { ... }  
  return;
}

// No sideband behavior for this position. Pass through.
if (data->enabled && data->callback) { ... }
```

app/src/kscan_sideband_behaviors.c
joelspadin: If this is a problem, then maybe you could call `kscan_config()` from `ksbb_configure()` instead of `ksbb_init()`?

app/dts/bindings/kscan/zmk,kscan-sideband-behaviors.yaml
petejohanson: I might be able to expand on this, but mostly it's things like tap holds, sticky keys, and any behavior that has non-trivial state stored as part of their use, that can't properly be used with sideband behaviors.

app/src/kscan_sideband_behaviors.c
petejohanson: > Is it okay to call a kscan device's config function before it is initialized? If not, is there anything here that ensures that the kscan we're attaching to is initialized first? (I'm not seeing anything, but I don't know if Zephyr is automatically determining device's dependencies or anything.)

The call to `device_is_ready` just above this will ensure that we're making the call into an already initialized device. Under the hood that calls `z_device_is_ready` which does:

```
bool z_device_is_ready(const struct device *dev)
{
	/*
	 * if an invalid device pointer is passed as argument, this call
	 * reports the `device` as not ready for usage.
	 */
	if (dev == NULL) {
		return false;
	}

	return dev->state->initialized && (dev->state->init_res == 0U);
}
```

we do this in the *init* because we want to catch events even if we're not configured ourselves, such as when wrapping a direct GPIO for a dedicated pin that's not used in the matrix at all.

petejohanson: Oh, and right now this works because this driver is in the `APPLICATION` init level, but once we move to Zephyr 3.5 we'll probably instead need to have a custom init priority lower than the `CONFIG_KSCAN_INIT_PRIORITY`.

app/boards/shields/zmk_uno/zmk_uno.overlay
```diff
@@ -19,4 +23,40 @@
...
+            tap-ms = <1>;
+            bindings = <&out OUT_BLE &bt BT_SEL 1>;
+        )
+    };
+
+    endpoint_sideband_behaviors {
+        compatible = "zmk,kscan-sideband-behaviors";
+        kscan = <&kscan_sp3t_toggle>;
+
+        usb {
```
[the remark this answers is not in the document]
petejohanson: Can do. Will probably name something generic like `first_toggle_sideband` or somesuch.

app/dts/bindings/kscan/zmk,kscan-sideband-behaviors.yaml
petejohanson: Tweaked the comment a bit. Thoughts?

app/src/kscan_sideband_behaviors.c
petejohanson: Yeah, this isn't really necessary. Removed.

app/src/kscan_sideband_behaviors.c
petejohanson: Added a new dedicated init priority for this instead of using the `APPLICATION` init level, so we don't have to fix this later.

app/src/kscan_sideband_behaviors.c
petejohanson: Done.

app/dts/bindings/kscan/zmk,kscan-sideband-behaviors.yaml
petejohanson: Done. Good call.

app/boards/shields/zmk_uno/zmk_uno_split.keymap
petejohanson: Yeah, updated and tweaked the comment above for clarity.

app/dts/bindings/kscan/zmk,kscan-sideband-behaviors.yaml
joelspadin: Seems reasonable to me. Thanks.

app/Kconfig
```diff
@@ -481,6 +494,21 @@ config ZMK_KSCAN_EVENT_QUEUE_SIZE
...
+    bool
+    default y
+    depends on DT_HAS_ZMK_KSCAN_SIDEBAND_BEHAVIORS_ENABLED
+    select KSCAN
+
+if ZMK_KSCAN_SIDEBAND_BEHAVIORS
+
+config ZMK_KSCAN_SIDEBAND_BEHAVIORS_INIT_PRIORITY
+    int "Keyboard scan sideband behaviors driver init priority"
+    # The default kscan init priority is 90, so be sure we are lower.
```
joelspadin, line 507: "be sure we are lower" paired with a value that is greater than 90 might be a bit confusing. Maybe say "initialized later" instead?

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,184 @@
...
+
+    return -ENODEV;
+}
+
+void ksbb_inner_kscan_callback(const struct device *dev, uint32_t row, uint32_t column,
+                               bool pressed) {
+    struct ksbb_entry *entry = NULL;
+    const struct device *ksbb = NULL;
+
+    if (find_ksbb_for_inner(dev, &ksbb) >= 0) {
```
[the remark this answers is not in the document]
petejohanson: Valid. Simplified.

app/Kconfig
petejohanson: Tweaked for clarity.

app/Kconfig
```diff
@@ -1,5 +1,5 @@
-# Copyright (c) 2020 The ZMK Contributors
-# SPDX-License-Identifier: MIT
+#Copyright(c) 2020 The ZMK Contributors
```
joelspadin, line 1: Were all the whitespace changes here intentional?

app/Kconfig
```diff
@@ -481,6 +494,21 @@ config ZMK_KSCAN_EVENT_QUEUE_SIZE
...
+    bool
+    default y
+    depends on DT_HAS_ZMK_KSCAN_SIDEBAND_BEHAVIORS_ENABLED
+    select KSCAN
+
+if ZMK_KSCAN_SIDEBAND_BEHAVIORS
+
+config ZMK_KSCAN_SIDEBAND_BEHAVIORS_INIT_PRIORITY
+    int "Keyboard scan sideband behaviors driver init priority"
+    #The default kscan init priority is 90, so be sure we are initialezed after.
```
joelspadin, line 507:
```suggestion
    # The default kscan init priority is 90, so be sure we are initialized after.
```

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,181 @@
...
+            return candidate;
+        }
+    }
+
+    return NULL;
+}
+
+void ksbb_inner_kscan_callback(const struct device *dev, uint32_t row, uint32_t column,
+                               bool pressed) {
+    struct ksbb_entry *entry = NULL;
```
joelspadin, line 73: Can this be const too?

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,181 @@
...
+    const struct device *ksbb = NULL;
+
+    if ((ksbb = find_ksbb_for_inner(dev)) != NULL) {
```
joelspadin, line 76: Optional: Not sure if there's a specific reason you wrote it this way, but these may be a bit easier to read like
```c
const struct device* ksbb = find_ksbb_for_inner(dev);
if (ksbb) {
  ...
}
```

app/Kconfig
petejohanson: Nope! Reverted.

app/src/kscan_sideband_behaviors.c
petejohanson: Sadly no, that leads the inner fields to be considered `const`, which fails when calling `behavior_keymap_binding_pressed`

app/src/kscan_sideband_behaviors.c
petejohanson: Totally fair. Refactored.

app/src/kscan_sideband_behaviors.c
```diff
@@ -0,0 +1,181 @@
...
+        .binding = ZMK_KEYMAP_EXTRACT_BINDING(0, e),                                               \
+    }
+
+#define KSBB_INST(n)                                                                               \
+    static struct ksbb_entry entries_##n[] = {                                                     \
+        DT_INST_FOREACH_CHILD_STATUS_OKAY_SEP(n, ENTRY, (, ))};                                    \
+    const struct ksbb_config ksbb_config_##n = {                                                   \
+        .kscan = DEVICE_DT_GET(DT_INST_PHANDLE(n, kscan)),                                         \
+        .entries = entries_##n,                                                                    \
+        .entries_len = DT_INST_FOREACH_CHILD_STATUS_OKAY_SEP(n, JUST_ONE, (+)),                    \
```
joelspadin, line 173: I think this would work?
```suggestion
        .entries_len = ARRAY_SIZE(entries_##n),                    \
```

joelspadin (approved):

app/src/kscan_sideband_behaviors.c
petejohanson: Indeed. Simplified and pushed.

petejohanson:
@caksoylar Ok, I've pushed updated docs to cover the new sideband behavior kscan driver and how to use it with simple direct GPIO and for matrix integrated hardware combos. Would appreciate your feedback when you have a sec.

docs/docs/behaviors/soft-off.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Soft Off Behavior
+sidebar_label: Soft Off
+---
+
+## Summary
+
+The soft off behavior is used to force the keyboard into an off state. Depending on the specific keyboard hardware, the keyboard can be turned back on again either with a dedicated on/off button that is available, or using the reset button found on the device.
+
+For more information, see the [Soft Off Feature](../features/soft-off.md) page.
```
caksoylar, line 10: I think we might want to note somewhere that `CONFIG_ZMK_PM_SOFT_OFF=y` is required for this behavior to work.

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,255 @@
...
+
+#### GPIO Key
+
+Zephyr's basic GPIO Key concept is used to configure the GPIO pin that will be used for both triggering soft off and waking the device later. Here is an example for a keyboard with a dedicated on/off push button that is a direct wire between the GPIO pin and ground:
+
+```
+/ {
+    keys {
+        compatible = "gpio-keys";
+        wakeup_key: wakeup_key {
```
caksoylar, line 71: Maybe rename this to `softoff_key` so it is clear it is used both for activating soft-off and waking up.

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,255 @@
...
+The kscan sideband behavior driver will be used to trigger the [soft off behavior](../behaviors/soft-off.md) "out of band" from the normal keymap processing. To do so, it will decorate/wrap an underlying kscan driver. What kscan driver will vary for simple direct pin vs. matrix-integrated hardware combo.
+
+##### Simple Direct Pin
+
+With a simple direct pin setup, the The [direct kscan](../config/kscan.md) driver can be used with a GPIO key, to make a small "side matrix":
+
+```
+    soft_off_direct_scan: soft_off_direct_scan {
+        compatible = "zmk,kscan-gpio-direct";
+        input-keys = <&wakeup_key>;
```
caksoylar, line 147: Can this also not be simplified to get rid of `input-keys` (and `gpio-keys` usage altogether) by asking the user to augment `input-gpios` instead? That seems to be the approach already for matrix input.

caksoylar: If we are using `input-keys`, it needs to be documented in the direct kscan config page.

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,255 @@
...
+        };
+    };
+};
+```
+
+GPIO keys are defined using child nodes under the `gpio-keys` compatible node. Each child needs just one property defined:
+
+- The `gpios` property should be a phandle-array with a fully defined GPIO pin and with the correct pull up/down and active high/low flags set. In the above example the soft on/off would be triggered by pulling the specified pin low, typically by pressing a switch that has the other leg connected to ground.
+
+#### Soft Off Waker
```
caksoylar, line 82: I am having some trouble understanding: Is this still necessary if you follow the "Simple Direct Pin" section under sideband?

docs/docs/features/soft-off.md
caksoylar, line 1: In general I think the structure of this page under "Firmware Changes" is hard to follow for me. The deep header nesting especially causes me issues since it is hard to distinguish header levels at that point. Perhaps we can remove `adding-dedicated-soft-onoff-gpio-pin-to-new-designs`, make both hardware and firmware sections h2 headers, so that we can reduce header levels by one and have more visible differences under firmware subsections.

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,255 @@
...
+};
+```
+
+Here are the properties for the node:
+
+- The `compatible` property for the node must be `zmk,gpio-key-wakeup-trigger`.
+- The `trigger` property is a phandle to the GPIO key defined earlier.
+- The `wakeup-source` property signals to Zephyr this device should not be suspended during the shutdown procedure.
+- An optional `output-gpios` property contains a list of GPIO pins (including the appropriate flags) to set active before going into power off, if needed to ensure the GPIO pin will trigger properly to wake the keyboard. This is only needed for matrix integrated combos. For those keyboards, the list should include the matrix output needs needed so the combo hardware is properly "driven" when the keyboard is off.
+
```
elpekenin, line 103: Unless im missing something (completely possible as im as much of a ZMK noob as someone can be), there is a name mismatch here:

```suggestion
- An optional `extra-gpios` property contains a list of GPIO pins (including the appropriate flags) to set active before going into power off, if needed to ensure the GPIO pin will trigger properly to wake the keyboard. This is only needed for matrix integrated combos. For those keyboards, the list should include the matrix output needs needed so the combo hardware is properly "driven" when the keyboard is off.
```

docs/docs/features/soft-off.md
petejohanson: > Can this also not be simplified to get rid of `input-keys` (and `gpio-keys` usage altogether) by asking the user to augment `input-gpios` instead? That seems to be the approach already for matrix input.

The advantage is avoiding duplicating the GPIO details if we can avoid it, since we use the GPIO key for the waker elsewhere too. Unfortunately, we can't easily "mix" these in a matrix definition.

docs/docs/features/soft-off.md
petejohanson: Very good suggestion! I just pushed this tweak, and it does help just that extra little bit. Thanks!

docs/docs/features/soft-off.md
petejohanson: So.. good question. *Yes*, but let me think if we could avoid it and just depend on the direct kscan itself to do the waking in that case.

petejohanson: Ok, so did a little testing, and no, it isn't actually required in that case, you can directly use the small dedicated direct kscan as the soft off waker.... Let me think about how best to capture this in the docs.

petejohanson: @caksoylar Ok, pushed docs tweaks and changed the ZMK Uno shield to reflect the updated recommended approach that adds the extra waker *only* when using the matrix integrated approach.

docs/docs/features/soft-off.md
petejohanson: @caksoylar Thoughts on the new structure?

docs/docs/features/soft-off.md
caksoylar: I think keeping it conceptually simple would have been an advantage but if you prefer it this way then I am OK with that.

docs/docs/behaviors/soft-off.md
caksoylar: Pinging @petejohanson on this, in case you missed it.

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,270 @@
...
+### Soft Off Waker
+
+Next, we need to add another device which will be enabled only when the keyboard is going into soft off state, and will configure the previously declared GPIO key with the correct interrupt configuration to wake the device from soft off once it is pressed.
+
+```
+/ {
+    wakeup_source: wakeup_source {
+        compatible = "zmk,gpio-key-wakeup-trigger";
+
+        trigger = <&wakeup_key>;
```
caksoylar, line 243:
```suggestion
        trigger = <&soft_off_key>;
```
Following the previous comment

docs/docs/behaviors/soft-off.md
petejohanson: Add a link to the actual config section where it's documented, rather than duplicating it here. Thoughts?

docs/docs/features/soft-off.md
petejohanson: I decided to go with `on_off_key` for this, which felt even more accurate. Thoughts?

docs/docs/features/soft-off.md
petejohanson: I think we stick with it. It gives us some flexibility, especially for designs that already have the GPIO keys defined for the upstream board definition, etc.

Added a note to the kscan config about this.

Thoughts?

docs/docs/features/soft-off.md
```diff
@@ -0,0 +1,270 @@
...
+The kscan sideband behavior driver will be used to trigger the [soft off behavior](../behaviors/soft-off.md) "out of band" from the normal keymap processing. To do so, it will decorate/wrap an underlying kscan driver. What kscan driver will vary for simple direct pin vs. matrix-integrated hardware combo.
+
+#### Simple Direct Pin
+
+With a simple direct pin setup, the The [direct kscan](../config/kscan.md) driver can be used with a GPIO key, to make a small "side matrix":
+
+```
+    soft_off_direct_scan: soft_off_direct_scan {
+        compatible = "zmk,kscan-gpio-direct";
+        input-keys = <&wakeup_key>;
```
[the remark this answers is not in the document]
petejohanson: Named `on_off_key` now. Good?

docs/docs/features/soft-off.md
petejohanson: Dittio

docs/docs/behaviors/soft-off.md
caksoylar: Sure, it seemed simple to note to me because it needed a single flag but this is fine too.

docs/docs/features/soft-off.md
caksoylar: Good call to note it there, sounds good to me.

docs/docs/features/soft-off.md
caksoylar: Makes sense, a bit clearer.

caksoylar (approved):

PR #224, opened 2020-09-30
Cradio

davidphilipbarr:
34 key keyboard with no diodes.

PR #2087, opened 2023-12-31
refactor: remove unused files; improve build UX

lesshonor:
The `Kconfig` files were mistakenly placed under `boards/shields` as part of #224.

Willing to break out the makefile change into a separate PR RFC-style, but IMO the "Did not find `${shield}`" message here is actively unhelpful:
  - Users mistakenly latch onto it as a reason for build failures because it is the first "error" they see written in plain English, and separated from the rest of the build output by a blank line
  - It displays for all out-of-tree shields
  - It displays for *in-tree* shields if they are splits, or built with an out-of-tree module/config repo (e.g. `my_shield nice_view_adapter nice_view`)
  - The build will fail within seconds of the warning if any shield specified is actually missing
  - Per #1989, it should not be `WARN`-worthy that a shield does not exist in the main repo when (emphasis mine):
      > [An independently managed Zephyr module repository] ***is the preferred way to handle boards and shields*** moving forward in ZMK [...] ZMK does have a collection of boards/shields in the ZMK main repository, ***which are planned to be phased out*** [...]


caksoylar:
I like these changes. Kconfig files do seem look the result of a mistake to me, so makes sense to remove them.

I also agree that that warning has caused a lot of confusion. I think it would be great to remove that (thanks for digging up the cause!).

petejohanson (approved):
Thanks. Agreed on all points.

PR #2088, opened 2023-12-31
Use `zmk_battery_state_of_charge` in battery widgets

jridgewell:
A small change that switches from `bt_bas_get_battery_level` to `zmk_battery_state_of_charge`.

The [Battery Level](https://zmk.dev/docs/config/battery#:~:text=BLE%20REPORTING%20ON%20MACOS) docs suggest that you can disable `CONFIG_BT_BAS` to stop the keyboard from constantly waking MacOS with battery notifications while still enabling `CONFIG_ZMK_BATTERY_REPORTING` monitoring for OLED widgets. But [this combination would break](https://github.com/jridgewell/zmk-config/actions/runs/7345827356/job/19999587898), because the widgets were calling a function only available when `CONFIG_BT_BAS` is enabled.


app/src/display/widgets/battery_status.c
```diff
@@ -64,7 +64,7 @@ void battery_status_update_cb(struct battery_status_state state) {
 
 static struct battery_status_state battery_status_get_state(const zmk_event_t *eh) {
     return (struct battery_status_state) {
-        .level = bt_bas_get_battery_level(),
+        .level = zmk_battery_state_of_charge(),
```
caksoylar, line 67: It looks like this change in particular was included in https://github.com/zmkfirmware/zmk/commit/0e2f94b73bc18e3be7d913d2a233ae6330b40624#diff-9f9737e080412a2f82ae1867e062e8d581ca4211af5b2a56e64b980668f322d1, which causes a merge conflict. Other instances covered in this PR are not changed, so this is still a helpful PR IMO.

jridgewell: Merged, keeping their `ev->state_of_charge`.

petejohanson (approved):
Indeed. Good catches on the rest of these! Thanks.

PR #1995, opened 2023-11-10
Zephyr 3.5 update

petejohanson:
Even more experimental than #1853 as this one has even less testing, but opening this for posterity. Still needs me to publish newer Docker images for CI to be happy, as this needs a newer SDK version.

Blog Post preview: https://deploy-preview-1995--zmk.netlify.app/blog/2024/01/31/zephyr-3-5

# Testing steps

## Local dev

1. If not building w/ VSCode docker integration, install https://github.com/zephyrproject-rtos/sdk-ng/releases/tag/v0.16.4
1. `gh pr checkout 1995`
2. `west update`
3. Use new build directory to build ZMK as usual.

## GitHub Actions

Make two edits in the branch of your config repo:

1. ~~Update `.github/workflowr/build.yaml` to contain:~~ No longer needed, the `stable` Docker image has been bumped so this step can be skipped now.

    ```
    on: [push, pull_request, workflow_dispatch]

    jobs:
      build:
        uses: petejohanson/zmk/.github/workflows/build-user-config.yml@core/zephyr-3.5-update
    ```
 1. Update `config/west.yaml` according to the https://zmk.dev/docs/features/beta-testing/ instructions to test this PR.

## To Do

- [x] https://github.com/zmkfirmware/zmk/pull/1853#issuecomment-1789915446
- [x] Blog post with upgrade details
- [x] Get a the retained bootmode change working.
- [x] Tag the proper SDK zmk-docker images and move to using them.
- [x] Lots of testing

PR #2091, opened 2024-01-02
Fix high idle/sleep current due to flash and use xiao-ble from upstream Zephyr

ldab:
I'm working a XIAO-BLE based keyboard and I noticed high idle and sleep currents, bearing in mind that my board:

- Uses matrix polling;
- has CA95xx gpio expander;

image

The 500uA current is similar to reported here: https://forum.seeedstudio.com/t/low-power-with-xiao-nrf52840-on-zephyr-rtos/270491/6

Which led me to https://github.com/zmkfirmware/zmk/pull/1927. Even tho the PR seems to address the issue somehow, I find it strange that it does since the nRF driver puts the flash on Deep Power-Down (DPD) [drivers/flash/nrf_qspi_nor.c#L1324](https://github.com/zmkfirmware/zephyr/blob/1ae0eb5ce8adafcec993e6fb8f4eeb6f818a7772/drivers/flash/nrf_qspi_nor.c#L1324), the reason for DPD not properly working I believe to be related to the fact that the ZMK fork of Zephyr use what seems the incorrect [boards/arm/seeeduino_xiao_ble/seeeduino_xiao_ble.dts#L97](https://github.com/zmkfirmware/zephyr/blob/1ae0eb5ce8adafcec993e6fb8f4eeb6f818a7772/boards/arm/seeeduino_xiao_ble/seeeduino_xiao_ble.dts#L97) board configuration (which does not match the upstream Zephyr).

I noticed the `seeeduino_xiao_ble` was added on https://github.com/zmkfirmware/zephyr/commit/1f19b89c79fe4c451ba5cd9ff42deb3fd5c3be3f and I do not know the reason for adding since it is already available on [boards/arm/xiao_ble](https://github.com/zmkfirmware/zephyr/tree/v3.2.0%2Bzmk-fixes/boards/arm/xiao_ble) and also on Zephyr 3.2 and newer.

With that being said, this PR:
1. ~~Remove `seeeduino_xiao_ble` and use the `xiao_ble` from Zephyr~~
2. Put the Flash to Deep Power Down mode and reduce the idle/sleep current. ~~(To add measurement)~~

Note/Question:
1. Another PR can be made on `zmkfirmware/zephyr` to revert https://github.com/zmkfirmware/zephyr/commit/1f19b89c79fe4c451ba5cd9ff42deb3fd5c3be3f
2. If removing the board is not desireable, I can change everything on the `app/boards/seeeduino_xiao_ble.overlay`

### DIFF
--- a/app/boards/seeeduino_xiao_ble.overlay
+++ b/app/boards/seeeduino_xiao_ble.overlay
@@ -31,5 +31,24 @@
 };
 
 &qspi {
-    status = "disabled";
+    status = "okay";
+    pinctrl-0 = <&qspi_default>;
+    pinctrl-1 = <&qspi_sleep>;
+    pinctrl-names = "default", "sleep";
+    p25q16h: p25q16h@0 {
+        compatible = "nordic,qspi-nor";
+        reg = <0>;
+        sck-frequency = <104000000>;
+        quad-enable-requirements = "S2B1v1";
+        jedec-id = [85 60 15];
+        sfdp-bfp = [
+            e5 20 f1 ff  ff ff ff 00  44 eb 08 6b  08 3b 80 bb
+            ee ff ff ff  ff ff 00 ff  ff ff 00 ff  0c 20 0f 52
+            10 d8 08 81
+        ];
+        size = <16777216>;
+        has-dpd;
+        t-enter-dpd = <3000>;
+        t-exit-dpd = <8000>;
+    };
 };


caksoylar:
I don't have a strong opinion on removing `seeeduino_xiao_ble`, but from a user/support POV it would be great if some sort of backwards compatibility shim existed.

At the very least, I think it would be helpful to split this cleanly into two commits: one which fixes the battery use and one which removes the `seeeduino_xiao_ble` board. Then it would be easier to test individually and commit one or both depending on the eventual decision.

xudongzheng:
Have you measured the power consumption with these changes? Ideally measure and compare with Bluetooth disabled but even figures with Bluetooth could be useful.

Also worth mentioning that upstream Zephyr definition for `xiao_ble` changed from SPI to QSPI in Zephyr 3.4 so your change will break with the Zephyr 3.5 upgrade in #1995.

ldab:
> Have you measured the power consumption with these changes? Ideally measure and compare with Bluetooth disabled but even figures with Bluetooth could be useful.
> 
> Also worth mentioning that upstream Zephyr definition for `xiao_ble` changed from SPI to QSPI in Zephyr 3.4 so your change will break with the Zephyr 3.5 upgrade in #1995.
> 
> I would even try measuring on Zephyr 3.5 without any changes. See [zephyrproject-rtos/zephyr#55199](https://github.com/zephyrproject-rtos/zephyr/pull/55199) for some related issues. I haven't looked at this closely but one of those issues won't be in upstream Zephyr until 3.6.

Measuring with BT off is pointless since the USB controller + Battery charger will drain a lot more power than this fix and it would overshadow the fix.

I do not see how that would break with the upstream QSPI Zephyr, in my view it is the opposite actually. Currently there is ZMK own board definition inside Zephyr repo that could contain flags that are no longer supported. On the upstream, QSPI is defined and configured and enter DPD upon init(), unless I missed what you meant.

ldab:
> At the very least, I think it would be helpful to split this cleanly into two commits: one which fixes the battery use and one which removes the `seeeduino_xiao_ble` board. Then it would be easier to test individually and commit one or both depending on the eventual decision.

ok, I have made a separate commit.

Please find below the difference:
* from 260uA to 180uA on "idle"
* from 180uA to **1uA** on deep sleep :fireworks: 

## Measured at 3.3V as the LDO is undocumented and cant be trusted:

### My board before



### With this commit



I have also flashed `hummingbird` to a bare module for comparison

### Before

hummingbird_after

### After

hummingbird_after

caksoylar:
I see just one commit in the PR now, with the qspi node changes. Is that enough or were the older changes dropped by accident?

ldab:
> I see just one commit in the PR now, with the qspi node changes. Is that enough or were the older changes dropped by accident?

The 1x commit is enough

petejohanson (commented):
This all seems reasonable to me. We have our own board definition for historical reasons... Sadly, removing it now would break a lot of folks' existing builds. It would be nice to at least get the upstream board definition fully supported (by adding the necessary config and overlay ZMK specific additions on `app/boards`) and changing our metadata file to guide folks to that board ID, so we can slowly phase it out.

For now... Does this in fact fix things still for our board definition? If so, happy to merge, and grateful for the fix!

Kikangh:
> This all seems reasonable to me. We have our own board definition for historical reasons... Sadly, removing it now would break a lot of folks' existing builds. It would be nice to at least get the upstream board definition fully supported (by adding the necessary config and overlay ZMK specific additions on `app/boards`) and changing our metadata file to guide folks to that board ID, so we can slowly phase it out.
> 
> 
> 
> For now... Does this in fact fix things still for our board definition? If so, happy to merge, and grateful for the fix!

Hi Pete, is this something that could be used with my revxlp ? If so I'd gladly test it to see if I get longer battery life.

ldab:
> For now... Does this in fact fix things still for our board definition? If so, happy to merge, and grateful for the fix!

Yes, the overlay fix the current board definition with:

* from 260uA to 180uA on "idle"
* from 180uA to 1uA on deep slee

petejohanson (approved):
Flashed to my revxlp here with a XIAO BLE, and runs no problem. I trust OP measurements here, so merging so others can get the fix easily. We can follow up further if folks still report high power draw.

Thanks for digging and for the fix @ldab !

KLingO-MS:
with latest changes to seeeduino_xiao_ble.overlay (instead qspi disabled) my split KB with XIAO BLEs refuse to go to sleep anymore after first sleep-wakeup cycle ... borads need to be powered off-on to get to sleep again

I again built with qspi overridden to disabled and it works again

caksoylar:
I am also seeing randomness in being able to sleep, sometimes it sleeps fine, sometimes it won't. The symptoms look to be the same as #1901.

ldab:
> with latest changes to seeeduino_xiao_ble.overlay (instead qspi disabled) my split KB with XIAO BLEs refuse to go to sleep anymore after first sleep-wakeup cycle ... borads need to be powered off-on to get to sleep again
> 
> I again built with qspi overridden to disabled and it works again

On my tests as shown on the measurements above, it worked ok.

To help you, can you please create an issue and include a detailed description of your issue, including which board you use, the schematic, measurements and enable USB logs and attach it?

ldab:
> I am also seeing randomness in being able to sleep, sometimes it sleeps fine, sometimes it won't. The symptoms look to be the same as #1901.

The sleep code only looks for VBus detected, no?

KLingO-MS:
> > with latest changes to seeeduino_xiao_ble.overlay (instead qspi disabled) my split KB with XIAO BLEs refuse to go to sleep anymore after first sleep-wakeup cycle ... borads need to be powered off-on to get to sleep again
> > I again built with qspi overridden to disabled and it works again
> 
> On my tests as shown on the measurements above, it worked ok.
> 
> To help you, can you please create an issue and include a detailed description of your issue, including which board you use, the schematic, measurements and enable USB logs and attach it?

I'm running handwire split (10 keys on each half, directly connected to pins and GND - no matrix) on XIAO BLEs with latest ZMK build 
sequence:
- power on
- typing, and leave to rest until sleep timer expires
- enter the sleep (boards disappear in battery widget in MacOS)
- wakeup (i can see battery percentages in widget again)
- typing and leave to rest -- not going to sleep anymore (no matter how long I wait)

I have split battery reporting on and also using Brave's RGB widget - that's how I noticed that they are not sleeping .. as after returning to PC they do not blink battery status and connection after pressing keys (what RGB widget always does after wakeup) - this led me to checknig it on battery widget in MacOS and indeed they always stay on after first wakeup

after setting QSPI to disabled behavior want back to normal and boards go to sleep after timeout

my assumption is that when I'll try running logging they will not go to sleep as USB is connected

ldab:
> my assumption is that when I'll try running logging they will not go to sleep as USB is connected

Your assumption is correct, but the logs may show something that we are missing, you can change this and it will go to sleep when USB is connected: https://github.com/zmkfirmware/zmk/blob/466cf920296d74dc9f22ef02fbccb43024f14c81/app/src/activity.c#L70

Do you have your keyboard configuration? I can try to reproduce it here.

KLingO-MS:
I do not have local env .... building via GHA, so editing .c files is out of my reach i guess

my repo is here
https://github.com/KLingO-MS/zmk-config-KgO-BLE28/tree/BLE20

ldab:
It prints 

```
[00:00:14.324,859] <err> pm: Device p25q16h@0 did not enter suspended state (-140)
```

So I suppose it relates to the patch I mentioned above. If I remove the power (and battery) it works no problem. so there is a point where the QSPI is not handled properly.

caksoylar:
I am [now using](https://github.com/caksoylar/zmk-config/blob/zephyr-3.5/build.yaml#L22) a Zephyr 3.5-based branch (#2027, that is based on #1995) and the deep sleep is working OK after waking up from sleep. 
I suppose the upgrade from 3.2 to 3.5 might have some fixes related to the flash chip's power management.

I am not sure if the power consumption is normal after waking up since I don't have a power profiler, but I'll track overall battery life.

ldab:
> I am not sure if the power consumption is normal after waking up since I don't have a power profiler, but I'll track overall battery life.

I didn't flash the board you mentioned to my keyboard because the GPIOs are different, but I have build based on https://github.com/zmkfirmware/zmk/pull/1995 [my keyboard ](https://github.com/ldab/split-or-not-keyboard/blob/6d2f1f56749ac5582d0e0f6cebb14ece578a17ed/build.yaml#L21C12-L21C23) and the result is:

image

caksoylar:
That's interesting that the sleep current went up but the usage consumption looks pretty good, thanks for measuring! 

> I guess we can open an issue to track it and investigate more when 3.5.0 is merged, what do you think @caksoylar ?

I think that would be a good idea indeed.

PR #2093, opened 2024-01-04
fix(ble): Disable Auto Sec Req again.

petejohanson:
* Auto security request actually makes macOS worse, so disable it,
  and remove our early request in favor of using GATT enforcement
 to ensure connections are secured.

# Summary

After reviewing again the AUTO_SEC_REQ Kconfig from Zephyr, I was reminded we had actually turned it *off* for better mac compatibility. From their [Accessory Guidelines](https://developer.apple.com/accessories/Accessory-Design-Guidelines.pdf#//apple_ref/doc/uid/DontLinkBookID_1-BK99CH59-BK99CH59SLK0551):

> 49.10 Pairing
The accessory should not request pairing until an ATT request is rejected using the Insufficient
Authentication error code. See the
Bluetooth 4.0 specification, Volume 3, Part F, Section 4 for details.

So, we want to not only keep that setting off, we want to remove our explicit `bt_conn_set_security` in favor of having the GATT/ATT *built in* security enforcement preventing connected centrals from accessing things before our security is established.

@joelspadin had previously evaluated doing this, and I do think it's the "Right Thing To Do" after reviewing the BT spec, reviewing the Zephyr code for ATT/GATT permissions checks, and from discussions on the Zephyr Discord server.

Because this is... sorta scary, I added a few changes to our BLE tests to actually try and access a GATT characteristic in a few ways and seeing that our peripheral properly sends an "Insufficient Authentication" when trying to access the characteristic right after connecting.


app/tests/ble/central/src/main.c
```diff
@@ -260,6 +272,13 @@ static void discover_conn(struct bt_conn *conn) {
     }
 }
 
+static struct read_params;
```
joelspadin, line 275: Not sure what this is for. It seems to be a forward declaration for a type, but then it's written to as a variable later (and then never read from)?

petejohanson: Sorry, I had pushed a totally broken WIP from my mac when I realized I couldn't run our BLE tests there. Just force pushed fixes to actually make these tests sane.

app/src/ble.c
```diff
@@ -447,12 +447,6 @@ static void connected(struct bt_conn *conn, uint8_t err) {
 
     LOG_DBG("Connected %s", addr);
 
-#if !IS_ENABLED(CONFIG_BT_GATT_AUTO_SEC_REQ)
-    if (bt_conn_set_security(conn, BT_SECURITY_L2)) {
-        LOG_ERR("Failed to set security");
-    }
-#endif // !IS_ENABLED(CONFIG_BT_GATT_AUTO_SEC_REQ)
-
```
cvinayak, line 455: > So, we want to not only keep that setting off, we want to remove our explicit `bt_conn_set_security` in favor of having the GATT/ATT _built in_ security enforcement preventing connected centrals from accessing things before our security is established.

1. I assume, Central initiating pairing is checked for in the application for being in bondable state, right? or we just accept new bonds all the time from new centrals?
2. When connected to a bonded peer, is it ensured that notifications and indications have the required security level? where in the implementation is the request to encrypt initiated from the peripheral?

app/tests/ble/central/src/main.c
```diff
@@ -59,6 +63,26 @@ static void ble_central_native_posix_options(void) {
...
+        {.is_switch = true,
+         .option = "skip_discovery_on_connect",
+         .type = 'b',
+         .dest = (void *)&skip_discovery_on_connect,
+         .descript = "Skip GATT characteristic discovery after connecting"},
+        {.is_switch = true,
+         .option = "read_directly_on_discovery",
+         .type = 'b',
+         .dest = (void *)&read_directly_on_discovery,
+         .descript = "Read HIDS report after GATT characteristic discovery"},
```
cvinayak, line 85: Add a test for peripheral initiated encryption setup when a bonded central does not encrypt after connection; i.e. scenarios where notifications were enabled in the GATT server in a previous connection with the central.

cvinayak (changes requested):

app/src/ble.c
petejohanson: @cvinayak

1. So, whenever we advertise, and the selected profile is "open" we accept pairing requests. You can see that in our `auth_pairing_accept` implementation:

```
static enum bt_security_err auth_pairing_accept(struct bt_conn *conn,
                                                const struct bt_conn_pairing_feat *const feat) {
    struct bt_conn_info info;
    bt_conn_get_info(conn, &info);

    LOG_DBG("role %d, open? %s", info.role, zmk_ble_active_profile_is_open() ? "yes" : "no");
    if (info.role == BT_CONN_ROLE_PERIPHERAL && !pairing_allowed_for_current_profile(conn)) {
        LOG_WRN("Rejecting pairing request to taken profile %d", active_profile);
        return BT_SECURITY_ERR_PAIR_NOT_ALLOWED;
    }

    return BT_SECURITY_ERR_SUCCESS;
};
```

2. We don't explicitly request encryption when trying to notify, but the Zephyr GATT code in `gatt_notify` (in `subsystem/bluetooth/host/gatt.c`) has a permission check that causes it to return `-EPERM` if our target connection is not encrypted yet. That should keep us from accidentally notifying if we don't yet have our connection encrypted. That being said, I know that *most* hosts will automatically upgrade their connection when re-connecting to previously bonded peers, but it *might* be necessary to have our code request encryption on notification as a fallback. Thoughts?

app/tests/ble/central/src/main.c
petejohanson: So... After reviewing the code in Zephyr's `bt_gatt_connected`, that happens *only* if we have AUTO_SEC_REQ set. It would seem that Apple guidelines state that you should *not* do this automatically, seemingly even for existing bonded peers, so perhaps we should *only* do this *on attempting to notify* and we see that we still haven't had our connection encrypted yet. That *should* cover our use cases and still have better macOS compat?

app/src/ble.c
cvinayak: > whenever we advertise, and the selected profile is "open" we accept pairing requests. You can see that in our auth_pairing_accept implementation:

This is ok then.

> it might be necessary to have our code request encryption on notification as a fallback. Thoughts?

Yes. Maybe you could check macOS/iOS (i do not remember on top of my mind) when bond is removed, and you connect back to a keyboard that has the previous bond and press keys, then what is the behavior? Now sending a SMP Security Request PDU may be of help?

app/tests/ble/central/src/main.c
petejohanson: > > we should only do this on attempting to notify and we see that we still haven't had our connection encrypted yet. That should cover our use cases and still have better macOS compat?
> 
> Yes, only send SMP Security Req PDU when already bonded to the peer and not already encrypted in the connection to the required security level. Also, not to accept any new pairing but restrict to only initiating encryption using the existing keys. If the encryption fails, it should be regarded as a spoof device (that may have used the public Bluetooth address of the genuine central or somehow generated a private address of that genuine central to establish the connection).

Yup, this effectively does this since we only try to upgrade that connection if it's one from a host address stored in the active/selected peripheral slot, and the call to set_security under the hood sends the SMP security request PDU for us.

PR #2095, opened 2024-01-06
add LPM009M360A display driver

WSTRN:
IMG_1794
zmk_lpm009m360a
LPM009M360A is a 0.85-inch reflective active-matrix liquid crystal display (LCD).
The active area of the LCD has 72(h) x 144(v) pixels.
It looks very nice, has high resolution and very low power consumption(about 2uW in no-update mode and 7uW in update mode). But it doesn't have a backlight, you can only see clearly in a bright room.
So, I think this may be a new choice on everyone's new keyboard!


ZitaoTech:
This is really nice! Is there any example or config file of your numpad to use this display?

WSTRN:
> This is really nice! Is there any example or config file of your numpad to use this display?

https://github.com/WSTRN/DUC_KB_BPad10

petejohanson:
@WSTRN I'm a *bit* wary to keep adding more drivers like this to ZMK... We did so with a couple so far, but would love to try to either:

1. Try to submit upstream to Zephyr, so they can be maintained upstream in case they make significant driver changes, they'd in theory fix up the new drivers simultaneously. We could *also* pull into our Zephyr fork *temporarily* once the driver is upstreamed, knowing we'd get to drop that cherry pick once we update to newer Zephyr, OR
2. Maintain as a seperate Zephyr module, which is fairly easy now for folks to include in their builds via west.

Thoughts? I saw you closed this today, so wanted to reach out. Thanks for the PR!

WSTRN:
> @WSTRN I'm a _bit_ wary to keep adding more drivers like this to ZMK... We did so with a couple so far, but would love to try to either:
> 
> 1. Try to submit upstream to Zephyr, so they can be maintained upstream in case they make significant driver changes, they'd in theory fix up the new drivers simultaneously. We could _also_ pull into our Zephyr fork _temporarily_ once the driver is upstreamed, knowing we'd get to drop that cherry pick once we update to newer Zephyr, OR
> 2. Maintain as a seperate Zephyr module, which is fairly easy now for folks to include in their builds via west.
> 
> Thoughts? I saw you closed this today, so wanted to reach out. Thanks for the PR!

Thanks! you are right, I'll try to support this display in another way.

yangxing844:
I copied the author's code, and ported it into a zmk driver module https://github.com/yangxing844/zmk-lpm009m360a-driver , I remove the required ext_com and disp since most of the screen on zmk will only use pinout compatible to nice!view

PR #2097, opened 2024-01-06
fix(split): Split peripherals should auto sec req still.

petejohanson:
* Ensure split peripherals have `BT_GATT_AUTO_SEC_REQ` enabled so that
  reconnects to centrals are automatically encrypted.


petejohanson:
Merging, I accidentally broke split reconnects when I merged #2093 and removed the explicit set security for peripheral connections.

PR #2098, opened 2024-01-06
fix(boards): Fix up EXT_POWER naming for adv360pro;

petejohanson:
* Restore setting loading by preserving old device name for the external power node.

### DIFF
--- a/app/boards/arm/adv360pro/adv360pro.dtsi
+++ b/app/boards/arm/adv360pro/adv360pro.dtsi
@@ -44,7 +44,9 @@
 
         >;
     };
-    ext-power {
+
+    // Node name must match original "EXT_POWER" label to preserve user settings.
+    EXT_POWER {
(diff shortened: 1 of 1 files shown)


app/boards/arm/adv360pro/adv360pro.dtsi
```diff
@@ -44,7 +44,7 @@
 
         >;
     };
-    ext-power {
+    EXT_POWER {
```
joelspadin, line 47: Might be helpful to include the
```c
// Node name must match original "EXT_POWER" label to preserve user settings.
```
comment that's used in other shields.

joelspadin (approved):

app/boards/arm/adv360pro/adv360pro.dtsi
petejohanson: Added

PR #2099, opened 2024-01-06
fix(battery): prevent bus fault when battery does not exist

xudongzheng:
zmk_battery_start_reporting() may be called from battery_event_listener(), which will result in a bus fault when attempting to read a battery that does not exist such as on a dongle.

### DIFF
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -87,7 +87,9 @@ static void zmk_battery_timer(struct k_timer *timer) {
 K_TIMER_DEFINE(battery_timer, zmk_battery_timer, NULL);
 
 static void zmk_battery_start_reporting() {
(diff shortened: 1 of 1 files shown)


app/src/battery.c
```diff
@@ -87,7 +87,9 @@ static void zmk_battery_timer(struct k_timer *timer) {
 K_TIMER_DEFINE(battery_timer, zmk_battery_timer, NULL);
 
 static void zmk_battery_start_reporting() {
-    k_timer_start(&battery_timer, K_NO_WAIT, K_SECONDS(CONFIG_ZMK_BATTERY_REPORT_INTERVAL));
+    if (device_is_ready(battery)) {
```
petejohanson, line 90: Seems reasonable, although I'm surprised that startup even succeeds when the init function below this fails with an error code using this same check.

xudongzheng: I believe the early return in the init function prevents zmk_battery_start_reporting() from being called on startup.

petejohanson (approved):

PR #2102, opened 2024-01-08
refactor: replace zmk,matrix_transform with zmk,matrix-transform

lesshonor:
* While functionally equivalent, the hyphenated form of this property is more consistent with other ZMK properties and adheres to [DTS style guidelines](https://buildmedia.readthedocs.org/media/pdf/devicetree-specification/latest/devicetree-specification.pdf).
* ...Since I'm already touching a number of files, might as well update links to use Zephyr 3.2 documentation instead of 2.5 where appropriate.

---

~~This PR intentionally refrains from updating any boards or shields to minimize disruption, as some forks and/or utilities may currently expect/require the underscore form ([Keymap Editor](https://nickcoutsos.github.io/keymap-editor/) being one).~~

### DIFF
--- a/app/boards/arm/adv360pro/adv360pro.dtsi
+++ b/app/boards/arm/adv360pro/adv360pro.dtsi
@@ -25,7 +25,7 @@
         zmk,kscan = &kscan0;
         zmk,backlight = &backlight;
         zmk,battery = &vbatt;
-        zmk,matrix_transform = &default_transform;
+        zmk,matrix-transform = &default_transform;
         zmk,underglow = &led_strip;
     };
 
--- a/app/boards/arm/bt60/bt60.dtsi
+++ b/app/boards/arm/bt60/bt60.dtsi
@@ -19,7 +19,7 @@
         zephyr,console = &cdc_acm_uart;
         zmk,battery = &vbatt;
         zmk,kscan = &kscan0;
-        zmk,matrix_transform = &default_transform;
+        zmk,matrix-transform = &default_transform;
     };
 
     sensors: sensors {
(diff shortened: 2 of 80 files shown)


joelspadin:
I think it may be confusing if the docs say one thing but all the code does something different. Can we check if it works to update the code to use hyphens and then build a user config repo that still uses an underscore when selecting a non-default transform? If we can confirm that it isn't a breaking change and the user config still selects the correct tranform, then we can update all the existing boards/shields to match this new documentation too.

caksoylar:
To analyze your results, using `zmk,matrix_transform = &five_column_transform;` in a keymap with the keyboard definition having `zmk,matrix-transform = &something` is working as expected -- as it results in excess array initializer errors with a 6 col keymap -- right?

caksoylar:
I think best would be to test and confirm that the transform choice was effective, but absent that the build warnings seem like a good sign.

caksoylar (approved):
Docs updates LGTM, thanks for updating the links as well -- they all seem to be live.

(I think it is a good idea in general too, given it is backwards compatible.)

caksoylar:
> I've done that; I personally use [this keymap](https://github.com/lesshonor/zmk-config/blob/0790f22e953b688cefbe96576687a44a100342ae/config/bb40.keymap#L23), which draws on [this module](https://github.com/lesshonor/mechwild-zmk-keyboards/blob/12c723c8935b0cc1c7d7d346c67d67d72acacb7b/boards/shields/bb40/bb40.overlay#L12) (and you're welcome to examine [its last build](https://github.com/lesshonor/zmk-config/actions/runs/7456325860/job/20286789750))—but how do you propose I demonstrate the results?

Maybe I missed it, but I didn't know that you had already tested it on a keyboard. Verbal confirmation is enough for me, personally :)

petejohanson (approved):
Thanks!

PR #2103, opened 2024-01-09
Change KP_MULTIPLY to ASTRK in corne.keymap

TheTuxis:
This change fixes the key mapping for the coner keyboard in the raise_layer.

The previous configuration works only for keyboards with an English layout, with this change it works in all languages.

### DIFF
--- a/app/boards/shields/corne/corne.keymap
+++ b/app/boards/shields/corne/corne.keymap
@@ -46,9 +46,9 @@
 // | SHFT |     |     |     |     |     |   |  _  |  +  |  {  |  }  | "|" |  ~   |
 //                    | GUI |     | SPC |   | ENT |     | ALT |
                         bindings = <
-   &kp  TAB  &kp EXCL &kp AT &kp HASH &kp DLLR &kp PRCNT   &kp CARET &kp AMPS  &kp KP_MULTIPLY &kp LPAR &kp RPAR &kp BSPC
-   &kp LCTRL &trans   &trans &trans   &trans   &trans      &kp MINUS &kp EQUAL &kp LBKT        &kp RBKT &kp BSLH &kp GRAVE
-   &kp LSHFT &trans   &trans &trans   &trans   &trans      &kp UNDER &kp PLUS  &kp LBRC        &kp RBRC &kp PIPE &kp TILDE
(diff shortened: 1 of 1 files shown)


TheTuxis:
Thank @lesshonor, I tried the original configuration in Spanish and the problem was that it strictly uses the *, however with the current configuration it considers "shift + 8".

caksoylar:
How about fixing this in other default keymaps as well? I see the same issue for:
- `microdox`
- `splitkb_aurora_corne`
- `lily58`
- `sofle`
- `iris`
- `corneish_zen`

TheTuxis:
> How about fixing this in other default keymaps as well? I see the same issue for:
> 
> * `microdox`
> * `splitkb_aurora_corne`
> * `lily58`
> * `sofle`
> * `iris`
> * `corneish_zen`

Do you want me to include it in the same PR?

caksoylar:
I think that would be appropriate, if you update the PR name accordingly.

TheTuxis:
> I think that would be appropriate, if you update the PR name accordingly.

I already changed them in **corne**, **microdox**, **iris** and **splitkb_aurora_corne**.

In **lily58** and **sofle** it would not correspond because the character that is represented is strictly the `*`, not `shift + 8`.

Note: I didn't find the corneish_zen keymap, will you use any of the others?

caksoylar:
Re: lily58 and sofle, I mean [e.g. this `KP_MULTIPLY` and `KP_PLUS`](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/lily58/lily58.keymap#L44C127-L44C142) which don't seem to belong there, since they aren't near the other keypad keycodes. You can see they are [fixed in the splitkb variants](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/splitkb_aurora_lily58/splitkb_aurora_lily58.keymap#L44C127-L44C135) already.

Here is the `corneish_zen` one: https://github.com/zmkfirmware/zmk/blob/main/app/boards/arm/corneish_zen/corneish_zen.keymap#L61C80-L61C92

For the PR title, something like "Replace some keypad keycodes in default keymaps" might be generic and informative enough to cover these changes.

TheTuxis:
All complete suggestions

caksoylar (approved):
LGTM, thanks.

TheTuxis:
@caksoylar @lesshonor 
I don't have permission to merge the pr, do I have to do something else?

caksoylar:
No, one of the core contributors will need to have a look and merge.

petejohanson (approved):
Thanks!

PR #2104, opened 2024-01-09
refactor: remove/replace deprecated properties

lesshonor:
A few overlays scattered across boards, shields and tests appear to have been missed when updating/upgrading these properties.

~~Tagging @ReFil who may or may not be `crbn`'s maintainer: `resolution` was replaced with sane defaults for `steps` and `triggers-per-rotation`, but I am unfamiliar with this shield. If there are specific known values that should go here, please suggest.~~ Handled in #2143.


petejohanson (approved):
Thanks!

PR #2106, opened 2024-01-10
refactor: Use Kconfig to conditionally compile behaviors

bryanforbes:
This pull request updates the current behaviors to use `Kconfig.behaviors` and `target_sources_ifdef()` within `app/CMakeLists.txt` to conditionally compile behaviors based on the behaviors used in the user's keymap


app/src/behaviors/behavior_transparent.c
```diff
@@ -14,8 +14,6 @@
 
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
-#if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
```
caksoylar, line 17: Should `DT_DRV_COMPAT` defines also be cleaned up?

docs/docs/development/new-behavior.md
```diff
@@ -320,38 +317,47 @@ The fifth cell of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if inst
...
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_CAPS_WORD app PRIVATE src/behaviors/behavior_caps_word.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_KEY_REPEAT app PRIVATE src/behaviors/behavior_key_repeat.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_MACRO app PRIVATE src/behaviors/behavior_macro.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_MOMENTARY_LAYER app PRIVATE src/behaviors/behavior_momentary_layer.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_MOD_MORPH app PRIVATE src/behaviors/behavior_mod_morph.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_OUTPUTS app PRIVATE src/behaviors/behavior_outputs.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_TOGGLE_LAYER app PRIVATE src/behaviors/behavior_toggle_layer.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_TO_LAYER app PRIVATE src/behaviors/behavior_to_layer.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_TRANSPARENT app PRIVATE src/behaviors/behavior_transparent.c)
+  target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_NONE app PRIVATE src/behaviors/behavior_none.c)
```
caksoylar, line 352: This is not quite matching https://github.com/zmkfirmware/zmk/blob/cb427b8a5c99bb79dfaff6be1673142b73bc0a7b/app/CMakeLists.txt, but it is a moving target anyway. How about a `/* ... */` in the beginning and end, and reduce the specific examples to a few?

caksoylar (commented):
LGTM in general from a docs perspective. Note that if you remove the `#define` as in the first comments the docs need to be updated for that too.

app/src/behaviors/behavior_transparent.c
bryanforbes: No, because `BEHAVIOR_DT_INST_DEFINE` uses it

docs/docs/development/new-behavior.md
bryanforbes: That's a good idea. I'll update it.

app/Kconfig.behaviors
```diff
@@ -1,17 +1,87 @@
...
+
+config ZMK_BEHAVIOR_OUTPUTS
+    bool
+    default y
+    depends on DT_HAS_ZMK_BEHAVIOR_OUTPUTS_ENABLED
+
+config ZMK_BEHAVIOR_RESET
+    bool
+    default y
+    depends on DT_HAS_ZMK_BEHAVIOR_RESET_ENABLED
```
petejohanson, line 78: So this one is a bit tricky... This is needs to be enables a bit more specifically for the right side of splits, even if it's not referenced in a keymap. There's a few of those, like the RBG behavior, that are either "local" or "global" so need this special handling.

app/Kconfig.behaviors
```diff
@@ -1,17 +1,87 @@
...
+
+config ZMK_BEHAVIOR_RGB_UNDERGLOW
```
petejohanson, line 80: Ditto.

app/dts/behaviors/ext_power.dtsi
```diff
@@ -7,7 +7,7 @@
 / {
     behaviors {
         // Behavior can be invoked on peripherals, so name must be <= 8 characters.
-        ext_power: extpower {
+        /omit-if-no-ref/ ext_power: extpower {
```
petejohanson, line 10: This was intentional, so probably needs an added comment, instead of adding the `omit-if-no-ref`.

app/dts/behaviors/reset.dtsi
```diff
@@ -9,13 +9,13 @@
 / {
     behaviors {
         // Behavior can be invoked on peripherals, so name must be <= 8 characters.
-        sys_reset: sysreset {
+        /omit-if-no-ref/ sys_reset: sysreset {
```
petejohanson, line 12: Ditto in this file.

petejohanson (changes requested):
A few high level comments on how this fits with splits.

app/Kconfig.behaviors
bryanforbes: This makes sense. I've removed the Kconfig options for the behaviors we talked about in Discord and reverted them back to using the main Kconfig option where available. This means the reset behaviors are always compiled in for both sides.

PR #2109, opened 2024-01-13
chore(docs): Bump docusaurus to 3 and update deps

caksoylar:
This PR updates docusaurus to version 3, while also updating other dependencies to latest where appropriate, following https://docusaurus.io/docs/migration/v3. I think it should resolve the 4 dependabot alerts as well.

When I tried this before I had run into some issues, related to mdx upgrade from v1 to v3. But only one page turned out to be affected so that was easy too fix in d070744ceddb5d01855b1e851df5b26c637a9bb9.

The `refactor` commits are updating things under the "Optional Changes" header in the migration guide so while they aren't strictly necessary now, they might be issues later on. I also snuck in a bug fix at 8295d764345b2b44bc7f377e073e98d8c99f554e since I was touching that line in the following commit.

This still needs extensive inspection to make sure nothing broke. (The migration guide recommends visual regression tests which is something we might want to invest into at some point, but it didn't seem worth it at this point to me.)


caksoylar:
This is ready for review, but both both `eslint` and `typecheck` are failing. Will look into how to update things there, but any fix suggestions are welcome since I am not familiar with them.

caksoylar:
I compared current and PR deploy results side-by-side for most of the pages -- they were all good except for the one issue solved by 33c6a37. I first tried converting those headers to regular markdown headers but that broke the TOC. Looks like https://github.com/facebook/docusaurus/pull/9684 might make creating our own `toc` unnecessary in the future.

petejohanson:
@caksoylar All seems reasonable to me, and some pragmatic fixes for the couple sticky issues. No concerns here.

caksoylar:
@petejohanson Thanks! If you are happy with it, do you mind giving a formal review? I can't approve the PR myself.

petejohanson (approved):

PR #2112, opened 2024-01-14
remove gd25q16 from xiao_ble

ldab:
Following https://github.com/zmkfirmware/zmk/pull/2091.. the QSPI from zephyr was still defined and prevented (i think) the device to sleep, logs showed `pm: Device GD25Q16 did not enter suspended state (-140)`

On my measurements it did indeed went to sleep, but this seems necessary anyway.

There is something still strange on Zephyr side, probably related to https://github.com/zephyrproject-rtos/zephyr/pull/64782

I will investigate further and tidy it up on Zephyr side and update the board overlay when fixed. For now this PR should be required.


KLingO-MS:
I put /delete-node/ &gd25q16; into kaymap file and using Your updated XIAO overlay with QSPI enabled ... unfortunatelly board does not go to sleep again

ldab:
> I put /delete-node/ &gd25q16; into kaymap file and using Your updated XIAO overlay with QSPI enabled ... unfortunatelly board does not go to sleep again

This is due the problem mentioned on the PR description. If you remove the power completely after flashing my test shows that it works as it should, have you tried that?

KLingO-MS:
Yes when I flash I laways have bettery disconnected .. so after removing USB it is without power until i switch it on ... it went to sleep , but after wakeup it does not go to sleep again .. as described in other thred

KLingO-MS:
and I chcecked generated devicetree file in GHA and gd25q16 is not present

ldab:
@KLingO-MS I moved to SPI, can you please let me know if that fixes for you?

ldab:
with this changes the sleep current significantly increases:

image

I will leave as it is. Unless my board is cooked and someone else can verify

KLingO-MS:
So with Your latest change to overlay file (spi2 , and qspi disabled) and /delete-node/ &gd25q16; in keymap file it is going properly to sleep every time.
I have no means to measure consumption, but I assume that in sleep it will consume less than not sleeping and being only in idle.

ldab:
> I have no means to measure consumption, but I assume that in sleep it will consume less than not sleeping and being only in idle.

Not on my tests. My keyboard does 40uA in idle and 1uA with sleep using QSPI.

With the changes on this PR it sleeps with 220uA.

PR #2117, opened 2024-01-15
docs: confirm some consumer codes on iOS/macOS

crides:
Tested on iOS 16.6, macOS 14.1


caksoylar (approved):
I could confirm the non-brightness ones here as well (haven't tested the others), thanks.

PR #2122, opened 2024-01-18
Input Listener: Auto Layer Toggle on mouse movement

infused-kim:
This PR adds a way to toggle a layer whenever a mouse moves.

It was designed for the PS/2 Mouse & Trackpoint PR, but since it's using the new input listener system, it can be used with any mouse driver that's using the new zephyr 3.5 input system.

## Configuration
In your `shield.dtsi`:

```dtsi
/ {
    mouse_ps2_input_listener: mouse_ps2_input_listener {
        compatible = "zmk,input-listener";
        status = "disabled";

        device = <&mouse_ps2>;
    };
};
```

As you can see the input config is disabled by default. Enable it only on the central half of your keyboard (such as `shield_right.overlay`:

```dtsi
&mouse_ps2_input_listener {
    status = "okay";
};
```
 
And then in your actual keymap:
```dtsi
&mouse_ps2_input_listener {
  xy-swap;
  x-invert;
  y-invert;

  // Set the layer that should automatically activate when the mouse is
  // moving.
  layer-toggle = <MOUSE_TP>;

  // How long the mouse needs to move for before the layer is activated (to
  // avoid accidental activations while typing) (Default: 250)
  layer-toggle-delay-ms = <250>;

  // How long to wait, after the last mouse movement, before deactivating
  // the layer (Default: 250)
  layer-toggle-timeout-ms = <250>;
};
```


infused-kim:
I updated the code to the new input listener system and updated the PR description with the new config.

Nick-Munnich:
This has been made obsolete by the input processors PR I believe.

PR #2126, opened 2024-01-22
refactor: Improve keymap upgrader

joelspadin:
Moved the keymap upgrader to a top-level page like the power profiler to make it more discoverable. It upgrades more things than key codes now, so putting it in the codes category doesn't make much sense.

Updated the page with syntax highlighting, highlighting of changed lines, and a more visible warning that it cannot handle codes in some locations such as inside `ZMK_MACRO()`.

Converted the upgrader code to TypeScript and split it up into smaller files to make it easier to add new upgrade functions.

Added upgrade functions to remove/replace "label" properties, rename matrix-transform.h to matrix_transform.h, and fix renamed behavior nodes in the unlikely event that someone was changing behavior settings using those instead of with node references.

Fixed key code replacement to only apply inside `bindings` properties, since identifiers such as `MOD_LCTL` should be upgraded when used as key codes, but they are still used in `mods` properties.


docs/src/css/custom.css
```diff
@@ -15,10 +15,15 @@
...
+
+  --docusaurus-highlighted-code-line-bg: rgb(0 0 0 / 8%);
+}
+
+[data-theme="dark"] {
+  --docusaurus-highlighted-code-line-bg: rgb(255 255 255 / 8%);
 }
 
 .docusaurus-highlight-code-line {
-  background-color: rgb(72, 77, 91);
```
caksoylar, line 21: This is a good change, especially for the light theme. Thanks for fixing. For others' reference, here is old vs. new highlights on the light theme:
image

docs/src/keymap-upgrade/textedit.ts
```diff
@@ -0,0 +1,182 @@
...
+  changedRanges: Range[];
+}
+
+interface TextChunk {
+  text: string;
+  changed?: boolean;
+}
+
+/**
+ * Returns a string with text replacements applied.
```
caksoylar, line 105: Would be good to update the docstring.

docs/src/keymap-upgrade/textedit.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    chunks.push({ text: text.substring(currentIndex, edit.startIndex) });
+    chunks.push({ text: edit.newText, changed: true });
+    currentIndex = edit.endIndex;
+  }
+
+  chunks.push({ text: text.substring(currentIndex) });
+
+  return chunks.reduce<EditResult>(
```
caksoylar, line 128: Perhaps a comment explaining this reduce would be helpful.

docs/src/keymap-upgrade/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    const startLine = positionToLineNumber(start, lineBreaks);
+    const endLine = positionToLineNumber(end, lineBreaks);
+
+    return startLine === endLine ? `${startLine}` : `${startLine}-${endLine}`;
+  });
+
+  return `{${changedLines.join(",")}}`;
+}
+
+function getLineBreakPositions(text: string) {
```
caksoylar, line 43: Can this be simplified by something like below?
```ts
Array.from(text).map((c, i) => c === '\n' ? i : -1).filter(i => i !== -1)
```
(Apparently `Array.map` [assigns the index to the second arg](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#callbackfn) for the callback)

caksoylar (commented):
I think this is very helpful, and the new highlighting of changed lines is great, thanks! I noted a few questions, some for my learning.

Also the highlighting if one line is deleted doesn't seem quite right, here is a reproducer:
```dts
        gresc: grave_escape {
            label = "xx";
            bindings = <&kp ESC>, <&kp GRAVE>;
            mods = <(MOD_LGUI|MOD_LSFT|MOD_RGUI|MOD_RSFT)>;
        };
```
~~Lines both 3 and 4 end up getting highlighted after `label = "xx";` line is deleted. There isn't a great way to show deleted lines by just highlighting lines in the output, but maybe highlighting the line after it is good enough. But it is pretty prone to edge cases, so if it is easier to not highlight anything then that might be fine too.~~

Edit: I missed that the upgrader replaces mod flags like `MOD_LGUI` with `LGUI` etc., hence line 4 also being highlighted. That seems like it would be good to fix, I guess we should remove them from the `CODES` map?

docs/src/keymap-upgrade/textedit.ts
```diff
@@ -0,0 +1,153 @@
...
+  constructor(startIndex: number, endIndex: number, newText: string);
+  /**
+   * Creates a text edit to replace a node with new text.
+   */
+  constructor(node: SyntaxNode, newText: string);
+  constructor(
+    startIndex: number | SyntaxNode,
+    endIndex: number | string,
+    newText?: string
+  ) {
```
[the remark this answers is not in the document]
joelspadin: JavaScript doesn't actually support overloaded functions, so the only way to emulate overloads is by checking the types of the parameters inside the function. Flipping the order around would make this code simpler though.

docs/src/keymap-upgrade/index.ts
joelspadin: It could be written more simply like that, but this implementation is [much faster](https://jsperf.app/yatiqa/3) on account of scanning through the original string without modifying it instead of turning it into an array of one-character strings.

joelspadin:
> ~Lines both 3 and 4 end up getting highlighted after `label = "xx";` line is deleted. There isn't a great way to show deleted lines by just highlighting lines in the output, but maybe highlighting the line after it is good enough. But it is pretty prone to edge cases, so if it is easier to not highlight anything then that might be fine too.~

Yeah. Ideally, we'd be able to put some sort of highlight in between the lines, but I couldn't find a way to do that. Stripping removed lines from the highlighting was harder than just having it highlight the following line.

> Edit: I missed that the upgrader replaces mod flags like `MOD_LGUI` with `LGUI` etc., hence line 4 also being highlighted. That seems like it would be good to fix, I guess we should remove them from the `CODES` map?

Will fix. I think it would be better to just restrict the replacements to within a `bindings` array, since `MOD_RSFT` is still valid inside `mods = <MOD_RSFT>` but no longer in `bindings = <&kp MOD_RSFT>`

docs/src/keymap-upgrade/index.ts
caksoylar: Interesting, that's indeed a very large difference. Thanks for the info.

docs/src/keymap-upgrade/textedit.ts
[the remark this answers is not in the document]
joelspadin: I decided to just simplify the constructor to only taking start/end indices and add a `fromNode()` factory function instead.

docs/src/keymap-upgrade/textedit.ts
joelspadin: Done.

docs/src/keymap-upgrade/textedit.ts
joelspadin: Done. I also pulled some more of this code out into a separate function to simplify it.

docs/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "@docusaurus/tsconfig/tsconfig.json",
+  "extends": "@docusaurus/tsconfig",
```
joelspadin, line 2: This didn't match Docusaurus's documentation before. I think it was failing to resolve the parent config until I fixed this, because TypeScript was complaining about uses of `import.meta` in CommonJS modules, but this config sets the module type to `esnext`.

caksoylar: Good catch, I am surprised I didn't run into any issues in #2109 related to that.

docs/src/keymap-upgrade/keycodes.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+/**
+ * Given a identifier, returns whether it is inside a "bindings" property value.
+ */
+function isInBindingsArray(identifier: SyntaxNode) {
+  let node = identifier.parent;
+  while (node) {
+    if (node.type === "property") {
+      return node.childForFieldName("name")?.text === "bindings";
```
caksoylar, line 159: Might be worth checking for `sensor-bindings` as well, if not a regex/suffix to cover future cases.

docs/src/pages/keymap-upgrader.mdx
```diff
@@ -0,0 +1,23 @@
...
+sidebar_label: Keymap Upgrader
+hide_title: true
+hide_table_of_contents: true
+---
+
+# Keymap Upgrader
+
+Some behaviors, key codes, and other features have been renamed to be more consistent with each other. This tool will upgrade most deprecated features to their replacements.
+
+Paste the contents of a `.keymap` file below. Then, hover your mouse over the upgraded keymap and click the `Copy` button in the upper-right corner to copy it to your clipboard.
```
caksoylar, line 12:
```suggestion
Paste the contents of a `.keymap` file below. Then, hover your mouse over the upgraded keymap and click the `Copy` button in the upper-right corner to copy it to your clipboard.
Lines that have modified content will be highlighted.
```
Not mentioning the line deletion scenario but I think we can keep it simple.

joelspadin: Done.

docs/src/keymap-upgrade/keycodes.ts
joelspadin: Good catch. Done.

joelspadin:
I added another upgrade function to handle renamed nodes. I don't expect many people were changing behavior settings this way, but in case they were, this should fix it for them.

docs/src/keymap-upgrade/parser.ts
```diff
@@ -0,0 +1,95 @@
...
+
+function getNodeQuery(names: string[], depth = 0): string {
+  if (names.length === 0) {
+    throw new TypeError();
+  }
+
+  const [name, ...rest] = names;
+
+  const nameCapture = `@_name_${depth}`;
+  const nameQuery = `name: (identifier) ${nameCapture} (#eq? ${nameCapture} "${name}")`;
```
caksoylar, line 88: How does tree-sitter distinguish node label from name? This doesn't seem to be recognized and replaced:
```dts
mt: behavior_mod_tap
```
while dropping the `mt: ` works fine.

joelspadin: Not in a very useful way... It wraps the labeled thing in a new syntax node. That's obnoxious for queries, so I just changed it to add a label field to the labeled thing instead (that change isn't pulled in to ZMK yet), but we'd still run into the same problem in cases like code wrapped by `#if ... #endif`, so I reworked this function to ignore all intermediate syntax nodes that aren't devicetree nodes.

joelspadin: Updated tree-sitter-devicetree to 0.9.0. No changes to the behavior here, but it should handle keymaps with `#if` and friends more correctly now.

caksoylar (approved):
Looks good to me, thanks for implementing the enhancements.

Given the utility functions you added, at some point in the future it might be good to pull out the DT-related helpers to a separate package, what you think? (Perhaps it can be utilized by e.g. ZMK Tools as well.)

joelspadin:
> Given the utility functions you added, at some point in the future it might be good to pull out the DT-related helpers to a separate package, what you think? (Perhaps it can be utilized by e.g. ZMK Tools as well.)

Might be useful at some point. Right now I'm not doing much devicetree inspection or editing in ZMK Tools.

joelspadin:
An upgrade function for `resolution` -> `steps` + `triggers-per-rotation` would be useful, but is much more complex, so I'll try that in a separate PR.

PR #2131, opened 2024-01-23
Refactor logging for behaviors

bryanforbes:
I've refactored the logging in behaviors to be more granular with the goal being to eventually use more granular logging everywhere. The refactor is as follows:

* Added `CONFIG_ZMK_BEHAVIORS_LOG_LEVEL` which, if unset, inherits the log level from `CONFIG_ZMK_LOG_LEVEL` (`behavior.c` and `behavior_queue.c` use this log level)
* Added `CONFIG_ZMK_BEHAVIOR_*_LOG_LEVEL` for each behavior which, if unset, inherit from `CONFIG_ZMK_BEHAVIORS_LOG_LEVEL`
* Ensured the behavior logging options show up as a sub-menu in `menuconfig`

With these additions, a developer can do the following:

* Completely turn off all behavior logging by setting `CONFIG_ZMK_BEHAVIORS_LOG_LEVEL_OFF=y`
* Only get logging for individual behaviors, such as caps word, by setting `CONFIG_ZMK_BEHAVIORS_LOG_LEVEL_OFF=y` and `CONFIG_ZMK_BEHAVIOR_CAPS_WORD_LOG_LEVEL_DBG=y`


caksoylar:
I would wait for input on the idea/implementation from others before committing to doing it, but I'll drop this note: This would need docs updates for the new behavior guide and possibly some pointers in the USB logging page.

nmunnich (changes requested):
I'm a fan of the rough idea of finer grained logging. That said, this would need further docs work as caksoylar mentioned. As it's been a while since this PR has been touched on, it would also need a rebase and the commit messages need to be brought in line with our guidelines.

PR #2135, opened 2024-01-26
fix(ci): Updated actions to move away Node 16.

petejohanson:


petejohanson:
Auditing things, looks like https://github.com/pre-commit/action/pull/188 needs to be resolved before we can fix the pre-commit/action usage that leads to the last lingering error about Node 16.

PR #2140, opened 2024-01-27
Update preonic_rev3.dts

senseored:
Quick fix to make USB-logging compile and work on the Preonic. Tested on my keyboard.


petejohanson (approved):
Thanks!

PR #2143, opened 2024-01-28
Add encoder resolution fix to keymap upgrader

joelspadin:
Updated the keymap upgrader to upgrade the encoder `resolution` property to `steps` and `triggers-per-rotation`. 

The `steps` property will be set assuming triggers-per-rotation is 20. To simplify the code, it does not go looking for an existing triggers value since it's unlikely that you would have set `triggers-per-rotation` but not `steps`.

If `steps` is added and `triggers-per-rotation` is not present, a new `&sensors { ... }` block is added with the property set to 20. To simplify the code, it does not go looking for an existing sensors node to insert the property into.

Both added properties include a comment explaining how to fix the value to match your encoder.

Fixed an issue where appending text to the end of the file would highlight the wrong lines.

Also updated the shields that still had a `resolution` property, and fixed shields whose `/sensors` node did not have a `sensors` label, since the upgrader requires that label or it will generate broke code.

### DIFF
--- a/app/boards/shields/kyria/kyria_rev3.dtsi
+++ b/app/boards/shields/kyria/kyria_rev3.dtsi
@@ -29,13 +29,11 @@
 };
 
 &left_encoder {
-    resolution = <2>;
     a-gpios = <&pro_micro 20 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
     b-gpios = <&pro_micro 21 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
 };
 
 &right_encoder {
-    resolution = <2>;
     a-gpios = <&pro_micro 20 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
     b-gpios = <&pro_micro 21 (GPIO_ACTIVE_HIGH | GPIO_PULL_UP)>;
 };
--- a/docs/src/keymap-upgrade/encoder.ts
+++ b/docs/src/keymap-upgrade/encoder.ts
@@ -1,0 +1,101 @@
+import type { SyntaxNode, Tree } from "web-tree-sitter";
+
+import {
+  getContainingDevicetreeNode,
+  getDevicetreeNodePath,
+  findDevicetreeProperty,
+} from "./parser";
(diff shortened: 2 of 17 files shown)


app/boards/shields/kyria/kyria_rev3.dtsi
```diff
@@ -29,13 +29,13 @@
 };
 
 &left_encoder {
-    resolution = <2>;
+    steps = <80>;
```
caksoylar, line 32: I think steps or resolution don't need to be set here since the common dtsi sets them: https://github.com/zmkfirmware/zmk/blob/72d569ee155ef9b78269ed0813596d5d758c7482/app/boards/shields/kyria/kyria_common.dtsi#L24

docs/src/keymap-upgrade/encoder.ts
```diff
@@ -0,0 +1,101 @@
...
+  // back to a default if the value is something more complex than a single int.
+  const resolution = value.text.trim().replaceAll(/^<|>$/g, "");
+  const steps =
+    (parseInt(resolution) || DEFAULT_RESOLUTION) * TRIGGERS_PER_ROTATION;
+
+  const hint = `/* Change this to your encoder's number of detents times ${resolution} */`;
+
+  return [
+    TextEdit.fromNode(name, "steps"),
+    TextEdit.fromNode(value, `<${steps}> ${hint}`),
```
caksoylar, line 76: Nit: Would be nice to put the comment after the semicolon but that doesn't play nicely with the parsing/editing approach, I think.

caksoylar (approved):
Seems to work well here and implementation looks good to me, see the added a comment on the Kyria steps property.

docs/src/keymap-upgrade/encoder.ts
joelspadin: Yeah, I could maybe do that by replacing the whole property node, but this way way easier to implement.

app/boards/shields/kyria/kyria_rev3.dtsi
joelspadin: Fixed.

caksoylar (approved):

PR #2150, opened 2024-02-04
fix(boards): Fix build warnings for Corne-ish Zen builds

caksoylar:
Fixes warnings for two separate issues:
- Missed include for battery header in display widget, after the change in 6f8d080b6 (similar to f71458aca that fixed it for nice!view)
  - Removed no-longer-used BAS header includes
    - Some other files (`nice_view` and stock battery widget) have these as well, but I can't test a build for these. If just making sure they build OK is fine I can add a commit for these too
- Disabled `ZMK_USB` in peripherals (didn't cause a build failure because it was forced off by its dependencies)


petejohanson:
@caksoylar Mind including the header removal for the others? As long as it still builds (which it should), that should be fine.

caksoylar:
Sure, will push it when I am home in the evening.

caksoylar:
Pushed those too now, `corne_left` and right builds with OLED or nice!views seem to build OK here.

petejohanson (approved):
Thanks!

Issue #2132, opened 2024-01-25
OLEDs do not fit into the nRF52833 flash anymore


krikun98:
Ever since `b021d192` OLED builds do not fit into `nrfmicro_13_52833`:
```
west update && west build -p -b nrfmicro_13_52833 -d build/jorne_test/left -- -DSHIELD=jorne_left
...
/home/screamingwhale/.local/zephyr-sdk-0.15.0/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.1.0/../../../../arm-zephyr-eabi/bin/ld.bfd: zephyr/zephyr_pre0.elf section `datas' will not fit in region `FLASH'
/home/screamingwhale/.local/zephyr-sdk-0.15.0/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.1.0/../../../../arm-zephyr-eabi/bin/ld.bfd: region `FLASH' overflowed by 4388 bytes
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```
Full logs are available here: https://github.com/krikun98/zmk-config/actions/runs/7005103094/job/19054193000
Builds with `CONFIG_ZMK_DISPLAY=y` for the standard SSD1306 fail to fit into the flash.
If I revert that change:
```
diff --git a/app/src/display/Kconfig b/app/src/display/Kconfig
index a2029481..d419300b 100644
--- a/app/src/display/Kconfig
+++ b/app/src/display/Kconfig
@@ -41,7 +41,6 @@ choice ZMK_DISPLAY_STATUS_SCREEN
 config ZMK_DISPLAY_STATUS_SCREEN_BUILT_IN
     bool "Built in status screen"
     select LV_OBJ_LABEL
-    imply LV_USE_THEME_MONO
     imply ZMK_WIDGET_LAYER_STATUS
     imply ZMK_WIDGET_BATTERY_STATUS
     imply ZMK_WIDGET_OUTPUT_STATUS
```
The firmware builds fine on the current `main` branch.

joelspadin:
`imply LV_USE_THEME_MONO` means that the config isn't forced. If you set `CONFIG_LV_USE_THEME_MONO=n` in your .conf file, does that build again?

krikun98:
@joelspadin, it does, yes.

joelspadin:
[...]

Issue #1848, opened 2023-06-24
corne: left display not working


gnapiorkowski:
Hello,
Recently I've rebased my local zmk repo to update my keyboard and it looks like left corne ( zmk/tree/main/app/boards/shields/corne ) display isn't working anymore (it used to work a while ago: sometime around Jan 6 it looks like: https://github.com/gnapiorkowski/zmk-config/commit/d3375828a459dd391a2f750d9a5f60329a3494e5). tested with a fresh corne-config with just the display line uncommented: https://github.com/gnapiorkowski/zmk-config-corne/actions/runs/5366593277
I can't figure out what caused it, because there has been quite a few commits since last time (January 6) I have changed anything.

shawnbitzaro:
I'm facing this too and was about to post an issue.

caksoylar:
Since it hasn't been reported more generally, it could be a `nrfmicro_13` specific issue due to the large Zephyr 3.2 upgrade. 

@shawnbitzaro do you also use a nrfmicro? 

Does it work OK for the right half both of you? Can you make sure that external power is turned on through the [`&ext_power EP_ON`](https://zmk.dev/docs/behaviors/power) binding, by pressing it, waiting one minute for the setting to be saved to flash, then resetting the keyboard (or turning it off then on)?

lsshawn:
I'm using nrfmicro. Right side works for me.

The weird thing with me is if I set `CONFIG_ZMK_DISPLAY=y`, my left side fails to flash. I drag the files into it and it doesn't unmount.

But on the right side, it flashes ok and display shows up.

Config here: https://github.com/lsshawn/corne-zmk

gnapiorkowski:
Same here, I have two identical cornes, one with nrfmicros and one with nice!nanos. I use the same config for both. Nice!nano one works as expected, but for the nrfmicro the left side oled display doesn't work (right side is fine). Checked the display, checked the connections... Display is getting 3.3V.
Config here:  https://github.com/gnapiorkowski/zmk-config-corne

PR #2152, opened 2024-02-07
Fix nrfMicro oled display

ton-im:
After Apr 6 2023(ae8299edb3d638f1332475b1da0fdf40afa43fe4), the nrfMicro can not driver oled, because of missing a line.

### DIFF
--- a/app/boards/arm/nrfmicro/nrfmicro_11.dts
+++ b/app/boards/arm/nrfmicro/nrfmicro_11.dts
@@ -31,6 +31,7 @@
     EXT_POWER {
         compatible = "zmk,ext-power-generic";
         control-gpios = <&gpio1 9 GPIO_ACTIVE_HIGH>;
+        init-delay-ms = <50>;
     };
 };
 
--- a/app/boards/arm/nrfmicro/nrfmicro_11_flipped.dts
+++ b/app/boards/arm/nrfmicro/nrfmicro_11_flipped.dts
@@ -31,6 +31,7 @@
     EXT_POWER {
         compatible = "zmk,ext-power-generic";
         control-gpios = <&gpio1 9 GPIO_ACTIVE_HIGH>;
+        init-delay-ms = <50>;
     };
 };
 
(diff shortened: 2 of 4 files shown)


ton-im:
@lesshonor 
> The commit you've linked to does not touch any files related to the nRFMicro board.
Modified files list:
app/boards/arm/nrfmicro/nrfmicro_11.dts
app/boards/arm/nrfmicro/nrfmicro_11_flipped.dts
app/boards/arm/nrfmicro/nrfmicro_13.dts
app/boards/arm/nrfmicro/nrfmicro_13_52833.dts

>  Are you sure you're not running something related to, say, #2132?
It is not about rom size. It is about some ssd1306 can not be driven.

ton-im:
@lesshonor 
ae8299e commits on Mar 26 2023 ，the next commits on Apr 6 2023  make oled can not driven.

#2132 can use sdk 0.13 on Zephyr 3.2 build well on my config . nRF52833 's 512 KB Flash can not include too much modules . oled is the straw that broke the camel's back.

caksoylar:
Maybe this is related to #1848, that one started happening after the Zephyr 3.2 upgrade. That sounds consistent with the description "https://github.com/zmkfirmware/zmk/commit/ae8299edb3d638f1332475b1da0fdf40afa43fe4 commits on Mar 26 2023 ，the next commits on Apr 6 2023 make oled can not driven."

petejohanson (approved):
Ok, so there is definitely a distinct likelyhood this would help on nrfMicro depending on the specifics of how it was built. That design uses a MOSFET to control the VCC on/off, 1.4 schematic states it uses an AO3407 that has the following in its datasheet:

image

So that's 13.6ms from toggling the power on until the OLED actually is powered, plus any time the OLED itself needs to actually be ready to be initialized. That assumes the original part, AO3407 is used, and not some alternative that may have slower turn-on time.

This brings the setup in here in line with other controller board DTS that add init delays for the same reason: without it the firmware may try to initialize a peripheral before it is actually powered on and ready to be initialized.

Thanks for the fix!

PR #1923, opened 2023-09-04
feat(sensors): Battery charging status detection

ReFil:
Adds the ability to detect if the battery is charging if the charger is connected to a GPIO on the microcontroller

Can tie into #1456 for an all encompassing battery state underglow effect.
the chg-gpios parameter is entirely optional so shouldn't break existing boards/configs and hasn't so far in my testing

PR #2153, opened 2024-02-09
feat(battery): decouple battery charging status from USB connection state

zhiayang:
reference/prior work: #1923 

this should be the first phase in a set of PRs to decouple the status of battery charging from the device being connected to USB. the end goal is to allow non-USB-based charging to be reflected in ZMK, eg. via (as in #1923) a GPIO pin connected to a charge IC, or (in future) reading from a fuel gauge IC.

this first PR adds the `zmk_battery_is_charging` function, which for now simply falls back to detecting USB. It also updates places that were using USB as a proxy to use this new function instead.

the RGB and underglow config flags currently explicitly name USB (eg . `CONFIG_ZMK_BACKLIGHT_AUTO_OFF_USB`), so I've opted not to change them -- at least for now, and they still use USB detection.

The display widgets (corne-ish zen, nice_view, generic zmk display) have been updated to use `zmk_battery_is_charging` as well.

### DIFF
--- a/app/src/activity.c
+++ b/app/src/activity.c
@@ -20,6 +20,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/events/position_state_changed.h>
 #include <zmk/events/sensor_event.h>
 
+#include <zmk/battery.h>
 #include <zmk/activity.h>
 
 #if IS_ENABLED(CONFIG_USB_DEVICE_STACK)
@@ -126,7 +127,7 @@ void activity_work_handler(struct k_work *work) {
     int32_t current = k_uptime_get();
     int32_t inactive_time = current - activity_last_uptime;
 #if IS_ENABLED(CONFIG_ZMK_SLEEP)
-    if (inactive_time > MAX_SLEEP_MS && !is_usb_power_present()) {
+    if (inactive_time > MAX_SLEEP_MS && !(is_usb_power_present() || zmk_battery_is_charging())) {
         // Put devices in suspend power mode before sleeping
         set_state(ZMK_ACTIVITY_SLEEP);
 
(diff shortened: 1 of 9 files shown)


zhiayang:
tested on real hardware: default display widget (not niceview)

existing behaviour seems to work just fine, charging icon appears (immediately) when USB goes in, disappears (immediately) when USB goes out.

ReFil:
I think it makes sense for RGB and backlighting to use USB power and not charging for their auto off triggers as otherwise the RGB and backlighting might randomly change state whilst the board is plugged in when the battery completes charging

app/src/activity.c
```diff
@@ -19,18 +19,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/events/sensor_event.h>
 
 #include <zmk/activity.h>
-
-#if IS_ENABLED(CONFIG_USB_DEVICE_STACK)
-#include <zmk/usb.h>
-#endif
-
-bool is_usb_power_present(void) {
-#if IS_ENABLED(CONFIG_USB_DEVICE_STACK)
```
ReFil, line 28: Why is this being removed? It can be used in conjunction with charging detection to detect when it's fully charged. If USB power is present but board is not charging then it can be inferred that the board is fully charged

zhiayang: you're right! didn't think of that. I only removed it because nobody was calling it (and it wasn't declared in a header anywhere LOL). now that you mention it, for the usecase i have in mind (charging but not via usb), i also want this distinction.

so i feel like we would want `is_charging` *and* `is_externally_powered` statuses?

zhiayang: edit: will probably wait for the zephyr 3.5 update and #2154 to be merged before continuing on this

zhiayang:
weirdge, it closed the pr.

app/src/activity.c
zhiayang: okay, i ended up removing this for a few reasons:
1. `zmk_is_externally_powered` subsumes this functionality, and covers both the battery-charging case and the usb-plugged-in case
2. nobody was actually calling this, and it wasn't declared in a header anywhere

for detecting fully-charged, you would be able to do `zmk_is_externally_powered() && !zmk_battery_is_charging()`.

PR #2154, opened 2024-02-09
fix(displays, battery): fix initial display of battery status on displays

zhiayang:
Fix the issue where the battery status on displays would not be correct until *after* at least one update interval (typically 1 minute) had passed. This was caused by the display subsystem being initialised after the first battery event is raised.

Also fixed a null dereference issue in the default display battery widget.


petejohanson (approved):
All seems very reasonable. Thanks!

PR #2155, opened 2024-02-09
feat(boards): add encoder support to planck

SethMilliken:
This adds support for an optional rotary encoder in the leftmost column of the Planck rev6. All positions for the encoder map to the same pins on the board, so this ought to work regardless of which position is chosen.

I have been using this regularly for the last week without any issues. FWIW, my encoder is in the bottom row.

### DIFF
--- a/app/boards/arm/planck/planck_rev6.dts
+++ b/app/boards/arm/planck/planck_rev6.dts
@@ -44,6 +44,20 @@
             ;
     };
 
+    encoder: encoder {
+        compatible = "alps,ec11";
+        a-gpios = <&gpiob 12 GPIO_PULL_UP>;
+        b-gpios = <&gpiob 13 GPIO_PULL_UP>;
+        steps = <80>;
+        status = "disabled";
+    };
+
+    sensors: sensors {
+        compatible = "zmk,keymap-sensors";
+        sensors = <&encoder>;
+        triggers-per-rotation = <20>;
+    };
+
 layout_grid_transform:
     keymap_transform_0 {
         compatible = "zmk,matrix-transform";
--- a/app/boards/arm/planck/planck_rev6.keymap
+++ b/app/boards/arm/planck/planck_rev6.keymap
@@ -7,6 +7,11 @@
 #include <behaviors.dtsi>
 #include <dt-bindings/zmk/keys.h>
 
+/* Uncomment this block if using an encoder */
+//&encoder {
+//    status = "okay";
+//};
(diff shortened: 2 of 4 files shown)


app/boards/arm/planck/planck_rev6.dts
```diff
@@ -44,6 +44,20 @@
             ;
     };
 
+    left_encoder: left_encoder {
+        compatible = "alps,ec11";
+        a-gpios = <&gpiob 12 GPIO_PULL_UP>;
+        b-gpios = <&gpiob 13 GPIO_PULL_UP>;
+        steps = <80>;
+        status = "okay";
```
SethMilliken, line 52: There did not appear to be a clear precedent for setting the status of an encoder on a unibody board. Most boards set this to "disabled" in the dts and then set the respective encoder statuses to "okay" in the overlays for each half. Let me know if there is a better way to handle this.

petejohanson: Ideally you'd have it disabled, with a commented out section in the *keymap* to enable the encoder node if you're using it.

SethMilliken: > Ideally you'd have it disabled, with a commented out section in the _keymap_ to enable the encoder node if you're using it.

Cool. Will do.

SethMilliken: Encoder is now disabled by default with commented-out node in keymap to enable it.

app/boards/arm/planck/planck_rev6.dts
```diff
@@ -44,6 +44,20 @@
             ;
     };
 
+    left_encoder: left_encoder {
```
joelspadin, line 47: If the keyboard only supports one encoder, should we just call it "encoder"?

SethMilliken: I had gone with that originally, but rev7 supports two encoders, one on each side. If we eventually support that, it seems like this would be a better fit if they share config. I can switch it back if you like though.

petejohanson: Whenever we do that, we can do that in a new `planck_rev7.dts`, etc. For now, keep this relevant to the rev6, so I'd suggest renaming back to just `encoder`.

SethMilliken: Aye, aye. Renamed to `encoder`.

PR #2159, opened 2024-02-10
feat(boards): allow use of blackpill_f401cc

lesshonor:
Zephyr's `blackpill_f401ce` and `blackpill_f411ce` board definitions were enabled for use in ZMK in 70aa4fa. I can think of no good reason why all three should not be available options, particularly as a lot of clones tend to use F401CC.


app/boards/blackpill_f401cc.conf
```diff
@@ -0,0 +1,7 @@
+CONFIG_CONSOLE=n
+CONFIG_SERIAL=n
+CONFIG_UART_CONSOLE=n
+CONFIG_UART_INTERRUPT_DRIVEN=n
+CONFIG_ZMK_USB=y
+CONFIG_ZMK_BLE=n
```
petejohanson, line 6: This is probably superfluous? If preset in the others I would suggest removing them.

app/boards/blackpill_f401cc.overlay
```diff
@@ -0,0 +1,50 @@
...
+            , <43 0 &gpiob 7  0>    /* PB7  */
+            , <45 0 &gpiob 8  0>    /* PB8  */
+            , <46 0 &gpiob 9  0>    /* PB9  */
+            ;
+    };
+};
+
+blackpill_i2c: &i2c1 {};
+blackpill_spi: &spi1 {};
+blackpill_serial: &usart1 {};
```
petejohanson, line 50: Are these peripherals already set up properly in the base board definition I assume?

petejohanson (commented):
Fully agreed. Just a couple comments.

petejohanson (approved):
Thanks!

PR #2161, opened 2024-02-12
refactor: remove redundant Kconfig defaults

lesshonor:
As discussed [on Discord](https://discord.com/channels/719497620560543766/1206391546090496060/1206398822675124254):
> **petejohanson:** We shouldn't ever explicitly set a default n
**petejohanson:** Because it causes the problem you describe
**petejohanson:** https://docs.zephyrproject.org/latest/build/kconfig/tips.html#redundant-defaults

Only ZMK symbols were modified; [`BT_GATT_NOTIFY_MULTIPLE`](https://github.com/zmkfirmware/zephyr/blob/1ae0eb5ce8adafcec993e6fb8f4eeb6f818a7772/subsys/bluetooth/host/Kconfig.gatt#L110) was intentionally omitted. While seemingly unlikely given the age of `git blame`...it's still *possible* that the default value of this symbol could be changed at some later date in Zephyr, so I left it alone.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -177,7 +177,6 @@ config ZMK_BLE_EXPERIMENTAL_FEATURES
 
 config ZMK_BLE_PASSKEY_ENTRY
     bool "Require passkey entry on the keyboard to complete pairing"
-    default n
     select RING_BUFFER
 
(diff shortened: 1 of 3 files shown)


petejohanson (approved):
Thanks!

PR #2164, opened 2024-02-14
fix(build): Used a fixed temp dir for caching.

petejohanson:
* Improve GHA caching by using a fixed temporary directory when using one.

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -68,7 +68,8 @@ jobs:
         run: |
           if [ -e zephyr/module.yml ]; then
             export zmk_load_arg=" -DZMK_EXTRA_MODULES='${GITHUB_WORKSPACE}'"
-            export new_tmp_dir=$(mktemp -d)
+            new_tmp_dir="${TMPDIR:-/tmp}/zmk-config"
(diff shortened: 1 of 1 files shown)


petejohanson:
Tested out and caching working using this branch: https://github.com/petejohanson/le-capybara-zmk-config/actions/runs/7906598751/job/21582180373

.github/workflows/build-user-config.yml
```diff
@@ -68,7 +68,8 @@ jobs:
         run: |
           if [ -e zephyr/module.yml ]; then
             export zmk_load_arg=" -DZMK_EXTRA_MODULES='${GITHUB_WORKSPACE}'"
-            export new_tmp_dir=$(mktemp -d)
+            export new_tmp_dir="${TMPDIR:-/tmp}/zmk-config"
```
caksoylar, line 71: Nit: I guess we don't need to `export` this variable, do we?

caksoylar (approved):
LGTM, noted a nit.

.github/workflows/build-user-config.yml
petejohanson: Nope. Not needed. Fixed, and re-ran the workflow with no error.

PR #2168, opened 2024-02-18
fix(boards): Tweaks for Ferris rev0.2 for Zephyr.

petejohanson:
* Enable missing clock and set up USB pinctrl.


dhoizner:
verifying that with this fix, Ferris rev0.2 works again :D

petejohanson:
Tested working by dandon on Discord.

PR #2177, opened 2024-02-27
feat(docs): Add pointer to shield folder in new shield guide

caksoylar:
- Add a tip that links to ZMK shields folder for users to use as starting points
  - I think it's a good idea, but I am open to discussion if anyone has concerns.
- Add some comments in the matrix transform example to look less like a complete overlay, but also add the closing brace just in case.


caksoylar:
Added a new fix commit to add `wakeup-source;` to the split example in new shield as well, which was initially missed.

petejohanson (approved):
Thanks

PR #2178, opened 2024-02-27
feat(docs): Document HID report descriptor refresh

caksoylar:
Added a section under the BT features page that describes when it might be necessary, and how to do it. This will be a more common ask with more mouse functionality getting integrated in the future.

Also took the opportunity to remove the guidance on Windows battery refresh Kconfig. This doesn't seem to be necessary anymore in W10/W11, according to reports.


docs/docs/features/bluetooth.md
```diff
@@ -36,6 +36,19 @@ Failure to manage the profiles can result in unexpected/broken behavior with hos
 
 Management of the bluetooth in ZMK is accomplished using the [`&bt` behavior](../behaviors/bluetooth.md). Be sure to refer to that documentation to learn how to manage profiles, switch between connected hosts, etc.
 
+## Refreshing the HID Descriptor
+
+Enabling certain features or behaviors in ZMK changes the data structure that ZMK sends over USB or BLE to host devices.
+This in turn requires [HID report descriptors](https://docs.kernel.org/hid/hidintro.html) to be modified for the reports to be parsed correctly.
```
caksoylar, line 42: An alternative reference here is https://en.wikipedia.org/wiki/Human_interface_device#Report_Descriptor. But the kernel page seems to be a bit more descriptive.

PR #549, opened 2020-12-29
Update docs to include macOS/Windows dual boot workaround

tominabox1:
This is a workaround to allow you to use a bluetooth device on macOS and Windows from a single keyboard profile. I have tested this on Big Sur and Windows 10 successfully. This is distilled from various reference sources but primarily [insanelymac](https://www.insanelymac.com/forum/topic/268837-dual-boot-bluetooth-pairing-solved/)

See the updated page here: https://deploy-preview-549--zmk.netlify.app/docs/troubleshooting/

PR #2179, opened 2024-02-28
feat(docs): Add a note on using BT with dual boot systems

caksoylar:
Added a remark noting issues with pairing in dual boot systems, any suggestions to improve technical correctness would be appreciated!

I wasn't around for #549 but I prefer this more concise approach, where we wouldn't have to maintain or take responsibility for an extended guide. This still remarks that the issue exists and hints at (unofficial) workarounds.


myst729:
This is so epic 🤣

petejohanson (approved):

PR #2180, opened 2024-02-28
fix(docs): Add Underglow Brightness Min/Max to Docs

awkannan:
I'm not sure if these exist somewhere else, but I couldn't find them.

Adding them here because I think it's appropriate.

Feel free to close if I am off base!


caksoylar (approved):
Good catch, thanks!

Issue #1416, opened 2022-08-01
Confusing error message when keymap node is missing


joelspadin:
If there is no keymap node (or you've misspelled the `compatible` property for your keymap node), you get a few dozen lines of error messages deep in the evaluation of the `ZMK_KEYMAP_LAYERS_LEN` macro. Ideally we should verify that there is a keymap before using that, and `#error` with a more useful message that tells you to check that you have a keymap and that it has `compatible = "zmk,keymap"` set properly.

joelspadin:
Looks like this was fixed a while back.

PR #2181, opened 2024-02-29
feat(build): Explicit keymap node error

ReFil:
Little fix to address #1416 to throw an explicit error if a node with `compatible = zmk,keymap` is not set. In testing, the `ZMK_KEYMAP_LAYERS_LEN` error is still present but the explicit error gets presented as well, which should still help users diagnose problems.

Some documentation on the error has also been added to the troubleshooting section of the docs.

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -40,6 +40,10 @@ West build errors usually indicate syntax problems in the `<keyboard>.keymap` fi
 If you are reviewing these errors in the GitHub Actions tab, they can be found in the `West Build` step of the build process.
 :::
 
+#### Keymap error
+
+If you get an error stating `Keymap node not found, check a keymap is available and is has compatible = "zmk,keymap" set` this is an indication that the build process cannot find the keymap. Double check that the `<keyboard>.keymap` file is present and has been discovered by the build process. This can be checked by looking for a line in the build log stating `-- Using keymap file: /path/to/keymap/file/<keyboard>.keymap`. Inside the keymap file ensure the keymap node has `compatible = zmk,keymap` and it's not misspelled. For more information see the [Keymap](features/keymaps.mdx) documentation.
+
 #### devicetree error
 
 A `devicetree error` followed by a reference to the line number on `<keyboard>.keymap` refers to an issue at the exact line position in that file. For example, below error message indicates a missing `;` at line 109 of the `cradio.keymap` file:
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -42,7 +42,7 @@ If you are reviewing these errors in the GitHub Actions tab, they can be found i
 
 #### Keymap error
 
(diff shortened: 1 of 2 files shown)


docs/docs/troubleshooting.md
```diff
@@ -40,6 +40,10 @@ West build errors usually indicate syntax problems in the `<keyboard>.keymap` fi
 If you are reviewing these errors in the GitHub Actions tab, they can be found in the `West Build` step of the build process.
 :::
 
+#### Keymap error
+
+If you get an error stating `Keymap node not found, check a keymap is available and is has compatible = "zmk,keymap" set` this is an indication that the build process cannot find the keymap. Double check that the `<keyboard>.keymap` file is present and has been discovered by the build process. This can be checked by looking for a line in the build log stating `-- Using keymap file: /path/to/keymap/file/<keyboard>.keymap`. Inside the keymap file ensure the keymap node has `compatible = zmk,keymap` and it's not misspelled. For more information see the [Keymap](features/keymaps.mdx) documentation.
```
caksoylar, line 45: If the main reason we think users get this is that the keymap file isn't found, we can also link to https://zmk.dev/docs/config in addition to the Keymap page.

ReFil: Added a link to the config documentation in addition to the keymap link that's currently there

docs/docs/troubleshooting.md
```diff
@@ -40,6 +40,10 @@ West build errors usually indicate syntax problems in the `<keyboard>.keymap` fi
 If you are reviewing these errors in the GitHub Actions tab, they can be found in the `West Build` step of the build process.
 :::
 
+#### Keymap error
+
+If you get an error stating `Keymap node not found, check a keymap is available and is has compatible = "zmk,keymap" set` this is an indication that the build process cannot find the keymap. Double check that the `<keyboard>.keymap` file is present and has been discovered by the build process. This can be checked by looking for a line in the build log stating `-- Using keymap file: /path/to/keymap/file/<keyboard>.keymap`. Inside the keymap file ensure the keymap node has `compatible = zmk,keymap` and it's not misspelled. For more information see the [Keymap](features/keymaps.mdx) and [Config](config) documentation.
```
caksoylar, line 45:
```suggestion
If you get an error stating `Keymap node not found, check a keymap is available and is has compatible = "zmk,keymap" set` this is an indication that the build process cannot find the keymap. Double check that the `<keyboard>.keymap` file is present and has been discovered by the build process. This can be checked by looking for a line in the build log stating `-- Using keymap file: /path/to/keymap/file/<keyboard>.keymap`. Inside the keymap file ensure the keymap node has `compatible = zmk,keymap` and it's not misspelled. For more information see the [Keymap](features/keymaps.mdx) and [Config](config/index.md) documentation.
```
Nit, we prefer file path linking as convention

ReFil: Ah ok, wasn't aware that was how the categories worked in the docs, will sort

caksoylar (commented):
Thanks! LGTM on the docs side.

petejohanson (approved):
Thanks!

PR #2185, opened 2024-03-03
feat(docs): Add warning to new shield docs how Kconfig treats whitspaces

### LINKED ISSUE #2183: Documentation: more warning for Kconfig.shield
Hello!

I've just created a new keyboard shield based on the documentation: https://zmk.dev/docs/development/new-shield

I have written an extra space character between "shields_list_contains," and "my_board":

```
config SHIELD_MY_BOARD
    def_bool $(shields_list_contains, my_board)
```

The valid format is without space:

```
config SHIELD_MY_BOARD
    def_bool $(shields_list_contains,my_board)
```

The caused error was that the board does not advertised itself over bluetooth. After 2 hours of debugging and reading solved issues in the discord server's history I managed to fix the issue. The documentation does warn about that "my_board" string should be set properly, but as a person not familiar with writing Kconfig files I went over it at least 5 times (and the error is hard to find visually).

I think other people will mess this up in the future, so I suggest to add a warning or a sentence in the documentation that there should be absolutely no space where I added.

huber-th:
Resolves #2183

Added a warning to the shield section explaining that Kconfig does not ignore whitespaces on function calls and therefore adding whitespaces after the comma will break functionality.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,12 @@ documentation to areas not currently covered are greatly appreciated.
 ZMK uses `prettier` to format documentation files. You can run prettier with `npm run prettier:format`.
 You can setup git to run prettier automatically when you commit by installing the pre-commit hooks: `pip3 install pre-commit`, `pre-commit install`.
 
+### Linting
+
+This repository utilizes ESLint for code linting to ensure consistent code style and identify potential errors or bugs early in the development process.
+
+You can run ESLint with `npm run lint` to verify your changes.
+
 ## Code Contributions
 
 ### Development Setup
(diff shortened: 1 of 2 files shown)


caksoylar (approved):
That seems perfect, thank you!

caksoylar:
Unfortunately eslint seems to be unhappy, given the message I assume it might be related to the leading space example? I can't dig into it right now since I am traveling, but if you figure out a way to resolve it it would be great.

huber-th:
I fixed the issue. I switched to using `<code></code>` instead of backticks for the example in order to being able to use HTML tag for the whitespace instead of a non-regular whitespace. Markdown trims leading and trailing whitespaces by default.

I added a second commit updating the `CONTRIBUTING.md` to mention the usage of ESLint for changes to the documentation.

Let me know in case that shouldn't be done as part of this PR but rather separately.

caksoylar:
Thanks, the new commit is fine in this PR IMO. I have to admit I have been forgetting to update contributing.md, but linting is documented in https://zmk.dev/docs/development/documentation#formatting-and-linting-your-changes.

caksoylar (approved):

PR #2188, opened 2024-03-07
feat(ble): allow dongle without Bluetooth profiles

xudongzheng:
Allow setting CONFIG_BT_MAX_CONN to CONFIG_ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS for dongle mode without Bluetooth profiles (USB output only).

### DIFF
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -162,6 +162,13 @@ bool zmk_ble_active_profile_is_connected(void) {
     advertising_status = ZMK_ADV_CONN;
 
 int update_advertising(void) {
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
+    if (CONFIG_ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS == CONFIG_BT_MAX_CONN) {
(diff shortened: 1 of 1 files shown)


ayroblu (approved):
I tested and verified this works
https://github.com/ayroblu/zmk-config/compare/fc9009595...ce2691ac
My keyboard types fine, and I no longer see my dongle advertising in bluetooth settings

xudongzheng:
This PR has been reworked such that profile advertising can be disabled with `CONFIG_ZMK_BLE_PROFILE_ADVERTISING=n`.

This makes configuration slightly less error prone as the user no longer has to make changes to `CONFIG_BT_MAX_CONN`, since the value ends up being different for dongle and dongleless.

Additionally, some parts of ZMK make the assumption that the number of profiles is non-zero. Having `CONFIG_ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS` set to `CONFIG_BT_MAX_CONN` can cause ZMK to behave unexpectedly as a result of that.

sigvebh:
Can verify that this works for me. Been using it for a while now.

PR #2189, opened 2024-03-07
feat(docs): Add a note on macOS v14.3 file copy error message

filterpaper:
The Finder error is reported by the current version of macOS Sonoma but it is innocuous.


docs/docs/troubleshooting.md
```diff
@@ -27,6 +27,10 @@ Variations of the warnings shown below occur when flashing the `<firmware>.uf2`
 
 macOS 13.0 (Ventura) Finder may report an error code 100093 when copying `<firmware>.uf2` files into microcontrollers. This bug is limited to the operating system's Finder. You can work around it by copying on Terminal command line or use a third party file manager. Issue is fixed in macOS version 13.1.
 
+### macOS Sonoma error
+
+macOS 14.3.1 (Sonoma) Finder may report an "error code -36" when copying `<firmware>.uf2` files into microcontrollers. A similar "fcopyfile failed: Input/output error" will also be reported when copying is performed using Terminal command line. These errors can be ignored because they are reported when the bootloader disconnects automatically after the uf2 file is copied successfully.
```
caksoylar, line 32:
```suggestion
macOS 14.3.1 (Sonoma) Finder may report an "Error code -36" when copying `<firmware>.uf2` files into microcontrollers. A similar "fcopyfile failed: Input/output error" will also be reported when copying is performed using Terminal command line. These errors can be ignored because they are reported when the bootloader disconnects automatically after the uf2 file is copied successfully.
```

caksoylar (commented):
Thank you for taking the time to add this, it is becoming frequently reported.

There are some reports that flashing actually fails (using the same error code) when the filenames are too long (like >256 chars). Since you have the hardware to test, can I ask you to confirm this? Because if that is the case we should document that caveat as well.

huber-th:
> Thank you for taking the time to add this, it is becoming frequently reported.
> 
> There are some reports that flashing actually fails (using the same error code) when the filenames are too long (like >256 chars). Since you have the hardware to test, can I ask you to confirm this? Because if that is the case we should document that caveat as well.

Filenames are limited to 255 characters on macOS. I tested this on a MacBook Air M2 running macOS Sonoma 14.4 and cannot confirm that flashing fails.

I did however encounter something odd and somewhat unexpected. I was shown the `Error code -36` for file names (not counting the extension `.uf2`) of length 253 characters or less. 

For file names of length 254 and 255 characters I don't get the error code. Flashing succeeds as expected and the drive disconnects with the firmware flashed and good to go.

filterpaper:
> Thank you for taking the time to add this, it is becoming frequently reported.
> 
> There are some reports that flashing actually fails (using the same error code) when the filenames are too long (like >256 chars). Since you have the hardware to test, can I ask you to confirm this? Because if that is the case we should document that caveat as well.

APFS file name constraint is 255 characters — is there a board and shield combination that will generate a long file name under normal circumstances?

caksoylar:
Thanks both, the reports I have for not flashing are from Glove80 users -- the layout editor at https://my.glove80.com exports firmware files with a GUID and layout name, like `2a774be9-eee1-4513-8583-7be1e84de402_v24.02_Stock layout.uf2`.

Digging a bit more, [one of the announcements](https://discord.com/channels/877392805654306816/1068659814529511505/1194534488500355112) mentions error code -43 instead of -36:
> NOTE: If you are not able to flash the firmware successfully on macOS, with an error code of “-43”, please first rename the UF2 file to a shorter name, while keeping the same extension (.uf2). macOS throws this error if the filename exceeds its length limit.

So in this case it seems safe to call -36 harmless.

caksoylar (approved):

PR #2195, opened 2024-03-08
fix(build): Properly board revision keymaps.

petejohanson:
* Handle board keymap location for boards with Zephyr board revisions included.
* Includes bare non-revision file and a revision specific keymap in case newer revision changes the layout/key positions.

### DIFF
--- a/app/keymap-module/modules/modules.cmake
+++ b/app/keymap-module/modules/modules.cmake
@@ -63,6 +63,13 @@ if(DEFINED SHIELD)
     string(REPLACE " " ";" SHIELD_AS_LIST "${SHIELD}")
 endif()
 
+string(FIND "${BOARD}" "@" REVISION_SEPARATOR_INDEX)
+if(NOT (REVISION_SEPARATOR_INDEX EQUAL -1))
+    math(EXPR BOARD_REVISION_INDEX "${REVISION_SEPARATOR_INDEX} + 1")
+    string(SUBSTRING ${BOARD} ${BOARD_REVISION_INDEX} -1 BOARD_REVISION)
+    string(SUBSTRING ${BOARD} 0 ${REVISION_SEPARATOR_INDEX} BOARD)
+endif()
+
 foreach(root ${BOARD_ROOT})
     set(shield_dir ${root}/boards/shields)
     # Match the Kconfig.shield files in the shield directories to make sure we are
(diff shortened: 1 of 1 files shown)


app/keymap-module/modules/modules.cmake
```diff
@@ -107,6 +114,21 @@ foreach(root ${BOARD_ROOT})
...
+    # Board provides revision handling.
+    include(${BOARD_DIR}/revision.cmake)
+elseif(BOARD_REVISION)
+    message(WARNING "Board revision ${BOARD_REVISION} specified for ${BOARD}, \
+                     but board has no revision so revision will be ignored.")
+endif()
+
+if(DEFINED BOARD_REVISION)
+    string(REPLACE "." "_" BOARD_REVISION_STRING ${BOARD_REVISION})
+    set(KEYMAP_BOARD_REVESION_PREFIX "${BOARD}_${BOARD_REVISION_STRING}")
```
elpekenin, line 127: Typo (and below) 😅 
```suggestion
    set(KEYMAP_BOARD_REVISION_PREFIX "${BOARD}_${BOARD_REVISION_STRING}")
```

app/keymap-module/modules/modules.cmake
```diff
@@ -107,6 +114,21 @@ foreach(root ${BOARD_ROOT})
     endif()
 endforeach()
 
+if(EXISTS ${BOARD_DIR}/revision.cmake)
+    # Board provides revision handling.
+    include(${BOARD_DIR}/revision.cmake)
+elseif(BOARD_REVISION)
+    message(WARNING "Board revision ${BOARD_REVISION} specified for ${BOARD}, \
+                     but board has no revision so revision will be ignored.")
```
elpekenin, line 122: This wording is a bit confusing imo.  My suggestion might not be much better tho...
```suggestion
                     but board has no revision. Setting will be ignored.")
```

elpekenin (changes requested):

app/keymap-module/modules/modules.cmake
petejohanson: Oh man, did this in my commit message and thought "Man, glad I caught that!".

Thanks.

app/keymap-module/modules/modules.cmake
petejohanson: This piece is copied verbatim from the equivalent code in Zephyr, so I'd favor keeping it so searching later more easily correlates things.

petejohanson:
I've tested this with the following Technikable board directory structure:

```
board.cmake
Kconfig
Kconfig.board
Kconfig.defconfig
revision.cmake
technikable_1_0_0.conf
technikable_1_0_0.overlay
technikable_1_1_0.conf
technikable_1_1_0.keymap
technikable_1_1_0.overlay
technikable_defconfig
technikable.dts
technikable.keymap
```

Where `revision.cmake` has:

```
board_check_revision(FORMAT MAJOR.MINOR.PATCH
                     DEFAULT_REVISION 1.0.0
                     VALID_REVISIONS 1.0.0 1.1.0)
```

I can build with `-b technikable` for the existing board target (thanks to default revision of `1.0.0`, or build for `-b technikable@1.1` to pull in the new version and it will find a proper keymap as expected.

app/keymap-module/modules/modules.cmake
elpekenin: Sounds like a great idea \:)

Issue #2198, opened 2024-03-09
[Feature Request]  Option to disable sleep while bluetooth connection is active


aroum:
The ZMK checks the USB connection before starting the sleep timer. I would like to be able to check the BLE connections.

Now the keyboard falls asleep when I watch a video and I have to wake it up every time to pause the video.

I increased the time before falling asleep to an hour. Now, after turning off the PC, the keyboard still tries to connect to it for a long time.

I would like the keyboard to fall asleep 5 minutes after losing connection to the PC.

petejohanson:
I think this is a really nice idea as a tweak or optional change to the current sleep code. Some folks may still prefer the current behavior, so probably work putting behavior a new Kconfig flag.

newx-it:
Hello! I submitted a PR to solve this issue and also add another related flag/feature. Please let me know how it looks. 

[pull 3286](https://github.com/zmkfirmware/zmk/pull/3286)

newx-it:
I've closed PR 3286 in favor of another PR with cleaned up commit history and a couple other fixes, It's available here: [PR 3288](https://github.com/zmkfirmware/zmk/pull/3288).

Apologies for the complication!

PR #2205, opened 2024-03-15
feat: stay awake while BLE connected

elpekenin:
Solves* #2198 by adding a new Kconfig option and its corresponding change on the "shall we sleep?" check.

\* Should solve, but i cant test as i dont have a battery-powered device (:

Note: May run into issues with the refactoring of this file on #2085


app/src/activity.c
```diff
@@ -126,7 +130,14 @@ void activity_work_handler(struct k_work *work) {
     int32_t current = k_uptime_get();
     int32_t inactive_time = current - activity_last_uptime;
 #if IS_ENABLED(CONFIG_ZMK_SLEEP)
-    if (inactive_time > MAX_SLEEP_MS && !is_usb_power_present()) {
+    if (inactive_time > MAX_SLEEP_MS &&
+        !is_usb_power_present()
```
elpekenin, line 134: I hate this formatting with the first `&&` on the same line but the second one (`#if`) being on the next one...

Not consistent, but it is what clangd-format wanted

caksoylar:
I think there is a question here whether this should check only the current profile's connection status (this implementation) or any of the profiles. If the former is preferred that's fine, but the docs should be more specific.

docs/docs/config/power.md
```diff
@@ -18,11 +18,12 @@ In the deep sleep state, the keyboard additionally disconnects from Bluetooth an
...
-| ------------------------------- | ---- | ----------------------------------------------------- | ------- |
-| `CONFIG_ZMK_IDLE_TIMEOUT`       | int  | Milliseconds of inactivity before entering idle state | 30000   |
-| `CONFIG_ZMK_SLEEP`              | bool | Enable deep sleep support                             | n       |
-| `CONFIG_ZMK_IDLE_SLEEP_TIMEOUT` | int  | Milliseconds of inactivity before entering deep sleep | 900000  |
+| Config                                 | Type | Description                                           | Default |
+| -------------------------------------- | ---- | ----------------------------------------------------- | ------- |
+| `CONFIG_ZMK_IDLE_TIMEOUT`              | int  | Milliseconds of inactivity before entering idle state | 30000   |
+| `CONFIG_ZMK_SLEEP`                     | bool | Enable deep sleep support                             | n       |
+| `CONFIG_ZMK_IDLE_SLEEP_TIMEOUT`        | int  | Milliseconds of inactivity before entering deep sleep | 900000  |
+| `CONFIG_ZMK_AWAKE_WHILE_BLE_CONNECTED` | bool | Whether to stay awake while BLE is connected          | n       |
```
huber-th, line 26: I notice that CONFIG_ZMK_AWAKE_WHILE_BLE_CONNECTED stands out due to its deviation from the consistent naming convention where the other options include "SLEEP."

Considering consistency aids clarity and contributes to a better user experience, I propose renaming `CONFIG_ZMK_AWAKE_WHILE_BLE_CONNECTED` to `CONFIG_ZMK_NO_SLEEP_WHILE_BLE_CONNECTED`

As description I would suggest `Prevent deep sleep while BLE is connected`

elpekenin: That was actually the first name i used, but pete suggested changing it.  See [here](https://discord.com/channels/719497620560543766/719544230497878116/1218308284801351852)

Im fine with any name as the relevant part is its functionality... Will change it to whichever is prefered.

elpekenin:
> [...] current profile [...] or any of the profiles [...]

Didnt even know that a profile could still be connected after changing to another. But even then, i agree that it makes more sense to (not) sleep based on the selected one, rather than having a "leftover" connection prevent power saving

huber-th:
> > [...] current profile [...] or any of the profiles [...]
> 
> […] i agree that it makes more sense to (not) sleep based on the selected one, rather than having a "leftover" connection prevent power saving

Agreed. I don’t think this should prevent deep sleep if a device is in range which isn’t selected as the active profile. 

Otherwise my phone would keep my keyboard up most of the time, being in range of the keyboard but not the selected profile.

jridgewell (changes requested):
There's a bug with the peripheral side of the keyboard not compiling because it can't find the `zmk_ble_active_profile_is_connected` function: https://github.com/jridgewell/zmk-config/actions/runs/8313322243/job/22749170208

app/src/activity.c
```diff
@@ -126,7 +130,14 @@ void activity_work_handler(struct k_work *work) {
...
+#if IS_ENABLED(CONFIG_ZMK_AWAKE_WHILE_BLE_CONNECTED)
+        // if user inactive and no USB
+        // keyboard will sleep as soon as BLE is disconnected
+        && !zmk_ble_active_profile_is_connected()
```
jridgewell, line 138: Got this working on the peripheral side with:
```suggestion
#if !IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
        && !zmk_ble_active_profile_is_connected()
#else
        && !zmk_split_bt_peripheral_is_connected()
#endif
```

aroum:
> peripheral side

Here is an example of checking the connection for the peripheral half

https://github.com/SplittedSpace/zmk-firmware/blob/913ecc0f21d5bebb9405ecd826cc61a375573385/app/boards/arm/harper/status_led.c#L287C1-L287C111

docs/docs/config/power.md
caksoylar: This should be updated to describe the behavior for peripheral as well.

caksoylar (commented):
Docs LGTM, thanks for the update.

app/src/activity.c
```diff
@@ -69,7 +73,18 @@ void activity_work_handler(struct k_work *work) {
...
+    /* if user inactive and USB is not connected,
+     * keyboard will sleep as soon as BLE is disconnected
+     */
+#if !IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
+        && !zmk_ble_active_profile_is_connected()
+#else
+        && !zmk_split_bt_peripheral_is_connected()
+#endif
+#endif
+    ) {
```
petejohanson, line 87: Wrapping this `if` with these preprocessor checks like this feels really brittle/painful.

Thoughts on something like:

```
bool prevent_sleep = is_usb_power_present();
#if IS_ENABLED(CONFIG_ZMK_NO_SLEEP_WHILE_BLE_CONNECTED)
prevent_sleep = prevent_sleep ||
#if !IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
   zmk_ble_active_profile_is_connected();
#else
   zmk_split_bt_peripheral_is_connected();
#endif
#endif
```

Then the check can be a simple `if (inactive_time > MAX_SLEEP_MX && !prevent_sleep) {`

Thoughts?

petejohanson (changes requested):
One thought on the code style. Thanks for the PR!

app/src/activity.c
```diff
@@ -68,8 +72,21 @@ int activity_event_listener(const zmk_event_t *eh) {
...
+    /* if user inactive and USB is not connected,
+     * keyboard will sleep as soon as BLE is disconnected
+     */
+#if !IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
+    prevent_sleep = prevent_sleep || zmk_ble_active_profile_is_connected();
+#else
+    prevent_sleep = prevent_sleep || zmk_split_bt_peripheral_is_connected();
+#endif
+#endif
+    if (inactive_time > MAX_SLEEP_MX && !prevent_sleep) {
```
jridgewell, line 89:
```suggestion
    if (inactive_time > MAX_SLEEP_MS && !prevent_sleep) {
```

Nick-Munnich:
Rather than preventing sleep entirely while BLE is connected, I think it would make more sense to add a new flag which sets a separate sleep timer when neither the current BLE profile nor a USB connection is present. That would solve the original issue, while not removing the current behaviour as an option.

elpekenin:
Current behavior is still an option (the default one, actually) if you dont enable the `ZMK_NO_SLEEP_WHILE_BLE_CONNECTED` flag (?)

Nick-Munnich:
Sorry, poorly phrased on my part: Instead of choosing one behaviour or the other, you could have both behaviours simultaneously.

zoriya (approved):

PR #2212, opened 2024-03-19
feat(docs): Add troubleshooting section for empty_file error

filterpaper:
Add a troubleshooting section for `empty_file.c` `lacks #binding-cells` error

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -56,6 +56,14 @@ A `devicetree error` followed by a reference to the line number on `<keyboard>.k
 devicetree error: /__w/zmk-config/zmk-config/config/cradio.keymap:109 (column 4): parse error: expected ';' or ','
 ```
 
+A `devicetree error` followed by an `empty_file.c` reference with `lacks #binding-cells` string indicates possible problems with improper parameters for specific bindings:
+
+```
+devicetree error: <Node /soc/gpio@50000300 in '/tmp/tmp.vJq9sMwkcY/zephyr/misc/empty_file.c'> lacks #binding-cells
+```
+
+This error can be triggered by incorrect binding syntax such as `&kp BT_SEL 0` instead of `&bt BT_SEL 0`.
+
 #### devicetree_unfixed.h error
 
 A `devicetree_unfixed.h` error that follows with an "undeclared here" string indicates a problem with key bindings, like behavior nodes (e.g. `&kp` or `&mt`) with incorrect number of parameters:


docs/docs/troubleshooting.md
```diff
@@ -74,6 +74,16 @@ Key positions are numbered starting from `0` at the top left key on the keymap,
 A common mistake that leads to this error is to use [key press keycodes](behaviors/key-press.md) without the leading `&kp` binding. That is, having entries such as `SPACE` that should have been `&kp SPACE`.
 :::
 
+#### empty_file.c error
```
huber-th, line 77: Personally, I think it would make sense to add this to the existing `#### devicetree error` section above as another error type instead of as a new section?

e.g. add the following to the above:

A `devicetree error` followed by a reference to `empty_file.c'>  with lacks #binding-cells` indicates possible problems with improper bindings for specific codes:
```
devicetree error: <Node /soc/gpio@50000300 in '/tmp/tmp.vJq9sMwkcY/zephyr/misc/empty_file.c'> lacks #binding-cells
```
This error can be triggered by incorrect binding syntax such as `&kp BT_SEL 0` instead of `&bt BT_SEL 0`.

caksoylar: I think this makes sense as well, since the line starts with `devicetree error:` just like that section.

docs/docs/troubleshooting.md
```diff
@@ -74,6 +74,16 @@ Key positions are numbered starting from `0` at the top left key on the keymap,
...
+
+An `empty_file.c` error with `lacks #binding-cells` string indicates possible problems with improper bindings for specific codes:
```
caksoylar, line 79:
```suggestion
An `empty_file.c` error with `lacks #binding-cells` string indicates possible problems with improper parameters for specific bindings:
```

filterpaper:
Thanks for the feedback. Suggestions are committed.

docs/docs/troubleshooting.md
```diff
@@ -56,6 +56,14 @@ A `devicetree error` followed by a reference to the line number on `<keyboard>.k
...
 
+A `devicetree error` followed by an `empty_file.c` reference with `lacks #binding-cells` string indicates possible problems with improper parameters for specific bindings:
+
+```
+devicetree error: <Node /soc/gpio@50000300 in '/tmp/tmp.vJq9sMwkcY/zephyr/misc/empty_file.c'> lacks #binding-cells
+```
+
+This error can be triggered by incorrect binding syntax such as `&kp BT_SEL 0` instead of `&bt BT_SEL 0`.
+
 #### devicetree_unfixed.h error
```
huber-th, line 67: Looking at this again, I wonder if this `devicetree_unfixed.h` is also a `devicetree error` which should be added to the above section rather than having its own header?

The example looks like the start `devicetree error:` was removed. Otherwise it looks like a similar error in which case it probably shouldn't have it's own heading?

caksoylar: Here is a recent variant of this error: <https://github.com/l2yuk3n/zmk-config/actions/runs/8303562600/job/22727892620#step:11:120> It indeed doesn't start with `devicetree error:`.

I believe the wording is a bit different since the last Zephyr upgrade or two, so it would be good to update it. For instance `unfixed` is now `generated`. I think it would make sense to merge this with this new section as well, since the root cause is very similar (or the same? mismatch between the binding and #parameters).

That being said we can do that in another PR.

caksoylar (approved):
Thanks again, I can merge this unless you decide you would like to address the new comment (it's fine to not do it right now, like I noted).

filterpaper:
> Thanks again, I can merge this unless you decide you would like to address the new comment (it's fine to not do it right now, like I noted).

Lets merge this particular PR if you are fine with the empty_file error. The current "devicetree_generated.h" message is a different context.

caksoylar (approved):
Thanks!

PR #2216, opened 2024-03-22
fix(devcontainer): Add packages required by grpcio

c10l:
When running `pip3 install --user -r zephyr/scripts/requirements.txt` (last step on https://zmk.dev/docs/development/setup), grpcio install fails with the following error:

    FileNotFoundError: [Errno 2] No such file or directory: c++: c++

This happens due to missing dev tooling in the container.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,3 +2,4 @@ FROM docker.io/zmkfirmware/zmk-dev-arm:3.5
 
 COPY .bashrc tmp
 RUN mv /tmp/.bashrc ~/.bashrc
+RUN apt-get update && apt-get install -y build-essential


huber-th:
Should this rather go into the `dev-generic` part of https://github.com/zmkfirmware/zmk-docker to include the dev tooling into the dev base image directly rather than just adding it to the `devcontainer` here?

c10l:
I'm not sure but I opened one there and referenced this one.

petejohanson:
I'm... a little hesitant to do this; the reality is, this is coming in from `requirements-extras.txt` in Zephyr, which has a lot of stuff that is *not* required for normal users to just build for their keyboards.

We used to only document installing the `requirements-base.txt` deps from Zephyr, but it looks like that got changed in 82ff6d8e857bf581714f2a2e23af3aa6bf4cf4e2 by Joel. @joelspadin is there a specific reason we changed that I don't recall, or could we go back to just recommending the base deps?

joelspadin:
I don't remember. I probably just copied it from the Zephyr docs somewhere?

PR #2222, opened 2024-03-23
feat: default layer setter

elpekenin:
I plan to do a PR to *persistently* change the default layer, for multi-layout uses (eg win/mac or qwerty/dvorak)

However, while activating the layer or to'ing into it is similar, it wouldn't mimic the behavior of default layer. Thus, the follow-up PR needs an API of this kind.

 Notes:
- New event for **default** layer changes?
- I dont quite like the function's name and log texts, ideas are welcome.

### DIFF
--- a/app/src/keymap.c
+++ b/app/src/keymap.c
@@ -108,6 +108,29 @@ static inline int set_layer_state(uint8_t layer, bool state) {
 
 uint8_t zmk_keymap_layer_default(void) { return _zmk_keymap_layer_default; }
 
+int zmk_keymap_layer_set_default(uint8_t layer) {
+    int ret = 0;
+    uint8_t prev_default = _zmk_keymap_layer_default;
+
+    ret = set_layer_state(layer, true);
+    if (ret < 0) {
+        LOG_WRN("Could not turn on the new default layer, bailing out.");
+        return ret;
+    }
+
+    _zmk_keymap_layer_default = layer;
+    ret = set_layer_state(prev_default, false);
+    if (ret < 0) {
+        LOG_WRN("Could not disable current default layer, undoing changes.");
+        _zmk_keymap_layer_default = prev_default;
+        set_layer_state(layer, false);
+        return ret;
+    }
+
+    LOG_DBG("default_layer_changed: %d", layer);
+    return 0;
+}
+
 zmk_keymap_layers_state_t zmk_keymap_layer_state(void) { return _zmk_keymap_layer_state; }
 
 bool zmk_keymap_layer_active_with_state(uint8_t layer, zmk_keymap_layers_state_t state_to_test) {
--- a/app/include/zmk/keymap.h
+++ b/app/include/zmk/keymap.h
@@ -15,6 +15,7 @@
 typedef uint32_t zmk_keymap_layers_state_t;
 
 uint8_t zmk_keymap_layer_default(void);
+int zmk_keymap_layer_set_default(uint8_t layer);
 zmk_keymap_layers_state_t zmk_keymap_layer_state(void);
 bool zmk_keymap_layer_active(uint8_t layer);
 uint8_t zmk_keymap_highest_layer_active(void);


app/src/keymap.c
```diff
@@ -108,6 +108,29 @@ static inline int set_layer_state(uint8_t layer, bool state) {
 
 uint8_t zmk_keymap_layer_default(void) { return _zmk_keymap_layer_default; }
 
+int zmk_keymap_layer_set_default(uint8_t layer) {
+    int ret = 0;
+    uint8_t prev_default = _zmk_keymap_layer_default;
+
+    ret = set_layer_state(layer, true);
+    if (ret < 0) {
+        LOG_WRN("Could not turn on the new default layer, bailing out.");
```
huber-th, line 117:
```suggestion
        LOG_WRN("Failed to activate the new default layer; aborting changes.");
```

app/src/keymap.c
```diff
@@ -108,6 +108,29 @@ static inline int set_layer_state(uint8_t layer, bool state) {
...
+        return ret;
+    }
+
+    _zmk_keymap_layer_default = layer;
+    ret = set_layer_state(prev_default, false);
+    if (ret < 0) {
+        LOG_WRN("Could not disable current default layer, undoing changes.");
```
huber-th, line 124:
```suggestion
        LOG_WRN("Unable to deactivate the current default layer; reverting changes.");
```

huber-th (commented):
Being able to set the default layer on startup would be wonderful for multi layout configurations.

huber-th:
If I may add an idea to be considered in your follow up PR. Might be worth to consider being able to being able to trigger automatic layout changes on BLE profile change.

Not sure how this would be best made configurable (maybe via KConfig?, Could ZMK remember default layer per profile?), but it would be convenient if someone with Windows and macOS system for example, could configure a default layer for each BLE profile and one for USB to automatically trigger a default layer change depending on which profile is activated.

huber-th:
I acknowledge the potential advantage of the planned behavior for users with dual systems, but after thinking about it a bit more I'm wondering whether this proposed change should be directly integrated into the feature's implementation itself rather than as its separate PR.

elpekenin:
I dont mind either way, the idea of making a small PR was to start rolling it and see if it had missed a small edge case or whatever.

Also making two different PRs with the respective self-contained logic seemed better review-wise.

I already have the code (or at least, its first revision) written q slightly tested it (with to instead of setting default layer at ZMK level). So, if collabs think it is better to keep everything together just let me know and i will just it on thus branch too

huber-th:
> I dont mind either way, the idea of making a small PR was to start rolling it and see if it had missed a small edge case or whatever.
> 
> Also making two different PRs with the respective self-contained logic seemed better review-wise.

Sorry, I have been quite busy lately. Personally, I would recommend adding everything related to this feature in one PR so when it's merged it adds the full feature.

I tested your code from your personal repo and I would recommend to incorporate it into this PR for full review/testing of the complete feature.

elpekenin:
Aight, will do.

Again, my main idea was go test if default layer setting really worked well before relying on it for the behavior, for easier debug when (if) edge cases poped up.

Will add it to make the process faster and keep it together. Glad to hear the current implementation (layer toggle instead of default layer, which is worse) is working fine for you

app/src/behaviors/behavior_default_layer.c
```diff
@@ -0,0 +1,176 @@
...
+
+#include <zmk/event_manager.h>
+#include <zmk/events/endpoint_changed.h>
+
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct default_layer_settings_t {
+    uint8_t usb[ZMK_ENDPOINT_USB_COUNT];
+    uint8_t ble[ZMK_ENDPOINT_BLE_COUNT];
+};
```
petejohanson, line 26: This is too tied to the *details* of the transports, IMHO. You should  be more agnostic, e.g.

```suggestion
struct default_layer_settings_t {
    uint8_t endpoint_defaults[ZMK_ENDPOINT_COUNT];
};
```

And then code that needs to set/load the setting would simply call  `zmk_endpoint_instance_to_index` to determine which setting to pull out of that array consistently.

The advantage here, whenever we do eventually support some other new transport (e.g. some custom 2.4Ghz protocol, or maybe HID over i2c), this code will  adjust automatically.

petejohanson (changes requested):
One major concern regarding the usage of the endpoints API. Thanks for the PR!

app/src/behaviors/behavior_default_layer.c
elpekenin: That's what i get for not looking closely at the codebase, i ended up reinventing the wheel 😝 

Im still a bit worried about issues when user has old settings stored (eg move layers around and whatnot).
Any ideas that could help with that? Imho code shouldn't worry too much about those scenarios and let the user be the one in charge of deteting/fixing. But if there is some easy/robust idea, i can sure add it.

If nothing gets added i will add some warn messge in docs as in "beware the stored settings might break when you tweak the keymap"

app/src/behaviors/behavior_default_layer.c
```diff
@@ -0,0 +1,176 @@
...
+        default_layers.usb[0] = layer;
+        break;
+
+    case ZMK_TRANSPORT_BLE:
+        __ASSERT(endpoint.ble.profile_index < ZMK_ENDPOINT_BLE_COUNT, "Unreachable");
+        default_layers.ble[endpoint.ble.profile_index] = layer;
+        break;
+    }
+
+    int ret = settings_save_one("default_layer/settings", &default_layers, sizeof(default_layers));
```
caksoylar, line 115: Do we want to apply debouncing here like most other settings that save to flash?

docs/docs/behaviors/layers.md
```diff
@@ -153,3 +153,29 @@ It is possible to use "toggle layer" to have keys that raise and lower the layer
 
 The "conditional layers" feature enables a particular layer when all layers in a specified set are active.
 For more information, see [conditional layers](../features/conditional-layers.md).
+
+## Default Layer
+
+The default layer behavior allows configuring a different default layer, for example to test DVORAK while keeping QWERTY on another layer, or moving a couple keycodes around for Windows/Mac usage.
```
caksoylar, line 159: I feel like this needs a little more information on the default layer concept, since it wasn't modifiable from 0 until this. I'd at least note here that the default is the first layer, that it cannot be changed or deactivated by other behaviors in this section.

caksoylar (commented):
Two other docs pieces need updating in addition to below comments:
- https://zmk.dev/docs/behaviors#layer-navigation-behaviors table
- The default layer remark in https://zmk.dev/docs/config/keymap#devicetree

docs/docs/behaviors/layers.md
ReFil: On this note, I've used the default layer concept differently on the ROTR to have an encoder scroll in a momentary layer change the layer you return to when that momentary layer is released, skipping layers tagged to be momentary only. I think there's reasoning behind an endpoint agnostic default layer mechanism too, perhaps a pair of bool props in the behaviour to choose whether it respects endpoints and saves to flash?

elpekenin: Thanks for the docs check, will get them updated tomorrow. 

As per the delayed writing, Im not sure whether it is needed. Based on my understanding of Zephyr's docs, it only triggers a write if the value has changed (eg: differs from what the flash contains already). 

If that is correct, I dont think an inmediate write is bad (wearing down memory - wise) but for sure is in terms of consistency with codebase, will update too.

Nick-Munnich:
#1500 seems like it could be relevant here and hasn't been mentioned yet, just dropping a link.

PR #2224, opened 2024-03-24
feat(build): Allow specifying snippets for a build.

petejohanson:
* Allow using snippets https://docs.zephyrproject.org/latest/build/snippets/using.html for user builds in a `snippets` array propertly.

You can see this in action in https://github.com/petejohanson/le-capybara-zmk-config/actions/runs/8412145898 which enables an extra EC calibration utility using a snippet there.

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -65,6 +65,7 @@ jobs:
           board: ${{ matrix.board }}
           shield: ${{ matrix.shield }}
           artifact_name: ${{ matrix.artifact-name }}
+          snippet: ${{ matrix.snippet }}
         run: |
           if [ -e zephyr/module.yml ]; then
             export zmk_load_arg=" -DZMK_EXTRA_MODULES='${GITHUB_WORKSPACE}'"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-user-config.yml
```diff
@@ -59,12 +59,16 @@ jobs:
         run: |
           echo "build_dir=$(mktemp -d)" >> $GITHUB_ENV
 
+      - name: Install yaml2json
+        run: apt-get update && apt-get install -y jq python3-pip && python3 -m pip install remarshal
```
joelspadin, line 63: This doesn't match the step with the same name on line 38. Should those match?

Also, would https://github.com/actions/setup-python be better for making sure pip is installed?

joelspadin (commented):
This is a bit inconsistent with how we pass shields. It seems weird to me that you would need to write

```yaml
- board: nice_nano_v2
  shield: corne_left nice_view_adapter nice_view
```

to use multiple shields, but

```yaml
- board: nice_nano_v2
  snippets: [snippet1, snippet2, snippet3]
```

to use multiple snippets

petejohanson:
After some discussion and discovery, simplified things a ton, and here's a successful build with the new simpler approach: https://github.com/petejohanson/le-capybara-zmk-config/actions/runs/8413180471/job/23035032995

joelspadin (approved):

PR #2193, opened 2024-03-08
feat(docs): Add Behavior overview page

huber-th:
Resolves #1044 
Resolves #525 

@petejohanson this is an initial draft of the behavior overview page, structured with groupings I deemed sensible. While I'm not entirely satisfied with all the descriptions yet, I aimed to steer clear of excessive repetition from the detailed behavior pages to prevent duplication of content yet give a good idea to readers what each behavior is for.

I opted not to include an example column because the majority of examples would be quite basic, typically consisting of bindings and keycodes. Additionally, the provided links typically direct users straight to the relevant section of the documentation explaining the usage.

As for behaviors that don't have their own binding, such as macros and sensors, I wasn't entirely sure where to put them. Currently, I've placed them at the end in a section titled 'Other'.

Let me know what you think.

PR #2225, opened 2024-03-25
refactor(docs): Refer to overview page for behaviors

caksoylar:
Thanks to @huber-th in #2193 we now have an overview page for behaviors. Add references to that page instead of mentioning key press page in places where we would like to refer to behaviors in general.


petejohanson (approved):
Good call!

caksoylar:
Thanks, forgot to force push my latest tweaks! Please see again.

petejohanson (approved):

PR #2229, opened 2024-03-27
fix(build): Modify function return type to remove compiler warning

huber-th:
Function [sticky_key_timeout](https://github.com/zmkfirmware/zmk/blob/main/app/src/behaviors/behavior_sticky_key.c#L124) is defined with return type `int`, yet it is not returning any value. Both calls to this function also ignore the return value. Therefore the return type is modified to `void` to remove this warning.


petejohanson (approved):
Thanks! Will merge once tests finish.

PR #2231, opened 2024-03-27
feat(behaviors): Add behavior metadata information.

petejohanson:
* For upcoming ZMK studio work, make a set of rich metadata available to provide a friendly name for a behavior, and allow super flexible descriptions of the parameters the behaviors take.

### DIFF
--- a/app/dts/behaviors/to_layer.dtsi
+++ b/app/dts/behaviors/to_layer.dtsi
@@ -9,6 +9,7 @@
         /omit-if-no-ref/ to: to_layer {
             compatible = "zmk,behavior-to-layer";
             #binding-cells = <1>;
+            display-name = "To Layer";
         };
     };
 };
--- a/app/dts/behaviors/toggle_layer.dtsi
+++ b/app/dts/behaviors/toggle_layer.dtsi
@@ -9,6 +9,7 @@
         /omit-if-no-ref/ tog: toggle_layer {
             compatible = "zmk,behavior-toggle-layer";
             #binding-cells = <1>;
+            display-name = "Toggle Layer";
         };
     };
 };
--- a/app/include/drivers/behavior.h
+++ b/app/include/drivers/behavior.h
@@ -23,6 +23,39 @@
  * (Internal use only.)
  */
 
+struct behavior_parameter_value_metadata {
+    char *display_name;
+
+    union {
+        uint32_t value;
+        struct {
+            int32_t min;
+            int32_t max;
+        } range;
+    };
+
+    enum {
+        BEHAVIOR_PARAMETER_VALUE_TYPE_NIL = 0,
+        BEHAVIOR_PARAMETER_VALUE_TYPE_VALUE = 1,
+        BEHAVIOR_PARAMETER_VALUE_TYPE_RANGE = 2,
+        BEHAVIOR_PARAMETER_VALUE_TYPE_HID_USAGE = 3,
+        BEHAVIOR_PARAMETER_VALUE_TYPE_LAYER_INDEX = 4,
+    } type;
+};
+
+struct behavior_parameter_metadata_set {
+    size_t param1_values_len;
+    const struct behavior_parameter_value_metadata *param1_values;
+
+    size_t param2_values_len;
+    const struct behavior_parameter_value_metadata *param2_values;
+};
+
+struct behavior_parameter_metadata {
+    size_t sets_len;
+    const struct behavior_parameter_metadata_set *sets;
+};
+
 enum behavior_sensor_binding_process_mode {
     BEHAVIOR_SENSOR_BINDING_PROCESS_MODE_TRIGGER,
     BEHAVIOR_SENSOR_BINDING_PROCESS_MODE_DISCARD,
@@ -37,6 +70,10 @@ typedef int (*behavior_sensor_keymap_binding_accept_data_callback_t)(
     struct zmk_behavior_binding *binding, struct zmk_behavior_binding_event event,
     const struct zmk_sensor_config *sensor_config, size_t channel_data_size,
     const struct zmk_sensor_channel_data channel_data[channel_data_size]);
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+typedef int (*behavior_get_parameter_metadata_t)(
+    const struct device *behavior, struct behavior_parameter_metadata *param_metadata);
+#endif // IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
 
 enum behavior_locality {
     BEHAVIOR_LOCALITY_CENTRAL,
@@ -51,23 +88,54 @@ __subsystem struct behavior_driver_api {
     behavior_keymap_binding_callback_t binding_released;
     behavior_sensor_keymap_binding_accept_data_callback_t sensor_binding_accept_data;
     behavior_sensor_keymap_binding_process_callback_t sensor_binding_process;
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+    behavior_get_parameter_metadata_t get_parameter_metadata;
+    const struct behavior_parameter_metadata *parameter_metadata;
+#endif // IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
 };
 /**
  * @endcond
  */
 
+struct zmk_behavior_metadata {
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+    const char *display_name;
+#endif // IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+};
+
 struct zmk_behavior_ref {
     const struct device *device;
+    const struct zmk_behavior_metadata metadata;
 };
+
+#define ZMK_BEHAVIOR_REF_DT_NAME(node_id) _CONCAT(zmk_behavior_, DEVICE_DT_NAME_GET(node_id))
+
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+
+#define ZMK_BEHAVIOR_METADATA_INITIALIZER(node_id)                                                 \
+    { .display_name = DT_PROP_OR(node_id, display_name, DEVICE_DT_NAME(node_id)), }
+
+#else
+
+#define ZMK_BEHAVIOR_METADATA_INITIALIZER(node_id)                                                 \
+    {}
+
+#endif // IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+
+#define ZMK_BEHAVIOR_REF_INITIALIZER(node_id, _dev)                                                \
+    { .device = _dev, .metadata = ZMK_BEHAVIOR_METADATA_INITIALIZER(node_id), }
+
+#define ZMK_BEHAVIOR_REF_DEFINE(name, node_id, _dev)                                               \
+    static const STRUCT_SECTION_ITERABLE(zmk_behavior_ref, name) =                                 \
+        ZMK_BEHAVIOR_REF_INITIALIZER(node_id, _dev)
+
+#define ZMK_BEHAVIOR_REF_DT_DEFINE(node_id)                                                        \
+    ZMK_BEHAVIOR_REF_DEFINE(ZMK_BEHAVIOR_REF_DT_NAME(node_id), node_id, DEVICE_DT_GET(node_id))
 
 /**
  * Registers @p node_id as a behavior.
  */
-#define BEHAVIOR_DEFINE(node_id)                                                                   \
-    static const STRUCT_SECTION_ITERABLE(zmk_behavior_ref,                                         \
-                                         _CONCAT(zmk_behavior_, DEVICE_DT_NAME_GET(node_id))) = {  \
-        .device = DEVICE_DT_GET(node_id),                                                          \
-    }
+#define BEHAVIOR_DEFINE(node_id) ZMK_BEHAVIOR_REF_DT_DEFINE(node_id)
 
 /**
  * @brief Like DEVICE_DT_DEFINE(), but also registers the device as a behavior.
@@ -88,6 +156,52 @@ struct zmk_behavior_ref {
 #define BEHAVIOR_DT_INST_DEFINE(inst, ...)                                                         \
     DEVICE_DT_INST_DEFINE(inst, __VA_ARGS__);                                                      \
     BEHAVIOR_DEFINE(DT_DRV_INST(inst))
+
+/**
+ * @brief Validate a given behavior binding is valid, including parameter validation
+ * if the metadata feature is enablued.
+ *
+ * @param binding The behavior binding to validate.
+ *
+ * @retval 0 if the passed in binding is valid.
+ * @retval -ENODEV if the binding references a non-existant behavior.
+ * @retval -EINVAL if parameters are not valid for the behavior metadata.
+ */
+int zmk_behavior_validate_binding(const struct zmk_behavior_binding *binding);
+
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+
+int zmk_behavior_get_empty_param_metadata(const struct device *dev,
+                                          struct behavior_parameter_metadata *metadata);
+
+/**
+ * @brief Validate a given behavior parameters match the behavior metadata.
+ *
+ * @param metadata The behavior metadata to validate against
+ * @param param1 The first parameter value
+ * @param param2 The second parameter value
+ *
+ * @retval 0 if the passed in parameters are valid.
+ * @retval -ENODEV if metadata is NULL.
+ * @retval -EINVAL if parameters are not valid for the metadata.
+ */
+int zmk_behavior_check_params_match_metadata(const struct behavior_parameter_metadata *metadata,
+                                             uint32_t param1, uint32_t param2);
+/**
+ * @brief Validate a given behavior parameter matches the behavior metadata parameter values.
+ *
+ * @param values The values to validate against
+ * @param values_len How many values to check
+ * @param param The value to check.
+ *
+ * @retval 0 if the passed in parameter is valid.
+ * @retval -ENODEV if values is NULL.
+ * @retval -EINVAL if parameter is not valid for the value metadata.
+ */
+int zmk_behavior_validate_param_values(const struct behavior_parameter_value_metadata *values,
+                                       size_t values_len, uint32_t param);
+
+#endif // IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
 
 /**
  * Syscall wrapper for zmk_behavior_get_binding().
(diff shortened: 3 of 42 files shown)


app/dts/behaviors/to_layer.dtsi
```diff
@@ -9,6 +9,7 @@
         /omit-if-no-ref/ to: to_layer {
             compatible = "zmk,behavior-to-layer";
             #binding-cells = <1>;
+            friendly-name = "Go To Layer";
```
caksoylar, line 12: We don't call this "Go To Layer" anywhere in the docs, just "To Layer". I think it'd be good to be consistent, do you prefer the former?

app/dts/behaviors/toggle_layer.dtsi
```diff
@@ -9,6 +9,7 @@
         /omit-if-no-ref/ tog: toggle_layer {
             compatible = "zmk,behavior-toggle-layer";
             #binding-cells = <1>;
+            friendly-name = "Layer Toggle";
```
caksoylar, line 12: This is also typically called "Toggle Layer." Having a proper noun phrase rather than verb+action might be better in general, but I'd again tend to consistency.

app/dts/behaviors/to_layer.dtsi
petejohanson: As I was reading it.. I realized it was a pretty odd name. I'll update to match our current language.

app/dts/behaviors/toggle_layer.dtsi
petejohanson: I'll defer to consistency for now, but I do think this reads better than "Toggle Layer", IMHO.

app/src/behaviors/behavior_bt.c
```diff
@@ -20,6 +20,73 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
 #if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
 
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+
+static const struct behavior_parameter_value_metadata no_arg_values[] = {
+    {
+        .position = 0,
+        .value = BT_NXT_CMD,
+        .friendly_name = "Next Profile",
+        .type = BEHAVIOR_PARAMETER_VALUE_METADATA_TYPE_VALUE,
```
joelspadin, line 30: Nit: I think it would be easier to read these if the type came before the value/range/etc. Maybe this would be a good order?

1. Position
2. Name
3. Type
4. Value/range/etc

app/include/drivers/behavior.h
```diff
@@ -51,23 +113,57 @@ __subsystem struct behavior_driver_api {
...
+#define ZMK_BEHAVIOR_REF_DT_DEFINE(node_id)                                                        \
+    ZMK_BEHAVIOR_REF_DEFINE(ZMK_BEHAVIOR_REF_DT_NAME(node_id), DEVICE_DT_GET(node_id),             \
+                            DT_PROP_OR(node_id, friendly_name, DEVICE_DT_NAME(node_id)))
+
+#else
+
+#define ZMK_BEHAVIOR_REF_INITIALIZER(_dev)                                                         \
+    { .device = _dev, }
+
+#define ZMK_BEHAVIOR_REF_DEFINE(name, ...)                                                         \
```
joelspadin, line 154: Does the compiler complain about unused macro parameters? If not, you could replace the variable args with an explicit parameter list and then just pass in everything but ignore the unneeded parameters whenever metadata is disabled. As this is currently written, it's quite difficult to understand what parameters this takes in either case.

Also, an idea for reducing the number of macros that need alternate implementations:

```c
struct zmk_behavior_metadata {
    // Making this still defined but empty when metadata is disabled so we can write
    // ".metadata = ..." and have it still compile when metadata is disabled.
#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
    const char *name;
    // any additional fields needed later go here...
#endif
};

#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
#define ZMK_BEHAVIOR_METADATA_INITIALIZER(name) \
    { .name = name }
#else
#define ZMK_BEHAVIOR_METADATA_INITIALIZER(name) {}
#endif    

struct zmk_behavior_ref {
    const struct device *device;
    struct zmk_behavior_metadata metadata;
};

#define ZMK_BEHAVIOR_REF_INITIALIZER(dev, name) \
    { .device = dev, .metadata = ZMK_BEHAVIOR_METADATA_INITIALIZER(name) }
```

(Test for that idea: https://godbolt.org/z/GzaeT9dYh)

app/src/behaviors/behavior_bt.c
petejohanson: Fixed. Definitely reads better, thanks!

app/include/drivers/behavior.h
petejohanson: Yeah, I dig it. Took a slightly different approach though and just passed the node_id into the macro, in case more gets added later, the `ZMK_BEHAVIOR_METADATA_INITIALIZER` can handle that "internally".

Thoughts on the updated take?

app/include/drivers/behavior.h
```diff
@@ -23,6 +23,64 @@
...
+    size_t sets_len;
+    struct behavior_parameter_metadata_custom_set sets[];
+};
+
+struct behavior_parameter_metadata {
+    union {
+        struct {
+            enum behavior_parameter_standard_domain param1;
+            enum behavior_parameter_standard_domain param2;
+        } standard;
```
joelspadin, line 71: If I'm understanding this correctly, this is effectively just a shortcut for 

```c
{
  .type = BEHAVIOR_PARAMETER_METADATA_CUSTOM,
  .custom = {
    .sets_len = 1,
    .sets = {{
      .values_len = 2,
      .values = {{
        .position = 0,
        .type = BEHAVIOR_PARAMETER_VALUE_METADATA_TYPE_STANDARD,
        .standard = <param1>,
      }, {
        .position = 1,
        .type = BEHAVIOR_PARAMETER_VALUE_METADATA_TYPE_STANDARD,
        .standard = <param2>,
      }},
    }},
  },
}
```

right? If so, that simplifies defining the data, but it adds a special case to the implementation that has to read this data. You also still have to use the more complex form for something like a MIDI note behavior that uses `param1 = range[0, 127]` (note) and `param2 = range[0, 127]` (velocity), since there's no way to represent a range type here.

Is there any way we could avoid the redundant ways to represent the same data? Maybe have some predefined constants and/or macros to let you build a metadata set that only has one item without all the boilerplate? Alternatively, could we support all possible value types in this inline form of writing the parameter data?

app/include/drivers/behavior.h
```diff
@@ -23,6 +23,64 @@
...
+        } range;
+
+        enum behavior_parameter_standard_domain standard;
+    };
+
+    enum {
+        BEHAVIOR_PARAMETER_VALUE_METADATA_TYPE_VALUE = 0,
+        BEHAVIOR_PARAMETER_VALUE_METADATA_TYPE_RANGE = 1,
+        BEHAVIOR_PARAMETER_VALUE_METADATA_TYPE_STANDARD = 2,
+    } type;
```
joelspadin, line 53: There seems to be a bit of overlap between what this and `enum behavior_parameter_standard_domain` conceptually represent. If we can get rid of the special case in `behavior_parameter_metadata` as suggested below, maybe "domain" and "type" could be merged together?

joelspadin (changes requested):

app/include/drivers/behavior.h
```diff
@@ -23,6 +23,64 @@
...
+struct behavior_parameter_value_metadata {
+    char *friendly_name;
+    uint8_t position;
+
+    union {
+        uint32_t value;
+        struct {
+            uint16_t min;
+            uint16_t max;
+        } range;
```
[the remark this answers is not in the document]
petejohanson: Hmm.... some good questions. I suppose for this, I was having a hard time imagining a range needing to be larger than this, or negative, but since we're trying to have this not have to change any time soon.... worth being a bit more flexible here.

Do we think `{ int32_t min; uint32_t max; } range` would work? Do we *really* think we need a range that's only negative numbers? or with a minimum above 2,147,483,647 ever?

app/include/drivers/behavior.h
```diff
@@ -23,6 +23,64 @@
  * (Internal use only.)
  */
 
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+
+enum behavior_parameter_standard_domain {
+    BEHAVIOR_PARAMETER_STANDARD_DOMAIN_NULL = 0,
+    BEHAVIOR_PARAMETER_STANDARD_DOMAIN_HID_USAGE = 1,
+    BEHAVIOR_PARAMETER_STANDARD_DOMAIN_LAYER_INDEX = 2,
+    BEHAVIOR_PARAMETER_STANDARD_DOMAIN_HSV = 3,
```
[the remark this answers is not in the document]
petejohanson: We had originally used "HSV" as the "native" format to make it easier to do increase/decrease brightness, IIRC. I don't have a strong opinion here, tbh. Maybe @Nicell has soe thoughts though.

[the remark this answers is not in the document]
joelspadin: My main concern is that you can only represent a small fraction of the possible RGB colors using integer HSV values. Assuming I correctly wrote the Python script I used to test this, rgb_underglow.c's `hsb_to_rgb()` function can represent 2,134,672 out of 16,777,216 possible colors (about 13%).

app/include/drivers/behavior.h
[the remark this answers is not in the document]
joelspadin: Mixing signed and unsigned math in C is generally a great way to have a bad time. I'd be in favor of just using signed values for both min and max. I can't think of any real-world range of values that wouldn't fit into a signed 32-bit integer.

The first thing I could think of that might need large numbers is millisecond durations. With 16-bit, that could be a problem since you'd be limited to ~32 s signed or ~65 s unsigned. With signed 32-bit, you get ~25 days, which I can't imagine being insufficient for anything (if that's the kind of time scale your behavior is working on, you should probably be using seconds instead of milliseconds).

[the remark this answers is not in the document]
petejohanson: Yeah, totally fair. I'll update for signed 32-bit values. It'll increase the size a small bit, but not enough I think to warrant limiting things.

app/include/drivers/behavior.h
```diff
@@ -23,6 +23,64 @@
...
+    uint8_t position;
+
+    union {
+        uint32_t value;
+        struct {
+            uint16_t min;
+            uint16_t max;
+        } range;
+
+        enum behavior_parameter_standard_domain standard;
```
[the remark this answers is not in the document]
petejohanson: So, since this is the union field that matches the `type` field, I'd like the naming to match, which explains my choice here. Thoughts?

app/include/drivers/behavior.h
petejohanson: So... this is a very valid question..... After adding the necessary complexity for custom params.... do we really need both ways of doing this? I think I'm almost convinced we don't. Let me sleep on this, but if I can't come up with a compelling reason, I will refactor to "one single way of doing this" with some naming tweaks to make this saner.

petejohanson: Ok, pushed a major refactor here to simplify. Thoughts?

app/include/drivers/behavior.h
```diff
@@ -23,6 +23,51 @@
...
+        struct {
+            int32_t min;
+            int32_t max;
+        } range;
+
+        enum behavior_parameter_standard_domain standard;
+    };
+
+    enum {
+        BEHAVIOR_PARAMETER_VALUE_METADATA_TYPE_VALUE = 0,
```
joelspadin, line 50: Do we need "metadata" here? This is the type of the parameter value, not the type of the metadata itself, so this naming could be a little confusing.

app/include/drivers/behavior.h
```diff
@@ -120,6 +246,40 @@ static inline int z_impl_behavior_keymap_binding_convert_central_state_dependent
...
+    if (behavior == NULL || param_metadata == NULL) {
+        return -EINVAL;
+    }
+
+    const struct behavior_driver_api *api = (const struct behavior_driver_api *)behavior->api;
+
+    if (api->get_parameter_metadata) {
+        return api->get_parameter_metadata(behavior, param_metadata);
+    } else if (api->parameter_metadata) {
+        memcpy(param_metadata, api->parameter_metadata, sizeof(struct behavior_parameter_metadata));
```
joelspadin, line 273: Minor:
```suggestion
        *param_metadata = api->parameter_metadata;
```

app/src/behaviors/behavior_macro.c
```diff
@@ -209,9 +213,98 @@ static int on_macro_binding_released(struct zmk_behavior_binding *binding,
     return ZMK_BEHAVIOR_OPAQUE;
 }
 
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_METADATA)
+
+static int get_macro_parameter_metadata(const struct device *macro,
```
joelspadin, line 218: This function is massive. Any way it could be broken into smaller pieces?

The combination of loop conditions with param1_found/param2_found and continue/break statements is particularly difficult to understand.

app/include/drivers/behavior.h
petejohanson: Shortened.

app/include/drivers/behavior.h
petejohanson: Done

app/include/drivers/behavior.h
petejohanson: Done.

app/src/behaviors/behavior_macro.c
petejohanson: Ok, cleaned this up a ton, after properly identifying the edges. Read better?

app/src/behavior.c
```diff
@@ -39,6 +41,144 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+            }
+            break;
+        }
+
+        if (found_match) {
+            break;
+        }
+    }
+
+    return found_match ? 0 : -ENOTSUP;
```
joelspadin, line 128: Any reason this one return -ENOTSUP while most of the others return -EINVAL on a failure to match?

app/src/behavior.c
```diff
@@ -39,6 +41,144 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+    bool found_match = false;
+
+    if (values_len == 0) {
+        return -ENODEV;
+    }
+
+    for (int v = 0; v < values_len; v++) {
+        const struct behavior_parameter_value_metadata *value_meta = &values[v];
+
+        switch (value_meta->type) {
```
joelspadin, line 89: This block (matching against a single `struct behavior_parameter_value_metadata` seems like a good candidate for breaking out into a separate function.

app/src/behavior.c
```diff
@@ -39,6 +41,144 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+        case BEHAVIOR_PARAMETER_VALUE_TYPE_NIL:
+            if (param == 0) {
+                found_match = true;
```
joelspadin, line 92: Given that we unconditionally break from the loop and return 0 upon reaching any of these `found_match = true` cases, maybe we should just return 0 directly instead? Could add something like `#define FOUND_MATCH 0` so this becomes `return FOUND_MATCH;` and stays clear that we're returning a value that indicates we found a match.

app/src/behavior.c
```diff
@@ -39,6 +41,144 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+    for (int s = 0; s < metadata->sets_len; s++) {
+        const struct behavior_parameter_metadata_set *set = &metadata->sets[s];
+        int param1_ret =
+            zmk_behavior_validate_param_values(set->param1_values, set->param1_values_len, param1);
+        int param2_ret =
+            zmk_behavior_validate_param_values(set->param2_values, set->param2_values_len, param2);
+
+        if ((param1_ret >= 0 || (param1_ret == -ENODEV && param1 == 0)) &&
+            (param2_ret >= 0 || (param2_ret == -ENODEV && param2 == 0))) {
+            LOG_DBG("ALL GOOD!");
```
joelspadin, line 150: Probably should either remove this or add some extra context (what did it correctly match?) so the log message is useful later.

app/src/behavior.c
```diff
@@ -39,6 +41,144 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+        break;
+    default:
+        LOG_WRN("Unsupported HID usage page %d", usage_page);
+        return -EINVAL;
+    }
+
+    return 0;
+}
+
+int zmk_behavior_validate_param_values(const struct behavior_parameter_value_metadata *values,
```
joelspadin, line 78: Given the name "validate param values", I first thought this was for validating the `struct behavior_parameter_value_metadata` array itself. Not sure how better to clarify this. Could add some documentation to explain it, or maybe rename it? Some ideas

- `zmk_behavior_matches_param_values()`
- `zmk_behavior_validate_param_with_param_values()`

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -652,9 +658,52 @@ static int on_hold_tap_binding_released(struct zmk_behavior_binding *binding,
...
+    err = behavior_get_parameter_metadata(zmk_behavior_get_binding(cfg->hold_behavior_dev),
+                                          &child_meta);
+    if (err < 0) {
+        LOG_WRN("Failed to get the hold behavior parameter: %d", err);
+        return err;
+    }
+
+    if (child_meta.sets_len > 0) {
+        data->set.param1_values = child_meta.sets[0].param1_values;
+        data->set.param1_values_len = child_meta.sets[0].param1_values_len;
```
joelspadin, line 678: Is it assumed that these are already zeroed out before calling this function, or is it intentional to leave it at its previous value in the case where the `behavior_get_parameter_metadata()` function returns an empty list?

app/src/behaviors/behavior_macro.c
joelspadin: Might warrant some extra comments explaining exactly what this is doing, but I *think* I correctly understand what it's doing now, so definitely an improvement.

Just to be sure, this is looping through the list of bindings in the macro, looking for the first thing it can find that uses each of the macro parameters, then returning a combination of the sets for each of those? I'm not sure I understand why the logic for param1 is different than param1 though. Does the param1 case not need to check if the bound param2 is valid?

app/src/behaviors/behavior_hold_tap.c
petejohanson: Yeah, the assumption is the set's got a default zero'd state before this gets called.

app/src/behavior.c
petejohanson: Fixed, in a split out function. Thanks!

app/src/behavior.c
petejohanson: Indeed. Split out.

app/src/behavior.c
petejohanson: Oversight. Corrected. Thanks!

app/src/behavior.c
petejohanson: Renamed to `zmk_behavior_check_params_match_metadata` for clarity. That work for you?

PR #2238, opened 2024-03-29
fix: Add wakeup-source; to kscan nodes for ZMK_SLEEP.

petejohanson:
* ZMK_SLEEP also enables PM_DEVICE, so be sure to mark kscan
  nodes as wakeup sources so they can wake keyboards from sleep.

### DIFF
--- a/app/boards/arm/adv360pro/adv360pro_left.dts
+++ b/app/boards/arm/adv360pro/adv360pro_left.dts
@@ -10,6 +10,7 @@
 /{
     kscan0: kscan {
         compatible = "zmk,kscan-gpio-matrix";
+        wakeup-source;
 
         diode-direction = "col2row";
         row-gpios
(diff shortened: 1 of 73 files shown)


petejohanson:
@caksoylar tried so hard to save me in https://github.com/zmkfirmware/zmk/pull/1942#issuecomment-1848732928 and I still missed this. Fixing up the in-tree boards/shields, and will broadcast more broadly on Discord and SM.

app/boards/shields/softie/Kconfig.defconfig
caksoylar, line 1: Looks like this shield was included by accident.

caksoylar (commented):
`snap` and `nibble` shields are missing changes, but I assume that is intentional given they use demux anyway?

app/boards/shields/softie/Kconfig.defconfig
petejohanson: Indeed. Removed. Thanks.

petejohanson:
> `snap` and `nibble` shields are missing changes, but I assume that is intentional given they use demux anyway?

Correct.

caksoylar (approved):

PR #2239, opened 2024-03-29
fix(docs): Fix bindings types in behaviors config page

caksoylar:
Fixes the incorrect types for `bindings` prop in `docs/config/behaviors`, for hold-tap, sticky key and sensor rotate behaviors.

In a separate commit, I split the sensor rotate table into two separate tables for `sensor-rotate` and `sensor-rotate-var` since there didn't seem to be enough common properties anymore to make the single table worth it.


petejohanson (approved):
Thanks!

Issue #1376, opened 2022-07-04
ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE disregards original on/off underglow state


slashfoo:
Kconfig affected:  `CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE`

Steps to reproduce:

1. Turn-on/Plug keyboard
2. Turn off underglow with `&rgb_ug RGB_OFF` (if they were not on before, can turn on with `&rgb_ug RGB_ON` to then turn off)
3. Wait for keyboard to go idle
4. Wake keyboard up.

Actual: lights turn on at wake-up from idle, regardless of the state of the lights at beginning of idle.
Expected: At wake-up, the board turns on the underglow to the same state it was when idle began. So if lights were off,  they won't turn on at wake-up.

DominikHorn:
I am facing the same issue, is there any update on this?

B-urb:
I have the same issue and additionally set the CONFIG_ZMK_RGB_UNDERGLOW_ON_START to n. It still turns on after every idle.

dimosped:
same issue with these overrides:
https://github.com/dimosped/glove80-zmk-config/blob/main/config/glove80.conf

@B-urb  @DominikHorn  did you have any luck with any specific combo?

slashfoo:
I have not, I stopped using the feature when I couldn't find a workaround.

DominikHorn:
I also gave up and stopped using the feature

elpekenin:
Im case Andrew's message on the PR didnt cause a notification... I wrote some code to hopefully fix this, would be great if you guys can test it, i dont have any RGB board to do so 🥴

jradtilbrook:
I think this is fixed with #2244 (and follow-up #2284) and can be closed now.

PR #2244, opened 2024-04-02
fix(rgb): auto-off logic

elpekenin:
When going idle or disconnecting USB, the "is RGB on?" state was not properly stored/reverted when waking up, this should fix it.

Cant test it myself as i dont have a RGB board, but according to @awkannan on discord, it works 🤓

### DIFF
--- a/app/src/rgb_underglow.c
+++ b/app/src/rgb_underglow.c
@@ -469,27 +469,30 @@ int zmk_rgb_underglow_change_spd(int direction) {
 #if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE) ||                                          \
     IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_USB)
 struct rgb_underglow_sleep_state {
-    bool sleeping;
-    bool pre_sleep;
+    bool is_awake;
+    bool rgb_state_before_sleeping;
 };
 
-static int rgb_underglow_auto_state(bool is_wakeup_event) {
-    static struct rgb_underglow_sleep_state sleep_state = {sleeping : false, pre_sleep : false};
+static int rgb_underglow_auto_state(bool target_wake_state) {
+    static struct rgb_underglow_sleep_state sleep_state = {
+        is_awake : true,
+        rgb_state_before_sleeping : false
+    };
 
     // wake up event while awake, or sleep event while sleeping -> no-op
-    if (!is_wakeup_event == sleep_state.sleeping) {
+    if (target_wake_state == sleep_state.is_awake) {
         return 0;
     }
-    sleep_state.sleeping = !is_wakeup_event;
+    sleep_state.is_awake = target_wake_state;
 
-    if (is_wakeup_event) {
-        if (sleep_state.pre_sleep) {
+    if (sleep_state.is_awake) {
+        if (sleep_state.rgb_state_before_sleeping) {
             return zmk_rgb_underglow_on();
         } else {
             return zmk_rgb_underglow_off();
         }
     } else {
-        sleep_state.pre_sleep = sleep_state.on;
+        sleep_state.rgb_state_before_sleeping = sleep_state.on;
         return zmk_rgb_underglow_off();
     }
 }
--- a/app/src/rgb_underglow.c
+++ b/app/src/rgb_underglow.c
@@ -468,17 +468,28 @@ int zmk_rgb_underglow_change_spd(int direction) {
 
 #if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE) ||                                          \
     IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_USB)
-static int rgb_underglow_auto_state(bool *prev_state, bool new_state) {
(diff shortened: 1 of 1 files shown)


awkannan:
I tried a fix without tracking "sleep state" as introduced in this PR, but it didn't work.

My guess was that the event handlers run and call into the rgb_auto function multiple times before ending up in their final state, making our tracking of the previous user state of RGB inaccurate.

Adding the sleep state bits here seems to fix it.

Tested USB auto off behavior as well as IDLE auto off behavior and both seem to work.

Tested them together, and they both still seem to work as expected, but I was not exhaustive in my testing. In any case - the behavior in this PR is better than what it was before (if it's not perfect).

Should fix this: https://github.com/zmkfirmware/zmk/issues/1376

app/src/rgb_underglow.c
```diff
@@ -468,17 +468,28 @@ int zmk_rgb_underglow_change_spd(int direction) {
...
-        state.on = *prev_state;
-        *prev_state = false;
-        return zmk_rgb_underglow_on();
+    sleep_state.sleeping = !is_wakeup_event;
+
+    if (is_wakeup_event) {
+        if (sleep_state.pre_sleep) {
+            return zmk_rgb_underglow_on();
+        } else {
+            return zmk_rgb_underglow_off();
```
awkannan, line 489: we can probably remove this line / return 0 instead here

elpekenin: If my understanding was correct a few days ago while reading zephyr's docs about settings system, it checks if data is different before writing.

And i think having an explicit `off()` instrad of `return 0` is better because is much more obvious and readable on a quick glance, even though it is equivalent logic-wise.

app/src/rgb_underglow.c
```diff
@@ -468,17 +468,28 @@ int zmk_rgb_underglow_change_spd(int direction) {
...
+struct rgb_underglow_sleep_state {
+    bool sleeping;
+    bool pre_sleep;
+};
+
+static int rgb_underglow_auto_state(bool is_wakeup_event) {
+    static struct rgb_underglow_sleep_state sleep_state = {sleeping : false, pre_sleep : false};
+
+    // wake up event while awake, or sleep event while sleeping -> no-op
+    if (!is_wakeup_event == sleep_state.sleeping) {
```
petejohanson, line 480: To be honest, the naming of `is_wakeup_event` and then using `sleeping` as the state leads to this line which is really counterintuitive when reading quickly. I would much rather either:

`bool  target_wake_state` and then store a `bool awake` in the structure, so that we are comparing the same logical  state, and don't need negation as found here.

Thoughts?

app/src/rgb_underglow.c
```diff
@@ -468,17 +468,28 @@ int zmk_rgb_underglow_change_spd(int direction) {
 
 #if IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE) ||                                          \
     IS_ENABLED(CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_USB)
-static int rgb_underglow_auto_state(bool *prev_state, bool new_state) {
-    if (state.on == new_state) {
+struct rgb_underglow_sleep_state {
+    bool sleeping;
+    bool pre_sleep;
+};
+
```
petejohanson, line 475: The mix of the *sleep* state and the *rgb* state as bools with similar naming makes this hard to easily scan and understand.

I would  rather something like:

```suggestion
struct rgb_underglow_sleep_state {
    bool is_awake;
    bool rgb_state_before_sleeping;
};

```

See my other comment about inconsistent use of "awake" bool verses "sleep" bool to track/describe things.

petejohanson (commented):
A few thoughts on the PR. I generally like the approach to harmonize things between  the two possible ways the state changes.

@ReFil any thoughts?

ReFil:
Haven't got a board to test this on but it looks good

petejohanson (approved):
Thanks!

app/src/rgb_underglow.c
```diff
@@ -468,17 +468,31 @@ int zmk_rgb_underglow_change_spd(int direction) {
...
+    if (sleep_state.is_awake) {
+        if (sleep_state.rgb_state_before_sleeping) {
+            return zmk_rgb_underglow_on();
+        } else {
+            return zmk_rgb_underglow_off();
+        }
     } else {
-        state.on = false;
-        *prev_state = true;
+        sleep_state.rgb_state_before_sleeping = sleep_state.on;
```
jradtilbrook, line 495: Hey folks, I just ran a build of my firmware with underglow and its now failing. You can see the logs at https://github.com/jradtilbrook/zmk-config/actions/runs/8830708927/job/24244430141?check_suite_focus=true#step:11:143

Looks like the refactor is trying to reference a struct field that is not present to set the rgb status before sleeping. 

I think this might need to be updated to 
```suggestion
        zmk_rgb_underglow_get_state(&sleep_state.rgb_state_before_sleeping);
```

jradtilbrook: Or maybe just
```suggestion
        sleep_state.rgb_state_before_sleeping = sleep.on;
```

jradtilbrook: @petejohanson I've created #2284 to remedy this

PR #2254, opened 2024-04-05
Studio RPC Infrastructure

petejohanson:
Initial PR for the RPC infrastructure to be used by ZMK Studio.

This includes the behavior metadata commit from #2231 as it's used by the first really fleshed out behavior subsystem.

## To Do

- [x] Document framing and protocol buffer messages.
- [x] Wrap up the sample client library.
- [x] Testing/validation from others.


and-elf:
Why not start with a cli tool, that handles the actual rpc calls, and layer studio gui ontop of it later?
I'm not sure the gui (library) should be a part of this repo.

I'm not a maintainer, but it seems to me that the software isn't a part of the firmware. 
When I do RPC, I usually keep protocol definitions in an own repo, which server and client references.

megafloras:
A new user left a comment. This user must be approved by a Netlify team owner before comments can be displayed.

[Approve this user](https://app.netlify.com/teams/zmkfirmware/reviewers)

petejohanson:
> Nit:
> 
> > From client to device
> > Perhaps "Studio Client to ZMK Device" or something similar to establish familiarity if folks aren't used to RPC
> 
> Browser metadata
> 
> [Open Deploy Preview](https://deploy-preview-2254--zmk.netlify.app/docs/development/studio-rpc-protocol) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=6681afec9bd7f00008755ee4&commentID=66835ae97ae60bdd40909568&resolution=resolved)

Updated.

.github/workflows/build.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     if: ${{ always() }}
     runs-on: ubuntu-latest
     container:
-      image: docker.io/zmkfirmware/zmk-build-arm:3.5
+      image: docker.io/zmkfirmware/zmk-build-arm:3.5-branch
```
petejohanson, line 26: This is temporary while I get our `zmk-docker` images for the `3.5` tag to actually publish.

app/boards/arm/bdn9/bdn9_rev2.dts
```diff
@@ -109,6 +109,9 @@ zephyr_udc0: &usb {
     status = "okay";
     pinctrl-0 = <&usb_dm_pa11 &usb_dp_pa12>;
     pinctrl-names = "default";
+
+    // Overriden to lower RAM usage.
```
caksoylar, line 113:
```suggestion
    // Overridden to lower RAM usage.
```

app/dts/behaviors/studio_unlock.dtsi
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
caksoylar, line 2:
```suggestion
 * Copyright (c) 2024 The ZMK Contributors
```

caksoylar, line 2:
```suggestion
 * Copyright (c) 2024 The ZMK Contributors
```

caksoylar, line 2:
```suggestion
 * Copyright (c) 2024 The ZMK Contributors
```

There are a few other instances of this, e.g. in the binding yaml. Didn't note all not to pollute the review too much

app/dts/behaviors/studio_unlock.dtsi
```diff
@@ -0,0 +1,17 @@
...
+
+/ {
+    behaviors {
+        studio_unlock: studio_unlock {
+            compatible = "zmk,behavior-studio-unlock";
+            #binding-cells = <0>;
+            display-name = "Studio Unlock";
+        };
+    };
+
```
caksoylar, line 15:
```suggestion
```

app/src/studio/uuid.h
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2024 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#pragma once
+
+#include <zephyr/bluetooth/uuid.h>
+
+#define ZMK_BT_STUDIO_UUID(num) BT_UUID_128_ENCODE(num, 0x0196, 0x6107, 0xc967, 0xc5cfb1c2482a)
```
caksoylar, line 11: I see this was also used in the split uuid, but some comments would be nice for these magic numbers, if they can't be made a `#define`. Or maybe it is completely arbitrary?

app/src/studio/Kconfig
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) 2024 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+menuconfig ZMK_STUDIO
+    bool "Studio Support"
+    select NANOPB
+    # These two save stack size
+    imply NANOPB_NO_ERRMSG
+    imply  NANOPB_WITHOUT_64BIT
```
caksoylar, line 9:
```suggestion
    imply NANOPB_WITHOUT_64BIT
```

app/src/studio/Kconfig
```diff
@@ -0,0 +1,86 @@
...
+
+menuconfig ZMK_STUDIO_LOCKING
+    bool "Lock Support"
+    default y
+
+if ZMK_STUDIO_LOCKING
+
+config ZMK_STUDIO_LOCK_IDLE_TIMEOUT_SEC
+    int "Idle Timeout"
+    default 500
```
caksoylar, line 30: Maybe 600 for an even 10 minutes?

app/src/studio/Kconfig
```diff
@@ -0,0 +1,86 @@
...
+config ZMK_STUDIO_RPC_THREAD_STACK_SIZE
+    int "RPC Thread Stack Size"
+    default 4096
+
+config ZMK_STUDIO_RPC_RX_BUF_SIZE
+    int "RX Buffer Size"
+    default 30
+
+config ZMK_STUDIO_RPC_TX_BUF_SIZE
+    int "RX Buffer Size"
```
caksoylar, line 81:
```suggestion
    int "TX Buffer Size"
```

docs/src/docusaurus-tree-sitter-plugin/index.js
```diff
@@ -6,6 +6,7 @@
 
 module.exports = function () {
   return {
+    name: "tree-sitter",
```
caksoylar, line 9: Did this sneak in from somewhere or is it related to other docs changes?

docs/docs/behaviors/studio-unlock.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Studio Unlock Behavior
+sidebar_label: Studio Lock
```
caksoylar, line 3:
```suggestion
sidebar_label: Studio Unlock
```

docs/docs/behaviors/studio-unlock.md
caksoylar, line 1: Needs to go in sidebars.js to be visible (and behaviors overview as well, unless you intentionally omitted it)

docs/docs/behaviors/studio-unlock.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+:::warn
+ZMK Studio is still in active development. This behavior is documented in preparation for it's general availability.
```
caksoylar, line 7:
```suggestion
ZMK Studio is still in active development. This behavior is documented in preparation for its general availability.
```

docs/docs/development/studio-rpc-protocol.md
```diff
@@ -0,0 +1,137 @@
...
+title: Studio RPC Protocol
+---
+
+:::warning
+ZMK Studio is still in active development, and the below information is for development purposes only. For up to date information, join the [ZMK Discord](https://zmk.dev/community/discord/invite) server and discuss in `#studio-development`.
+:::
+
+## Overview
+
+The ZMK Studio UI communicates with ZMK devices using a custom RPC protocol developed to be robust, reliable, while remaining simple and easy te extend with future enhancements.
```
caksoylar, line 11:
```suggestion
The ZMK Studio UI communicates with ZMK devices using a custom RPC protocol developed to be robust, reliable, while remaining simple and easy to extend with future enhancements.
```

docs/docs/development/studio-rpc-protocol.md
```diff
@@ -0,0 +1,137 @@
...
+
+The protocol consist of [protocol buffer](https://protobuf.dev/programming-guides/proto3/) messages which are encoded/decoded using message framing, and the transmitted using an underlying transport. Two transports are currently implemented, a BLE transport using a custom GATT service, and a serial port transport, which usually is used with CDC ADM devices over USB.
```
caksoylar, line 13:
```suggestion
The protocol consists of [protocol buffer](https://protobuf.dev/programming-guides/proto3/) messages which are encoded/decoded using message framing, and then transmitted using an underlying transport. Two transports are currently implemented: a BLE transport using a custom GATT service and a serial port transport, which usually is used with CDC ADM devices over USB.
```

docs/docs/development/studio-rpc-protocol.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Protobuf Messages
+
+The messages for ZMK Studio are defined in a dedicated [zmk-studio-messages](https://github.com/zmkfirmware/zmk-studio-messages) repository. Fundamentally, the [`Request`](https://github.com/zmkfirmware/zmk-studio-messages/blob/main/proto/zmk/studio.proto#L11) message is used to send any requests from the Studio client to the ZMK device, and the [`Response`](https://github.com/zmkfirmware/zmk-studio-messages/blob/main/proto/zmk/studio.proto#L21) messages are sent from the ZMK device to the Studio client.
+
+Responses can either be [`RequestResponses`](https://github.com/zmkfirmware/zmk-studio-messages/blob/main/proto/zmk/studio.proto#L28) that are are sent in response to an incoming `Request`, or a [`Notification`](https://github.com/zmkfirmware/zmk-studio-messages/blob/main/proto/zmk/studio.proto#L38) which is sent at any point from the ZMK device to the Studio client to inform the client about state changes on the device, e.g. device is unlocked.
```
caksoylar, line 19:
```suggestion
Responses can either be [`RequestResponses`](https://github.com/zmkfirmware/zmk-studio-messages/blob/main/proto/zmk/studio.proto#L28) that are sent in response to an incoming `Request` or a [`Notification`](https://github.com/zmkfirmware/zmk-studio-messages/blob/main/proto/zmk/studio.proto#L38) which is sent at any point from the ZMK device to the Studio client to inform the client about state changes on the device, e.g. that the device is unlocked.
```

docs/docs/development/studio-rpc-protocol.md
```diff
@@ -0,0 +1,137 @@
...
+
+Studio uses a simple framing protocol to easily identify the start and end of a given message, with basic escaping to allow for unrestricted content.
+
+The following special bytes are used for the framing protocol:
+
+- Start of Frame (SoF): `0xAB`
+- Escape Byte (Esc): `0xAC`
+- End of Frame (EoF): `0xAD`
+
+A message consists of a SoF byte, the payload, escaped as needed, followed by an EoF byte. Within the payload, and of the special encoding bytes will be escaped by being prefixed with an Esc byte.
```
caksoylar, line 31:
```suggestion
A message consists of a SoF byte, the payload, escaped as needed, followed by an EoF byte. Within the payload, any of the special encoding bytes will be escaped by being prefixed with an Esc byte.
```

app/src/studio/core.c
```diff
@@ -0,0 +1,53 @@
...
+static enum zmk_studio_core_lock_state state = IS_ENABLED(CONFIG_ZMK_STUDIO_LOCKING)
+                                                   ? ZMK_STUDIO_CORE_LOCK_STATE_LOCKED
+                                                   : ZMK_STUDIO_CORE_LOCK_STATE_UNLOCKED;
+
+enum zmk_studio_core_lock_state zmk_studio_core_get_lock_state(void) { return state; }
+
+static void set_state(enum zmk_studio_core_lock_state new_state) {
+    state = new_state;
+
+    raise_zmk_studio_core_lock_state_changed(
```
caksoylar, line 21: Maybe worth a check to see if the `state != new_state` before raising the event?

caksoylar (commented):
FYI I haven't properly reviewed the core Studio parts (in src/studio) but went over the docs and the external interfaces

docs/src/docusaurus-tree-sitter-plugin/index.js
petejohanson: QoL issue.... I hit this when I thought I needed to upgrade our Docusaurus version, and it errored out due to this missing `name` property. Since I was there, figured I'd fix so future Docusaurus upgraders don't need to.

caksoylar: Thanks, good (intentional) change :)

app/src/studio/uuid.h
petejohanson: They are truly arbitrary. Other than one "reserved" 128-bit "group", vendor specific 128-bit UUIDs should just be generated randomly.

docs/docs/behaviors/studio-unlock.md
petejohanson: TBH, I'm not really wanting to "surface" the studio stuff too visibly until it's readier. Thoughts?

app/dts/behaviors/studio_unlock.dtsi
petejohanson: Fixed them all up, and added a few missing ones. Thanks.

docs/docs/behaviors/studio-unlock.md
caksoylar: I am fine with that, though we should note it (at least mentally) for when it is ready.

caksoylar (approved):
LGTM, with the previous caveat that I haven't looked deep in `src/studio/*.c`.

app/src/studio/behavior_subsystem.c
```diff
@@ -0,0 +1,216 @@
...
+static struct encode_metadata_sets_state state = {};
+
+zmk_Response get_behavior_details(const zmk_Request *req) {
+    uint32_t behavior_id = req->subsystem.behaviors.request_type.get_behavior_details.behavior_id;
+
+    const char *behavior_name = zmk_behavior_find_behavior_name_from_local_id(behavior_id);
+
+    if (!behavior_name) {
+        LOG_WRN("No behavior found for ID %d", behavior_id);
+        return ZMK_RPC_SIMPLE_ERR(GENERIC);
```
joelspadin, line 177: What does the `type` parameter to this macro do? I don't see any definition of `GENERIC` in this PR.

app/src/studio/core.c
```diff
@@ -0,0 +1,57 @@
...
+
+void zmk_studio_core_unlock() {
+    set_state(ZMK_STUDIO_CORE_LOCK_STATE_UNLOCKED);
+
+    zmk_studio_core_reschedule_lock_timeout();
+}
+
+void zmk_studio_core_lock() { set_state(ZMK_STUDIO_CORE_LOCK_STATE_LOCKED); }
+void zmk_studio_core_initiate_unlock() {
+    set_state(ZMK_STUDIO_CORE_LOCK_STATE_UNLOCKING);
```
joelspadin, line 53: Is unlocking some asynchronous task that takes some time to complete, or is this handling a request from Studio to the user to unlock the keyboard, which they confirm with the `&studio_unlock` behavior?

If the latter, would `_UNLOCK_REQUESTED` be a better name for this enum than `_UNLOCKING`? Also, should this reset back to locked after a timeout? It looks like it can just get stuck in this state forever if the user never responds. Or is this state even necessary? I don't see it referenced anywhere else, so I'm not sure what a handler of the "unlock requested" event is expected to do.

app/include/zmk/studio/rpc.h
```diff
@@ -0,0 +1,158 @@
...
+#include <zephyr/sys/ring_buffer.h>
+
+#include <proto/zmk/studio.pb.h>
+
+#include <zmk/endpoints_types.h>
+#include <zmk/event_manager.h>
+#include <zmk/studio/core.h>
+
+struct zmk_studio_rpc_notification {
+    zmk_Notification notification;
```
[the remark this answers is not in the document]
petejohanson: Yeah, I'm happy to namespace in `zmk.studio` for that proto file.

app/src/studio/Kconfig
```diff
@@ -0,0 +1,86 @@
...
+    select BT_USER_DATA_LEN_UPDATE
+    depends on ZMK_BLE
+    default y
+
+config BT_CONN_TX_MAX
+    default 64 if ZMK_STUDIO_TRANSPORT_BLE
+
+config ZMK_STUDIO_TRANSPORT_BLE_PREF_LATENCY
+    int "BLE Transport preferred latency"
+    default 10
```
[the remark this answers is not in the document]
petejohanson: We correctly make no change if it matches... I'll check about greater/less than case.

app/src/studio/core.c
petejohanson: This is a half implemented feature where I could see a keyboard with a display showing a prompt to the user about unlocking, allowing input to respond. In that case, it's an async workflow that's initiated, then in progress, then aborted or completed. I can scrap for now instead of partially including, if you'd prefer.

app/CMakeLists.txt
```diff
@@ -105,4 +106,27 @@ target_sources(app PRIVATE src/main.c)
...
+
+    # Turn off the default nanopb behavior
+  set(NANOPB_GENERATE_CPP_STANDALONE OFF)
+
+  nanopb_generate_cpp(proto_srcs proto_hdrs RELPATH ${ZEPHYR_ZMK_STUDIO_MESSAGES_MODULE_DIR}
+                        ${ZEPHYR_ZMK_STUDIO_MESSAGES_MODULE_DIR}/proto/zmk/studio.proto
+                        ${ZEPHYR_ZMK_STUDIO_MESSAGES_MODULE_DIR}/proto/zmk/meta.proto
+                        ${ZEPHYR_ZMK_STUDIO_MESSAGES_MODULE_DIR}/proto/zmk/core.proto
+                        ${ZEPHYR_ZMK_STUDIO_MESSAGES_MODULE_DIR}/proto/zmk/behaviors.proto
+                        ${ZEPHYR_ZMK_STUDIO_MESSAGES_MODULE_DIR}/proto/zmk/keymap.proto
```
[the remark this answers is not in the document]
petejohanson: So... I dug on this a lot.. The general take in cmake in general seems to be "no wildcards", and the build stuff for nanopb really doesn't support it in the first place.

app/include/zmk/studio/rpc.h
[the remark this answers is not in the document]
petejohanson: Pushed this refactor, these are now in the `zmk.studio` package.

app/include/zmk/studio/rpc.h
```diff
@@ -0,0 +1,158 @@
...
+                if (sub_handler->secured &&                                                        \
+                    zmk_studio_core_get_lock_state() != ZMK_STUDIO_CORE_LOCK_STATE_UNLOCKED) {     \
+                    return ZMK_RPC_RESPONSE(meta, simple_error,                                    \
+                                            zmk_meta_ErrorConditions_UNLOCK_REQUIRED);             \
+                }                                                                                  \
+                return sub_handler->func(req);                                                     \
+            }                                                                                      \
+        }                                                                                          \
+        LOG_ERR("No handler func found for %d", which_req);                                        \
+        return ZMK_RPC_RESPONSE(meta, simple_error, zmk_meta_ErrorConditions_RPC_NOT_FOUND);       \
```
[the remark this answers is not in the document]
petejohanson: Pulled 95% of this out into a shared function, definitely an improvement. Won't mark this resolved until you see/review the updated code. Thanks.

app/src/studio/behavior_subsystem.c
```diff
@@ -0,0 +1,216 @@
...
+        case BEHAVIOR_PARAMETER_VALUE_TYPE_NIL:
+            desc.which_value_type = zmk_behaviors_BehaviorParameterValueDescription_nil_tag;
+            break;
+        case BEHAVIOR_PARAMETER_VALUE_TYPE_HID_USAGE:
+            desc.which_value_type = zmk_behaviors_BehaviorParameterValueDescription_hid_usage_tag;
+#if IS_ENABLED(CONFIG_ZMK_HID_CONSUMER_REPORT_USAGES_BASIC)
+            desc.value_type.hid_usage.consumer_max = 0xFF;
+#else
+            desc.value_type.hid_usage.consumer_max = 0xFFF;
+#endif
```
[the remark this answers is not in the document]
petejohanson: Pulled out two defines into `hid.h` and used those directly, and tweaked the hid.c `TOGGLE_CONSUMER` to use the one as well.

app/src/studio/behavior_subsystem.c
```diff
@@ -0,0 +1,216 @@
...
+            desc.which_value_type = zmk_behaviors_BehaviorParameterValueDescription_layer_id_tag;
+            break;
+        default:
+            LOG_ERR("Unknown value description type %d", val->type);
+            return false;
+        }
+
+        if (!pb_encode_submessage(stream, &zmk_behaviors_BehaviorParameterValueDescription_msg,
+                                  &desc)) {
+            LOG_WRN("Failed to encode submessage!");
```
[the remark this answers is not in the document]
petejohanson: Added.

app/include/zmk/studio/rpc.h
```diff
@@ -0,0 +1,158 @@
...
+};
+
+struct zmk_rpc_subsystem_handler {
+    rpc_func *func;
+    uint8_t subsystem_choice;
+    uint8_t request_choice;
+    bool secured;
+};
+
+#define ZMK_RPC_NO_RESPONSE() ZMK_RPC_RESPONSE(meta, no_response, true)
```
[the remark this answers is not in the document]
petejohanson: Documented most of the relevant ones in that file. Hopefully that helps clear things up a bit? Included several links to the relevant proto messages as references.

app/src/studio/Kconfig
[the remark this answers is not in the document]
petejohanson: Switched the condition to be less than so we don't do it unnecessarily. Good call.

app/include/zmk/studio/rpc.h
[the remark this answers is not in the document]
joelspadin: Much easier to understand.

app/include/zmk/studio/rpc.h
```diff
@@ -0,0 +1,209 @@
...
+        .subsystem_choice = zmk_studio_Request_##prefix##_tag,                                     \
+    };
+
+/**
+ * @brief Register an RPC subsystem handler handler a specific request within the subsystem.
+ * @param prefix The identifier for the subsystem, e.g. `core`, `keymap`, etc.
+ * @param request_id The identifier for the request ID, e.g. `save_changes`.
+ * @param _secured Whether the handler requires the device be unlocked to allow invocation.
+ *
+ * @note  A function with a name matching the request_id must be in-scope and will be used as the
```
joelspadin, line 88: Is this callback just a `void()` (seems unlikely)? The function signature or a typedef to a function pointer type would be useful here.

app/include/zmk/studio/rpc.h
```diff
@@ -0,0 +1,209 @@
...
+    STRUCT_SECTION_ITERABLE(zmk_rpc_subsystem, prefix##_subsystem) = {                             \
+        .func = subsystem_func_##prefix,                                                           \
+        .subsystem_choice = zmk_studio_Request_##prefix##_tag,                                     \
+    };
+
+/**
+ * @brief Register an RPC subsystem handler handler a specific request within the subsystem.
+ * @param prefix The identifier for the subsystem, e.g. `core`, `keymap`, etc.
+ * @param request_id The identifier for the request ID, e.g. `save_changes`.
+ * @param _secured Whether the handler requires the device be unlocked to allow invocation.
```
joelspadin, line 86: Consider making this an enum (possibly a bit field if there might be other options for this later, but that can easily be changed after the fact too).

For example:
```c
// Can't tell what the last parameter means at a glance
ZMK_RPC_SUBSYSTEM_HANDLER(behaviors, list_all_behaviors, false);
ZMK_RPC_SUBSYSTEM_HANDLER(behaviors, get_behavior_details, true);

// Immediately obvious
ZMK_RPC_SUBSYSTEM_HANDLER(behaviors, list_all_behaviors, ZMK_RPC_UNSECURE);
ZMK_RPC_SUBSYSTEM_HANDLER(behaviors, get_behavior_details, ZMK_RPC_SECURE);
```

app/src/studio/core.c
```diff
@@ -0,0 +1,57 @@
...
+    raise_zmk_studio_core_unlock_requested((struct zmk_studio_core_unlock_requested){});
+}
+
+void zmk_studio_core_complete_unlock() { set_state(ZMK_STUDIO_CORE_LOCK_STATE_UNLOCKED); }
```
joelspadin, line 57: If the keyboard is unlocked via this function, it won't re-lock with a timeout. Is that intentional?

app/src/studio/core.c
joelspadin: My first concern is that we have support for locking from unlocked after a timeout, but not this middle state. What happens if Studio requests an unlock, then without unlocking on the keyboard, you restart Studio? It looks like the keyboard gets stuck in this "unlocking" state unless CONFIG_ZMK_STUDIO_LOCK_ON_DISCONNECT is set? Might be a bit confusing if so.

My second and probably more concerning concern is that
```c
if (zmk_studio_core_get_lock_state() != ZMK_STUDIO_CORE_LOCK_STATE_LOCKED) {
  // we're unlocked, so allow things
}
```
looks perfectly reasonable at first glance yet is completely wrong. Would separate bools be any better? That does allow for the not very meaningful state of "is unlocked and unlock is requested", but it would make the above mistake impossible.

app/include/zmk/studio/rpc.h
petejohanson: I like it, just tweaked the naming a bit to use `ZMK_STUDIO_RPC_HANDLER_SECURED`.

app/src/studio/core.c
petejohanson: Nope. Will add the timeout reschedule for this too.

petejohanson:
> My first concern is that we have support for locking from unlocked after a timeout, but not this middle state. What happens if Studio requests an unlock, then without unlocking on the keyboard, you restart Studio? It looks like the keyboard gets stuck in this "unlocking" state unless CONFIG_ZMK_STUDIO_LOCK_ON_DISCONNECT is set? Might be a bit confusing if so.

I think adding a timeout for "unlocking" is a good idea. But yes, it does potentially present some funky states.

> 
> My second and probably more concerning concern is that
> 
> if (zmk_studio_core_get_lock_state() != ZMK_STUDIO_CORE_LOCK_STATE_LOCKED) {
>   // we're unlocked, so allow things
> }
> 
> looks perfectly reasonable at first glance yet is completely wrong. Would separate bools be any better? That does allow for the not very meaningful state of "is unlocked and unlock is requested", but it would make the above mistake impossible.

This is where I wish for proper tagged enums like Rust, where we could represent this sub-state properly, e.g.:

```
enum LockedState {
  Locked,
  Unlocking,
}

enum LockState {
  Locked(LockedState),
  Unlocked,
}
```

I think your idea of separate bools may be the best we can do. Let me think on this a bit though.

petejohanson:
Thinking about it... I think I will just axe the "Unlocking" state, we don't even need it yet, and there's no point holding this up on that one aspect.

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+
+    uint8_t notify_bytes[notify_size];
+
+    while (ring_buf_size_get(tx_buf) > 0) {
+        uint8_t added = 0;
+        while (added < notify_size && ring_buf_size_get(tx_buf) > 0) {
+            uint8_t *buf;
+            int len = ring_buf_get_claim(tx_buf, &buf, notify_size - added);
+
+            if (len > 0) {
```
joelspadin, line 149: Is len == 0 possible while ring_buf_size_get() > 0?

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+
+            added += len;
+            ring_buf_get_finish(tx_buf, len);
+        }
+
+        rpc_indicate_params.data = notify_bytes;
+        rpc_indicate_params.len = added;
+
+        uint8_t notify_attempts = 5;
+        do {
```
joelspadin, line 161: do-while works, but a for loop might be more conventional? Can also just use an int for the counter so you don't have to explicitly avoid it going negative.

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+    uint8_t notify_size = 23; // Default MTU size unless negotiated higher
+    struct bt_conn_info conn_info;
+    if (bt_conn_get_info(conn, &conn_info) >= 0) {
+        notify_size = conn_info.le.data_len->tx_max_len;
+    }
+
+    uint8_t notify_bytes[notify_size];
+
+    while (ring_buf_size_get(tx_buf) > 0) {
+        uint8_t added = 0;
```
joelspadin, line 144: Any reason to specifically use uint8_t here instead of something like int? (We're right between a couple of probably large variables on the stack, so probably not going to be packing variables together and saving stack size here.)

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+    struct bt_conn *conn = zmk_ble_active_profile_conn();
+    struct ring_buf *tx_buf = zmk_rpc_get_tx_buf();
+
+    if (!conn) {
+        LOG_WRN("No active connection for queued data, dropping");
+        ring_buf_reset(tx_buf);
+        return;
+    }
+
+    uint8_t notify_size = 23; // Default MTU size unless negotiated higher
```
joelspadin, line 135: tx_max_len is a uint16_t, this has potential to overflow.

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+static K_WORK_DEFINE(notify_tx_work, notif_rpc_tx_cb);
+
+struct gatt_write_state {
+    uint8_t pending_notify;
+};
+
+static void gatt_tx_notify(struct ring_buf *tx_buf, size_t added, bool msg_done, void *user_data) {
+    struct gatt_write_state *state = (struct gatt_write_state *)user_data;
+
+    state->pending_notify += added;
```
joelspadin, line 184: We are writing a size_t to a uint8_t. Can we guarantee that this is always <= 255?

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+static ssize_t write_rpc_req(struct bt_conn *conn, const struct bt_gatt_attr *attr, const void *buf,
+                             uint16_t len, uint16_t offset, uint8_t flags) {
+    if (handling_rx) {
+        uint32_t copied = 0;
+        struct ring_buf *rpc_buf = zmk_rpc_get_rx_buf();
+        do {
+            uint8_t *buffer;
+            uint32_t claim_len = ring_buf_put_claim(rpc_buf, &buffer, len - copied);
+
+            if (claim_len > 0) {
```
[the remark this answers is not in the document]
petejohanson: So... It will only infinitely look *if* there's no consumer pulling out of the ring buffer. I will tweak this to be a regular while loop though to be safe.

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+                memcpy(buffer, ((uint8_t *)buf) + copied, claim_len);
+                copied += claim_len;
+            }
+
+            ring_buf_put_finish(rpc_buf, claim_len);
+        } while (len - copied > 0);
```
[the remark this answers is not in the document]
petejohanson: I've tweaked it to the simpler `copied < len`

app/src/studio/gatt_rpc_transport.c
petejohanson: I don't believe so, on looking at this again. I'll simplify.

app/src/studio/gatt_rpc_transport.c
petejohanson: Not 100%. Updated to just store a `size_t` in the state.

app/src/studio/gatt_rpc_transport.c
petejohanson: Updated to `uint16_t` everywhere.

app/src/studio/gatt_rpc_transport.c
petejohanson: Nope. Changed.

app/src/studio/gatt_rpc_transport.c
```diff
@@ -0,0 +1,222 @@
...
+        LOG_WRN("No active connection for queued data, dropping");
+        ring_buf_reset(tx_buf);
+        return;
+    }
+
+    uint8_t notify_size = 23; // Default MTU size unless negotiated higher
+    struct bt_conn_info conn_info;
+    if (bt_conn_get_info(conn, &conn_info) >= 0) {
+        notify_size = conn_info.le.data_len->tx_max_len;
+    }
```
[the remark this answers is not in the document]
petejohanson: Moved to a shared function.

app/src/studio/gatt_rpc_transport.c
petejohanson: Switched to an int... I personally favor do/while style for "retries" like this.

app/src/studio/uart_rpc_transport.c
```diff
@@ -0,0 +1,167 @@
...
+        /* read until FIFO empty */
+        uint32_t last_read = 0;
+        struct ring_buf *buf = zmk_rpc_get_rx_buf();
+        do {
+            uint8_t *buffer;
+            uint32_t len = ring_buf_put_claim(buf, &buffer, buf->size);
+            if (len == 0) {
+                zmk_rpc_rx_notify();
+                continue;
+            }
```
joelspadin, line 113: Does this just busy loop until it gets a claim? If so, I think this would need to be running on a thread with a lower priority than whatever is consuming the data, or it may deadlock by never allowing that thread to run?

app/src/studio/uart_rpc_transport.c
```diff
@@ -0,0 +1,167 @@
...
+
+        zmk_rpc_rx_notify();
+    }
+
+    if (uart_irq_tx_ready(uart_dev)) {
+        struct ring_buf *tx_buf = zmk_rpc_get_tx_buf();
+        uint8_t len;
+        while ((len = ring_buf_size_get(tx_buf)) > 0) {
+            uint8_t *buf;
+            uint8_t claim_len = ring_buf_get_claim(tx_buf, &buf, tx_buf->size);
```
joelspadin, line 127: Wrong size.

app/src/studio/uart_rpc_transport.c
```diff
@@ -0,0 +1,167 @@
...
+
+            ring_buf_put_finish(buf, last_read);
+        } while (last_read > 0);
+
+        zmk_rpc_rx_notify();
+    }
+
+    if (uart_irq_tx_ready(uart_dev)) {
+        struct ring_buf *tx_buf = zmk_rpc_get_tx_buf();
+        uint8_t len;
```
joelspadin, line 124: Wrong size.

app/src/studio/uart_rpc_transport.c
petejohanson: A good question... The `zmk_rpc_rx_notify` will give on the underlying semaphore, but this is in an ISR context for the UART callback, IIRC, so we can't yield... I'm honestly not sure if we're hitting this (yet) in practice, and if so, how it's functioning. Let me dig a bit.

petejohanson: Ugh... Ok, so... I think the answer is: We need to size the RX ring buffer large enough to avoid this happening. From https://docs.zephyrproject.org/3.5.0/doxygen/html/group__uart__interrupt.html#gab10942076ac47ecff29e924098049398:

> That means that once [uart_irq_rx_ready()](https://docs.zephyrproject.org/3.5.0/doxygen/html/group__uart__interrupt.html#gad04073b1b8e3de13b43ae1194561377b) is detected, [uart_fifo_read()](https://docs.zephyrproject.org/3.5.0/doxygen/html/group__uart__interrupt.html#gab10942076ac47ecff29e924098049398) must be called until it reads all available data in the FIFO (i.e. until it returns less data than was requested).

So... we *have* to keep calling `uart_fifo_read` from this callback, even if we don't have anywhere to put it... And we're an an ISR so we can't just yield. I think the best we can actually do in this scenario is skip the bytes `LOG_ERR`
that we're doing so, and bail.

Thoughts?

petejohanson: Yup, here's a pretty authoritative Zephyr use of the async API for the shell work. It basically drops the data if there's no room in the ring buffer:

https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/shell/backends/shell_uart.c#L84

(modified to remove some conditional code for clarity)

```
	do {
		len = ring_buf_put_claim(&sh_uart->rx_ringbuf, &data,
					 sh_uart->rx_ringbuf.size);

		if (len > 0) {
			rd_len = uart_fifo_read(dev, data, len);

			/* If there is any new data to be either taken into
			 * ring buffer or consumed by the SMP, signal the
			 * shell_thread.
			 */
			if (rd_len > 0) {
				new_data = true;
			}

			int err = ring_buf_put_finish(&sh_uart->rx_ringbuf, rd_len);
			(void)err;
			__ASSERT_NO_MSG(err == 0);
		} else {
			uint8_t dummy;

			/* No space in the ring buffer - consume byte. */
			LOG_WRN("RX ring buffer full.");

			rd_len = uart_fifo_read(dev, &dummy, 1);
		}
	} while (rd_len && (rd_len == len));
```

petejohanson: Updated to match.

app/src/studio/msg_framing.h
```diff
@@ -0,0 +1,23 @@
...
+    FRAMING_STATE_ESCAPED,
+    FRAMING_STATE_ERR,
+    FRAMING_STATE_EOF,
+};
+
+#define FRAMING_SOF 0xAB
+#define FRAMING_ESC 0xAC
+#define FRAMING_EOF 0xAD
+
+bool studio_framing_process_byte(enum studio_framing_state *frame_state, uint8_t data);
```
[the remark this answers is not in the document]
petejohanson: Added some basic docs here.

app/src/studio/msg_framing.c
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#include "msg_framing.h"
+
+bool studio_framing_process_byte(enum studio_framing_state *rpc_framing_state, uint8_t c) {
```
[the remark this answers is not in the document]
petejohanson: Good call. Done.

app/src/studio/rpc.c
```diff
@@ -0,0 +1,339 @@
...
+
+    STRUCT_SECTION_FOREACH(zmk_rpc_transport, t) {
+        if (t->transport == transport) {
+            selected_transport = t;
+            if (selected_transport->rx_start) {
+                selected_transport->rx_start();
+            }
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
petejohanson: Tweaked.

Nick-Munnich:
Just as a small note, could you either finish (there's a TODO) & add the config and behavior doc pages to the sidebar or remove them from this PR? The RPC page looks good.

petejohanson:
> Just as a small note, could you either finish (there's a TODO) & add the config and behavior doc pages to the sidebar or remove them from this PR? The RPC page looks good.

I will move the *config* page to #2268 since that PR makes studio actually end user usable. The RPC docs I will add to the sidebar in this PR since it's part of this core work. That work @Nick-Munnich ?

Nick-Munnich:
> I will move the _config_ page to #2268 since that PR makes studio actually end user usable. The RPC docs I will add to the sidebar in this PR since it's part of this core work. That work @Nick-Munnich ?
Moving them is a good solution, but be aware that you've already added the RPC docs, the part which was added but not listed in the sidebar is "Studio Unlock" behaviour. I'd guess that should be moved too?

petejohanson:
> > I will move the _config_ page to #2268 since that PR makes studio actually end user usable. The RPC docs I will add to the sidebar in this PR since it's part of this core work. That work @Nick-Munnich ?
> 
> Moving them is a good solution, but be aware that you've already added the RPC docs, the other part which was added but not listed in the sidebar is "Studio Unlock" behaviour. I'd guess that should be moved too?

Yeah, I saw RPC was already in the sidebar. I've left that for now, but moved the config doc to #2268 and added in the sidebar there.

I wanted to be sure I had the unlock behavior documented, but don't really want to "expose it" until it's actually usable for something. I'd like to keep the doc, but add to sidebar in #2268. Any concerns?

Nick-Munnich:
> I wanted to be sure I had the unlock behavior documented, but don't really want to "expose it" until it's actually usable for something. I'd like to keep the doc, but add to sidebar in #2268. Any concerns?

Best practice would probably be to add it to sidebar and then comment it out in this PR, but I think that's not necessary unless you wanted to. All good from me.

petejohanson:
> > I wanted to be sure I had the unlock behavior documented, but don't really want to "expose it" until it's actually usable for something. I'd like to keep the doc, but add to sidebar in #2268. Any concerns?
> 
> Best practice would probably be to add it to sidebar and then comment it out in this PR, but I think that's not necessary unless you wanted to. All good from me.

Given the type check from Docusaurus is now angry about unlock docs referencing the config docs that don't exist, I'm just going to move that doc to #2268 also for cohesion.

PR #2256, opened 2024-04-06
Cleanup battery code

joelspadin:
Changed the zmk,battery-voltage-divider and zmk,battery-nrf-vddh drivers to support only SENSOR_CHAN_VOLTAGE, and changed the default battery fetch mode to ZMK_BATTERY_REPORTING_FETCH_MODE_LITHIUM_VOLTAGE. This allowed for removing the copy of `lithium_ion_mv_to_pct()` from the driver code, leaving only the version in battery.c.

Custom boards/shields that use a fuel gauge driver instead of a voltage sensor, will need the following added to their Kconfig.defconfig:

```
choice ZMK_BATTERY_REPORTING_FETCH_MODE
    default ZMK_BATTERY_REPORTING_FETCH_MODE_STATE_OF_CHARGE
endchoice
```

Alternatively, add the following to the _defconfig file:

```ini
ZMK_BATTERY_REPORTING_FETCH_MODE_STATE_OF_CHARGE=y
```

Also updated the battery docs to include missing options and recent changes to the battery code.


app/Kconfig
```diff
@@ -388,11 +388,17 @@ if ZMK_BATTERY_REPORTING
...
-config ZMK_BATTERY_REPORTING_FETCH_MODE_STATE_OF_CHARGE
-    bool "State of charge"
-
 config ZMK_BATTERY_REPORTING_FETCH_MODE_LITHIUM_VOLTAGE
     bool "Lithium Voltage"
+    help
+        Estimate the battery level by reading the voltage of a Lithium battery.
+        zmk,battery chosen node should be a sensor driver that implements SENSOR_CHAN_VOLTAGE.
+
+config ZMK_BATTERY_REPORTING_FETCH_MODE_STATE_OF_CHARGE
```
petejohanson, line 397: To avoid breaking existing out of tree boards that use a fuel gauge, I wonder if it might be reasonable to default this choice selection *if* our max17048 driver is enabled? Seems pragmatic until we can get the few out of tree boards to update their setup, then we remove that temporary work around?

petejohanson (changes requested):
Code seems reasonable to me. I wonder if we should update any of our in tree boards to the Zephyr voltage divider driver as well, to allow us to *really* test that and give folks some examples to work from that are updated?

Issue #2210, opened 2024-03-18
Right side of Corne suddenly dead.


jasper-at-windswept:
I was just configuring and flashing my Corne with Nice!Nano V2's and then after some point the right side won't connect to the left anymore, however its not a pairing issue because i tried plugging both the sides into USB and the right side specifically does not work.

I have triple checked the hardware and nothing has changed, i can flash firmware onto it with no issues. Getting desperate i reflashed both of the bootloader entirely and flashed the default ZMK Corne configuration but it did not work, I put my config back on and tried the ZMK guide for settings_resetting to fix what i thought at that time was a pairing issue but with no success!

What on earth has happened, can someone please help me!

jasper-at-windswept:
> The peripheral half will not send keystrokes over USB; it only communicates with the central half. So if you can still flash over USB (you say you can) and the battery still charges when plugged into USB (you haven't indicated this is a problem), that's all business as usual.
> 
> You could [enable USB logging](https://zmk.dev/docs/development/usb-logging) to gather more information, or perform a more complete flash erase with joric's [erase_flash.uf2](https://github.com/joric/nrfmicro/releases/download/1.4/erase_flash.uf2), but from where I'm standing it looks like you still have a pairing problem.

@lesshonor 
Thankyo very much for getting back to me.
I just tried the erase_flash.uf2 and then my previous working firmware but without success. I have also added a BT_CLR_ALL keybinding and the left side connects to my computer without issues but the right side still does not send keystrokes to the computer. What doesn't make sense is when i enable the display, firstly the left display won't turn on (which it used to before the right side stopped sending key inputs) but the right side display does work and it displays that both the sides are paired! Yet it isn't sending keystrokes, any ideas?

jasper-at-windswept:
Something I should mention. I can't remember properly but i think before the right side stopped working I tried to flash my firmware using a modifed `zmkfirmware` repository. The firmware did build with no issues but when i flashed it to both of the sides, neither of them worked at all. After re-installing the bootloader, flashing them with my firmware again, then trying the settings_reset method the left side would work but not the right side.
I truly hope there is some solution to this!

jasper-at-windswept:
@lesshonor 
Here are those logs from the left side of the board from after attempting to pair both the boards.

```
[08:54:17.357] Disconnected
[08:54:18.358] Warning: Could not open tty device (No such file or directory)
[08:54:18.358] Waiting for tty device..
[08:54:19.359] Connected
[...]

PR #2262, opened 2024-04-07
Build CI fixes/improvements

lesshonor:
[All `push` event runs of the Build action since 2024-01-25 have failed](https://github.com/zmkfirmware/zmk/actions/workflows/build.yml?query=event%3Apush+is%3Afailure) in part[^1] because `actions/checkout` is not being run prior to `tj-actions/changed-files`.

~~This primarily presents a problem for novice end-users looking for `settings_reset` firmware [as linked from the documentation](https://zmk.dev/docs/troubleshooting#option-2-download-reset-uf2-from-zmks-workflow). The outdated binaries appear to be entirely ineffective on Zephyr 3.5 firmware[^2].~~ All artifacts containing the outdated firmware have now expired.

<details>
<summary>Remaining commits are general improvements to security and efficiency:</summary>

### `concurrency`

Group builds by workflow and [github ref](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context).
> The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by `push`, this is the branch or tag ref that was pushed. For workflows triggered by `pull_request`, this is the pull request merge branch. 

`ref_name` is another possibility, but could collide with any eventual tags.

These groupings allow us to...

### `cancel-in-progress`

...cancel workflow runs which are subsequently obsoleted by later pushes.

In an ideal world, anyway. Unfortunately, the way ZMK currently handles the final `Build` and `Upload Artifact` steps means the cancellation of those particular steps is not immediate. If the five-minute post-cancellation timeout passes, something will error out. If not: the job will grind on to completion and display as "Cancelled".

This is something worth examining as part of a re-evaluation of CI generally, but it's out of scope for now. Any kind of "stop doing unnecessary work" constitutes improvement.

### `permissions`

Given what this workflow currently does, its `GITHUB_TOKEN` doesn't need anything more than read-only access to metadata.

### `persist-credentials`

It also doesn't need to keep the  `zmkfirmware` credentials handy.

### etc

- If the `compile-matrix` job has no inputs because the previous jobs were canceled, might as well cancel it.
- I don't see why forks should run nightly builds.
</details>

Suggestions welcome. I have been testing these changes [in my own fork](https://github.com/lesshonor/zmk/actions/workflows/build.yml).

[^1]: It might be "in total"; I haven't checked every run.
[^2]: #2210, [discord](https://discord.com/channels/719497620560543766/1221880567373496433/1221880567373496433), [discord](https://discord.com/channels/719497620560543766/1219493309429321728/1225906079393644626)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -413,7 +413,9 @@ jobs:
       board-changes: ${{ steps.board-changes.outputs.result }}
       core-changes: ${{ steps.core-changes.outputs.result }}
     steps:
-      - uses: tj-actions/changed-files@v42
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: tj-actions/changed-files@v44
         id: changed-files
         with:
           json: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,8 @@ on:
   schedule:
     - cron: "22 4 * * *"
 
+permissions: {}
+
 jobs:
   build:
     if: ${{ always() }}
@@ -25,6 +27,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - name: Cache west modules
         uses: actions/cache@v4
         env:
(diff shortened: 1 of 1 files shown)


caksoylar:
Might be worth pointing out this is a fix, maybe it will get higher priority during review?

petejohanson (approved):
Thanks for the CI improvements, and my apologies for the review delay. LGTM!

PR #2266, opened 2024-04-10
Update tg4x.overlay

shaggnanigans:
## Board/Shield Check-list


caksoylar:
I assume this change is to accomodate for the smaller right shift layout, where the bottom row has 12 keys. The current README states it is for "2u right shift": https://github.com/zmkfirmware/zmk/tree/main/app/boards/shields/tg4x/#board-revision-and-layout-notes

The right way to handle this would be to define multiple matrix transforms, corresponding to the different physical layout options. Having said that, shield PRs seem to be on a bit of a hiatus; so you might want to just create a custom shield in your config repo with these changes.

PR #437, opened 2020-11-27
Slice up the development environment setup

mcrosson:
- Give a brief overview of docker+vscode vs native os
- Dedicated page for docker+vscode setup
- Dedicated page for native os setup
- Dedicated page for setting up zmk sources for development

PR #2272, opened 2024-04-15
feat(docs): Improve the toolchain setup page

Nick-Munnich:
Making the [changes discussed on Discord](https://discord.com/channels/719497620560543766/883452966114324550/1229127206715002911). In particular:

 * Removing the 'install west' section
 * Renaming "Install Dependencies" to "Install Zephyr and West".
 * Consolidating the Linux based operating systems which aren't Raspberry OS to one tab

Further improvement could be made by identifying a better way to handle Raspberry OS, since it is an outlier, but I feel that this is good enough for now.


docs/docs/development/setup.mdx
```diff
@@ -273,50 +185,5 @@ This step pulls down quite a bit of tooling. Go grab a cup of coffee, it can tak
...
-
-```sh
-pip3 install --user -r zephyr/scripts/requirements.txt
-```
-
-</TabItem>
-<TabItem value="win">
-
-```sh
-pip3 install -r zephyr/scripts/requirements.txt
```
petejohanson, line 311: My main concern with this is related to my comment here https://github.com/zmkfirmware/zmk/pull/2216#issuecomment-2049051599 about really only needing `requirements-base.txt` deps, and our Docker images not having everything needed to install any "extras" that would get pulled in by following the upstream Zephyr docs.

Nick-Munnich: I see - valid point, I was just refactoring what is already there. I've added a note on this point, that's the best solution I can think of.

docs/docs/development/setup.mdx
```diff
@@ -203,21 +169,7 @@ Zephyr™ application is in the `app/` source directory:
...
-```
-
-</TabItem>
-<TabItem value="raspberryos">
-
-```sh
-cd zmk
-```
-
-</TabItem>
```
caksoylar, line 219: I suppose this was deleted by accident, currently it renders empty under the tab.

docs/docs/development/setup.mdx
```diff
@@ -85,10 +115,22 @@ Return to this guide once you are finished with each section.
...
-- [Install Dependencies](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html#install-dependencies)
+1. [Select and Update OS](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html#select-and-update-os)
+2. [Install Dependencies](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html#install-dependencies)
+3. [Get Zephyr and install Python dependencies](https://docs.zephyrproject.org/latest/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies)
+4. [Install Zephyr SDK](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html#install-zephyr-sdk)
+
+Return to this guide once you are finished with all four sections.
+
+:::info
+ZMK does not need all of the additional dependencies found in Zephyr's `requirements.txt`, only those found in `requirements-base.txt`
```
caksoylar, line 126:
```suggestion
ZMK does not need all of the additional dependencies found in Zephyr's `requirements.txt`, but only those found in `requirements-base.txt`:
```
Small improvement suggestion (for all 4 tabs).

caksoylar (commented):
This looks like a good improvement to me in general, to reduce confusion on what steps to follow and simply our docs.

I have one concern regarding following step 3 (Get Zephyr and install Python dependencies), it also makes you clone Zephyr and run `west zephyr-export` on the clone. Since fetching Zephyr will be handled by `west update` on the ZMK repo, we should skip that. How about adding some sub-bullet points like below:

```md
3. [Get Zephyr and install Python dependencies](https://docs.zephyrproject.org/latest/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies)
  - You can skip the steps starting with "Get the Zephyr source code" and "Export a Zephyr CMake package" since ZMK will fetch its own copy of Zephyr automatically
  - ZMK also does not need all of the additional dependencies found in Zephyr's `requirements.txt`, but only those found in `requirements-base.txt`. So you can modify parts of the related commands like below:

    ```diff
    - pip3 install -r zephyr/scripts/requirements.txt
    + pip3 install -r zephyr/scripts/requirements-base.txt
    ```
```
Note that I moved the extra info box here as well.

I believe we should keep the "Export Zephyr CMake package" section as well, in that case.

docs/docs/development/setup.mdx
Nick-Munnich: Resolved in reshuffling.

docs/docs/development/setup.mdx
Nick-Munnich: This and the below next comment were resolved in reshuffling.

Nick-Munnich:
Performed a larger redo of this by combining its content with that found in #437 

This updates the approach used there to the modern docs, and continues to leverage the Zephyr docs. The previous comments were both resolved in the process, but another careful look is likely due.

Note that there are some blocks written in #437 which I was uncertain about including, so I'm listing them here:

:::note Windows Users
Please note the ZMK builds run slower (up to 3-5 minutes on slower hardware) with Docker on Windows if you don't use the WSL2 filesystem to store files. If you run into performance problems you can checkout the ZMK sources inside a WSL2 environment and run `code .` to open the sources. This can make builds run at near-native speed.

This approach will also need the [Remote - WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) extension installed in VS Code.

Files stored within WSL2 can be accessed via Windows Explorer by navigating to `\\wsl$`.
:::

I don't know how accurate this is, and [Zephyr does not recommend WSL](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html), so I'm not sure this should be included.

:::danger The Docker environment will NOT run on arm CPUs like the Raspberry Pi or Apple Silicon. You must use the native environment if using an arm CPU. :::

Googling this seems to imply that there are workarounds, but I do not have access to a device to test it out. It is not included under the current docs, so I don't see a reason to include it until it is brought up by someone.

:::caution Windows Users If you're using the Docker environment on Windows, you must checkout the sources to a folder within C:\Users\[your_user_here] to avoid a potential permissions issue.

If you're using the WSL2 filesystem the sources should go under ~/ to avoid potential permissions issues. :::

I think this is true, but I have not tested. I've included it anyway, as it certainly does not hurt.

Thoughts on these would be appreciated. I am aware that there is duplicated content in the Native section - this is because I thought it looked nicer that way, but would be willing to change it upon request.

infused-kim:
I can confirm that it’s working fine on Apple Silicon. 

I don’t know if the zmk images have ARM versions. So it might be running using emulation, but it works just fine on Apple silicon and doesn’t need any special configuration.

caksoylar:
FYI I'll try to review this weekend.

docs/docs/development/new-shield.mdx
```diff
@@ -554,7 +554,7 @@ Add additional bindings as necessary to match the default number of encoders on
 
 ### GitHub Actions
 
-Using GitHub Actions to build your new firmware can save you from doing any local [development setup](./setup.mdx),
+Using GitHub Actions to build your new firmware can save you from doing any local [development setup](./setup/getting-started),
```
caksoylar, line 557:
```suggestion
Using GitHub Actions to build your new firmware can save you from doing any local [development setup](./setup/getting-started.md),
```

docs/docs/development/new-shield.mdx
```diff
@@ -566,7 +566,7 @@ Once pushed, click on the "Actions" tab of the repo you created in the first ste
 ### Local Build
 
 :::note
-To build locally, be sure you've followed the [development setup](./setup.mdx) guide first.
+To build locally, be sure you've followed the [development setup](./setup/getting-started) guide first.
```
caksoylar, line 569:
```suggestion
To build locally, be sure you've followed the [development setup](./setup/getting-started.md) guide first.
```

docs/docs/development/setup/docker.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Docker
+sidebar_label: Docker
+---
+
+:::note
+Currently the Docker approach is only documented for VSCode, Microsoft's proprietary version of [Code OSS](https://github.com/microsoft/vscode) - See [here](https://github.com/microsoft/vscode/wiki/Differences-between-the-repository-and-Visual-Studio-Code) for more info on different versions. There is an alternative approach using [devcontainers](https://containers.dev/), but this is not documented yet - contributions are welcome!
```
caksoylar, line 7:
```suggestion
Currently the Docker approach is only documented for [VS Code](https://github.com/microsoft/vscode). While it can be replicated using [devcontainers](https://containers.dev/) this is not documented yet - contributions are welcome!
```
I don't think we need this much detail here.

docs/docs/development/setup/docker.md
```diff
@@ -0,0 +1,77 @@
...
+
+![VS Code Dev Container Configuration Alert](../../assets/dev-setup/vscode_devcontainer.png)
+
+Click `Reopen in Container` in order to reopen the VS Code with the running container.
+
+The first time you do this on your machine, it will pull the docker image down from the registry and build the container. Subsequent launches are much faster!
+
+:::note
+If the alert fails to pop up or you accidentally close it, you can perform the same action from the remote menu: `ctrl+shift+p` and invoke `Remote: Show Remote Menu`.
+:::
```
caksoylar, line 51:
```suggestion
Click `Reopen in Container` in order to reopen the VS Code with the running container. If the alert fails to pop up or you accidentally close it, you can perform the same action by pressing `ctrl+shift+p` and selecting `Remote: Show Remote Menu`.

The first time you do this on your machine, it will pull the docker image down from the registry and build the container. Subsequent launches are much faster!
```

Trying to reduce the amount of special boxes. If we have too many, they lose their usefulness.

docs/docs/development/setup/docker.md
```diff
@@ -0,0 +1,77 @@
...
+
+:::tip
+This step pulls down quite a bit of tooling. Go grab a cup of coffee, it can take 10-15 minutes even on a good internet connection!
+:::
+
+:::info
+You must restart the container at this point. The easiest way to do so is to close the VS Code window, verify that the container has stopped in Docker Dashboard, and reopen the container with VS Code.
+
+Your setup is complete once your container has restarted.
+:::
```
caksoylar, line 77: I suppose `west zephyr-export` isn't necessary for Docker, right?

docs/docs/development/setup/getting-started.md
```diff
@@ -0,0 +1,31 @@
...
+
+:::info
+Throughout this documentation you will see commands like
+
+```sh
+sudo apt update
+```
+
+These commands should be run in a terminal such as Bash (Linux/MacOS/Docker), PowerShell, or Command Prompt (Windows).
+:::
```
caksoylar, line 14: Do you think this is necessary? I feel like this level of detail doesn't belong in dev documentation. I'd probably not recommend the user to use the local build option if this is a point of confusion.

docs/docs/development/setup/getting-started.md
```diff
@@ -0,0 +1,31 @@
...
+:::tip
+We recommend reading through the setup process before following it step by step, to ensure that you are happy with installing the required dependencies.
+:::
+
+## Environment Setup
+
+There are two ways to setup the ZMK development environment:
+
+- [Docker](/docs/development/setup/docker) ;
+- [Native](/docs/development/setup/native), i.e. using the operating system directly.
```
caksoylar, line 25:
```suggestion
There are two ways to setup the ZMK development environment: using [Docker](docker.md) or using a [native setup](native.mdx).
```
I'd say we can keep this small since the details are already elaborated on the following paragraphs.

(We also prefer file path links in the rest of the docs)

docs/docs/development/setup/getting-started.md
```diff
@@ -0,0 +1,31 @@
...
+
+The Docker approach is a self-contained development environment while the native approach will setup your local operating system for development. It uses the same [Docker image which is used by the GitHub action](https://github.com/zmkfirmware/zmk-docker) for local development. Beyond the benefits of [dev/prod parity](https://12factor.net/dev-prod-parity), this approach may be easier to set up for some operating systems. No toolchain or dependencies are necessary when using Docker; the container image has the toolchain installed and set up to use.
+
+The native approach installs the toolchain and dependencies on your system directly. This typically runs slightly faster than the Docker approach, and can be preferable for users who already have the dependencies on their system.
+
+Please see the [Docker](/docs/development/setup/docker) instructions or [native](/docs/development/setup/native) instructions to continue setup.
```
caksoylar, line 31:
```suggestion
Please see the [Docker](docker.md) instructions or [native](native.mdx) instructions to continue setup.
```

docs/static/_redirects
```diff
@@ -0,0 +1 @@
+/docs/development/setup /docs/development/setup/getting-started
```
caksoylar, line 1: I think we wouldn't need this if we made the getting-started page `index.md`. This is how it's handled for Behaviors and Configuration Overview pages.

docs/docs/development/setup/native.mdx
```diff
@@ -0,0 +1,509 @@
...
+Then step into the repository.
+
+```sh
+cd zmk
+```
+
+## 3. Get Zephyr and install Python dependencies
+
+:::note
+These steps are very similar to Zephyr's [Get Zephyr and install Python dependencies](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies) instructions, but specialised for ZMK.
```
caksoylar, line 84:
```suggestion
These steps are very similar to Zephyr's [Get Zephyr and install Python dependencies](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies) instructions, but specialized for ZMK.
```

docs/docs/development/setup/native.mdx
```diff
@@ -0,0 +1,509 @@
...
+Once activated your shell will be prefixed with `(.venv)`. The virtual environment can be deactivated at any time by running `deactivate`.
+
+:::note
+Remember to activate the virtual environment every time you start working.
+:::
+
+4. Install west:
+
+```sh
+pip3 install west
```
caksoylar, line 120:
```suggestion
pip install west
```
`pip3` isn't necessary since the virtualenv has Python 3 already.

docs/docs/development/setup/native.mdx
```diff
@@ -0,0 +1,509 @@
...
+```
+
+5. Initialise the application (This provides ZMK with its own copy of Zephyr):
```
caksoylar, line 123:
```suggestion
5. Initialize the application and update to fetch modules, including Zephyr:
```

caksoylar (commented):
While it seems like a decent improvement, I have to say incorporating #437 into this PR has made it much harder to review. It also is too verbose in some places IMO, which I tried to address in comments.

I also gave up on adding comments to the native page after a while because of the content duplication, I think it will make maintenance harder.

Thanks for making the requested changes from earlier.

caksoylar:
To address the points in your comment:

> :::note Windows Users
> [...]
> I don't know how accurate this is, and [Zephyr does not recommend WSL](https://docs.zephyrproject.org/3.5.0/develop/getting_started/index.html), so I'm not sure this should be included.

While the WSL un-recommendation doesn't apply to most ZMK users (because of uf2 flashing) I wouldn't put any WSL-specific instructions in our docs anyway. Users can simply follow the corresponding Linux distro instructions.

> :::danger The Docker environment will NOT run on arm CPUs like the Raspberry Pi or Apple Silicon. You must use the native environment if using an arm CPU. :::
> Googling this seems to imply that there are workarounds, but I do not have access to a device to test it out. It is not included under the current docs, so I don't see a reason to include it until it is brought up by someone.

I agree, it shouldn't be our responsibility to teach this to users.

> :::caution Windows Users If you're using the Docker environment on Windows, you must checkout the sources to a folder within C:\Users[your_user_here] to avoid a potential permissions issue.
> [...]
> I think this is true, but I have not tested. I've included it anyway, as it certainly does not hurt.

I don't think this is very necessary either, but we can keep it I guess.

docs/docs/development/setup/docker.md
Nick-Munnich: Not according to current documentation, at least.

docs/docs/development/setup/docker.md
Nick-Munnich: I'd prefer to reduce the number of special boxes by getting rid of the windows permissions note and the tooling tip. I think that this note ought to stay.

docs/docs/development/setup/getting-started.md
Nick-Munnich: It depends on the target audience. Anyone with dev experience shouldn't need this note, but I think it may be helpful for some less experienced windows users. It does contribute to making it too verbose though, so I'll get rid of it.

docs/docs/development/setup/getting-started.md
Nick-Munnich: Will change to file path links, but I think the bullet points break up the text and make it easier to read. Otherwise it's just a block of text in this section. Really, I'd prefer to break up the following paragraphs a bit to be more visually appealing, but can't think of a good way to do so.

docs/docs/development/setup/native.mdx
```diff
@@ -0,0 +1,509 @@
...
+
+```sh
+west init -l app/
+west update
+```
+
+:::tip
+This step pulls down quite a bit of tooling. Go grab a cup of coffee, it can take 10-15 minutes even on a good internet connection!
```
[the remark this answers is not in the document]
Nick-Munnich: I agree, it didn't take nearly as long for me -- Maybe they've managed to optimise the setup? I rephrased it to be a bit shorter and not be so exaggerated.

Nick-Munnich:
Did the necessary refactoring to make maintenance easier at the cost of appearance.

Two points from the previous review are not marked as resolved, the rest should be fine now I think,

docs/docs/development/setup/docker.md
Nick-Munnich: Completely missed that, sorry!

docs/docs/customization.md
```diff
@@ -40,7 +40,7 @@ If you need to, a review of [Learn The Basics Of Git In Under 10 Minutes](https:
 :::
 
 :::note
-It is also possible to build firmware locally on your computer by following the [toolchain setup](development/setup.mdx) and
+It is also possible to build firmware locally on your computer by following the [toolchain setup](development/setup) and
```
caksoylar, line 43:
```suggestion
It is also possible to build firmware locally on your computer by following the [toolchain setup](development/setup/index.md) and
```
I think this should work with the file path version.

docs/docs/development/new-shield.mdx
```diff
@@ -554,7 +554,7 @@ Add additional bindings as necessary to match the default number of encoders on
 
 ### GitHub Actions
 
-Using GitHub Actions to build your new firmware can save you from doing any local [development setup](./setup.mdx),
+Using GitHub Actions to build your new firmware can save you from doing any local [development setup](./setup),
```
caksoylar, line 557:
```suggestion
Using GitHub Actions to build your new firmware can save you from doing any local [development setup](./setup/index.md),
```

docs/docs/development/new-shield.mdx
```diff
@@ -566,7 +566,7 @@ Once pushed, click on the "Actions" tab of the repo you created in the first ste
 ### Local Build
 
 :::note
-To build locally, be sure you've followed the [development setup](./setup.mdx) guide first.
+To build locally, be sure you've followed the [development setup](./setup) guide first.
```
caksoylar, line 569:
```suggestion
To build locally, be sure you've followed the [development setup](./setup/index.md) guide first.
```

docs/docs/development/setup/index.md
```diff
@@ -0,0 +1,21 @@
...
+sidebar_label: Getting Started
+---
+
+:::tip
+We recommend reading through the setup process before following it step by step, to ensure that you are happy with installing the required dependencies.
+:::
+
+## Environment Setup
+
+There are two ways to setup the ZMK development environment:
```
caksoylar, line 12:
```suggestion
There are two ways to set up the ZMK development environment:
```

docs/docs/development/setup/native.mdx
```diff
@@ -0,0 +1,338 @@
...
+3. Export a [Zephyr CMake package](https://docs.zephyrproject.org/3.5.0/build/zephyr_cmake_package.html#cmake-pkg). This allows CMake to automatically load boilerplate code required for building Zephyr applications.
+
+```sh
+west zephyr-export
+```
+
+4. Install the additional dependencies found in Zephyr's `requirements-base.txt`.
+
+```sh
+pip3 install -r ~/zephyrproject/zephyr/scripts/requirements-base.txt
```
caksoylar, line 293: Is this the right path for macOS? I thought we'd want the one in `zephyr/scripts/requirements-base.txt` that comes with `west update`.

caksoylar (commented):
Thanks again, the deduplication helped a lot. Can we do the same for the `glob` tab? Steps 2 and 3 seem to be identical.

Regarding the bullet points in the getting started page: Maybe we can integrate the following paragraphs into the bullet points? This way there is less repetition, but still visual separation.

docs/docs/customization.md
Nick-Munnich: Tested and it does.

docs/docs/development/setup/native.mdx
Nick-Munnich: Woops, nice catch.

Nick-Munnich:
I deduplicated the glob tab, but in a slightly awkward way by getting rid of the function exporting the OsTabs environment. glob needing two tabs on 1 and 4 looks horrible, so I introduced a css class to just hide the second tab menu while keeping the switching functionality. There out to be be a better way to do this, but I don't know if or how.

Tried integrating the bullet points, I quite like it.

docs/docs/development/setup/native.mdx
```diff
@@ -0,0 +1,318 @@
...
+<TabItem value="cmd">
+
+```sh
+.venv\Scripts\activate.bat
+```
+
+</TabItem>
+<TabItem value="ps">
+
+```sh
```
caksoylar, line 139:
````suggestion
```powershell
````

File: docs/docs/development/setup/native.mdx [hunk 23]
```diff
@@ -0,0 +1,318 @@
...
+```
+
+
+</TabItem>
+  <TabItem value="win" label="Windows">
+1. Install `west`, and make sure `~/.local/bin` is on your `PATH` environment variable:
+
+```sh
+pip install -U west
+```
```
caksoylar, line 217:
````suggestion
1. Install `west`:

```sh
pip install -U west
```

Make sure the Python script directory is on your `PATH` environment variable by opening a PowerShell window and running the following commands:

```powershell
$Scripts = python -c "import sysconfig; print(sysconfig.get_path('scripts'))"
$Path = [Environment]::GetEnvironmentVariable('PATH', 'User')
[Environment]::SetEnvironmentVariable('PATH', "$Path;$Scripts", 'User')
$env:PATH += ";$Scripts"
```

````

I realized we lost the info in https://zmk.dev/docs/development/setup#install-west Windows tab on the PATH stuff here, even though it is mentioned in the bullet point.

(This should be under bullet 1., I am not sure the way I wrote it above works)

caksoylar (commented):
Thanks, I personally think the hack is fine but we can see if anyone disagrees.

I am generally happy with this, but given the scope of changes we should see if anyone else from @zmkfirmware/docs has comments?

Nick-Munnich:
I re-added the missing information and made the information in the ubuntu section a bit nicer too -- this is exactly what "note" blocks are for, imo.

caksoylar (approved):

caksoylar:
I'll do a final pass and if everything's good I'll merge tomorrow, unless anyone has objections.

PR #2265, opened 2024-04-09
nice!view: individual profile status

snoyer:
Allow the nice!view widget to display the status of each profile by:
- not adding a circle around the number if the profile is not bound
- drawing a dashed circle around the number if a profile is bound but not connected (device not live, explicitly disconnected with `BT_DISC`, ...)

Example:
niceview-profiles

- solid circle: connected
- dashed circle: not connected
- no circle: not bound
- filled circle: selected

---

This requires minor refactoring in BLE code to expose state methods by profile address rather than only by the current profile.

One issue I can see so far is that event's are indeed raised when the current profile changes but not when other profiles change from what I can tell (eg. a device other that the selected one disconnects).

I have tried addressing the `BT_DISC` case by adding a `raise_profile_changed_event();` in `zmk_ble_prof_disconnect()` and that seems to work but I'm pretty sure that's not the proper way to do it (if there is a proper way to do it currently, see comment). Maybe that needs to be addressed separately from this PR?

PR #2275, opened 2024-04-17
feat(ble): Add information to profile changed event

ReFil:
Currently the `ble_active_profile_changed` event will trigger if a profile is cleared or disconnected but it's up on the individual event handlers to call `zmk_ble_active_profile_is_open()` and `zmk_ble_active_profile_is_connected()` to actually retrieve that information, integrating all this inside the event streamlines downstream implementations as all they need to do is retrieve this information from the event rather than each event handler call the getter functions

### DIFF
--- a/app/include/zmk/events/ble_active_profile_changed.h
+++ b/app/include/zmk/events/ble_active_profile_changed.h
@@ -15,6 +15,8 @@
 struct zmk_ble_active_profile_changed {
     uint8_t index;
     struct zmk_ble_profile *profile;
+    bool open;
+    bool connected;
 };
 
 ZMK_EVENT_DECLARE(zmk_ble_active_profile_changed);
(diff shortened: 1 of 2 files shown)


snoyer:
Another consideration would be to generalize `ble_active_profile_changed` to `ble_profile_changed` as we may want to know about any profile changing instead of only the active one. Example use-case would be to display an up-to-date status for every available profile even when the change is initiated by the connected device and not ZMK (eg. a device bound to a non-selected profile disconnects), see #2265.

The current `raise_profile_changed_event(void)` method could be renamed `raise_current_profile_changed_event(void)` and a new `raise_profile_changed_event(...)` would take a profile address or index parameter.

caksoylar:
I also thought it might be nice to consolidate the profile status to an enum with three values: open, disconnected, connected. Three states are exclusive to each other, so two APIs to query it seems inelegant.

ReFil:
This PR is ready for review/merge, no additional reading etc is required

ReFil:

PR #2277, opened 2024-04-18
fix(behaviors): Add multiple soft-off instances properly.

petejohanson:
* Properly pass the node id for the unique
  soft-off behavior instance when defining it.

A user hit this bug when compiling a board with a hardware soft-off *and* one added in their keymap.


petejohanson (commented):
Simple change, confirmed working by the affected user.

Issue #767, opened 2021-04-21
Feature request: Select previously active Bluetooth profile


dxmh:
It would be great to be able switch to the previously active Bluetooth profile. 

For example — I’m typing away on profile 1, then I manually switch to profile 2 for a small task. I’d then like to get back to profile 1 using something like this:

```
&bt BT_SEL {{ previously active profile }}
``` 

This functionality would make the experience of switching back and forth between profiles similar to `ctrl`+`tab` window switching. 

This could be useful also in macros, for example to pause the media on another host and return back to the current host:

```
macro_toggle_media: macro_toggle_media {
  compatible = "zmk,behavior-macro";
  label = "macro_toggle_media";
  #binding-cells = <0>;
  bindings = <&bt BT_SEL 1>, <&kp C_PLAY_PAUSE>, <&bt BT_SEL {{ previously active profile }}>;
};

```

auipga:
I like the idea. **Prev** and also **next** (some sort of history) would be awesome.

Dimentium:
I've added PR #2278
It works as expected. You can use index 255 to switch back to the previously used profile.

NewbiZ:
What is the status of this issue? I can see at least 3 PRs (#2497, #2278 and #1413) implementing the feature, though none was merged.
Is there still interest for the feature? If so, I am willing to finish the work on on either of the existing PRs, it seems most of them were functional but just lacked the last extra mile from their initial contributor.

ZephWyrm:
This would be nice to be implemented. What's funny is that AIs suggest this as existing functionality (ask me how I got here 🥲 )

PR #2278, opened 2024-04-18
feat(ble): Fast switching between two recent profiles

Dimentium:
related to #767 

I'm not a developer, so this implementation is quite basic, but it works perfectly for me.

This update introduces the ability to switch back to the previously active bluetooth profile by passing index 255 to the profile selection function.  This feature facilitates quick toggling between two frequently used Bluetooth devices.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -3,6 +3,7 @@ name: pre-commit
 on:
   pull_request:
   push:
+  workflow_dispatch:
 
 jobs:
   pre-commit:
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -3,7 +3,6 @@ name: pre-commit
 on:
   pull_request:
   push:
-  workflow_dispatch:
 
 jobs:
   pre-commit:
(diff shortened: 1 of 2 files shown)


huber-th:
This is a great idea. I personally switch between two profiles a lot and can see the benefit of being able to only use a single key binding on the base layer rather than having to have two key bindings to swap between profiles.

I would recommend to add a new command, e.g. `&bt BT_SWAP`, instead of using `&bt BT_SEL 255` for this.

You could also extend the code to ensure the previous profile is persisted along side the active profile, so that swapping also works directly after startup.

Dimentium:
Hi @caksoylar,
Is there anything I need to do with this MR?

caksoylar:
I am not making the merge decision for this, but I don't see this being useful without the alias `BT_ALT` like @huber-th suggested. Then it also needs to be documented.

huber-th:
@Dimentium I really love the idea of this feature, especially since I switch between two profiles frequently. 

Since you mentioned that you're not a developer, I'd be happy to implement the improvements I suggested above as well as document the feature. We can then review the changes together.

What do you think?

Nick-Munnich:
#1413 is meant to implement the same thing, but a more recent implementation would be good. I like the naming used there though, `BT_LAST_DEV`.

PR #2280, opened 2024-04-21
fix(boards): nrf boards missing SPI definitions for pro_micro pins at the board level

sadekbaroudi:
nice nano didn't have definitions for SPI in pinctrl and dtsi, requiring users to manually define in their shield definitions if they wanted to use SPI. This should be defined at the board level.

### DIFF
--- a/app/boards/arm/nice_nano/nice_nano.dtsi
+++ b/app/boards/arm/nice_nano/nice_nano.dtsi
@@ -50,6 +50,13 @@
     pinctrl-names = "default", "sleep";
 };
 
+&spi1 {
+    compatible = "nordic,nrf-spim";
+    pinctrl-0 = <&spi1_default>;
+    pinctrl-1 = <&spi1_sleep>;
+    pinctrl-names = "default", "sleep";
+};
+
 &uart0 {
     compatible = "nordic,nrf-uarte";
     current-speed = <115200>;
--- a/app/boards/arm/bluemicro840/arduino_pro_micro_pins.dtsi
+++ b/app/boards/arm/bluemicro840/arduino_pro_micro_pins.dtsi
@@ -53,5 +53,5 @@
 
 pro_micro_d: &pro_micro {};
 pro_micro_i2c: &i2c0 {};
-pro_micro_spi: &spi0 {};
+pro_micro_spi: &spi1 {};
 pro_micro_serial: &uart0 {};
--- a/app/boards/arm/bluemicro840/bluemicro840_v1-pinctrl.dtsi
+++ b/app/boards/arm/bluemicro840/bluemicro840_v1-pinctrl.dtsi
@@ -36,4 +36,21 @@
             low-power-enable;
         };
     };
+
+    spi1_default: spi1_default {
+        group1 {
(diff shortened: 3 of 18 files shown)


app/boards/arm/nice_nano/nice_nano.dtsi
```diff
@@ -50,6 +50,13 @@
     pinctrl-names = "default", "sleep";
 };
 
+&spi0 {
```
petejohanson, line 53: Oh, can we please use `spi1`?

nRF52 doesn't allow using the same numbered SPI & i2c peripheral at the same time. So someone trying to use both would have problems if we stuck with this.

petejohanson (changes requested):
One comment on the code, and then if you're up for it, can you modify your commit msg to match our conventional commit style? If not I'll clean it up with the a squash when I merge.

Thanks for the PR!

app/boards/arm/nice_nano/nice_nano.dtsi
caksoylar: Looks like `nice_nano`, `nrfmicro`, `mikoto`, `bluemicro840` all use `pro_micro_spi: spi0`. I think it makes sense to use `pro_micro_spi` for these since they are standard Pro Micro pin locations so it might be nice to rename them to `spi1`.

I suppose (maybe separately) it would be nice to define pinctrl for others as well.

sadekbaroudi: @caksoylar - I can make these changes. I didn't realize that the other boards didn't have SPI defined as well.

So, I'm reading that I should:
* update the commit message to convention
* change pro_micro_spi for all nrf boards to be spi1
* change the definitions I made in this commit to be spi1 instead of spi0

sadekbaroudi:
@petejohanson @caksoylar 

Does this look good?

petejohanson:
> @petejohanson @caksoylar
> 
> Does this look good?

Seems reasonable to me! Have you had a chance to test at least the nano changes with an SPI peripheral?

sadekbaroudi:
@petejohanson @caksoylar 
> Seems reasonable to me! Have you had a chance to test at least the nano changes with an SPI peripheral?

I just confirmed it be working (with the help of someone in my discord server) with a nice!nano and cirque trackpad, hand wired to these pins on a VIK connector, which was defined to use the pro micro pins.

sadekbaroudi:
On a related note, I looked at the github action for build testing. Should it also have a test to verify that various peripherals are compiled in as intended? (SPI, I2C, etc)

petejohanson (approved):
LGTM. Thanks for sorting out the rest while you were at it.

petejohanson:
> On a related note, I looked at the github action for build testing. Should it also have a test to verify that various peripherals are compiled in as intended? (SPI, I2C, etc)

Yeah, would be nice. We'd probably need an "interconnect specific shield" that would do `&pro_micro_spi { status = "okay"; };` or `&xiao_spi { status = "okay"; };` and then build, to test that in a way that worked across our supported hardware. Bit tricky to get right generically, but doable. Merging this for now, lacking that kind of automated testing.

PR #2283, opened 2024-04-25
Feature: Add physical layout support

petejohanson:
This feature enhancement adds support for "physical layouts" that describe a given keyboard layout. Example uses cases for multiple physical layouts include:

* Devices with removable columns, e.g. Corne V2,
* Devices that can be built with multiple configurations, e.g. a 60% PCB that supports both ISO and ANSI, or a Planck-like with multiple bottom rows.

To make avoid code duplication, under the hood we are still using a "synthetic physical layout" when none are explicitly specified, using the `chosen` matrix transform and kscan nodes to set up a single physical layout that is then used.

A Physical Layout includes:
* The list of keys' physical attributes, including width, height, x, y, r, rx, and ry values.
* The matrix transform to use when the layout is active.
* The kscan device to use with this layout (this allows for freeing up kscan hardware (e.g. extra column pins) not eased for every layout for *other* purposes.

Selecting a physical layout will do the necessary setup for the layout's kscan instance, and when processing kscan events, use the associated matrix transform to map that kscan event to a key position.

The code here *also* has some tweaks to calculate the `ZMK_KEYMAP_LEN`, aka the maximum key position index, based on the *longest* transform available from all the physical layouts defined, which we will use down the road in the ZMK Studio work to ensure we have enough pre-allocated locations for keymaps the user sets up after picking a layout.

Not yet implemented, to come as follow up work:
* Selection of physical layout being propagated to split peripherals.
* Runtime selected layout persistence to settings subsystem.


app/src/matrix_transform.c
```diff
@@ -4,39 +4,74 @@
...
+    }
+
+    return 0;
+}
+
+SYS_INIT(init_synth_matrix_transform, POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT);
+
+#else
+
+#error "Need a matrix tranform or compatible kscan selected to determine keymap size!"
```
caksoylar, line 69:
```suggestion
#error "Need a matrix transform or compatible kscan selected to determine keymap size!"
```

app/src/matrix_transform.c
```diff
@@ -45,21 +80,21 @@ int32_t zmk_matrix_transform_row_column_to_position(uint32_t row, uint32_t colum
...
 
-    const uint32_t value = transform[matrix_index];
+    for (int i = 0; i < mt->entries_len; i++) {
+        const struct zmk_matrix_transform_entry *entry = &mt->entries[i];
 
-    if (!value) {
-        return -EINVAL;
+        if (entry->row == row && entry->column == column) {
+            return i;
+        }
```
caksoylar, line 90: Would it be worth constructing a reverse lookup table for each transform (or just for the dynamically chosen one at the time of choosing?) and using that for constant time lookups?

I am not sure how often this function is called, but if it is called on every key position change then it would add up.

app/src/physical_layouts.c
```diff
@@ -0,0 +1,201 @@
...
+                                                       n)[DT_INST_PROP_LEN_OR(n, keys, 0)] = {     \
+                                                                                                   \
+        LISTIFY(DT_INST_PROP_LEN_OR(n, keys, 0), ZKPA_INIT, (, ), n)};                             \
+    ZMK_MATRIX_TRANSFORM_EXTERN(DT_INST_PHANDLE(n, transform));                                    \
+    static const struct zmk_physical_layout _CONCAT(_zmk_physical_layout_, DT_DRV_INST(n)) = {     \
+        .display_name = DT_INST_PROP_OR(n, display_name, "Layout #" #n),                           \
+        .matrix_transform = ZMK_MATRIX_TRANSFORM_T_FOR_NODE(DT_INST_PHANDLE(n, transform)),        \
+        COND_CODE_1(DT_INST_PROP_LEN(n, kscan),                                                    \
+                    (.kscan = DEVICE_DT_GET(DT_INST_PHANDLE(n, kscan)), ), ())                     \
+            .keys = _CONCAT(_zmk_physical_layout_keys_, n),                                        \
```
caksoylar, line 46: Some indentation funkiness here?

app/src/matrix_transform.c
caksoylar: Oh, it looks like the `transform` in the old code was doing that? It does require knowing the max row/col values which isn't ideal.

petejohanson: Yeah. As I've slept on this more, maybe we:

* Use the max x/y values set on the *transform*
* Use that info to easily create that constant lookup table for a given transform.
* Stop inferring those max x/y from properties on the kscan node, which is a major headache and forces a dependency on kscan driver internals. Painful to use third party kscan like my EC analog driver.

Thoughts?

caksoylar: Yeah, sounds very reasonable to me. Setting `rows`/`columns` properties on matrix transforms + kscans (including composite ones) were always a bit confusing IMO, even for me. Do I understand correctly that none of those will be necessary in this case?

petejohanson: With this change, it would be required consistently *only* on the matrix transform.

caksoylar: Can that not be inferred from the item values in the `map`?

petejohanson: We could at *runtime* if we want to, but not in our preprocessor to create it statically.

petejohanson: Not necessarily, but it *does* force it to be non-const, so it'd be in RAM not ROM.

app/src/matrix_transform.c
petejohanson: Good catch. Fixed.

app/src/matrix_transform.c
petejohanson: Ok, figured out how to restore the constant time lookup table for all scenarios, and ditched the "entries" approach I'd used before.

Also took the time to fix up the col/row offsets so this should still work fine for splits.

app/src/physical_layouts.c
petejohanson: `clang-format` butchered this. Let me see if there's a nicer way that it will also be happy with.

app/src/matrix_transform.c
caksoylar: Great, FWIW it looks good to me!

petejohanson:
Ok, I've pushed some minor cleanups and fixes, as well as an implementation of this for ZMK Uno to show how this works. Tested working for the ZMK Uno split as well.

app/boards/shields/zmk_uno/zmk_uno_split.dtsi
```diff
@@ -31,18 +33,57 @@
...
+        keys
+            = <&key_physical_attrs 100 100   0   0 0 0 0>
+            , <&key_physical_attrs 100 100 100   0 0 0 0>
+            , <&key_physical_attrs 100 100   0 100 0 0 0>
+            , <&key_physical_attrs 100 100 100 100 0 0 0>
+            , <&key_physical_attrs 100 100 200   0 0 0 0>
+            , <&key_physical_attrs 100 100 300   0 0 0 0>
+            , <&key_physical_attrs 100 100 200 100 0 0 0>
+            , <&key_physical_attrs 100 100 300 100 0 0 0>
+            ;
```
caksoylar, line 64: I don't know how much you care about these at this stage, but this physical layout spec doesn't match the transform/keymap shape:

image

To fix that we'd either have to move the right 4 keys below the left in `keys`, or change matrix-transform+keymap order and reorder items in `keys`.

caksoylar:
In case anyone wants to test the physical layout, here is a helper script to convert from QMK-style info.json files or ortho-like specs to a devicetree snippet for `keys` property, based on https://github.com/caksoylar/keymap-drawer: https://gist.github.com/caksoylar/1f6809446ab2415d4116882ed1c60db2#file-physical_layout_to_dt-py

```sh
❯ python physical_layout_to_dt.py -k corne_rotated -l LAYOUT_split_3x5_3

keys  //                     w   h    x    y     rot   rx   ry
    = <&key_physical_attrs 100 100    0   37       0    0    0>
    , <&key_physical_attrs 100 100  100   12       0    0    0>
    , <&key_physical_attrs 100 100  200    0       0    0    0>
    , <&key_physical_attrs 100 100  300   12       0    0    0>
    , <&key_physical_attrs 100 100  400   25       0    0    0>
    , <&key_physical_attrs 100 100  800   25       0    0    0>
    , <&key_physical_attrs 100 100  900   12       0    0    0>
    , <&key_physical_attrs 100 100 1000    0       0    0    0>
    , <&key_physical_attrs 100 100 1100   12       0    0    0>
    , <&key_physical_attrs 100 100 1200   37       0    0    0>
    , <&key_physical_attrs 100 100    0  137       0    0    0>
    , <&key_physical_attrs 100 100  100  112       0    0    0>
    , <&key_physical_attrs 100 100  200  100       0    0    0>
    , <&key_physical_attrs 100 100  300  112       0    0    0>
    , <&key_physical_attrs 100 100  400  125       0    0    0>
    , <&key_physical_attrs 100 100  800  125       0    0    0>
    , <&key_physical_attrs 100 100  900  112       0    0    0>
    , <&key_physical_attrs 100 100 1000  100       0    0    0>
    , <&key_physical_attrs 100 100 1100  112       0    0    0>
    , <&key_physical_attrs 100 100 1200  137       0    0    0>
    , <&key_physical_attrs 100 100    0  237       0    0    0>
    , <&key_physical_attrs 100 100  100  212       0    0    0>
    , <&key_physical_attrs 100 100  200  200       0    0    0>
    , <&key_physical_attrs 100 100  300  212       0    0    0>
    , <&key_physical_attrs 100 100  400  225       0    0    0>
    , <&key_physical_attrs 100 100  800  225       0    0    0>
    , <&key_physical_attrs 100 100  900  212       0    0    0>
    , <&key_physical_attrs 100 100 1000  200       0    0    0>
    , <&key_physical_attrs 100 100 1100  212       0    0    0>
    , <&key_physical_attrs 100 100 1200  237       0    0    0>
    , <&key_physical_attrs 100 100  250  315       0    0    0>
    , <&key_physical_attrs 100 100  360  330    1500  409  380>
    , <&key_physical_attrs 100 150  476  325    3000  527  400>
    , <&key_physical_attrs 100 150  723  325 (-3000)  773  400>
    , <&key_physical_attrs 100 100  840  330 (-1500)  890  380>
    , <&key_physical_attrs 100 100  950  315       0    0    0>
    ;
```

app/boards/shields/zmk_uno/zmk_uno_split.dtsi
petejohanson: Good catch. Fixed the physical layout to match the keymap/transform.

app/boards/shields/zmk_uno/zmk_uno_split_right.overlay
```diff
@@ -11,7 +11,7 @@
 };
 
 &split_direct_matrix_transform {
-    row-offset = <2>;
+    col-offset = <4>;
```
caksoylar, line 14:
```suggestion
    row-offset = <1>;
```
If I am not mistaken, this seems to be the right value

app/boards/shields/zmk_uno/zmk_uno_split.dtsi
caksoylar: The direct wire one looks good now, but the matrix layout one is still side by side: 
image
I think it could be the same as the direct wire one, right?

petejohanson: Oh indeed. Fixed. Thanks.

app/boards/shields/zmk_uno/zmk_uno_split_right.overlay
petejohanson: Fixed. Thanks!

caksoylar: I think there might have been a mistake, you converted `col-offset` value to `row-offset` but the value is still `<4>`, not `<1>`.

petejohanson: Indeed. That's what I get for trying to quickly fix things during a rebase. Fixed?

caksoylar: LGTM now 😄

caksoylar (commented):
I left some comments, but the changes look good to me in general. 

Any plans to add docs for this, or any thoughts on it? (We can discuss that on Discord.)

app/dts/bindings/zmk,physical-layout.yaml
```diff
@@ -0,0 +1,26 @@
...
+properties:
+  display-name:
+    type: string
+    required: true
+    description: The name of this layout to display in the UI
+  transform:
+    type: phandle
+    required: true
+    description: The matrix transform to use along with this layout.
+  kscan:
```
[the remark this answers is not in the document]
petejohanson: That had been my intention, but I'm not sure if I remember to test that edge case yet. Will do so.

app/src/physical_layouts.c
```diff
@@ -0,0 +1,392 @@
...
+    }
+#endif
+
+    memset(map, UINT32_MAX, dest_layout->keys_len);
+
+    for (int b = 0; b < dest_layout->keys_len; b++) {
+        bool found = false;
+
+#if HAVE_POS_MAP
+        if (!found && src_pos_map && dest_pos_map) {
```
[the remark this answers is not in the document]
petejohanson: Indeed. Removed.

app/src/physical_layouts.c
```diff
@@ -0,0 +1,392 @@
...
+    const struct zmk_physical_layout *dest_layout = layouts[dest];
+
+#if HAVE_POS_MAP
+    const struct position_map_entry *src_pos_map = NULL;
+    const struct position_map_entry *dest_pos_map = NULL;
+
+    for (int pm = 0; pm < ARRAY_SIZE(positions_maps); pm++) {
+        if (positions_maps[pm].layout == src_layout) {
+            src_pos_map = &positions_maps[pm];
+        } else if (positions_maps[pm].layout == dest_layout) {
```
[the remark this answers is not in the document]
petejohanson: A funky edge case indeed... but updated to handle it just in case. Good catch!

app/dts/bindings/zmk,physical-layout.yaml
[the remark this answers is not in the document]
petejohanson: Ok, just pushed a fix for this, whereby we set the layout's `kscan` to the `zmk,kscan` chosen as a fallback if the property is not set on the physical layout node. Thanks!

app/include/zmk/matrix.h
```diff
@@ -9,15 +9,21 @@
...
-#if DT_HAS_CHOSEN(zmk_matrix_transform)
+#if DT_HAS_COMPAT_STATUS_OKAY(zmk_physical_layout)
+
+#define ZMK_PHYSICAL_LAYOUT_BYTE_ARRAY(node_id)                                                    \
+    uint8_t _CONCAT(prop_, node_id)[DT_PROP_LEN(DT_PHANDLE(node_id, transform), map)];
+
+#define ZMK_KEYMAP_LEN                                                                             \
+    sizeof(union {DT_FOREACH_STATUS_OKAY(zmk_physical_layout, ZMK_PHYSICAL_LAYOUT_BYTE_ARRAY)})
+
+#elif ZMK_MATRIX_HAS_TRANSFORM
```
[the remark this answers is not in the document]
petejohanson: Added a check. Thanks!

caksoylar (approved):

docs/docs/development/new-shield.mdx
```diff
@@ -377,10 +365,58 @@ Some important things to note:
...
+- If you have a keyboard with options for `2u` keys in certain positions, ANSI vs ISO layouts, or break away portions, the define one matrix transform for each possible arrangement, and they will be used in the physical layouts to let users select a given layout in they keymap file.
 
 See the [matrix transform section](../config/kscan.md#matrix-transform) in the Keyboard Scan configuration documentation for details and more examples of matrix transforms.
 
+## Physical Layout
+
+The physical layout is the top level entity that aggregates all details about a certain possibly layout for a keyboard: the matrix transform that defines the set of key positions, and what row/column they correspond to, what kscan driver is used for that layout, etc.
+For keyboards that support multiple layouts, setting a `chosen` node to a given physical layout in your keymap will allow selecting the given layout you've built.
```
caksoylar, line 375:
```suggestion
For keyboards that support multiple layouts, setting a `chosen` node to a defined physical layout in your keymap will allow selecting the specific layout that you've built.
```
"You've built" isn't necessarily true for most people but I couldn't find a more inclusive alternative that is still clear.

docs/docs/development/new-shield.mdx
```diff
@@ -377,10 +365,58 @@ Some important things to note:
...
+    };
+};
+```
+
+When supporting multiple layouts, define the multiple layout nodes and then set a `chosen` for the default:
+
+```
+/ {
+    chosen {
+        zmk,physical-layout = <&default_layout>;
```
caksoylar, line 394: Is this not `zmk,physical-layout = &default_layout;`? (or are they the same thing?)

docs/docs/development/new-shield.mdx
```diff
@@ -377,10 +365,58 @@ Some important things to note:
 
 - The `#include <dt-bindings/zmk/matrix_transform.h>` is critical. The `RC` macro is used to generate the internal storage in the matrix transform, and is actually replaced by a C preprocessor before the final devicetree is compiled into ZMK.
 - `RC(row, column)` is placed sequentially to define what row and column values that position corresponds to.
-- If you have a keyboard with options for `2u` keys in certain positions, or break away portions, it is a good idea to set the chosen `zmk,matrix-transform` to the default arrangement, and include _other_ possible matrix transform nodes in the devicetree that users can select in their user config by overriding the chosen node.
+- If you have a keyboard with options for `2u` keys in certain positions, ANSI vs ISO layouts, or break away portions, define one matrix transform for each possible arrangement to be used in the physical layouts. This will allow the users to select the right layout in their keymap files.
```
caksoylar, line 368:
```suggestion
- If you have a keyboard with options for `2u` keys in certain positions, ANSI vs. ISO layouts, or break away portions, define one matrix transform for each possible arrangement to be used in the physical layouts. This will allow the users to select the right layout in their keymap files.
```
Missed this earlier, apparently this is necessary in American English.

docs/docs/development/new-shield.mdx
petejohanson: Good catch. Fixed.

app/src/physical_layouts.c
```diff
@@ -0,0 +1,390 @@
...
+        if (pl->kscan) {
+            if (pm_device_wakeup_is_capable(pl->kscan)) {
+                pm_device_wakeup_enable(pl->kscan, true);
+            }
+        }
+    }
+#endif // IS_ENABLED(CONFIG_PM_DEVICE)
+
+#if IS_ENABLED(CONFIG_SETTINGS)
+    settings_subsys_init();
```
caksoylar, line 384: Is this necessary with it being called in `main()` now?

petejohanson: Nope. Nuking.

caksoylar (approved):

PR #2284, opened 2024-04-25
Correctly set underglow state when entering sleep

jradtilbrook:
This fixes a bug introduced in #2244


jradtilbrook:
I believe it now works 🙌 

This was my broken build with current main branch: https://github.com/jradtilbrook/zmk-config/actions/runs/8831214084/job/24245952124#step:11:144
And now, after patching and updating my config to use the patch: https://github.com/jradtilbrook/zmk-config/actions/runs/8831399440

petejohanson (approved):
Thanks!

PR #2285, opened 2024-04-27
shortening keymap_soft_off behavior label

### LINKED ISSUE #2274: Soft off only turn off the left side of my Corne split keyboard
I recently built a wireless Corne keyboard on the original v3 PCB design which doesn't have a power switch.

To turn my keyboard off I setup a keymap to trigger the Soft Off feature and it does work however only on the master side and the right side stays up.

I tried to create additional key map on the right side but that also only turn the left side off.

I am using Corne V3 and Nice!nano controller. My last build was through github action on 15th Apr 2024.

GermanG:
This PR is aiming to fix #2274 
The behavior label gets truncated because the len is 9, defined in app/include/zmk/split/bluetooth/service.h as ZMK_SPLIT_RUN_BEHAVIOR_DEV_LEN
Shortening the keymap_soft_off solves the issue. 
Thanks bravekarma (discord) for the advice.

### DIFF
--- a/app/dts/behaviors/soft_off.dtsi
+++ b/app/dts/behaviors/soft_off.dtsi
@@ -6,7 +6,7 @@
 
 / {
     behaviors {
-        /omit-if-no-ref/ soft_off: soft_off {
+        /omit-if-no-ref/ soft_off: z_so_off {
             compatible = "zmk,behavior-soft-off";
             #binding-cells = <0>;
             split-peripheral-off-on-press;
--- a/app/dts/behaviors/soft_off.dtsi
+++ b/app/dts/behaviors/soft_off.dtsi
@@ -6,7 +6,7 @@
 
 / {
     behaviors {
-        /omit-if-no-ref/ soft_off: keymap_soft_off {
+        /omit-if-no-ref/ soft_off: soft_off {
             compatible = "zmk,behavior-soft-off";
             #binding-cells = <0>;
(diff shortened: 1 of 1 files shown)


app/dts/behaviors/soft_off.dtsi
```diff
@@ -6,7 +6,7 @@
 
 / {
     behaviors {
-        /omit-if-no-ref/ soft_off: keymap_soft_off {
+        /omit-if-no-ref/ soft_off: soft_off {
```
petejohanson, line 9: This had originally been added to reduce the risk of folks adding additional DTS with the same node name accidentally... Perhaps something like `z_so_beh` so it's ever less likely of accidentally getting overriden?

joelspadin: Is that really necessary? We don't do that for any other behaviors. Including "behavior" in the node name isn't necessary either, since behaviors have their own lookup that's separate from other node types.

petejohanson: Probably overkill, but the only reason this one is a bit different than the others is the fact it is the first behavior that is likely to be used *directly* in board dts or shield overlay files, and if created there, accidentally clash with the "keymap version".

GermanG: It won't hurt to rename it, I guess, I updated the PR

PR #2291, opened 2024-05-03
fix(boards): Fix pulls on ZMK uno toggle switch

ReFil:
The devicetree pulls always add on to the extra pulls configured by toggle mode, so these should not have pulls defined in the devicetree. Saved ~200uA avg on another board with a 3t toggle switch.

AFAIK there are no other upstream board using `toggle-mode;` where this needs to be corrected


petejohanson (approved):
Thanks!

PR #2292, opened 2024-05-03
feat(docs): Document example toggle-mode implementation

ReFil:
The documentation for `toggle-mode` is unclear and leaves room for interpretation around the activation of the pull resistors.

This documents an example implementation, as well as writing up some of the quirks and features someone trying to implement toggle-mdoe functionality may not know (e.g. the power on behaviour and gpio requirements)

### DIFF
--- a/docs/docs/config/kscan.md
+++ b/docs/docs/config/kscan.md
@@ -94,7 +94,7 @@ Assuming the switches connect each GPIO pin to the ground, the [GPIO flags](http
     };
 ```
 
-By default, a switch will drain current through the internal pull up/down resistor whenever it is pressed. This is not ideal for a toggle switch, where the switch may be left in the "pressed" state for a long time. Enabling `toggle-mode` will make the driver flip between pull up and down as the switch is toggled to optimize for power. For `toggle-mode` to work correctly each pole of the switch needs a dedicated GPIO pin.
+By default, a switch will drain current through the internal pull up/down resistor whenever it is pressed. This is not ideal for a toggle switch, where the switch may be left in the "pressed" state for a long time. Enabling `toggle-mode` will make the driver enable and disable the internal pull up/down resistor as needed when the switch is toggled to minimise power draw. For `toggle-mode` to work correctly each pole of the switch needs a dedicated GPIO pin.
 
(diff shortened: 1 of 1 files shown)


docs/docs/config/kscan.md
```diff
@@ -98,6 +94,25 @@ Assuming the switches connect each GPIO pin to the ground, the [GPIO flags](http
     };
 ```
 
+By default, a switch will drain current through the internal pull up/down resistor whenever it is pressed. This is not ideal for a toggle switch, where the switch may be left in the "pressed" state for a long time. Enabling `toggle-mode` will make the driver flip between pull up and down as the switch is toggled to optimize for power. For `toggle-mode` to work correctly each pole of the switch needs a dedicated GPIO pin.
```
petejohanson, line 97: Don't really like the wording of this (you copied this directly, just noticing it now):

> Enabling `toggle-mode` will make the driver flip between pull up and down as the switch is toggled to optimize for power

Maybe "will make the driver enable/disable the internal pull up/down as needed when the switch is toggled to minimize power draw"?

docs/docs/config/kscan.md
```diff
@@ -98,6 +94,25 @@ Assuming the switches connect each GPIO pin to the ground, the [GPIO flags](http
     };
 ```
 
+By default, a switch will drain current through the internal pull up/down resistor whenever it is pressed. This is not ideal for a toggle switch, where the switch may be left in the "pressed" state for a long time. Enabling `toggle-mode` will make the driver enable and disable the internal pull up/down resistor as needed when the switch is toggled to minimise power draw. For `toggle-mode` to work correctly each pole of the switch needs a dedicated GPIO pin.
+
+`toggle-mode` applies to all switches handled by the instance of the driver. To use a toggle switch with other, non-toggle, direct GPIO switches, create two instances of the direct GPIO driver, one with `toggle-mode` and the other without. Then, use a [composite driver](#composite-driver) to combine them. The state of the switch is read on power on, so if the switch is moved whilst the board is off this will get correctly interpreted by the driver.
+
+When using `toggle-mode` the pull resistors get automatically set by the driver and should not be set in the devicetree. Assuming the common pole of the switch is connected to ground with an SP3T switch:
```
caksoylar, line 101:
```suggestion
When using `toggle-mode` the pull resistors get automatically set by the driver and should not be set in the devicetree via GPIO flags. Assuming the common pole of the switch is connected to ground with an SP3T switch:
```

caksoylar (approved):
LGTM, thanks for the update!

caksoylar:
@petejohanson you can merge if/when you are happy with the rewording.

petejohanson (approved):

PR #2302, opened 2024-05-12
behavior : programmable buttons

apewo:
implement programmable buttons with &pb 

programmable buttons are better described here : https://www.phoronix.com/news/Linux-HID-Programmable-Buttons

Supported only on linux (i think), they allow to define up to 32 shortcuts that will not overlap with any other software shortcuts.

I personally use them on i3 for all my gui shortcuts (rofi, ...), but they could be useful also for macropads, ...

They require a user land application to consume them so i uploaded a demo application here if anybody wants to test them : [pb demo](https://github.com/apewo/pbuttons_demo)


and-elf:
what is the big difference between this and macros?

Issue #2300, opened 2024-05-09
 Soft off on the right side of Corne split keyboard issues


GermanG:
This a sequel of #2274, which is partially solved in  #2285
The behavior is not consistent, most of the times works perfectly, but sometimes the right side is still on.
Reporting this now, I'll update my firmware to debug and update this issue when it happens.
On the other hand I don't know where to look to use the `wakeup-source` in my config, because there are some use cases that have no reset switch.

GermanG:
On the right side there isn't any event, on the left I saw this
```
[00:00:07.966,278] <dbg> zmk: position_state_changed_listener: 29 bubble (no undeci
ded hold_tap active)                                                               
  
[00:00:07.966,308] <dbg> zmk: zmk_keymap_apply_position_state: layer: 2 position: 2
9, binding name: soft_off  
[00:00:07.966,369] <dbg> zmk: split_bt_invoke_behavior_payload:                    
[00:00:07.966,430] <dbg> zmk: split_central_split_run_callback:                                                                                                      
[00:00:13.844,177] <dbg> zmk: kscan_matrix_read: Sending event at 2,5 state
```

But when it works I saw:
```
[00:00:10.537,170] <dbg> zmk: kscan_matrix_read: Sending event at 2,5 state off
[00:00:10.537,261] <dbg> zmk: zmk_ksc
```

My educated guess is that the left side goes down before completing the request to the right side.

GermanG:
I added this to app/src/behaviors/behavior_soft_off.c
```
           if (!IS_SPLIT_PERIPHERAL) {
                k_usleep(100000);
            }
```
the whole fragment looks like
```
        if (hold_time > config->hold_time_ms) {
            if (!IS_SPLIT_PERIPHERAL) {
                k_usleep(100000);
            }
            zmk_pm_soft_off();
        } else {
            LOG_INF("Not triggering soft off: held for %d and hold time is %d", hold_time,
                    config->hold_time_ms);
        }
```

So far so good. I think the issue is introduced because of the event is triggered when the key is released, so it's a matter of luck.

petejohanson:
> I added this to app/src/behaviors/behavior_soft_off.c
> 
> ```
>            if (!IS_SPLIT_PERIPHERAL) {
>                 k_usleep(100000);
>             }
> ```
> 
> the whole fragment looks like
> 
> ```
>         if (hold_time > config->hold_time_ms) {
>             if (!IS_SPLIT_PERIPHERAL) {
[...]

PR #2303, opened 2024-05-12
soft_off: Adding delay in central for split keyboards  

GermanG:
This should fix the issue I reported in #2300

### DIFF
--- a/app/src/behaviors/behavior_soft_off.c
+++ b/app/src/behaviors/behavior_soft_off.c
@@ -57,7 +57,7 @@ static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
         uint32_t hold_time = k_uptime_get() - data->press_start;
 
         if (hold_time > config->hold_time_ms) {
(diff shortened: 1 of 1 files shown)


app/src/behaviors/behavior_soft_off.c
```diff
@@ -57,6 +57,9 @@ static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
         uint32_t hold_time = k_uptime_get() - data->press_start;
 
         if (hold_time > config->hold_time_ms) {
+            if (IS_ENABLED(CONFIG_ZMK_SPLIT) && IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)) {
+                k_usleep(100000);
```
petejohanson, line 61: If you're simply trying to sleep for 100ms, just use:

```suggestion
                k_sleep(K_MSEC(100));
```

GermanG: done!

petejohanson (approved):
Thanks!

Issue #1599, opened 2022-12-20
feat: add option to disable storing state to settings


xudongzheng:
Currently ZMK uses settings to periodically remember the latest state for `&out`, `&ext_power`, and others.

Some users may be prefer for such state to not persist. Not using settings means you are guaranteed a consistent state after each reset. The 60 second default debounce period can sometimes make it difficult to know what the initial state is.

One approach is setting `CONFIG_ZMK_SETTINGS_SAVE_DEBOUNCE` to a really high value. However this is imperfect as setting that already exists on storage will persist indefinitely.

Ideally there would be a special value (such as `CONFIG_ZMK_SETTINGS_SAVE_DEBOUNCE=-1`) or a separate Kconfig value that disables state from being read from settings (and uses the default state). Instead of just `IS_ENABLED(CONFIG_SETTINGS)`, it would check the proposed setting as well https://github.com/zmkfirmware/zmk/blob/main/app/src/ext_power_generic.c#L144.

joelspadin:
I'm not sure I understand this. How would this be different from disabling `CONFIG_SETTINGS`?

petejohanson:
> I'm not sure I understand this. How would this be different from disabling `CONFIG_SETTINGS`?

Unfortunately, this would also break BT key storage... So we would need something more nuanced.

Timoyoungster:
Wouldn't it be enough to change this ```CONFIG_SETTINGS``` (line 44) to a new config value?

https://github.com/zmkfirmware/zmk/blob/e22bc7620cef763d9ad80e9b98182273de9973db/app/src/endpoints.c#L43-L49

If that's correct I'd also be happy to create a PR.
*(would be my first one though, so it might be a pretty bad one)*

Timoyoungster:
Ok, clear. But it won't break any other functionality this way right?

Timoyoungster:
If someone else is still waiting for this, I created a PR (#2308) 
and would be happy if also other people could test it. (I am only able to verify on one single keyboard)

PR #2308, opened 2024-05-23
feat: adding separate settings for automatic saving of the selected endpoint / ext_power state

Timoyoungster:
Hello,

it's my first pull request so I hope I did everything right :)

This adds two new config values for en-/disabling the automatic saving of the selected endpoint / ext_power state. (#1599)

Per default both are set to `y` to keep the default behaviour the same.

I tested it on my own keyboard and as far as I can tell it works, but I'd be grateful if also other people could test it for their boards.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -635,6 +635,16 @@ config ZMK_SETTINGS_SAVE_DEBOUNCE
     int "Milliseconds to debounce settings saves"
     default 60000
 
+config ZMK_SETTINGS_SAVE_ENDPOINTS
+    bool "save endpoints configuration"
+    depends on CONFIG_SETTINGS
+    default y
+
+config ZMK_SETTINGS_SAVE_EXT_POWER
+    bool "save ext power state"
+    depends on CONFIG_SETTINGS
+    default y
+
 #SETTINGS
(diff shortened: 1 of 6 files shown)


hyx0329:
Hi, it looks the automatic saving is disabled, but the loading of those values are not disabled(if they exist due to an old firmware and not erased). *It can be a feature XD, something disables automatic saving but allows manual saving.*

Quick notes for anyone interested:
- search `read_cb(cb_arg, &preferred_transport` for transport(USB/BT) loading code
- search `read_cb(cb_arg, &active_profile` for BT profile loading code

My current approach is just commenting out the related code(so they are disabled) and that leads to a hacky but satisfying result :)

Timoyoungster:
My primary reason for creating this PR was that I wanted to save write cycles to the eeprom's, since I sometimes switch a lot between bluetooth and usb when I'm working on my PC and laptop simultaneously.

Due to this regular switching I didn't really consider the default values and assumed that it would be set somewhere in the building/flashing process.

But if it's desired I could either add a config value for the defaults and disable the loading code or - as you mentioned - add a behavior to manually save the defaults.

PR #2313, opened 2024-05-27
fix(battery): Error sensor channel name

dymax-jp:
When using `CONFIG_ZMK_BATTERY_REPORTING_FETCH_MODE_LITHIUM_VOLTAGE` get a message in the log `Failed to fetch battery values: -...`

In [battery.c](https://github.com/zmkfirmware/zmk/blob/main/app/src/battery.c) sensor channel name: `SENSOR_CHAN_VOLTAGE` but in [battery_common.c](https://github.com/zmkfirmware/zmk/blob/main/app/module/drivers/sensor/battery/battery_common.c) : `SENSOR_CHAN_GAUGE_VOLTAGE`

```
battery_common.c :

int battery_channel_get(const struct battery_value *value, enum sensor_channel chan,
                        struct sensor_value *val_out) {
    switch (chan) {
    case SENSOR_CHAN_GAUGE_VOLTAGE:
        val_out->val1 = value->millivolts / 1000;
        val_out->val2 = (value->millivolts % 1000) * 1000U;
        break;

    case SENSOR_CHAN_GAUGE_STATE_OF_CHARGE:
        val_out->val1 = value->state_of_charge;
        val_out->val2 = 0;
        break;

    default:
        return -ENOTSUP;
    }

    return 0;
}
```

### DIFF
--- a/app/src/battery.c
+++ b/app/src/battery.c
@@ -69,14 +69,14 @@ static int zmk_battery_update(const struct device *battery) {
         return rc;
     }
 #elif IS_ENABLED(CONFIG_ZMK_BATTERY_REPORTING_FETCH_MODE_LITHIUM_VOLTAGE)
-    rc = sensor_sample_fetch_chan(battery, SENSOR_CHAN_VOLTAGE);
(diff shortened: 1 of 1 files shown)


Nick-Munnich:
Closing as no files are being changed.

PR #2315, opened 2024-05-29
pre-commit: use versioned clang-format hook

maximevince:
slight change to the `pre-commit` config.
the version of clang-format used determined the way the code is formatted.

I was having a lot of trouble because my local version of clang-format (Arch Linux) is more recent than the one running on Github Actions. Therefore, the pre-commit action would almost always fail because of difference in clang-format output.

Using this version of the `pre-commit` config, it specifies the exact version of `clang-format` to be used, which I believe matches what is currently using on Github.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,10 +5,11 @@ repos:
     hooks:
       - id: remove-tabs
         exclude: "vendor-prefixes\\.txt$"
-  - repo: https://github.com/pocc/pre-commit-hooks
-    rev: v1.3.5
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v18.1.8
     hooks:
       - id: clang-format
+        types_or: [c++, c]
         args:
           - -i
   - repo: https://github.com/pre-commit/mirrors-prettier
--- a/app/boards/arm/corneish_zen/widgets/battery_status.c
+++ b/app/boards/arm/corneish_zen/widgets/battery_status.c
@@ -68,7 +68,7 @@ void battery_status_update_cb(struct battery_status_state state) {
 static struct battery_status_state battery_status_get_state(const zmk_event_t *eh) {
     const struct zmk_battery_state_changed *ev = as_zmk_battery_state_changed(eh);
 
-    return (struct battery_status_state) {
+    return (struct battery_status_state){
         .level = (ev != NULL) ? ev->state_of_charge : zmk_battery_state_of_charge(),
 #if IS_ENABLED(CONFIG_USB_DEVICE_STACK)
         .usb_present = zmk_usb_is_powered(),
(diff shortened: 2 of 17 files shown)


caksoylar:
There was some investigation on Discord by @joelspadin on the versioning: https://discord.com/channels/719497620560543766/719544230497878116/1203479919821328395. In case you don't have access, his conclusion was that the CI seemed to be running 14.0.0.

joelspadin:
I think specifying a version is a good change, though I don't have strong feelings on what that version should be. If we are going to upgade from 14, it should ideally be to something that is easy to install on a majority of OSes.

Ubuntu 22.04's `llvm` package is version 14.0.0. Version 15.0.7 is also available.

Ubuntu 24.04 isn't available for WSL yet, so I don't know what versions it provides.

I'm not sure which other distros/versions of Linux are commonly used.

On Windows, winget has every release up through 18.1.6.

I don't have a Mac, so no idea what versions are easily installable there.

maximevince:
A lot of the popular Linux distro's are based on Ubuntu, so that's definitely a big one.

Here's an overview of what I could find:
- Ubuntu 24.04LTS (noble) has clang-format 18, it seems: https://packages.ubuntu.com/noble/clang-format
- Fedora 40 seems to have v18 as well: https://packages.fedoraproject.org/pkgs/clang/clang/
- Arch Linux, still seems to be stuck on version 17 for some reason: https://archlinux.org/packages/extra/x86_64/clang/
- On MacOS `brew` gives you v18 as well:  https://formulae.brew.sh/formula/clang-format

So I would suggest moving to clang-format v18, if everyone agrees?

petejohanson:
> A lot of the popular Linux distro's are based on Ubuntu, so that's definitely a big one.
> 
> Here's an overview of what I could find:
> 
>     * Ubuntu 24.04LTS (noble) has clang-format 18, it seems: https://packages.ubuntu.com/noble/clang-format
> 
>     * Fedora 40 seems to have v18 as well: https://packages.fedoraproject.org/pkgs/clang/clang/
> 
>     * Arch Linux, still seems to be stuck on version 17 for some reason: https://archlinux.org/packages/extra/x86_64/clang/
> 
>     * On MacOS `brew` gives you v18 as well:  https://formulae.brew.sh/formula/clang-format
> 
> 
> So I would suggest moving to clang-format v18, if everyone agrees?

I'd be happy to move to v18 and standardize there.

petejohanson:
@maximevince Do you have any interest in finishing this up with a move to v18? If not, are you okay if someone takes this over?

Thanks again for the PR!

maximevince:
@petejohanson absolutely!
I chose v18.1.8 now, which is the latest release for clang-format. Are you okay with using that?
It might require some reformatting of existing files.

PR updated.

petejohanson:
> @petejohanson absolutely!
> I chose v18.1.8 now, which is the latest release for clang-format. Are you okay with using that?
> It might require some reformatting of existing files.
> 
> PR updated.

Yes, and yes. Will review this weekend. Thanks.

petejohanson:
@maximevince Ok, after approving the GHA, it finds all the code that needs fixing after the update, indeed. Can you roll that into this PR, and we'll pull the bandaid and merge this ASAP?

Thanks!

maximevince:
Yay! Seems green!

petejohanson (approved):
Thanks!

PR #2318, opened 2024-05-31
fix(boards): Suppress spurious devicetree warning for nRF5x boards

ReFil:
nRF5x boards throw a spurious warning for duplicate unit-address when the devicetree gets processed

Upstream issue:  https://github.com/zephyrproject-rtos/zephyr/issues/29713

Warning suppressed as per: https://docs.zephyrproject.org/latest/build/dts/intro-input-output.html

Tested on a few nRF5x boards and i have no reason to think it would break things on the untested ones


ReFil:
Looking at the results of the CI it seems that the pre_dt_board.cmake file needs to be placed here:https://github.com/zmkfirmware/zephyr/tree/v3.5.0%2Bzmk-fixes/boards/arm/seeeduino_xiao_ble for the seeduino xiao BLE

petejohanson (approved):
Thanks!

Will need to fix the seeeduino XIAO BLE upstream and cherry pick in, to address that one. Good to fix these for now though to have good in tree reference boards.

PR #2319, opened 2024-06-03
feat(docs) Splitting up the troubleshooting page

Nick-Munnich:
This PR splits up the troubleshooting page into three parts:

- Building issues
- Flashing issues
- Connection issues

I intentionally made the minimal changes possible to make this easier to review than #2272 - all of the content is lifted directly from the current troubleshooting sections in getting-started/troubleshooting and features/bluetooth. Some additional lines were added to make it tidy, and some headers depth was changed.

I believe this would make it easier for users to find the information corresponding to a particular problem. Another goal is to make it easier to expand on the troubleshooting section in the future. Among other things I would like to add a "hardware issues" page to the three in a future PR.


docs/docs/troubleshooting/connection-issues.md
```diff
@@ -0,0 +1,115 @@
...
+
+1. [Open the GitHub `Actions` tab and select the `Build` workflow](https://github.com/zmkfirmware/zmk/actions?query=workflow%3ABuild+branch%3Amain+event%3Apush).
+1. Find one of the 'results' for which the core-coverage job was successfully run, indicated by a green checkmark in the core-coverage bubble like the image example below.
+1. From the next page under "Artifacts", download and unzip the `<board_name>-settings_reset-zmk` zip file for the UF2 image.
+
+| ![Successful core-coverage Job](../../docs/assets/troubleshooting/splitpairing/corecoverage.png) |
+| :----------------------------------------------------------------------------------------------: |
+|    An example of a successful core-coverage job which will produce a settings_reset firmware.    |
+
+### Reset Split Keyboard Procedure
```
caksoylar, line 50: I think you intentionally changed the header levels here compared to "Option" items, which makes sense. But it's a bit awkward without a level three header before this one. Maybe we add a `### Acquiring a Reset UF2` around line 22?

caksoylar (commented):
LGTM overall, with one nit, thanks. (And it was easy to review too.) I wasn't sure if it was worth splitting, but seeing the end result it seems useful and gives us more room to expand each item (or add a new one).

docs/docs/troubleshooting/connection-issues.md
Nick-Munnich: Inconsistent casing is annoying, I agree. Do you want to decide on a casing standard? A quick search pulled up [this tool](https://capitalizemytitle.com/style/ap). I'm not sure an issue needs to be raised for that sort of decision, personally.

I've added the suggested header, but be aware that I do want to reorganize this section some more, in particular using tabs for the reset UF2 acquisition bit. Another thing for the future, though!

caksoylar: > Inconsistent casing is annoying, I agree. Do you want to decide on a casing standard? A quick search pulled up [this tool](https://capitalizemytitle.com/style/ap). I'm not sure an issue needs to be raised for that sort of decision, personally.

Does "Title Case" up to a certain level of headers (1-3?) and "Sentence case" after would make sense, perhaps? I don't think it's a big deal what it is exactly, but consistency inside and across pages would be nice to have.

> I've added the suggested header, but be aware that I do want to reorganize this section some more, in particular using tabs for the reset UF2 acquisition bit. Another thing for the future, though!

If it's just a structural change, I think it might make sense to add it here but up to you.

Nick-Munnich: I agree with title case up to 3 and sentence case after. While not strictly within scope of this PR, I figured I may as well add it to this since it is a trivial change. Also added a note in dev/docs on it. 

Regarding reset UF2, I think I wanted to make more changes at some point but I can't remember what they were (aside from bolding and underlining that both halves need resetting). I added said change and the tabs, so won't go back to it until I remember what I wanted to do.

Nick-Munnich: I forgot to mention - the header cases that I did not change are in the FAQ and pre-commit, since I wasn't sure how to handle those. Title case FAQ just looks odd, and pre-commit is a lowercase thing. Also did not change any of the header level 4+ ones, those don't appear in the sidebar to the right so are more annoying to check (and I would need to look up how to change that).

caksoylar: Thanks for doing this, it's quite a lot of work! I agree with the decisions on FAQ and pre-commit.

I'll add a commit to handle the 4+ level headers (I'll also squash and rename commits as appropriate).

caksoylar (approved):

Nick-Munnich:
Had a look at the changed level 4+ headers just to double check - looks all good to me.

PR #2334, opened 2024-06-13
fix: adding option to separate implicit mod release from key release

Timoyoungster:
This adds a new config value `ZMK_HID_SEPARATE_MOD_RELEASE_REPORT` where, if enabled, the report for a key release is sent separately to the accompanying modifier release signals, which are then sent in a second report.

This fixes an issue where certain applications are unable to work with implicitly modified keys (e.g. colon) due to them registering the modifier release prior to the actual key release.

Have tested this on my personal keyboard and `wev` now shows the signals in the correct order.
=> **Previously:** ```LSHIFT (pressed) -> colon (pressed) -> LSHIFT (released) -> **semi**colon (released)```
=> **Now:** ```LSHIFT (pressed) -> colon (pressed) -> colon (released) -> LSHIFT (released)```

### DIFF
--- a/app/src/hid_listener.c
+++ b/app/src/hid_listener.c
@@ -66,6 +66,17 @@ static int hid_listener_keycode_released(const struct zmk_keycode_state_changed 
         return err;
     }
 
+#if IS_ENABLED(CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT)
+
+    // send report of normal key release early to fix the issue
+    // of some programs recognizing the implicit_mod release before the actual key release
+    err = zmk_endpoints_send_report(ev->usage_page);
+    if (err < 0) {
+        LOG_ERR("Failed to send key report for the released keycode (%d)", err);
+    }
+
+#endif // IS_ENABLED(CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT)
+
     explicit_mods_changed = zmk_hid_unregister_mods(ev->explicit_modifiers);
     // There is a minor issue with this code.
     // If LC(A) is pressed, then LS(B), then LC(A) is released, the shift for B will be released
@@ -73,7 +84,7 @@ static int hid_listener_keycode_released(const struct zmk_keycode_state_changed 
     // Solving this would require keeping track of which key's implicit modifiers are currently
     // active and only releasing modifiers at that time.
     implicit_mods_changed = zmk_hid_implicit_modifiers_release();
-    ;
+
     if (ev->usage_page != HID_USAGE_KEY &&
         (explicit_mods_changed > 0 || implicit_mods_changed > 0)) {
         err = zmk_endpoints_send_report(HID_USAGE_KEY);
--- a/docs/docs/config/system.md
+++ b/docs/docs/config/system.md
@@ -29,10 +29,11 @@ Making changes to any of the settings in this section modifies the HID report de
 
 :::
 
-| Config                                | Type | Description                                                    | Default |
-| ------------------------------------- | ---- | -------------------------------------------------------------- | ------- |
-| `CONFIG_ZMK_HID_INDICATORS`           | bool | Enable receipt of HID/LED indicator state from connected hosts | n       |
-| `CONFIG_ZMK_HID_CONSUMER_REPORT_SIZE` | int  | Number of consumer keys simultaneously reportable              | 6       |
+| Config                                       | Type | Description                                                    | Default |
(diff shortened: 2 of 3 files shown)


app/src/hid_listener.c
```diff
@@ -66,14 +66,25 @@ static int hid_listener_keycode_released(const struct zmk_keycode_state_changed
         return err;
     }
 
+#if IS_ENABLED(CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT)
+
+    // send report of normal key release early to fix the issue
+    // of some programs recognizing the implicit_mod release before the actual key release
+    err = zmk_endpoints_send_report(ev->usage_page);
+    if (err < 0) {
+        LOG_ERR("Failed to send key report for release keycode for consumer page event (%d)", err);
```
caksoylar, line 75: This log message seems incorrect.

Timoyoungster: Wow, that was quick.
I copy pasted that from below and changed 'changed modifiers' to released key as I wasn't really sure what else to log there.

caksoylar: You know the code change better than I do :) but the "consumer page event" part seems incorrect, at least.

caksoylar: "Failed to send separate mod release report (%d)" might be good enough.

Timoyoungster: or what about 'Failed to send key report for the released key (%d)'
because it's sending the released key first with the new call, and in the return it sends the modifier changes.

docs/package-lock.json
```diff
@@ -4897,11 +4897,12 @@
       }
     },
     "node_modules/braces": {
-      "version": "3.0.2",
```
petejohanson, line 4900: Please take this file change out of this commit/PR.

petejohanson (changes requested):
Code seems reasonable, just a comment on an unrelated file change that snuck in. Thanks for the PR!

docs/package-lock.json
Timoyoungster: I'll try👍

docs/docs/config/system.md
```diff
@@ -29,10 +29,11 @@ Making changes to any of the settings in this section modifies the HID report de
...
 
-| Config                                | Type | Description                                                    | Default |
-| ------------------------------------- | ---- | -------------------------------------------------------------- | ------- |
-| `CONFIG_ZMK_HID_INDICATORS`           | bool | Enable receipt of HID/LED indicator state from connected hosts | n       |
-| `CONFIG_ZMK_HID_CONSUMER_REPORT_SIZE` | int  | Number of consumer keys simultaneously reportable              | 6       |
+| Config                                       | Type | Description                                                    | Default |
+| -------------------------------------------- | ---- | -------------------------------------------------------------- | ------- |
+| `CONFIG_ZMK_HID_INDICATORS`                  | bool | Enable receipt of HID/LED indicator state from connected hosts | n       |
+| `CONFIG_ZMK_HID_CONSUMER_REPORT_SIZE`        | int  | Number of consumer keys simultaneously reportable              | 6       |
+| `CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT` | bool | Release the Modifiers separate from and after the modified key | n       |
```
petejohanson, line 36: IMHO, this description doesn't make sense to me. Perhaps "Release modifiers before the key they modify" or somesuch? Maybe @caksoylar has a better idea for a succinct description of this.

petejohanson (commented):
Just one comment on the docs portion.

docs/docs/config/system.md
caksoylar: "Send modifier key release events before non-modifiers"? Given this is practically happening only for modifier functions, we could add "for modifier functions" perhaps.

Timoyoungster: But isn't it the other way around?
We first send the normal key release event and then the modifier release after that.

Timoyoungster: May be something like: "Send Modifier release event after Non-Modifier release event."

caksoylar: > But isn't it the other way around?
We first send the normal key release event and then the modifier release after that.

Yes, I mixed it up in my comment. I also wrote the "before" version rather than "after" since technically the new report is the earlier one, but functionally either way works.

Timoyoungster: Ok perfect. Then my commit should resolve the issue. :)

Timoyoungster:
@petejohanson is the phrasing ok now?

petejohanson (approved):
Thanks!

PR #2335, opened 2024-06-17
feat(ble): Set device name at runtime

ReFil:
This allows for the device name to be set at runtime, possible uses for this include a custom display screen that lets you type it in using the keycode state event, future configuration in zmk studio, or altering it per profile as per #1169

Testing shows restarting advertising was necessary to update the new name, although some host OS's don't reflect the name change properly even with this in mind, also of note is some hosts don't update the name after pairing too


ReFil:
I can probably make a function to do it for USB, I don't think one function should do both as there's no logical place for that one function to live, whatever system the developer is using at the other end should call both to change both. I think there's also a good argument for keeping them separate as there are good reasons to want to change them independently e.g. per profile naming

petejohanson (commented):
So, upstream Zephyr already has a Kconfig symbol, `CONFIG_BT_DEVICE_NAME_GATT_WRITABLE` which will set up the GAP service to allow the GAP Device Name as writable, and will use `bt_set_name`
to set the provided value. 

Hosts that support it will detect a GAP service with the name characteristic having WRITE permissions/flags, and will actually write to the device after the user changes the device name.

Feels like we should be leveraging that existing work as much as possible.

ReFil:
Ah neat I didn't know that. I did find in my testing that results from this were inconsistent. iOS and Linux reflected the name change whilst scanning, android didn't. That should probably sort this. I've also got it saving and restoring from flash, and will incorporate that too

petejohanson:
> Ah neat I didn't know that. I did find in my testing that results from this were inconsistent. iOS and Linux reflected the name change whilst scanning, android didn't. That should probably sort this. I've also got it saving and restoring from flash, and will incorporate that too

Yeah, `bt_set_name` implementation will store to settings if it's enabled. See `hci_core.c:bt_set_name`

ReFil:
I've got it working now using `CONFIG_BT_DEVICE_NAME_DYNAMIC`. If i'm reading the PR associated with this function properly i think we get up to 27 chars for device name now too which is a neat bonus. The saving to flash can't be debounced, because it's outside of our control but i cant imagine that as a major problem. The advertising packet had to be changed, as ZMK used the advertising packet to report the name previously.
I suppose the final question is do we want the name to be writable by the host device? There's no downsides to it i guess

ReFil:
Just checked and `CONFIG_BT_DEVICE_NAME_GATT_WRITABLE` is enabled even when only enabling `CONFIG_BT_DEVICE_NAME_DYNAMIC`. Name can sucessfully be read and written using nRF connect. all seems to be working properly

app/include/zmk/ble.h
```diff
@@ -36,6 +36,8 @@ char *zmk_ble_active_profile_name(void);
 
 int zmk_ble_unpair_all(void);
 
+void zmk_ble_set_device_name(char *name);
```
petejohanson, line 39: This should really return `int` and propagate any errno returned by `bt_set_name`.

app/Kconfig
```diff
@@ -147,6 +147,7 @@ menuconfig ZMK_BLE
     select BT_SMP_APP_PAIRING_ACCEPT
     select BT_PERIPHERAL
     select BT_DIS
+    select BT_DEVICE_NAME_DYNAMIC
```
petejohanson, line 150: I'd rather we `imply` here in case folks want to disable this.

petejohanson (changes requested):
Couple minor comments.

app/include/zmk/ble.h
ReFil: Addressed. Also returned any errors from `update_advertising`

app/Kconfig
ReFil: Addressed

app/src/ble.c
```diff
@@ -56,19 +56,18 @@ enum advertising_type {
...
+                    BT_GAP_ADV_FAST_INT_MIN_2, BT_GAP_ADV_FAST_INT_MAX_2, NULL)
 
 static struct zmk_ble_profile profiles[ZMK_BLE_PROFILE_COUNT];
 static uint8_t active_profile;
 
 #define DEVICE_NAME CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
 
-BUILD_ASSERT(DEVICE_NAME_LEN <= 16, "ERROR: BLE device name is too long. Max length: 16");
+BUILD_ASSERT(DEVICE_NAME_LEN <= 27, "ERROR: BLE device name is too long. Max length: 27");
```
petejohanson, line 68: Sorry. As I think on this... I believe we limited this further because longer names would cause our advertising data to be too large, when combined with our services listed. So even if the name pieces allow this. I'm not sure we can actually use longer names.

Did you test with a longer name?

ReFil: Good spot, I tested it with an 18 char name, it worked and i called it good, I've just gone back to it, I can get a 20 char name to work but no longer. The characteristic doesn't seem to accept the write if it's longer

ReFil: Happy to cap it at the original 16 if you want. It's possible my laptop is accepting the overlength advertising data when it shouldnt be

petejohanson: > Happy to cap it at the original 16 if you want. It's possible my laptop is accepting the overlength advertising data when it shouldnt be

See https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/bluetooth/host/adv.c#L187

By default, asking Zephyr to advertise the name will put the name in the SD data instead of the AD packet itself, which is likely why this is working for longer names.

Two years ago or so, having name in the SD meant some hosts failed to display a name when pairing. We can test this again *now* and see if that's still a problem or not. SD certainly would be better, if it works.

petejohanson: IIRC, it was macOS at the time that wouldn't show a name if it was in SD

ReFil: Alternatively it looks like we can force it to put the name in AD and restrict to the original 16 chars

petejohanson: > Alternatively it looks like we can force it to put the name in AD and restrict to the original 16 chars

I'd vote for this for now. Can we do that for names that come in from a write to the GAP characteristic though?

ReFil: Just tested it and the new name is properly reflected in AD after writing to the GAP characteristic

ReFil: of note is after pairing android doesn't reflect the new name, it's got some local name caching because there's a mechanism where you can give it a name that stays on the host device to refer to it

petejohanson:
@ReFil Check the BLE test error.. maybe advertising data changes?

ReFil:
Yes I think that's the issue, name is now at the end of the ad packet not the beginning. Will sort that when I have a mo

petejohanson:
> Yes I think that's the issue, name is now at the end of the ad packet not the beginning. Will sort that when I have a mo

@ReFil Nudge.

ReFil:
@petejohanson Sorry for the delay in addressing it. I've tweaked the tests to remove the offending line from the logs and they all pass now

petejohanson (approved):
Thanks!

Issue #2356, opened 2024-07-03
`&macro_release` do not repsect the macros `wait-ms` resulting in lost keystrokes


SolidHal:
while messing around with midi support and macros, i noticed that `&macro_release` and `&macro_press` handle the `wait-ms` parameter differently. 

example macro:
```
        ma_Eb: major_Eb {
            compatible = "zmk,behavior-macro";
            wait-ms = <2>;
            tap-ms = <0>;
            #binding-cells = <0>;
            bindings
            = <&macro_press &mo MA_Eb>
            , <&macro_press &midi NOTE_Eb_4>
            , <&macro_press &midi NOTE_G_4>
            , <&macro_press &midi NOTE_Bb_4>
            , <&macro_pause_for_release>
            , <&macro_release &mo MA_Eb>
            , <&macro_release &midi NOTE_Eb_4>
            , <&macro_release &midi NOTE_G_4>
            , <&macro_release &midi NOTE_Bb_4>
            ;
        };
```

i would expect that when this macro is pressed there is a `2ms` wait between each press, and then a `2ms` wait between each release. 

but in midisnoop i observed the presses each has a `2ms` wait, the releases do not. in the following example we can see that one of the releases aka `Note Off` events gets missed entirely. 

![image](https://github.com/zmkfirmware/zmk/assets/12972568/0a638fe7-b507-4230-8b52-3c4dbc854664)

i put some debug prints into `behavior_macro.c` and confirmed the key release events were being queued with a `wait_ms` of `0`
and the presses were being queud with a `wait_ms` of `2`

```
static void queue_macro(uint32_t position, const struct zmk_behavior_binding bindings[],
                        struct behavior_macro_trigger_state state,
                        const struct zmk_behavior_binding *macro_binding) {
    LOG_DBG("Iterating macro bindings - starting: %d, count: %d", state.start_index, state.count);
    for (int i = state.start_index; i < state.start_index + state.count; i++) {
        if (!handle_control_binding(&state, &bindings[i])) {
            struct zmk_behavior_binding binding = bindings[i];
            replace_params(&state, &binding, macro_binding);

            switch (state.mode) {
            case MACRO_MODE_TAP:
                zmk_behavior_queue_add(position, binding, true, state.tap_ms);
                zmk_behavior_queue_add(position, binding, false, state.wait_ms);
                break;
            case MACRO_MODE_PRESS:
              LOG_ERR("MACRO MODE PRESS wait_ms: %d", state.wait_ms);
[...]

PR #2357, opened 2024-07-03
fix(macro): set the macro release states wait_ms

SolidHal:
when we create a release trigger state for each press state we miss setting the releases wait_ms field, resulting in key releases all occuring instantaneously

this can result in lost key release commands
see issue #2356


caksoylar:
It might be nice to add a test for this, perhaps?

PR #2358, opened 2024-07-03
fix: Improve startup time with proper settings loading.

petejohanson:
* Avoid doing duplicate calls to setings_load_subtree, which iterates NVS fully each time under the hood, and instead use on settings_load later in the lifecycle.

Found this during Studio work, after we ended up with many more settings entries. Pinned it down to the many calls to `settings_load_subtree` which under the hood was actually loading the full NVS data each time, doing a ton of duplicate work. This refactor moves away from duplicate subtree loads. Doing so requires a bit of care to make timing work and have the right things loaded when needed, especially for BT initialization.

### DIFF
--- a/app/src/split/bluetooth/central.c
+++ b/app/src/split/bluetooth/central.c
@@ -12,6 +12,7 @@
 #include <zephyr/bluetooth/uuid.h>
 #include <zephyr/bluetooth/gatt.h>
 #include <zephyr/bluetooth/hci.h>
+#include <zephyr/settings/settings.h>
 #include <zephyr/sys/byteorder.h>
 
 #include <zephyr/logging/log.h>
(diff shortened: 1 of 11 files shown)


app/src/split/bluetooth/central.c
```diff
@@ -865,13 +866,34 @@ int zmk_split_bt_update_hid_indicator(zmk_hid_indicators_t indicators) {
...
 static int zmk_split_bt_central_init(void) {
     k_work_queue_start(&split_central_split_run_q, split_central_split_run_q_stack,
                        K_THREAD_STACK_SIZEOF(split_central_split_run_q_stack),
                        CONFIG_ZMK_BLE_THREAD_PRIORITY, NULL);
     bt_conn_cb_register(&conn_callbacks);
 
-    return IS_ENABLED(CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START) ? 0 : start_scanning();
+#if IS_ENABLED(CONFIG_SETTINGS)
+    settings_register(&ble_central_settings_handler);
+    return 0;
```
caksoylar, line 893: Should this use `finish_init()` as well?

PR #2363, opened 2024-07-05
refactor(hog): Cleanup ble message sending

khoek:
We deduplicate a bunch of code removing about 80 lines (not clear from the diff due to dependence on other PR), fixing a very minor copy-paste typo in the process. No functional change.

Stacked on top of https://github.com/zmkfirmware/zmk/pull/2257 since we use a struct from there. Tested on an Adv360.

### DIFF
--- a/app/include/zmk/hid.h
+++ b/app/include/zmk/hid.h
@@ -254,6 +254,30 @@ struct zmk_hid_mouse_report {
 
 #endif // IS_ENABLED(CONFIG_ZMK_MOUSE)
 
+struct zmk_hid_report_body {
+    union {
+#if IS_ENABLED(CONFIG_ZMK_MOUSE)
+        struct zmk_hid_mouse_report_body mouse;
+#endif
+
+        struct zmk_hid_keyboard_report_body keyboard;
+        struct zmk_hid_consumer_report_body consumer;
+    } __packed;
+} __packed;
+
+struct zmk_hid_report {
+    union {
+#if IS_ENABLED(CONFIG_ZMK_USB_BOOT)
+        zmk_hid_boot_report_t boot;
+#endif
+
+        struct {
+            uint8_t report_id;
+            struct zmk_hid_report_body body;
+        } __packed;
+    } __packed;
+} __packed;
+
 zmk_mod_flags_t zmk_hid_get_explicit_mods(void);
 int zmk_hid_register_mod(zmk_mod_t modifier);
 int zmk_hid_unregister_mod(zmk_mod_t modifier);
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -648,6 +648,15 @@ config ZMK_SETTINGS_SAVE_DEBOUNCE
 #SETTINGS
 endif
 
+config ZMK_WORKAROUND_COPY_USB_TX_DATA_BUFFER
+    bool "Enable workaround for USB driver not copying data before transmit"
+    help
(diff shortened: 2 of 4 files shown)


app/include/zmk/hid.h
```diff
@@ -254,6 +254,30 @@ struct zmk_hid_mouse_report {
...
+#if IS_ENABLED(CONFIG_ZMK_MOUSE)
+        struct zmk_hid_mouse_report_body mouse;
+#endif
+
+        struct zmk_hid_keyboard_report_body keyboard;
+        struct zmk_hid_consumer_report_body consumer;
+    } __packed;
+} __packed;
+
+struct zmk_hid_report {
```
petejohanson, line 268: I kind of like this idea... I do wonder: Do we really need to keep around types like `struct zmk_hid_mouse_report` with this change?

petejohanson (changes requested):
Thanks for the PR! Seems like a very reasonable approach. See my one comment about some additional cleanup that's likely possibly with this change.

petejohanson:
@khoek Any chance you are still interested in working on this?

khoek:
@petejohanson Sorry about the radio silence, I forgot about your question, and remember interpreting it as slightly more contemplative at the time. Sure, I'll have a play around and see if I can get rid of them.

By the way while I have your attention: I've been having a bunch of trouble with applications correctly interpreting some taps coming out of hold-tap because of their very fast duration (coincidentally the same phenomenon which triggered the race condition bug which got me poking around in the first place)---mainly this is happening in e.g. video game engines which poll the key states themselves instead of handling events, and I guess the key is being press-released within a single frame. Anyhow, is there a natural way to integrate a configurable tiny delay into the current way ZMK works to alleviate this problem?---if it will be a bit awkward, is this more motivation for me to take another look at unblocking the main thread and adding a message queue for outbound USB messages (of course queuing a message to be sent later after some delay is then very cheap to implement in that setting)?

khoek:
By the way, besides the upsides about queuing multiple messages instead of waiting which I evangelized in that other thread, the only downside I could think of was that, naively implemented, there would be a tiny latency slowdown which people interested in maximum responsiveness wouldn't like. So I just wanted to note that I was attentive to that and I'd be changing the USB send function to immediately commit a USB transaction if there wasn't a pending operation in progress, and only queue in the alternative case (of if e.g. the message was to be transmitted after some delay).

petejohanson:
> @petejohanson Sorry about the radio silence, I forgot about your question, and remember interpreting it as slightly more contemplative at the time. Sure, I'll have a play around and see if I can get rid of them.
> 
> By the way while I have your attention: I've been having a bunch of trouble with applications correctly interpreting some taps coming out of hold-tap because of their very fast duration (coincidentally the same phenomenon which triggered the race condition bug which got me poking around in the first place)---mainly this is happening in e.g. video game engines which poll the key states themselves instead of handling events, and I guess the key is being press-released within a single frame. Anyhow, is there a natural way to integrate a configurable tiny delay into the current way ZMK works to alleviate this problem?---if that would be a bit awkward, is this more motivation for me to take another look at unblocking the main thread and adding a message queue for outbound USB messages (of course queuing a message to be sent later after some delay is then very cheap to implement in that setting)?

You'd need to modify the hold-tap behavior to add a devicetree configurable `release-delay` and implement a sleep for that amount in the decide code for hold-tap.

PR #2377, opened 2024-07-16
Fix max brightness in backlight metadata

ReFil:
Spotted this whilst working on a custom backlight behavior. The set brightness function in the backlighting code has a max of 100, as does the zephyr led-pwm driver https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/led/led_pwm.c

The range for the set brightness function should reflect this max, otherwise ~2/3rds of the possible brightness range will set the leds to the same brightness


petejohanson (approved):
Good catch. Thanks!

PR #2390, opened 2024-07-25
mini-feat(docs) Improvements to the layer documentation

### LINKED ISSUE #1284: Layer documentation is unclear
The layer documentation [1] does not clearly explain how layers are numbered. Also, it seems to give conflicting information: it explicitly gives a metaphor of layers as numbered by "[reverse-indexed] top-to-bottom" [1], but then the very first example of a keymap uses the exact opposite order with no clarification. [2]

For example: since the "layers" page is where users will go to if they're confused about how to define layers, it would help to say something like:

> Layers will be automatically assigned numbers, beginning with 0 (usually referred to as the 'default layer'), and increasing sequentially by 1 in the order they are defined in the keymap file. <<It would also help to clarify again the confusing fact that layers must be defined in the opposite order as was just explained in [1]>>

But AFAICT no such explanation exists, and so the ZMK user is left to head scratching and assumptions based off the very small number of examples.

And regarding the examples: consider that this is the official ZMK "Layers" documentation, and it gives only a measly 2-layer example. Adding an example with 3 or more will drive home the fact that layers are sequentially ordered; something that I, as a new ZMK user setting up a board for the first time, am still confused about.

[1] https://zmk.dev/docs/features/keymaps
[2] https://zmk.dev/docs/behaviors/layers

Nick-Munnich:
Rephased some things and added some clarification. The goal is to resolve #1284.

### DIFF
--- a/docs/docs/behaviors/layers.md
+++ b/docs/docs/behaviors/layers.md
@@ -9,7 +9,7 @@ Often, you may want a certain key position to alter which layers are enabled, ch
 Some of those behaviors are still in the works; the ones that are working now are documented here.
 
 :::note
-Activating a layer will not disable layers higher up in the "layer stack". See [Layers](../features/keymaps.mdx#layers) for more information.
+Multiple layers can be active at the same time and activating a layer will not deactivate layers higher up in the "layer stack". See [Layers](../features/keymaps.mdx#layers) for more information.
 :::
 
 ## Defines to Refer to Layers
--- a/docs/docs/behaviors/layers.md
+++ b/docs/docs/behaviors/layers.md
@@ -8,6 +8,10 @@ sidebar_label: Layers
 Often, you may want a certain key position to alter which layers are enabled, change the default layer, etc.
 Some of those behaviors are still in the works; the ones that are working now are documented here.
 
+:::note
+Activating a layer will not disable layers higher up in the "layer stack". See [Layers](../features/keymaps.mdx#layers) for more information.
+:::
+
 ## Defines to Refer to Layers
 
 When working with layers, you may have several different key positions with bindings that enable/disable those layers.
--- a/docs/docs/features/keymaps.mdx
+++ b/docs/docs/features/keymaps.mdx
@@ -53,7 +53,7 @@ When a key location is pressed/released, the _highest-valued currently active_ l
 Note that the _activation_ order isn't relevant for determining the priority of active layers, it is determined _only_ by the definition order.
 
 :::tip
-If you wish to use multiple base layers (with a [toggle](../behaviors/layers.md#toggle-layer)), e.g. one for qwerty and another for colemak, you will want these layers to have the lowest value possible. In other words, one should be layer `0`, and the other should be layer `1`. This allows other momentary layers activated on top of them to work with both.
+If you wish to use multiple base layers (with a [toggle](../behaviors/layers.md#toggle-layer)), e.g. one for QWERTY and another for Colemak layouts, you will want these layers to have the lowest value possible. In other words, one should be layer `0`, and the other should be layer `1`. This allows other momentary layers activated on top of them to work with both.
 :::
 
 ## Behavior Bindings
--- a/docs/docs/features/keymaps.mdx
+++ b/docs/docs/features/keymaps.mdx
@@ -46,9 +46,11 @@ in ZMK contains a set of bindings that bind a certain behavior to a certain key 
 
 All layers are assigned and referred to by a natural number, with the base layer being layer `0`. It is common to [use the C preprocessor to "name" layers](../behaviors/layers.md#defines-to-refer-to-layers), making them more legible.
 
-The default layer (usually the base layer) is always enabled. Certain bound behaviors may enable/disable additional layers.
+The default layer (the base layer with index 0) is always enabled. Certain bound behaviors may enable/disable additional layers.
 
 When a key location is pressed/released, the _highest-valued currently active_ layer is used. The press/release event is sent to the behavior bound at that position in said layer, for it to perform whatever actions it wants to in reaction to the event. The behavior can choose to "consume" the event, or "pass it along" and let the next highest-valued active layer _also_ get the event (whose behavior may continue "passing it along").
+
+Note that the _activation_ order isn't relevant for determining the priority of active layers, it is determined _only_ by the definition order.
 
 ## Behavior Bindings
 
@@ -132,7 +134,7 @@ Each layer of your keymap will be nested under the keymap node. Here is an examp
                 &kp A    &kp B    &kp C
             >;
         };
-    }
+    };
 ```
 
 Each layer should have:
@@ -149,7 +151,7 @@ Here is an example of a trio of layers for a simple 6-key macropad:
 <KeymapExample />
 
 :::note
-Even if layer `1` was to be activated after `2`, layer `2` would still have priority as it is higher valued. Behaviors such as [`&to`](../behaviors/layers.md#to-layer) can be used to enable one layer _and disable all other non-default layers_, though.
+Even if layer `1` was to be activated after `2`, layer `2` would still have priority as it is higher valued. Behaviors such as [To Layer (`&to`)](../behaviors/layers.md#to-layer) can be used to enable one layer _and disable all other non-default layers_, though.
 :::
 
 ### Complete Example
--- a/docs/docs/features/keymaps.mdx
+++ b/docs/docs/features/keymaps.mdx
@@ -52,6 +52,10 @@ When a key location is pressed/released, the _highest-valued currently active_ l
 
 Note that the _activation_ order isn't relevant for determining the priority of active layers, it is determined _only_ by the definition order.
 
+:::tip
+If you wish to use multiple base layers (with a [toggle](../behaviors/layers.md#toggle-layer)), e.g. one for qwerty and another for colemak, you will want these layers to have the lowest value possible. In other words, one should be layer `0`, and the other should be layer `1`. This allows other momentary layers activated on top of them to work with both.
+:::
+
 ## Behavior Bindings
 
 Binding a behavior at a certain key position may include up to two extra parameters that are used to
--- a/docs/docs/features/keymaps.mdx
+++ b/docs/docs/features/keymaps.mdx
@@ -4,7 +4,6 @@ sidebar_label: Keymaps
 ---
 
 import KeymapExample from "../keymap-example.md";
-import KeymapExampleFile from "../keymap-example-file.md";
 
 ZMK uses a declarative approach to keymaps instead of using C code for all keymap configuration.
 Right now, ZMK uses the devicetree syntax to declare those keymaps; future work is envisioned for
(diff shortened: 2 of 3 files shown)


docs/docs/features/keymaps.mdx
```diff
@@ -38,21 +37,18 @@ For the full set of possible behaviors, see the [overview page for behaviors](..
 ## Layers
 
 Like many mechanical keyboard firmwares, ZMK keymaps are composed of a collection of layers, with a
-minimum of at least one layer that is the default, usually near the bottom of the "stack". Each layer
+minimum of at least one layer that is the default, usually near the bottom of the "layer stack". Each layer
 in ZMK contains a set of bindings that bind a certain behavior to a certain key position in that layer.
 
 |                                                   ![Diagram of three layers](../assets/features/keymaps/layer-diagram.png)                                                    |
```
caksoylar, line 43: This should really be replaced as well, at least with one that doesn't use the `KC_` prefix. But maybe for another PR.

docs/docs/features/keymaps.mdx
```diff
@@ -4,7 +4,6 @@ sidebar_label: Keymaps
 ---
 
 import KeymapExample from "../keymap-example.md";
-import KeymapExampleFile from "../keymap-example-file.md";
```
caksoylar, line 7: It looks like we can delete this file now.

docs/docs/features/keymaps.mdx
```diff
@@ -121,18 +117,49 @@ Nested under the devicetree root, is the keymap node. The node _name_ itself is
...
+        default_layer { // Layer 0
+// ----------------------------------------------
+// |     Z     |     M     |     K     |
+// |     A     |     B     |     C     |
+            bindings = <
+                &kp Z    &kp M    &kp K
+                &kp A    &kp B    &kp C
+            >;
+        };
+    }
```
caksoylar, line 135:
```suggestion
    };
```

docs/docs/keymap-example.md
```diff
@@ -2,20 +2,34 @@
...
+// ----------------------------------------------
+// |        X       |      Y       |      Z     |
+// |   LEFT CTRL    |  LEFT ALT    |   &trans   |
+            bindings = <
+                &kp X        &kp Y       &kp Z
+                &kp LCTRL    &kp LALT    &trans
+            >;
+        };
+
+
```
caksoylar, line 33:
```suggestion
```

docs/docs/behaviors/layers.md
```diff
@@ -8,6 +8,10 @@ sidebar_label: Layers
 Often, you may want a certain key position to alter which layers are enabled, change the default layer, etc.
 Some of those behaviors are still in the works; the ones that are working now are documented here.
 
+:::note
+Activating a layer will not disable layers higher up in the "layer stack". See [Layers](../features/keymaps.mdx#layers) for more information.
```
caksoylar, line 12:
```suggestion
Multiple layers can be active at the same time and activating a layer will not deactivate layers higher up in the "layer stack". See [Layers](../features/keymaps.mdx#layers) for more information.
```

docs/docs/features/keymaps.mdx
```diff
@@ -38,21 +37,18 @@ For the full set of possible behaviors, see the [overview page for behaviors](..
...
-In addition to the base default layer (which can be changed), certain bound behaviors may also
-enable/disable additional layers "on top" of the default layer.
+All layers are assigned and referred to by a natural number, with the base layer being layer `0`. It is common to [use the C preprocessor to "name" layers](../behaviors/layers.md#defines-to-refer-to-layers), making them more legible.
 
-When a key location is pressed/released, the stack of all active layers from "top to bottom" is used,
-and the event is sent to the behavior bound at that position in each layer, for it to perform whatever
-actions it wants to in reaction to the event. Those behaviors can choose to "handle" the event, and stop
-it from being passed to any lower layers, or may choose to "pass it along", and let the next layer
-in the stack _also_ get the event.
+The default layer (usually the base layer) is always enabled. Certain bound behaviors may enable/disable additional layers.
```
caksoylar, line 49:
```suggestion
The default layer (the base layer with index 0) is always enabled. Certain bound behaviors may enable/disable additional layers.
```
I think this can be changed in the future if any changes regarding default layers are made.

docs/docs/features/keymaps.mdx
```diff
@@ -121,18 +117,49 @@ Nested under the devicetree root, is the keymap node. The node _name_ itself is
...
+### Multiple Layers
+
+Layers are numbered in the order that they appear in keymap node - the first layer is `0`, the second layer is `1`, etc.
+
+Here is an example of a trio of layers for a simple 6-key macropad:
+
+<KeymapExample />
+
+:::note
+Even if layer `1` was to be activated after `2`, layer `2` would still have priority as it is higher valued. Behaviors such as [`&to`](../behaviors/layers.md#to-layer) can be used to enable one layer _and disable all other non-default layers_, though.
```
caksoylar, line 152:
```suggestion
Even if layer `1` was to be activated after `2`, layer `2` would still have priority as it is higher valued. Behaviors such as [To Layer (`&to`)](../behaviors/layers.md#to-layer) can be used to enable one layer _and disable all other non-default layers_, though.
```
I am trying to be aware of future Studio users, where it isn't planned to surface the binding labels to the users at all. (We probably will need a lot more changes for that!)

docs/docs/keymap-example.md
```diff
@@ -2,20 +2,34 @@
...
         compatible = "zmk,keymap";
 
-        default_layer {
-// --------------------------------------------------------------------------------------------------------------------------------------------------------------------
-// |   ESC   |    Q    |    W    |    E    |    R    |    T    |                                          |    Y    |    U    |    I    |    O    |    P    |    \    |
-// |   TAB   |    A    |    S    |    D    |    F    |    G    |                                          |    H    |    J    |    K    |    L    |    ;    |    '    |
-// |  SHIFT  |    Z    |    X    |    C    |    V    |    B    | CTRL+A  | CTRL+C  |  |  CTRL+V |  CTRL+X |    N    |    M    |    ,    |    .    |    /    |  R CTRL |
-//                               |   GUI   |   DEL   | RETURN  |  SPACE  | ESCAPE  |  |  RETURN |  SPACE  |   TAB   |   BSPC  |  R ALT  |
+        default_layer { // Layer 0
+// ----------------------------------------------
```
caksoylar, line 6: Do you think it's worth getting rid of these comment blocks? They don't seem to add much in this example, at least.

docs/docs/features/keymaps.mdx
```diff
@@ -38,21 +37,18 @@ For the full set of possible behaviors, see the [overview page for behaviors](..
...
 
-When a key location is pressed/released, the stack of all active layers from "top to bottom" is used,
-and the event is sent to the behavior bound at that position in each layer, for it to perform whatever
-actions it wants to in reaction to the event. Those behaviors can choose to "handle" the event, and stop
-it from being passed to any lower layers, or may choose to "pass it along", and let the next layer
-in the stack _also_ get the event.
+The default layer (usually the base layer) is always enabled. Certain bound behaviors may enable/disable additional layers.
+
+When a key location is pressed/released, the _highest-valued currently active_ layer is used. The press/release event is sent to the behavior bound at that position in said layer, for it to perform whatever actions it wants to in reaction to the event. The behavior can choose to "consume" the event, or "pass it along" and let the next highest-valued active layer _also_ get the event (whose behavior may continue "passing it along").
```
caksoylar, line 52: I think it would be worth adding a "tip" box here, talking about multiple base layouts. That seems to be the most common scenario where people have issues.

i.e. if you have toggled layers like qwerty vs. colemak, you should always place them first in the stack (defined earlier) so that other layers that can be activated (momentarily) through them work as expected.

This might be worth a more elaborate section such as the one mentioned in #1953, but for now a short tip block would still be helpful. If we add that, then we can simply link there from here.

docs/docs/features/keymaps.mdx
```diff
@@ -38,21 +37,18 @@ For the full set of possible behaviors, see the [overview page for behaviors](..
...
```
caksoylar, line 51:
```suggestion
When a key location is pressed/released, the _highest-valued currently active_ layer is used. The press/release event is sent to the behavior bound at that position in said layer, for it to perform whatever actions it wants to in reaction to the event. The behavior can choose to "consume" the event, or "pass it along" and let the next highest-valued active layer _also_ get the event (whose behavior may continue "passing it along").

Note that the _activation_ order isn't relevant for determining the priority of active layers, it is determined _only_ by the definition order.
```
This remark usually helps when trying to explain to people, so I think we should insert it somewhere like this.

docs/docs/features/keymaps.mdx
Nick-Munnich: The file is still in use in the new shield guide. I thought about moving the file, or putting the content into the new shield guide. What are your thoughts?

docs/docs/features/keymaps.mdx
Nick-Munnich: I am not artistically talented, so someone else will need to sort that out I think.

docs/docs/keymap-example.md
Nick-Munnich: I think the comment blocks are used frequently enough in other keymaps that it's useful to have them so that new users get used to them, in a way.

docs/docs/features/keymaps.mdx
caksoylar: Ah, good catch thanks. I'd say let's leave it for now.

docs/docs/features/keymaps.mdx
caksoylar: Yeah, we could ask Thomas@splitkb to share the source (he was the original author) or reproduce it otherwise.

docs/docs/features/keymaps.mdx
Nick-Munnich: I agree on both points, added tip block for now. When #1953 is done we can revisit.

docs/docs/features/keymaps.mdx
```diff
@@ -38,21 +37,24 @@ For the full set of possible behaviors, see the [overview page for behaviors](..
...
-in the stack _also_ get the event.
+The default layer (the base layer with index 0) is always enabled. Certain bound behaviors may enable/disable additional layers.
+
+When a key location is pressed/released, the _highest-valued currently active_ layer is used. The press/release event is sent to the behavior bound at that position in said layer, for it to perform whatever actions it wants to in reaction to the event. The behavior can choose to "consume" the event, or "pass it along" and let the next highest-valued active layer _also_ get the event (whose behavior may continue "passing it along").
+
+Note that the _activation_ order isn't relevant for determining the priority of active layers, it is determined _only_ by the definition order.
+
+:::tip
+If you wish to use multiple base layers (with a [toggle](../behaviors/layers.md#toggle-layer)), e.g. one for qwerty and another for colemak, you will want these layers to have the lowest value possible. In other words, one should be layer `0`, and the other should be layer `1`. This allows other momentary layers activated on top of them to work with both.
```
caksoylar, line 56:
```suggestion
If you wish to use multiple base layers (with a [toggle](../behaviors/layers.md#toggle-layer)), e.g. one for QWERTY and another for Colemak layouts, you will want these layers to have the lowest value possible. In other words, one should be layer `0`, and the other should be layer `1`. This allows other momentary layers activated on top of them to work with both.
```

caksoylar (approved):

caksoylar:
I merged this to hopefully clarify #1284. As far as I can tell the specific concerns in there were addressed. Any follow ups are welcome to further improve any unclear points.

PR #2392, opened 2024-07-26
fix(docs): Fix toggle layer docs

caksoylar:
The example here was malformed and a bit weird with the extra `#define` etc. I thought I'd fix that, then decided that the example wasn't actually useful here and it was inconsistent with the rest of the page.

Also included a couple tweaks to other behavior descriptions to make them more technically correct.

### DIFF
--- a/docs/docs/behaviors/layers.md
+++ b/docs/docs/behaviors/layers.md
@@ -33,7 +33,7 @@ again.
 ### Behavior Binding
 
 - Reference: `&mo`
-- Parameter: The layer number to enable/disable, e.g. `1`
+- Parameter: The layer number to enable while held, e.g. `1`
 
 Example:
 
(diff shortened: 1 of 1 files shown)


docs/docs/behaviors/layers.md
```diff
@@ -99,7 +99,7 @@ Example:
 
 ## Toggle Layer
 
-The "toggle layer" behavior enables a layer until the layer is manually disabled.
+The "toggle layer" behavior enables a layer if it is currently disabled, or disables it if enabled.
```
Nick-Munnich, line 102: Do you know what happens if you `&tog` off a layer that is active due to an `&mo`? This reads to me as though it would disable the layer, while the original reads as if it would do nothing (though it is much less clear in that version)

caksoylar: I am 90% sure that it will deactivate the momentary layer, since they all use the same on/off functions. But it might be nice to double check, I'll report later.

caksoylar: I confirmed that tapping `&tog X` turns off `&mo X` even while the latter is held, so that's consistent with my understanding.

Nick-Munnich (approved):

PR #2396, opened 2024-07-31
feat(modularisation/docs) New hardware metadata version

### LINKED ISSUE #453: Towards a modular ZMK
Fellow developers, thank you for your patience.  The modular proof-of-concept (PoC) is ready for evaluation.

# Purpose

This is a (fully functional) **proof of concept**.  Its purpose is to:
- enable everyone to explore what a modular ZMK would entail.
- act as a point of reference for discussions about modularity.
- be the basis for properly evaluating a modular ZMK, including:
  - technical design
  - end-user experiences
  - developer workflows and experiences
  - maintenance
  - best practices
- be the basis for moving forward if we choose to go down this road (it's a high fidelity prototype).

# What's "a modular ZMK"?

- Modules:
  - can contain boards, shields, C (code), Kconfig, dts, soc, or arch.
  - can live anywhere in their own git repository (web or filesystems).
  - can be owned, licensed and/or maintained by anyone (distributed development).
  - are facilitated by Zephyr's `west` tool.
- We distill the ZMK repository down to core components only.
  - Non-core components (boards/shields,etc.) can be moved into external (optional) modules.
- We supersede `zmk-config` with a simple user "application".
  - This "application" module/repository is owned by the end-user.
  - It's specific to their own build.
  - It sits in the driver's seat.
    - It describes:
      - Which modules their application/build depends on (including boards, shields or extra features).
      - Any configuration or code specific to their build.
    - It's square one for the build process.
  - It can be as simple or complex as the user desires.
- ZMK effectively becomes a core module/library/SDK.

# Motivations

I'm suggesting we consider a modular approach for ZMK because of several pressing issues, including:
- Board and Shields
  - As ZMK grows in popularity we're seeing the number of board and shield PRs increase.
  - We're facing pressure around which contributions we should accept into the core.
  - Any time core contributors spend on boards/shields is time taken from core development, slowing the project down.
  - The more boards and shields we accept, the more bloated ZMK becomes and the more maintenance baggage we take on.
  - The more we take on, the more impractical or unaffordable it will become to test them as part of ZMK's CI.
  - Our current approach is great for popular boards/shields, but does little for the smaller players.  We risk artificially limiting discoverability.
  - We maintain pretty high standards for PRs, which isn't everyone's cup of tea.
  - What happens when contributions become abandoned?  How do we deprecate them?
- Distributed Extensions
  - Currently there are only these options for adding functionality:
    - Add it to your `zmk-config`.
    - Create a branch/fork of ZMK.
    - Submit a PR to add it to the core (which is an understandably slow process).
  - These approaches get the job done, but none of them are particularly great for sharing with others.
  - It's not easy for developers and testers who often live on the cutting edge from multiple sources.
- Flexibility
  - Modularity would let us explore and adopt a range of approaches for the future, such as:
    - Hardware developer owned & maintained modules (boards/shields).
    - Author owned & maintained modules.
    - Fully distributed boards/shields (each in their own module).
    - Community maintained modules/repositories.
  - Users could leverage other Zephyr modules such as the [Nordic nRF Connect SDK](https://github.com/nrfconnect/sdk-nrf) without licensing issues for ZMK itself.
  - Users could have more control over the build.

# First steps
- I've built a demo which uses an in-development version of the `tidbit` shield.  Thanks let us use it Kemonine!
- The shield itself lives in its own [repository](https://github.com/innovaker2/zmk-tidbit)/module.

## Local (developers with existing environment setup)
This includes developers using Windows, macOS, Linux or WSL2.
```
west init --mr master-v1 -m https://github.com/innovaker/my-zmk-keyboard zmk-workspace
cd zmk-workspace/my-zmk-keyboard
west update
west build
```

## GitHub Actions (cloud)
- Fork https://github.com/innovaker/my-zmk-keyboard
- Enable the workflow (this only because using a fork for the demo)
  - Go to the `Actions` tab.
  - Click `I understand my workflows, go ahead and enable them`.
  - Click `Build` (on the left).
  - Click `Enable workflow`.
- Run the workflow
  - Go to the `Actions` tab.
  - Click `Build` (on the left).
  - Click `Run workflow` (on the right) and run it on Branch: `master-v1`
  - Wait 10 seconds for the Build job to appear.
- Make a cup of tea whilst GitHub actions builds firmware for the `zmk-tidbit` (< 4 minutes).
- Checkout, edit/commit the files and push.
  - You probably only need to edit the comma delimited `matrix` in `.github/workflows/build.yml`
- Make another cup of tea whilst it rebuilds.

## Docker

If you want to try developing in Docker containers ...
- Install [Docker for Desktop](https://www.docker.com/products/docker-desktop)
- I recommend turning on the [WSL2 backend](https://docs.docker.com/docker-for-windows/wsl/) in the options.

### zmk-docker

As part of this work, I've developed a brand new container generation project: https://github.com/innovaker/zmk-docker
which I hope to be adopted by https://github.com/zmkfirmware

Both the GitHub CI (workflows) and VS Code Remote Containers (Dev Containers) leverage the images generated by the zmk-docker project.

### VS Code Remote Containers

Due to current shortcomings in VS Code, VS Code Remote Containers, Codespaces, Docker and the Zephyr build system, it's proving tricky to find an intuitive one-size-fits-all approach for containers.  However, their potential benefits make them worthy of further exploration.  Here's a guide to the known approaches at time of writing ...

#### Cloning _straight_ from a repository into a container with VS Code

This approach is my personal preference.  It's quick to setup, performant and I don't mind my files being inside volumes.   It lets you clone a repository straight into a Docker container/volumes without touching the local system or needing any tools besides VS Code.  I also use the WSL2 backend which makes it really fast.

- Open VS Code.  Make sure you have the `Remote - Containers` extension installed.
- If you press `F1` and enter: `> Remote-Containers: Clone`
  it'll give you a cloning option.
  Repository: https://github.com/innovaker/my-zmk-keyboard
- You can use either cloning options:
  - `Create a unique volume`
  - `Create a new volume ...`
     - Enter the volume name: `zmk-modular-v1` (it can be anything you like)
     - Enter the target folder name: `zmk-workspace/my-zmk-keyboard`
       - note the slash: it's a trick you can use to have multiple ZMK workspaces in a single container.
- After launching the container:
  - `west update` will run as part of the container setup - it will take a while the first time.
  - When the prompt pops up, click `Open Workspace`.
    - This will let you see the `zmk` folder and give you access to the other VS Code workspace benefits.
  - Open a terminal and run: `west build`

#### Cloning to a local disk first, then opening it in VS Code, then reopening it in a container

Some people prefer this method as it makes more sense to them.  But it's not as performant.

- Locally (in your OS): `git clone -b master-v1 https://github.com/innovaker/my-zmk-keyboard`
- Open the `my-zmk-keyboard` folder in VS Code.
- When the prompt pops up, click `Reopen in Container`.
- `west update` will run as part of the container setup - it will take a while the first time.
- When the prompt pops up, click `Open Workspace`.
    - This will let you see the `zmk` folder and give you access to the other VS Code workspace benefits.
- Open a terminal for `app` and run: `west build`

Note, I've configured it to mount `zephyr` and its dependencies into (shared) Docker volumes so that it doesn't take an eternity to run `west update`.  These volumes are shared across containers and not junked when you delete a container.  `zmk` is currently stored within the container itself (so it will be lost/re-cloned if you rebuild the container) but it too can be mounted as a volume by a savvy user (I've left it commented out).

### Codespaces + nested branch

I've not personally tested it on Codespaces yet but my cursory understanding is that it's built around the mono-repository paradigm and doesn't support untracked flat structure workspaces???  In anticipation of that, I've' been exploring a nested version of `zmk-app`/`my-zmk-keyboard`.  I'm not keen on it personally but I made it as an experiment.

The branch names are `nested-v1` if you want to try it.

# Next steps
- Edit: `CMakeList.txt`
- Drop in: your own overlay/keymap?
- Rebuild with: `west build -p` (pristine build)
- Explore: `west.yml`
- Try: making your own module?

# Going forward

- Whilst exciting, modular development is a complex topic which merits careful consideration and debate.
- It would entail a major paradigm shift to our current approach.  We shouldn't get carried away and rush into it.
- We should use this PoC to explore the pros and cons together.
- I encourage you to actively try developing with it, particularly to get a feel for its effects on workflow.
- We need to find the pain points and potential blockers.

Nick-Munnich:
This is an initial rough draft for a version 2 of the ZMK hardware metadata standard. The goal is to define the metadata such that the metadata files can be collected in a separate repository titled `zmk-modules`. [**This draft has not updated the metadata schema, but rather the documentation of it, to be easier to read and encourage discussion.**](https://deploy-preview-2396--zmk.netlify.app/docs/development/hardware-metadata-files) Also note that as the schema has been previously changed without updating the documentation, there are some properties to which I was uncertain of their use - I gave it my best guess.

The assumption is that (almost) all boards and shields currently in the ZMK tree will be moved to external module repositories, all interconnect definitions will be moved to `zmk-modules`, and only the boards and shields used for testing purposes remain in-tree.

ZMK's setup scripts and related docs would need to be adjusted accordingly, either also in this PR or in a separate one. An initial rough draft `zmk-modules` can be found [here](https://github.com/Nick-Munnich/zmk-modules), focusing on folder structure.

As this would obviously break compatibility with a lot of people's setups, I would suggest doing it simultaneously with whatever changes are necessary to move to Zephyr 3.7 LTS. When everything is completed, then I believe we can (finally) close #453. 

We could also include some optional properties that would allow for some cool tools to make use of zmk-modules. In particular, the data to enable tools [comparing keyboards to one another](https://yal-tools.github.io/ergo-keyboards/) might be very beneficial, as currently the developers of such tools have to input each new keyboard manually themselves.

TODO:
 - [ ] Agree on metadata v2
 - [ ] Create `zmk-modules` repository
 - [ ] Adjust ZMK's setup scripts to use `zmk-modules`
 - [ ] Document `zmk-modules` repository
 - [ ] Fill `zmk-modules` with existing modules
 - [ ] Move non-essential boards and shields from ZMK tree into modules

EDIT: Thanks to discussion, it seems that we also need:
 - [ ] Agree and implement a versioning system for ZMK


ssbb:
It would be cool to add some physical layout definitions for keyboards. For example, [QMK uses](https://github.com/qmk/qmk_firmware/blob/master/docs/reference_info_json.md#layouts-layouts) a KLE-like JSON format.

We could add some common presets like ANSI60, ANSI104, ISO60, ISO105, etc. Then, we might allow the xxx+x approach for ergonomic boards (e.g., 23332+2 is a Hummingbird layout). Additionally, we could use a KLE-like JSON for custom layouts.

Usage examples:

- Projects such as a keymap-drawer could benefit significantly from this (currently, you have to specify it manually).
- Projects like YellowAfterlife's ergo keyboard list could parse this information to determine the number of columns, rows, and thumbs.
- Some CLI tools could use it to render the proper layout, etc.

petejohanson:
> It would be cool to add some physical layout definitions for keyboards. For example, [QMK uses](https://github.com/qmk/qmk_firmware/blob/master/docs/reference_info_json.md#layouts-layouts) a KLE-like JSON format.
> 
> We could add some common presets like ANSI60, ANSI104, ISO60, ISO105, etc. Then, we might allow the xxx+x approach for ergonomic boards (e.g., 23332+2 is a Hummingbird layout). Additionally, we could use a KLE-like JSON for custom layouts.
> 
> Usage examples:
> 
> * Projects such as a keymap-drawer could benefit significantly from this (currently, you have to specify it manually).
> * Projects like YellowAfterlife's ergo keyboard list could parse this information to determine the number of columns, rows, and thumbs.
> * Some CLI tools could use it to render the proper layout, etc.

That information is already *now* possible to set in the devicetree with the new physics layouts refactor so it's available in firmware for the upcoming ZMK Studio work. It would be a net negative to duplicate that in the metadata, IMHO.

caksoylar:
I think the broad strokes of this organization makes sense. A separate repo where PRs can go into to add "references" from which tooling can discover things is a sensible approach.

One small suggestion I'd have for now would be to have `module` dict field instead of `murl`, where `url`, `commit`, `outdated` (also maybe project `name`, to be used in `west.yml`) can go as sub-entries.

I think there are a few things that warrant further discussion. I am not listing these because I necessarily have an opinion on them but discussion around them would be good to have.

1. Pinning modules to commits: The proposal mandates it, but there might be advantages to be able to point to a branch/tag.
2. How to handle compatibility with ZMK versions (or even between modules): The proposal has an `outdated` field which might be useful, but also is not a well-defined concept[^1]. This is a hard question to answer in general, and having ZMK use some sort of specific versioning scheme might be a prerequisite to a proper solution.
3. Maybe it is better to omit a `pull-request` field to limit "support" (since these will be used by official tooling) to the main branch/release.

[^1]: Do ZMK committers update these as things break? An automated tool can check for building but can't check for changed/broken functionality

Nick-Munnich:
> One small suggestion I'd have for now would be to have `module` dict field instead of `murl`, where `url`, `commit`, `outdated` (also maybe project `name`, to be used in `west.yml`) can go as sub-entries.

Great idea, I'll make this change.

>     1. Pinning modules to commits: The proposal mandates it, but there might be advantages to be able to point to a branch/tag.

The reason for this is specifically a security/malicious actor concern. By specifying a commit, we can prevent harmful images/code/etc from being present in the version of the module that we recommend (unless it is missed on review). I think SHA-1 should suffice here, though cybersecurity/cryptography isn't my specialty.

>     2. How to handle compatibility with ZMK versions (or even between modules): The proposal has an `outdated` field which might be useful, but also is not a well-defined concept[1](#user-content-fn-1-14bdca0f8be552f5c0b9ba62bc8be843). This is a hard question to answer in general, and having ZMK use some sort of specific versioning scheme might be a prerequisite to a proper solution.

I think I'll change it so that the `outdated` field points to the most recently working `zmk` commit and is omitted when up-to-date - then builds with outdated modules could still be attempted, while raising big warnings. A proper versioning scheme would be very nice, but I don't think it's necessarily necessary.

>     3. Maybe it is better to omit a `pull-request` field to limit "support" (since these will be used by official tooling) to the main branch/release.

I was quite torn on whether to include or remove it. I'll remove it.

> 
> ## Footnotes
> 
>     1. Do ZMK committers update these as things break? An automated tool can check for building but can't check for changed/broken functionality [↩](#user-content-fnref-1-14bdca0f8be552f5c0b9ba62bc8be843)

My thinking was that we could write a tool that goes through all of the modules in a `type` category, marks them as outdated, and opens issues on all the remote repos found. The owners/maintainers of the module could comment on the issue to discard it if it doesn't apply to them, causing the tool to automatically remove the outdated flag, or they can apply changes and then open a PR to update the pinned commit prompting manual review. It would be ZMK committers' responsibility to write the issue description if their commit breaks something, and ZMK maintainers' responsibility to run the tool on a breaking merge.

Ideally we could have it auto-accept particular changes (ex. adding `wakeup-source`), but that seems like a significantly more complex task and could be investigated far into the future. A proper versioning scheme as you mentioned would also help it feel smoother.

urob:
I have been thinking a bit about maintenance costs. Zephyr has a [policy](https://docs.zephyrproject.org/latest/develop/modules.html#requirements-for-allowed-practices) in place which only allows changes to modules as mergeable PRs and strictly forbids force-pushes to the main branch. Something similar could help reduce the review burden for the ZMK team.

On the flipside, to reduce maintenance costs for module maintainers, I'd second that some form of ZMK version system (with infrequent api-breaking changes) would be desirable. Even if not strictly necessary, forcing maintainers to check compatibility for *every* ZMK commit is a huge burden. 

Moreover, if modules aren't synchronized, which seems inevitable in a large ecosystem without proper upstream versioning, then the burden ultimately gets passed on to end-users: To update their firmware, they will need to:

1. Find the latest ZMK commit that's supported by all modules they are using
2. For each of their modules, find the corresponding commit that supports this largest-commonly-supported commit

Nick-Munnich:
Something similar would be an excellent policy to have in `zmk-modules`. 

You raise a good point and argument for versioning as well. If we introduced a versioning system, then the `commit` and `outdated` child properties of `module` could be cleanly replaced by a `version` property:
```yaml
module:
  url: https://github.com/module-owner/module-repo-name
  version:
    - 2-1-3: 093dj09sdapasdlkd09285c585e08pjonkjn7y8798
    - 1-0-12: c43a0365183c58f74c285c585e0cbc1ea99f8a12
``` 
That would allow tools like [zmk-cli](https://github.com/joelspadin/zmk-cli/tree/main) to completely remove the burden you described.

joelspadin:
If I'm understanding this correctly, with this proposal, in order to add or update a module, you would need to manually copy all the .zmk.yml files from the module into the central repository, then update each one to include the module repo's URL and the specific commit from which the data was pulled. Failing to do this correctly could lead to the data in the central repo not matching that of the module repo.

It seems to me like it would be a lot easier to maintain if the central repo contained only module URLs and commit hashes. Then to update a module, you would only need to update the commit hash in one file. This would mean you could not search the central repo directly for specific keyboards/behaviors/etc., but it would be possible to set up a website which is generated by collecting data from every module into a searchable database.

YellowAfterlife:
On keyboard geometry, that's such a topic - e.g. with Xx+Y notation, how do you describe something like Hillside? Even my current "key count, columns, rows, and extra keys" format doesn't _quite_ work because an extra index finger column and an extra pinky column serve different purposes.

Nick-Munnich:
> On keyboard geometry, that's such a topic - e.g. with Xx+Y notation, how do you describe something like Hillside? Even my current "key count, columns, rows, and extra keys" format doesn't _quite_ work because an extra index finger column and an extra pinky column serve different purposes.

The docs from #2268 found [here](https://deploy-preview-2268--zmk.netlify.app/docs/development/studio-setup) describe the physical layout system nicely. The current plan is to add tags to each key such as "PINKY_HOME" at some point in the future.

Nick-Munnich:
Closing this for the time being. While I do still feel like an update to the metadata would be nice, I no longer feel that it is necessary for the module ecosystem to mature. It would be better to revisit this topic at a much later stage and see what would be useful.

PR #2397, opened 2024-07-31
feat: Split physical layout selection sync.

petejohanson:
* Ensure the split peripherals have the same selected physical layout on connection and change.


app/src/split/bluetooth/central.c
```diff
@@ -56,6 +57,7 @@ struct peripheral_slot {
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
     uint16_t update_hid_indicators;
 #endif // IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
+    uint16_t select_physical_layout_handle;
```
caksoylar, line 60: IIUC, perhaps calling this `selected_*` would be more appropriate?

petejohanson: Fair. Updated.

PR #2398, opened 2024-08-02
feat(power) Add additional idle options

Nick-Munnich:
Added two Kconfig flags:

- `ZMK_EXT_POWER_IDLE_OFF` (disabled by default) to turn off external power when the keyboard goes into idle, and enable it again when it exits idle. Seems like it would be particularly useful for trackpoint users.
- `ZMK_IDLE_USB` (enabled by default) which enables/disables idling while connected to USB power. I'd be expecting RGB lovers to request this.

First flag is tested and confirmed working. Unable to test second flag, implementation is trivial though. Unable to test whether the display re-init bug mentioned in #1300 is still present, would expect this to interact poorly if so.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -163,8 +163,7 @@ config ZMK_BLE_EXPERIMENTAL_CONN
     bool "Experimental BLE connection changes"
     help
       Enables a combination of settings that are planned to be default in future versions of ZMK
-      to improve connection stability. This includes changes to timing on BLE pairing initiation,
-      restores use of the updated/new LLCP implementation, and disables 2M PHY support.
+      to improve connection stability. 
 
 config ZMK_BLE_EXPERIMENTAL_SEC
     bool "Experimental BLE security changes"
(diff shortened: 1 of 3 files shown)


caksoylar:
I think the combination of the two settings is overlapping with `CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE`. I think it could be nice to make the handling more structured by directly tying the behavior to power states, then configure the state's general behavior as with proposed settings. However it does reduce flexibility when you have multiple peripherals and want different behavior.

Nick-Munnich:
There is some ideological overlap, but currently no functional overlap: `CONFIG_ZMK_RGB_UNDERGLOW_AUTO_OFF_IDLE` merely turns brightness down to 0. There is use case for both currently - for example if the peripheral of a split is powered from the central and doesn't have its own battery, with a bit of underglow added on.

How this would look in the future when we have proper PM support I'm less certain about.

I'm not sure I follow the rest of your statement.

Nick-Munnich:
Going to close this for the time being - on further thinking, I think I'd rather revisit this once #1775 is merged and see if it's still a good thing to have then.

Issue #2388, opened 2024-07-23
Bug: keypresses not working after commit 0438cb0


Scalee:
After updating to this commit my keyboard stops working. (or just the latest commit, Just narrowed it down to this commit)

The keyboard will still connect and output logs but no keystrokes.

If I change chosen to &kscan1, the gpio-matrix will start working. If I switch to &kscan2 the charlieplex will start working.
Back to kscan_composite, nothing. Rolling back one commit and recompiling everything works as expected.

Was asked to create a new issue and link to #2376 since they might be related.

```
#include <dt-bindings/zmk/matrix_transform.h>

/ {
    chosen {
        zmk,kscan = &kscan0;
        zmk,matrix-transform = &comp_transform;
    };

    comp_transform: keymap_transform {
        compatible = "zmk,matrix-transform";
        
        rows = <12>;
        columns = <7>;
        
        map = <
            //Left side
            RC(0,0) RC(0,1) RC(0,2) RC(0,3) RC(0,4) RC(0,5) RC(0,6)        
            RC(1,0) RC(1,1) RC(1,2) RC(1,3) RC(1,4) RC(1,5) RC(1,6) 
            RC(2,0) RC(2,1) RC(2,2) RC(2,3) RC(2,4) RC(2,5) RC(2,6)       
            RC(3,0) RC(3,1) RC(3,2) RC(3,3) RC(3,4) RC(3,5) RC(3,6)        
            RC(4,0) RC(4,1) RC(4,2) RC(4,3) RC(4,4) RC(4,5) RC(4,6) 
            RC(5,0) RC(5,1) RC(5,2) RC(5,3) RC(5,4) RC(5,5) RC(5,6)
            
            //Right side
            RC(6,1) RC(7,0) RC(8,0) RC(9,0) RC(10,0) RC(11,0) RC(12,0)
            RC(6,2) RC(7,2) RC(8,1) RC(9,1) RC(10,1) RC(11,1) RC(12,1)
            RC(6,3) RC(7,3) RC(8,3) RC(9,2) RC(10,2) RC(11,2) RC(12,2)
            RC(6,4) RC(7,4) RC(8,4) RC(9,4) RC(10,3) RC(11,3) RC(12,3)
            RC(6,5) RC(7,5) RC(8,5) RC(9,5) RC(10,5) RC(11,4) RC(12,4)
            RC(6,6) RC(7,6) RC(8,6) RC(9,6) RC(10,6) RC(11,6) RC(12,5)
[...]

PR #2402, opened 2024-08-03
fix: Proper device PM support for composite kscan.

petejohanson:
* Clean up composite kscan to allow multiple instances properly.
* Implement PM hook and properly suspend/resume the child devices.

Fixes: #2388

Note: I don't have my Ferris on hand to properly test this, so hoping Scalee on #2388 can test.


Scalee:
This solved issue #2388

PR #2407, opened 2024-08-05
fix: Defer auto enabling of kscan sideband.

petejohanson:
* Now that device init of kscan sideband is in POST_KERNEL stage, use a separate SYS_INIT for auto enabling the device so processing of early/initial presses from referenced toggle mode kscan devices occurs at the proper time during init.

/cc @ReFil


ReFil:
Given this a test on a board with a toggle switch and direct kscan driver with toggle-mode and all seems to be working as expected

PR #2409, opened 2024-08-07
fix(split): Make locality work for nested behaviors

caksoylar:
This is another attempt to solve #1494. My understanding is that #1630 only covers the case of macros invoking global behaviors. I incorporated the changes in it and attempted to address the existing comments.

In addition to the other PR, this should cover sticky keys, hold-taps, mod-morphs and work with source+global locality behaviors. I tested the reset behaviors with hold-taps and tap dances so far.

I am not familiar with these parts of the code; so far I stuck an extra field `source` wherever I saw a `position` and passed it around. Specifically, it is a new field in `zmk_behavior_binding_event`. Any recommendations are welcome on how better to do it.

TODO:
- [x] Test source local behaviors (reset)
  - [x] macros
  - [x] mod-morphs
  - [x] hold-taps
  - [x] tap dances
  - [x] sticky keys
- [x] Test global behaviors (backlight)
  - [x] macros
  - [x] mod-morphs
  - [x] hold-taps
  - [x] tap dances
  - [x] sticky keys
- [x] Handle combos (at least get them working with global behaviors)

### DIFF
--- a/app/include/zmk/behavior.h
+++ b/app/include/zmk/behavior.h
@@ -26,6 +26,7 @@ struct zmk_behavior_binding_event {
     int layer;
     uint32_t position;
     int64_t timestamp;
+    uint8_t source;
 };
 
 /**
@@ -41,6 +42,19 @@ struct zmk_behavior_binding_event {
  * unrelated node which shares the same name as a behavior.
  */
 const struct device *zmk_behavior_get_binding(const char *name);
+
+/**
+ * @brief Invoke a behavior given its binding and invoking event details.
+ *
+ * @param src_binding Behavior binding to invoke.
+ * @param event The binding event struct containing details of the event that invoked it.
+ * @param pressed Whether the binding is pressed or released.
+ *
+ * @retval 0 If successful.
+ * @retval Negative errno code if failure.
+ */
+int zmk_behavior_invoke_binding(const struct zmk_behavior_binding *src_binding,
+                                struct zmk_behavior_binding_event event, bool pressed);
 
 /**
  * @brief Get a local ID for a behavior from its @p name field.
--- a/app/include/zmk/behavior.h
+++ b/app/include/zmk/behavior.h
@@ -26,7 +26,9 @@ struct zmk_behavior_binding_event {
     int layer;
     uint32_t position;
     int64_t timestamp;
+#if IS_ENABLED(CONFIG_ZMK_SPLIT)
     uint8_t source;
+#endif
 };
 
 /**
--- a/app/include/zmk/behavior_queue.h
+++ b/app/include/zmk/behavior_queue.h
@@ -10,5 +10,5 @@
 #include <stdint.h>
 #include <zmk/behavior.h>
 
-int zmk_behavior_queue_add(uint32_t position, const struct zmk_behavior_binding behavior,
-                           bool press, uint32_t wait);
+int zmk_behavior_queue_add(uint32_t position, uint8_t source,
+                           const struct zmk_behavior_binding behavior, bool press, uint32_t wait);
--- a/app/include/zmk/behavior_queue.h
+++ b/app/include/zmk/behavior_queue.h
@@ -10,5 +10,5 @@
 #include <stdint.h>
 #include <zmk/behavior.h>
 
-int zmk_behavior_queue_add(uint32_t position, uint8_t source,
+int zmk_behavior_queue_add(const struct zmk_behavior_binding_event *event,
                            const struct zmk_behavior_binding behavior, bool press, uint32_t wait);
--- a/app/include/zmk/split/bluetooth/service.h
+++ b/app/include/zmk/split/bluetooth/service.h
@@ -20,6 +20,7 @@ struct sensor_event {
 
 struct zmk_split_run_behavior_data {
     uint8_t position;
+    uint8_t source;
     uint8_t state;
     uint32_t param1;
     uint32_t param2;
--- a/app/src/behavior.c
+++ b/app/src/behavior.c
@@ -95,7 +95,7 @@ int zmk_behavior_invoke_binding(const struct zmk_behavior_binding *src_binding,
     case BEHAVIOR_LOCALITY_CENTRAL:
         return invoke_locally(&binding, event, pressed);
     case BEHAVIOR_LOCALITY_EVENT_SOURCE:
-#if ZMK_BLE_IS_CENTRAL
+#if ZMK_BLE_IS_CENTRAL // source is a member of event because CONFIG_ZMK_SPLIT is enabled
         if (event.source == ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL) {
             return invoke_locally(&binding, event, pressed);
         } else {
--- a/app/src/behavior.c
+++ b/app/src/behavior.c
@@ -17,10 +17,17 @@
 
 #endif
 
+#include <zmk/ble.h>
+#if ZMK_BLE_IS_CENTRAL
+#include <zmk/split/bluetooth/central.h>
+#endif
+
 #include <drivers/behavior.h>
 #include <zmk/behavior.h>
 #include <zmk/hid.h>
 #include <zmk/matrix.h>
+
+#include <zmk/events/position_state_changed.h>
 
 #include <zephyr/logging/log.h>
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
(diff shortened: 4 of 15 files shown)


app/include/zmk/keymap.h
```diff
@@ -27,6 +28,9 @@ const char *zmk_keymap_layer_name(uint8_t layer);
 int zmk_keymap_position_state_changed(uint8_t source, uint32_t position, bool pressed,
                                       int64_t timestamp);
 
+int zmk_invoke_behavior_binding(struct zmk_behavior_binding *binding,
```
Nick-Munnich, line 31: Not delved deep enough into the code to know the stylings typically used, but would it make sense to name this `invoke_zmk_behavior_binding`?

caksoylar: My impression (which can also be wrong) is that exposed methods (declared in the headers) get the `zmk_` prefix.

caksoylar:
This is ready for review: I tested all the combinations I can. (And fixed one bug where I didn't learned the purpose of the struct copy the hard way.)

One thing missing is sensor rotate since I don't have hardware with encoders but it is using the behavior queue just like the macros, so I assume it should be fine. I'll ask for testers on Discord just in case.

Nick-Munnich:
Missing a docs adjustment to the new split keyboard's page

caksoylar:
@Nick-Munnich removed the note, thanks!

> One thing missing is sensor rotate since I don't have hardware with encoders but it is using the behavior queue just like the macros, so I assume it should be fine.

This was naive, I looked into it and passing through the source information in the sensor events is not trivial and I don't feel confident enough to do it given I can't even test them. In https://github.com/zmkfirmware/zmk/pull/2409/commits/4e12392d918fafcc90fdff6195198f7cd9028e0a I hardcoded them to trigger on the central; compared to status quo it should fix using global behaviors like `&rgb_ug`, at least. I think making source local ones work can be tackled in the future.

app/include/zmk/behavior.h
```diff
@@ -26,6 +26,7 @@ struct zmk_behavior_binding_event {
     int layer;
     uint32_t position;
     int64_t timestamp;
+    uint8_t source;
```
petejohanson, line 29: I do wonder of the addition for the `source` in a few places should be conditional on `CONFIG_ZMK_SPLIT`. On more resource limited devices, we may be storing things like the queue items, etc. in RAM, and the extra 4-bytes will be not-insignificant, especially if we're buffering 32 of them, etc.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -77,6 +77,7 @@ struct behavior_hold_tap_data {
 // this data is specific for each hold-tap
 struct active_hold_tap {
     int32_t position;
+    uint8_t source;
```
petejohanson, line 80: Ditto.

app/src/keymap.c
```diff
@@ -174,22 +174,29 @@ int invoke_locally(struct zmk_behavior_binding *binding, struct zmk_behavior_bin
...
+        .source = source,
     };
 
-    LOG_DBG("layer: %d position: %d, binding name: %s", layer, position, binding.behavior_dev);
+    LOG_DBG("layer: %d position: %d, binding name: %s", layer, position, binding->behavior_dev);
+
+    return zmk_invoke_behavior_binding(binding, event, pressed);
+}
+
+int zmk_invoke_behavior_binding(const struct zmk_behavior_binding *src_binding,
```
petejohanson, line 190: If we're going to start having external callers of this, I'm not *convinced* this should stay in keymap, instead of going somewhere like `app/src/behavior.c`.

petejohanson (commented):
A few comments after a first pass. Thanks!

app/src/keymap.c
caksoylar: Makes sense, I'll move this to behavior.c and the corresponding header.

app/include/zmk/behavior.h
caksoylar: Valid concern, let me see if I can `#ifdef` guard these and whether it will be very ugly.

app/src/keymap.c
caksoylar: I moved the function and the `invoke_locally` method, I think it is better indeed.

app/include/zmk/behavior.h
caksoylar: I refactored all new `source` uses to be guarded by the split symbol, tested on split and non-splits. 

In some cases I thought it was nicer to pass a reference to the `zmk_behavior_binding_event` struct rather than the `position` (+`state` for split). Can you have another look?

caksoylar:
I got [confirmation](https://discord.com/channels/719497620560543766/1278388553163870333/1278396322495463557) from a user on Discord that encoders also work with a global behavior now. To sum up, this is the current state after this PR, according to my testing and others'.

Working with global and source-local behaviors:
- Mod-morphs
- Hold-taps
- Macros
- Tap dances
- Sticky keys

Working with global behaviors:
- Combos[^1]
- Sensors/encoders[^2]

I am not planning to tackle above two for source-local in this PR (see below) so it is complete from my POV.

[^1]: Making source local work with combos requires deciding the locality of combos from the set of key positions that trigger it, some discussion [here](https://discord.com/channels/719497620560543766/719544230497878116/1270736287589994497)
[^2]: The code change required for this is separate and since I don't have the hardware to iterate on, I think it'd be better to tackle this separately

app/include/zmk/behavior_queue.h
```diff
@@ -10,5 +10,5 @@
 #include <stdint.h>
 #include <zmk/behavior.h>
 
-int zmk_behavior_queue_add(uint32_t position, const struct zmk_behavior_binding behavior,
-                           bool press, uint32_t wait);
+int zmk_behavior_queue_add(struct zmk_behavior_binding_event *event,
```
petejohanson, line 13: Can probably be `const`:

```suggestion
int zmk_behavior_queue_add(const struct zmk_behavior_binding_event *event,
```

app/include/zmk/split/bluetooth/service.h
```diff
@@ -20,6 +20,7 @@ struct sensor_event {
 
 struct zmk_split_run_behavior_data {
     uint8_t position;
+    uint8_t source;
```
petejohanson, line 23: So, putting this in the *middle* of this packed data for sure could cause issues for this change and one or the other sides being out of sync. However, reviewing the code in `service.c`, I see that either way, when we grow this structure, if you flash only the central side, the peripheral side will drop/ignore the run GATT write due to the size being bigger than expected, which is probably a reasonable failure mode for this.

app/src/behavior.c
```diff
@@ -17,11 +17,18 @@
 
 #endif
 
+#include <zmk/ble.h>
+#if ZMK_BLE_IS_CENTRAL
```
petejohanson, line 21:
```suggestion
#if ZMK_BLE_IS_CENTRAL
#include <zmk/ble.h>
```

app/include/zmk/behavior.h
```diff
@@ -42,6 +45,19 @@ struct zmk_behavior_binding_event {
...
+ * @brief Invoke a behavior given its binding and invoking event details.
+ *
+ * @param src_binding Behavior binding to invoke.
+ * @param event The binding event struct containing details of the event that invoked it.
+ * @param pressed Whether the binding is pressed or released.
+ *
+ * @retval 0 If successful.
+ * @retval Negative errno code if failure.
+ */
+int zmk_invoke_behavior_binding(const struct zmk_behavior_binding *src_binding,
```
petejohanson, line 58: I would properly namespace this, e.g.:

```suggestion
int zmk_behavior_invoke_binding(const struct zmk_behavior_binding *src_binding,
```

app/src/behavior.c
```diff
@@ -49,6 +56,66 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+static int invoke_locally(struct zmk_behavior_binding *binding,
+                          struct zmk_behavior_binding_event event, bool pressed) {
+    if (pressed) {
+        return behavior_keymap_binding_pressed(binding, event);
+    } else {
+        return behavior_keymap_binding_released(binding, event);
+    }
+}
+
+int zmk_invoke_behavior_binding(const struct zmk_behavior_binding *src_binding,
```
petejohanson, line 68: Ditto:

```suggestion
int zmk_behavior_invoke_binding(const struct zmk_behavior_binding *src_binding,
```

petejohanson (changes requested):
Thanks for sticking with this! A few thoughts on the implementation.

app/include/zmk/split/bluetooth/service.h
caksoylar: I see, thanks for that -- I haven't tested backwards compatibility. Do you still prefer I move this to the end, or keeping as is?

petejohanson: As is should be fine.

app/src/behavior.c
caksoylar: This one is actually to get `ZMK_BLE_IS_CENTRAL` itself (and I believe nothing else, actually).

caksoylar:
Went over the review items, squashed commits to be a bit saner and gave it a drive on my test split to check the previously tested items (no encoder, but everything else).

app/src/behavior.c
petejohanson: Ah, then disregard. Thanks.

petejohanson (approved):
Thanks for sticking with this!

PR #2411, opened 2024-08-11
feat(docs) Minor improvements to the docs

Nick-Munnich:
Added a note on USB logging, pointing people towards output selection.

Added a note on the oscillator issue and how to resolve it for the nRF52 using the "internal oscillator".

Made the config/index page a bit more explicit about which files are searched for in the boards and shields folders.

### DIFF
--- a/docs/docs/config/index.md
+++ b/docs/docs/config/index.md
@@ -37,11 +37,9 @@ ZMK will search for config files in:
 
 ...where `<board>` is the name of the board and `<module>` is the root directory of any [included module](../features/modules.mdx). These files describe the hardware of the board.
 
-ZMK will search the board folder for the following config files:
+ZMK will search the board folder for the following config files _in addition_ to [Zephyr board-defining files](https://docs.zephyrproject.org/3.5.0/hardware/porting/board_porting.html#create-your-board-directory):
 
-- `<board>_defconfig` (Kconfig)
-- `<board>.conf` (Kconfig)
-- `<board>.dts` (Devicetree)
+- `<board>.conf` (Kconfig - For backwards compatibility only, do not use.)
 - `<board>.keymap` (Devicetree, keyboards with onboard controllers only)
 
 Shared config files (excluding any `_left` or `_right` suffix) are not currently supported in board folders.
@@ -59,10 +57,9 @@ When building with a shield, ZMK will search for config files in:
 
 ...where `<shield>` is the name of the shield and `<module>` is the root directory of any [included module](../features/modules.mdx). These files describe the hardware of the shield that the board is plugged into.
 
-ZMK will search the shield folder for the following config files:
+ZMK will search the shield folder for the following config files _in addition_ to [Zephyr shield-defining files](https://docs.zephyrproject.org/3.5.0/hardware/porting/shields.html#shield-porting-and-configuration):
 
-- `<shield>.conf` (Kconfig)
-- `<shield>.overlay` (Devicetree)
+- `<shield>.conf` (Kconfig - For backwards compatibility only, do not use.)
 - `<shield>.keymap` (Devicetree)
 
 Shared config files (excluding any `_left` or `_right` suffix) are not currently supported in shield folders.
--- a/docs/docs/config/index.md
+++ b/docs/docs/config/index.md
@@ -39,7 +39,7 @@ ZMK will search for config files in:
 
 ZMK will search the board folder for the following config files _in addition_ to [Zephyr board-defining files](https://docs.zephyrproject.org/3.5.0/hardware/porting/board_porting.html#create-your-board-directory):
 
-- `<board>.conf` (Kconfig - For backwards compatibility only, do not use.)
+- `<board>.conf` (Kconfig)
 - `<board>.keymap` (Devicetree, keyboards with onboard controllers only)
 
 Shared config files (excluding any `_left` or `_right` suffix) are not currently supported in board folders.
@@ -59,7 +59,7 @@ When building with a shield, ZMK will search for config files in:
 
 ZMK will search the shield folder for the following config files _in addition_ to [Zephyr shield-defining files](https://docs.zephyrproject.org/3.5.0/hardware/porting/shields.html#shield-porting-and-configuration):
 
-- `<shield>.conf` (Kconfig - For backwards compatibility only, do not use.)
+- `<shield>.conf` (Kconfig)
 - `<shield>.keymap` (Devicetree)
 
 Shared config files (excluding any `_left` or `_right` suffix) are not currently supported in shield folders.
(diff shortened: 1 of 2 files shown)


docs/docs/config/index.md
```diff
@@ -37,11 +37,9 @@ ZMK will search for config files in:
 
 ...where `<board>` is the name of the board and `<module>` is the root directory of any [included module](../features/modules.mdx). These files describe the hardware of the board.
 
-ZMK will search the board folder for the following config files:
+ZMK will search the board folder for the following config files _in addition_ to [Zephyr board-defining files](https://docs.zephyrproject.org/3.5.0/hardware/porting/board_porting.html#create-your-board-directory):
 
-- `<board>_defconfig` (Kconfig)
-- `<board>.conf` (Kconfig)
-- `<board>.dts` (Devicetree)
+- `<board>.conf` (Kconfig - For backwards compatibility only, do not use.)
```
caksoylar, line 42: Why is this and the below shield one "do not use"?

docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -105,6 +109,18 @@ Some devices and operating systems may have additional restrictions that they re
 Since ZMK associates pairing/bond keys with hardware addresses of hosts, you cannot pair to two different operating systems in a dual boot system at the same time.
 While you can find [documented workarounds](https://wiki.archlinux.org/title/bluetooth#Dual_boot_pairing) that involve copying pairing keys across operating systems and use both OS with a single profile, they can be fairly involved and should be followed with caution.
 
+### Mitigating Faulty Oscillator
```
caksoylar, line 112:
```suggestion
### Mitigating a Faulty Oscillator
```
?

docs/docs/config/index.md
Nick-Munnich: It's my current belief that the board/shield's configuration options should all be set in the .defconfig file, to then be overwritten by a .conf in the user's config. I think the usage of .conf in the board/shield definition folder should be discouraged. 

That said, upon looking at it more closely since you pointed it out, I stumbled across [this conversation](https://discord.com/channels/719497620560543766/1206391546090496060). Seems there is a use for it still, so I'll remove those lines and request that they be re-added in #1886?

caksoylar: > That said, upon looking at it more closely since you pointed it out, I stumbled across [this conversation](https://discord.com/channels/719497620560543766/1206391546090496060). Seems there is a use for it still, so I'll remove those lines and request that they be re-added in https://github.com/zmkfirmware/zmk/pull/1886?

That's one concern, where some setting just need to be set there right now like you found. Currently these boards/shields use this:
- zmk_uno
- waterfowl
- tidbit
- settings_reset
- nice_view
- nibble
- murphpad
- knob_goblin

The other concern is that these (the shared versions for splits) get copied by the setup script as the user .conf templates, like @honorless pointed out. They should only be used for that purpose, ideally, but for above list it isn't the case.

> ...Given that whatever succeeds the current setup script is unlikely to do the same and the paucity of boards which are genuinely reliant on said script[1](https://github.com/zmkfirmware/zmk/pull/2411#user-content-fn-1-e7a03aa433242dcc6d914ec14cde792c), I am not sure there is much point in perpetuating its existence.
I think you are underestimating this a bit; any keyboard like Corne that can use displays/underglow benefits from the discoverability, via the commented Kconfig options

IMO things are too messy right now, where it's hard to know what conf gets loaded and what doesn't, then what overrides what (especially since things work different for boards/shields/ split shields). 
There are two ways I can see to improve this situation:
- Disable loading of .conf files altogether, depend on `Kconfig.*`/`board_defconfig`. This seems ideal if after #1886 everything can be set alright.
- Only discourage using them via docs. This is *probably* fine to do even now, given users can always use the `config/*.conf` instead and it'd only be confusing if they are contributing to ZMK (doubtful at this point). It doesn't break neither the `settings_reset` kind of scenario, nor the template `corne.conf` scenario.

Nick-Munnich: >     * Disable loading of .conf files altogether, depend on `Kconfig.*`/`board_defconfig`. This seems ideal if after [refactor: rearrange Kconfig inclusion order #1886](https://github.com/zmkfirmware/zmk/pull/1886) everything can be set alright.
I'd pick this option, and leave them in for the time being. But I'd also say that in addition to that, it would be good for the "suggested" .conf settings to use a slightly different file extension. That would avoid the confusion while keeping the "suggestions" around.

caksoylar: I am fine to leave it the way it is for now. 

Maybe the files can have a .template extension or something, which can be removed after copy.

Nick-Munnich:
The offending line has been removed.

docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -22,6 +22,10 @@ export const Uf2Tabs = (props) => (
 
 );
 
+:::tip
+[USB logging](../development/usb-logging.mdx) can be very helpful for diagnosing issues with ZMK. However, when connected to USB your ZMK device will output to USB by default. To troubleshoot wireless connection issues using logging, you will need to [change the output](../behaviors/outputs.md).
```
caksoylar, line 26:
```suggestion
[USB logging](../development/usb-logging.mdx) can be very helpful for diagnosing issues with ZMK. However, when connected to USB your ZMK device will output to USB by default. To troubleshoot wireless connection issues using logging, you will need to [change the preferred output endpoint](../behaviors/outputs.md).
```

docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -105,6 +109,18 @@ Some devices and operating systems may have additional restrictions that they re
 Since ZMK associates pairing/bond keys with hardware addresses of hosts, you cannot pair to two different operating systems in a dual boot system at the same time.
 While you can find [documented workarounds](https://wiki.archlinux.org/title/bluetooth#Dual_boot_pairing) that involve copying pairing keys across operating systems and use both OS with a single profile, they can be fairly involved and should be followed with caution.
 
+### Mitigating a Faulty Oscillator
+
+Rarely, due to a manufacturing error, a device may have a faulty oscillator. This prevents it from functioning correctly wirelessly, and can be difficult to diagnose. Some microcontrollers, such as the `nRF52840`, allow for the usage of an internal oscillator instead of an external one. This results in increased power draw, but can be a solution if your device does indeed have a faulty oscillator.
```
caksoylar, line 114:
```suggestion
Rarely, due to a manufacturing error, a device may have a faulty oscillator. This prevents it from functioning correctly wirelessly and can be difficult to diagnose. Some microcontrollers such as the `nRF52840` allow for the usage of an internal oscillator instead of an external one. This results in increased power draw but can be a solution if your device does indeed have a faulty oscillator.
```

caksoylar (approved):

PR #2415, opened 2024-08-13
feat(docs): Add a note on column numbers for DTS errors

caksoylar:
Add a small remark to avoid confusing users with the column number in troubleshooting build errors.

### DIFF
--- a/docs/docs/troubleshooting/building-issues.md
+++ b/docs/docs/troubleshooting/building-issues.md
@@ -23,6 +23,8 @@ If you get an error stating `Keymap node not found, check a keymap is available 
 
 ### Devicetree Errors
 
+#### "parse error"
+
 A `devicetree error` followed by a reference to the line number on `<keyboard>.keymap` refers to an issue at the exact line position in that file. For example, below error message indicates a missing `;` at line 109 of the `cradio.keymap` file:
 
 ```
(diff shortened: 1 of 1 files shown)


docs/docs/troubleshooting/building-issues.md
```diff
@@ -29,6 +29,8 @@ A `devicetree error` followed by a reference to the line number on `<keyboard>.k
 devicetree error: /__w/zmk-config/zmk-config/config/cradio.keymap:109 (column 4): parse error: expected ';' or ','
 ```
 
+Note that the provided number for the column corresponds to the numbering after the preprocessor expansion and it will likely not match the column numbers in your keymap.
```
Nick-Munnich, line 32:
```suggestion
:::warning
The provided number for the column corresponds to the numbering after the preprocessor expansion and will likely not match the column numbers in your keymap.
:::
```
Warning block, maybe?

Nick-Munnich: Hmmm, a warning block doesn't feel quite right on re-read. I would like something to break apart the different devicetree compilation errors, though. Any ideas?

caksoylar: Yeah, I didn't think it was worth attracting any soecial attention to. For different errors, we can do subsections or bullet points?

Nick-Munnich: Would be some massive bullet points. Perhaps subsections if you can think of nice names, or maybe hrules?

caksoylar: I didn't quite like the hrules as a distinguisher: 
image

Pushing a version with subsection headers now.

Nick-Munnich (approved):
LGTM, though can you just double-check that the tip and the note are under the correct subheader before you merge? I can't recall that off the top of my head.

caksoylar:
Yeah, both the indexing tip and the missing `&kp` remark seem to be related to the last "devicetree_generated.h" section.

PR #2419, opened 2024-08-13
feat(docs): Add keyboard vs consumer variants note

caksoylar:
Add a info box on keycodes pages that include keycodes with both `K_` and `C_` variants in order to explain the difference. I thought this was the best place to present this information but comments are welcome.


Nick-Munnich:
The docs themselves look good to me, but I don't understand why you added the text using HTML+JS in `zmk/docs/src/components/codes` rather than using markdown in `zmk/docs/docs/codes`.

caksoylar:
TBH I used the SpellingCaution addition as an example, and this way you can include it in multiple files. But the approach you used in the low power states doc should also be valid, and is less convoluted? I don't know the history of this stuff enough to say if this is needed. But I can give refactoring a try.

caksoylar:
That turned out much better IMO, thanks!

Nick-Munnich (approved):

PR #2420, opened 2024-08-14
refactor(docs): Convert Codes and Setup index pages to category link

caksoylar:
Following the linkification of Behaviors and Config category headers, doing the same for Setup (seems like a net positive) and Codes (might be more controversial since "Full List" isn't exactly like an index).

### DIFF
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -80,8 +80,14 @@ module.exports = {
       ],
     },
     {
-      Codes: [
-        "codes/index",
+      type: "category",
+      label: "Codes",
+      link: {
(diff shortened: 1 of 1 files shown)


caksoylar:
I think that's good but we might want to solicit more feedback before a major change. For this PR I can either wait for that, or remove the codes link.

Nick-Munnich:
Up to you. If you did want to edit and wait for feedback, I think we should make sure to add some redirects as [urob pointed out](https://discord.com/channels/719497620560543766/719909884769992755/1272177149096230974).

PR #2425, opened 2024-08-15
Fix path to `run-test.sh` when ZMK is used as a module

### LINKED ISSUE #2421: `west test` fails if `zmk` is a module
`west test` assumes that `ZMK` is the topdir of the workspace and fails if it isn't due to `/app` not being where it expects it to be:
https://github.com/zmkfirmware/zmk/blob/ba0dee088f014911e0486288068b8a10284d73f9/app/scripts/west_commands/test.py#L37-L40

urob:
Fixes #2421


petejohanson (approved):
Thanks!

PR #2428, opened 2024-08-16
refactor: update obsolete keycodes

lesshonor:
## Board/Shield Check-list

There are a fair number of keymaps which still use deprecated keycodes. I think it's a good idea for new repositories not to start off with obsolete keycodes straight out of ZMK main.

...A lot of keymaps have decorative ASCII comments which also use obsolete keycodes. I think these are a bad idea but clearly others disagree—so I split the difference by updating commented-out keycodes where they wouldn't disturb the alignment, or if I already had to modify the file. The commit that touches any of the comments can be dropped if desired.


caksoylar (approved):
I went through the changes and LGTM, I agree that it'd be nice to get rid of deprecated keycodes in the default keymaps.

petejohanson (approved):
Thanks!

PR #2438, opened 2024-08-24
refactor(docs): refactored the development section

Nick-Munnich:
After discussion on Discord, this PR has changed in scope enough for me to change the title and main text. The original text can still be found below.

This PR refactors the development section. In particular it adds a "Hardware Integration" section and a "Local Toolchain" section under which many of the previous pages now fall. The "Guides" section has been removed.

<details> <summary>Previous text</summary>Added an advanced guides section. Also added module creation page so that there would be an item in the section already. Expect additional pages to be based off of this one in other commits.

I am not entirely happy with the modules page as-is, though I think a good part of that is because I'm imagining it being used together with a `zmk-modules` repo, which doesn't exist yet. The page would need updating when the time comes. It is mostly a cherry-picked version of the corresponding Zephyr page currently, but the far reduced length and focus only on the most relevant bits still makes it worthwhile to have imo.</details>

PR #2330, opened 2024-06-11
feat(docs) Adding some additional clarity around low power states

Nick-Munnich:
Minor improvements to resolve some questions that appeared on Discord. Also made a note of the soft-off/deep-sleep wakeup-sources issue, since I couldn't find that documented anywhere.

PR #2439, opened 2024-08-24
feat(docs): Added soft off setup guide under advanced guides

Nick-Munnich:
This is pretty much exactly the setup content taken from #2330, but with some simplifications to account for the changes made to that PR. This page now falls under the advanced guides section introduced in #2438, and is based on said PR.

EDIT: drafted until #2438 is merged, then I'll update this.


Nick-Munnich:
Folded content back into #2330.

PR #2440, opened 2024-08-25
fix(docs): wording change to improve clarity for split pairing

YodaDaCoda:


Nick-Munnich (approved):
Nice catch, thanks for making a PR!

PR #2449, opened 2024-09-01
chore(deps): npm update on docs

caksoylar:
This should hopefully resolve the `ws` security alert that can't be resolved by its own.

Only impact I see so far is the sidebar of blog posts are categorized by year.


Nick-Munnich (approved):
Think I prefer it with the year separation, actually.

PR #2451, opened 2024-09-01
refactor(blog): Add global authors file and truncate markers

caksoylar:
Build now complains about two things related to the blog:
- not using global authors with `authors.yml`
- not putting truncate markers in blog posts so they can be truncated in the main page

While both behaviors can be disabled, I think it's nicer to follow the recommendation in these cases.

When putting truncate markers I ran into the issue that `.md` files are also considered mdx by default, in Docusaurus 3. This causes an issue: We can't use `<!-- ... -->` in .md files because that's invalid mdx. But if we use the mdx format `{/* ... */}` instead, then prettier tries to fix it to `{/_ ... _/}` thinking it's markdown. 

The [`format` config](https://docusaurus.io/docs/api/docusaurus-config#markdown) controls this and can be set to `detect` to fix this, so I did that. The alternative would be to try to configure prettier instead but I think assuming the extension is correct is the better approach.

### DIFF
--- a/docs/blog/2020-05-24-wip.md
+++ b/docs/blog/2020-05-24-wip.md
@@ -1,9 +1,6 @@
 ---
 title: WIP
-author: Pete Johanson
-author_title: Project Creator
-author_url: https://gitlab.com/petejohanson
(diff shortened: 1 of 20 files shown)


Nick-Munnich (approved):
LGTM, agree with the approach.

Issue #841, opened 2021-06-15
BUG: USB Power detected as a host


hahuang65:
Having a USB-power cable drive the keyboard (not attached to a laptop, cable without data) still has the firmware thinking it's attached to a host, which sends data over USB, instead of BLE.

If the power source isn't detected to be an actual host, it would be nice to automatically switch over to BLE transmission.

i7d3v3l0p3r:
Temporary solution: use `&out OUT_BLE` binding (https://zmk.dev/docs/behaviors/outputs)

piit79:
Oh, I just got caught out by this issue. I see it's been going on for *a while*...

Is it realistically possible to get it fixed? QMK uses USB activity for sidedness detection, would it be possible to somehow monitor USB activity to decide if there's a host connected to USB?

angweekiat:
Hi @hahuang65 @piit79 @i7d3v3l0p3r , this may be premature, but I'm happy to say that the fix PR for this issue is merged into `main` above. As mentioned in the PR, existing documentations (https://zmk.dev/docs/keymaps/behaviors/outputs) will still mention this issue, until we have more real-world testing with the changes.

hahuang65:
Thanks for all the hard work, to ALL the ZMK contributors!

petejohanson:
> Thanks for all the hard work, to ALL the ZMK contributors!

Can you please confirm if current ZMK `main` addresses the issues you were encountering?

piit79:
I just tested with current `main` and I can confirm the board is no longer switching to USB mode when powered from a charger/power bank. Thanks @angweekiat, great job!

PR #2458, opened 2024-09-04
fix(usb): detect USB power mode to fallback to BLE

angweekiat:
Update `zmk_usb_is_hid_ready` to check that HID configuration from the host has been done, before allowing the keyboard to be used as a HID device.

This is done by adding `bool is_configured` and toggling it off when the device is disconnected, and enabling it when `usb_status == USB_DC_CONFIGURED`.

This is a naive solution for https://github.com/zmkfirmware/zmk/issues/841, where having a USB-power cable without data driving the keyboard, like a power bank, still has the firmware thinking that it's connected to a host and futilely sending data over USB.

When the cable to a power source is plugged in, the USB DC status (from Zephyr) goes through the following states:
```
USB_DC_CONNECTED -> USB_DC_SUSPEND
```
While for cables to a host machine:
```
USB_DC_CONNECTED -> USB_DC_SUSPEND -> USB_DC_RESUME -> USB_DC_RESET -> USB_DC_CONFIGURED
```
The status code list can be found at https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/usb/usb_dc.h#L33-L58

In the case of a power bank being plugged in, the code changes above allow `zmk_usb_is_hid_ready` to declare that HID isn't ready yet, and thus allows `endpoints.c` to fallback to BLE connection. 

Steps to reproduce the original issue
---------------------------------
1. (Setup) Have ZMK keyboard with Bluetooth profile enabled
2. Set the keyboard to prefer sending to USB (`&out OUT_USB`)
3. Power on the keyboard via USB via a power bank
4. Type, text should not show up anywhere ('consumed' by power bank)

Testing
------
- Built and flashed the firmware with the commit changes, ran the same steps above, and text showed up on the expected device.
- Tested using 2 laptops (Ubuntu / Windows) using 2 different power banks.
- Tested for regression, keyboard still works as expected for waking up the host machine (`CONFIG_USB_DEVICE_REMOTE_WAKEUP=y`) and for boot protocol (`CONFIG_ZMK_USB_BOOT=y`).

Concerns
--------
- I can't find any USB or USB HID specifications that explicitly state that the host machine will do the USB configuration step before the USB HID connection is considered complete.
- It's my first time contributing to ZMK, please let me know how the pull request can be improved!

### DIFF
--- a/app/src/usb.c
+++ b/app/src/usb.c
@@ -20,6 +20,7 @@
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 static enum usb_dc_status_code usb_status = USB_DC_UNKNOWN;
+static bool is_configured = false;
 
 static void raise_usb_status_changed_event(struct k_work *_work) {
     raise_zmk_usb_conn_state_changed(
@@ -33,8 +34,11 @@ enum usb_dc_status_code zmk_usb_get_status(void) { return usb_status; }
 enum zmk_usb_conn_state zmk_usb_get_conn_state(void) {
     LOG_DBG("state: %d", usb_status);
     switch (usb_status) {
+    case USB_DC_SUSPEND:
     case USB_DC_CONFIGURED:
     case USB_DC_RESUME:
+    case USB_DC_CLEAR_HALT:
+    case USB_DC_SOF:
         return ZMK_USB_CONN_HID;
 
     case USB_DC_DISCONNECTED:
@@ -44,6 +48,11 @@ enum zmk_usb_conn_state zmk_usb_get_conn_state(void) {
     default:
         return ZMK_USB_CONN_POWERED;
     }
+}
+
+bool zmk_usb_is_hid_ready(void) {
+    LOG_DBG("is_configured: %d", is_configured);
+    return zmk_usb_get_conn_state() == ZMK_USB_CONN_HID && is_configured;
 }
 
 void usb_status_cb(enum usb_dc_status_code status, const uint8_t *params) {
@@ -59,6 +68,11 @@ void usb_status_cb(enum usb_dc_status_code status, const uint8_t *params) {
     }
 #endif
     usb_status = status;
+    if (zmk_usb_get_conn_state() == ZMK_USB_CONN_HID) {
+        is_configured |= usb_status == USB_DC_CONFIGURED;
+    } else {
+        is_configured = false;
+    }
     k_work_submit(&usb_status_notifier_work);
 };
 
--- a/app/src/usb.c
+++ b/app/src/usb.c
@@ -51,7 +51,6 @@ enum zmk_usb_conn_state zmk_usb_get_conn_state(void) {
 }
 
 bool zmk_usb_is_hid_ready(void) {
-    LOG_DBG("is_configured: %d", is_configured);
     return zmk_usb_get_conn_state() == ZMK_USB_CONN_HID && is_configured;
 }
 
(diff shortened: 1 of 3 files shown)


app/src/usb.c
```diff
@@ -33,11 +33,8 @@ enum usb_dc_status_code zmk_usb_get_status(void) { return usb_status; }
 enum zmk_usb_conn_state zmk_usb_get_conn_state(void) {
     LOG_DBG("state: %d", usb_status);
     switch (usb_status) {
-    case USB_DC_SUSPEND:
     case USB_DC_CONFIGURED:
     case USB_DC_RESUME:
-    case USB_DC_CLEAR_HALT:
-    case USB_DC_SOF:
```
petejohanson, line 40: Have you had a chance to confirm the state it enters when a host is connected but sleeps/requests the USB device sleep? We should also confirm this is set up properly when making changes here.

Also, are we sure this is consistent across *all* the Zephyr USB drivers?

petejohanson (commented):
Thanks for the PR!

One major question about testing/statuses for another scenario.

app/src/usb.c
angweekiat: Hi, I tested these additional scenarios:
1. Connecting the keyboard to my laptop, sleeping the laptop, and waking it via keyboard, this works
2. Sleeping the laptop, then connect the keyboard, then wake it via keyboard, this works as well

Strangely, the keyboard didn't transition to any suspended state with the laptop, which feels concerning. If it helps, the board I'm testing with is nice nano v2, but I'm uncertain if this behaviour should be expected across all USB drivers.

To account for other behaviors from other devices, perhaps a better approach is to revert the mapping changes, and keep track that the keyboard has at least gone through the `USB_DC_CONFIGURED` state once, before allowing it as a possible endpoint? (Need to see if this works with `ZMK_USB_BOOT`)

petejohanson: > 
> I was being dumb and locked my laptop instead of sleeping it properly, the state does get set into suspended. When suspended, I wasn't able to wake the laptop from the keyboard, but I want to also check if this is expected behavior even from `main` branch.

It *should*, if you have `CONFIG_USB_DEVICE_REMOTE_WAKEUP=y` and the USB driver supports it (which the nrfx one does). In particular, since "suspended" is still considered USB active, then any attempt to send will still pick the USB endpoint, which ultimately triggers this code: https://github.com/zmkfirmware/zmk/blob/main/app/src/usb_hid.c#L132


> 
> To account for other behaviors from other devices, perhaps a better approach is to revert the mapping changes, and keep track that the keyboard has at least gone through the `USB_DC_CONFIGURED` state once, before allowing it as a possible endpoint? (Need to see if this works with `ZMK_USB_BOOT`)

Yeah, I think "has gone through configured state but hasn't gone to disconnected" may be the best way to really track this?

angweekiat: Thanks for the pointers! I changed the code to what we mentioned above, and regression tested that both the wake and boot protocol functionality still works as expected, and also updated the PR description to match.

I did it by adding another `is_configured` bool, but if space is a concern, it's possible to change `usb_status` to store both the status code and this flag as a bitfield and parsing out the values separately, at the cost of readability.

petejohanson (approved):
Finally had a chance to give this a test on a real device. Works like a champ from my few test scenarios. Thanks so much for the PR! I am going to suggest we leave some of our docs around this issue as-is for now, pending further real world testing with your change in place.

Issue #2408, opened 2024-08-07
Synced sleep bewteen central/peripherals


Genteure:
Would be nice to have an option to sync sleep timers.
Currently key presses on a peripheral keeps the central awake but not the reverse. I imagine with a multi-peripheral setup (e.g. with a dongle) key presses on peripheral A would also not keep peripheral B awake.
This leads to the peripheral side occasionally sleeping on it's own. Although wake-up and reconnect is pretty quick in my experience, there's still a few seconds of delay.

Nick-Munnich:
Personally I'd say your suggested function should be how it behaves, I don't see the reason why you would want to have the option to *not* sync sleep timers.

angweekiat:
Hi @Genteure , I took a stab at this, and shared my changes above, based on syncing the last activity timing, rather than sleep timers. One issue I realized is that in deep sleep mode, the BLE capability is turned off, so it's not possible to wake either device from either side (to my knowledge), but the changes do prevent the peripheral from going into deep sleep, if the central device is used, and the timing synced to the peripheral

angweekiat:
Hmm @Genteure , I was reading through some other issues when a realisation occurred, the last activity timers are synced, but the state is not. If the central is powered by USB, it'll never go to sleep, but the last activity time is still synced, and can result in the peripheral going to idle mode, while the central is in active state.

Some users may be ok with this (saves battery on the peripheral), while some may prefer the alternative where even the states are synchronised. Maybe it would be better to explicitly sync the states rather than indirectly through the last activity timers.

Just wanted to check if this use case crossed your mind when raising the issue, and your thoughts on it?

Genteure:
I don't really have a strong opinion but to me syncing time numbers feels better than completely synchronized sleep state.

I can think of four scenarios where the central is connected to USB but the peripherals are not:

1. Using as a USB keyboard w/ wireless split.
2. Dongle mode.
3. Simply charging the battery.
4. Wired split (when zmk fully supports it)

For scenario 4, users would probably not enable sleep to begin with.
For the other three, the peripheral(s) should still able to sleep when all parts are not actively used, otherwise it defeats the whole purpose of having a sleep feature isn't it. If a user doesn't want to have their keyboard go to sleep they can just set `CONFIG_ZMK_SLEEP=n`.


Also a peripheral should still be able to sleep on it's own to save battery when it's not connected to the central.

aeros281:
[...]

PR #2459, opened 2024-09-05
feat(split): sync central & peripherals last activity timing

angweekiat:
Sync central last activity timings to all devices, by having the central emit how long it's inactive, and the peripheral(s) using it to adjust the local last activity time.

Before this commit, key presses on a peripheral keep the central awake, but not vice versa. With this commit, key presses on the central (or hypothetical peripheral B) can keep peripheral A awake.

This is done by:
1. Adding a new `SYNC_ACTIVITY` GATT characteristic for central to sync its inactive time to the other peripheral(s).
2. Central's `activity.c` broadcasting the inactive time at a regular `ZMK_SPLIT_SYNC_LAST_ACTIVITY_INTERVAL_MS` interval
3. Peripheral's `service.c` receiving the data and sending it to its `activity.c`
4. Peripheral's `activity.c` determining the new `activity_last_uptime` value using the inactive duration as a relative difference.

Additionally:
- `central.c` is updated to sync the timers upon a new BLE connection, so new devices don't need to wait for the full interval to get synced.
- If a peripheral is in IDLE mode when the central comes online, it'll get synced to ACTIVE as well.

Cons:
- This solution doesn't handle cases where the peripheral(s) is already in deep sleep, since BLE is turned off. However, the sync-ing can prevent peripheral(s) from going into deep sleep if the central is used.
- If `ZMK_SPLIT_SYNC_SLEEP_TIMERS_INTERVAL_MS` > `ZMK_IDLE_TIMEOUT`, there can be situations where the peripheral(s) goes into IDLE before receiving the next sync from central. The original default interval was 5 minutes, primarily to prevent unnecessary deep sleeps, but it would run into the above case.
- This is a time-based periodic sync, so it won't be as accurate as an event-based sync, where each key press is sync-ed to the peripheral(s). The latter would potentially take up more of the BLE channel bandwidth. Maybe this accurate mode can be toggled as an option?

Test steps
---------
1. Build and flash both sides to the target keyboard, ideally with indicators such as backlight, battery, RGB underglow, displays, to detect when one side has gone to idle/active.
2. Allow the peripheral to go to idle state (default timeout: 30s)
3. Press any key on the central that causes it to switch to active. The peripheral should go to active state, **not immediately**, but within 30s.

Local test
------------
I added log lines to check the output, that the values are sent and received correctly on both ends. However, if the sync interval is set to a low value (30 seconds), the log lines would end up spamming the console, so I opt'ed to remove them. Let me know if logs can be introduced more nicely, or if there's a better way to check that the code works for the context of reviewing this pull request!

Raised as an issue in https://github.com/zmkfirmware/zmk/issues/2408


angweekiat:
I slept on it, and added a second sync point:
- `ZMK_SPLIT_SYNC_PERIODIC_INTERVAL_MS`: Original periodic sync, every x seconds
- `ZMK_SPLIT_SYNC_EVENT_MIN_INTERVAL_MS`: Syncs each time an event (key press/sensor) is triggered, with a minimum configurable interval to prevent message spamming.

The latter `ZMK_SPLIT_SYNC_EVENT_MIN_INTERVAL_MS` option allows the peripheral state to be synced _almost_ immediately and wait for the next sync periodic sync cycle.

Currently, both options are tied together, I can try making them distinct (either/or options rather than both)

kmobs:
It's a bummer that this never got merged. Been using event based sync for a bit now and it solves my problems when playing a game. Very useful.

lexizz:
May be you can do it??

PR #2464, opened 2024-09-07
feat: Adding simple versioning using GHA to ZMK

Nick-Munnich:
Adds a simple versioning system to ZMK using GHA. 

Conventional commit messages are used to identify when a new tag should be created:
 - commits starting with !, having !: in their commit message, or having BREAKING CHANGE in the message for major
 - commits starting with feat for minor
 - commits starting with fix for patch

 New tags are always accompanied by a release. `SKIP VERSIONING` in a commit message can be used to skip versioning for that commit.


petejohanson:
I think we'll really want to consider how we want to do this. I'm not sure we really want to commit to an automated semantic versioning scheme at this stage for ZMK or not.

If *not*, I've been really happy with the fairly simple https://github.com/googleapis/release-please we're using for zmk-studio-ts-client. More to ponder.

Nick-Munnich:
I like the [discussion](https://discord.com/channels/719497620560543766/719544230497878116/1282765371585069098) that this sprouted - the more I think about it, the more I favor the [trunk approach](https://discord.com/channels/719497620560543766/719544230497878116/1282942934102315051) raised by damselfly. I'd prefer some sort of automated system over a manual one, personally.

PR #2469, opened 2024-09-09
feat(boards): Generic 40%, 50%, 60%, 65%, 75% and tkl physical layouts

ReFil:
This PR adds physical layout definitions for all the common physical layouts for 40% ortho through tkl unibody keyboards as well as implementing them on the relevant boards.

All boards have been confirmed to build, but I don't have many of these boards to test on personally. I can test on all the boards I do have access to

### DIFF
--- a/app/boards/arm/bt60/bt60.dtsi
+++ b/app/boards/arm/bt60/bt60.dtsi
@@ -17,8 +17,6 @@
         zephyr,sram = &sram0;
         zephyr,flash = &flash0;
         zmk,battery = &vbatt;
-        zmk,kscan = &kscan0;
-        zmk,matrix-transform = &default_transform;
     };
 
     sensors: sensors {
--- a/app/boards/arm/bt60/bt60_v1.dts
+++ b/app/boards/arm/bt60/bt60_v1.dts
@@ -6,12 +6,13 @@
 
 /dts-v1/;
 #include "bt60.dtsi"
+#include <layouts/common/60percent/60percent.dtsi>
 
 
 / {
     chosen {
         zmk,kscan = &kscan0;
-        zmk,matrix-transform = &ansi_transform;
+        zmk,physical-layout = &layout_60_ansi;
     };
 
     ansi_transform: keymap_transform_0 {
@@ -66,19 +67,6 @@
         >;
     };
 
-    split_transform: keymap_transform_4 {
-        compatible = "zmk,matrix-transform";
-        columns = <15>;
-        rows = <5>;
-        map = <
-            RC(0,0) RC(0,1) RC(0,2) RC(0,3) RC(0,4) RC(0,5) RC(0,6) RC(0,7) RC(0,8) RC(0,9) RC(0,10) RC(0,11) RC(0,12)     RC(0,13) RC(0,14)
-            RC(1,0)   RC(1,1) RC(1,2) RC(1,3) RC(1,4) RC(1,5) RC(1,6) RC(1,7) RC(1,8) RC(1,9) RC(1,10) RC(1,11) RC(1,12)   RC(1,13)
-            RC(2,0)     RC(2,1) RC(2,2) RC(2,3) RC(2,4) RC(2,5) RC(2,6) RC(2,7) RC(2,8) RC(2,9) RC(2,10) RC(2,11)          RC(2,13)
-            RC(3,0)   RC(3,2) RC(3,3) RC(3,4) RC(3,5) RC(3,6) RC(3,7) RC(3,8) RC(3,9) RC(3,10)       RC(3,11)     RC(3,12) RC(3,14)
-            RC(4,0)   RC(4,1)   RC(4,2)                          RC(4,6)                  RC(4,10)  RC(4,11)   RC(4,12)    RC(4,13)
-        >;
-    };
-
     kscan0: kscan_0 {
         compatible = "zmk,kscan-gpio-matrix";
         wakeup-source;
@@ -111,4 +99,24 @@
             , <&gpio1 2 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>
             ;
     };
+};
+
+&layout_60_ansi {
+    status = "okay";
+    transform = <&ansi_transform>;
+};
+
+&layout_60_iso {
+    status = "okay";
+    transform = <&iso_transform>;
+};
+
+&layout_60_all1u {
+    status = "okay";
+    transform = <&all_1u_transform>;
+};
+
+&layout_60_hhkb {
+    status = "okay";
+    transform = <&hhkb_transform>;
 };
--- a/app/boards/arm/bt60/bt60_v1.keymap
+++ b/app/boards/arm/bt60/bt60_v1.keymap
@@ -2,26 +2,25 @@
 #include <dt-bindings/zmk/keys.h>
 #include <dt-bindings/zmk/bt.h>
 
-#define ANSI true
-//#define HHKB true
-//#define ISO true
-//#define ALL_1U true
-//#define SPLIT_BKSP_RSHFT true
+#define ANSI
+//#define HHKB
+//#define ISO
+//#define ALL_1U
 
 
 
 / {
     chosen {
     #ifdef ANSI
-        zmk,matrix-transform = &ansi_transform;
-    #elif defined(HHKB)
-        zmk,matrix-transform = &hhkb_transform;
+        zmk,physical-layout = &layout_60_ansi;
     #elif defined(ISO)
-        zmk,matrix-transform = &iso_transform;
+        zmk,physical-layout = &layout_60_iso;
     #elif defined(ALL_1U)
-        zmk,matrix-transform = &all_1u_transform;
+        zmk,physical-layout = &layout_60_all1u;
+    #elif defined(HHKB)
+        zmk,physical-layout = &layout_60_hhkb;
     #else
-        zmk,matrix-transform = &split_transform;
+    #error "Layout not defined, please define a layout by uncommenting the appropriate line in bt60_v1.keymap"
     #endif
     };
 
@@ -148,33 +147,7 @@
             sensor-bindings = <&inc_dec_kp C_VOL_UP C_VOL_DN>;
         };
         #else
-        default_layer {
-                // ------------------------------------------------------------------------------------------
-                // | ESC |  1  |  2  |  3  |  4  |  5  |  6  |  7  |  8  |  9  |  0  |  -  |  =  |BSPC| DEL |
-                // | TAB  |  Q  |  W  |  E  |  R  |  T  |  Y  |  U  |  I  |  O  |  P  |  [  |  ]  |    \    |
-                // | CAPS  |  A  |  S  |  D  |  F  |  G  |  H  |  J  |  K  |  L  |  ;  |  '  |     ENTER    |
-                // |  SHIFT    |  Z  |  X  |  C  |  V  |  B  |  N  |  M  |  ,  |  .  | /   |   SHIFT  |  1  |
-                // |  CTL  |  WIN  |  ALT  |                     SPACE               |   ALT   |  1  | CTRL |
-                // ------------------------------------------------------------------------------------------
-            bindings = <
-                &kp ESC &kp N1 &kp N2 &kp N3 &kp N4 &kp N5 &kp N6 &kp N7 &kp  N8   &kp  N9 &kp  N0  &kp MINUS &kp EQUAL &kp BSPC &kp DEL
-                &kp TAB  &kp Q  &kp W  &kp E  &kp R  &kp T  &kp Y  &kp U  &kp  I    &kp  O  &kp  P   &kp LBKT &kp RBKT  &kp BSLH
-                &kp CLCK  &kp A  &kp S  &kp D  &kp F  &kp G  &kp H  &kp J  &kp  K    &kp  L  &kp SEMI &kp SQT           &kp RET
-                &kp LSHFT  &kp Z  &kp X  &kp C  &kp V  &kp B  &kp N  &kp M  &kp COMMA &kp DOT &kp FSLH    &kp RSHFT    &mo 1
-                &kp LCTRL &kp LGUI &kp LALT            &kp SPACE                          &kp RALT &kp RGUI &kp C_MENU &kp RCTRL
-            >;
-            sensor-bindings = <&inc_dec_kp C_VOL_UP C_VOL_DN>;
-        };
-        raise {
-            bindings = <
-                &kp GRAVE &kp F1 &kp F2 &kp F3 &kp F4 &kp F5 &kp F6 &kp F7 &kp  F8   &kp  F9 &kp  F10  &kp F11 &kp F12   &kp DEL &trans
-                &trans &trans &kp UP &trans &trans &trans &trans &trans &kp INS &trans &kp PSCRN &kp SLCK &kp PAUSE_BREAK &sys_reset
-                &trans    &kp LEFT &kp DOWN &kp RIGHT &trans &trans &trans &trans &trans &trans &kp HOME  &kp PG_UP   &bootloader
-                &kp C_PREV &kp C_VOL_DN &kp C_VOL_UP &kp C_MUTE &trans &trans &trans &trans &trans &kp END &kp PG_DN  &kp C_NEXT &trans
-                &bt BT_PRV &bt BT_NXT  &trans                       &trans                            &trans  &trans &trans &bt BT_CLR
-            >;
-            sensor-bindings = <&inc_dec_kp C_VOL_UP C_VOL_DN>;
-        };
+        #error "Layout not defined, please define a layout by uncommenting the appropriate line in bt60_v2.keymap"
         #endif
     };
 };
--- a/app/boards/arm/ckp/bt60_v2.keymap
+++ b/app/boards/arm/ckp/bt60_v2.keymap
@@ -12,13 +12,13 @@
 / {
   chosen {
   #ifdef ANSI
-    zmk,matrix-transform = &ansi_transform;
+    zmk,physical-layout = &layout_60_ansi;
   #elif defined(ISO)
-    zmk,matrix-transform = &iso_transform;
+    zmk,physical-layout = &layout_60_iso;
   #elif defined(ALL_1U)
-    zmk,matrix-transform = &all_1u_transform;
+    zmk,physical-layout = &layout_60_all1u;
   #elif defined(HHKB)
-    zmk,matrix-transform = &hhkb_transform;
+    zmk,physical-layout = &layout_60_hhkb;
   #else
   #error "Layout not defined, please define a layout by uncommenting the appropriate line in bt60_v2.keymap"
   #endif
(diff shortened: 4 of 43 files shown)


app/boards/arm/bt60/bt60_v1.dts
```diff
@@ -6,15 +6,15 @@
...
 
 / {
     chosen {
-        zmk,kscan = &kscan0;
-        zmk,matrix-transform = &ansi_transform;
+        zmk,physical_layout = &layout_60_ansi;
     };
 
-    ansi_transform: keymap_transform_0 {
+    matrix_transform_60_ansi: matrix_transform_60_ansi {
```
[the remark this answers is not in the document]
ReFil: The boards in this pr that changed the matrix transform in the upstream keymap were the polarity works boards, which used funky preprocessor stuff to support multiple layouts, and the preonic. The Planck also provided multiple transforms for user selection. 

Polarity works have primarily been directing customers to their own editor and their hosted config repo redefines the whole board. Furthermore, I don't think the Preonic and Planck are especially popular boards for zmk. So the number of affected users would be limited, especially given only Planck users who change the default  transform would be affected.

The existing label could be retained using DT aliases I believe, I'll add that. I agree that this is a bit  of a blunt measure, but it only has to be done once.

ReFil:
the use of `&kscan0` has been ended up being a de facto standard across a lot of boards. standardising on the matrix transform naming scheme as ` matrix_transform_<size>_<quirks/specifics>` in documentation is a good idea, I'll add that

app/boards/arm/bt60/bt60.dtsi
```diff
@@ -17,8 +17,6 @@
         zephyr,sram = &sram0;
         zephyr,flash = &flash0;
         zmk,battery = &vbatt;
-        zmk,kscan = &kscan0;
```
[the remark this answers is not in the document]
ReFil: The `chosen` approach is described as a "fallback" in https://github.com/zmkfirmware/zmk/blob/6b4d591c3792c89653e88772f18a3b063fb16d27/app/dts/bindings/zmk%2Cphysical-layout.yaml#L23C78-L23C79 so I used the physical layout approach, happy to change this if it's not the way to do it

app/boards/arm/bt60/bt60_v1.dts
```diff
@@ -6,15 +6,15 @@
...
 /dts-v1/;
 #include "bt60.dtsi"
+#include <layouts/common/60percent.dtsi>
 
 
 / {
     chosen {
-        zmk,kscan = &kscan0;
-        zmk,matrix-transform = &ansi_transform;
+        zmk,physical_layout = &layout_60_ansi;
```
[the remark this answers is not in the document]
ReFil: `physical-layout` is used in the docs (https://zmk.dev/docs/development/hardware-integration/new-shield#physical-layout) and several currently upstream shields. https://github.com/zmkfirmware/zmk/blob/6b4d591c3792c89653e88772f18a3b063fb16d27/app/boards/shields/minivan_studio_tester/minivan_studio_tester.overlay#L19
https://github.com/zmkfirmware/zmk/blob/6b4d591c3792c89653e88772f18a3b063fb16d27/app/boards/shields/zmk_uno/zmk_uno.overlay#L16

app/boards/arm/bt60/bt60_v1.dts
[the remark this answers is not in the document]
ReFil: Im not keen on a breaking change, but I don't see any possible way to implement physical-layout without it being breaking currently on the btxx boards. 

That being said I have an idea that could resolve the issues. I can tweak the physical layout code to pick up the chosen transform if one is present, it's not an ideal solution as it won't handle multiple layouts but it will still work. Add in DT aliases to preserve old labels and then this pr won't be breaking

ReFil:
Unless I'm mistaken the kscan and matrix transform can be overridden in downstream boards by calling the layout node and overriding it, that was the impression under which I did this, if that's not the case then I agree it should be dropped from the layout and configured on a per board basis upstream. Perhaps this is also something that can be addressed through documentation. I'll do some testing quickly and get back on this point

app/boards/arm/bt60/bt60_v1.dts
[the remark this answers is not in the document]
joelspadin: I think in some of the boards Pete updated already, he did something more like

```dts
#include <layouts/whatever.dtsi>

&layout_whatever {
    transform = &my_keyboard_transform;
};
```

which would not require renaming the transforms. I also like it being more explicit rather than the layout defining a name for a transform which it then expects every keyboard using it to use.

app/dts/layouts/common/75percent.dtsi
```diff
@@ -0,0 +1,291 @@
...
+            , <&key_physical_attrs 625 100  375  500       0     0     0>
+            , <&key_physical_attrs 100 100 1000  500       0     0     0>
+            , <&key_physical_attrs 100 100 1100  500       0     0     0>
+            , <&key_physical_attrs 100 100 1200  500       0     0     0>
+            , <&key_physical_attrs 100 100 1300  500       0     0     0>
+            , <&key_physical_attrs 100 100 1400  500       0     0     0>
+            , <&key_physical_attrs 100 100 1500  500       0     0     0>
+            ;
+    };
+    layout_75_iso: layout_75_iso {
```
joelspadin, line 98: I think defining multiple layouts like this would cause all of the layouts in the included file to be compiled, even if only one of them is being used? IIRC Studio allows the user to select from all compiled layouts, in which case this would allow selecting layouts a keyboard might not support.

`/omit-if-no-ref/` could be used to avoid that, but then something would need to reference each layout that is supported somehow.

Would separating each layout into its own file be better? Alternatively, I think you could set `status = "disabled"` on every layout and make boards/shields explicitly set `status = "ok"` to enable the ones that are supported, but that's a bit awkward.

app/dts/layouts/common/tkl.dtsi
```diff
@@ -0,0 +1,101 @@
+#include <physical_layouts.dtsi>
```
joelspadin, line 1: Maybe not worth including until something needs them, but there are several TKL variants. Specifically, every combination of
- ANSI or ISO
- winkey (most common) or winkeyless
- 87 (most common) or 88 key (adds F13)

We may at least want to decide on how all of those variants should be named up front.

app/dts/layouts/common/75percent.dtsi
ReFil: I deal with this by using /delete-node/ for unsupported layouts. I'm not so sure about it though, maybe one file per layout is better

app/dts/layouts/common/tkl.dtsi
ReFil: For all the other layouts i've named them as `layout_<size>_<quirks/specifics>`, I think that can be extended to this too e.g. `layout_tkl_f13ansi`

ReFil:
Just given this a big shake up, removing transform names from the layouts, moving each layout to it's own file so no need for `/delete-node/` schenanigans and going back to chosen kscan nodes

app/dts/layouts/common/75percent.dtsi
petejohanson: Have pondered this for the day. There's a few things that complicate this:

Pros of "one file with related layouts":

* Position maps are really useful, especially if you have something like an ANSI with a split backspace, where if you lake a position map tweak, changing layouts will cause your "backspace" to disappear.
* For many designs, this would likely mean just one file to include.

Cons to "related layouts in one file":

* Risk of C&P because folks don't have good docs/references on how to only use the layouts they want.
* More "busy work" to enable/disable the layouts you want to use.
* Potentially *huge* shared layout files for some formats, e.g. 60%.
* Accidentally having variants enabled you don't need.

I *think* the benefits of users not having to figure out the position map bits would outweigh the other issues, if we can improve on the cons.

Perhaps:

* All layouts in the .dtsi files are `status = "disabled"` by default.
* Users will need to supply their own `transform` property for each layout the want to use *anyways* so the can also set `status = "okay"` at the same time. This matches the Zephyr convention for devices in SoCs, etc too.
* The physical position map code already handles ignoring mapping entries for layouts that are disabled. This means we can have a "complete" map for all possible variants that will only end up built with the pieces needed.
* We don't set a `kscan` or `transform` on any layouts in the layout .dtsi, and allow consumers to set them as appropriate.

Thoughts?

app/boards/arm/bt60/bt60_v1.dts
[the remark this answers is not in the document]
petejohanson: > > The boards in this pr that changed the matrix transform in the upstream keymap were the polarity works boards, which used funky preprocessor stuff to support multiple layouts
> 
> Fair enough for the people who are pointed at the PW repo, but the funky preprocessor stuff exacerbates the problem as it means [pretty much all BTXX keymaps have a `chosen` node with a `matrix-transform`](https://github.com/search?q=path%3A%2Fbt%5Cd%2B_.%2B%5C.keymap%2F+transform+NOT+path%3A%2Fapp%2F+NOT+%26reset&type=code) instead of just the nerds who got curious enough to examine the board definition.
> 
> No hard feelings if you still disagree and would rather just make the breaking change with no backwards compatibility, though. My comments aren't law, and there's already precedent; all 5-column Corne keymap builds were likely broken by the recent switch to `physical-layout`, and [that has a much greater impact](https://github.com/search?q=path%3A%2F%5C%2Fcorne.keymap%24%2F+transform+NOT+path%3Aapp+NOT+%26reset&type=code).

Yeah... I'm considering if there's any way to avoid this breakage, perhaps just *warning* when both are defined and falling back to the "use the assigned transform and ignore the layout" as a reasonable fix for this.

app/dts/layouts/common/75percent.dtsi
ReFil: Having rethought it and fiddled with testing things I'm tempted to suggest individual files per layout right now, we can always add a top level file per size (such as how behaviors.dtsi does it) down the line that incorporates the position mapping. Unless I'm mistaken if layouts are included but not provided a transform there's a compilation error, does that go away if they're disabled? I'm keen to make it as easy as possible for people downstream who just want an ansi 60 or hand wired 4x12.

petejohanson:
> Just given this a big shake up, removing transform names from the layouts, moving each layout to it's own file so no need for `/delete-node/` schenanigans and going back to chosen kscan nodes

Given the fix merged in #2490 is in, I think we should be moving to:

* Have split files like you have here.
* Also offer a "one stop shop" .dtsi that pulls in the unique files, *and* adds a position map. I'm on the fence on the `status` we should set here, but I suspect we should disable them all, and allow importers to enable just the layouts they use. That will allow us to add *more* variants to the aggregate .dtsi later and not break things.

Thoughts?

ReFil:
I was thinking about the position map earlier. I'm not sure if it should be done in a generic basis because the layouts each board supports are going to be different, and if a layout is omitted but the position map is tried to be used the build will fail

petejohanson:
> I was thinking about the position map earlier. I'm not sure if it should be done in a generic basis because the layouts each board supports are going to be different, and if a layout is omitted but the position map is tried to be used the build will fail

Can you describe a bit more the build failure you're talking about?

ReFil:
Would the build not fail if the position map references a disabled or absent layout? It's on my list to add position mapping to the ckp and 40&50% boards anyway so I'll be able to do testing but my understanding of device tree was that it would fail in processing that

petejohanson:
> Would the build not fail if the position map references a disabled or absent layout? It's on my list to add position mapping to the ckp and 40&50% boards anyway so I'll be able to do testing but my understanding of device tree was that it would fail in processing that

The code is set up so the position map can reference a disabled layout, and that layout will be skipped. It should compile and work fine. I've tested with posix minivan setup recently.

ReFil:
Ah ok awesome, I'll get to work on adding that later then

ReFil:
Physical layout maps for 40 through 75% have been added, and all multi layout boards now incorporate the physical layout map. I've tested 60, 65 and 75 and checked that 40 and 50 compile but haven't got a suitable board for testing to hand

app/boards/arm/bt60/bt60_v1.keymap
```diff
@@ -2,33 +2,9 @@
...
-    #elif defined(HHKB)
-        zmk,matrix-transform = &hhkb_transform;
-    #elif defined(ISO)
-        zmk,matrix-transform = &iso_transform;
-    #elif defined(ALL_1U)
-        zmk,matrix-transform = &all_1u_transform;
-    #else
-        zmk,matrix-transform = &split_transform;
-    #endif
-    };
```
petejohanson, line 26: It may be nice to preserve this for non-studio builds, but converted to select the physical layout?

app/boards/arm/ckp/bt60_v2.keymap
```diff
@@ -4,29 +4,9 @@
...
-  #elif defined(ISO)
-    zmk,matrix-transform = &iso_transform;
-  #elif defined(ALL_1U)
-    zmk,matrix-transform = &all_1u_transform;
-  #elif defined(HHKB)
-    zmk,matrix-transform = &hhkb_transform;
-  #else
-  #error "Layout not defined, please define a layout by uncommenting the appropriate line in bt60_v2.keymap"
-  #endif
-  };
```
petejohanson, line 25: Ditto

app/boards/arm/kbdfans_tofu65/kbdfans_tofu65_v2.dts
```diff
@@ -115,3 +117,6 @@ zephyr_udc0: &usbd {
     status = "okay";
 };
 
+&layout_65_ansi {
+    transform = <&default_transform>;
+};
```
petejohanson, line 122: Not necessary for you to do this right now, but for anyone finding this later, would be nice to add support for the other layouts found at https://kbdfans.com/products/tofu60-2-0?variant=41382653886603 e.g.:

image

app/dts/layouts/common/40percent/40percent.dtsi
```diff
@@ -0,0 +1,53 @@
+#include <layouts/common/40percent/1x2u.dtsi>
```
petejohanson, line 1: I don't really like the naming for this nor the 50percent folder... Although 60%, 65%, etc. have a fairly standard small set of standard layouts, the world of 50 and 40% keyboards has way too much variation, and trying to make catch all folders for these seems fraught.

I'd rather name thes something like `ortho_5x12` and `ortho_4x12`. Thoughts?

petejohanson (changes requested):
A few thoughts on the implementation. I think we are very close here though.

Thanks for working on this!

app/boards/arm/kbdfans_tofu65/kbdfans_tofu65_v2.dts
ReFil: That's for the tofu60, the tofu65 has only one layout https://kbdfans.com/products/tofu65-2-0?_pos=3&_sid=bb6d96f98&_ss=r

petejohanson: Oh indeed. Thanks

app/dts/layouts/common/40percent/40percent.dtsi
petejohanson: Two ways we could do this:

1. Still under the `40percent` folder, but have sub-aggregates like `40percent/ortho_4x12.dtsi` that pulls in `40percent/ortho_4x12_ortho.dtsi`, `40percent/ortho_4x12_mit.dtsi`, etc.
1. Have separate common folders, e.g. `common/ortho_4x12/` and `common/ortho_5x12/`.

app/boards/arm/bt60/bt60_v1.keymap
ReFil: Good idea. I'll incorporate this

app/dts/layouts/common/40percent/40percent.dtsi
ReFil: Maybe separate common folders is a better approach otherwise the 40 and 50% folders may get very cluttered with all sorts of layouts?

petejohanson: > Maybe separate common folders is a better approach otherwise the 40 and 50% folders may get very cluttered with all sorts of layouts?

WFM.

ReFil: What's your opinion on the individual layout node names, is `&layout_40_xxxx` still good, or should I change it to `&layout_ortho_4x12_xxxx`. I can envision a scenario where a 40 board has both ortholinear and staggered layouts so there may be a naming conflict. Perhaps it's better to be fully specific about the name?

petejohanson: I like the more specific, definitely.

app/boards/arm/preonic/preonic_rev3.dts
```diff
@@ -131,3 +132,18 @@ zephyr_udc0: &usb {
         };
     };
 };
+
+&layout_ortho_5x12_2x1u {
```
petejohanson, line 136: Not sure I love this naming, maybe:

```suggestion
&layout_ortho_5x12_all1u {
```

or:

```suggestion
&layout_ortho_5x12_grid {
```

`2x1u` doesn't fully cover the area a 2x2u changes, and is almost defined "in relation to just one of the other layouts" which seems funky. Thoughts? THe `all1u` matches the naming for the 60percent, which makes me favor that slightly.

app/boards/shields/boardsource5x12/boardsource5x12.overlay
```diff
@@ -6,9 +6,25 @@
 
 #include <dt-bindings/zmk/matrix_transform.h>
 
+#include <layouts/common/ortho_5x12/2x1u.dtsi>
```
petejohanson, line 9: Ditto on naming.

app/dts/layouts/common/60percent/60percent.dtsi
```diff
@@ -0,0 +1,72 @@
...
+&layout_60_all1u {
+    status = "disabled";
+};
+
+&layout_60_hhkb {
+    status = "disabled";
+};
+
+/ {
+    60_percent_position_map {
```
petejohanson, line 23: If someone wants to supplement these mappings in their board/shield where they add one more custom layout, they'll need a way to add new entries to the mapping, so we should give this labels folks can reference, e.g.:

```suggestion
    layouts_common_60percent_position_map: layouts_common_60percent_position_map {
```

Sorry for the late addition, just thought of this as I was going over this with fresh eyes.

petejohanson (changes requested):
A few more items. Thanks again!

app/boards/arm/bt60/bt60_v1.dts
```diff
@@ -66,19 +67,6 @@
         >;
     };
 
-    split_transform: keymap_transform_4 {
```
petejohanson, line 69: Is this layout not actually supported by this PCB? Was it an oversight in adding it originally?

petejohanson (commented):
Just one question, otherwise LGTM.

app/boards/arm/bt60/bt60_v1.dts
ReFil: It's a minor variation of the all 1U layout and can be configured as an all 1u layout. whilst being a technically possible layout option I figured it was better to streamline it and just drop it, doing a discrete physical layout for every possible layout isn't really practical. PW never offered an explicit transform for the BT60V2 etc for this layout and it's not been an issue. The all 1u layout covers everything

ReFil: Just searched for it https://github.com/search?q=SPLIT_BKSP_RSHFT&type=code and I don't think anyone is actually using that layout

petejohanson (approved):

PR #2478, opened 2024-09-14
fix(studio): Ensure null-termination of layer name read from settings

xingrz:
This fixes the string leak when a layer name is changed to a longer one, but is discarded and reverted to the original shorter one from ZMK Studio.

### DIFF
--- a/app/src/keymap.c
+++ b/app/src/keymap.c
@@ -787,12 +787,14 @@ static int keymap_handle_set(const char *name, size_t len, settings_read_cb read
             LOG_WRN("Found layer name for invalid layer ID %d", layer);
         }
 
-        int err = read_cb(cb_arg, zmk_keymap_layer_names[layer],
+        int ret = read_cb(cb_arg, zmk_keymap_layer_names[layer],
                           MIN(len, CONFIG_ZMK_KEYMAP_LAYER_NAME_MAX_LEN - 1));
-        if (err <= 0) {
-            LOG_ERR("Failed to handle keymap layer name from settings (err %d)", err);
-            return err;
+        if (ret <= 0) {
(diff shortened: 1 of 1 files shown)


app/src/keymap.c
```diff
@@ -793,6 +793,8 @@ static int keymap_handle_set(const char *name, size_t len, settings_read_cb read
             LOG_ERR("Failed to handle keymap layer name from settings (err %d)", err);
             return err;
         }
+
+        zmk_keymap_layer_names[layer][len] = 0;
```
petejohanson, line 797: We need to be careful just in case someone has changed the `CONFIG_ZMK_KEYMAP_LAYER_NAME_MAX_LEN` value and lowered it since a setting was stored. This should probably be:

```suggestion
        zmk_keymap_layer_names[layer][MIN(len, CONFIG_ZMK_KEYMAP_LAYER_NAME_MAX_LEN - 1)] = 0;
```

To be safe.

xingrz: Maybe using `zmk_keymap_layer_names[layer][err]` would be better? According to the comment in `settings_read_cb`, it represents the number of bytes read.

petejohanson: The length of the stored string may be longer than the allocated memory into which we are loading it. So we need to be more conservative. See the previous comment.

xingrz: Yes, but theoretically, `err` shouldn’t be longer than the length passed to `read_cb`, which is already `MIN(len, CONFIG_ZMK_KEYMAP_LAYER_NAME_MAX_LEN - 1)`.

petejohanson: Oh indeed! That should be fine.

petejohanson (approved):
Thanks!

PR #2480, opened 2024-09-16
Fixing compile bug in ble.c

abecodes:
Unable to run anything since this prevents any kind of build for me


ReFil:
This would only cause issues if you're building with `CONFIG_ZMK_BLE_CLEAR_BONDS_ON_START` enabled

abecodes:
Which I have for some reason (2 years old cfg, not sure what I was doing back then) xD
I will give it a shot without it, thanks for the hint <3 👍  @ReFil

petejohanson (approved):
Indeed. Thanks!

PR #2360, opened 2024-07-03
docs: Add a hardware issues page under troubleshooting

Nick-Munnich:
This PR adds a "hardware issues" page under troubleshooting. It discusses:
- Identifying problems with pins
- Firmware changes needed after soldering a bodge wire
- Common physical problems with antennas
- Advice for troubleshooting the physical problems

In the process, there are also two new shields to help with troubleshooting: `tester_pro_micro` and `tester_xiao`. This partially resolves #2410, though I would really like a more general solution eventually.

PR #2485, opened 2024-09-17
docs(feat): further generalize RGB information

lesshonor:
Using PIO SPI for the "Other Boards" example:
1. Does a better job of demonstrating how the RGB config for hardware X can differ from hardware Y (the current examples are nearly-identical).
2. Provides a reference configuration which an increasing number of hobbyist keyboard makers will inevitably end up looking for in the near term. (When/if ZMK moves to Zephyr 3.7, switching to `worldsemi,ws2812-pio` might be a good idea for further emphasis on point 1.)

---

## Etc.
- Since the primary motivation is to expand beyond a single hardware configuration, much of the rest of the page was edited with [TMTOWTDI](https://en.wiktionary.org/wiki/TMTOWTDI) in mind.
- That said, I don't know if any ZMKers have *ever* tested any of the other LED ICs in the list. Rather than affirmatively stating "ZMK supports X", it's probably better to gently guide users toward checking Zephyr.
- The adjective "per-key" is more widely-used than "under-key".
- ~~Reduce duplication. Instead of copying the `CONFIG_*` section, write it once for end-users then "yes, and" with an example of `Kconfig.defconfig` defaults for board/shield writers.~~
- Explain with examples. Linking to two of Kyria's different board configurations shines more light on the board/shield paradigm. I'm a little less sold on the Corne LED counts but it's accurate at this moment in time.
- Having an info box and a note box right next to each other deemphasises the "callout" aspect of both and looks a little unsightly, so I moved the low-frequency I/O pin disclaimer to a sentence near the beginning of that section.

## What Could Definitely Use Improvement
Maybe the hardware-specific implementations could be a set of tabs for nRF52 WS2812 SPI and RP2040 WS2812 PIO SPI; that would make it easier to pull out the common `led_strip` definition vs replicating.

The "Other Boards" description is genuinely an awful cop-out, but I'm not sure how to more delicately express that if you can't find an explanation or example of something in ZMK, you should check Zephyr. It doesn't help that Zephyr doesn't appear to have an "`led_strip`" index; just the sample code, which I'm not sold on linking to. Maybe it'd be better to link to `worldsemi`, but consider the first bullet in the et cetera list...

### DIFF
--- a/docs/docs/features/underglow.md
+++ b/docs/docs/features/underglow.md
@@ -6,16 +6,16 @@ sidebar_label: RGB Underglow
 RGB underglow is a feature used to control "strips" of RGB LEDs. Most of the time this is called underglow and creates a glow underneath the board using a ring of LEDs around the edge, hence the name. However, this can be extended to be used to control anything from a single LED to a long string of LEDs anywhere on the keyboard.
 
 :::info
-RGB underglow can also be used for under-key lighting. If you have RGB LEDs on your keyboard, this is what you want. For PWM/single color LEDs, see [Backlight](backlight.mdx).
+RGB underglow can also be used for per-key lighting. If you have RGB LEDs on your keyboard, this is what you want. For PWM/single color LEDs, see [Backlight](backlight.mdx).
 :::
 
-ZMK supports all the RGB LEDs supported by Zephyr. Here's the current list supported:
+ZMK relies on Zephyr's `led-strip` drivers for this feature. The following LEDs/LED families have been implemented:
 
-- WS2812-ish (WS2812B, WS2813, SK6812, or compatible)
+- WS2812 (includes WS2812B, WS2813, SK6812, and others)
 - APA102
-- LPD880x (LPD8803, LPD8806, or compatible)
+- LPD880x (includes LPD8803, LPD8806, and others)
 
-Of the compatible types, the WS2812 LED family is by far the most popular type. Currently each of these types of LEDs are expected to be run using SPI with a couple of exceptions.
+The WS2812 LED family is by far the most popular of these types, so this page will primarily focus on working with it.
 
 Here you can see the RGB underglow feature in action using WS2812 LEDs.
 
@@ -34,16 +34,15 @@ CONFIG_ZMK_RGB_UNDERGLOW=y
 CONFIG_WS2812_STRIP=y
 ```
 
-See [Configuration Overview](/docs/config) for more instructions on how to
-use Kconfig.
+See [Configuration Overview](/docs/config) for more instructions on how to use Kconfig.
 
 If your board or shield does not have RGB underglow configured, refer to [Adding RGB Underglow to a Board](#adding-rgb-underglow-to-a-board).
 
 ### Modifying the Number of LEDs
 
-A common issue when enabling underglow is that some of the installed LEDs do not illuminate. This can happen when a board's default underglow configuration accounts only for either the downward facing LEDs or the upward facing LEDs under each key. On a split keyboard, a good sign that this may be the problem is that the unilluminated LEDs on each half are symmetrical.
+The number of LEDs specified in the default configuration for your board or shield may not match the number you have installed. For example, the `corne` shield specifies only 10 LEDs per side while supporting up to 27. On a split keyboard, a good sign of this mismatch is if the lit LEDs on each half are symmetrical.
 
-The number of underglow LEDs is controlled by the `chain-length` property in the `led_strip` node. You can [change the value of this property](../config/index.md#changing-devicetree-properties) in the `<keyboard>.keymap` file by adding a stanza like this one outside of any other node (i.e. above or below the `/` node):
+The `chain-length` property of the `led_strip` node controls the number of underglow LEDs. If it is incorrect for your build, [you can change this property](../config/index.md#changing-devicetree-properties) in your `<keyboard>.keymap` file by adding a stanza like this one outside of any other node (i.e. above or below the `/` node):
 
 ```dts
 &led_strip {
@@ -51,7 +50,7 @@ The number of underglow LEDs is controlled by the `chain-length` property in the
 };
 ```
 
-where the value is the total count of LEDs (per half, for split keyboards).
+For split keyboards, set `chain-length` to the number of LEDs installed on each half.
 
 ## Configuring RGB Underglow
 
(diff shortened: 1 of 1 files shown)


docs/docs/features/underglow.md
```diff
@@ -34,41 +34,47 @@ CONFIG_ZMK_RGB_UNDERGLOW=y
...
+The `chain-length` property of the `led_strip` node controls the number of underglow LEDs. If it is incorrect for your build, [you can change this property](../config/index.md#changing-devicetree-properties) in your `<keyboard>.keymap` file by adding a stanza like this one outside of any other node (i.e. above or below the `/` node):
 
 ```dts
 &led_strip {
     chain-length = <21>;
 };
 ```
 
-where the value is the total count of LEDs (per half, for split keyboards).
+For split keyboards, set `chain-length` to the number of LEDs installed on each half.
```
Nick-Munnich, line 53: I'm not entirely happy with the subsection name considering the text. It makes sense to make a note of this somewhere for cases like the one described, but the name doesn't seem "troubleshooting" enough for my taste. It may also make sense to move this out into a "feature troubleshooting" section at some point.

docs/docs/features/underglow.md
```diff
@@ -34,41 +34,47 @@ CONFIG_ZMK_RGB_UNDERGLOW=y
...
+
+Shields written for boards which support RGB underglow should add a `boards/` folder underneath the shield folder. Inside this `boards/` folder, create a `<board>.overlay` for any of the boards the shield can be used with. Place all hardware-specific configurations in these `.overlay` files.
+
+For example: the `kyria` shield has a [`boards/nice_nano.overlay`](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/boards/nice_nano.overlay) and a [`boards/nrfmicro_13.overlay`](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/boards/nrfmicro_13.overlay), which configure a WS2812 LED strip for the `nice_nano` and `nrfmicro_13` boards respectively.
 
 ### nRF52-Based Boards
 
-With nRF52 boards, you can just use `&spi3` and define the pins you want to use.
+Using an SPI-based LED strip driver on the `&spi3` interface is the simplest option for nRF52-based boards. If possible, avoid using pins which are limited to low-frequency I/O for this purpose. The resulting interference may result in poor wireless performance.
```
Nick-Munnich, line 69: I think I'd move the warning back into the info box and turn the info box into a warning box, personally.

docs/docs/features/underglow.md
```diff
@@ -167,10 +192,11 @@ Once you have your `led_strip` properly defined you need to add it to the root d
...
-```ini
-CONFIG_ZMK_RGB_UNDERGLOW=y
-# Use the STRIP config specific to the LEDs you're using
-CONFIG_WS2812_STRIP=y
+```kconfig
+config ZMK_RGB_UNDERGLOW
+  default y
+  # If not using WS2812 LEDs, imply the appropriate *_STRIP.
+  imply WS2812_STRIP
 ```
```
Nick-Munnich, line 202: I'm iffy about this, as it blurs the lines between keyboard users and keyboard designers a bit too much for my taste. I think we want designers to make the feature available, but not enable it, and the kconfig.defconfig should be written accordingly.

docs/docs/features/underglow.md
```diff
@@ -130,34 +129,60 @@ If your board/shield uses LEDs that require the data sent in a different order,
...
+      chain-length = <10>; /* number of LEDs */
+      spi-one-frame = <0x70>;
+      spi-zero-frame = <0x40>;
+      color-mapping = <LED_COLOR_ID_GREEN
+                            LED_COLOR_ID_RED
+                            LED_COLOR_ID_BLUE>;
+    };
   };
 };
 ```
```
Nick-Munnich, line 181: This looks good to me at a rough glance, I just want to double-check that you tested it.

Nick-Munnich (commented):
Overall this looks like positive changes to me, just some minor nits. That said, I think there's room for some further improvements to the documentation of lighting:

Like [@caksoylar suggested](https://discord.com/channels/719497620560543766/883452966114324550/1282955504674406522), I think it would make sense to move the adding bits into the "Hardware integration" section (likely leaving a link in the lighting pages when doing so). I think after that is done, the pages would be short enough that it would make sense to merge the backlight and underglow pages into a "Lighting" (or similarly named) page, which could better discuss the differences between the two systems. The adding lighting stuff could also be merged under the hardware integration section.

Would making those changes (or similar) be something you'd be interested in doing/including in this PR?

docs/docs/features/underglow.md
caksoylar: I think making these 4-space indents would be good as well, given that's what we typically default to these days.

docs/docs/features/underglow.md
caksoylar: Same 4-space indent comment here.

caksoylar:
> Like [@caksoylar suggested](https://discord.com/channels/719497620560543766/883452966114324550/1282955504674406522), I think it would make sense to move the adding bits into the "Hardware integration" section (likely leaving a link in the lighting pages when doing so). I think after that is done, the pages would be short enough that it would make sense to merge the backlight and underglow pages into a "Lighting" (or similarly named) page, which could better discuss the differences between the two systems. The adding lighting stuff could also be merged under the hardware integration section.

> 
> Would making those changes (or similar) be something you'd be interested in doing/including in this PR?

I think it might be better to move things to the new "Hardware integration" wholesale in a new PR. But it would be good to keep that in mind while updating this page, so that it can be done as a simple copy/paste.

docs/docs/features/underglow.md
caksoylar: Maybe change it to something like "Troubleshooting Partially Working LEDs" (a mouthful, so maybe not the best) for now?

docs/docs/features/underglow.md
```diff
@@ -34,41 +34,47 @@ CONFIG_ZMK_RGB_UNDERGLOW=y
...
-Inside the `boards/` folder, you define a `<board>.overlay` for each different board.
-For example, the Kyria shield has a `boards/nice_nano.overlay` file that defines the RGB underglow for the `nice_nano` board specifically.
+Support for RGB underglow is always added to a board, not a shield. This is because the LED strip drivers rely on hardware-specific interfaces (e.g. SPI, I2S) and configurations, which shields do not control.
+
+Shields written for boards which support RGB underglow should add a `boards/` folder underneath the shield folder. Inside this `boards/` folder, create a `<board>.overlay` for any of the boards the shield can be used with. Place all hardware-specific configurations in these `.overlay` files.
+
+For example: the `kyria` shield has a [`boards/nice_nano.overlay`](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/boards/nice_nano.overlay) and a [`boards/nrfmicro_13.overlay`](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/boards/nrfmicro_13.overlay), which configure a WS2812 LED strip for the `nice_nano` and `nrfmicro_13` boards respectively.
```
caksoylar, line 65:
```suggestion
For example: the `kyria` shield has a [`boards/nice_nano_v2.overlay`](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/boards/nice_nano_v2.overlay) and a [`boards/nrfmicro_13.overlay`](https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/kyria/boards/nrfmicro_13.overlay), which configure a WS2812 LED strip for the `nice_nano_v2` and `nrfmicro_13` boards respectively.
```
Nit, I usually prefer v2 (as the more frequent target) to reduce potential mix-ups if people copy paste things like this.

docs/docs/features/underglow.md
Nick-Munnich: I think the entire warning and much of the section could probably be dropped if we can get #2508 merged, linking to that instead. 

I've seen a clip of the SPI nodes on the n!n causing noticeable input lag when used with a cirque, which I am of the (unconfirmed) belief is due to the low-frequency pins. I've not seen an example of RGB, but I think that's because pretty much all shields will just use the pro micro layout's RX or TX pins for that, and designers who stray are usually smart enough to check. 

In theory, a user could bodge a connection and adjust the definition. Probably not worth considering, though.

I semi-agree: I think we could note down low-frequency pins being used for high-frequency purposes, and I think it would be better positioned in #2360.

Nick-Munnich (approved):
I'm happy with this as-is for now. Working in the new pinctrl page is the only point I think I have remaining, and I'm happy for that to happen in parallel with reshuffling to move the setup details into hardware configuration.

caksoylar (approved):

PR #2487, opened 2024-09-18
fix(studio): Update position map querying

awkannan:
`DT_HAS_COMPAT_STATUS_OKAY` expects a compatible, what we actually want to do is check if the physical layout node in a position map is compatible with zmk_physical_layout and in ok status.

For that, we'd use `DT_NODE_HAS_COMPAT_STATUS` but `COND_CODE_1` only supports functions with one or two args, so we have to use UTIL_AND to accomplish the same behavior.


petejohanson (approved):
Tested here with the posix MiniVan, and also confirmed that disabling one of the layouts there still works as expected. Thanks for the fix!

PR #2491, opened 2024-09-19
fix: Properly calculate highest active layer for display.

petejohanson:


caksoylar:
Is this also changing back the layer name for layer without display-name to returning the integer index?

petejohanson:
> Is this also changing back the layer name for layer without display-name to returning the integer index?

Ok, found a further fix that was required for this. Tested on the nice!view with the stock widget and it now works as expected.

saponace:
This fix worked for me on a corne v2  with nice!view. Thank you @petejohanson !

amacleod:
For the sake of breadcrumbs, https://github.com/zmkfirmware/zmk/pull/2541 addresses #2540

Issue #1899, opened 2023-08-11
Handle simultaneous combos


dlip:
If you press a combo on the right side eg. j+k to output 'a' at the same time as a combo on the left side eg. d+f to output 'b', the first one wins and you get 'adf' instead of 'ab'.

My use case for this is to improve the implementation of the taipo layout, which uses a lot of combos and treats each side as a one handed keyboard so this type of roll happens quite often. Here are more details on the layout if you are curious https://inkeys.wiki/en/keymaps/taipo

PR #2494, opened 2024-09-22
fix(combos): clear combo timer only when affected keys are released

### LINKED ISSUE #2378: Releasing other key in middle of the combo breaks the combo trigger
I experience this a lot with combo V when typing fast. For example, in the `A B C` layout, you have a combo on A and B which outputs D. When you want to type CD, pressing C and then A+B often causes CAB to be output if C is released between the A and B presses.

Failing test: https://github.com/ssbb/zmk/commit/672a3995f6a69e934cfeabd12bcdd8317733af17

I think it's related to the https://github.com/zmkfirmware/zmk/issues/1899 but not related to split sides as reported.

ssbb:
Now, the key release listener checks if the released key is part of any combo candidate or active combo, instead of stopping the combo timer on any key release.

I also had to modify the `tap-dance/3b-tap-int-seq` test because it relied on the old combo release behavior, but I would like to double-check if this is the behavior we want to maintain.

Fixes #2378

### DIFF
--- a/app/src/combo.c
+++ b/app/src/combo.c
@@ -270,12 +270,25 @@ static int capture_pressed_key(const struct zmk_position_state_changed *ev) {
 
 const struct zmk_listener zmk_listener_combo;
 
-static int release_pressed_keys() {
+static int release_pressed_key(int32_t position) {
+    for (int i = 0; i < pressed_keys_count; i++) {
+        struct zmk_position_state_changed_event *ev = &pressed_keys[i];
+
+        if (ev->data.position == position) {
+            LOG_DBG("combo: releasing position event %d", ev->data.position);
+            return pressed_keys_count--;
+        }
+    }
+
+    return pressed_keys_count;
+}
+
+static int release_pressed_keys(bool release_first) {
     uint32_t count = pressed_keys_count;
     pressed_keys_count = 0;
     for (int i = 0; i < count; i++) {
         struct zmk_position_state_changed_event *ev = &pressed_keys[i];
-        if (i == 0) {
+        if (release_first && i == 0) {
             LOG_DBG("combo: releasing position event %d", ev->data.position);
             ZMK_EVENT_RELEASE(*ev);
         } else {
@@ -348,7 +361,7 @@ static void activate_combo(struct combo_cfg *combo) {
     struct active_combo *active_combo = store_active_combo(combo);
     if (active_combo == NULL) {
         // unable to store combo
-        release_pressed_keys();
+        release_pressed_keys(true);
         return;
     }
     move_pressed_keys_to_active_combo(active_combo);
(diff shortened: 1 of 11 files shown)


ssbb:
I added one more failing test and a fix for an annoying issue where, if a combo overlaps (so it can't be resolved immediately) and is followed by another combo, only the first one is resolved. I think thats what actually happens in #1899 (since taipo using 3-key overlapping combos of course).

ssbb:
Added a third test and a fix for the same issue as the original one, but for active combo release.

This PR now addresses the following issues:

1. Combo is interrupted if any other key (even without combos assigned to it) is released (tests/other-key-release).
2. Combo is interrupted if any key from an active combo is released (tests/other-key-release-2).
3. A combo with an overlapping combo on the same spot (making it unable to resolve immediately) is interrupted if another combo candidate is initialized, preventing quick combo rolls.

Nick-Munnich:
I wonder if we might run into issues with the inverse problem. Say you hypothetically have `shift, e, a` and a combo on `e` and `a` that outputs something else. I then want to type e.g."Each". There is a chance that with this change the combo will be triggered where it shouldn't be. Muscle memory and timings should prevent that from being a major issue, but should still discuss the possibility.

ssbb:
> I wonder if we might run into issues with the inverse problem. Say you hypothetically have `shift, e, a` and a combo on `e` and `a` that outputs something else. I then want to type e.g."Each". There is a chance that with this change the combo will be triggered where it shouldn't be. Muscle memory and timings should prevent that from being a major issue, but should still discuss the possibility.

Which condition do you mean - the first one, I guess? I would expect it to trigger if the conditions are met (timeout and prior idle), and I don’t see how shift should be involved here (I see how both behaviors could be desired). But I see what you mean. I’ve been using it for the last 4 months without any unexpected actuations, but I don’t use the real (held) shift – I use sticky shift instead. So, maybe it’s worth waiting for someone who uses a real shift to give it a try? In the worst case, this behavior could be hidden under some flag.

yexia0:
I applied this PR. And my hold-preferred layer key seems broken.

ssbb:
> I applied this PR. And my hold-preferred layer key seems broken.

Could you provide a bit more details?

yexia0:
```
        lthp: layer_tap_hold_preferred {
            #binding-cells = <2>;
            tapping_term_ms = <200>;
            quick_tap_ms = <0>;
            flavor = "hold-preferred";
            bindings = <&mo>, <&kp>;
        };
```

I have this layer hold-tap function flavored "hold-preferred" that triggers the hold behavior when the tapping-term-ms has expired or another key is pressed.

After applying the PR, the hold behavior will not trigger when another key is pressed before the layer key released. It is only triggered when tapping-term-ms has expired.

The layer key position is also a part of some other combo so I think this may cause the issue.

my patch branch: https://github.com/yexia0/zmk/tree/combo-fix

PR #2497, opened 2024-09-23
feat: Add support for switching to previously used bluetooth profile

willelind:
This is a PR to make it possible to switch to the index of the previously used profile. This is a feature I will need for my project (swapping between RPi and computer)

There are previous PRs offering this functionality, but they have comments that they didn't bother to fix, so I did it in this one.

I switched LAST_DEV to PRV_DEV as I think LAST could be misinterpreted to mean the last index of the list of profiles.

I also didn't add flash save for the previous_profile as exists for the active_profile. Instead I added initialize to 0, so each time you boot, the PRV_DEV will take you to index 0 for the first time, unless you switch profiles before pressing it with the other profile options. Maybe it should be saved but I thought it was a bit overkill for this, tell me if you want me to add this feature.


Nick-Munnich:
Regarding saving to memory, I think *if* you could get it to save both the current and previous profiles in a single write action to the same block, then it would be worth doing so. I'm not sure if that's possible, though, not looked into it.

Nick-Munnich:
On going through this again, could probably add a test for this.

app/src/behaviors/behavior_bt.c
```diff
@@ -33,6 +33,11 @@ static const struct behavior_parameter_value_metadata no_arg_values[] = {
         .type = BEHAVIOR_PARAMETER_VALUE_TYPE_VALUE,
         .value = BT_PRV_CMD,
     },
+    {
+        .display_name = "Previously Used Profile",
```
petejohanson, line 37:
```suggestion
        .display_name = "Previously Selected Profile",
```

app/include/dt-bindings/zmk/bt.h
```diff
@@ -7,9 +7,10 @@
 #define BT_CLR_CMD 0
 #define BT_NXT_CMD 1
 #define BT_PRV_CMD 2
-#define BT_SEL_CMD 3
-#define BT_CLR_ALL_CMD 4
-#define BT_DISC_CMD 5
+#define BT_PRV_DEV_CMD 3
+#define BT_SEL_CMD 4
+#define BT_CLR_ALL_CMD 5
+#define BT_DISC_CMD 6
```
petejohanson, line 13: Add this at the end with a value of 6 please, otherwise this will break folks using ZMK Studio who've saved these few commands to their keymap.

Also, I am not a huge fan of including `DEV` in the naming here, versus "selection" as a noun instead. I think `BT_PRV_SEL_CMD` is preferable.

app/include/zmk/ble.h
```diff
@@ -23,6 +23,7 @@
 void zmk_ble_clear_bonds(void);
 int zmk_ble_prof_next(void);
 int zmk_ble_prof_prev(void);
+int zmk_ble_prof_prev_dev(void);
```
petejohanson, line 26:
```suggestion
int zmk_ble_prof_prev_selection(void);
```

petejohanson (changes requested):
Thanks for the PR! Generally like it, just a few comments on the implementation.

PR #2502, opened 2024-09-24
feat(ext power): Support multiple ext_power GPIOS

ReFil:
Currently ext_power only supports a single GPIO, which causes issues for board with separate power management channels for RGB and displays. This is a small patch to allow all the pins in the gpios phandle array provided to ext_power to be switched. 

Tested working on a board with two ext_power gpios and also tested to confirm no regressions on single ext_power gpio boards


petejohanson (approved):
Easy! Thanks!

PR #2507, opened 2024-09-26
fix: Force Ubuntu 22.04, pending CI fixes.

petejohanson:
Address reported build failures in the community for those affected by `ubuntu-latest` roll out to 24.04: https://github.com/actions/runner-images/issues/10636


petejohanson:
In the interested of not getting "too far behind", I have only changed the two workflows that explicitly call `pip` to install things, since that is a specific area that changed on the newer Ubuntu.

PR #2508, opened 2024-09-26
docs: Added a page on pin control for ZMK

Nick-Munnich:
Should be fine aside from the part which I noted as needing to be tested. I think this is a worthwhile addition, as trying to piece things together from Zephyr's docs and examples can be very difficult, and an increasing number of designers are using features that make use of pin control.


ReFil:
Pinctrl master bindings can be outside of zephyr in the HAL, as I discovered when looking for bindings for the teensy and the samd21. It may be worth noting that in an info box in case someone is trying to find the available bindings for their specific MCU?

https://github.com/zephyrproject-rtos/hal_atmel/blob/master/include%2Fdt-bindings%2Fpinctrl%2Fsamd21-da1eXabcd-pinctrl.h

https://github.com/zephyrproject-rtos/hal_nxp/blob/master/dts%2Fnxp%2Fnxp_imx%2Frt%2Fmimxrt1061xvn5b-pinctrl.dtsi

Nick-Munnich:
> Pinctrl master bindings can be outside of zephyr in the HAL, as I discovered when looking for bindings for the teensy and the samd21. It may be worth noting that in an info box in case someone is trying to find the available bindings for their specific MCU?

Good shout, added

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+```dts title="<my-board>-pinctrl.dtsi"
+#include <dt-bindings/pinctrl/rpi-pico-rp2040-pinctrl.h>
+```
+
+If you are configuring pin control for a [shield](#boards-shields-and-modules), then this file may already be imported by the board's definition, in which case you shouldn't re-include it here.
+
+:::info
+Some MCUs/SoCs may have their headers located in a Zephyr HAL module. See for example the [NXP HAL module](https://github.com/zephyrproject-rtos/hal_nxp/tree/master/dts/nxp/nxp_imx/rt).
+:::
```
ReFil, line 111: Maybe it's better to use the SAMD21 HAL as an example? I think more people will be developing boards using the samd21 rather than the teensy, although admittedly it has some complications around it's SERCOM blocks rather than discrete spi,i2c etc blocks. The imx pinctrl is also rather complex in its own way

Nick-Munnich: I have no particular opinion either way, that change has been made.

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+  ├── boards/
+  │   ├── nice_nano.overlay
+  │   ├── nice_nano_v2.overlay
+  │   ├── nrfmicro_11.overlay
+  │   └── nrfmicro_13.overlay
+  └── <sofle shield defining files>
+  ```
+  Note that you will need to define a separate overlay _for each_ of the boards to be used with the shield.
+
+### External Shields
```
caksoylar, line 40: How about we pose this section in a different way, to side step the issue: We could call this section "adding pinctrl to existing keyboards". This way, we can come at this from a user perspective where things don't have to be 100% compliant, where we can then e.g. recommend putting it in `config/shield.overlay` (I think that should work).

In such a case I also don't know if we'd recommend a PR; they could simply be bodging a display or something to their keyboards.

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+:::
+
+A basic keyboard design as introduced in the [new shield guide](./new-shield.mdx) only uses its pins for the keyboard matrix. Many keyboard designs make use of advanced components or functionality, such as displays or shift registers. This results in the keyboard making use of communication protocols such as (but not limited to) SPI, I2C, or UART. Configuring pins for the usage of advanced functionality such as drivers for the previously named protocols is referred to as "Pin Control".
+
+:::warning
+The details of pin control can vary from vendor to vendor. An attempt was made to be as general as possible, but it isn't possible to cover all possible cases. The approaches for the nRF52840 and rp2040 MCUs/SoCs are documented in their entirety below. For other MCUs/SoCs, please refer to the [Zephyr documentation](https://docs.zephyrproject.org/3.5.0/index.html) and the examples and other files found in-tree of [ZMK](https://github.com/zmkfirmware/zmk/tree/main/app/boards) and [ZMK's fork of Zephyr](https://github.com/zmkfirmware/zephyr).
```
caksoylar, line 16: I think these links also apply to finding an existing pinctrl node/bus for a given board, assuming the user is writing a shield. It might be good info, given e.g. Xiao buses are defined like you note below.

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+Note that not all peripherals require all pins to be defined - for example, there are plenty of SPI devices which do not require either the MISO (Main In, Sub Out) or the MOSI (Main Out, Sub In) pin.
+
+## Pin Control File
+
+Pin control consists of two parts:
+
+1. Defining and grouping the pins together for a driver
+2. Defining a driver node and assigning pin groups to it
```
caksoylar, line 80: "Bus" sounds more familiar to me than "driver", I see you used both below. Maybe add bus here similarly?

One note, for 2. is "overriding/augmenting" rather than "defining" better, since MCU/SoC defines the bus?

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+                    <NRF_PSEL(SPIM_MOSI, 0, 2)>,
+                    <NRF_PSEL(SPIM_MISO, 0, 3)>;
+        };
+    };
+
+    /* configuration for spi0 device, sleep state */
+    spi0_sleep: spi0_sleep {
+        /* group 1 */
+        group1 {
+            /* main role: configure P0.1 as SPI clock, P0.2 as SPI MOSI, P0.3 as SPI MISO */
```
caksoylar, line 153:
```suggestion
            /* main role: configure P0.01 as SPI clock, P0.02 as SPI MOSI, P0.03 as SPI MISO */
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+{label: 'rp2040', value: 'rp2040'},
+]}>
+<TabItem value="nrf52840">
+```dts title="<board>-pinctrl.dtsi"
+&pinctrl {
+    /* configuration for spi0 device, default state */
+    spi0_default: spi0_default {
+        /* group 1 (name is arbitrary) */
+        group1 {
+            /* main role: configure P0.1 as SPI clock, P0.2 as SPI MOSI, P0.3 as SPI MISO */
```
caksoylar, line 142:
```suggestion
            /* main role: configure P0.01 as SPI clock, P0.02 as SPI MOSI, P0.03 as SPI MISO */
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+        };
+    };
+
+};
+
+````
+Pins are always selected via assignments to `psels`. `NRF_PSEL` is a helper macro with the following arguments:
+- Pin function configuration. This is the `name` portion of one of the `NRF_FUNC_{name}` macros found in `zephyr/include/zephyr/dt-bindings/pinctrl/nrf-pinctrl.h`.
+- Port (0 or 1).
+- Pin (0..31).
```
caksoylar, line 167:
```suggestion
- Pin (0 to 31).
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+The values that can be listed in `pinmux` to assign functionality to pins are listed in `zephyr/include/zephyr/dt-bindings/pinctrl/rpi-pico-rp2040-pinctrl.h`.
+
+The following pin properties can be assigned to groups:
+
+- bias-disable: Disable pull-up/down (default, not required).
+- bias-pull-up: Enable pull-up resistor.
+- bias-pull-down: Enable pull-down resistor.
+- input-enable: Enable input from the pin.
+- input-schmitt-enable: Enable input hysteresys.
```
caksoylar, line 224:
```suggestion
- input-schmitt-enable: Enable input hysteresis.
```
I assume?

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,345 @@
...
+- drive-strength: Set the drive strength of the pin, in milliamps. Possible values are: 2, 4, 8, 12 (default: 4mA)
+- slew-rate: If set to 0, slew rate is set to slow. If set to 1, it is set to fast.
```
caksoylar, line 226: Would be good to `code` the property names here.

caksoylar (commented):
This explainer seems pretty useful, and a lot of work!

So far I added some higher level remarks and small issues that caught my eye; I will do a more detailed review later assuming nothing major changes.

Finally, using "RP2040" everywhere would be better than "rp2040" IMO.

docs/docs/development/hardware-integration/pinctrl.mdx
Nick-Munnich: Hmmm. I think we end up wanting both, though from past experience I think we're better off putting it in `config/shield.keymap`. If people want to configure the standard bus then they should PR, otherwise they should `.keymap`.

docs/docs/development/hardware-integration/pinctrl.mdx
Nick-Munnich: While true, I think that would be overloading this block a bit. Not sure on best way to handle.

caksoylar: I realized that these are actually documented for the defined interconnects at 
https://deploy-preview-2508--zmk.netlify.app/docs/development/hardware-integration/new-shield#shield-overlays already. This might be a good opportunity to copy those images here, along with the bus node names as documented. Maybe it can be a separate subsection like "Existing pinctrl definitions for interconnects" (draft name) that replaces the tip box right below?

docs/docs/development/hardware-integration/pinctrl.mdx
caksoylar: The keymap is better because it overrides everything, but it isn't per side (and it seems bad to recommend the per-part keymap trick). The overlay is sometimes overridden by upstream but if the node wasn't defined at all it shouldn't be an issue?

docs/docs/development/hardware-integration/pinctrl.mdx
Nick-Munnich: Good idea, have done so. Had to mess with the interconnect tabs bit somewhat, but it works.

docs/docs/development/hardware-integration/pinctrl.mdx
Nick-Munnich: It turns out that my idea doesn't work after all, so I removed the section and added the forking note.

eldritch-elder:
here's a reminder, documentation or a mention on the required property `clock` in `pio0_spi` is missing :3

docs/src/components/interconnect-tabs.tsx
```diff
@@ -6,25 +6,26 @@ import { groupedMetadata, InterconnectDetails } from "./hardware-utils";
...
+      {gpio && <content.default />}
       {interconnect.node_labels && (
         <>
+          {!gpio && <br></br>}
           The following node labels are available:
           <ul>
-            <li>
-              GPIO: <code>&{interconnect.node_labels.gpio}</code>
-            </li>
+            {gpio && (
```
caksoylar, line 24: Can we document only GPIO in the new shield guide? I don't think other node labels are necessary in the kscan context, especially since the pinctrl docs will have them.

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,336 @@
...
+<InterconnectTabs items={Metadata} gpio={false} />
+
+Note that some boards may have these pins configured to fit the above pinouts, rather than the capabilities of the MCU. Most notably, boards using the nRF52840 often end up putting low-frequency pins at the locations of the SPI buses above. More on low-frequency pins in the next section.
+
+## Pin Selection
+
+Prior to setting up pin control, you will want to make sure that the pins you use are suitable for the driver that you are using. For this purpose, you will most likely need to consult the datasheet of your MCU/SoC directly. Different MCUs/SoCs have different restrictions in regards to this. The nRF52840 and RP2040 are excellent examples:
+
+- The nRF52840 has its pins marked as either low frequency or high frequency.
+  - Low frequency pins are limited to a frequency of 10KHz, making them unusable for I2C, UART, and (most of) SPI.
```
caksoylar, line 60:
```suggestion
  - Low frequency pins should not be used with signals with frequencies exceeding 10KHz, so they are not ideal for I2C, UART, and (most of) SPI.
```
I think it'd be better to document what happens when you do use low frequency pins (interference with radio) rather than say don't use them, given that many designs seem to be using them anyway. Would you agree?

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,336 @@
...
+Note that not all peripherals require all pins to be defined - for example, there are plenty of SPI devices which do not require either the MISO (Main In, Sub Out) or the MOSI (Main Out, Sub In) pin.
+
+## Pin Control File
+
+Pin control consists of two parts:
+
+1. Defining and grouping the pins together for a driver
+2. Augmenting a driver/bus node and assigning pin groups to it
+
+It is standard to do this in two separate files. Create a file called `<my-board>-pinctrl.dtsi`, which will be used for the first part. This file will later be imported into your board's `.dts`. To avoid namespace collisions, if you are configuring pin control for a [shield](#boards-shields-and-modules), then you should write the contents of both parts into the `.overlay` file directly.
```
caksoylar, line 73: Why would there be namespace collisions? (You'd do a relative import, which AFAIK shouldn't have issues)

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,336 @@
...
+
+A basic keyboard design as introduced in the [new shield guide](./new-shield.mdx) only uses its pins for the keyboard matrix. Many keyboard designs make use of advanced components or functionality, such as displays or shift registers. This results in the keyboard making use of communication protocols such as (but not limited to) SPI, I2C, or UART. Configuring pins for the usage of advanced functionality such as drivers for the previously named protocols is referred to as "Pin Control".
+
+:::warning
+The details of pin control can vary from vendor to vendor. An attempt was made to be as general as possible, but it isn't possible to cover all possible cases. The approaches for the nRF52840 and RP2040 MCUs/SoCs are documented in their entirety below. For other MCUs/SoCs, please refer to the [Zephyr documentation](https://docs.zephyrproject.org/3.5.0/index.html) and the examples and other files found in-tree of [ZMK](https://github.com/zmkfirmware/zmk/tree/main/app/boards) and [ZMK's fork of Zephyr](https://github.com/zmkfirmware/zephyr).
+:::
+
+## Boards, Shields, and Modules
+
+:::tip
```
caksoylar, line 23: Do we need this tip box anymore? Or can we move its contents to under the "Predefined Nodes" section?

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,336 @@
...
+
+## Driver/Bus Node
+
+Once pin control for a driver/bus has been defined, you'll need to adjust another node defining the driver/bus. This adjustment can be done in a number of places by convention:
+
+- If defining a unique board, `<board>.dts`
+- If defining boards with multiple revisions/versions that share pin control, `<board>-common.dtsi` (which is then included by each `<board>_<revision>.dtsi`)
+- If [configuring boards for a shield](#boards-shields-and-modules), directly in the `<board>.overlay` file
+
+You'll want to identify the correct node for you to be changing. The nRF52840 has nodes defined in `dts/arm/nordic/nrf52840.dtsi`, while the RP2040 has nodes defined in `dts/arm/rpi_pico/rp2040.dtsi`. **Make sure the node that you're using is not already in use elsewhere.**
```
caksoylar, line 230: What is the reason for the "not already in use" remark? I didn't understand it.

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,336 @@
...
+    };
+    device2: device@1 {
+        compatible = "manufacturer,device";
+        reg = <1>;
+        spi-max-frequency = <1000000>; /* conservatively set to 1MHz */
+    };
+};
+```
+
+Additional information on configuring specific SPI buses can be found in other pages of the ZMK documentation, or in the Zephyr documentation.
```
caksoylar, line 309: "other pages of the ZMK documentation" is this true? 😄 I can't think of a page off the top of my head where this is documented.

caksoylar: Maybe you have the shift register docs in mind?

Nick-Munnich: So in general, the content on this page is focused on setting up pinctrl and in doing so the bus for devices such as shift registers, RGB, displays, etc. The content on setting up the child node of the bus should be in their respective pages, which is what I'm referring to here. I'll rephrase to make it a bit clearer.

docs/docs/development/hardware-integration/pinctrl.mdx
Nick-Munnich: Both the rp2040 and the nrf52840 (and I'd imagine most other MCUs) have a limited number of each bus that can be used, even though the pins that can be tied to each bus are flexible. The nrf52 has a maximum of two I2C buses, and xiao ble sense (Zephyr's 3.7 version, not ours, ours isn't set up for this) uses one of these internally for the gyroscope. If you were to use the same node for I2C as the gyroscope uses, then you'd need to override pinctrl and disable the gyroscope. 

While the gyroscope is irrelevant, this might be relevant for a different board or if someone is defining multiple of the same bus on one device.

docs/docs/development/hardware-integration/pinctrl.mdx
Nick-Munnich: You're right, this is from when I thought my idea to edit existing shields would work and if it did then there would have been namespace collisions.

docs/docs/development/hardware-integration/pinctrl.mdx
caksoylar: That makes sense, OK.

docs/docs/development/hardware-integration/pinctrl.mdx
caksoylar: I see, I interpreted this as more like "don't modify `spi0` if it was already defined" which you will want to do in some/most cases. (There might not be any devices on the bus, or you don't care about them.) It might be better to add a generic warning like "pay attention to what bus you are modifying, which might already have devices defined for it".

Nick-Munnich:
Did some reworking, should have fixed all the issues in the process. Also added a bunch more examples at the bottom of the page, felt they would be a useful reference for people.

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+    pinctrl-0 = <&i2c0_default>;
+    pinctrl-1 = <&i2c0_sleep>;
+    pinctrl-names = "default", "sleep";
+    clock-frequency = <I2C_BITRATE_FAST>;
+
+    /* Nodes using the bus go here */
+};
+```
+
+#### I2C rp2040
```
caksoylar, line 503:
```suggestion
#### I2C RP2040
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+    status = "okay";
+    pinctrl-0 = <&i2c0_default>;
+    pinctrl-names = "default";
+    clock-frequency = <I2C_BITRATE_STANDARD>;
+
+    /* Nodes using the bus go here */
+};
+```
+
+#### I2C rp2040 PIO
```
caksoylar, line 537:
```suggestion
#### I2C RP2040 PIO
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+
+Zephyr currently does not have support for I2C using rp2040 PIO.
```
caksoylar, line 539:
```suggestion
Zephyr currently does not have support for I2C using RP2040 PIO.
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+```dts
+#include "<board>-pinctrl.dtsi"
+
+&uart0 {
+    pinctrl-0 = <&uart0_default>;
+    pinctrl-names = "default";
+};
+```
+
+#### UART rp2040 PIO
```
caksoylar, line 415:
```suggestion
#### UART RP2040 PIO
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+
+In the pin control file:
+
+```dts
+#include <dt-bindings/pinctrl/rpi-pico-rp2040-pinctrl.h>
+
+&pinctrl {
+    /* configuration for the usart0 "default" state */
+    uart0_default: uart0_default {
+        /* group 1 */
```
caksoylar, line 388:
```suggestion
```
I feel like these are unnecessary if they don't give additional info.

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+        group1 {
+            /* configure P0 as UART0 TX */
+            pinmux = <UART0_TX_P0>;
+            };
+        /* group 2 */
```
caksoylar, line 393:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+
+#### UART nRF52840
+
+In the pin control file:
+
+```dts
+&pinctrl {
+    /* configuration for uart0 device, default state */
+    uart0_default: uart0_default {
+        /* group 1 */
```
caksoylar, line 351:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+        group1 {
+            /* configure P0.1 as UART_TX and P0.2 as UART_RTS */
+            psels = <NRF_PSEL(UART_TX, 0, 1)>, <NRF_PSEL(UART_RTS, 0, 2)>;
+        };
+        /* group 2 */
```
caksoylar, line 356:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+Note that bias options are mutually exclusive.
+
+There is an additional child node for the "sleep" state, configuring the pins for low power. More on states will be explained later.
+</TabItem>
+<TabItem value="rp2040">
+```dts title="<board>-pinctrl.dtsi"
+&pinctrl {
+    /* configuration for spi0 device, default state */
+    spi0_default: spi0_default {
+        /* group 1 */
```
caksoylar, line 172:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+        group1 {
+            /* configure P18 as SPI0 clock, P19 as SPI0 MOSI */
+            pinmux = <SPI0_SCK_P18>, <SPI0_TX_P19>;
+        };
+        /* group 2 */
```
caksoylar, line 177:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+            /* configure P16 as SPI0 MISO */
+            pinmux = <SPI0_RX_P16>;
+            /* enable input on pin 1 */
+            input-enable;
+        };
+    };
+
+    /* configuration for an spi device using PIO0, default state  */
+    pio0_spi_default: pio0_spi_default {
+        /* group 1 */
```
caksoylar, line 188:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+        group1 {
+            /* Configure P13 and P14 for PIO, to be used for MOSI and SCK */
+            pinmux = <PIO0_P13>, <PIO0_P14>;
+        };
+        /* group 2 */
```
caksoylar, line 193:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+            /* main role: configure P0.01 as SPI clock, P0.02 as SPI MOSI, P0.03 as SPI MISO */
+            psels = <NRF_PSEL(SPIM_SCK, 0, 1)>,
+                    <NRF_PSEL(SPIM_MOSI, 0, 2)>,
+                    <NRF_PSEL(SPIM_MISO, 0, 3)>;
+        };
+    };
+
+    /* configuration for spi0 device, sleep state */
+    spi0_sleep: spi0_sleep {
+        /* group 1 */
```
caksoylar, line 138:
```suggestion
```

docs/docs/development/hardware-integration/pinctrl.mdx
```diff
@@ -0,0 +1,539 @@
...
+values={[
+{label: 'nRF52840', value: 'nrf52840'},
+{label: 'RP2040', value: 'rp2040'},
+]}>
+<TabItem value="nrf52840">
+```dts title="<board>-pinctrl.dtsi"
+&pinctrl {
+    /* configuration for spi0 device, default state */
+    spi0_default: spi0_default {
+        /* group 1 (name is arbitrary) */
```
caksoylar, line 127:
```suggestion
        /* node name is arbitrary */
```

caksoylar (approved):
All LGTM, again added a few cosmetic suggestions. Thanks for all the work, the new examples are useful as well. 

I didn't check them from a technical POV but they seem good at a glance. If anyone spots an issue we can fix.

PR #2517, opened 2024-09-30
WIP: feat(boards)!: refactor common layouts

joelspadin:
Removed the combined .dtsi file for each group of common layouts, and split the position map across the files in each group. This makes reading the position maps slightly more difficult, but it aligns the common layouts with the keyboard-specific ones such as foostan/corne.dtsi in that simply including each header adds the layout, with no need to set status = "okay".


petejohanson (commented):
Ok, seeing this in practice.... I really do like the fact that boards/shields can just explicitly import the layouts they want and run with it. The splitting up of the position map makes it perhaps a tiny bit harder to grok, but also is *cohesive* with everything for a given specific layout in the same file. I'd say we go with this, unless @ReFil has any concerns after their initial work on this.

ReFil:
I like this implementation, I can foresee it saving a lot of confusion when module developers are unsure whether or not to include `status = "okay";` for their specific board

joelspadin:
I _think_ this is ready for review now. I checked that all the position maps make sense by pasting all the files for each layout together into https://zmk-layout-helper.netlify.app/, but I haven't tested that every board builds yet.

Nick-Munnich:
Could you split the corne layout as well?

joelspadin:
> Could you split the corne layout as well?

I didn't notice that there were multiple layouts there before. Done.

joelspadin:
Looks like there are more things for me to update since the commit where I started this branch.

app/boards/shields/corne/corne.dtsi
```diff
@@ -6,7 +6,8 @@
 
 #include <dt-bindings/zmk/matrix_transform.h>
 
-#include <layouts/foostan/corne.dtsi>
+#include <layouts/foostan/corne/5column.dtsi>
+#include <layouts/foostan/corne/6column.dtsi>
```
caksoylar, line 11: Don't these want to include the position map as well?

petejohanson (approved):
Thanks!

PR #2515, opened 2024-09-29
docs: Streamlined the new shield guide

Nick-Munnich:
This PR builds on #2510 in an attempt to streamline our new shield guide somewhat. I do see some further improvements that can be made, but I think this version reads nicer and eases future improvements. In particular, the future improvements I'd like to see are:

- Once #2456 is merged, replacing the section on module creation with a link to the page created in that PR (with a note on modifying the personal zmk-config for truly personal designs)
- Making a page on kscans which pulls out the information in the config page (currently a bit of an ugly mix of "how to", "explainer", "reference") and referring to that more directly
- Making a page on matrix transformations and referring to that
- Pulling the information on physical transformations out into the relevant zmk-studio page, renaming said page to "Physical Layouts" and referring to that page

I think the above steps are likely required in order to be able to write a nice "New Keyboard Board" guide in the future.

PR #2519, opened 2024-09-30
docs: fixed broken anchors

Nick-Munnich:
Some broken anchors as a result of #2515 that didn't get picked up during `npm run build`.


caksoylar:
I believe the anchors are lower case (`#kconfigshield`), this doesn't seem to work on the deploy preview.

Nick-Munnich:
Noticed it and fixed it just as you pointed it out. Should be all good now.

caksoylar (approved):

PR #2522, opened 2024-10-01
feat(boards): PM RP2040 flash settings support.

petejohanson:
Haven't been able to test, waiting confirmation by other testers.


victorlucachi:
tested with ZMK Studio enabled reviung41 and rp2040 [rev.6 frood](https://github.com/piit79/Frood), no longer seeing errors in console when attempting to save keymap changes.

keymap updates work as expected and are persistent when disconnecting/reconnecting controller to usb.

PR #2527, opened 2024-10-03
docs: Studio setup to include `&studio_unlock`

petejohanson:
* Add a section about adding `&studio_unlock` to the stock keymap.

PR #2533, opened 2024-10-03
docs: Physical layout docs improvements

Nick-Munnich:
Tried to do a revamp for improved clarity and to match the refactoring plan from #2515.

- Added a new configuration page called layout
  - Added physical layout reference information
  - Added position map reference information
  - Moved matrix transform reference information and updated the description somewhat
- Changed studio setup section to physical layout section
  - Moved information out of new shield into said section
  - Reworked the position map stuff to (imo) be more clear
- Added a note in new shield that `&studio_unlock` should be added to the reference keymap if the device is intended for use with zmk studio

There is some duplicated information between physical layout section and the config section. I think it's fine to have it duplicated, but would also be fine with removing one of the copies.

Subsumes #2527


docs/docs/config/layout.md
```diff
@@ -0,0 +1,225 @@
...
+| `complete` | boolean | If the mapping complete describes the key mapping, and no position based mapping should be used. |         |
+
+The `zmk,physical-layout-position-map` node should have one child node per physical layout. Each child node should have the following properties:
+
+| Property          | Type    | Description                                                                       | Default |
+| ----------------- | ------- | --------------------------------------------------------------------------------- | ------- |
+| `physical-layout` | phandle | The physical layout that corresponds to this mapping entry                        |         |
+| `positions`       | array   | Array of key positions that match the same array entry in the other sibling nodes |         |
+
+## Matrix Transform
```
petejohanson, line 62: IMHO, the matrix transform information should be first, since physical layouts are mainly relevant for ZMK Studio, which isn't fully released yet, and not all boards/shields need to support out of the gate today.

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -442,33 +442,15 @@ The matrix transform is also used to "correct" pin orderings into something that
 
 See the [in-tree keyboards](https://github.com/zmkfirmware/zmk/tree/main/app/boards/shields) that ZMK defines for examples of more complex matrix transformations.
 
-Also see the [matrix transform section](../../config/kscan.md#matrix-transform) in the Keyboard Scan configuration documentation for further details and examples of matrix transforms.
+Also see the [matrix transform section](../../config/layout.md#matrix-transform) in the Keyboard Scan configuration documentation for further details and examples of matrix transforms.
 
 ### Physical Layout
 
-The physical layout is the top level entity that aggregates all details about a certain possible layout:
+Your keyboard will need to have a physical layout defined.
```
petejohanson, line 449: This isn't strictly a *need* and I am hesitant to push that for anyone just wanting to "get their keyboard working" and don't want/need alpha/beta level ZMK Studio at this point.

petejohanson (commented):
Thanks for the work on this! In general, I think this is a major imporvement, just a few thoughts on the layout bits vis-a-vis studio's alpha-almost-beta status.

docs/docs/development/hardware-integration/new-shield.mdx
Nick-Munnich: I think we should discourage the use of a chosen `zmk,matrix-transform` in favor of physical transforms, and a bunch of the changes made were done with that in mind. That said, the *need* aspect should be for the basic physical transform node without the `keys` property - I'm thinking adding an "(Optional)" to the header describing the keys property.

docs/docs/config/layout.md
Nick-Munnich: That's fair, I'll make that change. I think in the future we will want them swapped back though.

docs/docs/development/hardware-integration/new-shield.mdx
petejohanson: Yeah, I think that would work ok for me. Just don't want to unnecessarily complicate it for folks at this stage. Thanks!

docs/docs/development/hardware-integration/studio-setup.md
caksoylar, line 1: I thought it was nice to have a page for "ZMK Studio Setup" in general, even though the content was all about physical layouts. Do we foresee needing this in the future?

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+    default_layout: default_layout {
+        compatible = "zmk,physical-layout";
+        display-name = "Default Layout";
+        transform = <&default_transform>;
+        kscan = <&kscan0>;
+    };
+};
+```
+
+It is given a name, a matrix transform, and a kscan. If all of your physical layouts share the same kscan, then the `kscan` property can be omitted - in this case it needs to be set in the [`chosen` node](./new-shield.mdx#chosen-node). See the [configuration section on physical layouts](../../config/) for reference.
```
caksoylar, line 27:
```suggestion
It is given a name, a matrix transform, and a kscan. If all of your physical layouts share the same kscan, then the `kscan` property can be omitted - in this case it needs to be set in the [`chosen` node](./new-shield.mdx#chosen-node). See the [configuration section on physical layouts](../../config/index.md) for reference.
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+
+```dts
+
+#include <layouts/common/60percent/all1u.dtsi>
+#include <layouts/common/60percent/ansi.dtsi>
+#include <layouts/common/60percent/hhkb.dtsi>
+#include <layouts/common/60percent/iso.dtsi>
+
+
+\\ Assigning the suitable matrix transforms
```
caksoylar, line 99:
```suggestion
// Assigning suitable kscan and matrix transforms
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+    };
+};
+```
+
+### Predefined Layout Usage Example
+
+Here is an example of using the predefined physical layouts for a 60% keyboard:
+
+```dts
+
```
caksoylar, line 92:
```suggestion
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+&layout_60_hhkb {
+    transform = <&hhkb_transform>;
+    kscan = <&hhkb_kscan>;
+};
+```
+
+## Multiple Physical Layouts
+
+If a keyboard has multiple possible layouts (ex. you can snap off an outer column), then you should define multiple matrix transformations and multiple physical layouts - one for each possible layout.
```
caksoylar, line 123:
```suggestion
If a keyboard has multiple possible layouts (e.g. you can snap off an outer column), then you should define multiple matrix transformations and multiple physical layouts -- one for each possible layout.
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+// Needed if and only if keys property is used
+#include <physical_layouts.dtsi>
+
+/ {
+    default_layout: default_layout {
+        compatible = "zmk,physical-layout";
+        display-name = "Default Layout";
+        transform = <&default_transform>;
+        kscan = <&kscan0>;
+        keys = // List of key positions, optional
```
caksoylar, line 136:
```suggestion
        keys = <...>; // List of key positions, optional
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+    };
+
+    alt_layout: alt_layout {
+        compatible = "zmk,physical-layout";
+        display-name = "Alternate Layout";
+        transform = <&alt_transform>;
+        kscan = <&alt_kscan0>;
+        keys = // List of key positions, optional
```
caksoylar, line 144:
```suggestion
        keys = <...>; // List of key positions, optional
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+            = <&key_physical_attrs 100 100    0    0       0     0     0>
+            , <&key_physical_attrs 100 100  100    0       0     0     0>
+            , <&key_physical_attrs 100 100    0  100       0     0     0>
+            , <&key_physical_attrs 100 100  100  100       0     0     0>
+            ;
+    };
+};
+```
+
+### Predefined Layout Usage Example
```
caksoylar, line 87:
```suggestion
## Using Predefined Layouts
```
How about moving line 61 here and making this its own level 2 header?

As it is, it seemed a bit awkward to me to have this example with 4 layouts under "basic", *then* have a section on "multiple layouts"

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+        transform = <&alt_transform>;
+        kscan = <&alt_kscan0>;
+        keys = // List of key positions, optional
+    };
+};
+```
+
+### Position Map
+
+A position map is used for precise mapping between layouts. It is used to allow [ZMK Studio](../../features/studio.md) to accurately retain keymaps when switching layouts. A keyboard's position map has a child node for every potential layout of the keyboard, with the corresponding entries in `positions` property will be used to determine the mapping between layouts. A position map looks something like this:
```
caksoylar, line 151: Is this missing an example after the last sentence?

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+                = < 0  1  2  3>
+                , < 4  5  6  7>
+                , < 8  9 10   >
+                , <11 12 13 14>
+                , <15    16   >;
+        };
+    };
+};
+```
+
```
caksoylar, line 294: Is it worth adding a note here on why the former is preferable to this? i.e. that the extra positions can be recovered

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,351 @@
...
+    physical-layout = <&macropad_layout>;
+    positions // This is equivalent to `positions = <0 1 2 3>;`, reshaped for readability
+        = < 0  1 >
+        , < 2  3 >;
+};
+```
+
+Next, you'll need to write child nodes for every other layout, deriving the position map from the reference. Start by deleting any keys which aren't present in your layout, then reassign the numbers so that they match the matrix transform. Finally, (if the reference layout has more keys) add numbers back into the "empty"/"spare"/"nonexistent" positions so that both position maps count up to the same number. The below examples show this process more clearly.
+
+#### Example larger position map
```
caksoylar, line 197: Would it be worth adding an even more basic example as the first one that is not complete?

caksoylar (commented):
Like I mentioned offline, I'll also look into replacing the PNG with an SVG -- that should hopefully work well with both light/dark themes.

docs/docs/development/hardware-integration/physical-layouts.md
Nick-Munnich: Yes, woops. Fixed.

docs/docs/development/hardware-integration/physical-layouts.md
Nick-Munnich: Good idea. I've also added a quick example using the ferris' layout, since that's not a multiple layouts situation, and a sentence to make the situation there clearer.

docs/docs/development/hardware-integration/physical-layouts.md
Nick-Munnich: Personally I'm not a fan of non-`complete;` position maps, I think they're more likely to lead to confusion. I'd rather omit in favor of encouraging people to be explicit.

docs/docs/development/hardware-integration/physical-layouts.md
Nick-Munnich: Added a note.

docs/docs/development/hardware-integration/studio-setup.md
Nick-Munnich: I don't think we do *in this location*. I think a separate page under hardware integration would only be necessary if additional features distinct from physical layouts also needed to be added to keyboard definitions.

I think the role of dedicated "ZMK Studio Setup" should fall under a blog post and we default to keyboards being suitable for Studio (once we've reached beta or similar).

Nick-Munnich (commented):
One thing I didn't note in a comment but I also changed is the addition of labels to position map child nodes, following [this comment and conversation](https://discord.com/channels/719497620560543766/722441502948851741/1292835457763246220) on Discord.

docs/docs/development/hardware-integration/physical-layouts.md
petejohanson: Both are valid. iIMHO, and should be documented. Non-complete maps can save a ton of duplication, and not waste flash space for constrained devices. They *do* have a place, and should be documented.

petejohanson (commented):
Look good, generally. One minor comment/question.

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,429 @@
...
+    positions
+        = < 0  1  2  3>
+        , < 4  5  6  7>
+        , < 8  9 10   >
+        , <11 12 13 14>
+        , <15    16   >;
+};
+```
+
+As a reminder, the ordering of the physical layout and the position map is the same as in the matrix transform and the keymap (all are just one-dimensional arrays presented nicely through the use of whitespace).
```
Nick-Munnich, line 261:
```suggestion
As a reminder, the `positions` property is a one-dimensional array like the `keys` propertly, presented nicely through the use of whitespace.
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,429 @@
...
+            physical-layout = <&numpad_layout>;
+            positions = <7 14 15>;
+        };
+    };
+};
+```
+
+This is noticably simpler to write, and can be a useful way of saving flash space for memory-constrained devices. The above is a "lossy" mapping, though. While "lossless" non-`complete` mappings are possible, they can be counter-intuitive enough that it may be easier to write the full position map instead.
+
+For completion, the equivalent "lossless" non-`complete` position map is shown below:
```
caksoylar, line 353:
```suggestion
For compleness, the equivalent "lossless" non-`complete` position map is shown below:
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,429 @@
...
+    positions
+        = < 0  1  2  3>
+        , < 4  5  6  7>
+        , < 8  9 10   >
+        , <11 12 13 14>
+        , <15    16   >;
+};
+```
+
+As a reminder, the `positions` property is a one-dimensional array like the `keys` propertly, presented nicely through the use of whitespace.
```
caksoylar, line 261:
```suggestion
As a reminder, the `positions` property is a one-dimensional array like the `keys` property, formatted nicely through the use of whitespace and angle bracket groupings.
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,429 @@
...
+                , < 4  5  6  7>
+                , < 8  9 10   >
+                , <11 12 13 14>
+                , <15    16   >;
+        };
+    };
+};
+```
+
+The above example is "lossy" because (unlike the "lossless" example) if a user switches from the macropad layout to the numpad layout _and then_ switches from the numpad layout back to the macropad layout, the assignments to the keys present but not listed in the macropad's map are lost.
```
caksoylar, line 325:
```suggestion
The above example is "lossy" because (unlike the previous "lossless" example) if a user switches from the macropad layout to the numpad layout _and then_ switches from the numpad layout back to the macropad layout, the assignments to the keys present but not listed in the macropad's map are lost.
```
The only place lossless was mentioned was in the node name, so this probably helps

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,429 @@
...
+
+#### Example non-`complete` position map
+
+Consider the above device again - most of the positions have identical `keys` properties. For example, the numpad's `15` key and the macropad's `16` key would have the same physical property, and be mapped to each other automatically. The keys whose mappings are unable to be determined automatically are those with different physical characteristics - the 2u high and 2u wide keys, and their corresponding 1u counterparts.
```
caksoylar, line 329:
```suggestion
Consider the above device again -- most of the positions have identical `keys` properties. For example, the macropad's `0` key and the numpad's `0` key would have the same physical property, and be mapped to each other automatically. The keys whose mappings are unable to be determined automatically are those with different physical characteristics: the 2u high and 2u wide keys, and their corresponding 1u counterparts.
```
While I believe the previous version was technically true, the width of 16/15 were actually different so it had to be qualified. This is probably a simpler example. (also switched order to obey left/right)

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -0,0 +1,429 @@
...
+```dts
+macropad_map: macropad {
+    physical-layout = <&macropad_layout>;
+    positions // This is equivalent to `positions = <0 1 2 3>;`, reshaped for readability
+        = < 0  1 >
+        , < 2  3 >;
+};
+```
+
+Next, you'll need to write child nodes for every other layout, deriving the position map from the reference. Start by deleting any keys which aren't present in your layout, then reassign the numbers so that they match the matrix transform. Finally, (if the reference layout has more keys) add numbers back into the "empty"/"spare"/"nonexistent" positions so that both position maps count up to the same number. The below examples show this process more clearly.
```
joelspadin, line 225: I don't think this makes any sense because we haven't explained what `positions` actually does yet. This is a pretty difficult concept to explain in general, and I can't come up with a way to explain it that I think is very simple, so I think we should try to get as many ideas for this as possible and then see which one people find easiest to understand.

I think of it as the `positions` properties in all the layouts being parallel arrays. The values at the same index in each layout indicate which key positions are the same in each layout. If I switch from layout A to layout B, then the binding at key position `A.positions[i]` in the original layout is transferred to key position `B.positions[i]` in the new layout for every index `i` in the position map. 

Key positions that are past the end of the keymap can also be used in order to store a binding so that it is not lost when switching between layouts with different numbers of keys.

A diagram would probably be very helpful to explaining this too. For example, consider layout A which has three keys and layout B which replaces the first two keys with a 2U key. If the position map is defined like this

```dts
// key positions for layout_a
// +---+---+---+
// | 0 | 1 | 2 |
// +---+---+---+
layout_a {
    ...
    positions = <0 1 2>;
};

// key positions for layout_b
// +-------+---+
// |   0   | 1 |
// +-------+---+
layout_b {
    ...
    positions = <0 2 1>;
};
```
Then this is how the key bindings move when switching between layouts A and B
```
+---+---+---+
+ 0 | 1 | 2 |
+---+---+---+
  ^   ^   ^
  |   |   |
  |   +-- | -------+
  v       v        v
+-------+---+    .....
|   0   | 1 |    . 2 .
+-------+---+    .....
```
Note that the key at position 1 in layout A is moved to an unused position 2 in layout B so that it can be restored when switching back to layout A.

docs/docs/features/studio.md
```diff
@@ -109,3 +109,11 @@ By default, a build with ZMK Studio enabled will only allow as many layers as ar
 ```
 
 The reserved layers will be ignored during regular ZMK builds but will become available for ZMK Studio enabled builds.
+
+## Enabling ZMK Studio for a Keyboard
```
caksoylar, line 113:
```suggestion
## Adding ZMK Studio Support to a Keyboard
```
Maybe this is slightly clearer, as opposed to just turning the feature on?

Nick-Munnich:
Added a note on putting physical layouts into `keyboard-layouts.dtsi`. Also changed the TOC depth so that the examples show up in it.

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -34,7 +35,8 @@ It is given a name, a matrix transform, and a kscan. If all of your physical lay
 
 :::
 
-The `keys` property is required for [ZMK Studio](../../features/studio.md) support. It is used to describe the physical attributes of each key position present in that layout.
+The `keys` property is required for [ZMK Studio](../../features/studio.md) support. It is used to describe the physical attributes of each key position present in that layout. If this property is used, then you should define the physical layouts and any [position maps](#position-map) in a file called `<your keyboard>-layouts.dtsi`. This file should then be imported by the appropriate file (A `.overlay`, `.dts`, or a `.dtsi` which is itself imported by one of the previous).
```
caksoylar, line 38:
```suggestion
The `keys` property is required for [ZMK Studio](../../features/studio.md) support. It is used to describe the physical attributes of each key position present in that layout. If this property is used, then you should define the physical layouts and any [position maps](#position-map) in a file called `<your keyboard>-layouts.dtsi`. This file should then be imported by the appropriate file, such as an `.overlay`, `.dts`, or a `.dtsi` (last of which is itself imported by one of the previous).
```

petejohanson:
@Nick-Munnich Can you please mark as resolved any of the items you've addressed, so we can see what (if anything) is still outstanding here from the review comments?)

petejohanson (approved):
LGTM. @caksoylar any concerns?

caksoylar:
I am good, I wanted to give @joelspadin a day or so in case he has comments.

PR #2537, opened 2024-10-06
refactor(Kconfig): Extracted designer defaults out into new files

Nick-Munnich:
This is an alternative approach to #1886, with the same end goal. All of the "designer configurable" defaults are extracted out to separate files named `Kconfig.defaults`. The *definitions* stay in the same `Kconfig` file, only the defaults move. The defaults are imported after the keyboard config options, allowing them to be overridden by designers.

I used my own judgement to decide whether an option should be treated as "designer configurable". In general:
- Anything related to behaviors or how the keyboard would be *used* is a user config option and the designer has no business touching it
- Anything listed under our docs that isn't strictly a user config option is made available to designers
- ~~Specific additional options are made available to designers, such as USB vendor ID -- companies who have their own VID may want to override this one.~~

~~There are also some options that are listed in the docs, which I don't think should be listed in the docs as I don't understand why either designers or users should have any business touching them. For example:~~
- ~~`CONFIG_ZMK_USB_INIT_PRIORITY`~~
- ~~`CONFIG_ZMK_BLE_INIT_PRIORITY`~~

~~I have included such flags where I was uncertain as "designer configurable" out of caution.~~

While at it, I added a name to the `EC11` trigger mode choice, and updated some values listed in the docs to be accurate. 

Hopefully this more explicit approach is more "palatable" than #1886 and can get merged easier...

### DIFF
--- a/app/Kconfig.defaults
+++ b/app/Kconfig.defaults
@@ -1,0 +1,83 @@
+# Copyright (c) 2024 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+# HID
+if ZMK_HID_REPORT_TYPE_HKRO
+
+config ZMK_HID_KEYBOARD_REPORT_SIZE
+    default 6
+
+endif
+
+config ZMK_HID_CONSUMER_REPORT_SIZE
+    default 6
+
+# Underglow
+if ZMK_RGB_UNDERGLOW
+
+# This default value cuts down on tons of excess .conf files, if you're using GPIO, manually disable this
+config SPI
+    default y
+
+config ZMK_RGB_UNDERGLOW_EXT_POWER
+    default y
+
+config ZMK_RGB_UNDERGLOW_BRT_MIN
+    default 0
+
+config ZMK_RGB_UNDERGLOW_BRT_MAX
+    default 100
+
+config ZMK_RGB_UNDERGLOW_HUE_STEP
+    default 10
+
+config ZMK_RGB_UNDERGLOW_SAT_STEP
+    default 10
+
+config ZMK_RGB_UNDERGLOW_BRT_STEP
+    default 10
+
+config ZMK_RGB_UNDERGLOW_HUE_START
+    default 0
+
+config ZMK_RGB_UNDERGLOW_SAT_START
+    default 100
+
+config ZMK_RGB_UNDERGLOW_BRT_START
+    default ZMK_RGB_UNDERGLOW_BRT_MAX
+
+config ZMK_RGB_UNDERGLOW_SPD_START
+    default 3
+
+config ZMK_RGB_UNDERGLOW_EFF_START
+    default 0
+
+config ZMK_RGB_UNDERGLOW_ON_START
+    default y
+
+endif # ZMK_RGB_UNDERGLOW
+
+# Backlight
+if ZMK_BACKLIGHT
+
+config ZMK_BACKLIGHT_BRT_STEP
+    default 20
+
+config ZMK_BACKLIGHT_BRT_START
+    default 40
+
+config ZMK_BACKLIGHT_ON_START
+    default y
+
+endif # ZMK_BACKLIGHT
+
+# Ext_power
+config ZMK_EXT_POWER
+    default y
+
+# Battery
+config ZMK_BATTERY_REPORT_INTERVAL
+    default 60
+
+# Imports
+rsource "src/split/Kconfig.defaults"
--- a/app/src/split/bluetooth/Kconfig
+++ b/app/src/split/bluetooth/Kconfig
@@ -5,7 +5,7 @@ if ZMK_SPLIT && ZMK_SPLIT_BLE
 
 menu "BLE Transport"
 
-# Added for backwards compatibility. New shields/board should set `ZMK_SPLIT_ROLE_CENTRAL` only.
+# Added for backwards compatibility. New shields / board should set `ZMK_SPLIT_ROLE_CENTRAL` only.
 config ZMK_SPLIT_BLE_ROLE_CENTRAL
     bool
     select ZMK_SPLIT_ROLE_CENTRAL
@@ -24,7 +24,6 @@ if ZMK_SPLIT_ROLE_CENTRAL
 
 config ZMK_SPLIT_BLE_CENTRAL_PERIPHERALS
     int "Number of peripherals that will connect to the central."
-    default 1
 
 menuconfig ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING
     bool "Fetch Peripheral Battery Level Info"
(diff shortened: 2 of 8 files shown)


app/Kconfig.defaults
```diff
@@ -0,0 +1,168 @@
...
+    default y
+
+# USB
+if USB_DEVICE_STACK
+
+config ZMK_USB_INIT_PRIORITY
+    default 94
+
+config ZMK_USB_HID_INIT_PRIORITY
+    default 95
```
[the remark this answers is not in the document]
Nick-Munnich: I agree. I'm not sure why those settings are documented though, like I said. I'll move them back, thoughts on removing said settings from the docs as well?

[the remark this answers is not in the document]
Nick-Munnich: Moved init priorities, stack sizes, queue sizes, and studio transport details back. As mentioned on Discord, I'll look into moving said settings into an "Advanced" section under configuration in our docs in a future PR.

Nick-Munnich:
Moved the USB PID/VID stuff back to the standard Kconfig file. I realised that the `Kconfig.defaults` files should only be used for stuff that both the user and the designer may want to configure. For designer-only and user-only, they have their respective .conf files.

app/src/split/bluetooth/Kconfig
```diff
@@ -1,11 +1,11 @@
-# Copyright (c) 2022 The ZMK Contributors
-# SPDX-License-Identifier: MIT
+#Copyright(c) 2022 The ZMK Contributors
+#SPDX - License - Identifier : MIT
```
petejohanson, line 2: This looks like it was modified in error.

app/src/split/bluetooth/Kconfig
```diff
@@ -1,11 +1,11 @@
...
 
 if ZMK_SPLIT && ZMK_SPLIT_BLE
 
 menu "BLE Transport"
 
-# Added for backwards compatibility. New shields/board should set `ZMK_SPLIT_ROLE_CENTRAL` only.
+#Added for backwards compatibility.New shields / board should set `ZMK_SPLIT_ROLE_CENTRAL` only.
```
petejohanson, line 8: Same.

app/src/split/bluetooth/Kconfig
```diff
@@ -16,15 +16,14 @@ config ZMK_SPLIT_ROLE_CENTRAL
     select BT_GATT_AUTO_DISCOVER_CCC
     select BT_SCAN_WITH_IDENTITY
 
-# Bump this value needed for concurrent GATT discovery of splits
```
petejohanson, line 19: Ditto

app/src/split/bluetooth/Kconfig
```diff
@@ -92,7 +91,7 @@ config BT_MAX_PAIRED
 config BT_MAX_CONN
     default 1
 
-# Allow central to specify connection parameters.
+#Allow central to specify connection parameters.
```
petejohanson, line 94: Here

app/src/studio/Kconfig
```diff
@@ -1,5 +1,5 @@
-# Copyright (c) 2024 The ZMK Contributors
-# SPDX-License-Identifier: MIT
+#Copyright(c) 2024 The ZMK Contributors
+#SPDX - License - Identifier : MIT
```
petejohanson, line 2: And here.

app/src/studio/Kconfig
```diff
@@ -32,7 +32,7 @@ endif
 menuconfig ZMK_STUDIO_RPC
     bool "Remote Procedure Calls (RPC)"
     select NANOPB
-    # These two save stack size
+#These two save stack size
```
petejohanson, line 35: Ditto

petejohanson (changes requested):
As a whole, happy with this, just needs some reformatting funkiness fixed.

Nick-Munnich:
Fixed formatting of comments as requested, did a pass to catch any that were previously misformatted (bunch of wrong spacings). While I was at it, I did a pass over the if blocks, moving lonely ones into `depends on`, and moving another one into a menu (bottom of [this](https://docs.zephyrproject.org/latest/build/kconfig/tips.html#factoring-out-common-dependencies)). 

I also got rid of the if blocks in the `Kconfig.defaults`, as on reading the page again if blocks get turned into depends on internally, making the if statements in the defaults pages redundant. 

I left those last changes as a separate commit, to let it be removed easily if you'd prefer them not to change like this. Let me know if you want me to squash that or remove it.

app/Kconfig.defaults
```diff
@@ -0,0 +1,71 @@
...
+config ZMK_HID_KEYBOARD_REPORT_SIZE
+    default 6
+
+config ZMK_HID_CONSUMER_REPORT_SIZE
+    default 6
+
+# Underglow
+# This default value cuts down on tons of excess .conf files, if you're using GPIO, manually disable this
+config SPI
+    default y
```
petejohanson, line 14: This will have the effect of unconditionally enabling SPI, AFAICT. This is definitely *not* what we want.

petejohanson (changes requested):
One main item from reviewing the code. Will test locally a bit once that's fixed. Thanks!

app/Kconfig.defaults
Nick-Munnich: Good catch, fixed. It seems I may have been a bit overenthusiastic with the removal of if-blocks, went back over things and added some back where it seemed necessary.

Nick-Munnich:
Did some local testing, ended up popping the if-refactor. Turns out they are *not* equivalent, Zephyr docs are wrong there.

I built both halves of corne with n!nv2, nice60, and bt60_v2, and `diff`'d the resulting .config files against each other. For each of those, there was only one difference: The *location* of the `CONFIG_SPI` flag in the resulting output file, be it a comment that it was unset or it being set to `y`.

Currently it is being grouped under the RGB_UNDERGLOW section, with this change it gets moved under the advanced section alongside I2C etc, where it honestly *should* go I feel.

app/src/split/bluetooth/Kconfig
```diff
@@ -1,11 +1,11 @@
-# Copyright (c) 2022 The ZMK Contributors
+# Copyright(c) 2022 The ZMK Contributors
```
petejohanson, line 1: Superfluous whitespace change.

app/src/studio/Kconfig
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2024 The ZMK Contributors
+# Copyright(c) 2024 The ZMK Contributors
```
petejohanson, line 1: Ditto

petejohanson (changes requested):
Looks really good! One minor set of nitpicks left. Thanks!

app/src/split/bluetooth/Kconfig
Nick-Munnich: This and the below whitespace change is because the app/Kconfig file has a whitespace there and I like consistency.

petejohanson: Maybe my brain is rotten.. but isn't this *removing* the whitespace?

Nick-Munnich: No, you're right and I'm being silly. Fixed.

caksoylar:
Does this require any docs changes? Does it affect the warning note at https://zmk.dev/docs/development/hardware-integration/new-shield#user-configuration-files?

PR #2538, opened 2024-10-07
feat(shields): add physical layouts for Jian, Jiran and Jorne

krikun98:
## Board/Shield Check-list

Adding physical layouts and updating keymaps for Jian, Jiran and Jorne keyboards.

### DIFF
--- a/app/boards/shields/jian/jian.dtsi
+++ b/app/boards/shields/jian/jian.dtsi
@@ -6,10 +6,26 @@
 
 #include <dt-bindings/zmk/matrix_transform.h>
 
+#include <layouts/kgoh/jian/full.dtsi>
+#include <layouts/kgoh/jian/6column.dtsi>
+#include <layouts/kgoh/jian/5column.dtsi>
+
+&kgoh_jian_full_layout {
+    transform = <&default_transform>;
+};
+
+&kgoh_jian_6col_layout {
+    transform = <&crkbd_transform>;
+};
+
+&kgoh_jian_5col_layout {
+    transform = <&five_column_transform>;
+};
+
 / {
     chosen {
         zmk,kscan = &kscan0;
-        zmk,matrix-transform = &default_transform;
+        zmk,physical-layout = &kgoh_jian_full_layout;
     };
 
     default_transform: keymap_transform_0 {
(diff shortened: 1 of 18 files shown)


krikun98:
Jorne update was tested on working hardware with `nrfmicro_13` boards, Jian layouts were tested on the [Jorian840](https://github.com/krikun98/jorian840) (physically identical board), and the Jiran layouts were checked with https://zmk-physical-layout-converter.streamlit.app - currently looking for testers for these two shields.

krikun98:
Moved layouts to shield directories, added `feat(shields):` to commits

petejohanson (approved):
All seems reasonable to me. Please let me know once you've had testers confirm, or if you're ok with merging as is and allowing testing later. Thanks!

caksoylar:
In #2544 we are naming the individual position map labels `<layout_name>_posmap`, where `<layout_name>` is the one referred to in `physical-layout` property. It'd be good to conform to those here, at least for the shared layout.

petejohanson:
@krikun98 We've so far only been putting shared layout in place for layouts that are sufficiently common for reuse. I only see the Jian using the shared Jian layout in this PR. Is it used elsewhere that we need to elevate it at this point?

krikun98:
> @krikun98 We've so far only been putting shared layout in place for layouts that are sufficiently common for reuse. I only see the Jian using the shared Jian layout in this PR. Is it used elsewhere that we need to elevate it at this point?

The Jiran is re-using two of the Jian's layouts right in the PR: `app/boards/shields/jiran/layouts/6column.dtsi` and `app/boards/shields/jiran/layouts/jian.dtsi`. There are also out-of-tree boards that re-use the layout - Jorian840, Jiran BLE, etc.

krikun98:
> In #2544 we are naming the individual position map labels `<layout_name>_posmap`, where `<layout_name>` is the one referred to in `physical-layout` property. It'd be good to conform to those here, at least for the shared layout.

An earlier iteration of #2544 had the Corne mappings suffixed with `_mapping`, which is what I based all of these off. Refactored to `_posmap`.

petejohanson (approved):
Thanks!

PR #2541, opened 2024-10-07
Adding length check to layer_label in nice_view status widget, to enable numeric layer display. (fixes #2540)

### LINKED ISSUE #2540: Layer numbers absent from nice!view on Typeractive Corne and Lily58
See also https://github.com/zmkfirmware/zmk/issues/2467 and https://github.com/zmkfirmware/zmk/pull/2491

When using layers with no `display-name`, the widget at the very bottom of the left screen is totally blank. My expectation is that the numeric index of the layer would be printed there. I have reproduced this on my Typeractive Lily58 Choc wireless, and another person on the Typeractive Discord reported this behavior on their Typeractive Corne wireless.

It is possible to work around this by adding `display-name` values to every layer.

amacleod:
Fixes https://github.com/zmkfirmware/zmk/issues/2540

When display names are not present, the `layer_label` field may appear as an empty (zero-length) string rather than `NULL`; in either case, the status widget should show "LAYER N" where N is the 0-indexed layer number.

### DIFF
--- a/app/boards/shields/nice_view/widgets/status.c
+++ b/app/boards/shields/nice_view/widgets/status.c
@@ -179,7 +179,7 @@ static void draw_bottom(lv_obj_t *widget, lv_color_t cbuf[], const struct status
     lv_canvas_draw_rect(canvas, 0, 0, CANVAS_SIZE, CANVAS_SIZE, &rect_black_dsc);
 
     // Draw layer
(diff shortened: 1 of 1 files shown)


amacleod:
I tested this on my Typeractive Lily58 Choc wireless with nice!view displays.

petejohanson (approved):
Thanks!

PR #2543, opened 2024-10-08
docs: Remove download for settings_reset and note local build

caksoylar:
The downloaded `settings_reset` firmware frequently doesn't work for users, whether it is because they are picking it up from too old builds, or due to another issue. It can also be tricky to find the build for the board in question, especially for no-shield keyboards. This can lead to flashing the wrong firmware to corrupt partitions, although this hasn't seemed to have happened yet.

This PR removes the download option and replaces it with concise instructions for local builds which is not documented and has been asked before.


docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -41,14 +23,9 @@ This procedure will erase all settings, such as Bluetooth profiles, output selec
...
-### Acquiring a Reset UF2
-
-First, acquire the reset UF2 image file with one of the following options:
+### Building a Reset UF2
 
-<Uf2Tabs>
-<TabItem value="build">
-
-Find the `build.yaml` file in your `zmk-config` folder and add an additional settings reset build for the board used by your split keyboard. For example assuming that the config repo is setup for nice!nano v2 with Corne, append the `settings_reset` shield to the `build.yaml` file as follows:
+If you are using GitHub Actions to build your firmware as described in the [user setup](../user-setup.mdx), find the `build.yaml` file in your `zmk-config` folder and add an additional settings reset build for the board used by your split keyboard. For example assuming that the config repo is setup for nice!nano v2 with Corne, append the `settings_reset` shield to the `build.yaml` file as follows:
```
Nick-Munnich, line 28:
```suggestion
If you are using GitHub Actions to build your firmware as described in the [user setup](../user-setup.mdx), find the `build.yaml` file in your `zmk-config` folder and add an additional settings reset build for the board used by your split keyboard. For example, assuming that the config repo is setup for nice!nano v2 with Corne, append the `settings_reset` shield to the `build.yaml` file as follows:
```

docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -41,14 +23,9 @@ This procedure will erase all settings, such as Bluetooth profiles, output selec
 
 :::
 
-### Acquiring a Reset UF2
-
-First, acquire the reset UF2 image file with one of the following options:
+### Building a Reset UF2
```
Nick-Munnich, line 26: I've never been a big fan of the UF2 being in this header. "Building Reset Firmware" maybe?

docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -60,28 +37,18 @@ include:
     shield: settings_reset
 ```
 
-Save the file, commit the changes and push them to GitHub. Download the new firmware zip file build by the latest GitHub Actions job. In it you will find an additional `settings_reset` UF2 image file.
+If you are building for an onboard controller keyboard that doesn't use a shield, you can add a new `shield:` field with its `board:` entry.
```
Nick-Munnich, line 40:
```suggestion
If you are building for an onboard controller keyboard that doesn't use a shield, add the `shield: settings_reset` field to its `board:` entry.
```

docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -60,28 +37,18 @@ include:
...
-
-<TabItem value="download">
-1. [Open the `Build` workflow](https://github.com/zmkfirmware/zmk/actions/workflows/build.yml?query=event%3Apush+branch%3Amain+is%3Asuccess) from the `Actions` tab of the ZMK GitHub repository.
-1. Find one of the results for which the `core-coverage` job ran successfully, indicated by a green checkmark in the "core-coverage" bubble like the image example below.
-1. From the next page under "Artifacts", download and unzip the `<board_name>-settings_reset-zmk` zip file for the UF2 image.
-
-| ![Successful core-coverage Job](../../docs/assets/troubleshooting/splitpairing/corecoverage.png) |
-| :----------------------------------------------------------------------------------------------: |
-|    An example of a successful core-coverage job which will produce a settings_reset firmware.    |
+Save the file, commit the changes and push them to GitHub. Download the new firmware zip file build by the latest GitHub Actions job. In it you will find an additional `settings_reset` UF2 image file.
```
Nick-Munnich, line 42:
```suggestion
Save the file, commit the changes and push them to GitHub. Download the new `firmware.zip` file built by the latest GitHub Actions job. In it you will find an additional `settings_reset` UF2 image file.
```

docs/docs/troubleshooting/connection-issues.mdx
```diff
@@ -60,28 +37,18 @@ include:
...
 
-</TabItem>
-</Uf2Tabs>
+If you use a [local development environment](../development/local-toolchain/setup/index.md) to build firmware instead of GitHub Actions, you can replace your `-DSHIELD=` argument value with `settings_reset` in the [build command](../development/local-toolchain/build-flash.mdx), or add one if the keyboard doesn't normally use one.
```
Nick-Munnich, line 44:
```suggestion
If you use a [local development environment](../development/local-toolchain/setup/index.md) to build firmware instead of GitHub Actions, pass the `-DSHIELD=settings_reset` argument when [building](../development/local-toolchain/build-flash.mdx), omitting all other `-DSHIELD` arguments.
```

docs/docs/troubleshooting/connection-issues.mdx
caksoylar, line 26:
```suggestion
### Building a Reset Firmware
```

docs/docs/troubleshooting/connection-issues.mdx
caksoylar:
```suggestion
If you are building for an onboard controller keyboard that doesn't use a shield, add a new `shield:` field to its `board:` entry.
```
I think it's better to keep "a new" rather than "the" to signify the previous non-existence of the field?

docs/docs/troubleshooting/connection-issues.mdx
caksoylar: I think the zip file won't be literally called `firmware.zip`, hence the generic "the new firmware zip file" phrasing here. We should do "file that is built" rather than "file build".

Nick-Munnich (approved):
Happy with both of your suggested adjustments to my comments, approving in advance of you committing those.

PR #2544, opened 2024-10-09
refactor: Give position map children labels.

petejohanson:
* To make is easier to reuse layouts and override the position maps for them, add labels to all the position map children in our shared layouts.


petejohanson:
> I think it'd be nice to stay consistent with the naming, I propose `<layout_name>_posmap` which adds `pos` prefix to make it less ambiguous, compared to `<layout_name>_map` ones you have (Corne ones seem to be the exception).

@caksoylar Done.

PR #2546, opened 2024-10-10
feat(ble): Add the ability to enable and disable advertising

ReFil:
Currently advertising cannot be disabled, In some circumstances it may be desirable to disable advertising to the host e.g. seeking consistency between an endpoint switch and percieved board behaviour.

This adds 3 new options to &bt, ADV_ON, ADV_OFF and ADV_TOG as well as the metadata and documentation for the behaviour. `ble.c` has been updated with a static flag that gets checked in `update_advertising()` and doesn't permit any adveritising state other than `ZMK_ADV_NONE` unless it's set. it's defaulted to true so standard board behavior shouldn't change. 

The ability to get the current adverising state has also been added, it could have utility in custom display widgets. 

Tested on a development board, I've noticed some hosts cache the advertising info so it still shows up as a connectible device if the switch is used whilst the bluetooth devices menu is open, but it still cannot be connected to. Using an app like nRF connect or a different host it can be seen that the device properly stops advertising

### DIFF
--- a/app/include/dt-bindings/zmk/bt.h
+++ b/app/include/dt-bindings/zmk/bt.h
@@ -10,6 +10,9 @@
 #define BT_SEL_CMD 3
 #define BT_CLR_ALL_CMD 4
 #define BT_DISC_CMD 5
+#define BT_ADV_OFF_CMD 6
+#define BT_ADV_ON_CMD 7
+#define BT_ADV_TOG_CMD 8
 
 /*
 Note: Some future commands will include additional parameters, so we
@@ -22,3 +25,6 @@ defines these aliases up front.
 #define BT_SEL BT_SEL_CMD
 #define BT_CLR_ALL BT_CLR_ALL_CMD 0
 #define BT_DISC BT_DISC_CMD
+#define BT_ADV_OFF BT_ADV_OFF_CMD 0
+#define BT_ADV_ON BT_ADV_ON_CMD 0
+#define BT_ADV_TOG BT_ADV_TOG_CMD 0
--- a/app/include/zmk/ble.h
+++ b/app/include/zmk/ble.h
@@ -19,6 +19,12 @@
 #define ZMK_BLE_PROFILE_COUNT CONFIG_BT_MAX_PAIRED
 #endif
 
+enum advertising_type {
+    ZMK_ADV_NONE,
+    ZMK_ADV_DIR,
+    ZMK_ADV_CONN,
+};
+
 void zmk_ble_clear_bonds(void);
 int zmk_ble_prof_next(void);
 int zmk_ble_prof_prev(void);
@@ -43,6 +49,9 @@ char *zmk_ble_active_profile_name(void);
 int zmk_ble_unpair_all(void);
 
 int zmk_ble_set_device_name(char *name);
+
+void zmk_ble_adv_enabled_set(bool adv_enabled);
+bool zmk_ble_adv_enabled_get(void);
 
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
 int zmk_ble_put_peripheral_addr(const bt_addr_le_t *addr);
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -47,11 +47,7 @@ RING_BUF_DECLARE(passkey_entries, PASSKEY_DIGITS);
 
 #endif /* IS_ENABLED(CONFIG_ZMK_BLE_PASSKEY_ENTRY) */
 
-enum advertising_type {
-    ZMK_ADV_NONE,
-    ZMK_ADV_DIR,
-    ZMK_ADV_CONN,
-} advertising_status;
+enum advertising_type advertising_status;
 
 #define CURR_ADV(adv) (adv << 4)
 
@@ -62,6 +58,8 @@ enum advertising_type {
 
 static struct zmk_ble_profile profiles[ZMK_BLE_PROFILE_COUNT];
 static uint8_t active_profile;
+
+static bool permit_adv = true;
 
 #define DEVICE_NAME CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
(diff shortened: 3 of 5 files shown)


Nick-Munnich:
Regarding docs I think there might end up being some confusion when people see "This turns on advertising", use the behavior expecting to be able to connect a new device yet nothing happens because they haven't selected an unused profile.

ReFil:
I see your point, it would still advertise if it's in an occupied profile but wouldn't pair and work properly. Perhaps it would be better worded as "Permits advertising (note that other conditions must be met for the device to be visible)"

ReFil:
Functionally there's not really any difference I suppose

Nick-Munnich:
I think renaming and rephrasing it appropriately would be both clearer and make it usable by non-tech people, then. Sometimes a user might want to just turn off BT after all.

ReFil:
Hmm, I'm not sure about describing it as turning off Bluetooth when it doesn't do that. My concern is users may wonder if it affects split connections if it's described as shutting off Bluetooth

docs/docs/keymaps/behaviors/bluetooth.md
```diff
@@ -46,6 +46,9 @@ Here is a table describing the command for each define:
 | `BT_PRV`     | Switch to the previous profile, cycling through to the last one when the beginning is reached.                                                                                     |
 | `BT_SEL`     | Select the 0-indexed profile by number; must include a number as an argument in the keymap to work correctly, e.g. `BT_SEL 0`.                                                     |
 | `BT_DISC`    | Disconnect from the 0-indexed profile by number, if it's currently connected and inactive; must include a number as an argument in the keymap to work correctly, e.g. `BT_DISC 0`. |
+| `ADV_OFF`    | Disconnect from all profiles and disable advertising (The device will not show up as connectable on host devices)                                                                  |
+| `ADV_ON`     | Enable Bluetooth advertising, the device will show up as connectable                                                                                                               |
```
Nick-Munnich, line 50:
```suggestion
| `ADV_OFF`    | Disable advertising and disconnect from all profiles                                                                |
| `ADV_ON`     | Enable advertising and attempt to reconnect to profiles                                                                                                   |
```
Maybe like this?

nmunnich (commented):
True, so that's not an accurate description. Hmm.

docs/docs/keymaps/behaviors/bluetooth.md
ReFil: Yep i think that's better, thank you

nmunnich (approved):

app/include/dt-bindings/zmk/bt.h
```diff
@@ -22,3 +25,6 @@ defines these aliases up front.
 #define BT_SEL BT_SEL_CMD
 #define BT_CLR_ALL BT_CLR_ALL_CMD 0
 #define BT_DISC BT_DISC_CMD
+#define ADV_OFF BT_ADV_OFF_CMD 0
+#define ADV_ON BT_ADV_ON_CMD 0
+#define ADV_TOG BT_ADV_TOG_CMD 0
```
petejohanson, line 30: I'd rather these have the `BT_` prefix.

app/include/zmk/ble.h
```diff
@@ -41,6 +47,9 @@ int zmk_ble_unpair_all(void);
 
 int zmk_ble_set_device_name(char *name);
 
+void zmk_ble_adv_mode_set(bool mode);
+bool zmk_ble_adv_mode_get(void);
```
petejohanson, line 51: "mode" but using a boolean here feels off. Either return a struct for some mode , or use "adv_enabled" wording, IMHO.

petejohanson (changes requested):
Thanks for working on this!

docs/docs/keymaps/behaviors/bluetooth.md
```diff
@@ -46,6 +46,9 @@ Here is a table describing the command for each define:
 | `BT_PRV`     | Switch to the previous profile, cycling through to the last one when the beginning is reached.                                                                                     |
 | `BT_SEL`     | Select the 0-indexed profile by number; must include a number as an argument in the keymap to work correctly, e.g. `BT_SEL 0`.                                                     |
 | `BT_DISC`    | Disconnect from the 0-indexed profile by number, if it's currently connected and inactive; must include a number as an argument in the keymap to work correctly, e.g. `BT_DISC 0`. |
+| `ADV_OFF`    | Disable advertising and disconnect from all profiles                                                                                                                               |
+| `ADV_ON`     | Enable advertising and attempt to reconnect to profiles                                                                                                                            |
+| `ADV_TOG`    | Toggle advertising on and off                                                                                                                                                      |
```
caksoylar, line 52: I think it might be worth adding a small note here, saying that advertising is on by default.

I see this pattern in some users where they assume they need to set a setting for something to work, just because they see it exists.

app/include/zmk/ble.h
ReFil: The functions are now called `zmk_ble_adv_enabled_set(bool adv_enabled)` and `zmk_ble_adv_enabled_get()`

caksoylar (approved):

ReFil:
Rebased and ready for review

dapetcu21:
An improvement over this that I would personally love to see would be to persist the setting across reboots, similar to how bt profiles or output switching works.

kurtis-lew:
Cherry-picked on top of #3060, testing on Electronic Materials Office Ltd. Altar I's output selection SP3T switch.
Can confirm that **ADV_ON** and **ADV_OFF** work exactly as expected. Thanks for working on this!

app/src/ble.c
```diff
@@ -63,6 +59,8 @@ enum advertising_type {
 static struct zmk_ble_profile profiles[ZMK_BLE_PROFILE_COUNT];
 static uint8_t active_profile;
 
+static bool permit_adv = true;
```
yuannan, line 62: I would add a comment here to state this this is the default state. I personally have this disabled by default.

docs/docs/keymaps/behaviors/bluetooth.md
```diff
@@ -46,12 +46,19 @@ Here is a table describing the command for each define:
...
+| `BT_ADV_ON`  | Enable advertising and attempt to reconnect to profiles                                                                                                                            |
+| `BT_ADV_TOG` | Toggle advertising on and off                                                                                                                                                      |
 
 :::note[Selected profile persistence]
 The profile that is selected by the `BT_SEL`/`BT_PRV`/`BT_NXT` actions will be saved to flash storage and hence persist across restarts and firmware flashes.
 However it will only be saved after [`CONFIG_ZMK_SETTINGS_SAVE_DEBOUNCE`](../../config/system.md#general) milliseconds in order to reduce potential wear on the flash memory.
 :::
 
+:::note[Advertising]
+Bluetooth advertising is enabled in all situations by default, and doesn't have to be explicitly enabled via the bindings above.
```
yuannan, line 59: I would also add a comment here to state that it can be disabled by default.

yuannan (approved):
Works on my glove80!

In case anyone wants to add it to their configs as well you just have to add `$bt BT_ADV_TOG` (or another command) as a key into your keymap.

yuannan:
@petejohanson is there any changes you would like made to this before it can be merged? I'm happy to help out on this commit as well :)

PR #2554, opened 2024-10-15
feat: Added a flag to key press behavior to allow it to be inverted

nmunnich:
Currently if two of the same binding are pressed and then released, the first release will be treated as both keys being released. 

This adds a new optional flag to the key press behavior `invert-if-active` causing the behavior to invert its function if the binding is already pressed:
- If pressed and inverting key press is pressed, a release is triggered
- If released and inverting key press is released, a press is triggered

This flag resolves this inconsistent behavior, but is marked as optional for backwards compatibility and as it is less intuitive.

It is also a quite useful interaction, in my opinion:
- Quickly interrupting shift or a similar modifier
- Toggle W on and press W to stay in place for reduced key holding when WASD gaming

EDIT: I have a feeling that this may interact poorly with combos. This is a note telling me to test that interaction in the future when I get around to it.


petejohanson:
I am on the fence on this... I'm not convinced this is the correct solution for the issue at hand.  Additionally, this seems potentially useful for some scenarios, but feels *really* counter intuitive for a basic behavior like key press. I'm not sure I love this, versus a different behavior that does this. Need to ponder this.

Nick-Munnich:
> I am on the fence on this... I'm not convinced this is the correct solution for the issue at hand. Additionally, this seems potentially useful for some scenarios, but feels _really_ counter intuitive for a basic behavior like key press. I'm not sure I love this, versus a different behavior that does this. Need to ponder this.

Yeah, I was uncertain about it too. Originally I thought about making this a separate behavior in a module because I just wanted the functionality, but quickly realised that it's essentially just a slight variant of the standard `&kp` behavior. That put it more into the realm of "add a new flag" for me. Then while working on #2555 I stumbled across the weird interaction, and realised that this approach would also serve to bypass the interaction, in a way.

There are alternatives to resolving the interaction, which may be something to look into, but I would disagree with making the function a new behavior. I do agree that the *description* is very counterintuitive, but I find it very intuitive to use in practice, if that counts for anything.

nmunnich:
Closing this as identifying the correct solution to this edge case is unclear, but I think it won't end up being this one.

PR #2555, opened 2024-10-15
feat: Added `toggle-mode`, allowing toggle-on and toggle-off

### LINKED ISSUE #2523: Layer behaviors to only enable or disable a layer instead of toggling
The ability to only turn on or off a layer with dedicated behaviors would make complex layer setups and combos easier. Let me give a specific example:

I regularly work and switch between two computers. One is connected via USB and uses Windows, and one is a Mac I connect to with Bluetooth. To make my layers consistent across the operating systems, I've defined some patches to my basic layers to activate on a flag layer. I always activate this layer when switching to prefer bluetooth connections, and disable it when going back to USB. But the way I need to do it now, is with a layer toggle. Sometimes I mistype, and end up with the wrong layer and output device combo.

Instead of using a layer toggle, I'd want to be sure I am explicitly activating or deactivating a layer. This would also make it usable in a macro, where I can put both actions behind a single key.

Any thoughts? Maybe I've just missed something obvious :sweat_smile: and if that's the case, please do inform me. Many thanks for considering this!

Nick-Munnich:
Added `toggle-mode` flag to both the key toggle and toggle layer behaviors. Toggle mode can be set to
- "on"
- "off"
- "flip" (default when not present)

Tested locally with basic functionality. I don't see an edge case where this might cause issues.

Note that if key toggle on is spammed, it will repeatedly send over the "press" code. 

Resolves #2523.

### DIFF
--- a/app/dts/bindings/behaviors/zmk,behavior-toggle-layer.yaml
+++ b/app/dts/bindings/behaviors/zmk,behavior-toggle-layer.yaml
@@ -6,3 +6,13 @@ description: Toggle Layer
 compatible: "zmk,behavior-toggle-layer"
 
 include: one_param.yaml
+
+properties:
+  toggle-mode:
+    type: string
+    required: false
+    default: "flip"
+    enum:
+      - "on"
+      - "off"
+      - "flip"
--- a/app/src/behaviors/behavior_toggle_layer.c
+++ b/app/src/behaviors/behavior_toggle_layer.c
@@ -17,15 +17,33 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
 
-struct behavior_tog_config {};
-struct behavior_tog_data {};
+enum toggle_mode {
+    ON,
+    OFF,
+    FLIP,
+};
+
+struct behavior_tog_config {
+    enum toggle_mode toggle_mode;
+};
 
 static int behavior_tog_init(const struct device *dev) { return 0; };
 
 static int tog_keymap_binding_pressed(struct zmk_behavior_binding *binding,
                                       struct zmk_behavior_binding_event event) {
     LOG_DBG("position %d layer %d", event.position, binding->param1);
-    return zmk_keymap_layer_toggle(binding->param1);
+
+    const struct behavior_tog_config *cfg = zmk_behavior_get_binding(binding->behavior_dev)->config;
+    switch (cfg->toggle_mode) {
+    case ON:
+        return zmk_keymap_layer_activate(binding->param1);
+    case OFF:
+        return zmk_keymap_layer_deactivate(binding->param1);
+    case FLIP:
+        return zmk_keymap_layer_toggle(binding->param1);
+    default:
+        return -ENOTSUP;
+    };
 }
 
 static int tog_keymap_binding_released(struct zmk_behavior_binding *binding,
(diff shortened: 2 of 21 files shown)


petejohanson (commented):
Can we please add some tests that exercise this new functionality?

Nick-Munnich:
> Can we please add some tests that exercise this new functionality?

Yeah, I was just thinking that. Going to do so (and fix an oopsie I noticed) and then undraft.

docs/docs/keymaps/behaviors/key-toggle.md
```diff
@@ -24,3 +24,25 @@ Example:
...
+            compatible = "zmk,behavior-key-toggle";
+            #binding-cells = <1>;
+            display-name = "Key Toggle On";
+            toggle-mode = "on";
+        };
+    };
+};
+```
+
+You can then use `&kt_on` in place of `&kt` whenever you wish to only toggle a key on, and not toggle it off. The `"off"` version is analogous.
```
caksoylar, line 48:
```suggestion
You can then use `&kt_on` in place of `&kt` whenever you wish to only toggle a key on, and not toggle it off. An `"off"` version of the behavior can be defined similarly.
```

docs/docs/keymaps/behaviors/layers.md
```diff
@@ -116,6 +116,28 @@ Example:
...
+            compatible = "zmk,behavior-toggle-layer";
+            #binding-cells = <1>;
+            display-name = "Toggle Layer On";
+            toggle-mode = "on";
+        };
+    };
+};
+```
+
+You can then use `&tog_on` in place of `&tog` whenever you wish to only toggle a layer on, and not toggle it off. The `"off"` version is analogous.
```
caksoylar, line 139:
```suggestion
You can then use `&tog_on` in place of `&tog` whenever you wish to only toggle a layer on, and not toggle it off. An `"off"` version of the behavior can be defined similarly.
```

app/dts/bindings/behaviors/zmk,behavior-toggle-layer.yaml
```diff
@@ -6,3 +6,13 @@ description: Toggle Layer
...
+
+properties:
+  toggle-mode:
+    type: string
+    required: false
+    default: "on-off"
+    enum:
+      - "on"
+      - "off"
+      - "on-off"
```
caksoylar, line 18: I think a term like "invert" might be a better name than "on-off". The latter is a bit ambiguous that it "toggles"/flips the state.

app/src/behaviors/behavior_toggle_layer.c
```diff
@@ -17,22 +17,40 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+        return -ENOTSUP;
+    };
 }
 
 static int tog_keymap_binding_released(struct zmk_behavior_binding *binding,
                                        struct zmk_behavior_binding_event event) {
     LOG_DBG("position %d layer %d", event.position, binding->param1);
     return ZMK_BEHAVIOR_OPAQUE;
-}
+};
```
caksoylar, line 53: Typo?

caksoylar (commented):
https://zmk.dev/docs/config/behaviors needs to be updated as well.

app/dts/bindings/behaviors/zmk,behavior-toggle-layer.yaml
Nick-Munnich: I think invert could end up being confusing too. How about "switch"?

caksoylar: I'd say "toggle" is the best but in this case it might not be a good idea 😄 Switch sounds OK too, or "flip".

Nick-Munnich: I like flip, made that change.

caksoylar (approved):
LGTM on docs, and source code.

petejohanson (approved):
LGTM! Thanks!

PR #2564, opened 2024-10-17
fix(boards): Select hummingbird physical layout

Nick-Munnich:
The physical layout was defined previously, but never selected in the chosen node. I did not verify the correctness of the physical layout and matrix transform.


petejohanson (commented):
When no explicit layout is chosen, the first layout defined is chosen automatically. Since there's only one layout for this shield, it's picked automatically.

Not sure the value in being explicit when it's not necessary. On the fence on this.

Nick-Munnich:
I think if we're not explicit about it everywhere, then this needs to be documented. I'd prefer to be explicit everywhere to avoid needing to document it, but I'm also always biased towards "one way to do it".

Nick-Munnich:
Closing as it is part of #2582 and #2595.

PR #2565, opened 2024-10-17
fix: Optimize layout changes by doing runtime mapping

petejohanson:
* To avoid tons of migration, extra flash writes, etc, we keep the keymaps and settings using a key position index that's tied to the stock layout, and at runtime mapping key positions as needed.

### DIFF
--- a/app/include/zmk/physical_layouts.h
+++ b/app/include/zmk/physical_layouts.h
@@ -49,3 +49,12 @@ int zmk_physical_layouts_revert_selected(void);
 
 int zmk_physical_layouts_get_position_map(uint8_t source, uint8_t dest, size_t map_size,
                                           uint32_t map[map_size]);
+
+/**
+ * @brief Get a pointer to a position map array for mapping a key position in the selected
+ *        physical layout to the stock/chosen physical layout
+ *
+ * @retval a negative errno value in the case of errors
+ * @retval a positive length of the position map array that map is updated to point to.
+ */
+int zmk_physical_layouts_get_selected_to_stock_position_map(uint32_t const **map);
--- a/app/src/keymap.c
+++ b/app/src/keymap.c
@@ -13,6 +13,7 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <zmk/stdlib.h>
 #include <zmk/behavior.h>
 #include <zmk/keymap.h>
+#include <zmk/physical_layouts.h>
 #include <zmk/matrix.h>
 #include <zmk/sensors.h>
 #include <zmk/virtual_key_position.h>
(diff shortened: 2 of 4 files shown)


app/src/keymap.c
```diff
@@ -230,7 +231,16 @@ zmk_keymap_get_layer_binding_at_idx(zmk_keymap_layer_id_t layer_id, uint8_t bind
 
     ASSERT_LAYER_VAL(layer_id, NULL)
 
-    return &zmk_keymap[layer_id][binding_idx];
+    uint32_t *pos_map;
+    int ret;
+
+    ret = zmk_physical_layouts_get_selected_to_stock_position_map(&pos_map);
```
joelspadin, line 237:
```suggestion
    int ret = zmk_physical_layouts_get_selected_to_stock_position_map(&pos_map);
```
Can simplify this as is done later in the file

app/src/keymap.c
```diff
@@ -247,12 +257,27 @@ int zmk_keymap_set_layer_binding_at_idx(zmk_keymap_layer_id_t layer_id, uint8_t
...
     ASSERT_LAYER_VAL(layer_id, -EINVAL)
 
+    uint32_t *pos_map;
+    int ret = zmk_physical_layouts_get_selected_to_stock_position_map(&pos_map);
+    if (ret < 0) {
+        LOG_WRN("Failed to get the mapping to determine where to set the binding (%d)", ret);
+        return ret;
+    }
+
+    uint32_t storage_binding_idx = pos_map[binding_idx];
```
joelspadin, line 267: `ret` here is now the size of `pos_map`. Do we need to check that `binding_index` is in bounds?

Also, do we need to check that `storage_binding_index` is in bounds for `zmk_keymap`?

(If the answer to these is "yes, but it's an error in the code rather than a user configuration error if they're out of bounds", then I would probably just use asserts rather than returning an error.)

app/include/zmk/physical_layouts.h
```diff
@@ -49,3 +49,5 @@ int zmk_physical_layouts_revert_selected(void);
 
 int zmk_physical_layouts_get_position_map(uint8_t source, uint8_t dest, size_t map_size,
                                           uint32_t map[map_size]);
+
+int zmk_physical_layouts_get_selected_to_stock_position_map(uint32_t **map);
```
joelspadin, line 53: It would be good to document that the value written to `*map` is an array, and the return value is its size.

Also, this returns a pointer to a mutable array, but the only changes to this data I'm seeing are done to the static variable, not the return value of this function. Should this return a pointer to const pointer?

app/src/keymap.c
```diff
@@ -247,12 +257,27 @@ int zmk_keymap_set_layer_binding_at_idx(zmk_keymap_layer_id_t layer_id, uint8_t
 
     ASSERT_LAYER_VAL(layer_id, -EINVAL)
 
+    uint32_t *pos_map;
+    int ret = zmk_physical_layouts_get_selected_to_stock_position_map(&pos_map);
+    if (ret < 0) {
```
joelspadin, line 262: This case isn't currently possible. Will it ever be possible?

petejohanson: Rather than change the API later, I wanted to ensure we guard now in case things change later with how position maps get created.

app/include/zmk/physical_layouts.h
petejohanson: Added

app/src/keymap.c
petejohanson: Added checks. Thanks.

petejohanson:
Have tested this on a 75% with a split backspace layout, and on my revxlp, no issues. Also had testing on BakenekoGo with two layouts.

PR #2572, opened 2024-10-19
feat(core): Optionally disable automatic endpoint fallback

ReFil:
By default if only one endpoint is connected but the other is selected, the connected endpoint will be chosen as a fallback, for boards with physical endpoint selection slide switches this can lead to confusing behaviour where one is selected but the other is used.

This adds a kConfig option `CONFIG_ZMK_ENDPOINT_DISABLE_FALLBACK` that, when enabled, switches the board to a different endpoint selection system that will only return the preferred transport or anew transport `ZMK_TRANSPORT_NONE` that just swallows any HID data.

Tested with various combinations of USB and BLE and it seems to work as intended


caksoylar:
Can we add a link to https://zmk.dev/docs/keymaps/behaviors/outputs pointing to this? Perhaps an additional paragraph at the end of the summary section.

app/src/endpoints.c
```diff
@@ -118,6 +132,24 @@ int zmk_endpoints_toggle_transport(void) {
 
 struct zmk_endpoint_instance zmk_endpoints_selected(void) { return current_instance; }
 
+struct zmk_endpoint_instance zmk_endpoints_preferred(void) {
```
joelspadin, line 135: This looks like mostly duplicated code with `get_selected_instance()`. We should pull the common logic out into a separate function that both things can call so we don't keep the two functions in sync.

app/src/endpoints.c
```diff
@@ -282,6 +320,17 @@ static bool is_ble_ready(void) {
...
 
 static enum zmk_transport get_selected_transport(void) {
+#if IS_ENABLED(CONFIG_ZMK_ENDPOINT_DISABLE_FALLBACK)
+    switch (preferred_transport) {
+    case ZMK_TRANSPORT_BLE:
+        return is_ble_ready() ? ZMK_TRANSPORT_BLE : ZMK_TRANSPORT_NONE;
+    case ZMK_TRANSPORT_USB:
+        return is_usb_ready() ? ZMK_TRANSPORT_USB : ZMK_TRANSPORT_NONE;
+
+    default:
```
joelspadin, line 330: I'm not a fan of default cases with no logic. If you omit the default case, then the compiler can verify that you have a case for every enum value. Then, if you add a new enum value later but forget to update this code, it will give a diagnostic. 

I suggest having a case for ZMK_TRANSPORT_NONE which just returns ZMK_TRANSPORT_NONE, and then treating it as an error if none of the cases are hit. Maybe even split this into two completely separate implementations of the function like this?

```c
#if IS_ENABLED(CONFIG_ZMK_ENDPOINT_DISABLE_FALLBACK)

static enum zmk_transport get_selected_transport(void) {
    switch (preferred_transport) {
        case ZMK_TRANSPORT_BLE: 
            return ...
        case ZMK_TRANSPORT_USB: 
            return ...
        case ZMK_TRANSPORT_NONE: 
            return ZMK_TRANSPORT_NONE;
    }

    LOG_ERR("Unknown transport %d", preferred_transport);
    return ZMK_TRANSPORT_NONE;
}

#else

// original implementation

#endif
```

app/src/endpoints.c
ReFil: I've added a new function `endpoints_add_extra_data()` which takes the transport (whether selected or preferred), adds the ble profile on if applicable and returns the full endpoint instance

app/src/endpoints.c
ReFil: Both of these suggestions have now been implemented

app/src/endpoints.c
```diff
@@ -73,13 +80,17 @@ int zmk_endpoint_instance_to_str(struct zmk_endpoint_instance endpoint, char *st
...
+        return snprintf(str, len, "None");
+
     default:
         return snprintf(str, len, "Invalid");
     }
 }
 
 #define INSTANCE_INDEX_OFFSET_USB 0
 #define INSTANCE_INDEX_OFFSET_BLE ZMK_ENDPOINT_USB_COUNT
+#define INSTANCE_INDEX_OFFSET_NONE ZMK_ENDPOINT_USB_COUNT + ZMK_ENDPOINT_BLE_COUNT
```
joelspadin, line 93: I think it makes more sense for "none" to be index 0. (Endpoints instances are explicitly documented as something that can change between firmware versions, so there's no need to maintain USB as 0)

app/include/zmk/endpoints.h
```diff
@@ -33,7 +33,7 @@
  * Note that this value may change between firmware versions, so it should not
  * be used in any persistent storage.
  */
-#define ZMK_ENDPOINT_COUNT (ZMK_ENDPOINT_USB_COUNT + ZMK_ENDPOINT_BLE_COUNT)
+#define ZMK_ENDPOINT_COUNT (ZMK_ENDPOINT_USB_COUNT + ZMK_ENDPOINT_BLE_COUNT + 1)
```
joelspadin, line 36: I think it would be good to use

```c
#define ZMK_ENDPOINT_NONE_COUNT 1
...
#define ZMK_ENDPOINT_COUNT (ZMK_ENDPOINT_NONE_COUNT + ZMK_ENDPOINT_USB_COUNT + ZMK_ENDPOINT_BLE_COUNT)
```
since otherwise it's not obvious what the `+ 1` is for.

app/src/endpoints.c
```diff
@@ -118,6 +132,27 @@ int zmk_endpoints_toggle_transport(void) {
 
 struct zmk_endpoint_instance zmk_endpoints_selected(void) { return current_instance; }
 
+static struct zmk_endpoint_instance endpoints_add_extra_data(enum zmk_transport transport) {
```
joelspadin, line 135: "add extra data" makes me think that the function would take a struct as an input and populate some fields on it, but this generates the entire struct. Maybe a name more like "get instance from transport" would be clearer?

app/src/endpoints.c
```diff
@@ -23,8 +23,12 @@
 #include <zephyr/logging/log.h>
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
+#if IS_ENABLED(CONFIG_ZMK_ENDPOINT_DISABLE_FALLBACK)
+#define DEFAULT_TRANSPORT ZMK_TRANSPORT_NONE
```
joelspadin, line 27: Unless I'm missing somewhere that this gets used, this is unused when CONFIG_ZMK_ENDPOINT_DISABLE_FALLBACK is enabled, so it doesn't need to be defined.

app/src/endpoints.c
```diff
@@ -59,6 +63,9 @@ bool zmk_endpoint_instance_eq(struct zmk_endpoint_instance a, struct zmk_endpoin
 
     case ZMK_TRANSPORT_BLE:
         return a.ble.profile_index == b.ble.profile_index;
+
+    case ZMK_TRANSPORT_NONE:
```
joelspadin, line 67: Super nitpick: not really a big deal, but it would be nice for all of the switches to have the cases in the same order as the enum values are defined.

joelspadin (changes requested):

app/src/endpoints.c
ReFil: Good point, it's been removed

app/src/endpoints.c
ReFil: Name has been changed to `get_isntance_from_transport()`

joelspadin (changes requested):
I noticed one more issue. Otherwise, this looks good to me (though I haven't personally tested it).

ReFil:
@joelspadin sorry to be a pain, not sure if GitHub is playing up or something but I can't seem to see what changes you requested?

app/src/endpoints.c
```diff
@@ -78,11 +86,15 @@ int zmk_endpoint_instance_to_str(struct zmk_endpoint_instance endpoint, char *st
     }
 }
 
-#define INSTANCE_INDEX_OFFSET_USB 0
-#define INSTANCE_INDEX_OFFSET_BLE ZMK_ENDPOINT_USB_COUNT
+#define INSTANCE_INDEX_OFFSET_NONE 0
+#define INSTANCE_INDEX_OFFSET_USB ZMK_ENDPOINT_NONE_COUNT
+#define INSTANCE_INDEX_OFFSET_BLE ZMK_ENDPOINT_NONE_COUNT + ZMK_ENDPOINT_USB_COUNT
```
joelspadin, line 91:
```suggestion
#define INSTANCE_INDEX_OFFSET_BLE (ZMK_ENDPOINT_NONE_COUNT + ZMK_ENDPOINT_USB_COUNT)
```
This needs to be parenthesized, or you might get unexpected behaviors like
```c
INSTANCE_INDEX_OFFSET_BLE * 2  // I expect 2 * 2 = 4 but...
ZMK_ENDPOINT_NONE_COUNT + ZMK_ENDPOINT_USB_COUNT * 2  // 1 + 1 * 2 = 3
```

Better yet, each item can be written as the previous offset plus the previous count so the size of the expression doesn't grow with each item:

```c
#define INSTANCE_INDEX_OFFSET_NONE 0
#define INSTANCE_INDEX_OFFSET_USB (INSTANCE_INDEX_OFFSET_NONE + ZMK_ENDPOINT_NONE_COUNT)
#define INSTANCE_INDEX_OFFSET_BLE (INSTANCE_INDEX_OFFSET_USB + ZMK_ENDPOINT_USB_COUNT)
#define INSTANCE_INDEX_OFFSET_FOO (INSTANCE_INDEX_OFFSET_BLE + ZMK_ENDPOINT_BLE_COUNT)
...
```

joelspadin:
> @joelspadin sorry to be a pain, not sure if GitHub is playing up or something but I can't seem to see what changes you requested?

No idea where that comment went. I've added it again.

joelspadin (commented):
This looks good to me now, but @petejohanson should probably take a look too.

ReFil:
Rebased and ready for review

snoyer:
I've been testing this and it's been working as expected when it comes to selecting the endpoints :+1:

However it seems like there are cases where it's not possible to notice the change of preferred endpoint using events.
For example when switching between a non-connected BLE profile and non-HID USB:
- calling `zmk_endpoints_preferred()` will indeed return `BLE` or `USB`
- but no `zmk_endpoint_changed` event is raised

It would be nice to be able to listen to the change of preferred endpoint to reflect it on custom displays. (I don't know if that means a new event or just firing the existing `zmk_endpoint_changed` more)

caksoylar (approved):
Docs LGTM, I think it'd be nice to have the feature in.

kurtis-lew:
Cherry-picked on top of #3060, testing on Electronic Materials Office Ltd. Altar I. Feature works swimmingly, just had to remember for `CONFIG_ZMK_SETTINGS_SAVE_DEBOUNCE` to elapse between power cycles. Thanks for working on this!

snoyer:
are the changes to `keymap.h` and `keymap.c` in the latest commit related to this PR?

ReFil:
Sorry, they weren't meant to be there, should all be sorted and rebased now

petejohanson:
Needs rebasing now that #3140 is in. Thanks!

PR #2574, opened 2024-10-20
feat(boards): Add numpad layouts

joelspadin:
Added physical layouts for the following variants of numpads:

- With and without extra top row
- 2U plus key or 1U plus and backspace keys
- 2U 0 key or 1U 0 and 00 keys
- Full 1U grid/macropad layout

Other layouts exist, such as "southpaw" horizontally mirrored layouts, and layouts with a fifth column, but those seem to be much less common. There are a lot of layouts here already, so we can add the less common ones later if anyone asks for them.


Nick-Munnich:
LGTM

petejohanson (approved):
Thanks!

PR #2586, opened 2024-10-23
docs: fixing some more broken anchors

Nick-Munnich:


caksoylar (approved):
I think we usually omit `./` for relative paths in the same folder, but fine by me either way.

PR #2591, opened 2024-10-25
Kinesis advantage2 physical layout

dhruvinsh:
Adding physical layout support for Kinesis Advantage2.

This physical layout has been created from popular stapelberg mode from qmk community. Helper tools was,
https://zmk-physical-layout-converter.streamlit.app/

Reference:
https://github.com/qmk/qmk_firmware/blob/master/keyboards/kinesis/stapelberg/keyboard.json

### DIFF
--- a/app/dts/layouts/kinesis/adv2.dtsi
+++ b/app/dts/layouts/kinesis/adv2.dtsi
@@ -1,0 +1,100 @@
+#include <physical_layouts.dtsi>
+
+/ {
+    layout_0: layout_0 {
+        compatible = "zmk,physical-layout";
+        display-name = "Default";
+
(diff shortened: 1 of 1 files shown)


petejohanson:
Thanks for the PR! We are really not looking to merge arbitrary layouts into ZMK, much like we don't merge new shields/boards. The current shared layouts in ZMK are an *example*, and other shared layouts should be in modules, just like boards/shields are.

Long term, the shared layouts in ZMK will move *out* of ZMK along with the boards/shields we have, once we have versioning in place and can avoid breaking things for existing users.

Closing this for now, but please feel free to reopen if you have any other concerns.

PR #2593, opened 2024-10-26
Update web-tree-sitter and tweak keymap upgrader

joelspadin:
Updated web-tree-sitter and switched from having a copy of tree-sitter-devicetree.wasm in the repo to using the tree-sitter-devicetree module, which now includes the .wasm file.

Also fixed an issue with tsconfig.json which was causing imports to not resolve correctly in IDEs, and adjusted the keymap ugprader textarea so it can only be resized vertically.


Nick-Munnich (approved):
Gave it a bit of a test by introducing errors into a keymap. LGTM.

PR #2595, opened 2024-10-27
refactor: physical layout consistency

Nick-Munnich:
An alternative approach to #2582, as suggested by Joel and supported by Pete. Felt multiple PRs would be better here, to allow us to compare the different versions more directly.

Overall it looks nice, but future refactors to make a layout shared would still often be slightly more involved than dragging and dropping into the shared folder - multiple layouts present = need to be uncoupled.


app/boards/arm/adv360pro/adv360pro-layouts.dtsi
```diff
@@ -0,0 +1,87 @@
+#include <physical_layouts.dtsi>
+
+/{
+    default_layout: default_layout {
```
petejohanson, line 4: I am on the fence on the naming convention here of `default_layout`. Not opposed, just pondering. Some other ideas:

* `standard_layout`
* `std_layout`
* `stock_layout`
* `physical_layout`
* `physical_layout0`

Thoughts?

petejohanson (commented):
I generally favor this approach over #2582. One point of discussion.

app/boards/arm/adv360pro/adv360pro-layouts.dtsi
Nick-Munnich: `physical_layout0` would be the only contender aside from `default_layout` in my eyes.

petejohanson: Yeah, that matches the `kscan0` convention we've used, thus my suggestion. @joelspadin or @caksoylar any opinions?

caksoylar: I like `physical_layout0` also, since it is consistent with kscan (but breaks the matrix transform convention).

app/boards/arm/adv360pro/adv360pro-layouts.dtsi
```diff
@@ -0,0 +1,87 @@
+#include <physical_layouts.dtsi>
+
+/{
```
caksoylar, line 3: Nit: it'd be nice if we consistently used `/ {` with a space.

app/boards/arm/adv360pro/adv360pro-layouts.dtsi
Nick-Munnich: Changed everything over to `physical_layout0`. Also renamed some transform labels to `matrix_transform0` for consistency.

caksoylar: Sorry to bring this up after the changes, but I think we should add the new labels as additional rather than replace the existing ones. Othewrise we'll break existing configs which might use them in `chosen` nodes.

Nick-Munnich: I only changed the ones where we only have a single physical layout defined, or where the changed label is already the chosen one. I can't think of another reason why someone would need to refer to a physical layout or matrix transform in their config.

app/boards/shields/jiran/jiran-layouts.dtsi
```diff
@@ -1,6 +1,14 @@
+#include <layouts/kgoh/jian/6column.dtsi>
+#include <layouts/kgoh/jian/full.dtsi>
 #include <physical_layouts.dtsi>
 
-#include "position_map.dtsi"
+&layouts_kgoh_jian_position_map {
+    status = "disabled";
+};
```
petejohanson, line 7: Why are we not just enhancing the existing map here, and duplicating things? Not sure I agree with this, unless there's some good reason to do so.

app/boards/shields/jorne/jorne-layouts.dtsi
```diff
@@ -1,6 +1,10 @@
+#include <layouts/foostan/corne/6column.dtsi>
+#include <layouts/foostan/corne/5column.dtsi>
 #include <physical_layouts.dtsi>
 
-#include "position_map.dtsi"
+&layouts_foostan_corne_position_map {
+    status = "disabled";
+};
```
petejohanson, line 7: Ditto.

petejohanson (commented):
One question/issue with the map override approach taken here.

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -448,19 +448,10 @@ Also see the [matrix transform section](../../config/layout.md#matrix-transform)
...
 ```dts
-/ {
-    default_layout: default_layout {
-        compatible = "zmk,physical-layout";
-        display-name = "Default Layout";
-        transform = <&default_transform>;
-        kscan = <&kscan0>;
-    };
-};
+#import "my_keyboard-layouts.dtsi"
```
caksoylar, line 454:
```suggestion
#include "my_keyboard-layouts.dtsi"
```

docs/docs/development/hardware-integration/physical-layouts.md
```diff
@@ -10,22 +10,32 @@ It contains:
...
+Every physical layout needs a matrix transform, and optionally can also have a kscan. By convention, these are assigned to the physical layout in the file where the matrix transform and kscan are defined:
+
+```dts title="<your keyboard>.dts | <your keyboard>.dtsi | <your keyboard>.overlay"
+&physical_layout0 {
+    kscan = <&kscan0>;
+    transform = <&matrix_transform0>;
+};
+```
+
+The `kscan` property only needs to be assigned if some of your physical layouts use different kscans. Otherwise, it can be omitted and the `kscan` should be assigned in the [`chosen` node](./new-shield.mdx#chosen-node) instead. See the [configuration section on physical layouts](../../config/index.md) for reference.
```
caksoylar, line 38:
```suggestion
The `kscan` property only needs to be assigned if some of your physical layouts use different kscans. Otherwise, it can be omitted and the `kscan` can be assigned in the [`chosen` node](./new-shield.mdx#chosen-node) instead. See the [configuration section on physical layouts](../../config/layout.md#physical-layout) for reference.
```

"should -> can" is better since we aren't explicitly encouraging it, are we?

In any case, the link should be fixed

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -448,19 +448,10 @@ Also see the [matrix transform section](../../config/layout.md#matrix-transform)
...
 ```
 
 ### Chosen Node
```
caksoylar, line 457: The physical layout name in chosen node example below should be updated too.

caksoylar (changes requested):

Nick-Munnich:
Should be all good now.

caksoylar (approved):

PR #2596, opened 2024-10-29
Waveshare RP2040-Zero board addition

miketronic:
## Board/Shield Check-list

- [ ] N/A Keymaps do not use deprecated key defines (Check using the [upgrader tool](https://zmk.dev/docs/codes/keymap-upgrader))
- [ ] N/A `&pro_micro` used in favor of `&pro_micro_d/a` if applicable
- [ ] N/A If split, no name added for the right/peripheral half
- [ ] N/A `.conf` file has optional extra features commented out
- [x] Keyboard/PCB is part of a shipped group buy or is generally available in stock to purchase (OSH/personal projects without general availability should create a zmk-config repo instead) -- board is available from many sources.

This PR takes off from where glebsexy left off.  I attempted to make all changes that Pete mentioned in the comments of that PR (1871).


Nick-Munnich:
Are the outer pins of the zero footprint-compatible with the xiaos? Could you define an interconnect and GPIO nexuses for the zero?

I would also say that this is a board worth attempting to upstream to Zephyr, since waveshare already has some boards defined there.

miketronic:
> Are the outer pins of the zero footprint-compatible with the xiaos? Could you define an interconnect and GPIO nexuses for the zero?
> 
> I would also say that this is a board worth attempting to upstream to Zephyr, since waveshare already has some boards defined there.

The Xiao is even smaller, shorter by two pins on both sides and none of the pins opposite the USB port that the Zero has.

Nick-Munnich:
> The Xiao is even smaller, shorter by two pins on both sides and none of the pins opposite the USB port that the Zero has.

My question is more, if you have a keyboard designed for the Xiao RP2040, could you use the Zero with it or would the width/spacing not match up

Nick-Munnich:
As a consequence of them being compatible I think a xiao_d gpio nexus node should be added so that the zero can be used in place of the xiao on xiao shields.

miketronic:
Deleted the old branch and renamed the new one.

miketronic:
> As a consequence of them being compatible I think a xiao_d gpio nexus node should be added so that the zero can be used in place of the xiao on xiao shields.

Looking deeper, unfortunately they are not pin compatible unless you very specifically designed the shield with a cutout in the middle for the internals and USB connector, essentially having a "U" of the connecting pins.  5v/GND/3v3 are on opposite sides of the board for both of the MCUs.

app/boards/arm/waveshare_rp2040_zero/waveshare_rp2040_zero.dts
```diff
@@ -0,0 +1,162 @@
...
+        #clock-cells = <0>;
+    };
+
+    aliases {
+        watchdog0 = &wdt0;
+    };
+
+
+    zero_header: connector {
+        compatible = "raspberrypi,pico-header";
```
petejohanson, line 45: This is a compatible that should only be used if this matches the physical layout of a Pi Pico board, which the Zero does not. You'll need a new compatible bindings file for this.

app/boards/arm/waveshare_rp2040_zero/waveshare_rp2040_zero.dts
```diff
@@ -0,0 +1,162 @@
...
+zephyr_udc0: &usbd {
+    status = "okay";
+};
+
+// &vreg {
+// 	regulator-always-on;
+// 	regulator-allowed-modes = <REGULATOR_RPI_PICO_MODE_NORMAL>;
+// };
+
+// zero_spi: &spi0 {};
```
petejohanson, line 162: Remove these instead of leaving commented out.

app/boards/arm/waveshare_rp2040_zero/waveshare_rp2040_zero_defconfig
```diff
@@ -0,0 +1,26 @@
...
+# Enable GPIO
+CONFIG_GPIO=y
+
+# enable uart driver
+CONFIG_SERIAL=y
+CONFIG_UART_INTERRUPT_DRIVEN=y
+
+# enable console
+CONFIG_CONSOLE=y
+CONFIG_UART_CONSOLE=y
```
petejohanson, line 20: These defaults are what Zephyr recommends for their "stock board setup" but doesn't match ZMK's standards. Remove these please:

```suggestion
```

app/boards/arm/waveshare_rp2040_zero/waveshare_rp2040_zero_defconfig
```diff
@@ -0,0 +1,26 @@
...
+
+# Code partition needed to target the correct flash range
+CONFIG_USE_DT_CODE_PARTITION=y
+
+# Output UF2 by default, native bootloader supports it.
+CONFIG_BUILD_OUTPUT_UF2=y
```
petejohanson, line 26: Should add lines to enable settings, flash, flash NVS, etc. by default, so we can store various settings, e.g. RGB state or studio keymaps.

petejohanson (changes requested):

nmunnich:
See #2806

PR #2598, opened 2024-10-30
docs: studio features list

Nick-Munnich:
Trying to resolve [studio 70](https://github.com/zmkfirmware/zmk-studio/issues/70).

Likely there are things that I've missed and need to be listed. I think this table makes the most sense in this page, at least for now.

### DIFF
--- a/docs/docs/features/studio.md
+++ b/docs/docs/features/studio.md
@@ -10,6 +10,35 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
 
 ZMK Studio provides runtime update functionality to ZMK powered devices, allowing users to change their keymap layers without flashing new firmware to their keyboards. Studio is still under active development, and is not yet ready for casual end user use.
 
+## Capabilities
+
+ZMK Studio currently has the following capabilities:
+
+| Legend: | ✅ Supported | 🚧 Under Development | 💡 Planned | 🕯️ Low Priority | ❌ Not Planned |
+| :------ | :----------- | :------------------- | :--------- | :-------------- | -------------- |
+
+| Status | Feature/Capability                                                                                                                        |
+| :----: | ----------------------------------------------------------------------------------------------------------------------------------------- |
+|   ✅   | Making keymap changes while keyboard is in use                                                                                            |
+|   ✅   | Making changes while connected via USB                                                                                                    |
+|   ✅   | Native app for: Windows, Linux, MacOS                                                                                                     |
+|   ✅   | Making changes while connected via BLE (Linux web-app & native apps only)                                                                 |
+|   ✅   | Assigning [predefined behaviors](../keymaps/behaviors/index.mdx) to keys                                                                  |
+|   ✅   | Assigning [user-defined behaviors](../keymaps/behaviors/index.mdx#user-defined-behaviors) to keys                                         |
+|   💡   | [Configuring basic behavior properties](../config/behaviors.md)                                                                           |
+|   💡   | Configuring advanced behavior properties, e.g. [tap dance](../keymaps/behaviors/tap-dance.mdx), [macros](../keymaps/behaviors/macros.md), |
+|   ❌   | Defining new behaviors not specified in devicetree                                                                                        |
+|   💡   | Configuring [combos](../keymaps/combos.md)                                                                                                |
+|   💡   | Configuring [conditional layers](../keymaps/conditional-layers.md)                                                                        |
+|   🕯️   | Assigning behaviors to encoders                                                                                                           |
+|   ✅   | Selecting alternative pre-defined physical layouts for the keyboard                                                                       |
+|   ❌   | Defining new physical layouts for the keyboard                                                                                            |
+|   ✅   | Renaming layers & enabling [extra layers](#including-extra-layers)                                                                        |
+|   ❌   | Adding more layers than specified by devicetree                                                                                           |
+|   💡   | Host locale selection                                                                                                                     |
+
+Items listed as "planned", "under development", "low priority", or "not planned" can be configured using [devicetree](../config/index.md#devicetree-files) instead.
+
 ## Keymap Changes
 
 To unlock your keyboard to allow ZMK Studio to make changes, you'll need to add a [`&studio_unlock`](../keymaps/behaviors/studio-unlock.md) binding to the keymap.


docs/docs/features/studio.md
```diff
@@ -10,6 +10,34 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+
+| Studio Status | **Feature or Capability**                                                                         |
+| :-----------: | ------------------------------------------------------------------------------------------------- |
+|      ✅       | Making keymap changes while keyboard is in use                                                    |
+|      ✅       | Making changes while connected via USB                                                            |
+|      ✅       | Native app for: Windows, Linux, MacOS                                                             |
+|      ✅       | Making changes while connected via BLE (Linux web-app & native apps only)                         |
+|      ✅       | Assigning [predefined behaviors](../keymaps/behaviors/index.mdx) to keys                          |
+|      ✅       | Assigning [user-defined behaviors](../keymaps/behaviors/index.mdx#user-defined-behaviors) to keys |
+|      💡       | Defining new [user-defined behaviors](../keymaps/behaviors/index.mdx#user-defined-behaviors)      |
```
petejohanson, line 28: Not really... Sorta, but not really? We'll have support for "reserved" behaviors of a given type, much like we have reserved keymap layers, that can then be "added" and further configured from there.

docs/docs/features/studio.md
```diff
@@ -10,6 +10,34 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|       ?       | [Configuring predefined behaviors](../config/behaviors.md)                                        |
```
petejohanson, line 29: Yes, this is definitely planned.

docs/docs/features/studio.md
```diff
@@ -10,6 +10,34 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|      💡       | Defining [combos](../keymaps/combos.md)                                                           |
+|      💡       | Defining [conditional layers](../keymaps/conditional-layers.md)                                   |
+|       ?       | Assigning behaviors to encoders                                                                   |
```
petejohanson, line 32: I am not really interested in this *yet* since we're still up in the air on moving encoders to the input API, etc. so this is pretty low on my list at this time.

petejohanson (commented):
Thanks for starting on this!

Other items planned, as part of the "configuring behaviors" is tap-dance and macro editing support.

docs/docs/features/studio.md
Nick-Munnich: What I've done is added another category "low priority" and put this point under it.

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|      ✅       | Making changes while connected via USB                                                            |
+|      ✅       | Native app for: Windows, Linux, MacOS                                                             |
+|      ✅       | Making changes while connected via BLE (Linux web-app & native apps only)                         |
+|      ✅       | Assigning [predefined behaviors](../keymaps/behaviors/index.mdx) to keys                          |
+|      ✅       | Assigning [user-defined behaviors](../keymaps/behaviors/index.mdx#user-defined-behaviors) to keys |
+|      💡       | [Configuring predefined behaviors](../config/behaviors.md)                                        |
+|      💡       | [Configuring user-defined behaviors](../config/behaviors.md)                                      |
+|      💡       | [Configuring tap-dance](../keymaps/behaviors/tap-dance.mdx)                                       |
+|      💡       | [Configuring macros](../keymaps/behaviors/macros.md)                                              |
+|      ❌       | Defining and configuring behaviors beyond what is specified by devicetree                         |
```
petejohanson, line 32: This one *may* give the wrong impression, as I read it.. I wonder how we can succinctly capture the "reserved" concept for this without being too verbose.

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|      💡       | Defining [combos](../keymaps/combos.md)                                                           |
+|      💡       | Defining [conditional layers](../keymaps/conditional-layers.md)                                   |
+|      🕯️       | Assigning behaviors to encoders                                                                   |
```
petejohanson, line 35: Love the candle.

docs/docs/features/studio.md
Nick-Munnich: I think I would change the text "configuring user-defined behaviors" to "configuring user-defined and reserved behaviors" when that turns from planned into a reality, and change this point into "Adding additional reserved behaviors beyond [...]".

I don't think it's necessary to make that distinction yet, as reserved behaviors aren't a thing yet. But I see your point, so... "Defining new behaviors not specified in devicetree"?

petejohanson: "Defining new behaviors not specified in devicetree" WFM.

Nick-Munnich: Done.

petejohanson (approved):

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|      ✅       | Making keymap changes while keyboard is in use                                                    |
+|      ✅       | Making changes while connected via USB                                                            |
+|      ✅       | Native app for: Windows, Linux, MacOS                                                             |
+|      ✅       | Making changes while connected via BLE (Linux web-app & native apps only)                         |
+|      ✅       | Assigning [predefined behaviors](../keymaps/behaviors/index.mdx) to keys                          |
+|      ✅       | Assigning [user-defined behaviors](../keymaps/behaviors/index.mdx#user-defined-behaviors) to keys |
+|      💡       | [Configuring predefined behaviors](../config/behaviors.md)                                        |
+|      💡       | [Configuring user-defined behaviors](../config/behaviors.md)                                      |
+|      💡       | [Configuring tap-dance](../keymaps/behaviors/tap-dance.mdx)                                       |
+|      💡       | [Configuring macros](../keymaps/behaviors/macros.md)                                              |
```
caksoylar, line 31: I was a bit confused on why these were split from "user-defined behaviors" since they technically are. Is what we mean here is implementing a special UI to configure them?

If so, can they still be configured without a special UI when "configuring user-defined behaviors" is implemented? Or does it mean that former doesn't encapsulate these because e.g. `bindings` will not be editable in Studio?

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
 ZMK Studio provides runtime update functionality to ZMK powered devices, allowing users to change their keymap layers without flashing new firmware to their keyboards. Studio is still under active development, and is not yet ready for casual end user use.
 
+## Capabilities
+
+ZMK Studio currently has the following capabilities:
+
+| Legend: | ✅ Supported | 🚧 Under Development | 💡 Planned | 🕯️ Low Priority | ❌ Not Planned |
+| :------ | :----------- | :------------------- | :--------- | :-------------- | -------------- |
+
+| Studio Status | **Feature or Capability**                                                                         |
```
caksoylar, line 20:
```suggestion
| Status | Feature/Capability                                                                       |
```

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+| :-----------: | ------------------------------------------------------------------------------------------------- |
+|      ✅       | Making keymap changes while keyboard is in use                                                    |
```
caksoylar, line 22: I wonder if this is necessary. Is it just to constrast with flashing a new keymap?

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|      ✅       | Making changes while connected via USB                                                            |
+|      ✅       | Native app for: Windows, Linux, MacOS                                                             |
```
caksoylar, line 24: Similar question for this one, seems like it shouldn't be in the feature list and is already mentioned below.

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|      ✅       | Making changes while connected via BLE (Linux web-app & native apps only)                         |
+|      ✅       | Assigning [predefined behaviors](../keymaps/behaviors/index.mdx) to keys                          |
+|      ✅       | Assigning [user-defined behaviors](../keymaps/behaviors/index.mdx#user-defined-behaviors) to keys |
+|      💡       | [Configuring predefined behaviors](../config/behaviors.md)                                        |
+|      💡       | [Configuring user-defined behaviors](../config/behaviors.md)                                      |
+|      💡       | [Configuring tap-dance](../keymaps/behaviors/tap-dance.mdx)                                       |
```
caksoylar, line 30:
```suggestion
|      💡       | [Configuring tap-dances](../keymaps/behaviors/tap-dance.mdx)                                       |
```

docs/docs/features/studio.md
```diff
@@ -10,6 +10,37 @@ ZMK Studio is in beta. Although every effort has been made to provide a stable e
...
+|      💡       | [Configuring macros](../keymaps/behaviors/macros.md)                                              |
+|      ❌       | Defining new behaviors not specified in devicetree                                                |
+|      💡       | Defining [combos](../keymaps/combos.md)                                                           |
+|      💡       | Defining [conditional layers](../keymaps/conditional-layers.md)                                   |
```
caksoylar, line 34: Do the same caveats for behaviors (configuring existing/reserved vs. defining) apply here? This is a bit inconsistent if so.

caksoylar (commented):
If this is settled already I don't want to re-open this, but is there a  major difference between "configuring reserved X" vs. "defining new X" from an end user point-of-view? Also see the final comment here as related to this.

docs/docs/features/studio.md
petejohanson: Macros have a very specific functionality that's going to require some careful work to make editing the `bindings` easy for end users. It will definitely be an extension/additional effort on top of the generic behavior property editing we'll have to build out.

docs/docs/features/studio.md
petejohanson: Yeah, I'm ok if this isn't included in the list. It's not really a "feature" per se.

docs/docs/features/studio.md
petejohanson: Yes, for sure, this'll require the same reserved approach when we implement it.

docs/docs/features/studio.md
caksoylar: So should we qualify these as e.g. "Macro configuration interface", or does this mean we should edit the "configuring user-defined behaviors" item to exclude tap-dances+macros?

petejohanson: Since this is purely informational, I'm not sure the damage in keeping it separated so users can find it easily?

docs/docs/features/studio.md
Nick-Munnich: I was thinking of the possible features that Studio could be extended with in the future when mentioning this and the previous point. For completeness I felt it's nice to have them listed in that case.

docs/docs/features/studio.md
Nick-Munnich: I'm not sure it's worth making the distinction yet here, like in a previous point, but I'm happy to revise if you think it's needed.

caksoylar: Honestly, if we are sticking to the current distinction these should be split to two ("configuring existing" -> planned, "defining" -> not planned).

But like I point out in the main review comment, I could see us going the other way and not making that distinction in this table at all and consider "configure" the same as "defining" and omitting the "reserved" detail.

Nick-Munnich: I've changed it to "configuring" but chosen to omit the not planned "defining" for the time being.

docs/docs/features/studio.md
Nick-Munnich: Adjusted based on what I *think* the two of you agreed on on Discord.

caksoylar (approved):

PR #2602, opened 2024-11-01
fix(keymap): Set bindings with changed layout

petejohanson:
* Fix bug with setting keymap layer bindings when a non-stock layout is selected.

Should fix zmkfirmware/zmk-studio#73

Tested here on a Technikable with a different layout, where I was able to reproduce the issues, as well as testing the attached fix.


petejohanson:
Tested a ton locally and it's now working as expected. Original reporter is still having some issues, but based on the code this is a definite first step in improving things, while I dig more to see what issues they're still having.

PR #1595, opened 2022-12-17
Set C_AL_CALC for macOS and iOS to false

bttnns:
Docs update to set C_AL_CALC to false for macOS and iOS

PR #2606, opened 2024-11-03
Marking C_AL_CALENDAR / C_AL_CAL as not available on macOS and iOS.

amacleod:
Small docs update to mark the calendar consumer app launch keycode as unavailable on macOS and iOS.

I tested this with https://github.com/amacleod/zmk-config/tree/6d12fec60cee47fd24dd7174a30a76bc18428d81 including enabling `CONFIG_ZMK_HID_CONSUMER_REPORT_USAGES_BASIC`. I tested with my Lily58 connected by Bluetooth to a MacBook Pro and an iPad, both running fairly recent versions of their respective operating systems. Pressing the keys bound to `C_AL_CAL` and `C_AL_CALC` (see #1595) had no effect on either host.


Nick-Munnich (approved):
Thanks again for testing!

PR #2614, opened 2024-11-05
chore: Config dependabot for grouped npm updates

caksoylar:
Attempt to organize dependabot so that:
- we don't have individual "version" (as opposed to "security") updates
- PRs are grouped into separate production and dev groups
- for production, two groups for major vs. minor+patch
- for dev, only minor+patch
- security updates aren't touched, so there will be individual PRs as is

The resulting PRs can be seen at https://github.com/caksoylar/zmk/pulls
Config reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

Open question/points for discussion:
- should we also have a separate major updates group for dev deps? or alternatively, remove minor+patch constraint?
- should security updates be constrained to minor+patch?
- any need to create other groups for specific dependencies (e.g. `docusaurus*`)?
- should we touch the GH actions ecosystem entry? I am not sure if it has prod/dev but we can create groups (with semver constraints)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,25 +9,6 @@ updates:
     schedule:
       interval: "daily"
     groups:
-      production-major:
-        applies-to: "version-updates"
-        dependency-type: "production"
-        exclude-patterns:
-          - "@docusaurus/*"
-          - "tree-sitter-devicetree"
-          - "web-tree-sitter"
-        update-types:
-          - "major"
-      production-minor-patch:
-        applies-to: "version-updates"
-        dependency-type: "production"
-        exclude-patterns:
-          - "@docusaurus/*"
-          - "tree-sitter-devicetree"
-          - "web-tree-sitter"
-        update-types:
-          - "minor"
-          - "patch"
       docusaurus-major:
         applies-to: "version-updates"
         dependency-type: "production"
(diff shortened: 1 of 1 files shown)


Nick-Munnich:
Definitely like the look of this.

>     * should we also have a separate major updates group for dev deps? or alternatively, remove minor+patch constraint?

I think as long as we have security alerts set up I don't see the need to update dev deps major versions through dependabot.

>     * should security updates be constrained to minor+patch?

No, I think security stuff is important enough that we should put in the work to update major versions when they occur.

>     * any need to create other groups for specific dependencies (e.g. `docusaurus*`)?

I think a docusaurus group would be nice, yeah.
 
>     * should we touch the GH actions ecosystem entry? I am not sure if it has prod/dev but we can create groups (with semver constraints)

No opinion.

caksoylar:
> I think a docusaurus group would be nice, yeah.

Looking at the non-dev list https://github.com/zmkfirmware/zmk/blob/main/docs/package.json#L17-L34, we can either do `@docusaurus` and others, or perhaps we can do the following as separate:
- `@docusaurus/*`
- `@fort-awesome/*
- `react*` (maybe add `@mdx-js/react`?)
- `*tree-sitter*`
- others

The dev dependencies also have a `@docusaurus` group but it seems too fine-grained to split that as well.

Nick-Munnich:
I'd say do @docusaurus, but singling out tree-sitter would be good.

caksoylar:
I decided to only split prod to three (docusaurus, tree-sitter, other) and didn't split tree-sitter group to major/minor+patch. Let's see if it looks better.

@joelspadin any opinions on tree-sitter deps specifically, or in general?

Nick-Munnich (approved):

joelspadin:
Splitting out the tree-sitter packages seems good to me. That hasn't hit 1.x yet, so any updates could potentially be breaking.

PR #2615, opened 2024-11-06
fix: Fix kscan prop existence test in physical layouts

caksoylar:
Currently building with a chosen physical layout which has a kscan, but no chosen "shared" kscan fails. See example [build](https://github.com/rmuraglia/zmk-config/actions/runs/11695134268) and corresponding failing [overlay](https://github.com/rmuraglia/zmk-config/blob/7fabcb6e0e9413c782edf002593730bde92b9a40/boards/shields/menura/menura.dtsi).

I am guessing this is because `DT_INST_PROP_LEN` doesn't work for phandle type properties, and we can use `DT_INST_NODE_HAS_PROP` instead. (I haven't spent too much time looking into this, happy to change if there is a better check.)


caksoylar:
I think this might be a problem with an older ZMK version and doesn't currently reproduce, sorry about the noise.

PR #2620, opened 2024-11-08
feat(build): checkout submodules

nstetter:
To be able to share code between users submodules can come in very handy. 

https://github.com/bryanforbes/zmk-nodefree-config for example is best used as a submodule


nstetter:
Just found out this can be done in west.yaml with modules

PR #2626, opened 2024-11-09
docs: Fix typos in local-toolchain/setup

clbn:
Just noticed there's a typo in the docs. I believe it should be `/workspaces/*`, not `/workspace/*`.


Nick-Munnich (approved):
Thanks for catching this!

PR #2628, opened 2024-11-10
fix: Disable display feature for settings_reset

caksoylar:
Without this tweak, a build like `west build -b corneish_zen_v2_left -- -DSHIELD=settings_reset` fails because the status screens are enabled and they expect certain APIs related to e.g. BT profiles that don't get built. This can be an issue with any onboard controller keyboard that enables a display with status screen by default, like ones that live in modules. While good practice would be to `#if defined` guard the appropriate widgets, there is also no reason to have display enabled with settings reset.


joelspadin:
Yeah, fixing the status screens would be the ideal solution, but you're probably not going to think to do that when writing a custom status screen for a board that always supports BLE. This seems reasonable to me.

joelspadin (approved):

petejohanson (approved):
Thanks!

PR #2630, opened 2024-11-10
fix: Fix inconsistent column offset property

joelspadin:
Renamed the composite kscan's column-offset property to col-offset for consistency with other properties such as matrix transform's col-offset and matrix kscan's col-gpios.

column-offset remains supported but is deprecated.


caksoylar (approved):
Although trivial, I'll formally give an approval for the docs side.

petejohanson (approved):
LGTM. Thanks!

PR #2639, opened 2024-11-13
docs: move layer defines to reduce confusion

Genteure:
Moved the section about defines to the bottom, clearifing layer behaviors take a number as argument.

### DIFF
--- a/docs/docs/keymaps/behaviors/layers.md
+++ b/docs/docs/keymaps/behaviors/layers.md
@@ -12,22 +12,6 @@ Some of those behaviors are still in the works; the ones that are working now ar
 Multiple layers can be active at the same time and activating a layer will not deactivate layers higher up in the "layer stack". See [Layers](../index.mdx#layers) for more information.
 :::
 
-## Defines to Refer to Layers
-
-When working with layers, you may have several different key positions with bindings that enable/disable those layers.
-To make it easier to refer to those layers in your key bindings, and to change which layers are where later, you can
-add a set of `#define`s at the top of your keymap file, and use those layer in your keymap.
-
-For example, if you have three layers, you can add the following to the top of your keymap:
-
-```dts
-#define DEFAULT 0
-#define LOWER   1
-#define RAISE   2
-```
-
-This allows you to use those defines, e.g. `LOWER` later in your keymap.
-
 ## Momentary Layer
 
 The "momentary layer" behavior enables a layer while a certain key is pressed. Immediately upon
@@ -42,7 +26,7 @@ again.
 Example:
 
 ```dts
-&mo LOWER
+&mo 3
 ```
 
 ## Layer-Tap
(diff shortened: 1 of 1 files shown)


docs/docs/keymaps/behaviors/layers.md
```diff
@@ -42,7 +26,7 @@ again.
 Example:
 
 ```dts
-&mo LOWER
+&mo 3
 ```
```
caksoylar, line 31: Can I suggest you add a e.g. `note` admonition box here, reminding that layer numbers correspond to the order they are defined under the keymap node, starting from zero? e.g. the `3` here would be the 4th node, etc.

Having the `#define`s earlier sort of illustrated this before.

caksoylar (commented):
Thanks for remembering and following up! Good change, added one suggestion to improve.

docs/docs/keymaps/behaviors/layers.md
Genteure: I added a new line to the admonition box just above here. Having just one highlighted section feels cleaner than two right next to each other to me. The linked page also have more info about layer numbers with examples. What do you think?

docs/docs/keymaps/behaviors/layers.md
```diff
@@ -9,25 +9,11 @@ Often, you may want a certain key position to alter which layers are enabled, ch
 Some of those behaviors are still in the works; the ones that are working now are documented here.
 
 :::note
-Multiple layers can be active at the same time and activating a layer will not deactivate layers higher up in the "layer stack". See [Layers](../index.mdx#layers) for more information.
+Multiple layers can be active at the same time and activating a layer will not deactivate layers higher up in the "layer stack".
+Layer numbers start at 0 following the order they are defined in the keymap node, `&mo 3` would activate the 4th layer node defined in the keymap.
```
caksoylar, line 13:
```suggestion

Layer numbers start at 0 following the order they are defined in the keymap node, for example `&mo 3` would activate the 4th layer node defined in the keymap.
```

Added a line break and "for example".

docs/docs/keymaps/behaviors/layers.md
```diff
@@ -9,25 +9,11 @@ Often, you may want a certain key position to alter which layers are enabled, ch
 Some of those behaviors are still in the works; the ones that are working now are documented here.
```
caksoylar, line 9:
```suggestion
Below are the list of behaviors that can be used to activate and deactivate layers.
```

This is bothering me, so can we update while we are touching around it? :)

docs/docs/keymaps/behaviors/layers.md
caksoylar: I thought about the note box right after the first time it is used in an example, e.g. here after `&mo 3` but I guess they would still be pretty close. I think yours is fine, yeah.

Nick-Munnich:
Considering the rephrasing of the second sentence, can we link to sticky key in much the same way as conditional layers?

Nick-Munnich (approved):

caksoylar:
We'll handle above in a separate PR.

PR #2641, opened 2024-11-13
fix(studio): Improved error message when keyboard is missing a physical layout

Nick-Munnich:


app/src/physical_layouts.c
```diff
@@ -32,8 +32,11 @@ ZMK_EVENT_IMPL(zmk_physical_layout_selection_changed);
 
 BUILD_ASSERT(
     !IS_ENABLED(CONFIG_ZMK_STUDIO) || USE_PHY_LAYOUTS,
-    "ZMK Studio requires physical layouts with key positions, and no chosen zmk,matrix-transform. "
-    "See https://zmk.dev/docs/development/hardware-integration/studio-setup");
+    "DEAR USER: Keyboards require additional configuration to allow for firmware with ZMK "
```
petejohanson, line 35: Not sure I love the "DEAR USER". Could come off as patronizing, if viewed incorrectly, IMHO. I'd stick to plain language and skip it.
```suggestion
    "Keyboards require additional configuration to allow for firmware with ZMK "
```

Nick-Munnich: Most often people encountering build failures will see the massive error text and be completely lost when seeing it. My hope would be that having that in all-caps would make it more noticeable. I'm happy to remove, but I think we should consider some sort of standard text in our build asserts that stands out, so that users are likelier to notice it.

petejohanson: Yeah, totally valid... Perhaps something more generic, e.g. `IMPORTANT: ` ?

Nick-Munnich: `ISSUE FOUND:` maybe? Do you know if there are any issues with inserting new line characters into the error message? We could use those to separate out the message from the rest of the text a bit, making it pop out more. On re-reading the standard error message I think if we can't do so, then the additional caps wouldn't suffice for visibility.

petejohanson: I would avoid newlines, IMHO. I'm fine with "ISSUE FOUND" if you prefer that.

Nick-Munnich: Done

petejohanson (approved):

PR #2642, opened 2024-11-13
chore: Add codespell integration

elpekenin:
As per title/commit, very tiny changes :)

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = docs/package-lock.json,*.svg
+ignore-words = .codespellignorewords
--- a/docs/blog/2023-04-06-zephyr-3-2.md
+++ b/docs/blog/2023-04-06-zephyr-3-2.md
@@ -278,7 +278,7 @@ The following RP2040 powered controllers have board definitions for folks to tes
 
 - Raspberry Pi Pico (`rpi_pico`)
 - SparkFun Pro Micro RP2040 (`sparkfun_pro_micro_rp2040`)
-- Adafruit Keyboar/KB2040 (`adafruit_kb2040`)
+- Adafruit Keyboar/KB2040 (`adafruit_kb2040`) <!-- codespell:ignore keyboar -->
 - Seeeduino XIAO RP2040 (`seeeduino_xiao_rp2040`)
 - Adafruit Qt PY RP2040 (`adafruit_qt_py_rp2040`)
 - BoardSource blok (`boardsource_blok`)
--- a/docs/docs/config/battery.md
+++ b/docs/docs/config/battery.md
@@ -55,7 +55,7 @@ Driver for reading the voltage of a battery using an ADC connected to a voltage 
 
 Applies to: `compatible = "zmk,battery-voltage-divider"`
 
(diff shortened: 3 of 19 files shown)


docs/docs/config/battery.md
```diff
@@ -55,7 +55,7 @@ Driver for reading the voltage of a battery using an ADC connected to a voltage
 
 Applies to: `compatible = "zmk,battery-voltage-divider"`
 
-See [Zephyr's voltage divider documentation](https://docs.zephyrproject.org/3.5.0/build/dts/api/bindings/iio/afe/voltage-divider.html).
+See [Zephyr's voltage divider documentation](https://docs.zephyrproject.org/3.5.0/build/dts/api/bindings/iio/afe/voltage-divider.html). <!-- codespell:ignore afe -->
```
caksoylar, line 58: Does this not ignore URL's when checking spelling?

caksoylar (commented):
I'd be happy to have this for docs, thanks. For the rest (dts and C comments) I'll defer to others.

docs/docs/config/battery.md
elpekenin: Nope, it error'ed

caksoylar: It looks like it is possible to configure it to ignore them: https://github.com/codespell-project/codespell/blob/main/codespell_lib/_codespell.py#L471

I think it's a good idea to ignore them, if anyone else thinks otherwise let me know.

docs/blog/2023-04-06-zephyr-3-2.md
```diff
@@ -278,7 +278,7 @@ The following RP2040 powered controllers have board definitions for folks to tes
 
 - Raspberry Pi Pico (`rpi_pico`)
 - SparkFun Pro Micro RP2040 (`sparkfun_pro_micro_rp2040`)
-- Adafruit Keyboar/KB2040 (`adafruit_kb2040`)
+- Adafruit Keyboar/KB2040 (`adafruit_kb2040`) <!-- codespell:ignore keyboar -->
```
joelspadin, line 281: As far as I can tell, the correct name is not "Keyboard" but "Kee Boar Driver".

I expect that will result in flagging "Kee" though. If so, is it possible to ignore sequences of words such that "Kee" on its own is flagged but not "Kee Boar"?

elpekenin: > Kee Boar

Right, i didn't even Google it and assumed it was correctly spelled. It's a weird name either way 🙃 

> "Kee" is flagged but not "Kee Boar"

Good question, on my way to run a couple tests. Worst case scenario, seems like the "identifier" code-wise is `adafruit_kb2040`, so adding a couple of `codespell:ignore` comments on the few places where "kee boar" might appear does not sound like a big issue anyway

caksoylar (approved):
Thanks, I'll give a formal approval for the docs changes.

petejohanson (commented):
Thanks! Mostly just one comment/question on this.

.codespellrc
```diff
@@ -0,0 +1,4 @@
+[codespell]
+skip = docs/package-lock.json,*.svg
```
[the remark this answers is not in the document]
elpekenin: No reason, no. 

Those files are there just because they threw some error, thus i whitelisted them... There might be some other "special" files we want to ignore too.

PR #2643, opened 2024-11-14
fix: Fix warnings in nanopb encoding code

joelspadin:
The `arg` field on nanopb structs is a `void*` because it is shared between the encode and decode callbacks, even though the encode callback probably should not modify the data. We are passing const data using this non-const pointer, which causes warnings about discarding const. This commit explicitly casts to `void*` to suppress these warnings.


petejohanson (approved):
Thanks!

PR #2656, opened 2024-11-18
docs: remove dangling 0 in sticky keys docs

azaleacolburn:
There was a "0" after a sentence in the Stick-Key docs, all this PR does is remove it. 

If there's some larger-scale way that ZMK plans to remove mistakes like this going forward, then I don't see a reason to merge this PR.


caksoylar (approved):
I'd hope #2642 would catch this but I guess it didn't. Anyway, doesn't hurt to merge the fix.

PR #2664, opened 2024-11-20
Docs: Сorrecting a wrong file name in new-shield.mdx

LeFFaQ:
I happened to notice that in the documentation about creating a new keyboard shield, ``my_keyboard_left.conf``` is repeated twice. I decided to fix this small but minor issue.


Nick-Munnich (approved):
Thanks!

PR #2672, opened 2024-11-25
fix(boards): Disable uart serial node in Xiao BLE by default

### LINKED ISSUE #1887: Turning on USB logging interferes with TX/RX pins on Xiao BLE
I have this issue where turning on USB logging via `CONFIG_ZMK_USB_LOGGING=y` prevents the matrix pin I have at RX (D7/P1.12) from working. I am guessing this is because they are assigned to UART pins at https://github.com/zmkfirmware/zephyr/blob/v3.2.0%2Bzmk-fixes/boards/arm/seeeduino_xiao_ble/seeeduino_xiao_ble-pinctrl.dtsi#L7 and logging turns it on via https://github.com/zmkfirmware/zephyr/blob/v3.2.0%2Bzmk-fixes/boards/arm/seeeduino_xiao_ble/seeeduino_xiao_ble.dts#L17.

I see some shield configs deleting these properties, e.g. https://github.com/zmkfirmware/zmk/blob/main/app/boards/shields/hummingbird/hummingbird.overlay#L13. Should this be something that every shield does, or would it be better to solve it at the `seeeduino_xiao_ble.overlay` level?

caksoylar:
`xiao_serial` node corresponds to `uart0`, which uses D6 and D7 pins by default on Xiaos. When logging or Studio USB snippets are used, they turn on `CONFIG_SERIAL` and this prevents those pins from being used for other things, even though both use serial over USB. 

For RP2040 this is disabled in https://github.com/zmkfirmware/zmk/blob/main/app/boards/seeeduino_xiao_rp2040.overlay#L7, for Hummingbird shield this was disabled at the shield level but was (mistakenly?) removed in https://github.com/zmkfirmware/zmk/commit/e0339a2a57c33b86e351f31b04345ee8c984b075#diff-b0c154eb8d05802c1c61671a7e9e2ca6021c2ec81127f2c57194a84d7a85cdbfL57.

Is there a reason to not disable it by default for ZMK at the board level, like this? I can't think of any. Other boards like nice!nano don't set `status = "okay";` either.

Fixes #1887.

## PR check-list

### DIFF
--- a/app/boards/seeeduino_xiao_ble.overlay
+++ b/app/boards/seeeduino_xiao_ble.overlay
@@ -19,6 +19,8 @@
     };
 };
 
+&xiao_serial { status = "disabled"; };
+
 &adc {
     status = "okay";
 };


petejohanson (approved):
This should be fine. AFAIK, no one is using the serial port on these for anything, and when they do, we can check then if the overlay ordering causes any issues. Thanks!

PR #2680, opened 2024-11-30
fix(behaviors): Macros to include behavior metadata

petejohanson:
Ensure macros can be given metadata/studio display names by including behavior metadata in base macro bindings.


petejohanson:
Don't PR and (not) sleep. Not necessary.

PR #2704, opened 2024-12-11
docs: Move more sections under hardware integration

caksoylar:
Move portions of pages that describe hardware integration parts from feature pages into dedicated pages under Hardware Integration section.

- Defining battery sensing for battery levels
- Underglow and backlight (discussed in https://github.com/zmkfirmware/zmk/pull/2485#pullrequestreview-2327791146)
  - I decided to have a Lighting category index page and two subpages. The index page again emphasizes when to use each system
  - Although I added links to pin control page in both pages, this is (for now) a superficial integration and further refactoring to reduce duplication should be done in another PR

## PR check-list


Nick-Munnich:
I agree that the additional refactoring could be done in a future PR. I think that the config and feature pages ought to also be combined into one. Battery bit LGTM.

caksoylar:
> I think that the config and feature pages ought to also be combined into one. 

How do you envision this? I thought it's nice to have all the features' config as reference under configuration. If we move them to e.g. the feature page then the config reference section will be incomplete.

Nick-Munnich:
> > I think that the config and feature pages ought to also be combined into one.
> 
> How do you envision this? I thought it's nice to have all the features' config as reference under configuration. If we move them to e.g. the feature page then the config reference section will be incomplete.

I was unclear I think: feature pages combined into one and config pages combined into one, making two from four.

caksoylar:
Got it, yeah that makes sense. Will break quite a few links, and I should remember to add redirects.

caksoylar:
I (finally) consolidated the underglow+backlight feature and config pages as well. I added redirects for the old pages, though I don't think they work on local builds or in deploy previews.

While debugging any missing links, I also noticed the missing anchors on the dongle page, opened #2764.

Another extra is I removed an incorrect redirect from `docs/features/encoders` to the keymap section, since IIRC we didn't end up moving that page. I included that in this PR since I am already touching redirects and I couldn't cleanly cherry-pick it.

docs/docs/intro.md
```diff
@@ -28,8 +28,8 @@ ZMK is currently missing some features found in other popular firmware. This tab
...
 | [Encoders](features/encoders.md)                                                                                                                           | ✅  |    ✅     | ✅  |
 | [Display Support](features/displays.md)[^1]                                                                                                                | 🚧  |    🚧     | ✅  |
-| [RGB Underglow](features/underglow.md)                                                                                                                     | ✅  |    ✅     | ✅  |
-| [Backlight](features/backlight.mdx)                                                                                                                        | ✅  |    ✅     | ✅  |
+| [RGB Underglow](features/lighting.md#rgb-underglow)                                                                                                        | ✅  |    ✅     | ✅  |
+| [Backlight](features/lighting.md#backlight)                                                                                                                | ✅  |    ✅     | ✅  |
```
Nick-Munnich, line 32: Thoughts on consolidating this to "LED Lighting" or something similar?

Nick-Munnich (approved):
LGTM overall, just one little note.

docs/docs/intro.md
caksoylar: I can see that, maybe "LED-based Lighting"?

caksoylar: Changed now, thanks!

PR #2711, opened 2024-12-14
fix: Kconfig refactor now works correctly with external modules

Nick-Munnich:
The refactor of 2537 didn't account for external module boards and shields. This fixes it by pulling in the relevant Kconfig bits from `kconfig.zephyr` a bit earlier. Results in some duplication of defaults, but acceptable duplication I'd say.

## PR check-list

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -687,13 +687,22 @@ module = ZMK
 module-str = zmk
 source "subsys/logging/Kconfig.template.log_config"
 
+# This loads ZMK's internal board and shield Kconfigs
 rsource "boards/Kconfig"
 rsource "boards/shields/*/Kconfig.defconfig"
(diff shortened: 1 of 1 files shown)


caksoylar:
Do we want this before ZMK boards so external modules with upstream keyboard configs can override built ins?

Nick-Munnich:
> Do we want this before ZMK boards so external modules with upstream keyboard configs can override built ins?

I believe currently between e.g. internal `corne_left` and external `corne_left` the internal one has priority for the dts files. So I think this order is correct for consistency.

ctranstrum:
Tested this on my external-module-based keyboard and can confirm it works.

petejohanson (approved):
Thanks for the research and fix!

PR #2714, opened 2024-12-15
Feature: input processor behavior invocation

petejohanson:
Adds a new input processor that can trigger behaviors for certain input events.

## PR check-list


Nick-Munnich:
I'm immediately curious about the complex behavior/hold-tap limitations. Could you elaborate a bit?

petejohanson:
> I'm immediately curious about the complex behavior/hold-tap limitations. Could you elaborate a bit?

Those behaviors rely on tracking things based on the key position, which isn't part of the binding event details when there's no associated key position for this. This PR is still a draft because I'm next working on a "event to key position event mapper" processor that will allow certain input events to generate key position events, i.e. "key presses", to trigger keymap processing like normal, for folks interested in extending they keymaps to allow more complex behaviors.

petejohanson:
> > This PR is still a draft because I'm next working on a "event to key position event mapper" processor that will allow certain input events to generate key position events [...]
> 
> I recall a brief discussion on Discord a while ago where I suggested a (pair of) behavior to trigger behaviors from specific locations in the keymap, "Trigger the behavior found at my current index on this layer", "trigger the behavior found on layer X at index X". Though another processor is probably a cleaner way of resolving the issue, but as a backup/if you like the idea.

I want the processor to honor whatever layers are active though, so folks can combine the processor with various layer settings for effect.

petejohanson:
@Nick-Munnich Ok, I went with extending the virtual key positions to support this more extensively, and remove that limitation. I've removed that note, and added a test showing the use with hold tap.

docs/docs/keymaps/input-processors/behaviors.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Pre-Defined Instances
+
+One pre-defined instance of the out-of-band behaviors input processor is available:
+
+| Reference               | Description                                        |
+| ----------------------- | -------------------------------------------------- |
+| `&zip_button_behaviors` | Maps left/right/middle clicks to a given behavior. |
+
+Should you with to update the existing instance to trigger different behaviors for each mouse button, you can override the `bindings` property, e.g.:
```
minusfive, line 26:
```suggestion
Should you wish to update the existing instance to trigger different behaviors for each mouse button, you can override the `bindings` property, e.g.:
```

app/src/pointing/input_processor_behaviors.c
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    for (size_t i = 0; i < cfg->size; i++) {
+        if (cfg->codes[i] == event->code) {
+            struct zmk_behavior_binding_event behavior_event = {
+                .position = ZMK_VIRTUAL_KEY_POSITION_BEHAVIOR_INPUT_PROCESSOR(
+                    state->input_device_index, cfg->index),
+                .timestamp = k_uptime_get(),
+#if IS_ENABLED(CONFIG_ZMK_SPLIT)
+                .source = ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL,
```
caksoylar, line 48: Is there a way to associate the listener with the pointing device that creates the input event? It could let us assign a locality, at least if we assume one pointing device per listener (though that is an incorrect assumption, I think?)

If that's not possible or something we don't want to tackle yet, we should add a note like this one that it always triggers on central: https://zmk.dev/docs/keymaps/combos#advanced-usage

docs/docs/keymaps/input-processors/behaviors.md
```diff
@@ -0,0 +1,69 @@
...
+        zip_right_click_trigger_paste: zip_right_click_trigger_paste {
+            compatible = "zmk,input-processor-behaviors";
+            #input-processor-cells = <0>;
+            codes    = <INPUT_BTN_1>;
+            bindings = <&kp LC(V)  >;
+        };
+    };
+}
+```
+
```
caksoylar, line 56: Can we add an example that uses one of the movement events? Something like "convert pointer movement to arrow keys"? That's one thing that will probably commonly come up as a usage.

As an aside, there might also be requests to convert a certain amount of movement to `n` taps of a key. I can see how that is much more complicated to implement, but mmaybe something to think about.

docs/docs/keymaps/input-processors/behaviors.md
```diff
@@ -0,0 +1,69 @@
...
+### Compatible
+
+The code mapper input processor uses a `compatible` property of `"zmk,input-processor-behaviors"`.
```
caksoylar, line 59:
```suggestion
The behaviors input processor uses a `compatible` property of `"zmk,input-processor-behaviors"`.
```

docs/docs/keymaps/input-processors/behaviors.md
caksoylar: Maybe it is worth at least implementing a min threshold on the accumulated value before a press is triggered? I don't know enough about the input processors to say whether that is easy or not 😅

app/src/pointing/input_processor_behaviors.c
petejohanson: Yeah, I can note it only ever fires on the central. Tracking/identifying will require more work, and I'm not sure it's worth it this first go round.

docs/docs/keymaps/input-processors/behaviors.md
petejohanson: Honestly... At this point this realistically only supports "key" type events with on/off values, not movement events. To your point, we could have some "limit" for values, use the existing remainder tracking to handle this? We'd need special casing for "key" events though, since they should only trigger the behavior release once the same key code is released. Pondering this.

caksoylar: This was due to my confusion thinking it applied to movement type events as well -- some discussion followed in https://discord.com/channels/719497620560543766/845285481888743434/1318326844935901215. I think the outcome is that we'll document this limitation for now.

app/dts/input/processors/behaviors.dtsi
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+#include <zephyr/dt-bindings/input/input-event-codes.h>
+
+/ {
+    zip_button_behaviors: zip_button_behaviors {
+        compatible = "zmk,input-processor-behaviors";
+        #input-processor-cells = <0>;
+        codes =    <INPUT_BTN_0 INPUT_BTN_1 INPUT_BTN_2>;
+        bindings = <&none       &none       &none>;
```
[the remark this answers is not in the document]
caksoylar: Also, is it better to map all five keys?

docs/docs/keymaps/input-processors/behaviors.md
petejohanson: @caksoylar I added a more prominent note to the top of the behaviors input processor to explain the limitation. Thoughts?

app/dts/input/processors/behaviors.dtsi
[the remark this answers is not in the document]
petejohanson: Me only concern with using `&mkp` be default is the potential for confusion for folks thinking they need to use this to get buttons to send HID events, and also in theory the risk of an infinite loop if you applied this to the `&mkp_input_listener`.

Regarding all 5 keys... I considered this. I don't think the 4th and 5th keys get generated nearly as often, and wanted to avoid folks needing to have unneeded placeholders, and more flash space used, for those infrequently used buttons. Folks can always override the `codes` property if they really need to.

docs/docs/keymaps/input-processors/behaviors.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Pre-Defined Instances
+
+One pre-defined instance of the out-of-band behaviors input processor is available:
+
+| Reference               | Description                                        |
+| ----------------------- | -------------------------------------------------- |
+| `&zip_button_behaviors` | Maps left/right/middle clicks to a given behavior. |
+
+Should you wish to update the existing instance to trigger different behaviors for each mouse button, you can override the `bindings` property, e.g.:
```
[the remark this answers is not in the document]
petejohanson: Good call. Added.

docs/docs/keymaps/input-processors/behaviors.md
```diff
@@ -0,0 +1,77 @@
...
+sidebar_label: Behaviors
+---
+
+## Overview
+
+The behaviors input processor is used invoke standard behaviors when certain input events occur; most frequently this is used to trigger behaviors when certain mouse buttons are triggered by physical pointing devices.
+
+:::note
+
+The behaviors input processor is primarily intended for `INPUT_EV_KEY` type of events that have a binary on/off state, not vector types for relative or absolute movements.
```
caksoylar, line 12:
```suggestion
This input processor is primarily intended for `INPUT_EV_KEY` type of events that have a binary on/off state, not vector types for relative or absolute movements.
```
It was a bit repetitive since it mirrors the start of the previous paragraph.

app/src/pointing/input_processor_behaviors.c
petejohanson: Noted.

docs/docs/keymaps/input-processors/behaviors.md
```diff
@@ -0,0 +1,77 @@
...
+
+Should you wish to update the existing instance to trigger different behaviors for each mouse button, you can override the `bindings` property, e.g.:
+
+```dts
+&zip_button_behaviors {
+    bindings = <&kp A &kp B &kp C>;
+};
+```
+
+By default, the `bindings` property maps all the buttons to [`&none`](../behaviors/misc.md#none).
```
caksoylar, line 40:
```suggestion
By default, the `bindings` property maps all the buttons to [`&none`](../behaviors/misc.md#none), so you will want to override the `bindings` property like above if you use this processor by assigning it to an [input listener](usage.md).
```

This might be a bit clearer.

docs/docs/keymaps/input-processors/behaviors.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Behaviors Input Processor
+sidebar_label: Behaviors
+---
+
+## Overview
+
+The behaviors input processor is used invoke standard behaviors when certain input events occur; most frequently this is used to trigger behaviors when certain mouse buttons are triggered by physical pointing devices. For split devices, this is used with listeners on the central, and will never invoke behaviors on the split peripheral devices.
```
caksoylar, line 8: I think this is a bit imprecise when we consider global behaviors, and that they will do something on the peripherals. I like the wording in the note box in https://zmk.dev/docs/keymaps/combos#advanced-usage better.

petejohanson: Fair. Added a note that mimicked that wording.

caksoylar (approved):

PR #1984, opened 2023-11-02
feat: lock a momentary layer

ecstatic-morse:
Resolves #1299.

This adds a `&molock` (momentary layer lock) behavior which causes all active momentary layers _not_ to become deactivated on release.  `&molock` takes a single argument—a layer. If `&molock` is pressed but no momentary layers are active, `&molock LAYER` acts like `&to LAYER`. This can be used as a "panic switch" to return to a known layer
(typically the BASE layer) when some unknown combination of momentary layers are locked. Alternatively, users can press and release a locked momentary layer to disable it. If the "panic switch" behavior is unwanted for some reason, it can be disabled by passing a negative number instead of a layer (e.g. `&molock -1`).

PR #2717, opened 2024-12-16
feat: Allow layer behaviors to "lock" layers on

Nick-Munnich:
Adds a `locking` property to momentary, to, and toggle layer behaviors. Enabled by default for `&to` and `&tog`, a layer being "locked" prevents behaviors without the `locking` property from deactivating the layer.

Conditional layer activation and deactivation is always non-locking, which seems to interact nicely.

Common use case: with this PR you can place e.g. a `&tog_on 1` or a `&to 1` on layer 1, allowing you to "lock" it in place while it is active.

## PR check-list


caksoylar:
> Conditional layer activation and deactivation is always non-locking, which seems to interact nicely.

If you want to get the state of a conditionally activated `then-layer`, it seems useful to me to propagate the locked state from `if-layers` (i.e. it is locked if all of the if-layers are locked). This'd be necessary to replace the momentariness tracking I mentioned on Discord.

Nick-Munnich:
> If you want to get the state of a conditionally activated `then-layer`, it seems useful to me to propagate the locked state from `if-layers` (i.e. it is locked if all of the if-layers are locked). This'd be necessary to replace the momentariness tracking I mentioned on Discord.
>
> Can you lock on a `then-layer` using `&tog`? I would guess not, since it still needs all the `if-layers` to be active.

I realised that tracking that is unnecessary for this to work as expected. Rather, if both `if-layers` are locked then the `then-layer` is semi-locked: If it is deactivated, then on the event triggered by the deactivation the `then-layer` immediately becomes reactivated.

This also has the effect of allowing the `then-layer` to be locked using `&tog`, as if one of the `if-layers` becomes inactive then the conditional layer effect will try but fail to deactivate the `then-layer`.

caksoylar:
I will rewind a bit: I have this [momentariness tracking feature](https://github.com/caksoylar/zmk/commit/65bc8446c589468befdd2f590521a10898606d1d) which is being used to [filter layer change events](https://github.com/caksoylar/zmk/commit/895a3dfea97a7523641e6b7bbb3bbbed0fde9380) that aren't "permanent". I am hoping to replace the tracking of momentariness by checking the "locked" state of the highest active layer instead. For that to work:
- There'd need to be a function like `bool zmk_keymap_layer_locked(zmk_keymap_layer_id_t layer)` that'd return the state of a layer
- Conditional layers would need to set the default locked status (check if all the if-layers are locked), before it might be locked directly by e.g. a `&tog`

Nick-Munnich:
> I am hoping to replace the tracking of momentariness by checking the "locked" state of the highest active layer instead.

I've added what I believe are the necessary parts to get that to work with the locking system.

app/src/behaviors/behavior_momentary_layer.c
```diff
@@ -61,9 +67,16 @@ static const struct behavior_driver_api behavior_mo_driver_api = {
...
-                        POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT, &behavior_mo_driver_api);
+#define KT_INST(n)                                                                                 \
+    static const struct behavior_mo_config behavior_mo_config_##n = {                              \
+        .locking = DT_INST_PROP_OR(n, locking, false),                                             \
+    };                                                                                             \
+    BEHAVIOR_DT_INST_DEFINE(n, behavior_mo_init, NULL, &behavior_mo_data, &behavior_mo_config_##n, \
+                            POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT,                      \
+                            &behavior_mo_driver_api);
+
+DT_INST_FOREACH_STATUS_OKAY(KT_INST)
```
caksoylar, line 80:
```suggestion
#define MO_INST(n)                                                                                 \
    static const struct behavior_mo_config behavior_mo_config_##n = {                              \
        .locking = DT_INST_PROP_OR(n, locking, false),                                             \
    };                                                                                             \
    BEHAVIOR_DT_INST_DEFINE(n, behavior_mo_init, NULL, &behavior_mo_data, &behavior_mo_config_##n, \
                            POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT,                      \
                            &behavior_mo_driver_api);

DT_INST_FOREACH_STATUS_OKAY(MO_INST)
```

app/src/behaviors/behavior_to_layer.c
```diff
@@ -61,7 +66,13 @@ static const struct behavior_driver_api behavior_to_driver_api = {
...
-BEHAVIOR_DT_INST_DEFINE(0, behavior_to_init, NULL, NULL, NULL, POST_KERNEL,
-                        CONFIG_KERNEL_INIT_PRIORITY_DEFAULT, &behavior_to_driver_api);
+#define KT_INST(n)                                                                                 \
+    static const struct behavior_to_config behavior_to_config_##n = {                              \
+        .locking = DT_INST_PROP_OR(n, locking, false),                                             \
+    };                                                                                             \
+    BEHAVIOR_DT_INST_DEFINE(n, behavior_to_init, NULL, NULL, &behavior_to_config_##n, POST_KERNEL, \
+                            CONFIG_KERNEL_INIT_PRIORITY_DEFAULT, &behavior_to_driver_api);
+
+DT_INST_FOREACH_STATUS_OKAY(KT_INST)
```
caksoylar, line 76:
```suggestion
#define TO_INST(n)                                                                                 \
    static const struct behavior_to_config behavior_to_config_##n = {                              \
        .locking = DT_INST_PROP_OR(n, locking, false),                                             \
    };                                                                                             \
    BEHAVIOR_DT_INST_DEFINE(n, behavior_to_init, NULL, NULL, &behavior_to_config_##n, POST_KERNEL, \
                            CONFIG_KERNEL_INIT_PRIORITY_DEFAULT, &behavior_to_driver_api);

DT_INST_FOREACH_STATUS_OKAY(TO_INST)
```

app/src/behaviors/behavior_toggle_layer.c
```diff
@@ -84,6 +85,7 @@ static const struct behavior_driver_api behavior_tog_driver_api = {
 #define KT_INST(n)                                                                                 \
```
caksoylar, line 85:
```suggestion
#define TG_INST(n)                                                                                 \
```
(need to fix line 94 also)

app/include/zmk/keymap.h
```diff
@@ -36,12 +36,14 @@ zmk_keymap_layer_id_t zmk_keymap_layer_index_to_id(zmk_keymap_layer_index_t laye
...
+bool zmk_keymap_layer_locked(zmk_keymap_layer_id_t layer);
 zmk_keymap_layer_index_t zmk_keymap_highest_layer_active(void);
-int zmk_keymap_layer_activate(zmk_keymap_layer_id_t layer);
-int zmk_keymap_layer_deactivate(zmk_keymap_layer_id_t layer);
-int zmk_keymap_layer_toggle(zmk_keymap_layer_id_t layer);
-int zmk_keymap_layer_to(zmk_keymap_layer_id_t layer);
+int zmk_keymap_layer_activate(zmk_keymap_layer_id_t layer, bool locking);
+int zmk_keymap_layer_deactivate(zmk_keymap_layer_id_t layer, bool locking);
+int zmk_keymap_layer_toggle(zmk_keymap_layer_id_t layer, bool locking);
+int zmk_keymap_layer_to(zmk_keymap_layer_id_t layer, bool locking);
```
caksoylar, line 46: It might be worth thinking about adding single param variants as compatibility shims.

app/src/keymap.c
```diff
@@ -146,11 +147,20 @@ static inline int set_layer_state(zmk_keymap_layer_id_t layer_id, bool state) {
         return 0;
     }
 
+    // Non-forcing disables should not change a locked active layer
+    if (!locking && !state && ((_zmk_keymap_layer_locks >> layer_id) & 1)) {
```
caksoylar, line 151:
```suggestion
    if (!locking && !state && (BIT(layer_id) & _zmk_keymap_layer_locks)) {
```
Perhaps a bit clearer? (not tested)

app/src/keymap.c
```diff
@@ -146,11 +147,20 @@ static inline int set_layer_state(zmk_keymap_layer_id_t layer_id, bool state) {
...
+        return ret;
+    }
+
     zmk_keymap_layers_state_t old_state = _zmk_keymap_layer_state;
     WRITE_BIT(_zmk_keymap_layer_state, layer_id, state);
+    if (locking) {
+        WRITE_BIT(_zmk_keymap_layer_locks, layer_id, state);
+    }
     // Don't send state changes unless there was an actual change
     if (old_state != _zmk_keymap_layer_state) {
```
caksoylar, line 161: Should consider whether lock state change should raise a layer_state_changed event, or a different one.

app/src/conditional_layer.c
```diff
@@ -49,22 +49,24 @@ static const struct conditional_layer_cfg CONDITIONAL_LAYER_CFGS[] = {
 static const int32_t NUM_CONDITIONAL_LAYER_CFGS =
     sizeof(CONDITIONAL_LAYER_CFGS) / sizeof(*CONDITIONAL_LAYER_CFGS);
 
-static void conditional_layer_activate(int8_t layer) {
+static uint32_t layer_locked_by_conditional = 0;
```
caksoylar, line 52: Might be nicer to put inside the listener, on line 75 perhaps.

docs/docs/keymaps/behaviors/layers.md
```diff
@@ -125,6 +125,10 @@ Define a new behavior and assign `"on"` or `"off"` to `toggle-mode`:
 
 You can then use `&tog_on` in place of `&tog` whenever you wish to only toggle a layer on, and not toggle it off. An `"off"` version of the behavior can be defined similarly.
 
+## Layer Locking
+
+When the behaviors `&to` and `&tog` toggle a layer on, they will "lock" the layer in the active state. This prevents the layer from being deactivated by any behavior which is not a `&to` or a `&tog`.
```
caksoylar, line 130:
```suggestion
When the behaviors `&to` and `&tog` toggle a layer on, they will "lock" the layer in the active state. This prevents the layer from being deactivated by any behavior which is not a `&to` or a `&tog`.
In particular, if you activate a layer momentarily using e.g. `&mo 1`, tapping e.g. `&tog_on 1` as defined above will prevent the layer from deactivating after releasing `&mo 1`.
```

Might be good to call out this use case explicitly.

docs/docs/keymaps/behaviors/layers.md
```diff
@@ -125,6 +125,10 @@ Define a new behavior and assign `"on"` or `"off"` to `toggle-mode`:
```
caksoylar, line 125: Need to modify `tog_on` behavior in the above lines to add `locking;` right?

caksoylar (commented):
While I haven't tested, implementation looks good to me. I left some clean up remarks and some suggestions to think about. While I like the new default behavior of `&tog` and `&to`, also worth considering that it will technically constitute a breaking change (alongside the API break in `keymap.h`).

Also need to update https://zmk.dev/docs/config/behaviors to add the new property.

docs/docs/keymaps/behaviors/layers.md
caksoylar: Actually would just `&tog 1` work in this case? If so, that's a better example.

Nick-Munnich: I've edited the code slightly to allow `&tog` to work for this, since I think it's a common use case.

app/src/keymap.c
Nick-Munnich: I've added the `locked` property to the event. Good call.

app/include/zmk/keymap.h
Nick-Munnich: Thought about this for a bit. While it could be done, of the modules that I know of (aka those in the collection) the only one that is affected is auto-layer, which already has code in it to track momentariness. I think it *should* be ok to break API here, though @urob maybe can chime in?

urob: Thanks for letting me know. Agree, breaking the API here shouldn't be a big issue for `auto-layer`. (I adapted the upstream versioning for the module releases. So as long as users keep them in sync, there shouldn't be any compatibility issues).

As an aside, it would be great if breaking API changes like this one could be flagged in the release notes, ideally detailing how the API changed. Otherwise it could get tricky to figure out why things break given how many commits the next ZMK release is going to have.

docs/docs/config/behaviors.md
```diff
@@ -154,6 +154,29 @@ You can use the following node to tweak the default behavior:
...
+### Devicetree
+
+Definition file: [zmk/app/dts/bindings/behaviors/zmk,behavior-momentary-layer.yaml](https://github.com/zmkfirmware/zmk/blob/main/app/dts/bindings/behaviors/zmk%2Cbehavior-momentary-layer.yaml)
+
+Applies to: `compatible = "zmk,behavior-momentary-layer"`
+
+| Property         | Type | Description                                                     | Default |
+| ---------------- | ---- | --------------------------------------------------------------- | ------- |
+| `#binding-cells` | int  | Must be `<1>`                                                   |         |
+| `locking`        | bool | Whether the behavior can lock and unlock layers in the on state |         |
```
caksoylar, line 172:
```suggestion
| `locking`        | bool | Whether the behavior can lock and unlock layers in the on state | false   |
```
I think we explicitly put false defaults for bools like this in other places.

docs/docs/config/behaviors.md
```diff
@@ -166,17 +189,41 @@ Definition file: [zmk/app/dts/bindings/behaviors/zmk,behavior-layer-toggle.yaml]
...
 
-| Property         | Type | Description                    | Default |
-| ---------------- | ---- | ------------------------------ | ------- |
-| `#binding-cells` | int  | Must be `<1>`                  |         |
-| `toggle-mode`    |      | One of `on`, `off`, and `flip` | `flip`  |
+| Property         | Type | Description                                                     | Default |
+| ---------------- | ---- | --------------------------------------------------------------- | ------- |
+| `#binding-cells` | int  | Must be `<1>`                                                   |         |
+| `toggle-mode`    |      | One of `on`, `off`, and `flip`                                  | `flip`  |
+| `locking`        | bool | Whether the behavior can lock and unlock layers in the on state |         |
```
caksoylar, line 196:
```suggestion
| `locking`        | bool | Whether the behavior can lock and unlock layers in the on state | false   |
```

docs/docs/config/behaviors.md
```diff
@@ -166,17 +189,41 @@ Definition file: [zmk/app/dts/bindings/behaviors/zmk,behavior-layer-toggle.yaml]
...
+### Devicetree
+
+Definition file: [zmk/app/dts/bindings/behaviors/zmk,behavior-to-layer.yaml](https://github.com/zmkfirmware/zmk/blob/main/app/dts/bindings/behaviors/zmk%2Cbehavior-to-layer.yaml)
+
+Applies to: `compatible = "zmk,behavior-to-layer"`
+
+| Property         | Type | Description                                                     | Default |
+| ---------------- | ---- | --------------------------------------------------------------- | ------- |
+| `#binding-cells` | int  | Must be `<1>`                                                   |         |
+| `locking`        | bool | Whether the behavior can lock and unlock layers in the on state |         |
```
caksoylar, line 219:
```suggestion
| `locking`        | bool | Whether the behavior can lock and unlock layers in the on state | false   |
```

caksoylar (approved):

Nick-Munnich:
Just realised that this makes #1984 obsolete.

omulh:
Hello,

I see that the changes are ready to merge, but I wonder if the `locking` property could additionally be implemented for sticky layers. For the case of activating a sticky layer from within a momentary one, to finally return to the default layer after pressing the corresponding key or the after the timeout expires.

I haven't tried in my keyboard tbh, but as far as I understand, if a sticky layer is activated from a momentary layer, then the sticky layer would be deactivated on release of the &mo key, just as for the to and toggle layers.

nmunnich:
> Hello,
> 
> I see that the changes are ready to merge, but I wonder if the `locking` property could additionally be implemented for sticky layers. For the case of activating a sticky layer from within a momentary one, to finally return to the default layer after pressing the corresponding key or the after the timeout expires.
> 
> I haven't tried in my keyboard tbh, but as far as I understand, if a sticky layer is activated from a momentary layer, then the sticky layer would be deactivated on release of the &mo key, just as for the to and toggle layers.

Hi, sticky layer uses &mo internally. Using a custom &mo with the locking property enabled along with a custom &sl would allow this to work as you desire without further changes to this PR.

omulh:
> 
> Hi, sticky layer uses &mo internally. Using a custom &mo with the locking property enabled along with a custom &sl would allow this to work as you desire without further changes to this PR.

You are right about that. Thanks!

omulh:
Hi @petejohanson,
Any chance this gets merged to main?
It seems everything is ready for that.
I've also tested the implemented changes on my kb and it's working fine from my side.

I don't want to add pressure, just in case this has been forgotten :)
It would be great to have this feature in ZMK!

app/include/zmk/keymap.h
petejohanson: I would agree with this... I believe this needs the commit message to have a `refactor!(core): ` prefix and that will surface the change differently in the release notes that release-please generates.

app/src/behaviors/behavior_momentary_layer.c
```diff
@@ -35,17 +37,22 @@ static const struct behavior_parameter_metadata metadata = {
 };
 
 #endif
+struct behavior_mo_config {
```
petejohanson, line 41:
```suggestion

struct behavior_mo_config {
```

app/src/conditional_layer.c
```diff
@@ -49,28 +49,28 @@ static const struct conditional_layer_cfg CONDITIONAL_LAYER_CFGS[] = {
...
+        zmk_keymap_layer_activate(layer, locking);
     }
 }
 
-static void conditional_layer_deactivate(int8_t layer) {
+static void conditional_layer_deactivate(int8_t layer, bool locking) {
     // This may deactivate a then-layer that's already active via another mechanism (e.g., a
     // momentary layer behavior). However, the same problem arises when multiple keys with the same
     // &mo binding are held and then one is released, so it's probably not an issue in practice.
     if (zmk_keymap_layer_active(layer)) {
```
petejohanson: Should we also have this be conditional on the existing lock state of the layer?

app/src/conditional_layer.c
```diff
@@ -49,28 +49,28 @@ static const struct conditional_layer_cfg CONDITIONAL_LAYER_CFGS[] = {
...
         LOG_DBG("layer %d", layer);
-        zmk_keymap_layer_deactivate(layer);
+        zmk_keymap_layer_deactivate(layer, locking);
     }
 }
 
 static int layer_state_changed_listener(const zmk_event_t *ev) {
     static bool conditional_layer_updates_needed;
-
+    static uint32_t layer_locked_by_conditional = 0;
```
petejohanson: I'm not a *huge* fan of statics hidden within function bodies, as this makes it (IMHO) harder to see all the pieces in play. But I'm open to being convinced otherwise.

app/src/conditional_layer.c
```diff
@@ -101,14 +100,20 @@ static int layer_state_changed_listener(const zmk_event_t *ev) {
             if ((zmk_keymap_layer_state() & mask) == mask) {
                 then_layer_state |= BIT(cfg->then_layer);
             }
+            // Same as above, but for the lock status
+            if ((zmk_keymap_layer_locks() & mask) == mask) {
+                layer_locked_by_conditional |= BIT(cfg->then_layer);
```
petejohanson: I am more and more leaning toward the `WRITE_BIT` macro from Zephyr's `zephyr/sys/util_macro.h` for these bit operations, e.g.:

```suggestion
                WRITE_BIT(layer_locked_by_conditional, cfg->then_layer, true);
```

app/src/conditional_layer.c
```diff
@@ -101,14 +100,20 @@ static int layer_state_changed_listener(const zmk_event_t *ev) {
...
         for (uint8_t layer = 0; layer <= max_then_layer; layer++) {
             if ((BIT(layer) & then_layers) != 0U) {
+                bool locking = (BIT(layer) & layer_locked_by_conditional) != 0U;
                 if ((BIT(layer) & then_layer_state) != 0U) {
-                    conditional_layer_activate(layer);
+                    conditional_layer_activate(layer, locking);
                 } else {
-                    conditional_layer_deactivate(layer);
+                    conditional_layer_deactivate(layer, locking);
+                    layer_locked_by_conditional &= ~(BIT(layer));
```
petejohanson: And then:

```suggestion
                    WRITE_BIT(layer_locked_by_conditional, layer, false);
```

app/src/conditional_layer.c
```diff
@@ -101,14 +100,20 @@ static int layer_state_changed_listener(const zmk_event_t *ev) {
...
             }
+            // Same as above, but for the lock status
+            if ((zmk_keymap_layer_locks() & mask) == mask) {
+                layer_locked_by_conditional |= BIT(cfg->then_layer);
+            }
         }
 
         for (uint8_t layer = 0; layer <= max_then_layer; layer++) {
             if ((BIT(layer) & then_layers) != 0U) {
+                bool locking = (BIT(layer) & layer_locked_by_conditional) != 0U;
```
petejohanson, line 114:
```suggestion
                bool locking = IS_BIT_SET(layer_locked_by_conditional, layer);
```

And then update the line below as well.

app/src/keymap.c
```diff
@@ -146,12 +147,22 @@ static inline int set_layer_state(zmk_keymap_layer_id_t layer_id, bool state) {
         return 0;
     }
 
+    // Non-forcing disables should not change a locked active layer
+    if (!locking && !state && (_zmk_keymap_layer_locks & BIT(layer_id))) {
```
petejohanson, line 146:
```suggestion
    if (!locking && !state && IS_BIT_SET(_zmk_keymap_layer_locks, layer_id)) {
```

app/src/conditional_layer.c
nmunnich: I see your point. Here I decided to keep it simple and place the static next to the already existing static in order to minimise the changes made. I'm happy to pull them both out.

app/src/conditional_layer.c
nmunnich: On thinking about it harder, I think the only purpose for these methods is to have the debug statements trigger only when the layers are actually changed by the conditional layer system. The layer activation/deactivation functions would otherwise suffice on their own, being idempotent. With that as the reasoning for these functions, I agree that this should be conditional on the lock state.

app/src/conditional_layer.c
petejohanson: Yeah, new me wishes old me hadn't added that first one. :grinning:

PR #2724, opened 2024-12-19
feat: Parameterised Mod Morph and Tap Dance

nmunnich:
General idea is to have a `binding-params` array property which maps parameters input to the behavior to the behaviors found in the bindings array. Bit flags and some macros to tidy it all up.

Docs are included, I decided to substitute the references to the non-parameterised version with the parameterised one in the interest of "one way to do it". Current one was moved to a footnote.

In general I think the approach and user experience side of things is good. May add more tests. Unsure whether documentation needs further reworking or if the non-parameterised version should remain prominent.

## PR check-list

- [X] Branch has a [clean commit history](https://zmk.dev/docs/development/contributing/pull-requests#clean-commit-history)
- [X] Additional tests are included, if changing behaviors/core code that is testable.
- [X] [Pre-commit](https://zmk.dev/docs/development/local-toolchain/pre-commit) used to check formatting of files, commit messages, etc.

TODO:
- [ ] Identify a solution to the hold-tap issue.
- [ ] Add behavior metadata for Studio.


app/src/behaviors/behavior_tap_dance.c
```diff
@@ -258,20 +266,28 @@ static int behavior_tap_dance_init(const struct device *dev) {
...
 
 #define TRANSFORMED_BINDINGS(node)                                                                 \
-    {LISTIFY(DT_INST_PROP_LEN(node, bindings), _TRANSFORM_ENTRY, (, ), DT_DRV_INST(node))}
+    {LISTIFY(DT_PROP_LEN(node, bindings), _TRANSFORM_ENTRY, (, ), node)}
+
+#define BREAK_ITEM(i, inst) DT_PROP_BY_IDX(inst, binding_params, i)
+#define BREAK_ZERO(i, inst) 0
 
-#define KP_INST(n)                                                                                 \
+#define KP_INST(inst)                                                                              \
```
caksoylar, line 274: Might be a good opportunity to rename this to e.g. `TD_INST`, similar to other recently updated behaviors.

app/src/behaviors/behavior_mod_morph.c
```diff
@@ -84,26 +90,36 @@ static int behavior_mod_morph_init(const struct device *dev) { return 0; }
...
     }
 
-#define KP_INST(n)                                                                                 \
-    static struct behavior_mod_morph_config behavior_mod_morph_config_##n = {                      \
-        .normal_binding = _TRANSFORM_ENTRY(0, n),                                                  \
-        .morph_binding = _TRANSFORM_ENTRY(1, n),                                                   \
-        .mods = DT_INST_PROP(n, mods),                                                             \
-        .masked_mods = COND_CODE_0(DT_INST_NODE_HAS_PROP(n, keep_mods), (DT_INST_PROP(n, mods)),   \
-                                   (DT_INST_PROP(n, mods) & ~DT_INST_PROP(n, keep_mods))),         \
+#define KP_INST(inst)                                                                              \
```
caksoylar, line 100: Same, e.g. `MM_INST`

docs/docs/keymaps/behaviors/mod-morph.md
```diff
@@ -64,6 +70,37 @@ Example:
...
+### Binding Parameters
+
+The `binding-params` property determines how the parameters passed to the mod-morph behavior are passed on to the behaviors listed in `bindings`.
+It is an array which always has two `BINDING_PARAM(arg1,arg2)` elements - the first corresponds to the first behavior listed in `bindings`, the second corresponding to the second. The number chosen for `argX` determines what the Xth parameter passed to the behavior will be:
+
+- `1`: The Xth parameter passed will be the first parameter that the mod-morph behavior received
+- `2`: The Xth parameter passed will be the second parameter that the mod-morph behavior received
+- `0`: The Xth parameter will be that which is written in the `bindings` array.
+
+Note that any parameters in `bindings` behaviors which are to be replaced should be set to `PLACEHOLDER`.
```
caksoylar, line 82: It is worth noting here that if the behavior takes one/zero params, you should use `0` for `arg2`/`arg2` and `arg1`.

docs/docs/keymaps/behaviors/mod-morph.md
```diff
@@ -64,6 +70,37 @@ Example:
...
+binding-params = <BINDING_PARAM(2,0) BINDING_PARAM(1,0)>;
+```
+
+The first behavior receives the second input parameter as its first parameter, the second behavior receives the first input parameter as its first parameter.
+
+```dts
+binding-params = <BINDING_PARAM(0,0) BINDING_PARAM(0,0)>;
+```
+
+Neither behavior receives any input parameter. Note that mod-morph always requires two parameters to be passed, so you will still need to write e.g. `&my_mm 0 0` in your keymap - the parameters passed will be ignored[^1].
```
caksoylar, line 102: They still get any parameters you defined inside `bindings`, right?

docs/docs/keymaps/behaviors/tap-dance.mdx
```diff
@@ -8,23 +8,75 @@ import TabItem from "@theme/TabItem";
...
+A tap-dance key invokes a different behavior (e.g. `kp`) corresponding to how many times it is pressed.
+For example, you could configure a tap-dance key that acts as `LSHIFT` if tapped once, or Caps _Lock_ if tapped twice.
+The expandability of the number of [`bindings`](#bindings) attached to a particular tap-dance is a great way to add more functionality to a single key, especially for keyboards with a limited number of keys.
+Tap-dances accept two parameters from your keymap, which you can map to the behaviors within as you please.
 
 Tap-dances are designed to resolve immediately when interrupted by another keypress. Meaning, when a keybind is pressed other than any active tap-dances, the tap-dance will activate according to the current value of its counter before the interrupting keybind is registered.
 
-### Configuration
+## Keypress Tap Dance
```
caksoylar, line 18:
```suggestion
## Key Press Tap Dance
```

docs/docs/keymaps/behaviors/tap-dance.mdx
```diff
@@ -8,23 +8,75 @@ import TabItem from "@theme/TabItem";
...
 
-#### `tapping-term-ms`
+ZMK provides a simple two-tap tap dance for you to use. Press it once, and it will output a key press with the first parameter.
+Press it twice within 200ms, and it will output a key press with the second parameter.
+
+### Behavior Binding
+
+- Reference: `&td`
+- Parameters: The keycode usage IDs from the usage page, e.g. `N4` or `A`
```
caksoylar, line 26: This is a bit more confusing than it should be, but it looks like it is copied from https://zmk.dev/docs/keymaps/behaviors/key-press#behavior-binding. I'd still just call these "keycodes" and link to list of keycodes page.

docs/docs/keymaps/behaviors/tap-dance.mdx
```diff
@@ -8,23 +8,75 @@ import TabItem from "@theme/TabItem";
...
 
 The number of bindings in this array also determines the tap-dance's maximum number of keypresses. When a tap-dance reaches its maximum number of keypresses, it will immediately invoke the last behavior in its list of `bindings`, rather than waiting for [`tapping-term-ms`](#tapping-term-ms) to expire before the output is displayed.
 
-### Example Usage
+### `binding-params`
+
+The `binding-params` property determines how the parameters passed to the tap-dance behavior are passed on to the behaviors listed in `bindings`.
+It is an array of `BINDING_PARAM(arg1,arg2)` elements, one for each behavior in `bindings`.
+The Nth `BINDING_PARAM(arg1,arg2)` corresponds to the Nth behavior listed in `bindings`.
```
caksoylar, line 50: This is worded slightly differently compared to mod morph, it is probably worth keeping them same.

app/src/behaviors/behavior_mod_morph.c
```diff
@@ -84,26 +90,36 @@ static int behavior_mod_morph_init(const struct device *dev) { return 0; }
...
+    BUILD_ASSERT((COND_CODE_0(DT_NODE_HAS_PROP(inst, binding_params), (0),                         \
+                              ((((DT_PROP_BY_IDX(inst, binding_params, 0) >> 2) &                  \
+                                 (DT_PROP_BY_IDX(inst, binding_params, 0))) |                      \
+                                ((DT_PROP_BY_IDX(inst, binding_params, 1) >> 2) &                  \
+                                 (DT_PROP_BY_IDX(inst, binding_params, 1)))))) == 0),              \
+                 "Invalid binding parameters");                                                    \
```
caksoylar, line 106: What is this checking exactly? I assume it is checking that `arg1` and `arg2` values are not equal?

If so, it would be nice to make it more explicit, I think it is a bit too clever for me. It can be simplified by at least splitting the assert to two (for binding_params 1 and 2) and a more informative error message can be added.

caksoylar (commented):
I think this is pretty nice. A few things worth discussing:

- Forcing mod-morphs and tap-dances to always have two parameters prevents them from being used in hold-taps, or inside parametrized mod-morph/tap-dances. This is a big limitation, so it might be worth keeping around zero-param versions as the main variants.
- These are a bit inconsistent w.r.t. parametrized macros, they are more flexible so we can't easily unify them. One thing we can do is switch to `PLACEHOLDER` instead of `MACRO_PLACEHOLDER` in the macro docs.

docs/docs/keymaps/behaviors/mod-morph.md
Nick-Munnich: If the binding-params does not alter the parameter, then it gets the parameter inside bindings. Otherwise it will overwrite whatever parameter is inside bindings even if that parameter is not PLACEHOLDER/0.

I suppose it might be worthwhile to add an additional check somewhere to not overwrite non-0 parameters?  I should at least make this more explicit in the docs and maybe write an explicit test for it.

Nick-Munnich:
>     * Forcing mod-morphs and tap-dances to always have two parameters prevents them from being used in hold-taps, or inside parametrized mod-morph/tap-dances. This is a big limitation, so it might be worth keeping around zero-param versions as the main variants.

You should be able to use them inside the others with appropriate 0s being passed in the right places. I'll add some more tests to be sure. 

>     * These are a bit inconsistent w.r.t. parametrized macros, they are more flexible so we can't easily unify them. One thing we can do is switch to `PLACEHOLDER` instead of `MACRO_PLACEHOLDER` in the macro docs.

I think that both hold-tap and parameterised macros could be adjusted to use a `binding-params`, I just wanted to get the ones which aren't yet parameterised done before working on (understanding and) adjusting macro and hold-tap. Definitely agree on switching away from MACRO_PLACEHOLDER.

docs/docs/keymaps/behaviors/tap-dance.mdx
Nick-Munnich: The slightly different wording is due to one of them being of fixed length 2, while the other can vary. If this was extended to hold-tap and macros I'd use the mm wording for hold-tap and the td wording for macros.

pnoulis:
Hi,

Is this every going to get merged? :D 

From what I gather this feature would allow a user to define a generic behavior 
that accepts 2 symbols.  The behavior would output symbol_1 without the modifier being held and
symbol_2 if the opposite is true.

I think this is a very useful feature.

In the meantime is there any other solution to achieving the same behavior?

Could I just define a layer? Would that be sufficient?

Thank you in advance.

snoyer:
> In the meantime is there any other solution to achieving the same behavior?

this PR basically proposes a new syntax for the existing Mod-Morph so you can use that, see https://zmk.dev/docs/keymaps/behaviors/mod-morph

pnoulis:
> > In the meantime is there any other solution to achieving the same behavior?
> 
> this PR basically proposes a new syntax for the existing Mod-Morph so you can use that, see https://zmk.dev/docs/keymaps/behaviors/mod-morph

Thank you for your response.

The way currently things stand a User is allowed to define a `static` behavior.

Example: (from the docs)

bindings = `<&kp B>, <kp C>` -> When the symbol `B` is pressed output `B`.  When the symbol `B` `AND` `MODIFIER LCTL|RCTL` is pressed output `C`.

```
      morph_BC: morph_BC {
            compatible = "zmk,behavior-mod-morph";
            #binding-cells = <0>;
            bindings = <&kp B>, <&kp C>;
            mods = <(MOD_LCTL|MOD_RCTL)>;
        };
```

Essentially this behavior is static, it cannot be parameterized.

Compare it with the next example:

```
behavior_shift: behavior_shift {
  compatible = "zmk,behavior-mod-morph";
  label = "behavior_shift";
  #binding-cells = <2>;
  bindings = <&kp>, <&kp>;
  mods = <(LSHIFT|RSHIFT)>;
};
```

Defines a behavior that accepts 2 arguments (2 symbols).
bindings = `<&kp>, <&kp>`-> When symbol_1 is being pressed output symbol_1. When `symbol_1` is being pressed `AND` modifier `LSHIFT  | RSHIFT` output `symbol_2`.

This behavior is very different from the static ones.

How do you propose achieving the same result?

snoyer:
My point was that, as far as I understand, this PR would let you do the following, with a single parameterized `shift_morph` behavior:
```
/ {
  keymap {
    default_layer {
      bindings = <
        &shift_morph A B    &shift_morph C D    &shift_morph E F
      >;
    };
  };
  behaviors {
    shift_morph: shift_morph {
      compatible = "zmk,behavior-mod-morph-param";
      #binding-cells = <2>;
      bindings = <&kp PLACEHOLDER>, <&kp PLACEHOLDER>;
      mods = <(MOD_LSFT|MOD_RSFT)>;
      binding-params = <BINDING_PARAM(1,0) BINDING_PARAM(2,0)>;
    };
  };
};
```

but in the meantime you could achieve functionally the same thing with multiple non-parameterized `shift_morph_XY` behaviors:
```
/ {
  keymap {
    default_layer {
      bindings = <
        &shift_morph_AB    &shift_morph_CD    &shift_morph_EF
      >;
    };
  };
  behaviors {
    shift_morph_AB: shift_morph_AB {
      compatible = "zmk,behavior-mod-morph";
      #binding-cells = <0>;
      bindings = <&kp A>, <&kp B>;
      mods = <(MOD_LSFT|MOD_RSFT)>;
    };
    shift_morph_CD: shift_morph_CD {
      compatible = "zmk,behavior-mod-morph";
      #binding-cells = <0>;
      bindings = <&kp C>, <&kp D>;
      mods = <(MOD_LSFT|MOD_RSFT)>;
    };
    shift_morph_EF: shift_morph_EF {
      compatible = "zmk,behavior-mod-morph";
      #binding-cells = <0>;
      bindings = <&kp E>, <&kp F>;
      mods = <(MOD_LSFT|MOD_RSFT)>;
    };
  };
};
```

sorry if that wasn't relevant to the question

nmunnich:
What @snoyer describes is indeed what this PR would let you do, and also how to work around it for the time being. However, I wouldn't consider this PR ready for usage yet - somewhere in the force pushing the tests failed, and I still need to fix both hold-tap and allow the parameterised behaviours to be customisable with ZMK Studio. 

I will return to this PR eventually, but given the limited capacity for behaviour-focused PRs I will be waiting until #2717, #2848, and #2943 have been reviewed/merged first.

pnoulis:
Thank you both for your prompt and helpful responses!!!

PR #2730, opened 2024-12-22
feat: shields: add tester_nrf52840dk shield

hlord2000:
Adds similar shield as seen with XIAO, ProMicro to
the nRF52840-DK.

Signed-off-by: Helmut Lord <kellyhlord@gmail.com>

## PR check-list

### DIFF
--- a/app/boards/shields/tester_nrf52840dk/Kconfig.defconfig
+++ b/app/boards/shields/tester_nrf52840dk/Kconfig.defconfig
@@ -1,0 +1,12 @@
+if SHIELD_TESTER_NRF52840DK
+
+config ZMK_KEYBOARD_NAME
+    default "ZMK Tester"
+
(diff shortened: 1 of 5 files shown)


Nick-Munnich:
In general I like the idea of more testers. For dev kits I'm a bit sceptical though. Their purpose is to allow for debugging basic hardware faults by inexperienced users - for the dev kits I would expect that their users are experienced enough to know when something isn't working. Could you provide some more info on why you think this should be added?

hlord2000:
There isn't an easy way to get started with just a DK, please close if not useful

nmunnich (changes requested):
After thinking on it I would like to see this added. Could you adjust so that it makes use of physical layouts like the other testers were updated to do recently? That way you could get started with ZMK Studio using this and a DK.

PR #2741, opened 2025-01-02
feat(split): add peripheral changed event on central

ReFil:
Add a new event for when any peripheral connects or disconnects from the central

## PR check-list

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -95,7 +95,10 @@ target_sources_ifdef(CONFIG_ZMK_BATTERY_REPORTING app PRIVATE src/battery.c)
 
 target_sources_ifdef(CONFIG_ZMK_HID_INDICATORS app PRIVATE src/events/hid_indicators_changed.c)
 
-target_sources_ifdef(CONFIG_ZMK_SPLIT app PRIVATE src/events/split_peripheral_status_changed.c)
+if (CONFIG_ZMK_SPLIT AND (NOT CONFIG_ZMK_SPLIT_ROLE_CENTRAL))
+  target_sources(app PRIVATE src/events/split_peripheral_status_changed.c)
+endif()
+target_sources_ifdef(CONFIG_ZMK_SPLIT_ROLE_CENTRAL app PRIVATE src/events/split_central_peripheral_status_changed.c)
 add_subdirectory(src/split)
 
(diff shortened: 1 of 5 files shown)


app/CMakeLists.txt
```diff
@@ -95,7 +95,10 @@ target_sources_ifdef(CONFIG_ZMK_BATTERY_REPORTING app PRIVATE src/battery.c)
 
 target_sources_ifdef(CONFIG_ZMK_HID_INDICATORS app PRIVATE src/events/hid_indicators_changed.c)
 
-target_sources_ifdef(CONFIG_ZMK_SPLIT app PRIVATE src/events/split_peripheral_status_changed.c)
+if (CONFIG_ZMK_SPLIT AND (NOT CONFIG_ZMK_SPLIT_ROLE_CENTRAL))
+  target_sources(app PRIVATE src/events/split_peripheral_status_changed.c)
+endif()
+target_sources_ifdef(CONFIG_ZMK_SPLIT_ROLE_CENTRAL app PRIVATE src/events/split_central_peripheral_status_changed.c)
```
petejohanson, line 101: Honestly, it'd be clearer if using just conditional branches for both, IMHO:

```suggestion
if (CONFIG_ZMK_SPLIT)
  if(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
    target_sources(app PRIVATE src/events/split_central_peripheral_status_changed.c)
  else()
    target_sources(app PRIVATE src/events/split_peripheral_status_changed.c)
  endif()
endif()
```

app/include/zmk/split/bluetooth/central.h
```diff
@@ -4,6 +4,12 @@
 #include <zephyr/bluetooth/addr.h>
 #include <zmk/behavior.h>
 
+enum peripheral_slot_state {
```
petejohanson, line 7: These should be properly namespaced if published externally like this, e.g. `zmk_split_bt_central_peripheral_slot_state`

petejohanson: I'm also not convinced that the "slot" language should escape this, instead of using the more generic "source" concept used elsewhere.

petejohanson (changes requested):
Thanks for working on this! A few thoughts.

app/CMakeLists.txt
caksoylar: Seems addressed, resolving.

PR #2742, opened 2025-01-02
feat(split): Compile additional events on peripheral

ReFil:
In preparation for the re-do of #2036 events that are useful for displays/indicators on the peripheral should be compiled on both sides so they can be raised on both sides

## PR check-list


ReFil:
This pr is ready for review

nmunnich:
I agree with the changes made here. I wonder if we could take this opportunity to think about some convention about which events get compiled on all split parts Vs which get compiled only on the central?

Personally I think we may as well compile all events on all split parts, unless there is a strong argument otherwise (saving memory on the peripheral is not a convincing argument considering the tiny savings imo).

ReFil:
This PR is still ready for review and feedback prior to rebasing

PR #2753, opened 2025-01-06
feat(shields): Add physical layout for Lotus58

ralex2304:
Added physical layout and studio support for Lotus58.

I've tested changes on my board with this [firmware](https://github.com/ralex2304/lotus58-zmk-config/actions/runs/12624124266) and studio in chrome.

Layout is based on existing zmk layout for Lily58.

### DIFF
--- a/app/boards/shields/lotus58/lotus58.dtsi
+++ b/app/boards/shields/lotus58/lotus58.dtsi
@@ -6,17 +6,19 @@
 
 #include <dt-bindings/zmk/matrix_transform.h>
 
+#include "lotus58-layouts.dtsi"
+
 / {
     chosen {
         zephyr,display = &oled;
         zmk,kscan = &kscan0;
-        zmk,matrix-transform = &default_transform;
+        zmk,physical-layout = &physical_layout0;
     };
 
     default_transform: keymap_transform_0 {
         compatible = "zmk,matrix-transform";
         columns = <16>;
-        rows = <4>;
+        rows = <5>;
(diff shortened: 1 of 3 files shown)


caksoylar:
I am not sure if this should be a "common" layout under `dts/layouts`. It is probably better if it is a bespoke layout under the shield folder, like [hummingbird](https://github.com/zmkfirmware/zmk/tree/main/app/boards/shields/hummingbird).

ralex2304:
Fixed and tested.

app/boards/shields/lotus58/lotus58.dtsi
```diff
@@ -6,17 +6,19 @@
...
         zmk,kscan = &kscan0;
-        zmk,matrix-transform = &default_transform;
+        zmk,physical-layout = &physical_layout0;
     };
 
     default_transform: keymap_transform_0 {
         compatible = "zmk,matrix-transform";
         columns = <16>;
-        rows = <4>;
+        rows = <5>;
```
Nick-Munnich, line 21: Why was this changed?

ralex2304: I'm new to zmk, but as far as I understand, 4 is incorrect here, because board has 5 rows. Also Lily58 has 5 here.

Nick-Munnich: It's supposed to refer to the number of electrical rows, not physical rows, but that also seems to be 5 for the lotus on further investigation. Guess it's incorrect in its current state.

caksoylar (approved):
LGTM

Nick-Munnich (approved):

petejohanson (approved):
Thanks!

PR #2757, opened 2025-01-09
Optimize build workflow

urob:
This PR contains two performance tweaks for the GHA build workflow:

1. Use [treeless clones](https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/) when pulling in dependencies. This reduces the download size from 1.7 GB to 610 MB. 
2. Replace `yaml2json | jq` with `yq`. This saves the installation step shaving off a few more seconds.

The following table summarizes the performance impact for building a `reviung41` with the stock config on ZMK `v0.1`.

|     | cold cache | warm cache | cache size |
| --- | --- | --- | --- |
|upstream    | 5m 28s | 2m 02s | 1.7 GB |
|treeless PR | 3m 51s | 1m 45s | 610 MB |

(See https://github.com/urob/zmk-build-benchmarks/actions for the benchmarks.)

## Note

The dependency change is most impactful on cold caches. As Github caches expire after 7 days this should be a nice speed up for many `build-user-config` workflow runs.

For the upstream repo the gains are more limited since the cache should rarely be cold. On the other hand, there isn't really a downside since none of the workflows interacts with trees that aren't reachable from `HEAD`. For now I also applied the tweak to the ZMK `build` and `test` workflows but happy to revert that.

## PR check-list

### DIFF
--- a/.github/workflows/ble-test.yml
+++ b/.github/workflows/ble-test.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Enable babblesim group filter
         run: west config manifest.group-filter -- +babblesim
       - name: Update modules (west update)
-        run: west update
+        run: west update --fetch-opt=--filter=tree:0
       - name: Export Zephyr CMake package (west zephyr-export)
         run: west zephyr-export
       - name: Build BabbleSim components
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -34,13 +34,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install yaml2json
-        run: python3 -m pip install remarshal
-
       - name: Fetch Build Matrix
         run: |
-          echo "build_matrix=$(yaml2json '${{ inputs.build_matrix_path }}' | jq -c .)" >> $GITHUB_ENV
-          yaml2json "${{ inputs.build_matrix_path }}" | jq
+          echo "build_matrix=$(yq -oj -I0 '${{ inputs.build_matrix_path }}')" >> $GITHUB_ENV
+          yq -oj "${{ inputs.build_matrix_path }}"
 
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -117,7 +117,7 @@ jobs:
 
       - name: West Update
         working-directory: ${{ env.base_dir }}
-        run: west update
+        run: west update --fetch-opt=--filter=tree:0
 
       - name: West Zephyr export
         working-directory: ${{ env.base_dir }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Initialize workspace (west init)
         run: west init -l app
       - name: Update modules (west update)
-        run: west update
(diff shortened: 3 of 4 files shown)


urob:
As an aside, for local development using "blobless" clones gives most of the same performance gains without any of the performance penalties from interacting with trees that aren't reachable from `HEAD`. 

To use them, simply replace `west update` with 
```sh
west update --fetch-opt=--filter=blob:none
```
This reduces download size from 1.7 GB to around 800 MB.

urob:
One more aside: The ZMK's west manifest pulls in **a lot** of dependencies that aren't needed for most builds. For example, most `nice nano v2`-based builds only require 4 modules:
```yaml
- cmsis
- hal_nordic
- lvgl
- tinycrypt
```

Pulling in just those modules reduces build times on a cold cache from currently 5m 28s to 2m 11s!! Since most in-tree boards are nordic boards, this might be an argument to moving other boards to modules and specify those extra dependencies in the module manifest.

|     | cold cache | warm cache | cache size |
| --- | --- | --- | --- |
|upstream    | 5m 28s | 2m 02s | 1.7 GB |
|treeless PR | 3m 51s | 1m 45s | 610 MB |
|minimal + treeless | 2m 11s | 1m 36s | 240 MB |

petejohanson:
> One more aside: The ZMK's west manifest pulls in **a lot** of dependencies that aren't needed for most builds. For example, most `nice nano v2`-based builds only require 4 modules:
> 
> ```yaml
> - cmsis
> - hal_nordic
> - lvgl
> - tinycrypt
> ```
> 
> Pulling in just those modules reduces build times on a cold cache from currently 5m 28s to 2m 11s!! Since most in-tree boards are nordic boards, this might be an argument to moving other boards to modules and specify those extra dependencies in the module manifest.
> 	cold cache 	warm cache 	cache size
> upstream 	5m 28s 	2m 02s 	1.7 GB
> treeless PR 	3m 51s 	1m 45s 	610 MB
> minimal + treeless 	2m 11s 	1m 36s 	240 MB

Yeah, I've been on the fence for how best to handle this... We *mostly* have Nordic folks, but we do also have a fair number of folks on RP2040 and STM32 as well. 

Maybe we cut this down to Nordic, RP2040, STM32, and maybe microchip/samd21 required deps, and allow others to update their config's west.yml as needed? If we go that, route, I'd like to document that process for adding back in other Zephyr modules so it's understood how and why.

petejohanson (approved):
Thanks for the work on this, and for the clear PR summary with the metrics!

A definite win, since this is a common pain point for ZMK users. Thanks!

urob:
> Maybe we cut this down to Nordic, RP2040, STM32, and maybe microchip/samd21 required deps, and allow others to update their config's west.yml as needed? 

I haven't checked RP2040 but STM32 is one of the really big ones... So keeping it upstream would be somewhat costly.

> If we go that, route, I'd like to document that process for adding back in other Zephyr modules so it's understood how and why.

My suggestion would be to tackle this alongside an outsourcing of most boards from upstream ZMK to external modules. (Boards that are currently in upstream should probably be maintained by `zmkfirmware`.)

Then ZMK's `west.yml` only needs to include zephyr modules for the boards remaining in upstream. All other drivers could be sourced from the `west.yml` of the corresponding board module. So in the end, user's will need to update their `west.yml` to include the *board* module but won't need to worry about including the right zephyr drivers.

PR #2729, opened 2024-12-20
fix(pointing): Temp layer threading protection.

petejohanson:
* Ensure all layer operations occur on the system work queue thread.

Fixes: #2719

## PR check-list

- [x] Branch has a [clean commit history](https://zmk.dev/docs/development/contributing/pull-requests#clean-commit-history)
- [x] Additional tests are included, if changing behaviors/core code that is testable.
- [x] Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)
- [x] [Pre-commit](https://zmk.dev/docs/development/local-toolchain/pre-commit) used to check formatting of files, commit messages, etc.
- [x] Includes any necessary [documentation changes](https://zmk.dev/docs/development/contributing/documentation).

Issue #1110, opened 2022-01-31
Wired split keyboard support


matrixik:
I saw in few places users requesting wired connection between split halves of some keyboards using ZMK firmware (like Advantage360 Pro or Glove80) and developers said that it is not supported by ZMK for now so this is tracking issue for this request.

Personally I would also prefer wired connection when I'm using my stationary computer so that I don't need to think about charging or constantly using additional USB  port (beside the one used for central role) when I have LEDs turned on all the time. And then using Bluetooth only on the go.

BeatScherrer:
I'm currently using the nice nano wired without any issues on my sofle v2. firmware on my repo but it's basically default generated. I don't even have batteries attached.

caksoylar:
ZMK doesn't support wired *communication* between two halves. The data connection between the two sides go through Bluetooth, even if you are *powering* each side with wires. Maybe this is not clear [in the FAQ](https://zmk.dev/docs/faq/#does-zmk-support-wired-split) item as well.

ai212983:
Technically it is possible (and was done by Glove80 guys) to treat both halves as separate keyboards. Then one can have fully wired connection via USB splitter for example.

cgoates:
This in progress PR is working toward wired communication between the two halves: https://github.com/zmkfirmware/zmk/pull/1117  It doesn't look like it's ready for testing at this point, but it is actively being developed.

MrMarble:
What about charging the board through the jack cable? That way with only one cable you can charge and use both half at the same time. 

I only have one type-c cable and when the keyboard runs out of battery I have to keep changing the cable to still be able to use it wile charging.

caksoylar:
> What about charging the board through the jack cable? That way with only one cable you can charge and use both half at the same time.
> 
> I only have one type-c cable and when the keyboard runs out of battery I have to keep changing the cable to still be able to use it wile charging.

This is not related to firmware but to the board you are using. As far as I know nice!nano doesn't support charging through TRRS while for instance [Mikoto](https://github.com/zhiayang/mikoto#ext_5v) does.

aureumleonis:
I'm really excited about this feature! Having one side die on me drives me crazy. As far as hardware goes, will it be possible to mix and match? E.g having a nice!nano on on the central side but a plain pro micro (and no battery) on the peripheral?

Percentnineteen:
> 

[...]

PR #2766, opened 2025-01-14
Feature: Full-Duplex Wired Split

petejohanson:
~~Opening this as a draft for now, pending more extensive testing, documentation, etc.~~

~~I've tested this with a very basic ZMK Uno setup, but nothing extensive dailying on a keyboard, etc.~~

Preview docs: https://deploy-preview-2766--zmk.netlify.app/docs/config/split#wired-splits

Has gotten a fair amount of testing at this point, so moving this out of Draft.

## PR check-list

## Other To-Dos

- [x] Investigate report of stuck/broken taps on peripheral Cirque. Recovered by restarting left side only.
- [x] Address stack size issues for RP2040 exacerbated by this PR?


caksoylar:
Related: #1110.

Nick-Munnich (commented):
Couple of notes on the first commit. Haven't dealt with the split ble code bits enough to be a trustworthy reviewer there, but those bits look fine at a glance.

Not looked at the second commit at all.

app/include/zmk/split/transport/types.h
```diff
@@ -0,0 +1,73 @@
...
+    ZMK_SPLIT_TRANSPORT_PERIPHERAL_EVENT_TYPE_SENSOR_EVENT,
+    ZMK_SPLIT_TRANSPORT_PERIPHERAL_EVENT_TYPE_INPUT_EVENT,
+};
+
+struct zmk_split_transport_peripheral_event {
+    union {
+        struct {
+            uint8_t position;
+            uint8_t pressed;
+        } key_position;
```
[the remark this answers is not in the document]
petejohanson: I think probably `key_position_event` would maybe be better.

app/include/zmk/split/bluetooth/central.h
```diff
@@ -1,22 +1,11 @@
...
-
-#if IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
-
-int zmk_split_bt_update_hid_indicator(zmk_hid_indicators_t indicators);
-
-#endif // IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
-
 #if IS_ENABLED(CONFIG_ZMK_SPLIT_BLE_CENTRAL_BATTERY_LEVEL_FETCHING)
 
 int zmk_split_get_peripheral_battery_level(uint8_t source, uint8_t *level);
```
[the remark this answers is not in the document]
petejohanson: Yeah, that's still on my list to move, just was trying to avoid distraction from the core changes until it settled. Good call out.

app/include/zmk/split/central.h
```diff
@@ -0,0 +1,42 @@
...
+
+#endif
+
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_WIRED)
+#define WIRED_PERIPHERAL_COUNT 1
+#else
+#define WIRED_PERIPHERAL_COUNT 0
+#endif
+
+#define ZMK_SPLIT_CENTRAL_PERIPHERAL_COUNT MAX(BLE_PERIPHERAL_COUNT, WIRED_PERIPHERAL_COUNT)
```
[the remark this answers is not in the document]
petejohanson: This will definitely need changing in the event we do that, but I am not interested in trying to *guess* how that will work, and put something half baked now. I'd rather keep it simple, and refactor once we actually need/use it.

app/include/zmk/split/central.h
```diff
@@ -0,0 +1,42 @@
...
+
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
+#include <zmk/hid_indicators_types.h>
+#endif // IS_ENABLED(CONFIG_ZMK_SPLIT_PERIPHERAL_HID_INDICATORS)
+
+int zmk_split_cental_invoke_behavior(uint8_t source, struct zmk_behavior_binding *binding,
```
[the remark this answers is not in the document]
petejohanson: What's wrong with `cental`?! /s

(thanks)

app/src/split/central.c
```diff
@@ -0,0 +1,138 @@
...
+
+        return raise_zmk_sensor_event(sensor_ev);
+    }
+    default:
+        LOG_WRN("GOT AN UNKNOWN EVENT TYPE %d", ev.type);
+        return -ENOTSUP;
+    }
+}
+
+int zmk_split_central_invoke_behavior(uint8_t source, struct zmk_behavior_binding *binding,
```
Nick-Munnich, line 60:
```suggestion
int zmk_split_central_invoke_behavior(uint8_t target, struct zmk_behavior_binding *binding,
```
Minor suggestion, but I just remembered thinking the other day that this name for this variable (and similar variables in functions around the transport) would be nicer. Feel free to ignore and hit resolve.

app/include/linker/zmk-split-transport-central.ld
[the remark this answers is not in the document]
petejohanson: Ok, so Zephyr offers a functionality called [iterable sections](https://docs.zephyrproject.org/3.5.0/kernel/iterable_sections/index.html) that allow you to define a section, with a certain type expected to be placed in that section, and allows other code to add their own entry(ies) to that section. It's a very convenient way to add extension mechanisms, because you decouple the code that needs to consume a set of items, from the code that needs to include items in the set.

In particular, it allows adding iterable sections to be stored in different places, either RAM or ROM, and handles all the linker magic, so things get placed in the right place in flash or memory, etc. as needed.

So, these files "declare" the iterable sections to the linker, and work along with the other C macros to allow inserting items, iterating the items in a given section at runtime, etc.

app/src/split/central.c
petejohanson: I'm on the fence on this one... It requires just a *tad* bit more mental load to understand source/target are the same thing, just from different perspectives, but I can see where you are coming from. Leaving unresolved for now to sleep on it. Thanks.

app/include/zmk/split/transport/types.h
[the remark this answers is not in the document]
petejohanson: Done.

petejohanson:
Ok, taking this out of draft at this point, since it's gotten a fair amount of use and testing by myself and a few others. I just pushed some of the initial docs work as well, which needs review still.

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+See [Configuration Overview](index.md) for instructions on how to change these settings.
+
+## Kconfig
+
+Following [split keyboard](../features/split-keyboards.md) settings are defined in [zmk/app/src/split/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/src/split/Kconfig).
+
+| Config                                       | Type | Description                                                              | Default |
+| -------------------------------------------- | ---- | ------------------------------------------------------------------------ | ------- |
+| `CONFIG_ZMK_SPLIT`                           | bool | Enable split keyboard support                                            | n       |
+| `CONFIG_ZMK_SPLIT_ROLE_CENTRAL`              | bool | `y` for central device, `n` for peripheral                               |         |
```
Nick-Munnich, line 17:
```suggestion
| `CONFIG_ZMK_SPLIT_ROLE_CENTRAL`              | bool | `y` for central device, `n` for peripheral                               |  n       |
```

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+:::note
+
+Hardware UARTs have a few different modes/approaches to sending and receiving data, with different levels of complexity and performance. Not all hardware nor drivers support all modes, so ZMK has code to support different interaction modes with the UART as needed. The default mode should be properly selected based on the platform's report support, but you can choose to override the mode if needed.
+
+:::
+
+| Config                                | Type   | Description                                                       | Default                                                       |
+| ------------------------------------- | ------ | ----------------------------------------------------------------- | ------------------------------------------------------------- |
+| `CONFIG_ZMK_SPLIT_WIRED`              | bool   | Use wired connection to communicate between split keyboard halves | y (if no BLE split and devicetree is set appropriately)       |
+| `ZMK_SPLIT_WIRED_UART_MODE`           | choice | The choice of which UART mode to use for TX/RX of data.           |                                                               |
```
Nick-Munnich, line 51:
```suggestion
```
I don't think we expose choice options to people (yet)

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+| `ZMK_SPLIT_WIRED_UART_MODE_ASYNC`     | bool   | Async (DMA) mode                                                  | y if the driver supports it (excluding nRF52 with known bugs) |
+| `ZMK_SPLIT_WIRED_UART_MODE_INTERRUPT` | bool   | Interrupt mode                                                    | `y` if the hardware supports it                               |
+| `ZMK_SPLIT_WIRED_UART_MODE_POLLING`   | bool   | Polling mode                                                      | `y` it neither other mode is supported                        |
```
Nick-Munnich, line 54: Missing prefix

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Async (DMA) Mode
+
+The following settings only apply when using wired split in async (DMA) mode:
+
+| Config                             | Type | Description                                                 | Default |
+| ---------------------------------- | ---- | ----------------------------------------------------------- | ------- |
+| `ZMK_SPLIT_WIRED_ASYNC_RX_TIMEOUT` | int  | RX Timeout (in microseconds) before reporting received data | 20      |
```
Nick-Munnich, line 62: As above

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Polling Mode
+
+The following settings only apply when using wired split in polling mode:
+
+| Config                              | Type | Description                                          | Default |
+| ----------------------------------- | ---- | ---------------------------------------------------- | ------- |
+| `ZMK_SPLIT_WIRED_POLLING_RX_PERIOD` | int  | Number of ticks between calls to poll for split data | 10      |
```
Nick-Munnich, line 70: As above

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Devicetree
+
+### Wired Split
+
+Wired splits require a properly configured UART to function. If writing a shield, you may be able to use the standard UART already provided by the board, e.g. `&pro_micro_serial`. See shield overlays section in the [new shield guide](../development/hardware-integration/new-shield.mdx#shield-overlays) for details on the UART node labels provided by various interconnects. If creating your own board, or using custom pins for the UART, see the documentation on [pin control](../development/hardware-integration/pinctrl.mdx#additional-examples) to configure the pins for your UART.
```
Nick-Munnich, line 76:
```suggestion
Wired splits require a properly configured UART to function. If writing a shield, you may be able to use the standard UART already provided by the board, e.g. `&pro_micro_serial`. See the shield overlays section in the [new shield guide](../development/hardware-integration/new-shield.mdx#shield-overlays) for details on the UART node labels provided by various interconnects. If you are creating your own board, or using custom pins for the UART, see the documentation on [pin control](../development/hardware-integration/pinctrl.mdx#additional-examples) to configure the pins for your UART.
```

docs/docs/features/split-keyboards.md
```diff
@@ -6,10 +6,9 @@ sidebar_label: Split Keyboards
 ZMK supports setups where a keyboard is split into two or more physical parts (also called "sides" or "halves" when split in two), each with their own controller running ZMK. The parts communicate with each other to work as a single keyboard device.
 
 :::note[Split communication protocols]
-Currently ZMK only supports split keyboards that communicate with each other wirelessly over BLE.
-As such, only controllers that support BLE can be used with ZMK split keyboards.
+Currently ZMK only supports split keyboards that communicate with each other wirelessly over BLE or with a wired, full-duplex UART.
 
-Supporting split communication over wired protocols is planned, allowing for ZMK split keyboards using non-wireless controllers.
+Future single-wire, half-duplex UART support, which is planned, will allow using wired ZMK with designs like Corne, Sweep, etc. that use only a single GPIO pin for bidirectional communication between split sides.
```
Nick-Munnich, line 11: This is the bit that was commented on on Discord. Best to mark it down as experimental initially, and change it later, rather than implying full support from the get-go.

Nick-Munnich (commented):
Minor comments on the docs.

docs/docs/config/split.md
caksoylar, line 76: The first one should probably link to https://zmk.dev/docs/development/hardware-integration/pinctrl#predefined-nodes instead, which shows non-GPIO bus labels.

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+
+Once you have a properly configured UART device, it needs to be assigned to the `zmk,split-uart` property in the `chosen` node. For example:
+
+```dts
+/ {
+    chosen {
+      zmk,split-uart = &pro_micro_serial;
+    };
+}
+```
```
Nick-Munnich, line 86: I think you move this over to the new shield guide, as a subsection under the chosen node (make sure that the headers don't break the TOC on the right) and rephrase it to flow with the new shield guide. Then you can make this text much more succinct.

That way readers of the new shield guide don't have to click to come over here, and then click again to go to pinctrl if they're doing somethign more advanced. Readers of this section should be more prepared to do click-hopping, so to speak.

docs/docs/config/split.md
petejohanson: Yeah, good call, forgot that had moved. Updated.

docs/docs/features/split-keyboards.md
petejohanson: Tweaked the wording here. Thoughts?

docs/docs/config/split.md
```diff
@@ -0,0 +1,85 @@
...
+
+| Config                                     | Type | Description                                          | Default |
+| ------------------------------------------ | ---- | ---------------------------------------------------- | ------- |
+| `CONFIG_ZMK_SPLIT_WIRED_POLLING_RX_PERIOD` | int  | Number of ticks between calls to poll for split data | 10      |
+
+## Devicetree
+
+### Wired Split
+
+Wired splits require a properly configured UART to function. If writing a shield, you may be able to use the standard UART already provided by the board, e.g. `&pro_micro_serial`. See [predefined nodes](../development/hardware-integration/pinctrl#predefined-nodes) for details on the UART node labels provided by various interconnects. If you are creating your own board, or using custom pins for the UART, see the documentation on [pin control](../development/hardware-integration/pinctrl.mdx#additional-examples) to configure the pins for your UART.
```
caksoylar, line 75:
```suggestion
Wired splits require a properly configured UART to function. If writing a shield, you may be able to use the standard UART already provided by the board, e.g. `&pro_micro_serial`. See [predefined nodes](../development/hardware-integration/pinctrl.mdx#predefined-nodes) for details on the UART node labels provided by various interconnects. If you are creating your own board, or using custom pins for the UART, see the documentation on [pin control](../development/hardware-integration/pinctrl.mdx#additional-examples) to configure the pins for your UART.
```

docs/docs/config/split.md
petejohanson: Ok, so I added to the new shield guide, but didn't really see much way to trim this down without just fully removing it, so I left it as is.

Thoughts?

app/include/zmk/split/bluetooth/central.h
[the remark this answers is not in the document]
petejohanson: Ok, refactored this to cover battery levels over any transport, and added wired handling for this.

app/dts/bindings/zmk,wired-split.yaml
Nick-Munnich, line 1: Looking at this file, I'm not a fan of the current method of selecting a UART as chosen. It seems to me that both approaches will exist in the future simultaneously if that is kept. I think while it might be a *little* more complicated *now*, making wired split only go through this node will simplify things in the future. No reason to introduce two ways of doing the same thing from the get-go imo.

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -511,6 +511,19 @@ If all of your physical layouts use the same `kscan` node under the hood, you ca
...
+
+```dts
+/ {
+    chosen {
+        zmk,split-uart = &pro_micro_serial;
+        // Other chosen items
+    };
+};
+```
+
```
Nick-Munnich, line 524: Would be good to have a direct link to the predefined nodes here.

docs/docs/config/split.md
Nick-Munnich: Resolving this thread so that we can discuss it on the actual file in question,

Nick-Munnich (commented):
Some comments. Still working my way through codey-bits slowly, but can't seem to write two reviews at the same time.

app/src/split/wired/wired.h
```diff
@@ -0,0 +1,95 @@
...
+void zmk_split_wired_poll_out(struct ring_buf *tx_buf, const struct device *uart);
+
+int zmk_split_wired_poll_in(struct ring_buf *rx_buf, const struct device *uart,
+                            struct k_work *process_data_work,
+                            zmk_split_wired_process_tx_callback_t process_data_cb,
+                            size_t envelope_size);
+
+#endif
+
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_WIRED_UART_MODE_INTERRUPT)
```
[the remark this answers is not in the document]
petejohanson: I had originally done this, but it makes it much harder to move/reorganize files it they become context dependent like that, at the expense of an extra few `#endif`s. There's also a theoretical time when we have multiple UARTs (e.g. one PIO based and one dedicated PL011 UART on RP2040, and we need to refactor things to allow multiples with the type set in DTS, for example, so I'd prefer to leave this .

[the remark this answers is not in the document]
Nick-Munnich: I'd suggested this because `ZMK_SPLIT_WIRED_UART_MODE` is a `choice`. I think if there is a world where we would want multiple enabled at the same time, then it should not be marked as `choice`. In that event, we should rather have the Kconfig mark interrupt and async as mutually exclusive flags due to the note on the Zephyr UART page.

app/dts/bindings/zmk,wired-split.yaml
petejohanson: Yeah... I'm open to this. I had originally suspected I wouldn't get to anything that required the dedicated node soon enough, but now that I have, I'm open to just documenting the dedicated node and using that as the canonical way.

docs/docs/development/hardware-integration/new-shield.mdx
petejohanson: Added.

docs/docs/config/split.md
```diff
@@ -0,0 +1,86 @@
...
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_PRIORITY`              | int  | Priority of the BLE split peripheral notify thread                         | 5                                          |
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_POSITION_QUEUE_SIZE`   | int  | Max number of key state events to queue to send to the central             | 10                                         |
+
+### Wired Splits
+
+Following wired [split keyboard](../features/split-keyboards.md) settings are defined in [zmk/app/src/split/wired/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/src/split/wired/Kconfig).
+
+:::note
+
+Hardware UARTs have a few different modes/approaches to sending and receiving data, with different levels of complexity and performance. Not all hardware nor drivers support all modes, so ZMK has code to support different interaction modes with the UART as needed. The default mode should be properly selected based on the platform's report support, but you can choose to override the mode if needed.
```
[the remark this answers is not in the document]
petejohanson: Expanded on this a ton, hopefully it's clear.

docs/docs/features/split-keyboards.md
```diff
@@ -6,7 +6,7 @@ sidebar_label: Split Keyboards
 ZMK supports setups where a keyboard is split into two or more physical parts (also called "sides" or "halves" when split in two), each with their own controller running ZMK. The parts communicate with each other to work as a single keyboard device.
 
 :::note[Split communication protocols]
-Currently ZMK only supports split keyboards that communicate with each other wirelessly over BLE.
+ZMK supports split keyboards that communicate with each other wirelessly over BLE, and includes experimental wired split support.
```
Nick-Munnich, line 9:
```suggestion
ZMK supports split keyboards that communicate with each other wirelessly over BLE.
```
Don't think the wired bit is necessary considering the next sentence.

docs/docs/config/split.md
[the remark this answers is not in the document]
Nick-Munnich: Perfect. By the sounds of it the nrf52 should use polling until we upgrade to 3.7?

docs/docs/features/split-keyboards.md
petejohanson: Tweaked.

app/src/split/wired/wired.h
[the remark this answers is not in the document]
Nick-Munnich: If you wanted to ensure that there isn't an unsupported combination using Kconfig, couldn't you just use `$(dt_compat_any_has_prop,<compatible string>,<prop>[,<value>])`? 

I'm always in favor of moving things to be more future-proof while we can do so without breaking user setups (aside from beta testers obv). Especially since it's a *when* and not an *if*.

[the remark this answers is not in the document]
petejohanson: We can probably nix' the `choice` piece today, and use the fact that the "other modes" all have APIs that should return `-ENOTSUP` when first setting up a device, so we can have multiple enabled and still have fallbacks without duplicating things in DTS.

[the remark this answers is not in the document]
petejohanson: Fundamentally, I don't think this Kconfig would require any major breakage when we do (eventually, I don't see it happening near term) support multiple UART peripherals, that we should get derailed on this aspect today (as opposed to the `chosen` discussion which I'm leaning toward agreeing with you on)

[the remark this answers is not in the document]
petejohanson: We could avoid some risk here by avoiding giving the `choice`  a name so no one attempts to override it in board/shield defconfig files.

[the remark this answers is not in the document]
Nick-Munnich: To be clear, I'm not opposed to the kconfig stuff existing, I just don't think it should exist as a user facing thing. I guess it would make sense to keep the mutually exclusive modes as a choice, but leave polling out of it. Something like
```
config POLL
 default $(dt_compat_any_has_prop,<compatible string>,<prop>[,<value>])

choice UART_FANCY

config ASYNC
 default $(dt_compat_any_has_prop,<compatible string>,<prop>[,<value>])

...
```
and then put all of that in the standard Kconfig, not in Kconfig.defaults.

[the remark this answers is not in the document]
petejohanson: I don't see what compatible or props you expect to check on? Can you fill that out, because I'm not seeing your vision without it.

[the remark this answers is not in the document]
Nick-Munnich: $(dt_compat_any_has_prop,zmk,wired-split,uart-mode,polling)
I was imagining something like this. (I do hope I haven't pulled something impossible out of thin air)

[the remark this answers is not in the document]
petejohanson: Yeah, and my issue with that is that putting that polling mode *in the devicetree* leads to folks who can set a mode in DTS that the UART they're using doesn't support, and we end up with a major mismatch that we probably can only sorta catch at compile time, making it painful at best. The only two ways that "what does my UART support" is surfaced are:

* Each UART driver Kconfig will select some symbols that indicate what is supported, Sadly, this isn't "scoped", but global, so if *any* enabled UART supports async/DMA, then we'll have that Kconfig symbol enabled. See https://github.com/zmkfirmware/zephyr/blob/v3.5.0%2Bzmk-fixes/drivers/serial/Kconfig.sam0#L12 for an example
* At runtime, if you try to use the async API on a serial device that doesn't support async, we'll get back an `-ENOTSUP`.

So... adding some random DTS prop for this then *further* muddies things, which I'd prefer to avoid.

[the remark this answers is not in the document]
petejohanson: Which is why I suggested we:

* Not over-complicate this now, if we can have something *today* that won't require breakage later,
* Not try to dictate this with additional props, but eventually support having multiple "Mode codes" enabled, that properly fall back to the "less ideal" mode after attempting to use the more efficient mode and checking for `-ENOTSUP` at runtime.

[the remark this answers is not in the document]
Nick-Munnich: I see, all of that makes sense, thanks for the thorough explanation. I'm still slightly iffy on the mismatch between what the Kconfig states and what the preprocessor states, but you're right in that it could be resolved when it comes to that.

Sorry for the massive sidetracking here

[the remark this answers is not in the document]
petejohanson: No worries. I *am* going to remove the name/symbol for the choice, so no one attempts to override the default for it, to avoid breakage later once we support multiple here, so still good coming from it!

docs/docs/config/split.md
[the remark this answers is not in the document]
Nick-Munnich: Probably best to have a hazard box in features/split-keyboards, and then link to it everywhere that wired split is mentioned.

caksoylar:
Note for docs: https://zmk.dev/docs/faq#does-zmk-support-wired-split needs to be updated (in this PR or a future one)

docs/docs/config/split.md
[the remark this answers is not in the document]
petejohanson: Added a warning. Would appreciate your thoughts.

petejohanson:
> Note for docs: https://zmk.dev/docs/faq#does-zmk-support-wired-split needs to be updated (in this PR or a future one)

Tweaked that note, would appreciate feedback. Thanks!

app/dts/bindings/zmk,wired-split.yaml
petejohanson: Pushed this tweak in a few spots, still need to figure out best approach to the new shield guide.

docs/docs/config/split.md
```diff
@@ -82,12 +82,13 @@ The following settings only apply when using wired split in polling mode:
...
-Once you have a properly configured UART device, it needs to be assigned to the `zmk,split-uart` property in the `chosen` node. For example:
+Once you have a properly configured UART device, it needs to be assigned in a new node with a compatible value of `zmk,wired-split`. For example:
 
 ```dts
 / {
-    chosen {
-      zmk,split-uart = &pro_micro_serial;
+    wired_split {
+      compatible = "zmk,wired-split";
+      device = <&pro_micro_serial>;
```
caksoylar, line 91:
```suggestion
        compatible = "zmk,wired-split";
        device = <&pro_micro_serial>;
```

docs/docs/faq.md
```diff
@@ -55,7 +55,7 @@ Please see the [explainer on boards & shields](development/hardware-integration/
 
 ### Does ZMK support wired split?
 
-Currently, ZMK only supports wireless split, but wired split is possible and we welcome contributions!
+Currently, ZMK only supports wireless split. Experimental wired split support for some specific hardware designs is available for advanced users to test.
```
caksoylar, line 58:
```suggestion
Currently, ZMK only supports wireless [split keyboards](features/split-keyboards.md). Experimental wired split support for some specific hardware designs is available for advanced users to test.
```

docs/docs/features/split-keyboards.md
```diff
@@ -6,10 +6,17 @@ sidebar_label: Split Keyboards
...
+ZMK supports split keyboards that communicate with each other wirelessly over BLE.
+
+Full-duplex UART, wired split support is currently experimental, and is available for advanced/technical users to test.
+
+Future single-wire, half-duplex UART support, which is planned, will allow using wired ZMK with designs like Corne, Sweep, etc. that use only a single GPIO pin for bidirectional communication between split sides.
+:::
+
+:::warning[Hot Plugging Cables]
+
+Many popular cables, in particular, TRRS/TRS cables, can cause irreparable damage to controllers if they are inserted or removed when power is already present on them. Whether or not you are using the wired split functionality or not, _never_ insert or remove such a cable when a controller is powered by USB _or_ battery.
```
caksoylar, line 18:
```suggestion
Many popular wired interconnects used between split keyboards, in particular, TRRS/TRS cables, can cause irreparable damage to controllers if they are inserted or removed when power is already present on them. Whether or not you are using the wired split functionality or not, _never_ insert or remove such a cable when a controller is powered by USB _or_ battery.
```

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -511,6 +511,19 @@ If all of your physical layouts use the same `kscan` node under the hood, you ca
 
 :::
 
+Additionally, if testing the experimental [wired split](../../features/split-keyboards.md) support, you should assign a [predefined](./pinctrl.mdx#predefined-nodes) or [pinctrl configured](./pinctrl.mdx) UART to the `zmk,split-uart` chosen property:
```
caksoylar, line 514: Just leaving a comment here as a reminder to update this section.

docs/docs/config/split.md
Nick-Munnich: Do we not also mostly have `phandle` types written without the `<>`?

caksoylar: That's the `path` type, usually used for chosen and aliases nodes: https://docs.zephyrproject.org/3.5.0/build/dts/bindings-syntax.html#type

docs/docs/features/split-keyboards.md
petejohanson: I'm on the fence using "interconnect" here since we use that term elsewhere to describe board/shield abstractions.

docs/docs/development/hardware-integration/new-shield.mdx
petejohanson: I added this in a new section, since it's no longer just an additional chosen node... But that means a placeholder text for unibody shields... Thoughts?

app/dts/bindings/zmk,wired-split.yaml
petejohanson: Ok, removed the `chosen` use, tweaked the docs, warned folks on Discord. Now the dedicated node is the only way to configure this.

docs/docs/features/split-keyboards.md
caksoylar: Fair, maybe something like "physical connectors" instead of "wired interconnects"?

docs/docs/config/split.md
```diff
@@ -0,0 +1,94 @@
...
+| `CONFIG_ZMK_SPLIT_BLE_CENTRAL_SPLIT_RUN_QUEUE_SIZE`     | int  | Max number of behavior run events to queue to send to the peripheral(s)    | 5                                          |
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_STACK_SIZE`            | int  | Stack size of the BLE split peripheral notify thread                       | 756                                        |
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_PRIORITY`              | int  | Priority of the BLE split peripheral notify thread                         | 5                                          |
+| `CONFIG_ZMK_SPLIT_BLE_PERIPHERAL_POSITION_QUEUE_SIZE`   | int  | Max number of key state events to queue to send to the central             | 10                                         |
+
+### Wired Splits
+
+Following wired [split keyboard](../features/split-keyboards.md) settings are defined in [zmk/app/src/split/wired/Kconfig](https://github.com/zmkfirmware/zmk/blob/main/app/src/split/wired/Kconfig).
+
+:::note
```
caksoylar, line 42: I think I'd make these regular paragraphs without the note box, then move "Following wired [split keyboard](../features/split-keyboards.md) settings..." to below, right above the table.

docs/docs/config/split.md
```diff
@@ -0,0 +1,94 @@
...
+  - STM32 (e.g. stm32f072)
+
+:::
+
+| Config                                       | Type | Description                                                       | Default                                                       |
+| -------------------------------------------- | ---- | ----------------------------------------------------------------- | ------------------------------------------------------------- |
+| `CONFIG_ZMK_SPLIT_WIRED`                     | bool | Use wired connection to communicate between split keyboard halves | y (if no BLE split and devicetree is set appropriately)       |
+| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_ASYNC`     | bool | Async (DMA) mode                                                  | y if the driver supports it (excluding nRF52 with known bugs) |
+| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_INTERRUPT` | bool | Interrupt mode                                                    | `y` if the hardware supports it                               |
+| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_POLLING`   | bool | Polling mode                                                      | `y` it neither other mode is supported                        |
```
caksoylar, line 61:
```suggestion
| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_INTERRUPT` | bool | Interrupt mode                                                    | y if the hardware supports it                               |
| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_POLLING`   | bool | Polling mode                                                      | y it neither other mode is supported                        |
```
Consistency pass

docs/docs/config/split.md
```diff
@@ -0,0 +1,94 @@
...
+
+Once you have a properly configured UART device, it needs to be assigned in a new node with a compatible value of `zmk,wired-split`. For example:
+
+```dts
+/ {
+    wired_split {
+        compatible = "zmk,wired-split";
+        device = <&pro_micro_serial>;
+    };
+}
```
caksoylar, line 93:
```suggestion
};
```

docs/docs/config/split.md
```diff
@@ -0,0 +1,94 @@
...
+| ------------------------------------------ | ---- | ---------------------------------------------------- | ------- |
+| `CONFIG_ZMK_SPLIT_WIRED_POLLING_RX_PERIOD` | int  | Number of ticks between calls to poll for split data | 10      |
+
+## Devicetree
+
+### Wired Split
+
+Wired splits require a properly configured UART to function. If writing a shield, you may be able to use the standard UART already provided by the board, e.g. `&pro_micro_serial`. See [predefined nodes](../development/hardware-integration/pinctrl.mdx#predefined-nodes) for details on the UART node labels provided by various interconnects. If you are creating your own board, or using custom pins for the UART, see the documentation on [pin control](../development/hardware-integration/pinctrl.mdx#additional-examples) to configure the pins for your UART.
+
+Once you have a properly configured UART device, it needs to be assigned in a new node with a compatible value of `zmk,wired-split`. For example:
```
caksoylar, line 85:
```suggestion
Once you have a properly configured UART device, it needs to be assigned in a new node with a compatible value of `"zmk,wired-split"`. For example:
```

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -38,6 +38,22 @@ export const SplitInvisTabs = (props) => (
 
 );
 
+export const UnibodyInvisTabs = (props) => (
```
caksoylar, line 41: This component doesn't seem necessary and is unused, right?

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -511,6 +527,28 @@ If all of your physical layouts use the same `kscan` node under the hood, you ca
 
 :::
 
+### Wired Split
+
+<SplitTabs>
+<TabItem value="unibody">This is only required for wired split keyboards</TabItem>
```
caksoylar, line 533:
```suggestion
<TabItem value="unibody">This is only required for wired split keyboards.</TabItem>
```

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -511,6 +527,28 @@ If all of your physical layouts use the same `kscan` node under the hood, you ca
...
+<TabItem value="split">
+
+If testing the experimental [wired split](../../features/split-keyboards.md) support, you should assign a [predefined](./pinctrl.mdx#predefined-nodes) or [pinctrl configured](./pinctrl.mdx) UART to the `device` property of a new node with `compatible` of `zmk,wired-split`:
```
caksoylar, line 536:
```suggestion
If testing the experimental [wired split](../../features/split-keyboards.md) support, you should assign a [predefined](./pinctrl.mdx#predefined-nodes) or [pinctrl configured](./pinctrl.mdx) UART to the `device` property of a new node with `compatible` value of `"zmk,wired-split"`:
```

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -511,6 +527,28 @@ If all of your physical layouts use the same `kscan` node under the hood, you ca
...
+
+```dts
+/ {
+    wired_split {
+      compatible = "zmk,wired-split";
+      device = <&pro_micro_serial>;
+    };
+}
```
caksoylar, line 544:
```suggestion
};
```

docs/docs/features/split-keyboards.md
Nick-Munnich: Personally I think this particular warning should be made *more* specific, not less, as this problem specifically affects TRS-style cables. Things like USB, RJ45, MIDI, etc are all exempt.

petejohanson: > Personally I think this particular warning should be made _more_ specific, not less, as this problem specifically affects TRS-style cables. Things like USB, RJ45, MIDI, etc are all exempt.

Beyond my original wording?

Nick-Munnich: Ideally yes, but I was thinking about it a bunch yesterday and I couldn't find a phrasing I liked. Essentially making the warning sound more explicit towards the "bad" cable types, rather than warning against all. I think it's fine as-is though.

What I was also considering is a "unless explicitly stated otherwise by the designer" because I think you *can* create a circuit which doesn't have this sort of flaw. I worry that might overcomplicate the warning and invite lots of questions from inexperienced folks though, opening up a can of worms we dont want to tackle (yet)

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -511,6 +511,28 @@ If all of your physical layouts use the same `kscan` node under the hood, you ca
...
+<TabItem value="split">
+
+If testing the experimental [wired split](../../features/split-keyboards.md) support, you should assign a [predefined](./pinctrl.mdx#predefined-nodes) or [pinctrl configured](./pinctrl.mdx) UART to the `device` property of a new node with `compatible` value of `"zmk,wired-split"`:
+
+```dts
+/ {
+    wired_split {
+      compatible = "zmk,wired-split";
+      device = <&pro_micro_serial>;
+    };
```
caksoylar, line 527:
```suggestion
    wired_split {
        compatible = "zmk,wired-split";
        device = <&pro_micro_serial>;
    };
```
Sorry, missed this on the previous pass.

docs/docs/config/split.md
petejohanson: Tweaked. Thought?

docs/docs/development/hardware-integration/new-shield.mdx
```diff
@@ -511,6 +511,28 @@ If all of your physical layouts use the same `kscan` node under the hood, you ca
...
+/ {
+    wired_split {
+        compatible = "zmk,wired-split";
+        device = <&pro_micro_serial>;
+    };
+};
+```
+
+See the [wired split](../../config/split.md#wired-split) configuration for more details.
+
```
caksoylar, line 532:
```suggestion

For wireless split keyboards, this step is **not** required and might want to be skipped, especially since the UART pins on your controller might already be in use for other functionality.
```

It doesn't hurt to be clearer here, since people blindly copy paste code blocks and Pro Micro UART pins are D1/D0, I believe.

caksoylar (approved):
One last note, otherwise docs LGTM.

docs/docs/development/hardware-integration/new-shield.mdx
caksoylar: Open question whether we say "should be skipped" instead.

petejohanson: > Open question whether we say "should be skipped" instead.

I think probably that wording is better, to be really clear.

petejohanson: Added with the tweaked language.

app/include/zmk/split/transport/types.h
```diff
@@ -0,0 +1,78 @@
...
+    } data;
+} __packed;
+
+enum zmk_split_transport_central_command_type {
+    ZMK_SPLIT_TRANSPORT_CENTRAL_CMD_TYPE_POLL_EVENTS,
+    ZMK_SPLIT_TRANSPORT_CENTRAL_CMD_TYPE_INVOKE_BEHAVIOR,
+    ZMK_SPLIT_TRANSPORT_CENTRAL_CMD_TYPE_SET_PHYSICAL_LAYOUT,
+    ZMK_SPLIT_TRANSPORT_CENTRAL_CMD_TYPE_SET_HID_INDICATORS,
+} __packed;
+
```
Nick-Munnich, line 57: I wonder if it might make sense to abstract things in the transport a bit further, taking some ideas from https://github.com/zmkfirmware/zmk/pull/2036#discussion_r1771981768. Probably for a future PR, but might be worth a bit of thought to see if there're any minor changes that could be done to make that work easier in the future.

Nick-Munnich (commented):
Last pair of minor queries from me. I think it's unlikely I'll notice any further things from looking at it more.

I will say I don't think this should be merged until we've had another release, which I consider #2729 to be a prerequisite for.

app/src/split/wired/central.c
```diff
@@ -0,0 +1,335 @@
...
+
+#endif
+
+static void begin_tx(void) {
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_WIRED_UART_MODE_INTERRUPT)
+    uart_irq_tx_enable(uart);
+#elif IS_ENABLED(CONFIG_ZMK_SPLIT_WIRED_UART_MODE_ASYNC)
+    zmk_split_wired_async_tx(&async_state);
+#else
+    k_work_submit(&wired_central_tx_work);
```
paulshir, line 112: This isn't defined until line 240 so fails building

app/src/split/wired/central.c
```diff
@@ -0,0 +1,335 @@
...
+}
+
+#elif IS_ENABLED(CONFIG_ZMK_SPLIT_WIRED_UART_MODE_POLLING)
+
+static void send_pending_tx_work_cb(struct k_work *work) {
+    zmk_split_wired_poll_out(&tx_buf, uart);
+}
+
+static void read_timer_cb(struct k_timer *_timer) {
+    zmk_split_wired_poll_in(&rx_buf, uart, &publish_events, sizeof(struct event_envelope));
```
paulshir, line 236: Signature has changed. Should it be`zmk_split_wired_poll_in(&rx_buf, uart, &publish_events, NULL, sizeof(struct event_envelope));`?

app/src/split/wired/peripheral.c
```diff
@@ -0,0 +1,303 @@
...
+#elif IS_ENABLED(CONFIG_ZMK_SPLIT_WIRED_UART_MODE_POLLING)
+
+static void send_pending_tx_work_cb(struct k_work *work) {
+    zmk_split_wired_poll_out(&chosen_tx_buf, uart);
+}
+
+static K_WORK_DEFINE(send_pending_tx, send_pending_tx_work_cb);
+
+static void wired_peripheral_read_tick_cb(struct k_timer *timer) {
+    zmk_split_wired_poll_in(&chosen_rx_buf, uart, &publish_commands,
```
paulshir, line 138: Signature has changed here. Using the callback instead of the workqueue to get the message expansion.

`zmk_split_wired_poll_in(&chosen_rx_buf, uart, NULL, &process_tx_callback,`

app/src/split/wired/wired.c
```diff
@@ -0,0 +1,322 @@
...
+            all_read = true;
+            break;
+        }
+
+        read++;
+    }
+
+    ring_buf_put_finish(rx_buf, read);
+
+    if (ring_buf_size_get(rx_buf) >= envelope_size) {
```
paulshir, line 55: Envelope size no longer applies here since the update to compress the messages. Should it be`> 0`? Otherwise it takes about 3 messages before any are actually processed.

paulshir (commented):
Leaving comments here as opposed to discord as it's easier to point to the issues. Tested and debugged the latest changes in polling mode with a rpi pico. Events and commands working with the changes I commented on here.

app/src/split/wired/wired.c
petejohanson: Yeah, thanks for all the above catches, I'd missed re-testing the polling code after several refactors. About to push some fixes.

paulshir: Thanks, latest fixes work.
Just a small note, you've set this to `>= 0` as opposed to `> 0`.

petejohanson: > Thanks, latest fixes work. Just a small note, you've set this to `>= 0` as opposed to `> 0`.

Indeed, good catch. Fixed.

app/include/zmk/split/transport/types.h
petejohanson: Yes, I am probably going to do a "custom payload" enhancement eventually, to allow this. We'll probably reserve the lower IDs for future "core" commands/events, and have the upper half of the IDs used for extension. Not for this PR though, as you note.

app/include/zmk/split/transport/types.h
```diff
@@ -0,0 +1,78 @@
...
+    ZMK_SPLIT_TRANSPORT_PERIPHERAL_EVENT_TYPE_BATTERY_EVENT,
+};
+
+struct zmk_split_transport_peripheral_event {
+    enum zmk_split_transport_peripheral_event_type type;
+
+    union {
+        struct {
+            uint8_t position;
+            uint8_t pressed;
```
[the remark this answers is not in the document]
petejohanson: I actually noticed this oddity after the fact as well, but wanted to minimize any *potential* risk to the behavior by also changing this when so much was in flight. Likely this should be unified, and just use the local/central timestamp, but I didn't want to risk any more in this PR.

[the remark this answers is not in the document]
Nick-Munnich: Personally I'd say it should use the timestamp when the event occurs at the source. I think I'd encourage you to use the same approach between sensor event and key position event either way though; having one transfer the timestamp while the other doesn't feels like it's asking for trouble to appear *somewhere*.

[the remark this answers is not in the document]
petejohanson: So, the issue here, is if timestamps are used to compare the time between "key presses" somewhere on the central, and a random timestamp that has a different "zero" will *completely* mess with that logic, so there's definitely some good reason to use the central timestamp for when it gets a key press event.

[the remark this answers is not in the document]
Nick-Munnich: That is a fair argument. Perhaps a TODO comment somewhere, talking about having timestamps taken from source once we have timestamp syncing set up properly? I recall some discussion around that being a necessity for e.g. synced fancy RGB across halves, so we'll probably have that eventually.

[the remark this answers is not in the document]
petejohanson: > That is a fair argument. Perhaps a TODO comment somewhere, talking about having timestamps taken from source once we have timestamp syncing set up properly? I recall some discussion around that being a necessity for e.g. synced fancy RGB across halves, so we'll probably have that eventually.
> 
> EDIT: I do think that sensor timestamps should be central too though for now, in that case.

Reasonable. Done.

docs/docs/config/split.md
```diff
@@ -0,0 +1,90 @@
...
+
+| Config                                       | Type | Description                                                       | Default                                                       |
+| -------------------------------------------- | ---- | ----------------------------------------------------------------- | ------------------------------------------------------------- |
+| `CONFIG_ZMK_SPLIT_WIRED`                     | bool | Use wired connection to communicate between split keyboard halves | y (if no BLE split and devicetree is set appropriately)       |
+| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_ASYNC`     | bool | Async (DMA) mode                                                  | y if the driver supports it (excluding nRF52 with known bugs) |
+| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_INTERRUPT` | bool | Interrupt mode                                                    | y if the hardware supports it                                 |
+| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_POLLING`   | bool | Polling mode                                                      | y it neither other mode is supported                          |
```
thazhemadam, line 57:
```suggestion
| `CONFIG_ZMK_SPLIT_WIRED_UART_MODE_POLLING`   | bool | Polling mode                                                      | y if neither other mode is supported                          |
```

petejohanson: Thanks. Fixed.

FearlessSpiff:
This broke my unibody board. Had to fall back to 5ba7e260f49d7277ade882410e7e97fda6e815d4 for it to work again. 
Unfortunatley I can't share my repo as I have other things there I need to fix first. Maybe it's just me?!

nmunnich:
> This broke my unibody board. Had to fall back to [5ba7e26](https://github.com/zmkfirmware/zmk/commit/5ba7e260f49d7277ade882410e7e97fda6e815d4) for it to work again. Unfortunatley I can't share my repo as I have other things there I need to fix first. Maybe it's just me?!
> 
> It was not even working over USB. Got some errors in `dmesg` when connecting it.

Could you provide some details such as a repo link so that we can investigate?

FearlessSpiff:
I would love to but I pushed some "private" stuff which I first have to get rid of somehow.

petejohanson:
> This broke my unibody board. Had to fall back to [5ba7e26](https://github.com/zmkfirmware/zmk/commit/5ba7e260f49d7277ade882410e7e97fda6e815d4) for it to work again. Unfortunatley I can't share my repo as I have other things there I need to fix first. Maybe it's just me?!
> 
> It was not even working over USB. Got some errors in `dmesg` when connecting it.

Can you share any high level details? What controller, etc?

FearlessSpiff:
Ok. Fixed it. I am using a Nicenano V2 and a cirque trackpad. But I removed all it's config when I looked for the problem. So here's the repo which works when using a commit before this one: https://github.com/FearlessSpiff/raveniung-zmk-config

Also what gets built with the actions is not working. So that's why I used a build locally.

And here is what dmesg told me:
```
 8511.628105] usb 7-1: device descriptor read/64, error -110
[ 8511.844931] usb 7-1: new full-speed USB device number 8 using xhci_hcd
```

petejohanson:
> Ok. Fixed it. I am using a Nicenano V2 and a cirque trackpad. But I removed all it's (trackpad) config when I looked for the problem. So here's the repo which works when using a commit before this one: https://github.com/FearlessSpiff/raveniung-zmk-config

Just to be sure I understand here... Did this work on latest `main` if you remove the cirque bits?

> 
> Also what gets built with the actions is not working. So that's why I used a build locally.
> 
> And here is what dmesg told me:
> ```
>  8511.628105] usb 7-1: device descriptor read/64, error -110
> [ 8511.844931] usb 7-1: new full-speed USB device number 8 using xhci_hcd
> ```


I assume it's also not advertising/connecting over BT?

FearlessSpiff:
> > Ok. Fixed it. I am using a Nicenano V2 and a cirque trackpad. But I removed all it's (trackpad) config when I looked for the problem. So here's the repo which works when using a commit before this one: https://github.com/FearlessSpiff/raveniung-zmk-config
> 
> Just to be sure I understand here... Did this work on latest `main` if you remove the cirque bits?
No, it didn't. Sorry to be unclear. I also needed to go back from this change to make it work.
> 
> > Also what gets built with the actions is not working. So that's why I used a build locally.
> > And here is what dmesg told me:
> > ```
> >  8511.628105] usb 7-1: device descriptor read/64, error -110
> > [ 8511.844931] usb 7-1: new full-speed USB device number 8 using xhci_hcd
> > ```
> 
> I assume it's also not advertising/connecting over BT?
Correct. Looked for it but there is no BT advertisement.

petejohanson:
@FearlessSpiff Reproduced here, and I just opened https://github.com/zmkfirmware/zmk/pull/2884 to fix this. Will merge as soon as the tests are all green. Sorry for the disruption.

FearlessSpiff:
Saw that #2884 got merged and can confirm it is working again for me. Thanks.

PR #2767, opened 2025-01-14
docs: Update intro page

caksoylar:
Attempt to update the intro page to docs, converting the comparison table to a feature/capabilities list. Then, split it into categories and add new items for in-progress work.

The last two are split into different commits, if you want to see intermediate states (but I'll squash them all into one before merging).

### DIFF
--- a/docs/docs/intro.md
+++ b/docs/docs/intro.md
@@ -11,35 +11,41 @@ ZMK's goal is to provide a modern and powerful firmware that is wireless-first a
 
 Below table lists major features/capabilities currently supported in ZMK, as well as ones that are currently under development and not planned.
 
-| Legend: | ✅ Supported | 🚧 Under Development | 💡 Planned | ❌ Not Planned |
-| :------ | :----------- | :------------------- | :--------- | -------------- |
+| Legend: | ✅ Supported | 🚧 Under Development | ❌ Not Planned |
+| :------ | :----------- | :------------------- | -------------- |
 
-| Feature                                                                                                                                                | Support |
+| Hardware                                                                                        | Support |
+| ----------------------------------------------------------------------------------------------- | :-----: |
+| [Split Keyboards](features/split-keyboards.md)                                                  |   ✅    |
+| [Low Active Power Usage](/power-profiler)                                                       |   ✅    |
+| [Encoders](features/encoders.md)                                                                |   ✅    |
+| [LED-based Lighting](features/lighting.md)                                                      |   ✅    |
+| [Displays](features/displays.md)                                                                |   🚧    |
+| [Pointing Devices](features/pointing.md)                                                        |   ✅    |
+| [Low Power Sleep States](features/low-power-states.md)                                          |   ✅    |
+| [Low Power Mode (VCC Shutoff)](keymaps/behaviors/power.md)                                      |   ✅    |
+| [Battery Level Reporting](features/battery.md)                                                  |   ✅    |
+| [Support for a Wide Range of ARM Chips](https://docs.zephyrproject.org/3.5.0/boards/index.html) |   ✅    |
+| Support for AVR/8-Bit Chips                                                                     |   ❌    |
+
+| Connectivity                                                                 | Support |
+| ---------------------------------------------------------------------------- | :-----: |
+| Low-Latency BLE Support                                                      |   ✅    |
+| [Multi-Device Simultaneous BLE Connectivity](features/bluetooth.md#profiles) |   ✅    |
+| [USB Connectivity](keymaps/behaviors/outputs.md)                             |   ✅    |
+
+| Keymap Features                                                                                                                                        | Support |
 | ------------------------------------------------------------------------------------------------------------------------------------------------------ | :-----: |
-| Low-Latency BLE Support                                                                                                                                |   ✅    |
-| [Multi-Device BLE Support](features/bluetooth.md#profiles)                                                                                             |   ✅    |
-| [USB Connectivity](keymaps/behaviors/outputs.md)                                                                                                       |   ✅    |
 | [User Configuration Repositories](user-setup.mdx)                                                                                                      |   ✅    |
 | [Keymaps and Layers](keymaps/index.mdx)                                                                                                                |   ✅    |
-| [Split Keyboard Support](features/split-keyboards.md)                                                                                                  |   ✅    |
+| [Wide Range of Keycodes Including Media and Consumer Codes](keymaps/list-of-keycodes.mdx)                                                              |   ✅    |
 | [Hold-Taps](keymaps/behaviors/hold-tap.mdx) (including [Mod-Tap](keymaps/behaviors/mod-tap.md) and [Layer-Tap](keymaps/behaviors/layers.md#layer-tap)) |   ✅    |
 | [Tap-Dances](keymaps/behaviors/tap-dance.mdx)                                                                                                          |   ✅    |
-| [Wide Range of Keycodes Including Media and Consumer Codes](keymaps/list-of-keycodes.mdx)                                                              |   ✅    |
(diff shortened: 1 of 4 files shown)


docs/docs/intro.md
```diff
@@ -4,49 +4,53 @@ sidebar_label: Introduction
...
+| --------------------------------------------------------------- | :-----: |
+| Low-Latency BLE Support                                         |   ✅    |
+| [Multi-Device BLE Connectivity](features/bluetooth.md#profiles) |   ✅    |
+| [USB Connectivity](keymaps/behaviors/outputs.md)                |   ✅    |
+
+| Keymap Features                                                                                                                                        | Support |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------ | :-----: |
+| [User Configuration Repositories](user-setup.mdx)                                                                                                      |   ✅    |
+| [Keymaps and Layers](keymaps/index.mdx)                                                                                                                |   ✅    |
+| [Wide Range of Keycodes (Including Media and Consumer)](keymaps/list-of-keycodes.mdx)                                                                  |   ✅    |
```
Nick-Munnich, line 44: I wonder if it might be good to mention that these are HID keycodes here. Might help clarify the locales situation for people.

docs/docs/intro.md
```diff
@@ -4,49 +4,53 @@ sidebar_label: Introduction
...
+| [Support for a Wide Range of ARM Chips](https://docs.zephyrproject.org/3.5.0/boards/index.html) |   ✅    |
+| Support for AVR/8-Bit Chips                                                                     |   ❌    |
+
+| Connectivity                                                    | Support |
+| --------------------------------------------------------------- | :-----: |
+| Low-Latency BLE Support                                         |   ✅    |
+| [Multi-Device BLE Connectivity](features/bluetooth.md#profiles) |   ✅    |
+| [USB Connectivity](keymaps/behaviors/outputs.md)                |   ✅    |
+
+| Keymap Features                                                                                                                                        | Support |
```
Nick-Munnich, line 40: I wonder if there's some way to describe how our behavior system is flexible and allows things like hold-taps inside of mod-morph. I also wonder what justifies a particular behavior/feature being in this table vs not being in it.

Nick-Munnich (commented):
Agree that this has been overdue. Aside from my minor comments, my biggest complaint is how the tables appear on the page. If it's possible to center the legend table and move the keymap and connectivity tables to be on the right of the hardware table when there is the space for it, that'd improve the situation massively for me. Other alignment stuff too.

docs/docs/intro.md
```diff
@@ -4,49 +4,53 @@ sidebar_label: Introduction
 slug: /
 ---
 
-ZMK Firmware is an open source (MIT) keyboard
-firmware built on the [Zephyr™ Project](https://zephyrproject.org/) Real Time Operating System (RTOS). ZMK's goal is to provide a modern, wireless, and powerful firmware free of licensing issues.
+ZMK Firmware is an open source (MIT) keyboard firmware built on the [Zephyr™ Project](https://zephyrproject.org/) Real Time Operating System (RTOS).
+ZMK's goal is to provide a modern and powerful firmware that is wireless-first and free of licensing issues.
```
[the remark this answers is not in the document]
caksoylar: I am fine either way, I think wireless-first is fair given the history. I'll defer to @petejohanson on the final wording.

It'd be nice to use less exclusive language but calling it something like "HID firmware" is too inaccessible IMO. I'll vote to kick that can down the line, again :P

docs/docs/intro.md
caksoylar: These are informed by the old comparison, even though we don't have the comparison columns anymore. I thought of adding e.g. mod-morph but that isn't likely to be a feature people would know about before coming to ZMK, whereas others are.

This is the reason I added the "major" qualifier for features/capabilities at the top of the table. I don't think we should document everything, so we have to add a cut-off somewhere.

caksoylar:
> Aside from my minor comments, my biggest complaint is how the tables appear on the page. 

I agree with this, it doesn't look good. I don't know if we can easily put tables side-by-side. My initial thought was making the tables full width like https://zmk.dev/docs/keymaps/list-of-keycodes, which seems to require using a custom element rather than a markdown table, to add a bit of CSS.

docs/docs/intro.md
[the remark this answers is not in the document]
Nick-Munnich: Agree on the terminology of HID input device being too inaccessible. I meant more philosophically, and then using nicer phrasing in the docs. That's still something for the future, I don't think we're there yet, but something to think about.

docs/docs/intro.md
Nick-Munnich: I think most people wouldn't have any clue about "consumer". They could guess "media" *maybe*, but I'd say include HID or drop the brackets

docs/docs/intro.md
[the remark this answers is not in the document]
petejohanson: If we want to nix the "wireless-first" qualifier, then I think something equivalent, maybe.

> ZMK's goal is to provide a modern and powerful firmware that is designed for power-efficiency, flexibility, and broad hardware support.

?

docs/docs/intro.md
```diff
@@ -4,49 +4,53 @@ sidebar_label: Introduction
...
+| [Encoders](features/encoders.md)                                                                |   ✅    |
+| [LED-based Lighting](features/lighting.md)                                                      |   ✅    |
+| [Displays](features/displays.md)                                                                |   🚧    |
+| [Pointing Devices](features/pointing.md)                                                        |   ✅    |
+| Multitouch Touchpads (PTP)                                                                      |   🚧    |
+| [Low Power Sleep States](features/low-power-states.md)                                          |   ✅    |
+| [Low Power Mode (VCC Shutoff) for Peripherals](keymaps/behaviors/power.md)                      |   ✅    |
+| Improved Power Handling for Multiple Peripherals                                                |   🚧    |
+| [Battery Level Reporting](features/battery.md)                                                  |   ✅    |
+| [Support for a Wide Range of ARM Chips](https://docs.zephyrproject.org/3.5.0/boards/index.html) |   ✅    |
```
petejohanson, line 31: Given we're starting to have some RISC-V support, Xtensa, etc. perhaps take this time to adjust?

```suggestion
| [Support for a Wide Range of 32-bit Microcontrollers](https://docs.zephyrproject.org/3.5.0/boards/index.html) |   ✅    |
```

docs/docs/intro.md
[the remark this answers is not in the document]
Nick-Munnich: I don't think we want to drop "wireless" from the description, I just think we should be placing them on equal level to each other. Not including it makes it seem like it's *not* capable of wireless imo, for someone who knows nothing about the area.

docs/docs/intro.md
caksoylar: Fair, dropping the part in parentheses.

docs/docs/intro.md
caksoylar: I added a new item "flexible behavior system" that is a bit vague but maybe still worth noting?

docs/docs/intro.md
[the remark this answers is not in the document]
caksoylar: I split (?) the difference and added both sentences, let me know what you think.

caksoylar:
I gave the two column suggestion @Nick-Munnich had a shot, using the implementation from https://docusaurus.community/knowledge/component-library/new/Columns/. It still isn't perfect but seems better, I am open to other ideas.

Nick-Munnich (approved):
I think it looks good, I'm happy.

petejohanson (approved):
LGTM! Thanks!

PR #2770, opened 2025-01-15
feat(core): Make physical layout key rotation optional

petejohanson:
To be able to save on flash space, for layouts on space constrained devices that don't require rotation, make key rotation props optional behind a new Kconfig flag.

## PR check-list

### DIFF
--- a/app/src/studio/keymap_subsystem.c
+++ b/app/src/studio/keymap_subsystem.c
@@ -278,9 +278,11 @@ static bool encode_layout_keys(pb_ostream_t *stream, const pb_field_t *field, vo
             .height = layout_kp->height,
             .x = layout_kp->x,
             .y = layout_kp->y,
+#if IS_ENABLED(CONFIG_ZMK_PHYSICAL_LAYOUT_KEY_ROTATION)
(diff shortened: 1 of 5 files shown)


app/src/studio/keymap_subsystem.c
```diff
@@ -278,9 +278,11 @@ static bool encode_layout_keys(pb_ostream_t *stream, const pb_field_t *field, vo
             .height = layout_kp->height,
             .x = layout_kp->x,
             .y = layout_kp->y,
+#if IS_ENABLED(CONFIG_ZMK_PHYSICAL_LAYOUT_KEY_ROTATION)
             .r = layout_kp->r,
             .rx = layout_kp->rx,
             .ry = layout_kp->ry,
+#endif
```
caksoylar, line 285: I guess Studio client handles not getting these properties appropriately?

petejohanson: Yeah, if they're not included in the PB payload, clients will treat them as having their default values, which for integer fields is `0` which is what we'd want.

PR #2784, opened 2025-01-21
fix(ble): enforce maximum length for dynamic device name

xudongzheng:


app/Kconfig
```diff
@@ -238,6 +238,11 @@ config BT_PERIPHERAL_PREF_LATENCY
 config BT_PERIPHERAL_PREF_TIMEOUT
     default 400
 
+# The device name should be 16 characters or less so it fits within the
+# advertising data.
```
joelspadin, line 242: Does this need to be configurable? The comment implies that the value should always be 16.

petejohanson (approved):
Thanks, good catch!

PR #2787, opened 2025-01-21
test(split): remove build banner from snapshot

xudongzheng:
The test snapshot will otherwise be outdated each time there is a new commit on https://github.com/zmkfirmware/zephyr.


petejohanson (approved):
Good catch. Thanks!

Issue #2792, opened 2025-01-24
Segfault of `as_zmk_battery_state_changed` on `native_posix_64`


mlouielu:
At the following config, when building on `native_posix_64`, it will segfault during init process:

```
CONFIG_ZMK_DISPLAY=y
CONFIG_ZMK_BLE=y
CONFIG_ZMK_BATTERY_REPORTING=y
```

GDB backtrace:

```gdb
[00:00:00.000,000] <dbg> zmk: initialize_display: 

Thread 6 "zmk.elf" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff73a66c0 (LWP 1280968)]
0x0000000000414315 in as_zmk_battery_state_changed (eh=eh@entry=0x0)
    at /zmk/app/src/events/battery_state_changed.c:10
10	ZMK_EVENT_IMPL(zmk_battery_state_changed);
(gdb) bt
#0  0x0000000000414315 in as_zmk_battery_state_changed (eh=eh@entry=0x0)
    at /zmk/app/src/events/battery_state_changed.c:10
#1  0x0000000000414d3c in battery_status_get_state (eh=eh@entry=0x0)
    at /zmk/app/src/display/widgets/battery_status.c:66
#2  0x0000000000414d77 in widget_battery_status_refresh_state (eh=eh@entry=0x0)
    at /zmk/app/src/display/widgets/battery_status.c:76
#3  0x0000000000414e23 in widget_battery_status_init ()
    at /zmk/app/src/display/widgets/battery_status.c:76
#4  0x0000000000414e69 in zmk_widget_battery_status_init (
    widget=widget@entry=0x4d01c0 <battery_status_widget>, 
    parent=parent@entry=0x4d0db8 <lvgl_heap_mem+2000>)
    at /zmk/app/src/display/widgets/battery_status.c:89
#5  0x0000000000414b7a in zmk_display_status_screen ()
    at /zmk/app/src/display/status_screen.c:42
#6  0x0000000000414a57 in initialize_display (work=work@entry=0x456500 <init_work>)
    at /zmk/app/src/display/main.c:116
#7  0x0000000000445a44 in work_queue_main (workq_ptr=0x4d00a0 <k_sys_work_q>, 
    workq_ptr@entry=<error reading variable: value has been optimized out>, 
    p2=<error reading variable: value has been optimized out>, 
    p3=<error reading variable: value has been optimized out>)
    at /zmk/zephyr/kernel/work.c:672
[...]

PR #2793, opened 2025-01-24
fix(events): Add NULL check for `as_##event_type`.

mlouielu:
See #2792 where this cause segfault.

## PR check-list

- [0] Branch has a [clean commit history](https://zmk.dev/docs/development/contributing/pull-requests#clean-commit-history)
- [-]Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)

### DIFF
--- a/app/include/zmk/event_manager.h
+++ b/app/include/zmk/event_manager.h
@@ -57,6 +57,9 @@ struct zmk_event_subscription {
         return ZMK_EVENT_RAISE(ev);                                                                \
     };                                                                                             \
     struct event_type *as_##event_type(const zmk_event_t *eh) {                                    \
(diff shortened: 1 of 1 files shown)


Nick-Munnich:
I'd like to see a logging message added to this.

joelspadin:
See also my comments on #2792. This is being used as a workaround to what I consider to be a design flaw in the display code. I wouldn't be opposed to adding this with an error log message, but then we'd log a bunch of errors every time the firmware starts up when using a display.

Nick-Munnich:
> This is being used as a workaround to what I consider to be a design flaw in the display code.

Agree with this point. I'd like to see both fixes made simultaneously.

PR #2795, opened 2025-01-24
fix(usb): Fix `-Wincompatible-pointer-types` compiler error

### LINKED ISSUE #2794: Having `-Wincompatible-pointer-types` when building with `CONFIG_ZMK_USB=y`
When enabled `CONFIG_ZMK_USB=y`, the following error `-Wincompatible-pointer-types` shows:

```zsh
/zmk/app/src/usb_hid.c: In function ‘get_report_cb’:
/zmk/app/src/usb_hid.c:92:41: error: passing argument 1 of ‘get_keyboard_report’ from incompatible pointer type [-Wincompatible-pointer-types]
   92 |             *data = get_keyboard_report(len);
      |                                         ^~~
      |                                         |
      |                                         int32_t * {aka int *}
compilation terminated due to -Wfatal-errors.

```

Example

```zsh
$ echo "CONFIG_ZMK_USB=y" > tests/none/normal/native_posix_64.conf
$ west build --pristine --board native_posix_64 -- -DZMK_CONFIG=tests/none/normal/
...
/zmk/app/src/usb_hid.c:92:41: error: passing argument 1 of ‘get_keyboard_report’ from incompatible pointer type [-Wincompatible-pointer-types]
...
```

mlouielu:
Fix #2794, where `get_keyboard_report(len)` didn't convert `len` from `int32_t *` to `size_t *`.

## PR check-list

### DIFF
--- a/app/src/usb_hid.c
+++ b/app/src/usb_hid.c
@@ -89,7 +89,7 @@ static int get_report_cb(const struct device *dev, struct usb_setup_packet *setu
     case HID_REPORT_TYPE_INPUT:
         switch (setup->wValue & HID_GET_REPORT_ID_MASK) {
         case ZMK_HID_REPORT_ID_KEYBOARD: {
-            *data = get_keyboard_report(len);
+            *data = get_keyboard_report((size_t *)len);
             break;
         }
         case ZMK_HID_REPORT_ID_CONSUMER: {


app/src/usb_hid.c
```diff
@@ -89,7 +89,7 @@ static int get_report_cb(const struct device *dev, struct usb_setup_packet *setu
     case HID_REPORT_TYPE_INPUT:
         switch (setup->wValue & HID_GET_REPORT_ID_MASK) {
         case ZMK_HID_REPORT_ID_KEYBOARD: {
-            *data = get_keyboard_report(len);
+            *data = get_keyboard_report((size_t *)len);
```
joelspadin, line 92: This doesn't actually fix the issue that the compiler is warning you about. It just says, "I know what I'm doing", so the compiler stops warning about it.

`len` is still a pointer to `int32_t`. Casting it to a pointer to `size_t` means that `get_keyboard_report()` is going to write a `size_t` to the memory that is actually an `int32_t`. On a platform such as `native_posix_64` where `size_t` is `uint64_t`, it will write a 64-bit value to the given address, but the variable is only 32 bits, so it clobbers some memory after it.

Some safe alternatives:
- Change `get_keyboard_report()` to also use an `int32_t` for its output parameter (requires also updating any other places that call the function to use `int32_t` as well).
- Create a new `size_t` variable here, pass its address to `get_keyboard_report()`, then assign its value to `len`.

joelspadin (changes requested):

app/src/usb_hid.c
petejohanson: Note: I took @joelspadin second approach when I hit this and created #2812

petejohanson:
Thanks for the PR! Closing, since we addressed this a slightly different way.

PR #2806, opened 2025-02-01
feat(boards): Add waveshare rp2040 board definition

Nick-Munnich:
This PR builds off of #2596, fixing the problems Pete noted there alongside some other changes. As of two weeks ago, Zephyr's main branch has this board as well, so the name was adjusted to match with it. 

Figured since I needed to update the PR for my own stuff anyway I may as well PR the updated version.

## PR check-list

### DIFF
--- a/app/boards/arm/waveshare_rp2040_zero/Kconfig.board
+++ b/app/boards/arm/waveshare_rp2040_zero/Kconfig.board
@@ -1,0 +1,6 @@
+# Copyright (c) 2024 The ZMK Contributors
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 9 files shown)


xudongzheng:
The pins defined for I2C and SPI are a bit different from upstream Zephyr. If the plan is for ZMK to eventually use the upstream board, that could be confusing.

nmunnich:
https://github.com/nmunnich/zmk-component-rp2040-zero This component is made available in a module. As we are looking into upgrading to Zephyr 4.1 currently, which includes this component, I see no reason to further pursue putting this into ZMK directly.

PR #2456, opened 2024-09-03
feat(docs): Added a page on module creation

Nick-Munnich:
Added a page on module creation, much nicer this time.

This page makes use of another repository, `zmk-module-template`, which we should fork into the project if this approach is approved.

~~This page also makes reference to a second repository, `zmk-modules`. Work regarding `zmk-modules` is still in progress, but it is there to read and provide feedback on. If this is merged before `zmk-modules` is ready, then I suggest we comment out the parts on `zmk-modules` for the time being.~~

PR #2811, opened 2025-02-03
chore: Add a deprecation warning to config/boards

Nick-Munnich:
Thought it was about time for at least a warning to be spat out. My suggestion would be that after some more time we make using `config/boards` spit out an error that can be lifted with an extra cmake argument, but I think we should have a warning message first.

## PR check-list


caksoylar:
I think it'd be better to be more explicit here, someone who doesn't follow ZMK might not even know what a "module" is, given this warning will be emitted by GH Actions where config repo == module. Is there a docs page we can link to?

Nick-Munnich:
With #2456 merged I have now updated this with a link to the appropriate docs page.

caksoylar:
I think the new shield guide might be more straight-forward to link to, given the module creation page might be a bit overwhelming for people who just want to use a custom shield. Maybe we can add both, with new shield first, what do you think?

Nick-Munnich:
Good idea, made that change.

caksoylar (approved):

petejohanson (approved):
Thanks!

PR #2814, opened 2025-02-07
feat(display): Add ability to set display on/off pin.

petejohanson:
Zephyr is still working on the plan upstream for generically controlling display "backlight" pins with GPIO/PWM, so in the meantime, add our own chosen property `zmk,display-led` that is set to an LED device child to allow blanking/unblanking of devices that use a dedicated backlight control pin.

Basically looks like this in the devicetree:

```
/ {
    chosen {
        zephyr,display = &vik_st7789v;
        zmk,display-led = &disp_led;
    };

    leds {
        compatible = "gpio-leds";

        disp_led: disp_led {
            gpios = <&vik_conn 3 GPIO_ACTIVE_HIGH>;
            label = "Display LED";
        };
    };
};
```

And then you need to enable the `CONFIG_LED=y` Kconfig flag to build that driver subsystem in.

## PR check-list

### DIFF
--- a/docs/docs/config/displays.md
+++ b/docs/docs/config/displays.md
@@ -62,3 +62,12 @@ See the Devicetree bindings for your display. Here are the bindings for common d
 - [SSD1306 (spi)](https://docs.zephyrproject.org/3.5.0/build/dts/api/bindings/display/solomon,ssd1306fb-spi.html)
 
 A full list of drivers provided by Zephyr can be found in [Zephyr's Devicetree bindings index](https://docs.zephyrproject.org/3.5.0/build/dts/api/bindings.html).
+
+### Chosen nodes
+
+Applies to: [`/chosen` node](https://docs.zephyrproject.org/3.5.0/build/dts/intro-syntax-structure.html#aliases-and-chosen-nodes)
+
+| Property          | Type | Description                                                                                              |
+| ----------------- | ---- | -------------------------------------------------------------------------------------------------------- |
+| `zephyr,display`  | path | The display device to use.                                                                               |
+| `zmk,display-led` | path | The LED device to use for on/off blanking, if the hardware requires it. Can be a PWM or GPIO LED device. |
--- a/app/src/display/main.c
+++ b/app/src/display/main.c
@@ -13,6 +13,7 @@
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #include <zephyr/drivers/display.h>
+#include <zephyr/drivers/led.h>
 #include <lvgl.h>
 
 #include "theme.h"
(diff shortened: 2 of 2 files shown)


docs/docs/config/displays.md
```diff
@@ -62,3 +62,12 @@ See the Devicetree bindings for your display. Here are the bindings for common d
 - [SSD1306 (spi)](https://docs.zephyrproject.org/3.5.0/build/dts/api/bindings/display/solomon,ssd1306fb-spi.html)
 
 A full list of drivers provided by Zephyr can be found in [Zephyr's Devicetree bindings index](https://docs.zephyrproject.org/3.5.0/build/dts/api/bindings.html).
+
```
caksoylar, line 65: FYI this is how we normally document these in config pages: https://zmk.dev/docs/config/kscan#devicetree

Nick-Munnich:
Would it not make more sense for `&disp_led` to be attached to the `&vik_st7789v` node? Or does Zephyr not support multiple displays from one device anyway, so it doesn't matter.

petejohanson:
> Would it not make more sense for `&disp_led` to be attached to the `&vik_st7789v` node? Or does Zephyr not support multiple displays from one device anyway, so it doesn't matter.

It does not support more than one. Upstream Zephyr folks are exploring how to more fully integrate it, but there's various pros/cons to a more invasive change, so this is the pragmatic way to support this today, with existing drivers, that won't cause conflicts once upstream settles on an approach.

docs/docs/config/displays.md
petejohanson: Thanks. Missed that. I'll update.

petejohanson: @caksoylar Tweaked.

docs/docs/config/displays.md
```diff
@@ -62,3 +62,12 @@ See the Devicetree bindings for your display. Here are the bindings for common d
...
+### Chosen nodes
+
+Applies to: [`/chosen` node](https://docs.zephyrproject.org/3.5.0/build/dts/intro-syntax-structure.html#aliases-and-chosen-nodes)
+
+| Property          | Type | Description                                                                                              |
+| ----------------- | ---- | -------------------------------------------------------------------------------------------------------- |
+| `zephyr,display`  | path | The display device to use.                                                                               |
+| `zmk,display-led` | path | The LED device to use for on/off blanking, if the hardware requires it. Can be a PWM or GPIO LED device. |
```
Nick-Munnich, line 73:
```suggestion
| `zephyr,display`  | phandle | The display device to use.                                                                               |
| `zmk,display-led` | phandle | The LED device to use for on/off blanking, if the hardware requires it. Can be a PWM or GPIO LED device. |
```
Phandle as type is used in e.g. https://zmk.dev/docs/config/kscan#devicetree-5, not sure if path appears anywhere.

petejohanson: I'm happy to do this, but we should update the kscan docs too to match as well.

Nick-Munnich: Mmm there's more appearances than I thought. Maybe leave it as-is and we'll do another PR at some point making it all consistent.

caksoylar: I think this is a path and not phandle indeed, it is not in angle brackets and can use an absolute path as well: https://zmk.dev/docs/config#path

PR #2815, opened 2025-02-08
docs: Add a dedicated page on ZMK events

Nick-Munnich:
This PR adds a dedicated page on events in ZMK and the ZMK event handler. Not sure how events interact with modules, needs that figuring out before this can be merged.

Mostly wrote this for my own reference since I kept forgetting, I don't plan on checking the interaction with modules myself anytime soon. If someone knows the interaction details, please share.

## PR check-list


nmunnich:
I have now checked the interaction of ZMK events with modules, and documented this accordingly.

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
+---
+title: ZMK Events
+sidebar_label: ZMK Events
+---
+
+ZMK makes use of events to decouple individual components such as behaviors and peripherals from the core functionality. For this purpose, ZMK has implemented its own event manager. This page is a (brief) overview of the functionality and methods exposed by the event manager. To see what events exist and what data they contain, it is best to view the corresponding [event header files](https://github.com/zmkfirmware/zmk/tree/main/app/include/zmk/events) directly.
```
caksoylar: A small intro sentence here would be good, regarding who this is targeted for. I'd like to clarify this is documenting the C API and is aimed towards users writing new behaviors (link to guide) or other features (link to modules creation page?).

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+
+Including `zmk/event_manager.h` is required for any interaction with the event system.
```
caksoylar:
```suggestion
Including the event_manager header via `#include <zmk/event_manager.h>` is required for any interaction with the event system.
```

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Generic Events
```
caksoylar, line 10: I keep going back and forth on whether this section is necessary for users, since it goes into `ZMK_EVENT_DECLARE`/`ZMK_EVENT_IMPL` implementation details and is confusing to me. I guess it is necessary for wherever we mention `zmk_event_t`, "header", "data" etc. below.

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+ZMK_SUBSCRIPTION(combo, zmk_keycode_state_changed);
+```
+
+The first parameter is the name of the listener created with `ZMK_LISTENER`, while the second is the name of the _struct_ that defines the event's data, which was declared in the corresponding header file. By convention the header file for an event will be named `specific_thing_happened`, with the struct named `zmk_specific_thing_happened`.
+
+Of course, you will also need to import the corresponding event header at the top of your file.
+
+### Listener Callback
+
+The listener will be passed a raised `zmk_event_t` pointer (as described previously) as an argument, and should have `int` as its return type. By convention, the `zmk_event_t` is called
```
caksoylar: Missing the final part of the sentence?

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+
+The listener should return one of three values (which are of type `int`) back to the event manager:
+
+- `ZMK_EV_EVENT_BUBBLE`: Keep propagating the event `struct` to the next listener.
+- `ZMK_EV_EVENT_HANDLED`: Stop propagating the event `struct` to the next listener. The event manager still owns the `struct`'s memory, so it will be `free`d automatically. Do **not** free the memory in this function.
+- `ZMK_EV_EVENT_CAPTURED`: Stop propagating the event `struct` to the next listener. The event `struct`'s memory is now owned by your code, so the event manager will not free the event `struct` memory. Make sure your code will release or free the event at some point in the future. (Use the `ZMK_EVENT_*` macros described [below](#raising-events).)
+
+If an error occurs during the listener call, it should of course return the error.
```
caksoylar:
```suggestion
If an error occurs during the listener call, it should return a negative value indicating the appropriate error code.
```

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+
+As mentioned previously, the same callback will be called when any event that is subscribed to occurs. To obtain the underlying event from the generic event passed to the listener, the previously described `as_zmk_specific_thing_happened` function should be used:
+
+```c
+int behavior_hold_tap_listener(const zmk_event_t *eh) {
+    if (as_zmk_position_state_changed(eh) != NULL) {
+        return position_state_changed_listener(eh);
+    } else if (as_zmk_keycode_state_changed(eh) != NULL) {
+        return keycode_state_changed_listener(eh);
```
caksoylar:
```suggestion
    if (as_zmk_position_state_changed(eh) != NULL) {
        // it is a position_state_changed event, handle it with my_position_state_handler
        return my_position_state_handler(eh);
    } else if (as_zmk_keycode_state_changed(eh) != NULL) {
        // it is a keycode_state_changed event, handle it with my_keycode_state_handler
        return my_keycode_state_handler(eh);
```

This snippet didn't make it clear to me that `*_listener` functions are custom and not provided by ZMK.

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+    }
+    return ZMK_EV_EVENT_BUBBLE;
+}
+```
+
+The priority of the listeners is determined by the order in which the linker links the files. Within ZMK, this is the order of the corresponding files in `CMakeLists.txt`. External modules targeting `app` are linked prior to any files within ZMK itself, making them the highest priority. It is thus the module maintainer's responsibility to both ensure that their module does not cause issues by being first in the listener queue. For example, [hold-tap](../keymaps/behaviors/hold-tap.mdx) is the first listener to `position_state_changed`, and may behave inconsistently if a behavior defined in a module listens to `position_state_changed` and calls `hold_tap`.
```
caksoylar: "calls `hold_tap`" at the end is a bit ambiguous, what do you have in mind?

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+
+When considering multiple modules, priority is determined by the order in which the modules are present in the user's `west.yml`. Hence there should be no order dependencies between modules, only within a module.
+
+## Raising Events
+
+There are several different ways to raise events, with slight differences between them.
+
+- `int raise_zmk_specific_thing_happened(struct zmk_specific_thing_happened event)`: This function will take an event data structure, add a header to it, and then start handling the event with the first registered event listener.
+
+The following macros can also be used. These will each take in an event `ev` which already consists of the header & data combination, i.e. `ev` has the type `struct zmk_specific_thing_happened_event`.
```
caksoylar:
```suggestion
The following macros can also be used for advanced use cases. These will each take in an event `ev` which already consists of the header & data combination, i.e. `ev` has the type `struct zmk_specific_thing_happened_event`.
```

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+```
+
+## Creating New Events
+
+### Header File
+
+Your event's header file should have four things:
+
+- A copyright comment
+- Any required header files (along with `#pragma once`)
```
caksoylar:
```suggestion
- Any required header includes (along with `#pragma once`)
```

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+
+#pragma once
+
+#include <zephyr/kernel.h>
+
+#include <zmk/endpoints_types.h>
+#include <zmk/event_manager.h>
+
+struct zmk_endpoint_changed {
+struct zmk_endpoint_instance endpoint;
```
caksoylar:
```suggestion
    struct zmk_endpoint_instance endpoint;
```

docs/docs/development/new-behavior.mdx
```diff
@@ -215,38 +215,7 @@ Including `zmk/event_manager.h` is required for the following dependencies to fu
...
-```c title="app/src/behaviors/behavior_tap_dance.c (Lines 192-197, 225)"
-static int tap_dance_position_state_changed_listener(const zmk_event_t *eh);
-
-ZMK_LISTENER(behavior_tap_dance, tap_dance_position_state_changed_listener);
-ZMK_SUBSCRIPTION(behavior_tap_dance, zmk_position_state_changed);
-
-static int tap_dance_position_state_changed_listener(const zmk_event_t *eh){
-    // Do stuff...
-}
-```
```
caksoylar, line 233: Keeping just this example in this page makes sense to me.

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+```
+
+In memory, a raised event **always** consists of a `zmk_event_t` struct **followed immediately afterwards** by the struct containing the data for the actual event. The contents of `const struct zmk_event_type *event` allows us to identify which event a particular event actually is, so that we may safely access its data. The type of the combination of a `zmk_event_t` struct with a specific event data struct, e.g. `struct zmk_specific_thing_happened`, is `struct zmk_specific_thing_happened_event`:
+
+```c
+struct zmk_specific_thing_happened_event {
+    zmk_event_t header;
+    struct zmk_specific_thing_happened data;
+};
+```
```
caksoylar:
````suggestion
In memory, the struct for a specific raised event `struct zmk_specific_thing_happened_event` **always** consists of a `zmk_event_t` struct **followed immediately afterwards** by the struct containing the data for the actual event. 

```c
struct zmk_specific_thing_happened_event {
    zmk_event_t header;
    struct zmk_specific_thing_happened data;
};
```
````

File: docs/docs/development/events.md [hunk 13]
```diff
@@ -0,0 +1,178 @@
...
+
+To obtain the underlying data from a generic event, the following function is used:
+
+```c
+struct zmk_specific_thing_happened *as_specific_thing_happened(const zmk_event_t *eh);
+```
+
+This method takes in a pointer to a `zmk_event_t` (which is actually a pointer to a specific event, such as `zmk_specific_thing_happened_event`), and will return the underlying `zmk_specific_thing_happened` data struct if the `zmk_event_t` header indicates that the generic event pointer is indeed a pointer to a `zmk_specific_thing_happened_event`. If the type of the event does not match the function, then the function will return `NULL`. By convention, `zmk_event_t` pointer arguments are named `eh`, short for "event header".
```
caksoylar, line 36: This is perhaps outside the scope of the docs but it is confusing me as I read this so I'll ask: Why is the arg type `zmk_event_t` here? It seems like this is not correct, as it is actually a specific event type, like you mentioned here.

docs/docs/development/events.md
```diff
@@ -0,0 +1,178 @@
...
+In memory, a raised event **always** consists of a `zmk_event_t` struct **followed immediately afterwards** by the struct containing the data for the actual event. The contents of `const struct zmk_event_type *event` allows us to identify which event a particular event actually is, so that we may safely access its data. The type of the combination of a `zmk_event_t` struct with a specific event data struct, e.g. `struct zmk_specific_thing_happened`, is `struct zmk_specific_thing_happened_event`:
+
+```c
+struct zmk_specific_thing_happened_event {
+    zmk_event_t header;
+    struct zmk_specific_thing_happened data;
+};
+```
+
+To obtain the underlying data from a generic event, the following function is used:
```
caksoylar:
```suggestion
The contents of `header.event` allows us to identify which type a particular event actually is, so that we may safely access its data in `data`. This is handled by the following function, which allows us to obtain the underlying data from a generic event:
```

docs/docs/development/events.md
nmunnich: I believe this is to accept a generic "event type" without needing to know any details about the event itself. A union type would work if we didn't have to consider external modules as well.

This exact point was confusing me, hence why I explained the "what" of what's happening, without going into detail on the "why", since that's not really relevant to someone writing a module. Explaining this confusion is the main reason why I wanted to have this section.

docs/docs/development/events.md
nmunnich: Ah, I moved this sentence above - this was supposed to be "By convention, `zmk_event_t` pointer arguments are named `eh`, short for "event header".". Removed.

docs/docs/development/events.md
nmunnich: https://github.com/zmkfirmware/zmk/blob/5144de677d6474ca40695c1841b725ba3160ccba/app/src/behaviors/behavior_hold_tap.c#L611-L615 May run into this, which currently can never occur in ZMK natively as hold-tap is the first in the queue. That's too much detail to go into in this page imo, hence why I kept it vague.

docs/docs/development/events.md
nmunnich: Done.

docs/docs/development/events.md
caksoylar: Right, what's confusing to me is how the same `eh` can act as both of these at the same time within `as_specific_thing_happened`:
- `zmk_event_t`: to check `eh->event` to determine the specific type (**not** `eh->header.event` which'd be consistent with below)
- `zmk_specific_thing_happened_event`: to fetch the data using `&((struct event_type##_event *)eh)->data`

without a linter complaining about it.

Anyway this is likely not crucial for the users to understand, but this section is going to be confusing for readers. If you have any improvements in mind it'd be good, but otherwise I guess I'll let it go.

docs/docs/development/events.md
caksoylar: Got it, can you expands upon the actual calling? Do you mean invokes a hold-tap behavior through e.g. `zmk_behavior_invoke_event`?

docs/docs/development/events.md
nmunnich: I'm not sure we have linters checking that bit, it's inside of a preprocessor macro on top of everything after all. I'm not sure linters could understand what's happening exactly, but it works fine because the way everything gets laid out in memory means that `*zmk_event_t` and `*zmk_specific_thing_happened_event` are the same address if `*zmk_event_t` is exactly  `*zmk_specific_thing_happened_event->header`, and so the original `*zmk_event_t` can be safely cast even though the type it is being cast to requires more memory.

docs/docs/development/events.md
nmunnich: Done. Do you think the new version is clear enough?

caksoylar: That seems good enough, thanks.

docs/docs/development/events.md
caksoylar: I see, "are the same address" happens through `header` being the first item in the struct, I assume. Thanks for explaining!

PR #2839, opened 2025-02-25
refactor(behaviors): Minor RAM usage tweaks

petejohanson:
Follow up cleanup of some config structs not marked as const.

## PR check-list


Nick-Munnich:
Couple more points in the example code in the docs which could use a `const` added, along with a note under the corresponding section in the docs that `_config` should always be `const`.

petejohanson:
> Couple more points in the example code in the docs which could use a `const` added, along with a note under the corresponding section in the docs that `_config` should always be `const`.

Added the note, but I'm not seeing any other declared instances of the config data in the new behavior guide that would get updated.

Nick-Munnich (approved):
You're right, mb, I misread a bit. LGTM

PR #2840, opened 2025-02-25
docs: Add query strings to tabs

caksoylar:
Apparently docusaurus tabs support query strings, which lets you persist the selected tab to the URL: https://docusaurus.io/docs/markdown-features/tabs#query-string. I think this will be useful to share links to the docs where we can point to the right tab value, so I added it in most places that tabs are used -- skipped a few that didn't seem necessary.

Note: when value of `queryString` is skipped it falls back to using `groupId`, so I preferred that whenever the latter exists.

### DIFF
--- a/docs/docs/development/hardware-integration/dongle.mdx
+++ b/docs/docs/development/hardware-integration/dongle.mdx
@@ -146,6 +146,7 @@ If your keyboard is not Studio-ready or you have no interest in using ZMK Studio
 
 <Tabs
     groupId="studio-ready"
+    queryString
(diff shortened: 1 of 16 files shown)


Nick-Munnich (approved):
I like this. I haven't checked whether there were any missed tabs, but I like the look of the preview and haven't noticed any issues scanning the commit details.

caksoylar:
For reference, I think I only skipped https://zmk.dev/docs/features/modules#examples and https://zmk.dev/docs/development/new-behavior#submitting-a-pull-request.

Issue #2368, opened 2024-07-08
Hold-taps should honor transparency within the bindings.


amacleod:
When one of the bindings of a hold-tap is `&trans`, the behavior should look at lower layers' bindings when that portion of the hold-tap is activated.

```dts
behaviors {
  hmt: homerow_mod_transparent {
    compatible = "zmk,behavior-hold-tap";
    #binding-cells = <2>;
    bindings = <&kp>, <&trans>;
    // ... and any other mandatory fields
  };
};
keymap {
  base_layer { bindings = <&kp A>; };
  other_layer { bindings = <&kp B>; };
  overlay_mods_layer {
    bindings = <&hmt LSHFT 0>;
  };
};
```

Expected behavior: when the overlay mods layer is active, tapping the key should either send `a` or `b` depending on which lower layer is active.

Actual behavior: tapping the key does nothing when the overlay layer is active, just as if the second binding were `&none`. The hold behavior works fine.

---

https://github.com/amacleod/zmk-config/blob/hold-tap-trans-repro/config/corne.keymap#L381-L391 is an example of me trying to use a transparent hold-tap to make a floating home row mods layer.

amacleod:
Linking the locality-preservation issue https://github.com/zmkfirmware/zmk/issues/1494 as @caksoylar mentioned that it might be related

caksoylar:
I looked into this a bit and fixing #1494 apparently didn't fix it. `&trans` behavior doesn't do anything special besides returning a positive code, and that is processed by the keymap logic to fall through to other layers. But hold-tap doesn't propagate up the return code of some of the behaviors it invokes due to its complex decision logic. Maybe this can be fixed, but it seems difficult (to me) for not much gain.

amacleod:
If the effort:effect ratio would be too low, I am okay with closing this issue as "won't fix". Any action needed from me in that regard?

caksoylar:
I guess this isn't an architectural decision to "won't fix" so we can leave this open if you want.

amacleod:
Looks like #2848 does indeed solve my use case.

PR #2848, opened 2025-02-27
feat: Refine behavior_binding_event, nested trans

Nick-Munnich:
This PR essentially does multiple things in one commit. These are all quite linked to one another, I wasn't *quite* able to untangle them from each other.

- Refactors the existing sensor code, making encoder rotational tracking happen outside of behaviors and inside `sensors.c`. 
- Revamps `zmk_behavior_binding_event` so that it is actually a zmk event, i.e. is raised and managed by `event_manager`.
- Allow the nesting of `&trans` inside of other behaviors, getting rid of `ZMK_BEHAVIOR_TRANSPARENT` in the process.

Taking the explanation from #573:

 - position_state_changed
     - keycode_state_changed
     - layer_state_changed

 -  position_state_changed
     -  behavior_state_changed
          - keycode_state_changed
          - layer_state_changed

This refactor/feat will allow for much more advanced and elegant handling in the future. My goal is to use this to rework combos a bit further. Nesting `&trans` is also a benefit that should not be underestimated, it will make things like home-row mods much easier to set up. The sensor refactor should allow further work on that part to be simpler as well.

Needs documentation and lots of testing, particularly around the encoder bits as the tests we have are very rudimentary. There are likely some stylistic bits that could use tidying up as well.

Closes: #573, #532, #2368

## PR check-list


app/src/behavior.c
```diff
@@ -56,29 +58,58 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+        .timestamp = event.timestamp,
+        .type = pressed ? PRESS : RELEASE,
+#if IS_ENABLED(CONFIG_ZMK_SPLIT)
+        .source = event.source,
+#endif
+    });
+}
+
+// TODO: Pass the event in as pointers, rather than copying in the struct
+// Make this change as part of a breaking release
```
caksoylar, line 79: Isn't this passing it as a pointer already?

app/src/behavior.c
```diff
@@ -56,29 +58,58 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
-                                struct zmk_behavior_binding_event event, bool pressed) {
-    // We want to make a copy of this, since it may be converted from
-    // relative to absolute before being invoked
-    struct zmk_behavior_binding binding = *src_binding;
-
-    const struct device *behavior = zmk_behavior_get_binding(binding.behavior_dev);
+int behavior_listener(const zmk_event_t *eh) {
+    struct zmk_behavior_binding_event *event = as_zmk_behavior_binding_event(eh);
+    if (event == NULL) {
+        return -EINVAL;
```
caksoylar, line 97: Maybe add a log statement here.

app/src/behavior.c
```diff
@@ -93,29 +124,32 @@ int zmk_behavior_invoke_binding(const struct zmk_behavior_binding *src_binding,
...
+        return invoke_locally(&binding, event);
     case BEHAVIOR_LOCALITY_EVENT_SOURCE:
 #if ZMK_BLE_IS_CENTRAL // source is a member of event because CONFIG_ZMK_SPLIT is enabled
-        if (event.source == ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL) {
-            return invoke_locally(&binding, event, pressed);
+        if (event->source == ZMK_POSITION_STATE_CHANGE_SOURCE_LOCAL) {
+            return invoke_locally(&binding, event);
         } else {
-            return zmk_split_bt_invoke_behavior(event.source, &binding, event, pressed);
+            return zmk_split_bt_invoke_behavior(event->source, &binding, event);
```
caksoylar, line 133: Can get rid of the `else` due to `return`

caksoylar:
What documentation changes do you have in mind, besides updating the new behavior guide?

app/include/zmk/keymap.h
```diff
@@ -73,8 +74,9 @@ int zmk_keymap_save_changes(void);
 int zmk_keymap_discard_changes(void);
 int zmk_keymap_reset_settings(void);
 
-int zmk_keymap_position_state_changed(uint8_t source, uint32_t position, bool pressed,
-                                      int64_t timestamp);
+int zmk_keymap_raise_binding_event_at_layer_idx(zmk_keymap_layer_id_t layer_id, uint8_t source,
```
caksoylar, line 77: Probably should be `..._layer_id`, to disambiguate between `layer_id` and `layer_index`?

app/src/behavior.c
```diff
@@ -56,29 +58,58 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+    struct zmk_behavior_binding_event *event = as_zmk_behavior_binding_event(eh);
+    if (event == NULL) {
+        return -EINVAL;
+    }
 
+    const struct device *behavior = zmk_behavior_get_binding(event->binding->behavior_dev);
     if (!behavior) {
-        LOG_WRN("No behavior assigned to %d on layer %d", event.position, event.layer);
-        return 1;
+        LOG_WRN("No behavior assigned to %d on layer %d", event->position, event->layer);
```
caksoylar, line 102: Wouldn't this be an error rather than a warning?

app/src/behavior.c
```diff
@@ -56,29 +58,58 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
     }
 
-    int err = behavior_keymap_binding_convert_central_state_dependent_params(&binding, event);
+    /*
+     * Copying the behavior is necessary so that
+     * behavior_keymap_binding_convert_central_state_dependent_params (used for e.g. certain
+     * RGB behaviors) does not modify it, e.g. if the behavior is stored in a combo_cfg
+     */
+    struct zmk_behavior_binding binding = *(event->binding);
+    int err = behavior_keymap_binding_convert_central_state_dependent_params(&binding, *event);
```
caksoylar, line 112: Is this still making a copy somewhere? I remember I had to tweak the signature of `zmk_behavior_invoke_binding` so that it used a struct that was copy-by-value, rather than a pointer

app/src/behavior_queue.c
```diff
@@ -30,24 +30,29 @@ static K_WORK_DELAYABLE_DEFINE(queue_work, behavior_queue_process_next);
...
         if (item.press) {
-            zmk_behavior_invoke_binding(&item.binding, event, true);
+            event.type = PRESS;
         } else {
-            zmk_behavior_invoke_binding(&item.binding, event, false);
+            event.type = RELEASE;
+        }
+        ret = raise_zmk_behavior_binding_event(event);
+        if (ret < 0) {
+            LOG_DBG("Error %d occurred while processing behavior in queue.", ret);
```
caksoylar, line 54: Perhaps an error here as well

app/src/behavior_queue.c
```diff
@@ -30,24 +30,29 @@ static K_WORK_DELAYABLE_DEFINE(queue_work, behavior_queue_process_next);
...
 #endif
         };
 
         if (item.press) {
-            zmk_behavior_invoke_binding(&item.binding, event, true);
+            event.type = PRESS;
         } else {
-            zmk_behavior_invoke_binding(&item.binding, event, false);
+            event.type = RELEASE;
+        }
```
caksoylar, line 51: I like ternary statements for very simple if/elses like these.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -402,60 +404,72 @@ static inline const char *decision_moment_str(enum decision_moment decision_mome
 }
 
 static int press_hold_binding(struct active_hold_tap *hold_tap) {
+    struct zmk_behavior_binding binding = {.behavior_dev = hold_tap->config->hold_behavior_dev,
+                                           .param1 = hold_tap->param_hold};
     struct zmk_behavior_binding_event event = {
+        .binding = &binding,
         .position = hold_tap->position,
         .timestamp = hold_tap->timestamp,
+        .layer = hold_tap->layer,
```
caksoylar, line 413: Thanks for adding these, I actually wanted to depend on them for a custom behavior but behaviors weren't dilige enough in passing through `layer`.

app/src/behaviors/behavior_transparent.c
```diff
@@ -9,26 +9,23 @@
...
-static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
+static int on_keymap_binding_trigger(struct zmk_behavior_binding *binding,
                                      struct zmk_behavior_binding_event event) {
-    return ZMK_BEHAVIOR_TRANSPARENT;
-}
-
-static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
-                                      struct zmk_behavior_binding_event event) {
-    return ZMK_BEHAVIOR_TRANSPARENT;
+    return zmk_keymap_raise_binding_event_at_layer_idx(event.layer - 1, event.source,
```
caksoylar, line 22: Shouldn't this `event.layer - 1` be on the layer index, not ID? If so, it might be easier to handle the substitution inside the function itself. Or have a helper function to get you the layer ID of the previous layer by index.

app/src/keymap.c
```diff
@@ -685,146 +686,83 @@ int zmk_keymap_reset_settings(void) { return -ENOTSUP; }
...
+                    continue;
+                }
+                LOG_DBG("layer_id: %d sensor_index: %d, binding name: %s", candidate_layer,
+                        sensor_index, binding->behavior_dev);
+#else
+                return -ENOTSUP;
+#endif
             } else {
-                return ret;
+                binding = zmk_keymap_get_layer_binding_at_idx(candidate_layer, position);
```
caksoylar, line 719: Any performance/stack concerns with recursive calls, especially when we have >32 layers?

app/src/combo.c
```diff
@@ -290,28 +290,34 @@ static int release_pressed_keys() {
 
 static inline int press_combo_behavior(const struct combo_cfg *combo, int32_t timestamp) {
     struct zmk_behavior_binding_event event = {
+        .binding = &combo->behavior,
+        .layer = 0, // Combos don't have layers, so their layer is set to be the base layer.
```
caksoylar, line 294: That seems fair, though we could consider highest active layer that obey combo's layer filter.

app/src/combo.c
```diff
@@ -502,6 +508,7 @@ static int position_state_changed_listener(const zmk_event_t *ev) {
 static int keycode_state_changed_listener(const zmk_event_t *eh) {
     struct zmk_keycode_state_changed *ev = as_zmk_keycode_state_changed(eh);
     if (ev->state && !is_mod(ev->usage_page, ev->keycode)) {
+        LOG_DBG("help");
```
caksoylar, line 511: Seems accidental

app/src/events/behavior_binding_event.c
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2020 The ZMK Contributors
```
caksoylar, line 2: Copyright date

caksoylar (commented):
I haven't really reviewed the sensor parts since I am not familiar with those bits, but otherwise the broad strokes LGTM.

Nick-Munnich (commented):
Thanks for the review! Regarding documentation changes, it is pretty much just the new behavior guide that I was thinking of, at least for now. Some further changes in the future might result in a nested trans making for a nice example in e.g. hold-taps, but that shouldn't be a part of this PR imo.

app/src/behavior.c
caksoylar: It prints with `<err>` instead of `<wrn>` etc. in the logs, and you can filter the displayed logs via `*_LOG_LEVEL_*` Kconfigs.

app/src/behaviors/behavior_transparent.c
caksoylar: Layer IDs are the underlying identifier for a layer, whereas layer indices indicate actually where the layer sits on the layer stack. These are normally equal, unless the user reorder layers in the Studio (IIRC) in which case the layer indices are updated (but not IDs).

Most code doesn't care about using indices since layer ordering doesn't matter to them, but specifically for this case (transparent fallthroughs) we need to be using indices so that the layer ordering in Studio is respected. In this case, you seem to be using IDs only, IIUC.

app/src/behavior.c
caksoylar: You are right, I suppose L111 should be making a copy, similar to what was happening in L72 before. It'd be good to confirm via testing to be sure -- I was testing backlight behavior with nano LEDs on a simple split.

app/src/keymap.c
```diff
@@ -685,146 +686,83 @@ int zmk_keymap_reset_settings(void) { return -ENOTSUP; }
...
-    }
-
+int zmk_keymap_raise_binding_event_at_layer_idx(zmk_keymap_layer_id_t layer_id, uint8_t source,
+                                                uint32_t position, enum trigger_type type,
+                                                int64_t timestamp) {
     // We use int here to be sure we don't loop layer_idx back to UINT8_MAX
-    for (int layer_idx = ZMK_KEYMAP_LAYERS_LEN - 1;
-         layer_idx >= LAYER_ID_TO_INDEX(_zmk_keymap_layer_default); layer_idx--) {
-        zmk_keymap_layer_id_t layer_id = LAYER_INDEX_TO_ID(layer_idx);
+    for (int layer_idx = layer_id; layer_idx >= LAYER_ID_TO_INDEX(_zmk_keymap_layer_default);
```
caksoylar, line 693: Related to the other comment on ID vs. index, I believe this one should be looping over indices, not IDs, like before.

app/src/behavior.c
caksoylar: Got it, so the comment applies to the function calls inside. Though I don't know enough to say whether passing a pointer rather than a struct gives a non-trivial performance gain. I'd be interested to learn if e.g. @joelspadin has thoughts on it.

app/src/behavior.c
Nick-Munnich: I see. I think for this particular point we should leave it as a warning, because we may end up creating a system where a keymap is allowed to have blank spots at *some* point in the future.

app/src/behaviors/behavior_transparent.c
Nick-Munnich: Ahh, very good point. I hadn't considered that at all. Will fix.

app/include/zmk/keymap.h
Nick-Munnich: Considering the other conversationy bits I think renaming this to `..._layer_index` might be better?

caksoylar: I guess that function interface will be changed to use indices anyway, so it makes sense.

app/include/zmk/events/behavior_binding_event.h
```diff
@@ -0,0 +1,35 @@
...
+#endif // IS_ENABLED(CONFIG_ZMK_BEHAVIOR_LOCAL_IDS_IN_BINDINGS)
+    const char *behavior_dev;
+    uint32_t param1;
+    uint32_t param2;
+};
+
+enum trigger_type { PRESS, RELEASE, SENSOR };
+
+struct zmk_behavior_binding_event {
+    const struct zmk_behavior_binding *binding;
```
petejohanson: So.... In theory, none of these events need the local ID at all for what they're doing, so we're probably wasting a bit of stack size here by using the binding directly like this.

Also, this being a pointer to a binding.... Gets *messy* if any consumers are capturing these events, by copying them locally into a store and then swallowing the event... If the `bindings` points to a binding on the stack (like I see hold tap doing), then it'll point to random memory at a later time.

This probably needs to actually store a whole copy of the binding, we try to avoid pointers like this in events for this very reason.

app/include/zmk/events/behavior_binding_event.h
```diff
@@ -0,0 +1,35 @@
...
+struct zmk_behavior_binding {
+#if IS_ENABLED(CONFIG_ZMK_BEHAVIOR_LOCAL_IDS_IN_BINDINGS)
+    zmk_behavior_local_id_t local_id;
+#endif // IS_ENABLED(CONFIG_ZMK_BEHAVIOR_LOCAL_IDS_IN_BINDINGS)
+    const char *behavior_dev;
+    uint32_t param1;
+    uint32_t param2;
+};
+
+enum trigger_type { PRESS, RELEASE, SENSOR };
```
petejohanson: The enum name *and* the entries should get proper namespace prefixies, e.g.:

```suggestion
enum zmk_behavior_trigger_type {
    ZMK_BEHAVIOR_TRIG_TYPE_PRESS,
    ZMK_BEHAVIOR_TRIG_TYPE_RELEASE,
    ZMK_BEHAVIOR_TRIG_TYPE_SENSOR,
};
```

app/include/zmk/events/behavior_binding_event.h
```diff
@@ -0,0 +1,35 @@
...
+
+struct zmk_behavior_binding_event {
+    const struct zmk_behavior_binding *binding;
+    int layer;
+    uint32_t position;
+    int64_t timestamp;
+    enum trigger_type type;
+    uint8_t source;
```
petejohanson, line 39: Is there a reason the conditional for splits was removed from this?

app/src/behavior.c
```diff
@@ -55,29 +57,59 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
     return NULL;
 }
 
+// TODO: Delete this method as part of a breaking release
+int zmk_behavior_invoke_binding(const struct zmk_behavior_binding *src_binding,
+                                struct zmk_behavior_binding_event event, bool pressed) {
```
petejohanson:
```suggestion
__deprecated int zmk_behavior_invoke_binding(const struct zmk_behavior_binding *src_binding,
                                struct zmk_behavior_binding_event event, bool pressed) {
```

Probably needs to be in the correspondig header declaration too.

app/src/behavior.c
```diff
@@ -55,29 +57,59 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
+    LOG_DBG("`zmk_behavior_invoke_binding` is deprecated. Please raise a "
```
petejohanson:
```suggestion
    LOG_WRN("`zmk_behavior_invoke_binding` is deprecated. Please raise a "
```

app/src/behavior.c
```diff
@@ -55,29 +57,59 @@ const struct device *z_impl_behavior_get_binding(const char *name) {
...
-    // We want to make a copy of this, since it may be converted from
-    // relative to absolute before being invoked
-    struct zmk_behavior_binding binding = *src_binding;
-
-    const struct device *behavior = zmk_behavior_get_binding(binding.behavior_dev);
+int behavior_listener(const zmk_event_t *eh) {
+    struct zmk_behavior_binding_event *event = as_zmk_behavior_binding_event(eh);
+    if (event == NULL) {
+        LOG_ERR("An invalid event was passed as an argument somehow.");
+        return -EINVAL;
```
petejohanson: Catching these kinds of really unexpected conditions (that indicate a systemic issue or programmer error somehow) should really be an `__ASSERT` that would get caught during our tests, where `CONFIG_ASSERT` is enabled, but not incur the small runtime hit otherwise.

There's a good chance there's existing checks like this, which would benefit from the same treatment at some point.

app/src/behavior.c
petejohanson: Yup, that dereference like that will copy. No concerns there.

app/include/zmk/events/behavior_binding_event.h
```diff
@@ -0,0 +1,35 @@
...
+    const char *behavior_dev;
+    uint32_t param1;
+    uint32_t param2;
+};
+
+enum trigger_type { PRESS, RELEASE, SENSOR };
+
+struct zmk_behavior_binding_event {
+    const struct zmk_behavior_binding *binding;
+    int layer;
```
petejohanson: Probably should take the time to specifically use a smaller type here:

```suggestion
    uint8_t layer;
```

Also, moving these adjacent to the `source` may lead the compiler to pack this a bit smaller.

app/src/behaviors/behavior_hold_tap.c
```diff
@@ -402,60 +404,72 @@ static inline const char *decision_moment_str(enum decision_moment decision_mome
...
     };
 
-    struct zmk_behavior_binding binding = {.behavior_dev = hold_tap->config->hold_behavior_dev,
-                                           .param1 = hold_tap->param_hold};
-    return zmk_behavior_invoke_binding(&binding, event, true);
+    return raise_zmk_behavior_binding_event(event);
 }
 
 static int press_tap_binding(struct active_hold_tap *hold_tap) {
+    struct zmk_behavior_binding binding = {.behavior_dev = hold_tap->config->tap_behavior_dev,
```
petejohanson: Here's where I'm concerned about the event pointing to a stack allocated binding.

app/src/behaviors/behavior_sensor_rotate_common.c
```diff
@@ -4,73 +4,23 @@
...
 
     const int sensor_index = ZMK_SENSOR_POSITION_FROM_VIRTUAL_KEY_POSITION(event.position);
 
-    if (mode != BEHAVIOR_SENSOR_BINDING_PROCESS_MODE_TRIGGER) {
-        data->triggers[sensor_index][event.layer] = 0;
-        return ZMK_BEHAVIOR_TRANSPARENT;
-    }
-
-    int triggers = data->triggers[sensor_index][event.layer];
+    struct zmk_sensor_data *data = zmk_sensor_get_data(sensor_index);
```
petejohanson: This API should be returning `const` pointer probably, not good to return pointers that callers then change values of.

app/src/behaviors/behavior_transparent.c
```diff
@@ -9,26 +9,28 @@
...
+static int on_keymap_binding_trigger(struct zmk_behavior_binding *binding,
                                      struct zmk_behavior_binding_event event) {
-    return ZMK_BEHAVIOR_TRANSPARENT;
-}
-
-static int on_keymap_binding_released(struct zmk_behavior_binding *binding,
-                                      struct zmk_behavior_binding_event event) {
-    return ZMK_BEHAVIOR_TRANSPARENT;
+    // Avoid uint8_t overflow resulting in an infinite loop
+    if (LAYER_ID_TO_INDEX(event.layer) == 0) {
```
petejohanson: Store this to a local if we're gonna be using this twice, since it may involve some lookups.

app/src/keymap.c
petejohanson: We'll want to actually verify the generated assembly is doing this as a tail call.

nmunnich: Verified that this is function ends in a tail call. Built the nice60 with the current version. The `raise_zmk_behavior_binding_event` call becomes a `b.w` instruction, which I believe is a wide branch instruction, not a branch with link.

app/include/zmk/events/behavior_binding_event.h
nmunnich: No particular reason; I just missed it. Fixed.

caksoylar:
Can you update the PR description and the new behavior docs for the latest changes?

nmunnich:
I've moved the behavior_dev and the params into the event now. This allowed for a significant refactoring and simplification of lots of parts. I think writing a behavior has become simpler, but interacting with the behavior queue has become slightly more annoying. Documentation still to come.

PR #2849, opened 2025-02-27
Combo RAM and Kconfig reduction 

petejohanson:
Keeping this draft for now, but this is a collection of improvements to remove a few combo Kconfig flags that are instead auto-calculated at build time, and reduce the RAM usage on more constrained devices.

A Le Chiffre STM32 Build on `main`:

```
[198/198] Linking C executable zephyr/zmk.elf
           FLASH:       42636 B       128 KB     32.53%
             RAM:        8120 B        15 KB     52.86%
     RetainedMem:          0 GB          1 B      0.00%
        IDT_LIST:          0 GB         2 KB      0.00%
```

Same keymap with this PR:

```
❯ west build -d build/lc_stm32_tap_dance
[7/7] Linking C executable zephyr/zmk.elf
           FLASH:       42732 B       128 KB     32.60%
             RAM:        7736 B        15 KB     50.36%
     RetainedMem:          0 GB          1 B      0.00%
        IDT_LIST:          0 GB         2 KB      0.00%
```

For a savings of 384 bytes.

Needs some more cleanup, and there's room for some performance improvements for the access to the new bit fields, but this should be ready for early testing.

## PR check-list

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -444,12 +444,16 @@ config ZMK_COMBO_MAX_PRESSED_COMBOS
     default 4
 
 config ZMK_COMBO_MAX_COMBOS_PER_KEY
-    int "Maximum number of combos per key"
-    default 5
+    int
+    default 0
+    help
+      Deprecated: Storage for combos is now determined automatically
 
 config ZMK_COMBO_MAX_KEYS_PER_COMBO
-    int "Maximum number of keys per combo"
-    default 4
+    int
+    default 0
+    help
+       Deprecated: This is now auto-calculated based on `key-positions` in devicetree
 
 # Combo options
 endmenu
--- a/app/src/combo.c
+++ b/app/src/combo.c
@@ -9,6 +9,7 @@
 #include <zephyr/device.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/sys/dlist.h>
+#include <zephyr/sys/util.h>
 #include <zephyr/kernel.h>
 
 #include <drivers/behavior.h>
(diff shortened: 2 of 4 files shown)


app/src/combo.c
```diff
@@ -48,27 +66,66 @@ struct active_combo {
...
+                        .virtual_key_position = ZMK_VIRTUAL_KEY_POSITION_COMBO(__COUNTER__),       \
+                        .slow_release = DT_PROP(n, slow_release),                                  \
+                        .layer_mask = NODE_PROP_BITMASK(n, layers),                                \
+                    }, ),                                                                          \
+                ())
+
+#define COMBO_CONFIGS_WITH_MATCHING_POSITIONS_LEN(positions, _ignore)                              \
+    DT_INST_FOREACH_CHILD_VARGS(0, COMBO_INST, positions)
+static const struct combo_cfg combos[] = {
+    LISTIFY(10, COMBO_CONFIGS_WITH_MATCHING_POSITIONS_LEN, (), 0)};
```
caksoylar, line 106: Should `10` be `MAX_COMBO_KEYS` instead? I have 30% confidence that I understand what this is doing, but it sounds like it is iterating over possible `key_positions` lengths.

petejohanson: This deserved a comment: `MAX_COMBO_KEYS` is not a pre-processor constant, but a C union trick, so can't be used as a parameter to `LISTIFY`, so we basically pick an arbitrary high value that won't tax the pre-processor too much. I picked 10, since it's unlikely for someone to combo with more fingers than they have.

caksoylar: I guess it doesn't hurt to set this to a high number, so maybe we go 'til 255 (devicetree limit, IIRC)? Unless you think preprocessor can slow down the build enough.

petejohanson: > I guess it doesn't hurt to set this to a high number, so maybe we go 'til 255 (devicetree limit, IIRC)? Unless you think preprocessor can slow down the build by a significant amount.

It *might*, and honestly, this is iterating by key-positions *length*, so I'm having a hard time imagining we need to support a combo triggered by 256 different key positions?

caksoylar: Right, just thought it was perhaps a number with lesser-magic involved than 10, it's fine to keep that too (perhaps with a comment and noting the limitation in docs).

app/src/combo.c
```diff
@@ -145,66 +168,59 @@ static bool is_quick_tap(const struct combo_cfg *combo, int64_t timestamp) {
...
+                if (candidate->key_positions[kp] == position) {
+                    matches++;
+                    found = true;
+                    break;
+                }
+            }
+
+            if (!found) {
+                sys_bitfield_clear_bit((mem_addr_t)&candidates, i);
+            }
```
Nick-Munnich, line 203: Shouldn't we be able to just `&` the candidates and the combo lookup bitfields with one another?

app/src/combo.c
```diff
@@ -250,17 +257,13 @@ static int filter_timed_out_candidates(int64_t timestamp) {
...
-        if (candidates[i].combo == NULL) {
-            return i;
-        }
-        candidates[i].combo = NULL;
-    }
-    return CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY;
+    memset(candidates, 0, BYTES_FOR_COMBOS_MASK);
+
+    return MAX_COMBO_KEYS;
 }
```
Nick-Munnich, line 263: I'd say remove this function and just stick it inline. The return value is pointless.

app/src/combo.c
```diff
@@ -26,52 +27,104 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+
+#define COMBOS_KEYS_BYTE_ARRAY(node_id)                                                            \
+    uint8_t _CONCAT(combo_prop_, node_id)[DT_PROP_LEN(node_id, key_positions)];
+
+#define MAX_COMBO_KEYS sizeof(union {DT_INST_FOREACH_CHILD(0, COMBOS_KEYS_BYTE_ARRAY)})
+
 struct combo_cfg {
-    int32_t key_positions[CONFIG_ZMK_COMBO_MAX_KEYS_PER_COMBO];
+    int32_t key_positions[MAX_COMBO_KEYS];
     int32_t key_position_len;
```
Nick-Munnich, line 50: There is also the possibility of using bit fields for this as well.

app/src/combo.c
```diff
@@ -361,8 +367,8 @@ static void deactivate_combo(int active_combo_index) {
         memcpy(&active_combos[active_combo_index], &active_combos[active_combo_count],
                sizeof(struct active_combo));
     }
-    active_combos[active_combo_count].combo = NULL;
     active_combos[active_combo_count] = (struct active_combo){0};
+    active_combos[active_combo_count].combo_idx = INT32_MAX;
```
Nick-Munnich, line 371: This feels a bit weird stylistically. Is there a reason we need to apply the value after creation?

app/Kconfig
```diff
@@ -448,12 +448,16 @@ config ZMK_COMBO_MAX_PRESSED_COMBOS
     default 4
 
 config ZMK_COMBO_MAX_COMBOS_PER_KEY
-    int "Maximum number of combos per key"
-    default 5
+    int
+    default 0
+    help
+      Deprecated: Storage for combos is now determined automatically
```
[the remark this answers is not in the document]
petejohanson: When we have future studio use, we'll likely "pre-allocate" things by setting placeholders up in the devicetree, like we do with reserved layers, so our auto-calculation will still be usable there as well.

app/src/combo.c
```diff
@@ -26,52 +27,104 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+
+#define COMBO_CHILDREN_COUNT (0 DT_INST_FOREACH_CHILD(0, COMBO_ONE))
+
+// TODO: Expand this more. Need an actual constant value to be used for LISTIFY later.
+#if COMBO_CHILDREN_COUNT > 32
+#define BYTES_FOR_COMBOS_MASK 5
+#else
+// The smallest we can go is 4 bytes, to avoid unaligned access issues with bitfield ops
+#define BYTES_FOR_COMBOS_MASK 4
+#endif
```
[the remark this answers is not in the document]
petejohanson: Ah, I needed this "pre-processor constant" when I was attempting to make the `struct combo_cfg`'s `key_positions` a bit field as well, but that caused a wide range of issues. Will be simplifying this to `#define BYTES_FOR_COMBOS_MASK MAX(4, DIV_ROUND_UP(COMBO_CHILDREN_COUNT, 8))` which is basically equivalent.

app/src/combo.c
petejohanson: This got messy *fast*. I attempted to do that and hit lots of funky edge cases by doing so. I think it *might* be doable, but think we should try that in follow up work.

This was actually the feature that required the `MAX_COMBO_KEYS` to be a pre-processor constant since it required some painful magic to generate the constant uint8_t array bitfield content.

app/src/combo.c
petejohanson: It's a placeholder special value that indicates this entry is actually "empty". The alternative would be to store values offset by `1` so that `0` could be "unset"/"empty"

app/Kconfig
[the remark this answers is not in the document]
Nick-Munnich: Closer inspection, seems doable for `MAX_COMBOS_PER_KEY`, but less so for vice-versa: what if DT has only 2 and 3 key combos, and then someone wants to set up a 5 key one in Studio?

Nick-Munnich:
Joel had an interesting comment [here](https://github.com/zmkfirmware/zmk/pull/2846#issuecomment-2689598874) that might also apply to this PR.

app/src/combo.c
petejohanson: Indeed, with some tweaks for counting remaining bits.

petejohanson:
> Joel had an interesting comment [here](https://github.com/zmkfirmware/zmk/pull/2846#issuecomment-2689598874) that might also apply to this PR.

Moved to uint32_t type for the bit fields to avoid alignment issues.

app/src/combo.c
```diff
@@ -105,17 +105,17 @@ static const struct combo_cfg combos[] = {
 #define COMBO_CHILDREN_COUNT (0 DT_INST_FOREACH_CHILD(0, COMBO_ONE))
 
 // We need at least 4 bytes to avoid alignment issues
-#define BYTES_FOR_COMBOS_MASK MAX(4, DIV_ROUND_UP(COMBO_CHILDREN_COUNT, 8))
+#define BYTES_FOR_COMBOS_MASK MAX(4, DIV_ROUND_UP(COMBO_CHILDREN_COUNT, 32))
```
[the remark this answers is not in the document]
petejohanson: Nope. Removed.

caksoylar:
Would daily testing be useful in the current state of the PR?

petejohanson:
> Would daily testing be useful in the current state of the PR?

I believe so, but I need to do so myself first to be 100% sure, even though tests are all green.

app/src/combo.c
```diff
@@ -145,66 +162,56 @@ static bool is_quick_tap(const struct combo_cfg *combo, int64_t timestamp) {
...
+    while (field) {
+        field &= (field - 1);
+        count++;
     }
-    // clear unmatched candidates
-    for (int i = matches; i < CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY; i++) {
-        candidates[i].combo = NULL;
+
+    return count;
+}
```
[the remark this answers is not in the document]
petejohanson: Good call. I tweaked this a bit, roughly modeling what you described, to save on some cycles. Thanks!

app/src/combo.c
petejohanson: Done.

app/src/combo.c
```diff
@@ -26,53 +27,99 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+// reasonably press simultaneously with 10 fingers is 20 keys, two keys per finger.
+static const struct combo_cfg combos[] = {
+    LISTIFY(20, COMBO_CONFIGS_WITH_MATCHING_POSITIONS_LEN, (), 0)};
+
+#define COMBO_ONE(n) +1
 
-uint32_t pressed_keys_count = 0;
+#define COMBO_CHILDREN_COUNT (0 DT_INST_FOREACH_CHILD(0, COMBO_ONE))
+
+// We need at least 4 bytes to avoid alignment issues
```
Nick-Munnich, line 107:
```suggestion
```

app/src/combo.c
```diff
@@ -26,53 +27,99 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
+    DT_INST_FOREACH_CHILD_VARGS(0, COMBO_INST, positions)
+
+// We do some magic here to generate the `combos` array by "key position length", looping
+// by key position length and on each iteration, only include entries where the `key-positions`
+// length matches.
+// Doing so allows our bitmasks to be "shorted key positions list first" when searching for matches.
+// `20` is chosen as a reasonable limit, since the theoretical maximum number of keys you might
+// reasonably press simultaneously with 10 fingers is 20 keys, two keys per finger.
+static const struct combo_cfg combos[] = {
+    LISTIFY(20, COMBO_CONFIGS_WITH_MATCHING_POSITIONS_LEN, (), 0)};
```
Nick-Munnich, line 101: I think with the changes to `filter_candidates` this sorting is no longer necessary.

app/src/combo.c
```diff
@@ -26,53 +27,99 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
...
-    // the virtual key position is a key position outside the range used by the keyboard.
-    // it is necessary so hold-taps can uniquely identify a behavior.
-    int32_t virtual_key_position;
-    int32_t layers_len;
-    int8_t layers[];
 };
 
 struct active_combo {
-    const struct combo_cfg *combo;
+    int16_t combo_idx;
```
Nick-Munnich, line 61: Why not uint16?

app/src/combo.c
```diff
@@ -145,66 +162,58 @@ static bool is_quick_tap(const struct combo_cfg *combo, int64_t timestamp) {
...
-        // LOG_DBG("combo timeout %d %d %d", position, i, candidates[i].timeout_at);
     }
+
     return number_of_combo_candidates;
 }
 
+static uint8_t zero_one_or_more_bits(uint32_t field) {
+    uint8_t count = 0;
+    while (field && count < 2) {
+        field &= (field - 1);
```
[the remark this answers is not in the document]
petejohanson: I'm not sure I'm seeing the extra loop... We need to iterate up until 2, since we need to know if there was more than 1 match as another condition to check for.

app/src/combo.c
petejohanson: Yeah, we could, not sure we'd ever hit that many combos, but refactored anyways. Thanks.

app/src/combo.c
petejohanson: Not necessary, but I think still will lead to faster matching for combos with fewer keys.

app/src/combo.c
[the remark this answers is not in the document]
nmunnich: Say that field=111. we have:
```
count = 0
check field and count (pass, would return 0 if fail)
field = 110
count = 1
check field and count (pass, would return 1 if fail)
field = 101
count = 2
check field and count (fails)
return 2
```
while with the code I showed in the earlier review, we have
```
check field (fail, would return 0 if passed)
field = 110
check field (fail, would return 1 if passed)
return 2
```
There are only 2 checks of field necessary.

app/src/combo.c
```diff
@@ -145,66 +162,58 @@ static bool is_quick_tap(const struct combo_cfg *combo, int64_t timestamp) {
...
-            candidate_idx++;
-        }
-    }
-    // clear unmatched candidates
-    for (int i = matches; i < CONFIG_ZMK_COMBO_MAX_COMBOS_PER_KEY; i++) {
-        candidates[i].combo = NULL;
     }
-    // LOG_DBG("combo matches after filter %d", matches);
+
+    LOG_DBG("combo matches after filter %d", matches);
```
nmunnich, line 200:
```suggestion
if(matches < 2){
    LOG_DBG("combo matches after filter %d.", matches);
   }
else {
LOG_DBG("combo matches after filter >= 2");
}
```

app/src/combo.c
nmunnich: I think it only matters for position press down, but sure we can keep it in to save on that.

app/Kconfig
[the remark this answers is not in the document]
nmunnich: I suppose we can just handle that once we get around to adding combos to Studio.

nmunnich (approved):
The two remaining comments of mine are both minor nits in my eyes. I'd be happy to see this go through as-is.

app/src/combo.c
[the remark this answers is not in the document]
petejohanson: Ah, indeed. Replaced with your original implementation, locally tests are still all green.

app/src/combo.c
petejohanson: I dislike this, as it adds conditional code that *might* not get optimized out if logging is disabled entirely. Would prefer as is.

nmunnich: Maybe add on "Match numbers above 2 are presented as 2." to the end of the line, then? I was wavering on that but couldn't find a nice phrasing. Bad phrasing is better than no phrasing, though.

caksoylar:
A heads up: If you set the removed settings, the build fails because the help message somehow didn't work?

> error: ZMK_COMBO_MAX_COMBOS_PER_KEY (defined at /home/.../zmk/app/Kconfig:446) is assigned in a
configuration file, but is not directly user-configurable (has no prompt).

PR #2854, opened 2025-03-04
refactor(split): use LOG_HEXDUMP_DBG() to print position state data

xudongzheng:
The previous code prints one line per byte, making debugging difficult.


petejohanson (approved):
Indeed. Thanks.

PR #2855, opened 2025-03-05
feat: key press behavior supports mouse buttons

klesh:
# Summary

This is a Proof of Concept (PoC) demonstrating the concept of key - press behavior supporting mouse buttons. I'll explain the rationale and the implementation method. If the maintainer team deems this idea acceptable, my next steps will be to further develop the document and work on ZMK Studio.
My plan is to enable ZMK Studio to display Mouse Buttons along with LayerTap information, similar to the following:

image

# Why

Currently, mouse button support is implemented through `behavior-mouse-key-press` alongside `behavior-key-press`,  while it works for most of the cases,  I ran into some problems when working on my trackball :

1. I tried to implement LayerTap with Mouse Button 4(History back on most browser). However, the `&mkp` can not be used in conjunction with `&lt` and other composite behavior: `&it` assumes `param2` is a `key-press key`.
2. The ZMK Studio cannot display the `Mouse Button`. It just showing nothing in the key box while it should be something like `BTN1` intuitively.

# What

According to [the document from usb.org](https://usb.org/sites/default/files/hut1_6.pdf), there is a **Button Page** in the HID Usage. So I think it makes more sense to integrate mouse button support into the `key-press-behavior` rather than using the standalone `mouse-key-press-behavior`. So I implemented it on my local machine and tested it and it seemed to be working on my setup.

1. Added 5 `Mouse Button IDs` and defined `Mouse Button Keys` accordingly
2. Added an interceptor into `behavior-key-press` to handle the `Button Page` events.
3. Added an `input-listener` to the `kp` device to process the button-events

# How

With the above approach, the Mouse button can be defined by using the `&kp` behavior and used in conjunction with the `&lt` too.
And, with a simple patch to ZMK Studio we can make it support mouse button display without too much work.

## PR check-list


Nick-Munnich:
Hi! While I cannot comment on the approach (Pete probably has a reason for not tackling it this way), I would like to point out that neither of your "Why" reasons really apply:

* `&lt` is a special case of hold-tap. Hold-tap can be used with `&mkp`.
* This is more an issue on the side of ZMK Studio. This should be solved by improvements to the Studio UI, not through changes to ZMK.

klesh:
> Hi, thanks for the PR! While I cannot comment on the approach (Pete probably has a reason for not tackling it this way), I would like to point out that neither of your "Why" reasons really apply:

@Nick-Munnich Thanks for the information. I just started looking into ZMK, please bear with me if I ask something dumb 🙂


> 
> * `&lt` is a special case of hold-tap. Hold-tap can be used with `&mkp`.

Thanks for pointing that out. You're right; it didn't occur to me. So, I need to define a new behavior. Alongside `Layer - Tap`, perhaps `Layer - Click`? That could work. However, I still think it would be more intuitive to allow the key - press behavior to accept mouse buttons. I'm curious if you could prompt Pete for more input?

> * This is more an issue on the side of ZMK Studio. This should be solved by improvements to the Studio UI, not through changes to ZMK.

Apologies for not clearly conveying the rationale. My point is that the Mouse Buttons are part of the `hid usage` definition. It would be more concise if ZMK output the correct `hid usage id` for them instead of `0`.

klesh:
Hi, @nmunnich  @petejohanson 

Sorry for bothering you. 
I would need the information on why the approach is not acceptable to move the PR forward or close it:

1. I would like to investigate the solution to the reason if it is possible.
2. Or it should fail, I would like to switch to focus on the ZMK studio to mitigate the situation.

Please show me some pointers if you have time.

klesh:
> My two cents is that the separation between `&kp` and `&mkp` is nice[1](#user-content-fn-1-62b5e1b9e4539d08775bcb715224a21e) and adding Studio support to `&mkp` is a less invasive change than this. I am not opposed strongly, but I would have wished this to happen on the initial mouse keys merge if it was going to.
> 
> ## Footnotes
> 1. Counterpoint: Removal of `&cp` a few years ago to unify consumer and keyboard keycodes. [↩](#user-content-fnref-1-62b5e1b9e4539d08775bcb715224a21e)

I see, thanks for the input. Closing the PR now.

PR #2888, opened 2025-03-20
docs: update hold-tap page

nmunnich:
Hold-tap page was in need of some editing and improving. Should be clearer now. I removed the mod-tap page and the layer-tap section in favor of folding both into the hold-tap page, to make the connection and configuration clearer. I also decided that considering the complexity of hold-tap, it would be better to show some examples before the configuration options are listed.


docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -8,261 +8,175 @@ import TabItem from "@theme/TabItem";
...
-    keymap {
-        compatible = "zmk,keymap";
-        default_layer {
-            bindings = <
-                &bhm LCTRL A &bhm LGUI S &bhm LALT D &bhm LSHIFT F
-            >;
+            hold-trigger-key-positions = < ... >; // List of keys on the left hand side of the keyboard
+            hold-trigger-on-release;
         };
     };
```
caksoylar, line 164:
```suggestion
        hmr: home_row_mod_right {
            compatible = "zmk,behavior-hold-tap";
            #binding-cells = <2>;
            flavor = "balanced";
            require-prior-idle-ms = <150>;
            tapping-term-ms = <280>;
            quick-tap-ms = <175>;
            bindings = <&kp>, <&kp>;
            hold-trigger-key-positions = < ... >; // List of keys on the left side of the keyboard
            hold-trigger-on-release;
        };
    };
```
If we are encouraging copy pasting, why not define both?

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -8,261 +8,175 @@ import TabItem from "@theme/TabItem";
...
 };
 ```
 
+If you wish to use this configuration of homerow mods, you will need to define the corresponding left and right behaviors. Below is a brief overview of the options set.
```
caksoylar, line 168:
```suggestion
If you wish to use this configuration of homerow mods, you will need to define the corresponding left and right behaviors as above, then use `&hml` for homerow keys on the left side of the keyboard, and `&hmr` for ones on the right side.

Below is a brief overview of the options set.
```

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -8,261 +8,175 @@ import TabItem from "@theme/TabItem";
...
-        };
-    };
-
-    keymap {
-        compatible = "zmk,keymap";
-        default_layer {
-            bindings = <
-                &bhm LCTRL A &bhm LGUI S &bhm LALT D &bhm LSHIFT F
-            >;
+            hold-trigger-key-positions = < ... >; // List of keys on the left hand side of the keyboard
```
caksoylar, line 161:
```suggestion
            hold-trigger-key-positions = < ... >; // List of keys on the right side of the keyboard
```

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -8,261 +8,175 @@ import TabItem from "@theme/TabItem";
...
     };
 };
 ```
 
+If you wish to use this configuration of homerow mods, you will need to define the corresponding left and right behaviors. Below is a brief overview of the options set.
+
+- The ["balanced" flavor](#interrupt-flavors) produces a "hold" if another key is both pressed and released within the tapping-term.
+  This interrupt flavor is primarily used to decide between hold and tap, rather than using [tapping-term-ms](#tapping-term-ms).
+  This matches how modifiers are used a majority of the time (hold shift, tap a key, release shift).
+- [`require-prior-idle-ms`](#require-prior-idle-ms) is set to 125ms to resolve to taps immediately when typing at speed, which all but completely eliminates any typing delay.
```
caksoylar, line 173:
```suggestion
- [`require-prior-idle-ms`](#require-prior-idle-ms) is set to 125ms to resolve to taps immediately when typing at speed, which helps eliminate typing delay.
```

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -8,261 +8,175 @@ import TabItem from "@theme/TabItem";
...
+- [Positional hold taps](#positional-hold-tap-and-hold-trigger-key-positions) with [`hold-trigger-on-release`] are used to avoid accidental hold resolutions when typing sequences of letters all with the same hand.
+- [`tapping-term-ms`](#tapping-term-ms) is set to a higher value of 280ms, but not unreasonably high, so that same hand modifiers can still be used with intention.
+- [`quick-tap-ms`](#quick-tap-ms) is set to 175ms, for the event where a key may wish to be held down.
```
caksoylar, line 176:
```suggestion
- [`quick-tap-ms`](#quick-tap-ms) is set to 175ms, for the event where a tapped key may wish to be held down.
```

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -8,261 +8,175 @@ import TabItem from "@theme/TabItem";
...
+
+Depending on your typing behavior, you may wish to adjust this configuration somewhat. For example, adjusting the specific timings used, or switching the flavor to `tap-unless-interrupted`.
```
caksoylar, line 178:
```suggestion
Depending on your typing behavior, you may wish to adjust this configuration. For example, you can adjust the specific timings used, or define another set of hold-taps for more dextrous fingers like the index with different parameters.
```

I guess this is a bit subjective opinion territory but `tap-unless-interrupted` is niche and I am not sure it is worth mentioning here.

caksoylar (commented):
Thank you for tackling this; it's a huge amount of work but it will be a really nice improvement to one of the most important behavior doc pages.

A couple additional things:
- Need to do a pass on "home-row" vs. "homerow", I am fine either way as long as it's consistent
- We might consider adding one-line examples under each of the configuration options, like `tapping-term-ms = <200>;` (maybe excluding the ones with large examples already, for which I suggested a `details` box)

docs/docs/keymaps/behaviors/hold-tap.mdx
nmunnich: Because they are identically barring hold-trigger-key-positions, which they need to fill out themselves anyway.

docs/docs/keymaps/behaviors/hold-tap.mdx
nmunnich: I think it might be useful to mention the niche cases where it *can* come in useful. I feel like people often struggle to understand "I want this sort of tapping behavior -> Ok this particular flavor is ideal for that".

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -337,6 +251,160 @@ This hold-tap example implements a [momentary-layer](layers.md#momentary-layer)
...
+
+:::info
+Note that `hold-trigger-key-positions` is an array of key position indexes. Key positions are numbered sequentially according to your keymap, starting with 0. So if the first key in your keymap is Q, this key is in position 0. The next key (probably W) will be in position 1, et cetera.
+:::
+
+#### `hold-trigger-on-release`
+
+If set, instead of the keys listed in `hold-trigger-key-positions` producing a tap when _pressed_ before `tapping-term-ms` expires, they instead produce a tap when _released_ before `tapping-term-ms` expires.
+
+#### Example
```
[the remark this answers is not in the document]
nmunnich: I see your point, but I'm not personally a fan of using details boxes here. I think with the reordering of the config settings it should be fine.

docs/docs/keymaps/behaviors/hold-tap.mdx
caksoylar: I was being polite, I don't think this flavor should be recommended given how counter-intuitive it is with the tapping term. (It is already documented in the flavors.)

docs/docs/keymaps/behaviors/hold-tap.mdx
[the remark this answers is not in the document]
caksoylar: I think they take up too much real estate and inflates their importance relative to other settings, just because they end up being documented for a full page height. I'd try to at least make them more concise in this case.

docs/docs/keymaps/behaviors/hold-tap.mdx
caksoylar: True, but I think the "use `&hml`/`&hmr`" remark on the other suggestion is necessary and not sure how to do that without defining `hmr` in some way.

docs/docs/keymaps/behaviors/hold-tap.mdx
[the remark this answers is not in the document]
nmunnich: That's fair. I'll shove them in details boxes.

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -337,6 +251,160 @@ This hold-tap example implements a [momentary-layer](layers.md#momentary-layer)
...
+
+If `retro-tap` is enabled, the tap behavior is triggered when releasing the hold-tap key if no other key was pressed in the meantime. The hold key does not activate until another key is pressed, meaning that it cannot be used for mouse events like Shift Click to select from your cursor position to mouse position.
+
+For example, if you press `&mt LEFT_SHIFT A` and then release it without pressing another key, it will output `a`.
+
+```dts
+&mt {
+    retro-tap;
+};
+```
```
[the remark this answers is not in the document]
nmunnich: This is counter to your other suggestion of putting small example configurations under everything. I've added small examples under everything now, take a look and see what you think. I think we do all or nothing.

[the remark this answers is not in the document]
caksoylar: I was aware when I added this comment, I imagined the examples without `&mt {` part but that's fine too, thanks.

docs/docs/keymaps/behaviors/hold-tap.mdx
caksoylar: OK, I disagree but I won't push it further.

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -337,6 +263,192 @@ This hold-tap example implements a [momentary-layer](layers.md#momentary-layer)
...
-### Comparison to QMK
+## Custom Hold-Tap Configuration
+
+### `tapping-term-ms`
 
-The `hold-preferred` flavor works similar to the `HOLD_ON_OTHER_KEY_PRESS` setting in QMK. The `balanced` flavor is similar to the `PERMISSIVE_HOLD` setting, and the `tap-preferred` flavor is the QMK default.
+This value defines how long a key must be pressed to trigger the "hold" behavior, alongside other factors described in [interrupt flavors](#interrupt-flavors). The default is 200ms.
+
+```dts
+&my_mt{
```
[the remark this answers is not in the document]
nmunnich: Mostly me copypasting different bits. Which one do you think would be preferable?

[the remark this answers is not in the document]
caksoylar: I think it's better to avoid referring to non-predefined behaviors if we can, so I'd prefer `&mt`.

caksoylar (approved):
Thanks again!

PR #2889, opened 2025-03-20
feat(power): Allow runtime toggling of deep sleep functionality

awkannan:
Opening this as a draft just to get eyes on it.

There's nothing really preventing runtime toggling of deep sleep functionality - and it allows users more configurability without making them recompile / reflash.

I struggled getting automated tests to work - for some reason, my `CONFIG_ZMK_SLEEP=y` was not getting picked up when the test suite built - I left a note in discord about it.

I have tested the change on real hardware (including on a split keyboard, to make sure both halves act properly). From my testing, it seems to work.

Once I get some feedback, I can add docs.

## PR check-list

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -43,6 +43,7 @@ target_sources_ifdef(CONFIG_USB_DEVICE_STACK app PRIVATE src/events/usb_conn_sta
 target_sources(app PRIVATE src/behaviors/behavior_reset.c)
 target_sources_ifdef(CONFIG_ZMK_EXT_POWER app PRIVATE src/behaviors/behavior_ext_power.c)
 target_sources_ifdef(CONFIG_ZMK_BEHAVIOR_SOFT_OFF app PRIVATE src/behaviors/behavior_soft_off.c)
+target_sources_ifdef(CONFIG_ZMK_SLEEP app PRIVATE src/behaviors/behavior_set_sleep.c)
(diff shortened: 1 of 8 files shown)


caksoylar:
I think this is a bit orthogonal to the functionality/code added here, but once we have such a behavior it might be nice to have a `SLEEP_NOW` parameter that triggers deep sleep.

badjeff:
You might be interested in checking out this module. https://github.com/badjeff/zmk-behavior-insomnia/

awkannan:
@caksoylar I don't think adding that would be too difficult - I'm thinking of setting "activity_last_uptime" to 0 in a helper function that's called from a SLEEP NOW behavior. This should force sleep as long as sleep is enabled.

@badjeff that is super neat - if I understand correctly, the main difference is insomnia only applies if the board is actually connected to something (peripheral to central, or central to PC via BLE).  If peripheral isn't paired to a central, it'll still sleep. Similarly, if the central isn't connected to a PC, it'll also sleep.

I think badjeff's change is a bit more complex, but also makes total sense. 

One thing that I do like about working in activity.c is there is a single place where sleep/wake is handled vs. inside the behavior - but also fully understand why it's in the behavior (because nothing has to be merged into core ZMK).

Unsure if I want to keep working on this one, or just pull in the insomnia behavior to my board (which essentially provides a version of this behavior)

Any thoughts from the ZMK team?

caksoylar:
I personally think this feature would be good to have in core, but that decision is up to @petejohanson.

PR #2893, opened 2025-03-21
fix: changed shebang to make scripts more platform independent

Tygo-van-den-Hurk:
See for example [this thread](https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash) on why its better to use `#!/usr/bin/env bash` instead of `#!/bin/bash`. Without this change these scripts will not run on some platforms. This is not a breaking change for the platforms it already currently works on. 

## PR check-list

### DIFF
--- a/app/run-ble-test.sh
+++ b/app/run-ble-test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Copyright (c) 2023 The ZMK Contributors
 # SPDX-License-Identifier: MIT
--- a/docs/src/templates/setup.sh.mustache
+++ b/docs/src/templates/setup.sh.mustache
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Copyright (c) 2020 The ZMK Contributors
 # SPDX-License-Identifier: MIT


caksoylar (approved):

caksoylar:
Out of curiosity, under what systems does it not work? I am aware of this practice but I'd like to learn when this becomes an issue in practice for `bash`.

nmunnich:
Should this change also be applied to e.g. `app/run-test.sh`?

Tygo-van-den-Hurk:
No but good catch, this is cause that’s using `sh` instead of `bash`, and it’s defined in the POSIX standard that `sh` **must** be at `/bin/sh`. So any computer that understands shebangs will also have `sh` at that location, for any other program other then `sh`, such as `bash` this isn’t the case. Though technically `#!/usr/bin/env sh` would work.

Tygo-van-den-Hurk:
Haha no I get it. It feels cleaner if they’re all the same...

As for why, I think that’s cause people just use bash, so they put `#!/bin/bash` down. I don’t think they use any bash specific features though... so changing it to `#!/bin/sh` most likely wouldn’t break anything since they follow the same POSIX standard.

caksoylar:
There are a few [bashisms](https://mywiki.wooledge.org/Bashism) used in the bash scripts, usually you need a bit more diligence when writing POSIX sh. In my experience, it isn't too hard to make them POSIX compatible but replacing some bash features like arrays can be quite annoying.

petejohanson (approved):
Thanks!

PR #2901, opened 2025-03-31
fix(split): add source to battery event

englmaxi:
147c340 introduces a bug where the peripheral battery level is always reported with index 0. This should not be the case if there are more than one peripherals.


caksoylar:
Good catch, this looks like a regression due to #2766.

caksoylar (approved):

petejohanson (approved):
Thanks!

PR #2909, opened 2025-04-04
chore(eslint): upgrade to eslint v9 and flat config

willhack:
# What it does

This is a replacement for the dependabot eslint bump PR  #2896 

## Notes on smaller changes

- the prettier plugin is no longer needed: eslint has move all stylistic rules into a [separate config](https://eslint.style/guide/why) (not sure if it's desired)
- ignores is [now additive](https://eslint.org/docs/latest/use/configure/ignore#name-the-global-ignores-config) so the only thing required is the `build` directory
- file formats can be added directly in the config so now the npm script is simply `eslint` (you _could_ even just run `npx eslint` if you didn't want to go through the boilerplate of `npm run lint`)
- the ecma parser [now defaults](https://eslint.org/docs/latest/use/configure/language-options#specifying-javascript-options) to `latest`, that in combination with the cjs globals allows us to drop the previously explicit parser options

---

I noticed that even though there are `.ts(x)` files there's no linting on them other than the tsc typechecking. Might be useful to add that in the future? For the time being I figured it best to keep this PR pared down.

## PR check-list


docs/package.json
```diff
@@ -48,15 +48,16 @@
...
     "@types/react-helmet": "^6.1.6",
     "@types/react-router-dom": "^5.1.7",
-    "eslint": "^8.39.0",
-    "eslint-config-prettier": "^9.1.0",
-    "eslint-plugin-mdx": "^3.1.5",
-    "eslint-plugin-react": "^7.37.2",
+    "eslint": "^9.23.0",
+    "eslint-plugin-mdx": "^3.3.2",
+    "eslint-plugin-react": "^7.37.5",
+    "globals": "^16.0.0",
```
caksoylar, line 60: Is it needed to add this explicitly? AFAICT this is only used for the eslint config file, and eslint also has it as a dependency.

caksoylar (commented):
Thank you for lending your expertise and tackling this; with our level of knowledge and the _long_ migration guide we probably wouldn't have tackled this ourselves anytime soon.

One question noted in the review, and another issue I noticed. When running `npm run lint` after `npm run build`, the generated file in `.docusaurus` emits an error:

```sh
/home/cem/zmk/docs/.docusaurus/routes.js
  1:8  error  'React' is defined but never used  no-unused-vars
```

I'd probably solve this by adding `.docusaurus` to the `globalIgnores`, but let me know how you'd handle it instead.

docs/package.json
willhack: [Unfortunately yes](https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables), it seems odd that they would now externalize a dependency that most folks are using. Feels very much like ExpressJS externalizing `connect`...but then again it might just be to alleviate [downstream issues](https://github.com/eslint/eslint/issues/19391)

willhack:
> One question noted in the review, and another issue I noticed. When running `npm run lint` after `npm run build`, the generated file in `.docusaurus` emits an error:
> 
> ```shell
> /home/cem/zmk/docs/.docusaurus/routes.js
>   1:8  error  'React' is defined but never used  no-unused-vars
> ```

Total oversight on my part, good catch! We definitely don't want to be linting generated files.

docs/package.json
caksoylar: I think this is due to my ignorance on how node modules work. I thought of it like Python where dependencies are installed and available everywhere, even if they are transitive. But e.g. importing `globals` doesn't work if it is not explicitly listed in `package.json` even though it is depended on by a few other dependencies:

```sh
❯ npm ls globals
docs@0.0.0 /home/cem/zmk/docs
├─┬ @docusaurus/core@3.7.0
│ └─┬ @docusaurus/babel@3.7.0
│   ├─┬ @babel/preset-env@7.26.9
│   │ └─┬ @babel/plugin-transform-classes@7.25.9
│   │   └── globals@11.12.0
│   └─┬ @babel/traverse@7.26.10
│     └── globals@11.12.0
├─┬ @docusaurus/theme-mermaid@3.7.0
│ └─┬ mermaid@11.5.0
│   └─┬ @iconify/utils@2.3.0
│     └── globals@15.15.0
└─┬ eslint@9.23.0
  └─┬ @eslint/eslintrc@3.3.1
    └── globals@14.0.0
```

caksoylar (approved):
Thanks! If you want to tackle adding linting for `.ts(x)` files in the future, it would be a welcome addition.

docs/package.json
willhack: Gotcha, unfortunately (or fortunately depending on your pov) not! There are a few package managers [like pnpm](https://pnpm.io/faq) who have attempted to fix the issue of duplicate dependencies, at least from a space cost, but then we wouldn't have gems like:

PR #2922, opened 2025-04-19
fix(docs): docs home img alt text

romil-soni:
Img alt text on the home page was shown as alt="[object Object]"
since title was a react node and not text

This PR fixes that

## PR check-list

## Screenshots

image

### DIFF
--- a/docs/src/pages/index.js
+++ b/docs/src/pages/index.js
@@ -37,7 +37,7 @@ function Feature({ imageUrl, title, description }) {
     <div className={classnames("col col--4", styles.feature)}>
       {imgUrl && (
         <div className="text--center">
-          <img className={styles.featureImage} src={imgUrl} alt={title} />
+          <img className={styles.featureImage} src={imgUrl} alt="" />
         </div>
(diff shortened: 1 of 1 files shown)


willhack:
Solid call-out in the busted alt rendering, but watch out. This fixes the bug, but surfaces another issue: the titles really don't work as alt text.

> [!TIP]
>  Since they're decorative (or don't provide additional information to the text below), it would probably be best to use an empty alt tag instead to just remove them from the accessibility tree.
> 
> [Helpful alt tag decision tree](https://www.w3.org/WAI/tutorials/images/decision-tree/)



A few options could be to add an optional `alt` prop with fallback to the `Feature` something like:
```Feature({ title, alt = "", ... })```

but since this is only used in this one file it might be just as clean to hardcode the empty alt tag inline.

romil-soni:
Thanks @willhack for your insights have made necessary changes

nmunnich (approved):
Thanks for the fix, and thanks @willhack for the notes!

PR #2923, opened 2025-04-19
chore(eslint): Added typescript support, Fixed all errors

Aasim-A:
This PR is a continuation of #2909 that adds typescript linting support, as well as fix all linting errors.
Was tested with `npm run lint` and `npm run typecheck` with no errors or warnings.

## PR check-list

### DIFF
--- a/docs/.prettierrc.js
+++ b/docs/.prettierrc.js
@@ -1,3 +1,4 @@
 module.exports = {
   endOfLine: "auto",
+  trailingComma: "all",
 };
--- a/docs/.prettierrc.js
+++ b/docs/.prettierrc.js
@@ -1,4 +1,4 @@
 module.exports = {
   endOfLine: "auto",
-  trailingComma: "all",
+  trailingComma: "es5",
 };
--- a/docs/eslint.config.mjs
+++ b/docs/eslint.config.mjs
@@ -6,7 +6,12 @@ import pluginReact from "eslint-plugin-react";
 import * as mdx from "eslint-plugin-mdx";
 
 export default defineConfig([
-  globalIgnores([".cache-loader/", ".docusaurus/", "build/"]),
+  globalIgnores([
+    ".cache-loader/",
+    ".docusaurus/",
+    "build/",
+    "src/hardware-metadata.d.ts",
+  ]),
   { files: ["**/*.{js,jsx,ts,tsx,md,mdx}"] },
   jseslint.configs.recommended,
   tseslint.configs.strict,
--- a/docs/eslint.config.mjs
+++ b/docs/eslint.config.mjs
@@ -9,7 +9,7 @@ export default defineConfig([
   globalIgnores([".cache-loader/", ".docusaurus/", "build/"]),
   { files: ["**/*.{js,jsx,ts,tsx,md,mdx}"] },
   jseslint.configs.recommended,
-  tseslint.configs.recommended,
+  tseslint.configs.strict,
   mdx.flat,
   pluginReact.configs.flat.recommended,
   pluginReact.configs.flat["jsx-runtime"],
--- a/docs/eslint.config.mjs
+++ b/docs/eslint.config.mjs
@@ -1,16 +1,15 @@
 import { defineConfig, globalIgnores } from "eslint/config";
 import globals from "globals";
-import js from "@eslint/js";
+import jseslint from "@eslint/js";
+import tseslint from "typescript-eslint";
 import pluginReact from "eslint-plugin-react";
 import * as mdx from "eslint-plugin-mdx";
 
 export default defineConfig([
   globalIgnores([".cache-loader/", ".docusaurus/", "build/"]),
-  { files: ["**/*.{js,jsx,md,mdx}"] },
-  {
-    plugins: { js },
-    extends: ["js/recommended"],
-  },
+  { files: ["**/*.{js,jsx,ts,tsx,md,mdx}"] },
+  jseslint.configs.recommended,
+  tseslint.configs.recommended,
   mdx.flat,
   pluginReact.configs.flat.recommended,
   pluginReact.configs.flat["jsx-runtime"],
@@ -18,7 +17,10 @@ export default defineConfig([
     languageOptions: {
       globals: { ...globals.browser, ...globals.commonjs, ...globals.node },
     },
-    rules: { "react/no-unescaped-entities": "off" },
+    rules: {
+      "react/no-unescaped-entities": "off",
+      "@typescript-eslint/no-require-imports": "off",
+    },
     settings: {
       react: {
         version: "detect",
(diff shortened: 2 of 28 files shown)


docs/eslint.config.mjs
```diff
@@ -1,24 +1,26 @@
...
 
 export default defineConfig([
   globalIgnores([".cache-loader/", ".docusaurus/", "build/"]),
-  { files: ["**/*.{js,jsx,md,mdx}"] },
-  {
-    plugins: { js },
-    extends: ["js/recommended"],
-  },
+  { files: ["**/*.{js,jsx,ts,tsx,md,mdx}"] },
+  jseslint.configs.recommended,
```
willhack, line 11: huh....this is way cleaner, I wonder why the eslint docs use the more verbose plugins/extends combo. Just tested locally and works as expected.

Do you have any resources besides the tseslint docs where this pattern is used? ...tempted to raise a discussion/pr on the eslint page for their intro and would love something to cite.

docs/eslint.config.mjs
```diff
@@ -1,24 +1,26 @@
 import { defineConfig, globalIgnores } from "eslint/config";
 import globals from "globals";
-import js from "@eslint/js";
+import jseslint from "@eslint/js";
```
willhack, line 3: FWIW this doesn't actually do anything except change the alias (both refer to the same default export).

docs/eslint.config.mjs
```diff
@@ -1,24 +1,26 @@
...
 export default defineConfig([
   globalIgnores([".cache-loader/", ".docusaurus/", "build/"]),
-  { files: ["**/*.{js,jsx,md,mdx}"] },
-  {
-    plugins: { js },
-    extends: ["js/recommended"],
-  },
+  { files: ["**/*.{js,jsx,ts,tsx,md,mdx}"] },
+  jseslint.configs.recommended,
+  tseslint.configs.recommended,
```
willhack, line 12: If you haven't yet, I might suggest trying the `tseslint.configs.strict` since strict mode is turned on anyway in the `tsconfig`.

docs/.prettierrc.js
```diff
@@ -1,3 +1,4 @@
 module.exports = {
   endOfLine: "auto",
+  trailingComma: "all",
```
willhack, line 3: for the sake of this PR I would suggest:
```suggestion
  trailingComma: "es5",
```
that'll reduce the noise from all the extra commas. (better to keep linting and formatting as separate PRs)

docs/src/components/hardware-list.tsx
```diff
@@ -18,11 +18,11 @@ function itemHasMultiple(item: HardwareMetadata) {
...
 function itemIds(item: HardwareMetadata) {
   if (item.type == "board" || item.type == "shield") {
-    let nodes = (item.siblings ?? [item.id])
-      .map<ElementOrString>((id) => <code key={id}>{id}</code>)
+    const nodes = (item.siblings ?? [item.id])
+      .map((id) => <code key={id}>{id}</code>)
       .reduce(
         (prev, curr, index) => [...prev, index > 0 ? ", " : "", curr],
-        [] as ElementOrString[]
+        [] as ElementOrString[],
```
willhack, line 25: This change doesn't seem related to linting updates, might be better in a separate PR?

If anything I would also suggest moving the opposite direction - using generics in reduce as well instead of casting with `as`.

```
  .map<ElementOrString>...
  .reduce<ElementOrString[]>()
```
it'll give the same result but also provide the benefit of type checking on the logic of the reduce function callback

docs/src/hardware-metadata-static-plugin/index.js
```diff
@@ -22,9 +22,9 @@ const METADATA_GLOB = path.posix.join(BASE_DIR, "app/boards/**/*.zmk.yml");
  * @param {string} file
  */
 async function readMetadata(file) {
-  /** @type HardwareMetadata[] */
-  // @ts-ignore
-  const documents = yaml.loadAll(await fs.readFile(file, "utf-8"));
+  const documents = /** @type HardwareMetadata[] */ (
+    yaml.loadAll(await fs.readFile(file, "utf-8"))
+  );
```
willhack, line 27: 🔥

willhack (commented):
Looks good overall!

Some small things here and there

docs/eslint.config.mjs
Aasim-A: That's right. The reason behind the change is to follow the convention used by typescript-eslint. Can revert it if you like.

docs/eslint.config.mjs
Aasim-A: AFAIK, that's how the default config is setup when using eslint init.

willhack: using
```
❯ npm init @eslint/config@latest
✔ What do you want to lint? · javascript
✔ How would you like to use ESLint? · problems
✔ What type of modules does your project use? · esm
✔ Which framework does your project use? · react
✔ Does your project use TypeScript? · no / yes
✔ Where does your code run? · browser
The config that you've selected requires the following dependencies:

eslint, @eslint/js, globals, eslint-plugin-react
✔ Would you like to install them now? · No / Yes
✔ Which package manager do you want to use? · npm
☕️Installing...
```

 ended up generating:


```
import js from "@eslint/js";
import globals from "globals";
import pluginReact from "eslint-plugin-react";
import { defineConfig } from "eslint/config";


export default defineConfig([
  { files: ["**/*.{js,mjs,cjs,jsx}"], plugins: { js }, extends: ["js/recommended"] },
  { files: ["**/*.{js,mjs,cjs,jsx}"], languageOptions: { globals: globals.browser } },
  pluginReact.configs.flat.recommended,
]);
```

If it works it works though, I was just curious why eslint wouldn't write it the way you (and tseslint) did.

docs/eslint.config.mjs
willhack: all good, just figured I'd note it for learning for you and others who might not know

willhack (approved):
Looking good! 🚀

nmunnich (approved):
Thank you to both of you!

petejohanson (approved):
Thanks!

caksoylar (changes requested):
When I run `lint` after a docs build (e.g. via `npm run build`), there is a generated file that causes a warning due to unused ignore:

> /.../zmk/docs/src/hardware-metadata.d.ts
>   1:1  warning  Unused eslint-disable directive (no problems were reported)
> 
> ✖ 1 problem (0 errors, 1 warning)
>   0 errors and 1 warning potentially fixable with the `--fix` option.

This is generated in https://github.com/zmkfirmware/zmk/blob/main/docs/src/hardware-schema-typescript-plugin/index.js during the build. I think the right thing to do is to exclude the file from linting, can you add that to the config?

(There are other solutions, like changing the banner, or configuring eslint to disable the unused directive warning. But excluding the file seems best to me.)

Aasim-A:
Nice catch! it's fixed now.

caksoylar (approved):
Thanks!

PR #2933, opened 2025-04-29
docs: Add secondary functions for the keypad number keys

omulh:
## PR check-list

## Description

This PR is simply to add the information for the alternate or secondary function of the keypad number keys used in the List of Keycodes docs page.  
With the current state of the docs, I was not aware that those keycodes will only serve as numbers when the Numlock is enabled.

### DIFF
--- a/docs/src/data/hid.js
+++ b/docs/src/data/hid.js
@@ -2376,7 +2376,7 @@ export default [
   },
   {
     names: ["KP_NUMBER_1", "KP_N1"],
-    description: "1",
+    description: "1 and End",
     context: "Keypad",
     clarify: false,
     usages: [
@@ -2397,7 +2397,7 @@ export default [
   },
   {
     names: ["KP_NUMBER_2", "KP_N2"],
-    description: "2",
+    description: "2 and ⬇ [Down Arrow]",
     context: "Keypad",
     clarify: false,
     usages: [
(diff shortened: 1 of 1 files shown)


nmunnich (changes requested):
Hi, thanks for the PR. I agree with the concept, this would be good to have. Could you adjust the descriptions to match the descriptions in the [USB HID tables](https://www.usb.org/sites/default/files/documents/hut1_12v2.pdf) (p55-56) exactly? I.e. `Keypad 2 and Down Arrow`, without the arrow symbol (that renders very weirdly on my browser at least).

omulh:
Hi,
I actually just copied the text from the description for the arrow keys in this same file.
We could leave both as they are now, or I could change it as you suggest in both the keypad and normal arrow keys.
Let me know :)

nmunnich:
> Hi, I actually just copied the text from the description for the arrow keys in this same file. We could leave both as they are now, or I could change it as you suggest in both the keypad and normal arrow keys. Let me know :)

I think the arrow keys are fine as-is, but the keypad text gets warped due to the length, so I'd only want that text changed personally.

omulh:
Please check and merge if you agree.

nmunnich (approved):
LGTM, thanks again!

PR #2934, opened 2025-05-05
fix(pointing): Avoids mutex leak for default layer toggle event

cormoran:
Temp layer freezes by mutex leak if layer state change event happens for default layer.

## PR check-list

### DIFF
--- a/app/src/pointing/input_processor_temp_layer.c
+++ b/app/src/pointing/input_processor_temp_layer.c
@@ -134,9 +134,6 @@ static int handle_layer_state_changed(const struct device *dev, const zmk_event_
     if (ret < 0) {
         return ret;
     }
-    if (data->state.toggle_layer == 0) {
-        return ZMK_EV_EVENT_BUBBLE;
-    }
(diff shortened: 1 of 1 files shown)


app/src/pointing/input_processor_temp_layer.c
```diff
@@ -134,9 +134,6 @@ static int handle_layer_state_changed(const struct device *dev, const zmk_event_
     if (ret < 0) {
         return ret;
     }
-    if (data->state.toggle_layer == 0) {
-        return ZMK_EV_EVENT_BUBBLE;
-    }
```
cormoran, line 139: Special handling looks not required, since `zmk_keymap_layer_active` properly handles default layer.

caksoylar: Is the main issue that the mutex is stuck if we hit this branch, since it `return`s without unlocking?

petejohanson: Yeah, this seems reasonable. Especially since we may *change* what is the default/"base" layer later, and don't really want a special check here.

petejohanson (approved):
Good catch, thanks!

PR #2952, opened 2025-06-01
chore(docs): Fix prettier warnings

petejohanson:
Clean up lingering prettier warnings. These are causing failures in our zmk-docker worflows, which check that prettier runs properly. See https://github.com/zmkfirmware/zmk-docker/actions/runs/15376917506/job/43263738558?pr=168

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,8 +12,8 @@ repos:
         types_or: [c++, c]
         args:
           - -i
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.7.1
+  - repo: https://github.com/rbubley/mirrors-prettier
+    rev: 787fb9f542b140ba0b2aced38e6a3e68021647a3
     hooks:
       - id: prettier
         exclude: |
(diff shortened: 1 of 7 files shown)


petejohanson:
Ugh. We appear to have dueling prettier issues. Trying to figure out the "right" way to handle this.

petejohanson:
Ok, got our pre-commit setup actually using the same version of prettier that `npm run prettier:check` does, so we're consistent. Previously, our pre-commit was stuck on an older version.

willhack (approved):
🚀 🔥

.pre-commit-config.yaml
```diff
@@ -12,8 +12,8 @@ repos:
         types_or: [c++, c]
         args:
           - -i
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.7.1
+  - repo: https://github.com/rbubley/mirrors-prettier
+    rev: v3.5.3
```
caksoylar:
```suggestion
    rev: 787fb9f542b140ba0b2aced38e6a3e68021647a3 # v3.5.3
```
How about this? This repo doesn't look very official, so might be prudent to pin to [commit](https://github.com/rbubley/mirrors-prettier/commit/787fb9f542b140ba0b2aced38e6a3e68021647a3).

Issue #1920, opened 2023-09-03
RGB: Setting explicit HSB/RGB doesn't persist new state after 1min debounce


petejohanson:
As noted by @caksoylar in #1918, when explicitly setting the underglow HSB/RGB, we ultiately call `zmk_rgb_underglow_set_hsb` which isn't triggering the delayed state saving like most other functions there.

PR #2955, opened 2025-06-03
Fix rgb underglow hsb not saving in settings.

yiancar:
I have noticed that changing HSB for the rgb underglow subsystem would not save the new value in settings except if an effect change was involved.

This is because the save function is not called in the `zmk_rgb_underglow_set_hsb` function.
Interestingly it is called in all other set functions of the API.

Even more interestingly, https://github.com/zmkfirmware/zmk/blob/239baa487509ace108d36f0e5c627d61a3d95f53/app/src/behaviors/behavior_rgb_underglow.c#L135 in the behavior binding file, even tho each keycode could correspont to one of the other functions of the API that directly maps to their intention; they all get remapped to the `zmk_rgb_underglow_set_hsb`.

In either case, `zmk_rgb_underglow_set_hsb` should call the save function.

## PR check-list

### DIFF
--- a/app/src/rgb_underglow.c
+++ b/app/src/rgb_underglow.c
@@ -381,7 +381,7 @@ int zmk_rgb_underglow_set_hsb(struct zmk_led_hsb color) {
 
     state.color = color;
 
-    return 0;
+    return zmk_rgb_underglow_save_state();
 }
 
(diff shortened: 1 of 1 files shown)


caksoylar:
Related: #1920

PR #2968, opened 2025-06-16
feat(behaviors): add universal snap tap behavior for gaming SOCD

sqdbruh:
Implements SOCD (Simultaneous Opposing Cardinal Directions) functionality that provides competitive gaming advantages through "last key wins" behavior for opposing directional keys.

Features:
- Universal syntax: &st <key> <opposing_key> works with any key pair
- Dynamic state management supports up to 16 key pairs simultaneously
- Memory efficient: only stores state for active pairs
- Gaming optimized: perfect for FPS strafing, platformer movement, MOBA control
- Zero performance impact when behavior is not used

Usage example:
&st A D    // A key with D as opposing key
&st W S    // W key with S as opposing key

The behavior solves the problem where pressing opposing keys (like A+D) causes character to stop moving by automatically releasing the previous key when its opposing key is pressed, enabling instant direction changes without dead zones


lesshonor:
related: #2391

caksoylar:
This seems a bit specific to me for a built-in behavior in ZMK. See modules like 
https://github.com/nguyendown/zmk-behavior-socd for an out-of-tree implementation.

peterjc:
How would this look for multiple SOCD pairs? eg these six pairs:
* A and D,
* W and S, 
* left-cursor and right-cursor,
* up-cursor and down-cursor,
* page-up and page-down,
* home and end?

(My understanding of the docs for the module https://github.com/nguyendown/zmk-behavior-socd is that would need a behavior for each pair and then use those six different behaviours in the keymap for those 12 keys - workable but verbose - and this could not be used for say *all* letters being mutually exclusive.)

PR #2975, opened 2025-06-20
ci: Support nested `config_path` inputs

robbowland:
# What

The current `config_path` `input` option is a bit of a misnomer - it's better described as `config_dir_name` as providing a path that is more than a single level deep will result in the [following error](https://github.com/robbowland/keymaps/actions/runs/15767071715/job/44445463861#step:9:13) when reaching the `west update` stage:

```sh
FATAL ERROR: no west workspace found from "/__w/keymaps/keymaps"; "west update" requires one.
Things to try:
  - Change directory to somewhere inside a west workspace and retry.
  - Set ZEPHYR_BASE to a zephyr repository path in a west workspace.
  - Run "west init" to set up a workspace here.
  - Run "west init -h" for additional information.
```

This occurs because `west init -l <dir>` creates a `.west` folder in the parent directory of `<dir>` marking it as the west `topdir`.

In subsequent `west` commands, `west` will search **up** the directory tree (I think) for this `.west` folder.
If it doesn't find one, it'll throw the above error.

For level `config_path` inputs this is fine, as the parent directory is the same as the base directory the workflow is executing it. However, if we provide a multilevel path e.g. `crkbd/config` the command will fail as the `west` workspace and the base directory are no longer equivalent.

This PR fixes this by propagating the `west` workspace directory correctly to subsequent commands. 
[Example workflow](https://github.com/robbowland/keymaps/actions/runs/15782596683).

# Why

I use a nested [monorepo](https://github.com/robbowland/keymaps) for my keyboard configurations and want the option to use the workflow for multiple different configurations without:
- sacrificing my directory structure
- having to have a separate repo per board

This is kind of a 'power user' feature so to speak so I haven't written any docs for it.
I couldn't find any docs for non-standard workflow set-ups anyway.


.github/workflows/build-user-config.yml
```diff
@@ -71,15 +71,24 @@ jobs:
             export zmk_load_arg=" -DZMK_EXTRA_MODULES='${GITHUB_WORKSPACE}'"
```
robbowland, line 68: I suspect I may need to also change the `zephyr/module.yml` check to check the config parent directory.

PR #2983, opened 2025-06-24
feat(display): Always display current BLE profile

AndrewMSHowe:
Display USB or BLE active transport symbol (USB if BLE is disabled)

To the right, If BLE is enabled, always show the current BLE profile index and its connection status (not bonded / disconnected / connected).

The previous behavior was to show the USB symbol only, if that was the active transport. It's useful to see the active BLE profile index and status even if it is not the active transport. It is what you will connect to if you toggle USB to BLE. It is possible to browse for a free profile without unplugging USB.

## PR check-list


AndrewMSHowe:
Something was wrong with the checks, and also I noticed I pushed this from my main branch and it doesn't seem like there's a way to fix it. I am so bad at all this git stuff, there are so many ways to screw it up :( I'l be back!

PR #2986, opened 2025-06-27
feat(backlight): Add backlight brightness curves

AndrewMSHowe:
Add brightness correction curves to the backlight module.

The perceived brightness of the backlight LEDs is nonlinear. With a fixed step size, BL_INC/DEC give a linear change in brightness, which is too steep at low brightness and too shallow at high brightness.

This commit corrects the linear brightness values using a configurable curve. This allows for a more natural adjustment of the backlight brightness level.

The feature can be enabled by choosing one of the exponent options: ZMK_BACKLIGHT_BRIGHTNESS_CURVE_GAMMA_TWO
ZMK_BACKLIGHT_BRIGHTNESS_CURVE_GAMMA_TWO_POINT_TWO ZMK_BACKLIGHT_BRIGHTNESS_CURVE_GAMMA_THREE
ZMK_BACKLIGHT_BRIGHTNESS_CURVE_GAMMA_CUSTOM
ZMK_BACKLIGHT_BRIGHTNESS_CURVE_SRGB
ZMK_BACKLIGHT_BRIGHTNESS_CURVE_CIE_L

The _CUSTOM option allows for a custom exponent value to be set via ZMK_BACKLIGHT_BRIGHTNESS_CURVE_GAMMA_VALUE

The _TWO and _THREE options use integer math only.

The SRGB option uses the sRGB electro-optical transfer function, which is linear below 4.045% brightness and gamma 2.4 above. The CIE L option uses the (inverse) CIE L* transfer function, which is linear below 8% brightness and gamma 3.0 above.

Future work:
- Values near zero are still difficult
  - Bit depth is limited; ideally we could have more precision at low brightness levels.
  - On the author's hardware, brightness levels 1..2 are not visible
    - It might be good to have a configurable minimum brightness level that's not completely off, and have the linear parts of the functions start there rather than at 0.
  - There may be electronic reasons for extra nonlinearity at low brightness levels, in addition to the perceptual reasons.
- RGB underglow could use a similar correction curve.
  - Ths author's hardware does not have RGB underglow
  - Configuring that would probably need a separate Kconfig section, but some of the code could be shared. However it couldn't be shared as it is, because it's full of the Kconfig options for backlight.

## PR check-list


AndrewMSHowe:
lol no-one bothered to look at it, what a waste of my time

PR #2993, opened 2025-07-10
refactor(ble): add functions to check if profile is open/connected by address

snoyer:
This PR implements `zmk_ble_active_profile_is_open()` and `zmk_ble_active_profile_is_connected()` in terms of newly introduced `zmk_ble_profile_is_open(addr)` and `zmk_ble_profile_is_connected(addr)`

## PR check-list

### DIFF
--- a/app/include/zmk/ble.h
+++ b/app/include/zmk/ble.h
@@ -33,6 +33,9 @@ bt_addr_le_t *zmk_ble_profile_address(uint8_t index);
 bt_addr_le_t *zmk_ble_active_profile_addr(void);
 struct bt_conn *zmk_ble_active_profile_conn(void);
 
+bool zmk_ble_profile_is_connected(const bt_addr_le_t *addr);
+bool zmk_ble_profile_is_open(const bt_addr_le_t *addr);
+
 bool zmk_ble_active_profile_is_open(void);
 bool zmk_ble_active_profile_is_connected(void);
 char *zmk_ble_active_profile_name(void);
--- a/app/include/zmk/ble.h
+++ b/app/include/zmk/ble.h
@@ -33,8 +33,8 @@ bt_addr_le_t *zmk_ble_profile_address(uint8_t index);
 bt_addr_le_t *zmk_ble_active_profile_addr(void);
 struct bt_conn *zmk_ble_active_profile_conn(void);
 
-bool zmk_ble_profile_is_connected(const bt_addr_le_t *addr);
-bool zmk_ble_profile_is_open(const bt_addr_le_t *addr);
+bool zmk_ble_profile_is_connected(uint8_t index);
+bool zmk_ble_profile_is_open(uint8_t index);
 
 bool zmk_ble_active_profile_is_open(void);
 bool zmk_ble_active_profile_is_connected(void);
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -98,6 +98,9 @@ K_WORK_DEFINE(raise_profile_changed_event_work, raise_profile_changed_event_call
 bool zmk_ble_active_profile_is_open(void) { return zmk_ble_profile_is_open(active_profile); }
 
 bool zmk_ble_profile_is_open(uint8_t index) {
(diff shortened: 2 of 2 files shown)


app/include/zmk/ble.h
```diff
@@ -33,6 +33,9 @@ bt_addr_le_t *zmk_ble_profile_address(uint8_t index);
 bt_addr_le_t *zmk_ble_active_profile_addr(void);
 struct bt_conn *zmk_ble_active_profile_conn(void);
 
+bool zmk_ble_profile_is_connected(const bt_addr_le_t *addr);
+bool zmk_ble_profile_is_open(const bt_addr_le_t *addr);
```
petejohanson: I'm definitely confused, I would have expected these to accept a profile index as a parameter, not an address. This feels off to me.

snoyer: The existing `bool zmk_ble_active_profile_is_xxx(void)` functions used `&profiles[active_profile].peer` internally so it felt like the most straight forward (and least invasive) refactor to allow checking arbitrary profiles was to just extract that as a parameter.

The indirection from index to address can be done by the caller with `bt_addr_le_t *zmk_ble_profile_address(uint8_t index)` from #2992

But yes, these 2 new functions could take the profile index instead of the address:
- pros: simpler
- cons: can't explicitly handle the invalid index case? `zmk_ble_profile_address` returns `BT_ADDR_LE_NONE` but here we'd just silently return `false`?

petejohanson: The correct way to handle this would be to return `int` from the function, with:

`0`: not connected/not open
`1`: connected/open
negative value: errno

snoyer: I see, but then we have different return types for `active_profile_is_xxx()` and `profile_is_xxx(index)` :thinking:

So, choices are:
1. PR as is is, callers have to do the `index -> address` indirection themselves and should error check it
2. use index instead of address and just return `false` on bad index. This make sense too: an non-existing profile is not connected and it's not open either
3. use index instead of address and return `int` with negative errnos

Personally I'd go with 2 for simplicity and consistency (and I think had I done that to begin with we wouldn't be having this conversation), 3 is more proper in theory but IMO not worth the added complexity.

But it's not my call to make in the end so let me know whether you want to go with option 2 or 3

petejohanson: I'm good with #2.

snoyer: done

app/src/ble.c
```diff
@@ -115,9 +117,13 @@ void set_profile_address(uint8_t index, const bt_addr_le_t *addr) {
 }
 
 bool zmk_ble_active_profile_is_connected(void) {
+    return zmk_ble_profile_is_connected(active_profile);
+}
+
+bool zmk_ble_profile_is_connected(uint8_t index) {
```
petejohanson, line 126: Need to check the `index` is in the valid range, and return `false` otherwise.

app/src/ble.c
```diff
@@ -95,8 +95,10 @@ static void raise_profile_changed_event_callback(struct k_work *work) {
 
 K_WORK_DEFINE(raise_profile_changed_event_work, raise_profile_changed_event_callback);
 
-bool zmk_ble_active_profile_is_open(void) {
-    return !bt_addr_le_cmp(&profiles[active_profile].peer, BT_ADDR_LE_ANY);
+bool zmk_ble_active_profile_is_open(void) { return zmk_ble_profile_is_open(active_profile); }
+
+bool zmk_ble_profile_is_open(uint8_t index) {
+    return !bt_addr_le_cmp(&profiles[index].peer, BT_ADDR_LE_ANY);
```
petejohanson, line 104: Needs a sanity check on the `index` value to avoid index out of bound error.

app/src/ble.c
snoyer: Totally forgot to do the one thing I said needed doing :facepalm: sorry!

petejohanson (approved):

PR #2995, opened 2025-07-12
Disable high voltage DC-DC by default

### LINKED ISSUE #2990: nice!nano v2 high power consumption after enabling `SOC_DCDC_NRF52X_HV` became default
commit f0b20c1c938da6dfd51f4639a2a71b587cab4205: feat(boards): Add nRF52 high voltage DC/DC config

```
feat(boards): Add nRF52 high voltage DC/DC config
Added a Kconfig option to enable SOC_DCDC_NRF52X_HV for nice_nano_v2
and mikoto. According to Nordic's documentation, the DC/DC regulator is
more efficient than the LDO regulator, so this is enabled by default.
```

Is this option actually tested?
I tested `SOC_DCDC_NRF52X_HV` on/off on both a genuine official nice!nano v2 and a nice!nano v2 aliexpress clone (dubbed `NRF52840 Development Board Compatible With Nice! Nano V2.0`).

The test is performed on a pair of corne split keyboard (no light, wireless split), with a multimeter in series with the battery connection.
Tested battery current draw on both standby (not in sleep mode, but not typing anything either. with active bluetooth connection to 1 computer), and pressed state(pressing a key without letting go continuously). Battery is at 4.2V, both board is tested on the same pair of battery, back to back.

The result showed a noticeable current draw increase with the option ON compared to OFF on the genuine nice!nano v2, with even more drastically increased current draw on the aliexpress nice!nano v2 clone board.

## Genuine nice!nano v2
shown an increase in current draw across the broad
|                | OFF    | ON     | diff    | diff% |
|----------------|--------|--------|---------|-------|
| Left(standby)  | 0.48mA | 0.60mA | +0.12mA | +25%  |
| Right(standby) | 0.05mA | 0.06mA | +0.01mA | +20%  |
| Left(pressed)  | 1.10mA | 1.58mA | +0.48mA | +44%  |
| Right(pressed) | 0.79mA | 1.22mA | +0.43mA | +54%  |

## aliexpress nice!nano v2 clone board
way worse. basically tanked the battery life:
|                | OFF    | ON     | diff    | diff%  |
|----------------|--------|--------|---------|--------|
| Left(standby)  | 0.60mA | 1.90mA | +1.30mA | +216%  |
| Right(standby) | 0.08mA | 1.40mA | +1.32mA | +1650% |
| Left(pressed)  | 1.37mA | 2.60mA | +2.23mA | +607%  |
| Right(pressed) | 1.01mA | 2.10mA | +1.09mA | +107%  |

Considering it's effect has been negative on both the genuine and the clone board(added here because the number of people who might've bought the clone board thinking it's the same as the genuine one might be significant), at least on my units. Maybe it's worthwhile to do more tests before enabling it by default?

----------------

For anyone who stumbles upon this issue trying to find a solution, you can add this option to your board's `.conf` file, compile and reflash it, and measure with a multimeter to see if it's the problem:
```
CONFIG_BOARD_ENABLE_DCDC_HV=n
```

> edit(20250721): fix typo, used to say battery is 8.2V which is incorrect. it's 4.2V

joelspadin:
Given a typical setup of a battery at ~4 V and a nice!nano v2 running at 3.3 V, I measured a small increase in power usage (using a Nordic PPK2) when SOC_DCDC_NRF52X_HV was enabled. This changes the nice_nano_v2 and mikoto boards so this option is disabled by default.

Fixes #2990

### DIFF
--- a/app/boards/arm/mikoto/Kconfig
+++ b/app/boards/arm/mikoto/Kconfig
@@ -7,7 +7,7 @@ config BOARD_ENABLE_DCDC
 config BOARD_ENABLE_DCDC_HV
     bool "High voltage DCDC converter"
     select SOC_DCDC_NRF52X_HV
-    default y
+    default n
     depends on (BOARD_MIKOTO)
 
 choice BOARD_MIKOTO_CHARGER_CURRENT
--- a/app/boards/arm/nice_nano/Kconfig
+++ b/app/boards/arm/nice_nano/Kconfig
@@ -9,5 +9,5 @@ config BOARD_ENABLE_DCDC
 config BOARD_ENABLE_DCDC_HV
     bool "High voltage DCDC converter"
     select SOC_DCDC_NRF52X_HV
-    default y
+    default n
     depends on (BOARD_NICE_NANO_V2)


nmunnich (approved):

nmunnich:
Considering that the only tests where DCDC actually won out were with a custom board, I think this is justified.

PR #3011, opened 2025-07-30
feat(boards): add support for nrf52840dongle_nrf52840

dapetcu21:
This adds additional required configs for the Nordic NRF52840 Dongle (PCA10059), which was already supported by Zephyr, but didn't have flash settings and Bluetooth enabled.

## PR check-list


app/boards/nrf52840dongle_nrf52840.conf
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2022 The ZMK Contributors
```
caksoylar:
```suggestion
# Copyright (c) 2025 The ZMK Contributors
```

dapetcu21: Oh yeah. Oops

app/boards/arm/nrf52840dongle_nrf52840/nrf52840dongle_nrf52840.zmk.yml
```diff
@@ -0,0 +1,10 @@
+file_format: "1"
+id: nrf52840dongle_nrf52840
+name: Nordic nRF52840 Dongle
+type: board
+arch: arm
+outputs:
+  - usb
+  - ble
+url: https://www.nordicsemi.com/Products/Development-hardware/nrf52840-dongle
+exposes: []
```
nmunnich:
```suggestion
```

dapetcu21:
Fixed and rebased :)

dapetcu21:
Just realised: Since this doesn't `expose` any standard interconnect (it's a usb dongle with a non-standard gpio pinout as far as I know), nor does it have a `keys` feature (it's not an on-board controller keyboard), it won't show up in Supported Hardware in the documentation. How do you propose this is handled?

nmunnich:
I think it would be better for it not to show up in the hardware list anyway, as it will most likely only be used by people actively searching for it. Putting it in the hardware list will only generate confusion imo.

However, it seems possible that our build scripts are unhappy with the situation too. See if you can resolve that, please.

dapetcu21:
Fixed. I used the `settings_reset` shield to test boards of this sort. Tested the CI workflow and it seems to work on my own fork.

nmunnich:
Hmmm. Sorry, I think I was misleading in my comment - I assumed there was something in the board that could be adjusted to get it to pass, but it seems that's not the case.

For things like CI changes, that would need to be in a different PR, but I don't really think the solution presented is the right one. For something like this, I think we would want to adjust our metadata schema to account for dongles, and provide test cases to dongles instead. I think that's beyond the scope of this ticket and would need some further discussion and thought though, and I think the focus currently lies elsewhere.

For now, I think if you reduce this PR down to only the .conf file, similar to how the rpi_pico exists, then the tests should pass and I would be ok with it - though I'd want @caksoylar to also give input on whether he thinks that is an acceptable solution too.

dapetcu21:
Hi! Sure, I can do that for now. I am also willing to volunteer the implementation, once there is a plan for dongle boards. (I agree that it's not the best solution, just seemed the simplest without a major change).

Unfortunately I don't know enough about the goals of the metadata file to suggest a solution. Technically, any board that supports BLE can be a dongle. At the extreme, there could even be a "dongle" with UART transport if you want to separate your keyboard like that. What I'm saying is that the line between a central in a split configuration and a separate dongle is very blurry. Maybe a better question is how to handle boards with non-standard interconnects?

dapetcu21:
(old implementation left at https://github.com/dapetcu21/zmk/tree/nrf52840dongle_nrf52840-old)

nmunnich (approved):

PR #3015, opened 2025-08-01
docs: correct a misleading _defconfig description

efogdev:
## PR check-list

### DIFF
--- a/docs/docs/config/index.md
+++ b/docs/docs/config/index.md
@@ -78,7 +78,7 @@ CONFIG_EC11=y
 CONFIG_EC11_TRIGGER_GLOBAL_THREAD=y
 ```
 
-The list of available settings is determined by various files in ZMK whose names start with `Kconfig`. Files ending with `_defconfig` use the same syntax, but are intended for setting configuration specific to the hardware which users typically won't need to change. Note that options are _not_ prefixed with `CONFIG_` in these files.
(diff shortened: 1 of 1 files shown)


docs/docs/config/index.md
```diff
@@ -78,7 +78,7 @@ CONFIG_EC11=y
 CONFIG_EC11_TRIGGER_GLOBAL_THREAD=y
 ```
 
-The list of available settings is determined by various files in ZMK whose names start with `Kconfig`. Files ending with `_defconfig` use the same syntax, but are intended for setting configuration specific to the hardware which users typically won't need to change. Note that options are _not_ prefixed with `CONFIG_` in these files.
+The list of available settings is determined by various files in ZMK whose names start with `Kconfig`. Files ending with `_defconfig` use the same syntax, but are intended for setting configuration specific to the hardware which users typically won't need to change.
```
joelspadin: I think someone may have inserted the `_defconfig` line into the middle of what used to be a valid description of the `Kconfig` files. The first and third sentences make sense together, but the second one is related to the `.conf` stuff above it.

We can probably also do a better job of describing the difference between `_defconfig` and `.conf` here too. Maybe something like this?

```suggestion
Files ending with `_defconfig` use the same syntax as `.conf` files. They set the default configuration for the hardware, which is then overridden by anything in a `.conf` file.

The list of available settings is determined by various files in ZMK whose names start with `Kconfig`. Note that options are _not_ prefixed with `CONFIG_` in these files.
```

efogdev: I'll step away and stand in the corner :) Not familiar with ZMK at all, just noticed this misdirection.

PR #3016, opened 2025-08-01
feat(ci): Automatically bump version in user config template on release-please

nmunnich:
Refactored the existing release-please workflow into multiple jobs. Then added a new job to automatically bump the version of https://github.com/zmkfirmware/unified-zmk-config-template. 

The new job was tested https://github.com/nmunnich/actions_test_repo. I did not test the complete workflow, but it should be equivalent to the existing one. 

This requires the secret token ZMK_RELEASE_PLEASE_TOKEN to be given access to the unified zmk config template repo with write permissions for contents *and* workflows. I did consider using a second token for the unified zmk config template repo, but I decided against it. Let me know if you think of a reason for a different token/wish to use a second token, and I can adjust accordingly.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -11,28 +11,70 @@ permissions:
 name: release-please
 
 jobs:
-  release-please:
+  handle-commit:
+    name: Handle new commit
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


petejohanson (approved):
Thanks!

petejohanson:
Added, but changed the scope to `chore` since CI change like that shouldn't bump the ZMK version as much.

PR #3029, opened 2025-08-07
fix typo for pip requirements extras file

maxchaston:
When trying to setup the a environment, trying to pip install the extra requirements did not work as the file was slightly misnamed.


nmunnich (approved):
Thanks!

PR #2873, opened 2025-03-13
feat(ble): Add event `zmk_ble_advertising_status_changed`

adolto:
Add event `zmk_ble_advertising_status_changed` to notify about advertising status changes.
New advertising status will be send within the event data.

Closes: #2872

## PR check-list

- [x] Branch has a [clean commit history](https://zmk.dev/docs/development/contributing/pull-requests#clean-commit-history)
- [ ] Additional tests are included, if changing behaviors/core code that is testable.
- [x] Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)
- [x] [Pre-commit](https://zmk.dev/docs/development/local-toolchain/pre-commit) used to check formatting of files, commit messages, etc.
- [ ] Includes any necessary [documentation changes](https://zmk.dev/docs/development/contributing/documentation).

Issue #2872, opened 2025-03-13
Feature request: Add ZMK event to get notified about BLE advertising state changes


adolto:
I'm missing a ZMK event to get notified about BLE advertising state changes. You will only be notified when the active profile changes through the `zmk_ble_active_profile_changed` event

adolto:
Still relevant feature request + pull request.

I'm developing a behaviour displaying the selected BT profile with unique LEDs for each profile. A not connected profile should be indicated with a blinking led. However, at the moment there is no way to recognize a connection status change of a profile except for polling `zmk_ble_active_profile_is_connected` cyclical.

PR #3032, opened 2025-08-11
feat(ble): Add event for non-active BLE profile changes

snoyer:
Currently `zmk_ble_active_profile_changed` events are raised when the state of the active profile changes but we don't get notified if the state of any other profiles changes.
This means that, for example, the nice!view display of individual profile states (#2265) will not update if a device that is not on the active profile disconnects on its own.

This PR adds a `zmk_ble_other_profile_changed` (similar to the existing `zmk_ble_active_profile_changed`) that gets raised from the `connected()` and `disconnected()` methods when the connection is not the active one.

## PR check-list


caksoylar:
Does this also address #2872 similar to #2873?

snoyer:
> Does this also address #2872 similar to #2873?

According to the implementation of #2873, no, this one doesn't expose or react to the advertising state.
But according to the description of #2872, I'd say yes, it does let you know about profiles other than the active one. Would need @adolto to confirm whether they actually need the advertising details.

petejohanson (changes requested):
I don't love the "other" naming... I think perhaps "inactive" would be a better adjective for these?

snoyer:
> > Does this also address #2872 similar to #2873?
> 
> According to the implementation of #2873, no, this one doesn't expose or react to the advertising state. But according to the description of #2872, I'd say yes, it does let you know about profiles other than the active one. Would need @adolto to confirm whether they actually need the advertising details.

I have done some more testing with #2873 and found that when a device connected to an inactive profile disconnects we get `<dbg>    zmk:    update_advertising: advertising from 2 to 2` and no `zmk_ble_advertising_status_changed` is fired because the advertising state value hasn't actually _changed_.

So, revised answer to the best of my understanding:
- if we want to listen to changes of advertising state explicitly, as per the title #2872: we need #2873 and this PR is unrelated
- if we want to listen to profile/connection changes, as per the comments of #2872: we want this PR (or drop the `if (previous_adv != advertising_status)` in #2873 which could make it noisy?)

PR #3037, opened 2025-08-17
feat(split): Allow wired split for Ergodash

seanlano:
The Ergodash shield uses a Pro Micro compatible TRRS wired split configuration. This small change enables this with the existing Ergodash shield definition.

## PR check-list

- [ ] ~~Additional tests are included, if changing behaviors/core code that is testable.~~
- [ ] ~~Includes any necessary [documentation changes](https://zmk.dev/docs/development/contributing/documentation).~~

### DIFF
--- a/app/boards/shields/ergodash/ergodash.dtsi
+++ b/app/boards/shields/ergodash/ergodash.dtsi
@@ -12,6 +12,11 @@
         zmk,matrix-transform = &default_transform;
     };
 
+    wired_split {
+        compatible = "zmk,wired-split";
+        device = <&pro_micro_serial>;
+    };
+
     default_transform: keymap_transform_0 {
         compatible = "zmk,matrix-transform";
         columns = <14>;


seanlano:
I've very new to Zephyr and ZMK (although I'm otherwise an experience firmware developer), so I'm not 100% sure that this is the right way to do this – suggestions gladly welcomed!

I had tried putting the block:
```
    wired_split {
        compatible = "zmk,wired-split";
        device = <&pro_micro_serial>;
    };
```
into a local `zmk-config` keymap file, but that didn't work (you can see my attempts [here](https://github.com/seanlano/seanlano-ergodash-zmk))

I found that it only worked when in the `.dtsi` file, so I figured I'd create a PR for it.

seanlano:
Oh and I have a KB2040 driver, which seems to not enable serial by default - so I also added this to my keymap file:
```
&pro_micro_serial {
    status = "okay";
};
```

nmunnich:
I think we'll need to think a bit about how to best begin supporting wired split keyboards in-tree -- this would be the first such keyboard. I'm not sure our setup scripts/schema/stuff are configured to allow everything to work smoothly. That said, thanks for the PR, it'll help keep an eye on the subject.

Regarding your local attempts, the reason it didn't work in the keymap is because you didn't enable 
```
&pro_micro_serial {
    status = "okay";
};
```
in the keymap as well at the time -- you only did so after copying over and modifying the shield definition.

seanlano:
Thanks @nmunnich , yeah I've been playing some more and do have it working now just with a local keymap and no changes to the Ergodash shield.

I'll keep an eye out for other changes and help test where I can, but yeah this seems to work for me without needing the PR.

_(If anyone happens to find this in the future, you can see the config I changed for a KB2040 and Ergodash [in this commit](https://github.com/seanlano/seanlano-ergodash-zmk/commit/f1669e9b5bcddc14b5485144cd8a1f46ea76ad51))_

PR #3060, opened 2025-09-04
Move to zephyr v4.1

petejohanson:
Opening this as a draft for now, as there is still work to be done here, along extensive testing, but at this point I want a place where progress, testing results, etc can be tracked.

This is considered beta level at this point, so basic support questions will not be answered on this PR/branch. I welcome testing by anyone willing to, given this level of readiness. TIA.

# Prelim Blog Post

@nmunnich prepared an update post, which I am working to round out and finalize: https://deploy-preview-3060--zmk.netlify.app/blog/2025-xx-xx-zephyr-4-1

# Testing steps

## Local dev

1. If not building w/ VSCode docker integration, install https://github.com/zephyrproject-rtos/sdk-ng/releases/tag/v0.16.9
1. `gh pr checkout 3060`
2. `west update`
3. Use new build directory to build ZMK as usual.

## GitHub Actions

Make two edits in the branch of your config repo:

1. ~~Update `.github/workflowr/build.yaml` to contain:~~

    ```
    on: [push, pull_request, workflow_dispatch]

    jobs:
      build:
        uses: petejohanson/zmk/.github/workflows/build-user-config.yml@core/move-to-zephyr-4-1
    ```
 1. Update `config/west.yaml` according to the https://zmk.dev/docs/features/beta-testing/ instructions to test this PR.

## Outstanding Items

- [x] Bring in blog post from https://github.com/petejohanson/zmk/pull/48
- [x] Finish updating existing (nRF52 & RP2040) boards to support newer bootloader approach with retained mem + bootmode APIs from Zephyr
- [x] Ensure blog post covers HWMv2 upgrades, bootloader changes, nRF52840 DCDC and NFC pin DTS changes, LVGL API changes, LED strip changes, input support for keyboard matrixes (testing only), and anything else discovered during testing
- [x] Add note to blog post about RP2040 board upgrades: Need to enable CONFIG_CLOCK_CONTROL, adjust DTS include path, boot mode retention.
- [x] Fix up display screens, including nice!view custom status screen + widgets
- [x] Integrate split BLE API change code and test thoroughly.
- [x] Test, test, more testing.
- [x] Document CMake v3 requirements for some edge cases (nRF5340 where `libmetal` gets used)
- [x] Investigate fault on connecting with studio over BLE (reported with Win11 host): https://discord.com/channels/719497620560543766/1433362939217444874/1433632744235339867
- [x] Move away from deprecated BT DIS Kconfigs
- [x] restore cmake discovery of `<module>/boards/<board_id>.overlay` files.

## PR check-list


.github/workflows/build-user-config.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           echo "extra_west_args=${extra_west_args}" >> $GITHUB_ENV
           echo "extra_cmake_args=${shield:+-DSHIELD=\"$shield\"}${zmk_load_arg}" >> $GITHUB_ENV
           echo "display_name=${shield:+$shield - }${board}" >> $GITHUB_ENV
-          echo "artifact_name=${artifact_name:-${shield:+$shield-}${board}-zmk}" >> $GITHUB_ENV
+          echo "artifact_name=${artifact_name:-${shield:+$shield-}${board//\//_}}-zmk}" >> $GITHUB_ENV
```
lesshonor: [`dash` doesn't implement this syntax for parameter expansion](https://github.com/lesshonor/zmk-config/actions/runs/17480137983/job/49648828340#step:5:35), so either the `shell` invocation for this step needs to be switched from [`sh -x` to `bash -x`](https://github.com/zmkfirmware/zmk/pull/3060/files#diff-f091e8bd59ea9020ccd60bff28b4314b79a9d92de351843ddbbaf68a459f46ddR60) or the substitution performed by a utility like `tr` (...or `sed`)

```suggestion
          echo "artifact_name=${artifact_name:-${shield:+$shield-}${board//\//_}-zmk}" >> $GITHUB_ENV
```

app/boards/polarityworks/bt60/bt60.dts
```diff
@@ -1,53 +1,23 @@
...
-        compatible = "gpio-leds";
-        blue_led: led_0 {
-            gpios = <&gpio0 15 GPIO_ACTIVE_HIGH>;
-        };
-    };
-
     vbatt: vbatt {
         compatible = "zmk,battery-voltage-divider";
         io-channels = <&adc 2>;
-        output-ohms = <2000000>;
```
joelspadin, line 49: Was deleting this intentional? I noticed this while editing all the boards using zmk,battery-voltage-divider for one of my two approaches for implementing fuel gauge API support.

petejohanson: Yes, because this board is moved over to using multiple versions, so the specific settings on that come from the version specific overlays. The prop is still set, it's just coming from somewhere else.

app/boards/joric/nrfmicro/nrfmicro_nrf52840_1_1_0_defconfig
```diff
@@ -1,6 +1,5 @@
 # SPDX-License-Identifier: MIT
 
-CONFIG_SOC_SERIES_NRF52X=y
 CONFIG_SOC_NRF52840_QIAA=y
 CONFIG_BOARD_NRFMICRO_11=y
```
nmunnich:
```suggestion
```

app/boards/xiao_ble.overlay
nmunnich, line 1: xiao_rp2040 also needs its overrides renamed.

petejohanson: Thanks. Fixed.

galaxyeden:
Great to see this making progress, thanks for all your hard work on ZMK! I wanted to move to `zephyr,input` to get my own board ready for Zephyr 4.2 and I ran into a few issues and notes:

- There is a maximum of eight rows by default, users may need to select `INPUT_KBD_MATRIX_16_BIT_ROW` – I did
- The Zephyr documentation for `gpio-kbd-matrix` assumes the matrix is active low. In my previous testing, I found this provided a small but noticeable improvement over the active high ZMK-documented configuration, so my keyboard was already set up this way, but it may confuse some users looking at the Zephyr documentation and not understanding the difference, especially newbies
- `gpio-kbd-matrix` only supports scanning "columns" and listening on "rows" – this may also confuse some, especially if this doesn't match their physical layout
- Initially, I could only get five rows to work, @petejohanson suggested it was likely running out of hardware interrupts set on the falling (or rising, for active high) edge of the pin, as `gpio-kbd-matrix` sets. I'm on an nRF5340 with 8 interrupts available, which is comparable to most common ZMK users' hardware – so anyone with more than (hardware interrupts - defined buttons - other defined interrupt pins) is likely to encounter this. This can be fixed by emulating this behavior in software with port-based interrupts, more below.
- In `physical_layouts.c` there is a use of the `ZMK_KSCAN_EVENT_QUEUE_SIZE` that applies to the `zephyr,input` scan driver as well. I renamed this Kconfig symbol (for aesthetics) and removed the `if ZMK_KSCAN` conditional in my fork to resolve this, and it works as expected.

To emulate the falling edge interrupts in software, each GPIO port needs a `sense-edge-mask` defined. I set this to `< 0xffffffff >` for both ports and that resolved the issue and my nRF5340-based board is working as expected with `zephyr,input`.

On the very good news side, I've seen dramatically better power consumption – about 25% lower at idle than before total (≈240uA to ≈180uA), about 40% lower with keys held forcing scanning (at the same 1ms scan interval, my initial, much better results were due to the 5ms default scan interval), and similar – hard to tell but slightly higher if anything – during my "average" typing – with the amount of time spent idle for most users, I expect this to be a significant, though variable, overall power savings if it holds true across multiple SoCs and matrix configurations.

docs/docs/development/local-toolchain/setup/native.mdx
caksoylar, line 1: I think this change might let us merge some of the separate tab contents, need to look into it later. Does this get the Studio dependencies OK? I think Joel had mentioned issues with that on 4.1.

petejohanson: Tested with a fresh new virtual environment and it installed deps and built a Studio firmware fine.

lesshonor:
Since the original suggestion has been marked as resolved, I'll reiterate it: [`build-user-config.yml` is currently broken](https://github.com/lesshonor/zmk-config/actions/runs/18235050609/job/51927071288#step:5:35) because [`dash` doesn't implement `${string/substitution/replacement}`](https://mywiki.wooledge.org/Bashism#Parameter_Expansions).

The easiest fix would be to replace the `sh -x` invocation with `bash -x`. This will allow GHA users to test Zephyr 4.1 without requiring a ZMK fork.

petejohanson:
> Since the original suggestion has been marked as resolved, I'll reiterate it: [`build-user-config.yml` is currently broken](https://github.com/lesshonor/zmk-config/actions/runs/18235050609/job/51927071288#step:5:35) because [`dash` doesn't implement `${string/substitution/replacement}`](https://mywiki.wooledge.org/Bashism#Parameter_Expansions).
> 
> The easiest fix would be to replace the `sh -x` invocation with `bash -x`. This will allow GHA users to test Zephyr 4.1 without requiring a ZMK fork.

Apologies, did that in error late last night!

petejohanson:
> Since the original suggestion has been marked as resolved, I'll reiterate it: [`build-user-config.yml` is currently broken](https://github.com/lesshonor/zmk-config/actions/runs/18235050609/job/51927071288#step:5:35) because [`dash` doesn't implement `${string/substitution/replacement}`](https://mywiki.wooledge.org/Bashism#Parameter_Expansions).
> 
> The easiest fix would be to replace the `sh -x` invocation with `bash -x`. This will allow GHA users to test Zephyr 4.1 without requiring a ZMK fork.

Switched it to explicitly invoke `bash` there.

wlo2:
Can confirm `&bootloader` is now correctly puts dongle into the bootloader if called from macro (`Central` locality) for Seeed XIAO nRF52840 with a Adafruit bootloader, solid work!

app/Kconfig
```diff
@@ -186,8 +189,8 @@ config BT_CTLR_PHY_2M
     default n if ZMK_BLE_EXPERIMENTAL_CONN
 
 # BT_TINYCRYPT_ECC is required for BT_SMP_SC_PAIR_ONLY when using HCI
-config BT_TINYCRYPT_ECC
-    default y if BT_HCI && !BT_CTLR
+# config BT_TINYCRYPT_ECC
+    # default y if BT_HCI && !BT_CTLR
```
caksoylar: Do you want to keep this or remove?

app/Kconfig
```diff
@@ -705,7 +717,7 @@ rsource "boards/shields/*/Kconfig.shield"
 # Duplicated from Kconfig.zephyr
 osource "$(KCONFIG_BINARY_DIR)/Kconfig.shield.defconfig"
 
-source "$(BOARD_DIR)/Kconfig.defconfig"
+# source "$(BOARD_DIR)/Kconfig.defconfig"
```
caksoylar: Same, should delete?

app/boards/joric/nrfmicro/nrfmicro_nrf52840_1_3_0_defconfig
```diff
@@ -1,8 +1,6 @@
 # SPDX-License-Identifier: MIT
 
-CONFIG_SOC_SERIES_NRF52X=y
-CONFIG_SOC_NRF52840_QIAA=y
-CONFIG_BOARD_NRFMICRO_13=y
+# CONFIG_BOARD_NRFMICRO_13=y
```
caksoylar:
```suggestion
```

app/boards/nicekeyboards/nice_nano/Kconfig.defconfig
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) 2021 The ZMK Contributors
 # SPDX-License-Identifier: MIT
 
-config BOARD_S40NC
-    bool "S40NC"
-    depends on SOC_NRF52840_QIAA
+if BOARD_NICE_NANO
+
+endif # BOARD_NICE_NANO
```
caksoylar: Can be removed?

app/boards/shields/murphpad/murphpad.conf
caksoylar, line 1: Since the default y for ZMK_DISPLAY was removed from Kconfig.defconfig, we might want to edit line 10 here (I can't add a comment there)

app/src/boot/Kconfig.defaults
```diff
@@ -0,0 +1,23 @@
+
+config ZMK_DBL_TAP_BOOTLOADER
+    default y if !SOC_SERIES_NRF52X && RETENTION_BOOT_MODE
+
+if ZMK_DBL_TAP_BOOTLOADER
+
+config ZMK_DBL_TAP_BOOTLOADER_TIMEOUT
```
caksoylar: Given this might be a user-facing value, I'd recommend adding a `_MS` prefix here

app/keymap-module/modules/modules.cmake
```diff
@@ -59,163 +59,3 @@ if (ZMK_CONFIG)
         list(APPEND DTS_ROOT ${ZMK_CONFIG})
     endif()
 endif()
-
-
-if(DEFINED SHIELD)
-    string(REPLACE " " ";" SHIELD_AS_LIST "${SHIELD}")
```
caksoylar, line 65: Is this logic preserved?

app/src/display/main.c
```diff
@@ -110,10 +110,10 @@ static void initialize_theme() {
 #if IS_ENABLED(CONFIG_LV_USE_THEME_MONO)
     lv_disp_t *disp = lv_disp_get_default();
     lv_theme_t *theme =
-        lv_theme_mono_init(disp, IS_ENABLED(CONFIG_ZMK_DISPLAY_INVERT), CONFIG_LV_FONT_DEFAULT);
-    theme->font_small = CONFIG_ZMK_LV_FONT_DEFAULT_SMALL;
+        lv_theme_mono_init(disp, IS_ENABLED(CONFIG_ZMK_DISPLAY_INVERT), CONFIG_LV_FONT_DEFAULT,
+                           CONFIG_ZMK_LV_FONT_DEFAULT_SMALL, CONFIG_LV_FONT_DEFAULT);
```
caksoylar, line 114: I couldn't find an API reference for this function that accepts two font pointers, but if you tested feel free to mark resolved

petejohanson: I had to make some minor LVGL changes to keep supporting our functionality, they're pushed here: https://github.com/zmkfirmware/lvgl/commit/f1db87ee98f1810328a8419572fa42a3b5f352ae

app/src/boot/Kconfig.defaults
petejohanson: Added, and added a small section to the "system" config doc to document these options.

app/Kconfig
petejohanson: Removed.

app/boards/shields/murphpad/murphpad.conf
petejohanson: Inverted the logic in the commented config line, and adjusted the comment accordingly. Thanks.

app/keymap-module/modules/modules.cmake
petejohanson: This whole thing got reworked into `app/boards/post_boards_shields.cmake`

app/src/display/main.c
caksoylar: Ah OK, that's a bit unfortunate to break the API for any modules etc. but I wouldn't expect this function to be used much? I assume you considered alternatives, so I am happy to leave as is.

app/keymap-module/modules/modules.cmake
caksoylar: I didn't spot an equivalent statement to this in there (with my admittedly bad cmake knowledge), but it's working so must be happening somewhere :)

app/src/physical_layouts.c
```diff
@@ -180,6 +213,55 @@ static struct zmk_kscan_msg_processor {
 K_MSGQ_DEFINE(physical_layouts_kscan_msgq, sizeof(struct zmk_kscan_event),
               CONFIG_ZMK_KSCAN_EVENT_QUEUE_SIZE, 4);
 
+#if IS_ENABLED(CONFIG_INPUT)
```
caksoylar: Is there a different check we can put here, like `DT_HAS_CHOSEN(zmk_matrix_input)`? This is resulting in a warning when CONFIG_INPUT is turned on by other means, like with pointing.

petejohanson: Fixed this with some more subtle pre-preprocessor checks here.

app/boards/lowprokb/corneish_zen/corneish_zen.zmk.yml
```diff
@@ -11,6 +11,9 @@ features:
 outputs:
   - usb
   - ble
+revisions:
```
joelspadin, line 14: The ID and name still list "v2".

There are a few other places that still have "_v1" or "_v2" in them too, including

- corneish_zen.yaml
- bt60_1_0_0.keymap
- bt60_2_0_0.keymap
- bt65.keymap
- bt75.keymap
- Tons of `nice_nano_v2` instead of `nice_nano@2` in various README.md files

app/boards/polarityworks/bt60/bt60.zmk.yml
```diff
@@ -1,5 +1,5 @@
 file_format: "1"
-id: bt60_v2
+id: bt60
```
joelspadin, line 2: Missing revisions for this keyboard, even though all the files have revisions.

app/boards/lowprokb/corneish_zen/corneish_zen.zmk.yml
petejohanson: Thanks. Fixed al of those.

PR #3075, opened 2025-09-28
feat: add dongle configuration support for splitkb aurora sofle

sashasimkin:
Added shields `splitkb_aurora_sofle_dongle` and `splitkb_aurora_sofle_peripheral_left` to be able to build dongled configuration for this keyboard.

FWIW - the change was tested in my configuration build: https://github.com/sashasimkin/zmk-aurora-sofle-v2/actions/runs/18071698605

## PR check-list


caksoylar:
Hi there, thanks for the PR. I don't believe we intend to accept additional dongle shields for keyboards, and instead encourage folks to follow the recipe at https://zmk.dev/docs/development/hardware-integration/dongle within their own user config if they want to add a dongle to their keyboards that are maintained upstream. I think an exception could be considered if the keyboard ships with a dongle as a product (but that's not the case here AFAIK).

sashasimkin:
Hi! Sorry for the long reply. And totally understood.

The reason I thought PR would be useful because when going thru official
instruction I had to copy the .dtsi file into my own project, which wasn’t
very clean.

Is there a way to implement dongle without copying dtsi into my repo?


пн, 6 жовт. 2025 р. о 23:22 Cem Aksoylar ***@***.***> пише:

caksoylar:
I don't think you need to copy the whole .dtsi file, just the relevant parts of it to the new overlay. But there is no way to `#include` it, as far as I know.

PR #3085, opened 2025-10-03
Add config for enabling/disabling wpm widget for nice view shield

andrewscodes:
## PR check-list

### DIFF
--- a/app/boards/shields/nice_view/widgets/status.c
+++ b/app/boards/shields/nice_view/widgets/status.c
@@ -51,8 +51,10 @@ static void draw_top(lv_obj_t *widget, lv_color_t cbuf[], const struct status_st
 
     lv_draw_label_dsc_t label_dsc;
     init_label_dsc(&label_dsc, LVGL_FOREGROUND, &lv_font_montserrat_16, LV_TEXT_ALIGN_RIGHT);
+#if IS_ENABLED(CONFIG_NICE_VIEW_WPM_WIDGET)
     lv_draw_label_dsc_t label_dsc_wpm;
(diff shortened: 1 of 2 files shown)


app/boards/shields/nice_view/widgets/status.c
```diff
@@ -325,9 +330,11 @@ struct wpm_status_state wpm_status_get_state(const zmk_event_t *eh) {
     return (struct wpm_status_state){.wpm = zmk_wpm_get_state()};
 };
 
+#if IS_ENABLED(CONFIG_NICE_VIEW_WPM_WIDGET)
```
snoyer, line 333:
```suggestion
#if IS_ENABLED(CONFIG_ZMK_WPM)
```
I believe you should use the existing `CONFIG_ZMK_WPM` (https://zmk.dev/docs/config/system#general) here and around `widget_wpm_status_init();`, and `CONFIG_ZMK_WIDGET_WPM_STATUS` to show the widget

andrewscodes: I remembered why I added a new config, I looks like using "select zmk_wpm" will always enable zmk_wpm regardless If we disable it in the config
image

snoyer: Ideally that would need to be conditional because having it not selected would allow the backend WPM computation code to not be compiled at all instead of just not used when the widget isn't displayed.

I assume there is a way to have `ZMK_WPM` selected when `CONFIG_ZMK_WIDGET_WPM_STATUS` is set but I don't know enough KConfig to be sure how

snoyer: @petejohanson disagrees (see Discord) so I was wrong and you can disregard

PR #3102, opened 2025-10-21
fix(pm): add state check to soft off device suspension

carrefinho:
Check device state before suspension during soft-off to avoid unexpected behavior from incorrectly suspended devices, using the same existing pattern in `zmk_pm_suspend_devices`. 
Fixes high current consumption on XIAO nRF52840 upon wakeup from soft-off, likely caused by the QSPI flash being put in a weird state.

Tested on XIAO nRF52840 and XIAO nRF52840 Plus:
- Before: 4.6mA current draw after wake from soft-off (requires multiple additional resets to clear)
- After: Normal 40µA current draw after wake from soft-off.

## PR check-list

- [X] Branch has a [clean commit history](https://zmk.dev/docs/development/contributing/pull-requests#clean-commit-history)
- [X] [Pre-commit](https://zmk.dev/docs/development/local-toolchain/pre-commit) used to check formatting of files, commit messages, etc.

### DIFF
--- a/app/src/pm.c
+++ b/app/src/pm.c
@@ -108,6 +108,11 @@ int zmk_pm_soft_off(void) {
     for (int i = 0; i < device_count; i++) {
         const struct device *dev = &devs[i];
 
+        if (!device_is_ready(dev) || pm_device_is_busy(dev) || pm_device_state_is_locked(dev) ||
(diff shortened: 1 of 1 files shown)


carrefinho:
Closing because the root cause of this particular power consumption issue is fixed in Zephyr 4.1 update.

PR #3107, opened 2025-11-04
feat(docs): Update New Behavior Guide

kurtis-lew:
Updates new behavior guide, focusing on modules as per the [February 2025 Contributor Sync](https://zmk.dev/blog/2025/03/20/contributor-sync-1#cems-summary)

Key changes:
- Separate the "Behavior Creation" into module setup and source file development sections
    - Allows for more detail regarding underlying Zephyr APIs and ZMK-specific features
- Describe methods of performing common behavior actions:
    - Invoking keycodes
    - Working with layers
    - Invoking other behaviors
    - Working with behavior queue

Some text has been rewritten for more conciseness. Efforts in this area are ongoing.

Certain code blocks have been substituted because their source files change quite frequently (e.g., not referring to hold-tap code as often).

### DIFF
--- a/docs/docs/development/new-behavior.mdx
+++ b/docs/docs/development/new-behavior.mdx
@@ -12,20 +12,27 @@ Before reading this section, it is **vital** that you read through our [clean ro
 
 ## Overview
 
-This document outlines how to develop a [behavior](../keymaps/behaviors/index.mdx) for ZMK and prepare the changes for a pull request.
+[Behaviors](../keymaps/behaviors/index.mdx) refer to the actions that are invoked when a key is pressed or released.
+This guide outlines how to create a [ZMK module](../development/module-creation.md) that contains a new [behavior](../keymaps/behaviors/index.mdx).
 
-Behaviors are assigned to key positions and determine what happens when they are pressed and released. They are implemented in Zephyr as "devices": they consist of a devicetree binding file, which specifies the properties of the behavior, and a driver written in C code. This allows for the ability to create unique instances of these behaviors in [keymaps](../keymaps/index.mdx) or devicetree-source-include files (`.dtsi`). While instances of behaviors stored in keymaps are created by end-users for their personal needs, the instances that live in the .dtsi files are stored and documented in ZMK directly, which removes the need for end-users to set up common use-cases of these behaviors in their personal keymaps.
+:::info
+If an out-of-tree behavior's use-case is deemed widespread enough to be merged into upstream ZMK, a new pull request may be issued instead.
+While this page describes practices that maximize compatibility between module-focused development and pull-request-based workflows, we also note specific modifications required for the latter to work properly.
+:::
+
+In the context of the Zephyr RTOS, behaviors are implemented as "devices", which consist of:
+
+- A devicetree binding file, which declares the behavior's properties
+- A device driver written in C code
+- Optionally, devicetree-source-include files (`.dtsi`), which contain predefined instances of the behavior that may be included directly in keymaps
 
 The general process for developing behaviors is:
 
-1. [Create the behavior](#creating-the-behavior)
-   1. [Create the devicetree binding (`.yaml`)](#creating-the-devicetree-binding-yaml)
-   1. [Create the driver (`.c`)](#creating-the-driver-c)
-   1. [Update `app/CmakeLists.txt` to include the new driver](#updating-appcmakeliststxt-to-include-the-new-driver)
-   1. [Define common use-cases for the behavior (`.dtsi`) (Optional)](#defining-common-use-cases-for-the-behavior-dtsi-optional)
+1. [Create a new behavior repository](#creating-a-new-behavior-repository)
+1. [Develop the behavior functionality](#developing-the-behavior-functionality)
 1. [Test changes locally](#testing-changes-locally)
 1. [Document behavior functionality](#documenting-behavior-functionality)
-1. [Create a pull request for review and inclusion into the ZMK sources](#submitting-a-pull-request)
+1. [Provide licensing information](#licensing-information)
 
 :::info
 Before developing new behaviors, developers should have a working knowledge of the Embedded Linux Devicetree.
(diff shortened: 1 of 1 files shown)


docs/docs/development/new-behavior.mdx
```diff
@@ -12,20 +12,46 @@ Before reading this section, it is **vital** that you read through our [clean ro
 
 ## Overview
 
-This document outlines how to develop a [behavior](../keymaps/behaviors/index.mdx) for ZMK and prepare the changes for a pull request.
+Behaviors refer to the actions that are invoked when a key is pressed or released.
```
caksoylar:
```suggestion
[Behaviors](../keymaps/behaviors/index.mdx) refer to the actions that are invoked when a key is pressed or released.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -12,20 +12,46 @@ Before reading this section, it is **vital** that you read through our [clean ro
...
 
-Behaviors are assigned to key positions and determine what happens when they are pressed and released. They are implemented in Zephyr as "devices": they consist of a devicetree binding file, which specifies the properties of the behavior, and a driver written in C code. This allows for the ability to create unique instances of these behaviors in [keymaps](../keymaps/index.mdx) or devicetree-source-include files (`.dtsi`). While instances of behaviors stored in keymaps are created by end-users for their personal needs, the instances that live in the .dtsi files are stored and documented in ZMK directly, which removes the need for end-users to set up common use-cases of these behaviors in their personal keymaps.
+:::info
+If an out-of-tree behavior's use-case is deemed widespread enough to be merged into upstream ZMK, a new pull request may be issued instead.
+While this page describes practices that maximize compatibility between module-focused development and pull-request-based workflows, we also note specific modifications required for the latter to work properly.
+:::
+
+In the context of the Zephyr RTOS, behaviors are implemented as "devices", which contain:
```
caksoylar:
```suggestion
In the context of the Zephyr RTOS, behaviors are implemented as "devices," which consist of:
```

docs/docs/development/new-behavior.mdx
```diff
@@ -12,20 +12,46 @@ Before reading this section, it is **vital** that you read through our [clean ro
...
+
+- A devicetree binding file, which declare behavior's properties
```
caksoylar:
```suggestion
- A devicetree binding file, which declares behavior's properties
```

docs/docs/development/new-behavior.mdx
```diff
@@ -12,20 +12,46 @@ Before reading this section, it is **vital** that you read through our [clean ro
...
+- A device driver written in C code
+- Optionally, devicetree-source-include files (`.dtsi`), which contain instances of the behavior that may be included directly in keymaps, rather than being explicitly defined in a `behaviors` node
```
caksoylar:
```suggestion
- Optionally, devicetree-source-include files (`.dtsi`), which contain predefined instances of the behavior that may be included directly in keymaps
```
I think it's cleaner this way, the "behaviors" node name is not technically necessary etc. so we don't need to get into it

docs/docs/development/new-behavior.mdx
```diff
@@ -12,20 +12,46 @@ Before reading this section, it is **vital** that you read through our [clean ro
...
 
 The general process for developing behaviors is:
 
-1. [Create the behavior](#creating-the-behavior)
-   1. [Create the devicetree binding (`.yaml`)](#creating-the-devicetree-binding-yaml)
-   1. [Create the driver (`.c`)](#creating-the-driver-c)
-   1. [Update `app/CmakeLists.txt` to include the new driver](#updating-appcmakeliststxt-to-include-the-new-driver)
-   1. [Define common use-cases for the behavior (`.dtsi`) (Optional)](#defining-common-use-cases-for-the-behavior-dtsi-optional)
+1. [Create a new behavior repository](#creating-a-new-behavior-repository)
```
caksoylar, line 31: I think it might be better to just leave the top-level items here, what do you think? This looks too much like a ToC.

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+
+### Initializing a new Behavior Module
+
+1. Navigate to [the ZMK module template repository](https://github.com/zmkfirmware/zmk-module-template)
+1. Select **"Use this template"** in the upper right corner, followed by **"Create a new repository"**
+1. Choose an appropriate name for your new repository.
+   ZMK behavior modules should follow the naming convention, **`zmk-behavior-<name-of-behavior>`**, using all lowercase letters and dashes to separate words
+1. Complete the module's creation by selecting the repository's visibility, before clicking **"Create repository"**
+1. Clone a copy of your module to your development environment.
+   It is easiest if your behavior resides in the same folder as your copy of the ZMK source code, as shown in the tree below.
```
caksoylar: Is this explained somewhere? I can't think of why it is particularly easier this way

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+// highlight-end
+// highlight-next-line
+├── west.yml (optional)
+└── zephyr
+    └── module.yml
+```
 
-### Creating the Devicetree Binding (`.yaml`)
+For more information on module preparation, refer to the page on [module creation](./module-creation.md).
+This is especially the case for procedures involving the `west.yml` and `zephyr/module.yml` files.
```
caksoylar:
```suggestion
For more information on module preparation, such as details on `west.yml` and `zephyr/module.yml` contents, refer to the page on [module creation](./module-creation.md).
```

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
 
-#### `DT_DRV_COMPAT`
+</TabItem>
+</Tabs>
 
-Replace `zmk_<behavior_name>` in the `#define DT_DRV_COMPAT` statement with the name of your behavior. (e.g. `zmk_behavior_caps_word`)
+A more thorough explanation of the contents of a behavior source file can be [found below](#developing-the-behavior-functionality).
 
-#### Dependencies
+Other source files may be created as well, as is often the case when creating new features or **events** from scratch.
```
caksoylar:
```suggestion
Other source files may be created as well, as is often the case when creating new features or [events](events.md) from scratch.
```
Is this intentionally bolded? I think it'd be better to link the events page here.

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+};
+
+// Initialization Function (Optional)
+static int <name_of_behavior>_init(const struct device *dev) {
+    return 0;
+};
+
+// API struct
+static const struct behavior_driver_api <name_of_behavior>_driver_api = {
+
```
caksoylar: Should we add stubs for press/release functions and refer to them here? There aren't explicitly mentioned in the dedicated section, though there is an example is in the "Keycodes" header.

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
 
-Replace `zmk_<behavior_name>` in the `#define DT_DRV_COMPAT` statement with the name of your behavior. (e.g. `zmk_behavior_caps_word`)
+A more thorough explanation of the contents of a behavior source file can be [found below](#developing-the-behavior-functionality).
 
-#### Dependencies
+Other source files may be created as well, as is often the case when creating new features or **events** from scratch.
+For parity with upstream ZMK, these files will generally be placed in the root `src/` directory, or `src/events/`.
 
-The dependencies required for any ZMK behavior are:
+### Updating `KConfig`
```
caksoylar: Can we `s/KConfig/Kconfig/` throughout the document? That's the usual capitalization I see we and Zephyr use.

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
-ZMK_SUBSCRIPTION(behavior_tap_dance, zmk_position_state_changed);
-static int tap_dance_position_state_changed_listener(const zmk_event_t *eh){
-    // Do stuff...
-}
+For more examples of behavior-specific KConfig settings, see [Behavior Configuration](../config/behaviors.md).
+:::
+
+### Updating `CMakeLists.txt`
+
+`CMakeLists.txt` files are used in Zephyr's [**configuration stage**](https://docs.zephyrproject.org/3.5.0/build/cmake/index.html) when building firmware.
```
caksoylar:
```suggestion
`CMakeLists.txt` files are used in Zephyr's [configuration stage](https://docs.zephyrproject.org/3.5.0/build/cmake/index.html) when building firmware.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+This is what allows us to specify which source files to include, which may depend on the KConfig settings shown previously.
```
caksoylar:
```suggestion
These specify which source files are included in the build, and may depend on the KConfig settings shown previously.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+| -------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Locality is unibody, Central side of split, peripheral side of split | if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_ROLE_CENTRAL)                                                                                      |
+| Locality is **only** on the central unit of a split keyboard         | if (CONFIG_ZMK_SPLIT AND CONFIG_ZMK_SPLIT_ROLE_CENTRAL)                                                                                           |
+| Locality is **only** on the peripheral unit of a split keyboard      | if (CONFIG_ZMK_SPLIT AND (NOT CONFIG_ZMK_SPLIT_ROLE_CENTRAL))                                                                                     |
+| KConfig Requirement must be met                                      | Use `target_sources_ifdef(CONFIG_<Configuration Requirement> app PRIVATE <name_of_behavior>.c)` instead of `target_sources(<name_of_behavior>.c)` |
```
caksoylar:
```suggestion
| Locality is unibody, or the central part of split | `if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_ROLE_CENTRAL)`                                                                                      |
| Locality is **only** on the central part of a split keyboard         | `if (CONFIG_ZMK_SPLIT AND CONFIG_ZMK_SPLIT_ROLE_CENTRAL)`                                                                                           |
| Locality is **only** on the peripheral part of a split keyboard      | `if (CONFIG_ZMK_SPLIT AND (NOT CONFIG_ZMK_SPLIT_ROLE_CENTRAL))`                                                                                     |
| KConfig Requirement must be met                                      | Use `target_sources_ifdef(CONFIG_<Configuration Requirement> app PRIVATE <name_of_behavior>.c)` instead of `target_sources(<name_of_behavior>.c)` |
```
(Will need realigning)
a few changes:
- "part" terminology instead of "unit"
- fixing what I assume is a bug on the first row condition
- code fencing

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+        };
+    };
+};
 
-DT_INST_FOREACH_STATUS_OKAY(KP_INST)
 ```
 
-Note that in the hold-tap example, the instance number, `0`, has been replaced by `n`, signifying the unique `node_id` of each instance of a behavior. Furthermore, the DT_INST_FOREACH_STATUS_OKAY(KP_INST) macro iterates through each compatible, non-disabled devicetree node, creating and applying the proper values to any instance-specific configurations or data by invoking the KP_INST macro for each instance of the new behavior.
+The format of a behavior's `.dtsi` file is identical to declaring an instance of the behavior in a user's keymap.
+However, a major difference is that the value `/omit-if-no-ref/` should be placed adjacent to the label and name of the behavior, as highlighted in the example.
```
caksoylar, line 466:
```suggestion
However, a major difference is that the value `/omit-if-no-ref/` should be placed adjacent to the label and name of the behavior, as highlighted in the example.
This enables the behavior to only be compiled if it is used in the keymap.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
 
-##### Initialization function (optional)
+:::warning
+If submitting a pull request, any `.c` files should be formatted according to `clang-format` to ensure that automated checks run smoothly.
+:::
+
+### Compatible: `#define DT_DRV_COMPAT`
 
-Comes in the form `static int <behavior_name>_init(const struct device *dev)`. Initialization functions preconfigure any data, like resetting timers and position for hold-taps and tap-dances. All initialization functions `return 0;` once complete.
+This field should match the [`compatible` field of your devicetree binding](#compatible).
```
caksoylar:
```suggestion
This field should match the [`compatible` field of your devicetree binding](#compatible), with commas and hyphens replaced by underscores.
```
It'd be better if we can make it more precise but it's good enough, IMO

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
 
-##### API structure
+For example, the Caps Word behavior's devicetree binding lists `compatible: "zmk,behavior-caps-word"`, so the `DT_DRV_COMPAT` is `zmk_behavior_caps_word`.
 
-Comes in the form `static const struct behavior_driver_api <behavior_name>_driver_api)`. Common items to include in the API Structure are:
+### Dependencies
+
+The dependencies required for any ZMK behavior are:
+
+- `device.h`: [Zephyr Device APIs](https://docs.zephyrproject.org/apidoc/latest/group__device__model.html)
```
caksoylar:
```suggestion
- `zephyr/device.h`: [Zephyr Device APIs](https://docs.zephyrproject.org/apidoc/3.5.0/group__device__model.html)
```

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+- `drivers/behavior.h`: ZMK Behavior Functions (e.g. [`locality`](#zmk-api-struct), `behavior_keymap_binding_pressed`, `behavior_keymap_binding_released`, `behavior_sensor_keymap_binding_triggered`)
```
caksoylar:
```suggestion
- `drivers/behavior.h`: ZMK Behavior Functions (e.g. [locality](#zmk-api-struct), `behavior_keymap_binding_pressed`, `behavior_keymap_binding_released`, `behavior_sensor_keymap_binding_triggered`)
```

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
+- `logging/log.h`: [Zephyr Logging APIs](https://docs.zephyrproject.org/3.5.0/services/logging/index.html) (for more information on USB Logging in ZMK, see [USB Logging](usb-logging.mdx)).
```
caksoylar:
```suggestion
- `zephyr/logging/log.h`: [Zephyr Logging APIs](https://docs.zephyrproject.org/3.5.0/services/logging/index.html) (for more information on USB Logging in ZMK, see [USB Logging](usb-logging.mdx)).
```

docs/docs/development/new-behavior.mdx
```diff
@@ -37,270 +63,495 @@ The following resources are provided for those seeking further understanding:
...
 
 - [Key press](https://github.com/zmkfirmware/zmk/blob/main/app/src/behaviors/behavior_key_press.c#L21)
 - [RGB underglow](https://github.com/zmkfirmware/zmk/blob/main/app/src/behaviors/behavior_rgb_underglow.c#L23)
 - [Hold-tap](https://github.com/zmkfirmware/zmk/blob/main/app/src/behaviors/behavior_hold_tap.c#L680), which is dynamic based on what behaviors are set up in the hold-tap bindings
 
 Behavior metadata consists of one or more metadata sets, where each metadata set has a set of values for the parameter(s) used with the behavior.
 
-For example, a common approach for behaviors is to have a set of possible first parameters that identify the "command" to invoke for the behavior, and the second parameter is a detail/sub-parameter to the action. You can see this with the `&bt` behavior.
+For example, a common approach for behaviors is to have a set of possible first parameters that identify the "command" to invoke for the behavior, and the second parameter is a detail/sub-parameter to the action.
+You can see this with the `&bt` behavior.
```
caksoylar:
```suggestion
You can see an example of this with the `&bt` behavior.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+
+Behaviors also require the following parameters of `BEHAVIOR_DT_INST_DEFINE` to be changed:
+
+#### Optional: Initialization function
+
+Comes in the form `static int <name_of_behavior>_init(const struct device *dev)`.
+Initialization functions preconfigure any data, like resetting timers and position for hold-taps and tap-dances.
+All initialization functions `return 0;` once complete.
+
+The **second** cell of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if an initialization function is not required.
```
caksoylar:
```suggestion
The **second** argument of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if an initialization function is not required.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
 ```c
 const struct device *dev = zmk_behavior_get_binding(binding->behavior_dev);
-struct behavior_<behavior_name>_data *data = dev->data;
+struct behavior_<name_of_behavior>_data *data = dev->data;
 ```
 
 The variables stored inside the data `struct`, `data`, can be then modified as necessary.
 
-The fourth cell of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if instance-specific data is not required.
+The **fourth** cell of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if instance-specific data is not required.
```
caksoylar:
```suggestion
The **fourth** argument of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if instance-specific data is not required.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+As seen in the `#define KP_INST(n)` of the hold-tap example, the configuration `struct`, `behavior_<name_of_behavior>_config_##n`, for each instance number, `n`, can be initialized using the [Zephyr Devicetree Instance-based APIs](https://docs.zephyrproject.org/3.5.0/build/dts/api/api.html#instance-based-apis),
+which extract the values from the `properties` of each instance of the [devicetree binding](#devicetree-bindings-yaml) from a user's keymap or [predefined use-case `.dtsi` files](#optional-defining-common-use-cases-for-the-behavior-dtsi) stored in `app/dts/behaviors/`.
+We illustrate this further by comparing the [`#define KP_INST(n)` from the hold-tap driver](#invoking-behavior_dt_inst_define) and the [`properties` of the hold-tap devicetree binding](#devicetree-bindings-yaml).
+The config structure instances should always be declared `const`
 so they are placed into flash, not RAM, by the linker.
 
-The fifth cell of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if instance-specific configurations are not required.
+The **fifth** cell of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if instance-specific configurations are not required.
```
caksoylar:
```suggestion
The **fifth** argument of `BEHAVIOR_DT_INST_DEFINE` can be set to `NULL` instead if instance-specific configurations are not required.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+### Layers
 
-- Behavior applies to unibody, or central or peripheral half of keyboard: place `target_sources(app PRIVATE <behavior_name>.c)` line _before_ `if ((NOT CONFIG_ZMK_SPLIT) OR CONFIG_ZMK_SPLIT_ROLE_CENTRAL)`
-- Behavior applies to _only_ central half of split keyboard: place `target_sources(app PRIVATE <behavior_name>.c)` after `if (CONFIG_ZMK_SPLIT AND CONFIG_ZMK_SPLIT_ROLE_CENTRAL)`
-- Behavior applies to _only_ peripheral half of split keyboard: place `target_sources(app PRIVATE <behavior_name>.c)` after `if (CONFIG_ZMK_SPLIT AND (NOT CONFIG_ZMK_SPLIT_ROLE_CENTRAL))`
-- Behavior requires certain condition in a keyboard's `.conf` file to be met: use `target_sources_ifdef(CONFIG_<Configuration Requirement> app PRIVATE <behavior_name>.c)` instead of `target_sources(<behavior_name>.c)`
+All functions that interact with layers can be found in `keymap.h`.
 
-### Defining Common Use-Cases for the Behavior (`.dtsi`) (Optional)
+Layers can be identified in two ways: a **order-independent** `zmk_keymap_layer_id_t`, and an `order-dependent` `zmk_keymap_layer_index_t`.
```
caksoylar:
```suggestion
Layers can be identified in two ways: an "order-independent" `zmk_keymap_layer_id_t`, and an "order-dependent" `zmk_keymap_layer_index_t`.
The latter is used by ZMK Studio and the two values will be different when layers are re-ordered through Studio interfaces.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+Additionally, a bitmask, `zmk_keymap_layers_state_t`, is used to keep track of which layers are active or not, based on their **order-independent IDs**.
```
caksoylar: I think this might be unnecessary, given it's just the type definition returned by `zmk_keymap_layer_state(void)`

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+| `zmk_keymap_highest_layer_active(void)`                    | Returns a `zmk_keymap_layer_index_t` the ordered layer index of the highest active layer.                                    |
+| `zmk_keymap_layer_activate(zmk_keymap_layer_id_t layer)`   | Activates the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs.                                  |
+| `zmk_keymap_layer_deactivate(zmk_keymap_layer_id_t layer)` | Deactivates the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs.                                |
+| `zmk_keymap_layer_toggle(zmk_keymap_layer_id_t layer)`     | Toggles the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs.                                    |
+| `zmk_keymap_layer_to(zmk_keymap_layer_id_t layer)`         | Deactivates every layer, before activating the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs. |
+| `zmk_keymap_layer_name(zmk_keymap_layer_id_t layer)`       | Returns a C-string containing the layer's name.                                                                              |
```
caksoylar:
```suggestion
| `zmk_keymap_layer_name(zmk_keymap_layer_id_t layer)`       | Returns a C-string containing the layer's name.                                                                              |
| `zmk_keymap_layer_index_to_id(zmk_keymap_layer_index_t layer_index)` | Returns the order-independent ID for a given layer index.                                                                              |
```
I thought this is worth adding given the distinction made above, but feel free to reword

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
-```
+Some examples of events that are the most relevant to behavior development can be seen below.
 
-The format of a behavior's `.dtsi` file is identical to declaring an instance of the behavior in a user's keymap. The only major difference is that the value `/omit-if-no-ref/` should be placed adjacent to the label and name of the behavior, as seen in line 11 of the `gresc` example.
+| Event                       | Description                                                                                       |
+| --------------------------- | ------------------------------------------------------------------------------------------------- |
+| `hid_indicators_changed.h`  | The current HID indicators (Num Lock, Caps Lock, Scroll Lock, Compose, Kana) as a bitmask         |
+| `keycode_state_changed.h`   | [Keycode events' state (on/off)](#keycodes), usage page, keycode value, modifiers, and timestamps |
+| `layer_state_changed.h`     | [Layer events' state (bitmask)](#layers), layer index, and timestamps                             |
+| `modifiers_state_changed.h` | Modifier events' state (on/off) and modifier values                                               |
```
caksoylar:
```suggestion

```
AFAIK nothing raises this event, so it is not currently working

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+| `zmk_keymap_layer_activate(zmk_keymap_layer_id_t layer)`   | Activates the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs.                                  |
+| `zmk_keymap_layer_deactivate(zmk_keymap_layer_id_t layer)` | Deactivates the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs.                                |
+| `zmk_keymap_layer_toggle(zmk_keymap_layer_id_t layer)`     | Toggles the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs.                                    |
+| `zmk_keymap_layer_to(zmk_keymap_layer_id_t layer)`         | Deactivates every layer, before activating the chosen layer. Returns 0 if successful. Returns values < 0 if an error occurs. |
+| `zmk_keymap_layer_name(zmk_keymap_layer_id_t layer)`       | Returns a C-string containing the layer's name.                                                                              |
 
-For the purpose of this section, we will discuss the structure of `app/dts/behaviors/gresc.dtsi` below.
+### ZMK Event Manager
```
caksoylar:
```suggestion
### ZMK Events
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
 
-```dts
-// Behavior can be invoked on peripherals, so name must be <= 8 characters.
-/omit-if-no-ref/ gresc: gresc { ... };
-```
+#### `#include <zmk/behavior.h>`
 
-:::
+These functions work with behaviors at a **device** level.
+They are used to retrieve the device associateed with a keymap binding, or invoke the behaviors in a parameter list.
```
caksoylar:
```suggestion
They are used to retrieve the device associated with a keymap binding, or invoke other behaviors, such as ones provided as a parameter to the current behavior.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
-#include <behaviors/caps_word.dtsi>
-#include <behaviors/key_repeat.dtsi>
-#include <behaviors/backlight.dtsi>
-#include <behaviors/macros.dtsi>
-// highlight-next-line
-#include <behaviors/new_behavior_instance.dtsi>
-```
+| Function                      | Description                                                     |
+| ----------------------------- | --------------------------------------------------------------- |
+| `zmk_behavior_get_binding`    | Get a const struct device\* for a behavior from its name field. |
```
caksoylar:
```suggestion
| `zmk_behavior_get_binding`    | Get a `const struct device*` for a behavior from its name field. |
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+| `zmk_behavior_invoke_binding` | Invoke a behavior given its binding and invoking event details. |
+
```
caksoylar, line 827: Should we also describe `zmk_behavior_invoke_binding` and `zmk_behavior_binding_event` structs in this section? We don't need to necessarily go item-by-item, but it can be one sentence each describing what they contain.

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+#### `#include <zmk/behavior_queue.h>`
+
+The behavior queue is leveraged by [macros](../keymaps/behaviors/macros.md) and [sensor rotation](../keymaps/behaviors/sensor-rotate.md) behaviors.
+This is used to invoke behaviors in sequence with specific timing.
```
caksoylar:
```suggestion
This is used to invoke behaviors in sequence with specific timing.
Since it is processed on a separate thread, it does not block the rest of the keyboard functionality while it is waiting for time-based triggers.
```
Is "on a separate thread" technically true? Not sure what's a better way to put this. But I think this is a critical piece on why it is used

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+
+| Function                 | Description                              |
+| ------------------------ | ---------------------------------------- |
+| `zmk_behavior_queue_add` | Adds the behavior to the behavior queue. |
 
 ## Testing Changes Locally
 
-Create a new folder in `app/tests/` to develop virtual test sets for all common use cases of the behavior. Behaviors should be tested thoroughly on both virtual testing environments using `west test` and real hardware.
+Create a new folder in `app/tests/` to develop virtual test sets for all common use cases of the behavior.
```
caksoylar:
```suggestion
Create a new folder in `tests/` (or `app/tests/` if submitting a pull request) to develop virtual test sets for all common use cases of the behavior.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -390,135 +641,206 @@ static const struct behavior_driver_api behavior_bt_driver_api = {
...
+Behaviors should be tested thoroughly on both virtual testing environments using `west test` and real hardware.
```
caksoylar:
```suggestion
For pull requests, behaviors should be tested thoroughly on both virtual testing environments using `west test` and real hardware.
```

docs/docs/development/new-behavior.mdx
```diff
@@ -536,24 +858,25 @@ Zephyr currently does not support logging over Bluetooth, so any use of the seri
...
-Consider also including visual aids alongside written documentation if it adds clarity.
+Including visual aids alongside written documentation for additional clarity may be helpful.
 
 :::info
-See [Documentation](contributing/documentation.md) for more information on writing, testing, and formatting ZMK documentation.
+If submitting a pull request, see [Documentation](contributing/documentation.md) for more information on writing, testing, and formatting ZMK documentation.
 :::
 
-## Submitting a Pull Request
+## Updating Copyright Information
```
caksoylar: Not sure if we should specialize this for "creating PR" case, or leave as is

docs/docs/development/new-behavior.mdx
kurtis-lew: Agreed. With the increase in the number of sub-headers, the preamble is a touch bulky now.

docs/docs/development/new-behavior.mdx
kurtis-lew: I think I initially wanted to include a section to prompt the user to include it as an extra module when building/testing.

https://zmk.dev/docs/development/local-toolchain/build-flash#building-with-external-modules

Though in hindsight, since building with extra modules requires an absolute path, there's no real incentive to place it near the ZMK source code.

docs/docs/development/new-behavior.mdx
kurtis-lew: I think I meant to link to the page or the header referring to behaviors' frequent use of the event manager, but it slipped my mind. Thanks for catching that!

docs/docs/development/new-behavior.mdx
kurtis-lew: Agreed. For users looking for changes in modifiers, we'd just use the implicit/explicit fields of `keycode_state_changed`, correct?

docs/docs/development/new-behavior.mdx
kurtis-lew: I'm not sure it's on a separate thread but the note that the behavior queue is non-blocking is important. Can isolate the non-blocking section as a mention.

docs/docs/development/new-behavior.mdx
kurtis-lew: Perhaps a better version would be:
```
1. Clone a copy of your module to your development environment.
   The cloned repository should be easily accessible when [building and testing firmware using a local toolchain](./local-toolchain/build-flash.mdx#building-with-external-modules).
```
Thoughts?

docs/docs/development/new-behavior.mdx
caksoylar: Yes, AFAIK

docs/docs/development/new-behavior.mdx
kurtis-lew: I've seen a few modules shared in the Discord or in the docs that lack a license. Some recent examples that come to mind include:
* Pete's cirque driver: https://github.com/petejohanson/cirque-input-module
* badjeff's DRV2605 driver: https://github.com/badjeff/zmk-drv2605-driver

I've made it a habit recently to reach out to the creators of each module where possible to ask explicitly for permission to reference their work, but if the author of a module is comfortable with others referencing their work, the licensing helps. For non-upstream work, we could note that this doesn't necessarily have to be under "The ZMK Contributors".

docs/docs/development/new-behavior.mdx
caksoylar: Sounds good

docs/docs/development/new-behavior.mdx
kurtis-lew: Just to confirm: do you mean `zmk_behavior_binding`, instead of `zmk_behavior_invoke_binding`?

Otherwise, I think it's worth including for more completeness.

docs/docs/development/new-behavior.mdx
kurtis-lew: Maybe unnecessary in this particular location but I think it's worth mentioning that the layer state refers to the aggregate state of _all_ layers represented as a bitmask. Perhaps it belongs in the table instead?

```
Layers can be identified in two ways: an "order-independent" `zmk_keymap_layer_id_t`, and an "order-dependent" `zmk_keymap_layer_index_t`.

| Function                                                             | Description                                                                                                                             |
| -------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
| `zmk_keymap_layer_default(void)`                                     | Returns a `zmk_keymap_layer_id_t` of the default layer.                                                                                 |
| `zmk_keymap_layer_state(void)`                                       | Returns the current keyboard's layer state: a bitmask where each bit represents the state of the corresponding `zmk_keymap_layer_id_t`. |
```

caksoylar: Yes, I think the function is worth mentioning in the table. :+1:

docs/docs/development/new-behavior.mdx
kurtis-lew: Just to confirm what you mean by adding stubs: are you suggesting something like below?
```diff
static const struct behavior_driver_api <name_of_behavior>_driver_api = {
+    .binding_pressed = on_<name_of_behavior>_binding_pressed,
+    .binding_released = on_<name_of_behavior>_binding_pressed,
};
```

docs/docs/development/new-behavior.mdx
caksoylar: Ah yes, sorry about the typo. I mean the two structs, which we can refer to in the table also:

`zmk_behavior_invoke_binding(zmk_behavior_binding, zmk_behavior_binding_event)` etc.

docs/docs/development/new-behavior.mdx
caksoylar: I think adding a license is worth pointing out, agreed. I was raising this because it doesn't necessarily have to be MIT or have copyright headers, if not making a PR.

docs/docs/development/new-behavior.mdx
caksoylar: Yes, in addition to adding `on_<name_of_behavior>_binding_pressed` etc. function stubs above that can be empty, or just return opaque.

docs/docs/development/new-behavior.mdx
kurtis-lew: Also agreed. Will workshop this section a bit more with that in mind.

docs/docs/development/new-behavior.mdx
kurtis-lew: I like this idea. Thanks for the suggestion!

docs/docs/development/new-behavior.mdx
kurtis-lew: Updated this section to be a more general "Provide licensing information" step in the Overview tied to a generic "Licensing Information" header. Want to emphasize I'm not a lawyer, but I've provided some resources that seem most relevant to the topics.

Separated licenses into
- Upstream ZMK/community sharing via MIT: we already had the bulk of this from before, not many changes needed
- Other licenses: Very brief. I just mention Apache and GPL exist but am stuck on what other information is relevant without getting too deep into the weeds with it. One thing that comes to mind is the "one-way" nature of copyleft/permissive licenses.

kurtis-lew: Another thought I had was repeating the clean room disclaimer in the MIT licensing section to make it more recent in readers' minds, but would like others' thoughts on if this redundancy is necessary.

docs/docs/development/new-behavior.mdx
```diff
@@ -536,24 +855,50 @@ Zephyr currently does not support logging over Bluetooth, so any use of the seri
...
+## Licensing Information
+
+:::danger
+The ZMK Project and its contributors do not claim to be legal representatives, and any material below should not considered official legal advice.
+When distributing your work, please review the terms and conditions associated with any relevant licenses thoroughly.
+:::
+
+Developers may wish to share their work with the public, which is often done by sharing a link to a GitHub repository.
+However, making a repository public does **not** automatically qualify the repository as open source, or permit others to use the works as they see fit.
+Author must provide a license that in addition to their source code, satisfy criteria that includes but is not limited to:
```
caksoylar:
```suggestion
Authors should provide a license that in addition to their source code, ideally satisfies criteria that includes but is not limited to:
```
I softened this a bit, am I missing anything about the "must" clause?

docs/docs/development/new-behavior.mdx
caksoylar: I am happy with this in general, but I'll give a chance for others to comment (e.g. @nmunnich, FYI).

I kinda think it'd be good to drop another link to clean room in the last section re: ZMK PR, like "please also consider the constraints noted in ... for contributing to ZMK"

docs/docs/development/new-behavior.mdx
kurtis-lew: The "must" clause is specifically for qualifying the module as open source. 
https://opensource.org/osd
https://opensource.guide/legal/

Was trying to steer readers towards the open-source route given ZMK's disposition by setting up this sentence with the previous statement on public != open source

caksoylar: Works for me with that additional qualification, thanks.

kurtis-lew:
Sounds good to me! Thanks again for the quick turnaround on reviews!

More than happy to get other eyes on this for any other fixes/updates/clarifications worth adding.

PR #3103, opened 2025-10-26
feat(split): Add HID indicators and physical layout set to Wired split

ReFil:
This adds support for the set hid indicators and set physical layout commands to the split transport handler, I've tested the hid indicators and all seems to work as expected

- [x] Branch has a [clean commit history](https://zmk.dev/docs/development/contributing/pull-requests#clean-commit-history)
- [x] Additional tests are included, if changing behaviors/core code that is testable.
- [x] Proper Copyright + License headers added to applicable files (Generally, we stick to "The ZMK Contributors" for copyrights to help avoid churn when files get edited)
- [x] [Pre-commit](https://zmk.dev/docs/development/local-toolchain/pre-commit) used to check formatting of files, commit messages, etc.
- [x] Includes any necessary [documentation changes](https://zmk.dev/docs/development/contributing/documentation).

PR #3120, opened 2025-11-22
fix(split): Properly return from command handler

ReFil:
#3103 introduced a small bug where spurious HID indicator events with garbage data would be raised on the peripheral when behaviours are invoked as the switch statement didn't return in each case and continued to chain down. This properly returns from each case to stop this happening

## PR check-list


petejohanson (approved):
Good catch! Thanks.

PR #3122, opened 2025-11-25
docs: Add release version to the docs, with links and warning banner

petejohanson:
Add versions to the sidebar of the documentation, and when viewing the deployment on zmk.dev, which tracks `main`, add a banner warning about possibly prefering the docs for a particular stable release.

If we like this strategy, I can add commits to v0.2-branch and v0.1-branch so those deploy as well, then include them in the list of release versions.

## PR check-list

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -1,6 +1,32 @@
+const { execSync } = require("child_process");
 const path = require("path");
 const theme = require("./src/theme/prism/themes/github");
 const darkTheme = require("./src/theme/prism/themes/github-dark-dimmed");
+const { releaseVersions } = require("./zmk-release-versions.json");
+
+const gitBranch =
+  process.env.BRANCH ||
+  execSync("git branch --show-current", { encoding: "utf-8" }).trim();
+const isDevelopmentVersion =
+  gitBranch === "main" || !/^v\d+\.\d+-branch$/.test(gitBranch);
+const versionNavbarItems = [
+  {
+    label: "Pre-Release",
+    description: "Current active development branch (main)",
+    href: "https://zmk.dev/docs/",
+  },
+  ...releaseVersions.map((r) => ({
+    label: "v" + r,
+    description: "Stable release v" + r,
+    href: `https://v${r.replaceAll(".", "-")}-branch.zmk.dev/docs/`,
+  })),
+];
+
+const versionDropDownLabel = isDevelopmentVersion
+  ? versionNavbarItems[0].label
+  : versionNavbarItems.find((item) => {
+      return item.label === gitBranch.replace("-branch", "");
+    });
 
 module.exports = {
   title: "ZMK Firmware",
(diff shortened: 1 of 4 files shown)


docs/sidebars.js
```diff
@@ -1,3 +1,20 @@
+const versions = require("./versions.js");
+
+const releaseLinks = [
+  {
+    type: "link",
+    label: "development (main)",
```
caksoylar:
```suggestion
    label: "Development (main)",
```

docs/sidebars.js
```diff
@@ -215,5 +232,17 @@ module.exports = {
         "development/events",
       ],
     },
+    {
+      type: "category",
+      label: "Versions",
+      link: {
+        type: "generated-index",
+        title: "ZMK Versions",
+        description: "Available documentation versions",
```
caksoylar:
```suggestion
        description: "Below are the available documentation versions, corresponding to ZMK versions.",
```

I think this needs a little elaborating. It's awkward without being able to provide more information on ZMK versions but I guess that's fine for now.

docs/docusaurus.config.js
```diff
@@ -11,6 +13,12 @@ module.exports = {
   trailingSlash: "false",
   organizationName: "zmkfirmware", // Usually your GitHub org/user name.
   projectName: "zmk", // Usually your repo name.
+  customFields: {
+    gitBranch:
+      process.env.BRANCH ||
```
caksoylar: Does this cover deploy previews by any chance?

petejohanson: AFAIK, the `BRANCH` will be populated by Netlify in that case, yes, but won't be `main` so our check will assume this *isn't* the development version...

caksoylar: That's probably OK, with the netlify banner at the bottom. But it'd be bonus if the banner indicated it's a deploy preview :)

petejohanson: I tweaked this to be the behavior basically any time we're not in a version branch, so it should default everywhere to using the "I'm showing a dev version" mode now most places.

docs/sidebars.js
petejohanson: Removed anywhere now that we moved things into the navbar.

docs/versions.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  releaseVersions: ["0.3"],
```
nmunnich: Could you add the bits to the pipeline to automatically update this when a new release happens?

nmunnich (commented):
I think this commit should be picked into our v0.3 branch as well.

docs/docusaurus.config.js
```diff
@@ -1,6 +1,34 @@
...
+  })),
+];
+
+const versionDropDownLabel = isDevelopmentVersion
+  ? versionNavbarItems[0].label
+  : versionNavbarItems.find((item) => {
+      return item.label === gitBranch.replace("-branch", "");
+    });
+
+console.error(versionDropDownLabel);
```
caksoylar: Errant debugging command?

petejohanson: Indeed. Removed.

docs/versions.js
petejohanson: I'm happy to do that in a follow up. I'm not sure it'll be trivial, and don't really want to block this on that automation

petejohanson: I did just tweak this, so that the version information is in a *JSON* file, which we should be able to automate appending to via something like: `jq '.releaseVersions[.releaseVersions | length] |= . + "0.4"' docs/zmk-release-versions.json`.

We'll want to do so *only* when we create a release that includes a new branch, e.g. `v0.4-branch` that would be deployed as well.

PR #3125, opened 2025-12-02
chore(tests): Write test builds to logfile

nmunnich:
* tempfile necessary because running build command makes directory pristine, interfering with the build log being written.
* build.log should be picked up by the test job artifact export step

### DIFF
--- a/app/run-test.sh
+++ b/app/run-test.sh
@@ -8,7 +8,6 @@
 #  ZMK_SRC_DIR:             Path to zmk/app (default is ./)
 #  ZMK_BUILD_DIR:           Path to build directory (default is $ZMK_SRC_DIR/build)
 #  ZMK_EXTRA_MODULES:       Path to at most one module (in addition to any in west.yml)
-#  ZMK_TESTS_VERBOSE:       Be more verbose
 #  ZMK_TESTS_AUTO_ACCEPT:   Replace snapshot files with new key events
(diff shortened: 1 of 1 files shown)


app/run-test.sh
```diff
@@ -42,13 +41,12 @@ build_cmd="west build ${ZMK_SRC_DIR:+-s $ZMK_SRC_DIR} -d ${ZMK_BUILD_DIR}/tests/
...
-fi
+build_log_tmp="${ZMK_BUILD_DIR}/tmp/$testcase/build.log"
+build_log="${ZMK_BUILD_DIR}/tests/$testcase/build.log"
+mkdir -p $(dirname $build_log_tmp)
+$build_cmd >"$build_log_tmp" 2>&1
+mv "$build_log_tmp" "$build_log"
+rmdir -p $(dirname $build_log_tmp) 2>/dev/null || true
 
 if [ $? -gt 0 ]; then
     echo "FAILED: $testcase did not build" | tee -a ${ZMK_BUILD_DIR}/tests/pass-fail.log
```
petejohanson: It would be nice to point callers to the location of the build log when this happens, so they can easily dig into what happened, e.g.:

```suggestion
    echo "FAILED: $testcase did not build (see ${build_log})" | tee -a ${ZMK_BUILD_DIR}/tests/pass-fail.log
```

nmunnich: Good idea, made that change.

PR #3129, opened 2025-12-07
feat(behaviors): Update hold-tap to be affected by sensors

nickconway:
Reopen of #1368. Allows a sensor to affect a hold-tap's decision the same way a key press would.


nmunnich:
Could you add some tests to go along with this?

PR #3136, opened 2025-12-13
docs: Switch to CLI for setup

nmunnich:
* setup script is broken, remove it
* docs is a merger of old and experimental cli pages

I think I removed the old setup script in its entirety, but I'm not 100% sure.


nmunnich:
Joel has a WIP coming up soon, so this isn't needed.

PR #3137, opened 2025-12-13
docs: Replace user setup script with ZMK CLI

joelspadin:
This deletes the Bash and PowerShell setup scripts in favor of using ZMK CLI. The old setup scripts are broken, since the config repo template uses ZMK version 0.3, while the setup scripts use hardware metadata from the main branch. ZMK CLI doesn't have this issue, because it clones the version of ZMK listed in the config repo to get hardware metadata.

Much of the "Installing ZMK" page has been rewritten to give instructions for installing ZMK CLI and using it to create and modify a config repo. The previously-hidden page on ZMK CLI has been added to the sidebar and repurposed into general documentation for the tool.


joelspadin:
> LGTM. I haven't kept up with the developments in the area but in the future we can consider using something like pyinstaller to reduce the dependencies, especially for Windows.

Looks like pyinstaller doesn't support any form of cross-compile, but if we can get GitHub actions to run it for each OS, and if we can get it to install into a standard location that's on the PATH variable, that is worth considering.

I also wonder if uv might be simpler to use. I'm not sure yet if uv can automatically manage installing Python, but if so, the Windows (11) install would be as simple as

```
winget install --id Git.Git -e
winget install --id astral-sh.uv -e
uv tool install zmk
```

> Also worth considering adding pointers to the CLI in the new shield guide.

Agreed, though I wanted to do that separately so this part could be done more quickly.

joelspadin:
Confirmed on another Windows 11 PC that never had Python installed: uv will automatically download the latest python version when you run `uv tool install zmk`. You do need to close and reopen the terminal after installing uv though.

The full install process would then be
```
winget install --id Git.Git -e
winget install --id astral-sh.uv -e
```
Restart terminal
```
uv tool install zmk
uv tool update-shell
```
Restart terminal
```
zmk init
```

caksoylar:
Do you want to update that? I like the `winget` approach and I don't have an issue recommending `uv`. You can also merge this and we can update later if you like.

joelspadin:
I think we can go with this for now and investigate uv later. Winget is only preinstalled on Windows 11, so if we do want to switch the documentation to use it, we'll want to either have a fallback for Windows 10 or to explicitly state that you're on your own if you're still using that.

We'll also want to decide if Linux and Mac instructions should use uv as well. I'm not sure what is the best way to install uv on other OSes.

PR #3140, opened 2025-12-13
feat(endpoints): add "no endpoint" value

joelspadin:
This adds ZMK_TRANSPORT_NONE, which can be set as the preferred endpoint transport if you wish to prevent the keyboard from sending any output. More usefully, it also is used to indicate that the preferred endpoint is not available and it could not fall back to an available one. To go along with this, many endpoint functions are renamed for consistency, and a few new functions are added:

- zmk_endpoint_get_preferred_transport() returns the value that was set with zmk_endpoint_set_preferred_transport().

- zmk_endpoint_get_preferred() returns the endpoint that will be used if it is available. This endpoint always has the same transport as zmk_endpoint_get_preferred_transport().

- zmk_endpoint_is_connected() is a shortcut to check if the keyboard is actually connected to an endpoint.

This change is based on #2572 but without the option to disable endpoint fallback. It does refactor code to allow adding that feature later.

Since this changes the meanings of `enum zmk_transport` values that are saved to settings, the setting for storing the preferred transport is renamed. If the old setting exists, it is interpreted using the old enum values and translated to the new ones. The old setting is then deleted, and the upgraded value is written to the new setting immediately.

### DIFF
--- a/app/src/endpoints.c
+++ b/app/src/endpoints.c
@@ -38,10 +38,18 @@ LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 // Full name of the deprecated setting which stored preferred_transport with an older type.
 #define SETTING_PREFERRED_TRANSPORT_V1 SETTING_SUBTREE "/" SETTING_PREFERRED_TRANSPORT_V1_KEY
 
+#if IS_ENABLED(CONFIG_ZMK_USB)
+#define DEFAULT_TRANSPORT ZMK_TRANSPORT_USB
+#elif IS_ENABLED(CONFIG_ZMK_BLE)
+#define DEFAULT_TRANSPORT ZMK_TRANSPORT_BLE
+#else
+#define DEFAULT_TRANSPORT ZMK_TRANSPORT_NONE
+#endif
+
 static struct zmk_endpoint_instance current_instance = {};
 
 // Transport to use if multiple endpoints are ready
-static enum zmk_transport preferred_transport = ZMK_TRANSPORT_USB;
+static enum zmk_transport preferred_transport = DEFAULT_TRANSPORT;
 
 static void update_current_endpoint(void);
 


app/tests/ble/split/set-hid-indicators/snapshot.log
```diff
@@ -9,4 +9,5 @@ peripheral 0 <inf> zmk: Welcome to ZMK!
 peripheral 0 <dbg> zmk: security_changed: Security changed: FD:9E:B2:48:47:39 (random) level 2
 peripheral 0 <dbg> zmk: split_svc_pos_state_ccc: value 1
 peripheral 0 <dbg> zmk: split_svc_select_phys_layout_callback: Selecting physical layout after GATT write of 0
+peripheral 0 <dbg> zmk: split_svc_update_indicators_callback: Raising HID indicators changed event: 0
```
joelspadin, line 12: I'm not really sure why this changed, but `split_svc_update_indicators_callback` is running once before `zmk_hid_indicators_process_report` now.

app/src/endpoints.c
```diff
@@ -282,41 +322,40 @@ static bool is_ble_ready(void) {
...
+        return ZMK_TRANSPORT_NONE;
+
+    case ZMK_TRANSPORT_USB:
         if (is_usb_ready()) {
-            LOG_DBG("Both endpoint transports are ready. Using %d", preferred_transport);
-            return preferred_transport;
+            LOG_DBG("USB is preferred and ready");
+            return ZMK_TRANSPORT_USB;
         }
+        if (is_ble_ready()) {
```
joelspadin, line 435: To implement the "no fallback" option later, we would simply disable this block and the similar one in the BLE case.

nmunnich:
I definitely agree with the renaming of the functions, and the introduction of the new functions. Could you perhaps separate that out into a different commit, though? Would make it much easier to review imo.

app/include/zmk/endpoints_types.h
```diff
@@ -10,6 +10,7 @@
  * The method by which data is sent.
  */
 enum zmk_transport {
+    ZMK_TRANSPORT_NONE,
```
petejohanson: I can't easily check from my phone, but IIRC this enum value gets saved to settings when a user user sets a preferred endpoint with the behavior, so this new value should get added to the end of this enum if so.

petejohanson (changes requested):
On a whole this looks good. One minor comment from my first pass at this.

app/include/zmk/endpoints_types.h
joelspadin: I opted to keep this as value 0, because "none" is the most reasonable value for anything that is implicitly initialized to 0. Instead, I changed the setting name and added an upgrade routine if the old setting is present.

Tested on my nrf5340dk. With "endpoints/preferred" set to 1, `endpoint_settings_load_preferred_v1()` gets called, translates the value to 2, deletes the old setting, and writes to "endpoints/preferred2". (This also resulted in `endpoint_settings_load_preferred_v2()` being called, but I wasn't sure if I could rely on that happening, so I made sure `preferred_transport` is set in both cases.) On a subsequent power cycle, the v2 function is called and the v1 function is not.

snoyer: `NONE == 0` makes sense and I get why changing the enum value is bad in theory but what would be the actual consequences in this case it you just did it without adding the whole migration system?

Are we talking about bricking keyboards? or just some one-off inconvenience such as shutting down having saved `USB -> 0` and rebooting loading `0 -> NONE` so we need to manually press `&out OUT_xxx` the first time around before being able to type again (or need to flash `settings_reset` if there's no output selection at all in the keymap)?

joelspadin: Everyone who had set their keyboard to BLE would find their keyboard now preferring USB.

Everyone who had set their keyboard to USB would find their keyboard nonfunctional unless they happened to have `&out ...` on their keyboard and pressed it, because it would switch from preferring USB to preferring nothing. They would need to add this binding or do a full settings reset.

joelspadin: (You are correct about the behavior for if USB was preferred, but I would hardly consider that to be *just* an inconvenience. That's basically signing us up for a lot of people asking for help on Discord.)

snoyer: > but I would hardly consider that to be _just_ an inconvenience. That's basically signing us up for a lot of people asking for help on Discord.

That's true, maybe I'm personally too tolerant to having to fiddle with the `&out` keys :)

app/src/endpoints.c
```diff
@@ -15,26 +15,40 @@
...
+#define SETTING_PREFERRED_TRANSPORT_V1_KEY "preferred"
+// Full name of the deprecated setting which stored preferred_transport with an older type.
+#define SETTING_PREFERRED_TRANSPORT_V1 SETTING_SUBTREE "/" SETTING_PREFERRED_TRANSPORT_V1_KEY
 
 static struct zmk_endpoint_instance current_instance = {};
-static enum zmk_transport preferred_transport =
-    ZMK_TRANSPORT_USB; /* Used if multiple endpoints are ready */
+
+// Transport to use if multiple endpoints are ready
+static enum zmk_transport preferred_transport = ZMK_TRANSPORT_USB;
```
petejohanson: Let's take a moment to fix this up, and adjust the preferred based on `CONFIG_ZMK_USB` being enabled or not... With a few upcoming supported SoCs, they've got BLE but no USB, so we should gracefully handle that case where only `CONFIG_ZMK_BLE` is enabled as a HID/endpoint transport.

petejohanson (changes requested):
LGTM, with one minor tweak we should make around preferred if ZMK_USB isn't set, since we're fixing this up anyways. Thanks for working on this!

app/src/endpoints.c
joelspadin: I don't think this actually matters right now, since `preferred_transport` only gets used when USB and BLE are both connected, but I could see that becoming more problematic if new endpoints are added later.

joelspadin: Done.

petejohanson (approved):

PR #3145, opened 2025-12-15
refactor(core): Adjust our approach for upstream Zephyr boards

petejohanson:
Move to using proper HWMv2 board extensions
https://docs.zephyrproject.org/4.1.0/hardware/porting/board_porting.html#board-extensions

for extending upstream Zephyr boards for use with ZMK. With this change, using upstream Zephyr board IDs directly, e.g. `seeeduino_xiao` will be stock versions as found upstream. To use a board variant that is tuned for ZMK use, use the `zmk` variant, e.g. `seeeduino_xiao//zmk` which is shorthand for `seeeduino_xiao/samd21g18a/zmk`.

I've updated a few docs as part of this to note the variant change:

* https://deploy-preview-3145--zmk.netlify.app/blog/2025/12/09/zephyr-4-1#zmk-board-variant
* https://deploy-preview-3145--zmk.netlify.app/docs/hardware

## PR check-list


nmunnich:
Re: discussion on user clarity: do you think it might make sense to introduce //zmk variants for our in-tree boards which are also interconnects, and use the base version as something that matches what Zephyr would accept?

So we would have 
- boards which are also keyboards, name without //zmk
- interconnect boards without tweaks, regardless of being in Zephyr or ZMK name without //zmk
- interconnect boards with tweaks, name always with //zmk

petejohanson:
> Re: discussion on user clarity: do you think it might make sense to introduce //zmk variants for our in-tree boards which are also interconnects, and use the base version as something that matches what Zephyr would accept?


I have been debating this internally for quite some time now. There guiding principles, as I see them, for this discussion, should be:

* Clean expectation setting for users of ZMK, that is as intuitive as possible, and as well documented/supported by tooling as possible where things are less intuitive.
* Promote flexibility and re-use (e.g. right now it's hard to build a Zephyr sample for something like the nice!60, since the ZMK specific headers, etc cause problems) that allows boards to be used very flexibly, we can easily use upstream boards for various users without random things dragged in that break things, etc.
* Minimizing breakage, or at least consolidating it into one release (we're already going to have this with the planned v0.4, so further refactors in `main` to satisfy the first two bullets are worth it *now*, IMHO)

> 
> So we would have
> 
>     * boards which are also keyboards, name without //zmk

So... If we do this, we perpetuate the existing challenges here, where you can't flash basic Zephyr samples to our "keyboard boards" from ZMK, without hacking apart the .dts files temporarily. Not a *huge* deal, but if we're putting in this effort, it might be worth improving this. I find this useful for targetted "get this display working" or "test the RGB" or "verify this sensor".

> 
>     * interconnect boards without tweaks, regardless of being in Zephyr or ZMK name without //zmk

Define "tweaks" here. At a minimum, we usually have the following changes that differ from upstream:

* Enabling one or more HID transports.
* Defining settings storage (*many* upstream boards do this, but not all, and many of those do so with an MCUBoot style partition setup) in devicetree and enabling related Kconfig OOTB for those boards.
* (Newer) Bootloader/retained boot mode setup by default.

> 
>     * interconnect boards with tweaks, name always with //zmk

If we do this, I think we may introduce confusing for users, surrounding when and when not to use a `zmk` variant specifier.

Appreciate you helping brainstorm and nail this down!

nmunnich:
I 100% agree with the guiding principles.

> So... If we do this, we perpetuate the existing challenges here, where you can't flash basic Zephyr samples to our "keyboard boards" from ZMK, without hacking apart the .dts files temporarily. Not a huge deal, but if we're putting in this effort, it might be worth improving this. I find this useful for targetted "get this display working" or "test the RGB" or "verify this sensor".

I could certainly be convinced to have //zmk variants for all of our boards, so that we *could* flash Zephyr samples to them. My concern here is mainly around out-of-tree boards. I think for boards which are also interconnects, we can generally be convinced to add them in-tree of ZMK, if Zephyr does not accept them already, so my expectation is that out-of-tree boards are mostly just keyboards. Considering how much chaos and inconsistency there already is around out-of-tree boards, I see some value in keeping the expected definition for such boards simpler. I could really go either way, though.

> Define "tweaks" here.

Essentially what you listed, though we also usually disable the UART node. I see an extra benefit of the separation as making it much easier to upstream boards to Zephyr since the definition has already been written, if a board gains popularity/prominence enough that it would be accepted.

> If we do this, I think we may introduce confusing for users, surrounding when and when not to use a zmk variant specifier.

I actually see the opposite: End users would be taught "If you have a board+shield combo, your board always has a //zmk at the end", without needing to look things up. The finer details would be more for advanced users.

petejohanson:
> I 100% agree with the guiding principles.
> 
> > So... If we do this, we perpetuate the existing challenges here, where you can't flash basic Zephyr samples to our "keyboard boards" from ZMK, without hacking apart the .dts files temporarily. Not a huge deal, but if we're putting in this effort, it might be worth improving this. I find this useful for targetted "get this display working" or "test the RGB" or "verify this sensor".
> 
> I could certainly be convinced to have //zmk variants for all of our boards, so that we _could_ flash Zephyr samples to them. My concern here is mainly around out-of-tree boards. I think for boards which are also interconnects, we can generally be convinced to add them in-tree of ZMK, if Zephyr does not accept them already, so my expectation is that out-of-tree boards are mostly just keyboards. Considering how much chaos and inconsistency there already is around out-of-tree boards, I see some value in keeping the expected definition for such boards simpler. I could really go either way, though.

Having tried this just the past couple days while playing with my stm32c0pad prototype, I can definitely say that having a "Zephyr compatible" based board for it was really helpful, and that adding a `zmk` variant was not a significant additional effort.

Not that I think we should expect everyone to do this with every board definition, but it might be nice to document it as an *option* for folks to follow.

Not sure how best to communicate this to users though? For sure, listing the `//zmk` variant in the .zmk.yml files is a good first start, with integration with the ZMK CLI helping here.
> 
> > Define "tweaks" here.
> 
> Essentially what you listed, though we also usually disable the UART node. I see an extra benefit of the separation as making it much easier to upstream boards to Zephyr since the definition has already been written, if a board gains popularity/prominence enough that it would be accepted.

:+1: 

> 
> > If we do this, I think we may introduce confusing for users, surrounding when and when not to use a zmk variant specifier.
> 
> I actually see the opposite: End users would be taught "If you have a board+shield combo, your board always has a //zmk at the end", without needing to look things up. The finer details would be more for advanced users.

I think my only real concern here, is where we draw the line for "accepting boards that support some interconnect", if we want to have that really be consistent/enforced.... The [lemon controllers](https://ryanis.cool/cosmos/lemon/) ? Halcyon? How about the tenth random RP2040 board we theoretically support? I'd love to have some planning around what that really looks like.

I do think that your proposed "user expectation" *does* seem like it would be nice to get to. How do we do so?

nmunnich:
> Not that I think we should expect everyone to do this with every board definition, but it might be nice to document it as an option for folks to follow.
> 
> Not sure how best to communicate this to users though? For sure, listing the //zmk variant in the .zmk.yml files is a good first start, with integration with the ZMK CLI helping here.
>
> I do think that your proposed "user expectation" does seem like it would be nice to get to. How do we do so?

Proposed plan:

- All in-tree boards get a //zmk variant
- We introduce a Kconfig flag that gets set in the variants (`CONFIG_ZMK`?). During build time we have the ZMK app check for this flag and throw a warning in the build process, something like "You are building ZMK firmware with a board that is missing ZMK-specific tweaks. You may experience some unexpected behavior. Have you checked if there is a <your_board>//zmk variant of your board?"
- The blog post and docs are updated to have //zmk everywhere, and we add a "patch note" to the blog post
- Probably we should adjust our supported hardware script to list boards which don't have in-tree shields as well.
- We add a "New board" page under hardware integration, which essentially states "Go to Zephyr and follow their instructions *exactly* to create a Zephyr board. <before continuing, here are some helpful Zephyr sample apps that you can flash to your board to ensure things are working at this point> Then return here for instructions on making a //zmk variant of your board with ZMK-specific tweaks allowing for Studio etc. If your board is a keyboard, once you've made the //zmk variant, take guidance (but don't follow exactly) the steps in the new shield guide to add a kscan etc." 

> I think my only real concern here, is where we draw the line for "accepting boards that support some interconnect", if we want to have that really be consistent/enforced.... The [lemon controllers](https://ryanis.cool/cosmos/lemon/) ? Halcyon? How about the tenth random RP2040 board we theoretically support? I'd love to have some planning around what that really looks like.

I edited this reply to open #3172 for this discussion instead, for better organisation.

petejohanson:
> Proposed plan:
> 
>     * All in-tree boards get a //zmk variant

Including "complete keyboards", that only ever are used with ZMK? We refactor so they can be used with Zephyr directly with the "stock" variant?

caksoylar:
It seems that (some?) board extensions are currently broken on main:
```sh
> west build -b sparkfun_pro_micro_rp2040 -d build/rp2040-test -- -DSHIELD=naked60
> grep CONFIG_ZMK_USB build/rp2040-test/zephyr/.config
# CONFIG_ZMK_USB is not set
```
@carrefinho also reported that board extensions are not applied on all boards whose name starts with vendor name, like `adafruit_kb2040`, `boardsource_blok`.

Do we want to fix that or get this in quicker?

petejohanson:
> It seems that (some?) board extensions are currently broken on main:
> 
> ```shell
> > west build -b sparkfun_pro_micro_rp2040 -d build/rp2040-test -- -DSHIELD=naked60
> > grep CONFIG_ZMK_USB build/rp2040-test/zephyr/.config
> # CONFIG_ZMK_USB is not set
> ```
> 
> @carrefinho also reported that board extensions are not applied on all boards whose name starts with vendor name, like `adafruit_kb2040`, `boardsource_blok`.
> 
> Do we want to fix that or get this in quicker?

We'll fix that here. Anyone with issues can pin to v0.3 in the meantime.

petejohanson:
Ok, I've just pushed fixes for the nRFMicro to create a "vanilla" board definition for it in the `app/module/` directory, that's completely ZMK-free, and then use the same HWMv2 extension approach to add ZMK variants.

With this. you can now do:

```
west build -p -d build/nrfmicro_hello_world -b nrfmicro/nrf52840 samples/hello_world -- -DBOARD_ROOT=/home/peter/git/zmk/app/module
```

from within the Zephyr repo. I'm not *sure* that this whole process is completely necessary for every in-tree *keyboard* board, e.g. nice!60, I think there we would *only* have the `/zmk` variant for those living in their existing location in `app/boards/`.

.github/workflows/build.yml
```diff
@@ -161,9 +161,14 @@ jobs:
...
             const combinedUnique = [...new Map(combined.map(el => [JSON.stringify(el), el])).values()];
 
+            console.log(combinedUnique);
+
             const perBoard = {};
+            idx = 0;
 
             for (const configuration of combinedUnique) {
+              console.log(idx++);
+              console.log(configuration);
```
joelspadin: Leftover debug code

app/boards/adafruit/kb2040/adafruit_kb2040_zmk.dts
```diff
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: MIT
  */
 
+#include <../boards/adafruit/kb2040/adafruit_kb2040.dts>
```
joelspadin, line 7: What directory are all these includes relative to?

app/boards/joric/nrfmicro/board.yml
```diff
@@ -1,14 +1,9 @@
...
-    revisions:
-      - name: 1.3.0
-      - name: 1.1.0
+  extend: nrfmicro
+  variants:
+    - name: zmk
+      qualifier: nrf52840
+    - name: flipped_zmk
+      qualifier: nrf52840
+    - name: zmk
```
joelspadin, line 8: Is having two variants with the same name okay?

app/module/boards/keycapsss/puchi_ble/puchi_ble_defconfig
joelspadin, line 1: On most of the other boards, the ZMK-specific configs were pulled out into a board extension. Was putting these in the base board definition here intentional?

app/boards/adafruit/kb2040/board.yml
nmunnich, line 1: Building this one with the tester I get 
```
CMake Warning at /workspaces/zmk/zephyr/cmake/modules/dts.cmake:425 (message):
  dtc raised one or more warnings:

  /workspaces/zmk/app/build/zephyr/zephyr.dts:39.26-42.5: Warning
  (simple_bus_reg): /soc/memory@20000000: simple-bus unit address format
  error, expected "20000004"
```

Doesn't occur on the non //zmk version. No idea about the effects, haven't delved deep at all. Seems to apply to some of the others as well

.github/workflows/build.yml
petejohanson: Indeed. Removed. Thanks.

app/boards/adafruit/kb2040/adafruit_kb2040_zmk.dts
petejohanson: So... This is an annoying hack, which is the only way I could find to do this, because these are processed from some subdirectory of the Zephyr repo, and doing this weird relative import of the system path is how I could best find to include the original board files from the Zephyr board definition of the board.

app/boards/joric/nrfmicro/board.yml
petejohanson: Yes, when a board has multiple possible SoCs, you can no longer do the shorthand that omits the SoC portion of the qualifier, so instead of `nrfmicro//zmk` you *must* do `nrfmicro/nrf52840/zmk` or `nrfmicro/nrf52833/zmk` depending on which you want to target.

app/module/boards/keycapsss/puchi_ble/puchi_ble_defconfig
petejohanson: Ah, no. Oversight. Will fix. Thanks, good catch.

app/boards/adafruit/kb2040/board.yml
petejohanson: Yes, that's a side effect of the ZMK variants doing the setup for retained boot mode, which carves out some of the memory for noinit storage of the boot mode, triggering this error. It's not fatal, but I can look into use `pre_dt_board.cmake` tweaks if we really want.

app/core-coverage.yml
nmunnich, line 1: do the boards at the start of the file not also need to be zmk variants?

app/boards/weact/blackpill_f411ce/blackpill_f411ce_zmk.dts
```diff
@@ -1,9 +1,10 @@
 /*
- * Copyright (c) 2022 The ZMK Contributors
+ * Copyright (c) 2024 The ZMK Contributors
```
nmunnich: if we're going to number bump we should probably bump it up more

app/boards/weact/blackpill_f401ce/blackpill_f401ce_zmk.dts
```diff
@@ -1,9 +1,10 @@
 /*
- * Copyright (c) 2022 The ZMK Contributors
+ * Copyright (c) 2024 The ZMK Contributors
```
nmunnich: same

app/boards/extensions/xiao_ble/xiao_ble.conf
```diff
@@ -1,19 +1,29 @@
 
-CONFIG_CONSOLE=n
-CONFIG_SERIAL=n
-CONFIG_UART_CONSOLE=n
-CONFIG_UART_INTERRUPT_DRIVEN=n
```
nmunnich, line 5: Why are we re-enabling these? Does Zephyr now have them disabled by default?

app/boards/raspberrypi/rpi_pico/rpi_pico_rp2040_zmk.dts
```diff
@@ -1,3 +1,10 @@
+/*
+ * Copyright (c) 2025 The ZMK Contributors
```
nmunnich: 2026?

app/boards/nordic/nrf52840dongle/nrf52840dongle_nrf52840_zmk_defconfig
nmunnich, line 1: Do we need adjustments to console? probably not because it's a dongle?

app/boards/nordic/nrf52840dongle/nrf52840dongle_nrf52840_zmk.dts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2025 The ZMK Contributors
```
nmunnich: 2026?

app/boards/adafruit/kb2040/board.yml
nmunnich: Probably would be nice just to avoid all the help requests we'll likely be getting.

docs/blog/2025-12-09-zephyr-4-1.md
```diff
@@ -61,6 +61,12 @@ The following steps will get you building ZMK locally against Zephyr 4.1:
 
 From there, you should be ready to build as normal!
 
+## ZMK Board Variant
+
+The new Zephyr version includes a [standard mechanism for extending boards](https://docs.zephyrproject.org/4.1.0/hardware/porting/board_porting.html#board-extensions) found in Zephyr (like the Seeeduino XIAO, or Raspberry Pi Pico) with new variants that include additional configuration/defaults that are relevant to the application. This also means the original "stock" board can still be used as is, if consumers don't want to use those customizations for any reason.
+
+As a result, all board definitions found in the ZMK tree now must be used with a `zmk` variant. For example, the previous board ID of `adv360pro_left` would now be `adv360pro_left/nrf52840/zmk`, which can be elided to `adv360pro_left//zmk` since there's only one SoC on the board.
```
nmunnich: Dislike "elided" because it's a fairly uncommon word that non-native speakers most likely will be unfamiliar with, even if the meaning can be guessed. "shortened"?

docs/blog/2025-12-09-zephyr-4-1.md
```diff
@@ -429,5 +435,13 @@ Once a ZMK version based on Zephyr 4.1 is released, we'll be working towards upd
 - Removing ZMK's use of the deprecated, and now removed, `kscan` APIs in favor of the newer matrix input API. ZMK already supports the matrix input API, but has not yet converted our existing drivers to that API. Advanced users looking to test can try leveraging the upstream Zephyr drivers today to test that functionality, but that is _not_ officially supported and is likely to have bugs or untested side effects.
 - Move the new USB stack, that includes better High-Speed USB support.
 
+:::note
```
nmunnich: I think I'd prefer to have this note at/near the top, rather than the bottom.

nmunnich (commented):
Are we leaving the "base" versions of full keyboards for later?

There are some references to `<shield>/boards/nice_nano.overlay` in the docs which could use renaming.

My comments on the serial bits and copyright years will apply to lots of places, putting them everywhere would be unhelpful I think.

docs/blog/2025-12-09-zephyr-4-1.md
petejohanson: Fair. Moved.

app/core-coverage.yml
petejohanson: I'm sorry, which boards specifically?

app/boards/adafruit/kb2040/board.yml
petejohanson: Yeah, let me see how this could be tweaked.

app/core-coverage.yml
nmunnich: Nevermind, going commit by commit failed me on this file 😅 All good!

app/boards/nordic/nrf52840dongle/nrf52840dongle_nrf52840_zmk.dts
petejohanson: Fixed up all the copyright years I could find.

petejohanson:
> Are we leaving the "base" versions of full keyboards for later?

Are you talking about not having non-ZMK variants for the onboard boards?

app/boards/extensions/xiao_ble/xiao_ble.conf
petejohanson: We're not. We're now:

* Leaving the original `xiao_ble` board definition untouched/changed.
* Adding a new `xiao_ble/nrf52840/zmk` variant, and that doesn't inherit the `xiao_ble_defconfig`, but has it's own, authoritative `xiao_ble_nrf52840_defconfig` file, which doesn't inherit from anything, so it doesn't have the serial/uart settings enabled simply by not having them included in that file.

petejohanson:
Unfortunately, it doesn't appear that adding a `pre_dt_board.cmake` file in a board *extension* directory gets picked up, so we're out of luck squashing those DTC warnings, AFAICT.

app/boards/nordic/nrf52840dongle/nrf52840dongle_nrf52840_zmk_defconfig
petejohanson: I'm not sure what adjustments you would expect? See my other comment here, which also applies here: https://github.com/zmkfirmware/zmk/pull/3145#discussion_r2750447340 when it comes to this no longer "inheriting" the defconfig file for the bare/non-variant board definition

nmunnich: Resolved by your other comment, yes.

nmunnich:
I was indeed talking about non-ZMK versions of keyboards which are boards.

Unfortunate that the warnings couldn't be squashed. Could we open an issue with Zephyr to resolve that? Obviously not worth waiting for a fix from them, but I wouldn't want the warnings long-term.

petejohanson:
> I was indeed talking about non-ZMK versions of keyboards which are boards.

Tbh. This is lower priority for me. If the vendor/designer of a given board that's in tree wants to do it, they can submit a PR modeled after this one.

> 
> Unfortunate that the warnings couldn't be squashed. Could we open an issue with Zephyr to resolve that? Obviously not worth waiting for a fix from them, but I wouldn't want the warnings long-term.

Yeah. Fixing this just requires me to tweak how the included boot mode retention modified the RAM, which will squash that error. Will look at putting that together.

nmunnich (approved):
Well, minus the warning everything else looks fine to me, so I'll drop an approve and leave it up to you if you want to sort the warning now or later.

peterjc:
I mentioned this on Discord, but repeating here as it is more public: The blog post https://zmk.dev/blog/2025/12/09/zephyr-4-1 currently covers updating board names like `nice_nano_v2` in v0.3 -> `nice_nano@2.0.0` (or similar), but not this new change -> `nice_nano@2.0.0//zmk` (or similar).

I'd suggest updating the December 2025 blog post, or a second post for the Feb 2026 breaking changes (here in #3145). A second post would be best if any further changes from v0.3 are expected soon?

petejohanson:
> I mentioned this on Discord, but repeating here as it is more public: The blog post https://zmk.dev/blog/2025/12/09/zephyr-4-1 currently covers updating board names like `nice_nano_v2` in v0.3 -> `nice_nano@2.0.0` (or similar), but not this new change -> `nice_nano@2.0.0//zmk` (or similar).
> 
> I'd suggest updating the December 2025 blog post, or a second post for the Feb 2026 breaking changes (here in #3145). A second post would be best if any further changes from v0.3 are expected soon?

This is now fixed in the original blog post, via #3247

PeterMortensen:
### Broken links

The `deploy-preview-3145--zmk.netlify.app` links are broken (404). Would it be something like [this](https://zmk.dev/blog/2025/12/09/zephyr-4-1#zmk-board-variant) and [this](https://zmk.dev/docs/hardware)?

caksoylar:
The deploy previews are deployed from the PR branch (3145 is the github PR number, in this case). Since the PR is merged, you can instead check the paths in the main docs like the ones you shared, yes.

PR #3154, opened 2025-12-20
feat(shields): Add Studio support to boardsource3x4

snoyer:
I added matrix transforms for both horizontal (MCU on top) and vertical (MCU on the left) positioning, and corresponding physical layouts. I put `&studio_unlock` on a free key on the firmware layer but I'm open to other suggestions.

Both the default keymap and a custom vertical keymap seem to work and are editable with Studio, however I'm not sure how (or to what extent) switching layouts at runtime is supposed to be working. Is there something wrong with my changes or is support for this limited to begin with? 

## PR check-list

### DIFF
--- a/app/boards/shields/boardsource3x4/boardsource3x4-layouts.dtsi
+++ b/app/boards/shields/boardsource3x4/boardsource3x4-layouts.dtsi
@@ -46,4 +46,25 @@
             , <&key_physical_attrs 100 100    200 300    0 0 0>
             ;
     };
+
+    position_map {
+        compatible = "zmk,physical-layout-position-map";
+
+        horizontal_map: horizontal {
+            // A  7  8  9
+            // B  4  5  6
+            // C  1  2  3
+            physical-layout = <&horizontal_physical_layout>;
+            positions = </* 0 */ 1 2 3 4 5 6 7 8 /* 9 10 11 */>;
+        };
+        vertical_map: verctical {
+            // A  B  C
+            // 7  8  9
+            // 4  5  6
+            // 1  2  3
+            physical-layout = <&vertical_physical_layout>;
+            positions = </* 0 */ 3 4 5 1 6 7 8 2 /* 9 10 11 */>;
+        };
+    };
+
 };


app/boards/shields/boardsource3x4/boardsource3x4.overlay
```diff
@@ -1,14 +1,47 @@
...
+    transform = <&vertical_transform>;
+};
 
 / {
     chosen {
         zmk,kscan = &kscan0;
+        zmk,physical-layout = &horizontal_physical_layout;
+    };
+
+    default_transform: default_transform {
```
snoyer: should this be `default_transform` for some kind of compatibility/convention reason or should I make it `horizontal_transform` to be more explicit?

nmunnich: There is no compatibility reason here. I would go with the horizontal naming.

nmunnich:
Thanks for working on this. To enable full support when swapping layouts at runtime, you'll need to write a position map: https://zmk.dev/docs/development/hardware-integration/physical-layouts#position-map

caksoylar:
I see this for the position map:

image

It makes more sense to me to keep the a 3x3 square the same, then remap the extra row to column or vice versa. The square can be rotated or kept the same in the mapping. Was the above mapping intentional?

snoyer:
@caksoylar what tool is that visualization from? and should I be using it to create the position map in the first place? 

My intended mapping was:
```c
            // A  7  8  9
            // B  4  5  6
            // C  1  2  3
```
```c
            // A  B  C
            // 7  8  9
            // 4  5  6
            // 1  2  3
```
and this is what I see in zmk.studio:
image
image

caksoylar:
Ah OK, that mapping makes sense to me now. The screenshot was from https://zmk-layout-helper.netlify.app/, which we link from the docs for position maps.

nmunnich:
Thanks!

snoyer:
Thank you both for the pointers :)

PR #3159, opened 2025-12-22
chore(ci): Add warning about being on main branch when build fails

nmunnich:
Add an extra step to our build job which puts a warning about pinning your ZMK version if you are building off of main and the build fails.

See https://github.com/nmunnich/zmk-config-akohekohe/actions/runs/20440843682 (I switched the == to a != for this workflow run) and https://github.com/nmunnich/zmk-config-akohekohe/actions/runs/20441867277 (run with the current version).

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -122,6 +122,13 @@ jobs:
         working-directory: ${{ env.base_dir }}
         run: west update --fetch-opt=--filter=tree:0
 
+      - name: Check ZMK revision
+        working-directory: ${{ env.base_dir }}
+        run: |
+          zmk_revision=$(west list -f "{revision}" zmk)
+          echo "zmk_revision=${zmk_revision}" >> $GITHUB_ENV
+          echo "ZMK revision: ${zmk_revision}"
+
       - name: West Zephyr export
         working-directory: ${{ env.base_dir }}
         run: west zephyr-export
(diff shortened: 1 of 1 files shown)


.github/workflows/build-user-config.yml
```diff
@@ -131,6 +131,15 @@ jobs:
         shell: sh -x {0}
         run: west build -s zmk/app -d "${{ env.build_dir }}" -b "${{ matrix.board }}" ${{ env.extra_west_args }} -- -DZMK_CONFIG=${{ env.base_dir }}/${{ inputs.config_path }} ${{ env.extra_cmake_args }} ${{ matrix.cmake-args }}
 
+      - name: Warn about building from main if build fails
+        if: failure() && github.event.workflow.ref == 'refs/heads/main'
```
petejohanson: This isn't actually checking for ZMK main in the west.yml file. Afaict.

petejohanson: You can do `west list -f "{revision}" zmk` to get the info easily.

.github/workflows/build-user-config.yml
```diff
@@ -131,6 +138,22 @@ jobs:
...
+        if: failure() && github.event.workflow.ref == 'refs/heads/main' && env.zmk_revision == 'main'
+        run: |
+          echo "# Consider Pinning ZMK" >> $GITHUB_STEP_SUMMARY
+          echo "Your recent build failure might be the result of breaking changes made to ZMK's main branch." >> $GITHUB_STEP_SUMMARY
+          echo "Consider [pinning your ZMK version](https://zmk.dev/blog/2025/06/20/pinned-zmk) to a release for increased stability." >> $GITHUB_STEP_SUMMARY
+          echo "See also the [list of released versions](https://github.com/zmkfirmware/zmk/releases)." >> $GITHUB_STEP_SUMMARY
+          echo "If you wish to stay on main, check the most recent pending release PR for breaking changes. [Our blog](https://zmk.dev/blog) may have upgrade information if breaking changes are significant." >> $GITHUB_STEP_SUMMARY
+
+      - name: Warn about ZMK revision mismatch
+        if: always() && env.zmk_revision != github.ref_name
```
petejohanson: I'm really confused why we specifically care if someone has a config branch named "foo-bar" and are using ZMK v0.3?

For both of these checks, I don't see any reason to care at all about what branch naming is used within the user's config repo.

nmunnich: I typod this, was meant to check if the branch used for the workflow differs from the branch used in west.

petejohanson: Is there actually a way to know the revision of a workflow that was included from elsewhere?

nmunnich: I thought so, but suddenly I'm not so sure. Let me doublecheck with more experimenting. Setting to draft for now, will ping on Discord.

nmunnich: Yeah, I think I was fooled by something I misread. Removed this. Could probably check the workflow file itself in the branch, but not worth the effort.

petejohanson:
Lgtm. Feel free to merge if you're happy with it.

PR #3163, opened 2025-12-23
fix(core): support magic bootloader values for nRF52833

krikun98:
Enable ZMK_BOOTMODE_MAGIC_VALUE_BOOTLOADER_TYPE_ADAFRUIT_NRF52 for nRF52833, add nrf52833_uf2_boot_mode.dtsi as a copy of the nRF52840 file for easy inclusion.

Enable retention for all nrfmicro subvariants (as this is the only in-tree board that uses the nRF52833).

This fixes &bootloader for nRF52833 boards post-4.1 upgrade.

## PR check-list


krikun98:
I could also just rename the `nrf52840_uf2_boot_mode.dtsi` to `nrf52_uf2_boot_mode`, but that would entail more changes in existing files.

nmunnich:
I think I would be in favor of the renaming solution personally. We've not released yet after all.

petejohanson:
> I think I would be in favor of the renaming solution personally. We've not released yet after all.

In case there are *future* differences, I merged this as is, I don't mind this being explicit for each target at this point.

PR #3164, opened 2025-12-23
chore: Add basic tests for Studio's layer manipulation

nmunnich:
We currently don't have any test coverage for the layer manipulation that Studio can perform.

Use a bit of a hacky workaround by creating test behaviors for the methods that Studio calls to manipulate layers. Test behaviors are only compiled when we set a Kconfig flag that is set during the test script.

Then use these hacky test behaviors to test some basic interactions. More tests should probably be added, but I want to verify that we're ok with this approach to testing non-behavior functionality first.


app/module/test-behaviors/Kconfig
```diff
@@ -0,0 +1,3 @@
+config ZMK_TEST_BEHAVIORS
+    bool "Include behaviors used for testing purposes"
+    default n
```
petejohanson: Superfluous, this is the default, and adding it can actually complicate some scenarios. See https://docs.zephyrproject.org/latest/build/kconfig/tips.html#redundant-defaults

app/module/test-behaviors/behavior_add_layer.c
```diff
@@ -0,0 +1,46 @@
...
+                                        struct zmk_behavior_binding_event event) {
+    int new_layer = zmk_keymap_add_layer();
+    switch (new_layer) {
+    case -ENOSPC:
+        LOG_ERR("No more layers can be added");
+        return -ENOSPC;
+    default:
+        LOG_DBG("Added layer %d", new_layer);
+        return 0;
+    }
```
petejohanson, line 34: To be future proof, this should really only consider it success if `new_layer >= 0`, in case other errno start getting returned from the called API down the road.

app/module/test-behaviors/behavior_remove_layer.c
```diff
@@ -0,0 +1,46 @@
...
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+#if (IS_ENABLED(CONFIG_ZMK_KEYMAP_LAYER_REORDERING)) && (DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT))
+
+static int on_remove_layer_binding_pressed(struct zmk_behavior_binding *binding,
+                                           struct zmk_behavior_binding_event event) {
+    int result = zmk_keymap_remove_layer(binding->param1);
+
+    switch (result) {
+    case -EINVAL:
```
petejohanson, line 28: Same.

app/run-test.sh
```diff
@@ -38,9 +38,16 @@ testcase=$(realpath $path | sed -n -e "s|.*/tests/||p")
 echo "Running $testcase:"
 
 build_cmd="west build ${ZMK_SRC_DIR:+-s $ZMK_SRC_DIR} -d ${ZMK_BUILD_DIR}/tests/$testcase \
-    -b native_sim/native/64 -p -- -DCONFIG_ASSERT=y -DZMK_CONFIG="$(realpath $path)" \
+    -b native_sim/native/64 -p -- -DCONFIG_ASSERT=y -DCONFIG_ZMK_TEST_BEHAVIORS=y \
```
petejohanson: Why not have the tests themselves set this in the .conf file for the test that needs it? Don't think this should get enabled by default yet.

nmunnich:
Thanks for the review. All points should be addressed now.

PR #3165, opened 2025-12-23
Runtime combos work

petejohanson:
Draft for now, just to kickstart discussions.

1. Adds new messages to the core service for "changes" to move this out of the keymap service. The thinking would be to have the updated UI try the core version of "check unsaved changes" on connection, and if working, use the core variants of those messages, otherwise fall back to the keymap versions, for backwards compat support for keyboards not running the new code.
2. Adds a new combos service, linked to the new runtime combos functionality.

Like with keymap layers, combos assumes you crease some reserved combos, and you can set the max number of supported "key positions" for combos by setting key positions on one of the reserved combos. 

No UI support, but you can test with https://github.com/electronicmaterialsoffice/zmk-studio-cli/pull/2 for now.

Lots more testing needed, just a start for discussion on the design.

## PR check-list

### DIFF
--- a/app/west.yml
+++ b/app/west.yml
@@ -33,7 +33,7 @@ manifest:
       path: modules/lib/gui/lvgl
       remote: zmkfirmware
     - name: zmk-studio-messages
-      revision: 6cb4c283e76209d59c45fbcb218800cd19e9339d
+      revision: add-combos
       path: modules/msgs/zmk-studio-messages
       remote: zmkfirmware
   group-filter: [+optional]


app/include/zmk/keymap.h
```diff
@@ -73,10 +73,10 @@ int zmk_keymap_set_layer_name(zmk_keymap_layer_id_t id, const char *name, size_t
 /**
  * @brief Check if there are any unsaved keymap changes.
  *
- * @retval 0 if there are no changes.
- * @retval 1 if there are changes.
+ * @retval true if there are no changes.
+ * @retval false if there are changes.
  */
-int zmk_keymap_check_unsaved_changes(void);
+bool zmk_keymap_check_unsaved_changes(void);
```
nmunnich, line 79: `has_unsaved_changes` seems like a better naming convention, with the boolean returns inverted.

petejohanson:
> Good refactoring.
> 
> At a skim of the actual runtime support commits, my main concerns would be around the organisation of the extra bits to support runtime. I think as-is if we continue like this all of our files will become even more annoying to read and understand.

Fair. Let me exploring moving the runtime bits into a separate file for at least some clarity/organization.
> 
> I also wonder if there's some abstraction/metadata-based approach we could use to provide editing of the settings for all behaviors, and squint a bit to equate `combo_cfg` to `behavior_config`. If we want to allow out-of-tree behaviors to be customisable with Studio then we'd definitely need some approach like that, and we could probably reuse most of that to add combo customisation.

Yeah, a generic "editable object" metadata approach has been on my mind for a while, to support behavior editing, and could be used for combos as well. I envision it a layer more abstract than just editing behaviors, but any singleton item (e.g. "lighting settings" or "EC Settings") or collections (e.g. combos or hold-taps/behavior instances).

I'm definitely open to playing with that, with combos as a first test case.

PR #3167, opened 2025-12-24
docs: Fix layers property of config combo page

### LINKED ISSUE #3166: Can no longer set combo property defaults explicitly for layers
I hope I'm not way off on this.  Apologies if so or if I have not supplied enough info.

I have a [glove80.keymap](https://github.com/mzs13/glove80-zmk-config/blob/mods/config/glove80.keymap) that I have built without issue many times, but have not rebuilt since mid 2024.  Just today I tried to build and got parsing errors on my config I couldn't makes sense of.    I narrowed it down to two things, 

1.  Setting -1 for combos::layers explicitly when I desire all layers
2. Setting -1 for combos::require-prior-idle-ms explicitly when I want to disable this functionality

I set explicit defaults on other behaviors (e.g. behavior-hold-tap) and it is not an issue.  It only seems to be an issue for combos now. 

I included the link to my config in case you wanted to look at that.  Although my config is kind of complicated and the current version is hacked so it compiles.  Diff with versions prior to today will show what I changed to get it to compile again (i.e remove the explicit defaults).  But if you want me to reproduce with a simpler example I should be able to do that.  I'm hoping from my description and the commit I list below it will be obvious to someone familiar with zmk code. 

Lastly, just glancing at the code in your repo I think below is the pull request that introduced the issue.  But honestly I don't know much about your code to really pinpoint what the exact issue is.  

[issue 2849](https://github.com/zmkfirmware/zmk/pull/2849)

Thank you!

nmunnich:
Has outdated info.

Fixes #3166

### DIFF
--- a/docs/docs/config/combos.md
+++ b/docs/docs/config/combos.md
@@ -32,4 +32,4 @@ Each child node can have the following properties:
 | `timeout-ms`            | int           | All the keys in `key-positions` must be pressed within this time in milliseconds to trigger the combo                                     | 50            |
 | `require-prior-idle-ms` | int           | If any non-modifier key is pressed within `require-prior-idle-ms` before a key in the combo, the key will not be considered for the combo | -1 (disabled) |
 | `slow-release`          | bool          | Releases the combo when all keys are released instead of when any key is released                                                         | false         |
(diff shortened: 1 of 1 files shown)


caksoylar:
Is it intentional to not have removed the default value in the table?

PR #3168, opened 2025-12-24
docs: Add redirect setup script informing users to switch to new user setup page

nmunnich:
See https://discordapp.com/channels/719497620560543766/883452966114324550/1452956056149037096

Old setup script has been removed.

### DIFF
--- a/docs/static/setup.sh
+++ b/docs/static/setup.sh
@@ -1,1 +1,2 @@
-echo "This setup script has been removed. Please use the ZMK CLI instead. Visit https://zmk.dev/docs/user-setup for more information."
+echo "This setup script has been removed. Please use the ZMK CLI instead."
+echo "Visit https://zmk.dev/docs/user-setup for more information."


joelspadin:
Since the old instructions have people running a command to download and execute this script without ever opening the script directly in a browser, won't this just download and attempt to execute the HTML instead, resulting in a confusing error?

Maybe we should instead create a new script which simply prints a message indicating to go to the new setup page?

Also, this is missing the powershell script.

nmunnich:
Good point, I forgot about that detail. Good suggestion, made that change.

docs/static/setup.sh
```diff
@@ -0,0 +1 @@
+echo "This setup script has been removed. Please use the ZMK CLI instead. Visit https://zmk.dev/docs/user-setup for more information."
```
caksoylar:
```suggestion
echo "This setup script has been removed. Please use the ZMK CLI instead."
echo "Visit https://zmk.dev/docs/user-setup for more information."
```
Consistency with ps1 :)

PR #3175, opened 2025-12-28
feat(boards): Add ZMK specific tweaks for upstream rp2040-zero board

nmunnich:
RP2040-zero is a popular enough board that we should have these tweaks in-tree.


petejohanson:
I think we should hold off on this pending the discussions in #3145

peterjc:
I'm working on a `tester_rpi_zero` for the [Waveshare PR2040-Zero](https://www.waveshare.com/wiki/RP2040-Zero) or same pin-out [Waveshare RP2350-Zero](https://www.waveshare.com/wiki/RP2350-Zero), abusing the `rpi_pico` board for now with `&gpio0 29` etc since this exposes a different set of GPIOs (all of GP0 to GP29 inclusive except GP16 used for the RGB LED).

Should I make a PR with it now, or wait on the `rp2040_zero` board in ZMK?

tester_rpi_zero

peterjc:
As per @nmunnich via Discord, I'll wait on this PR, and then submit the tester using `rp2040-zero` as the board.

Note to self: Move the back pads to the left to match front view, or off to one side perhaps?

peterjc:
My Raspberry Pi Zero tester is currently on the branch for https://github.com/peterjc/zmk-keyboard-graph-theory/pull/12 - originally that was for testing my Raspberry Pi PR2040 designs with the RP2350, but is now cleaned up to just focus on the new RP2040-Zero tester.

I'll turn that into a ZMK PR once this issue is done (and it should then use the new `rp2040-zero` board rather than abusing the `rpi_pico` board).

petejohanson:
Should be rebased now that #3145 is in.

peterjc:
Does this need any additional testing? I'm still happy to work on the tester using it once this is merged.

PR #3177, opened 2025-12-29
load settings on POST_KERNEL step

greengrocer98:
Load settings not in main but on POST_KERNEL step for better module support.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -18,14 +18,20 @@ LOG_MODULE_REGISTER(zmk, CONFIG_ZMK_LOG_LEVEL);
 #include <lvgl.h>
 
 #endif
+#if IS_ENABLED(CONFIG_SETTINGS)
+static int settings_init(const struct device *dev) {
+    ARG_UNUSED(dev);
+
+    settings_subsys_init();
+    settings_load();
+    return 0;
+}
+
+SYS_INIT(settings_init, POST_KERNEL, 0);
+#endif
 
 int main(void) {
     LOG_INF("Welcome to ZMK!\n");
-
-#if IS_ENABLED(CONFIG_SETTINGS)
-    settings_subsys_init();
-    settings_load();
-#endif
 
 #ifdef CONFIG_ZMK_DISPLAY
     zmk_display_init();


caksoylar:
Can you elaborate why this is needed? What is the current problem, and are there any trade-offs with the new approach? I am missing context.

petejohanson:
Curious the specific need/use case where this timing is a problem?

And generally, the idea is to use static handlers for loading, and use the commit callback to know when that's completed if you need any processing/work once that's done.

greengrocer98:
What if I need to load settings on my module. I can use settings_load_subtree(), but I can't call this function before settings_subsys_init(). I thing that we should call settings_subsys_init() in the POST_KERNEL step so that modules may load settings from flash with settings_load_subtree()

caksoylar:
> And generally, the idea is to use static handlers for loading, and use the commit callback to know when that's completed if you need any processing/work once that's done.

This seems to be answering "how" and should work within a module. (I don't know enough but maybe Pete can elaborate.) There are modules that utilize settings already, such as https://github.com/george-norton/zmk-behavior-sensor-attr-cycle/blob/main/src/behavior_sensor_attr_cycle.c and https://github.com/efogdev/zmk-pointer-2s-mixer/blob/50264f67bda937e585bf8c69a7412cf2dcda6549/src/behaviors/behavior_p2sm_accel_adj.c, maybe they would help.

petejohanson:
> What if I need to load settings on my module. I can use settings_load_subtree(), but I can't call this function before settings_subsys_init(). I thing that we should call settings_subsys_init() in the POST_KERNEL step so that modules may load settings from flash with settings_load_subtree()

The correct way is to define a static handler for your specific prefix, and let the call from main load everything *once* which will include invoking your callback as needed: https://github.com/zmkfirmware/zmk/blob/main/app/src/keymap.c#L967

Making a call to `settings_load_subtree` from within each module would actually cause the NVS/flash to be re-iterated, the whole thing, multiple times, on startup. Using the static handler and letting ZMK initiate the load from main allows all modules to get their settings loaded during startup without slowing startup of the device.

greengrocer98:
> > What if I need to load settings on my module. I can use settings_load_subtree(), but I can't call this function before settings_subsys_init(). I thing that we should call settings_subsys_init() in the POST_KERNEL step so that modules may load settings from flash with settings_load_subtree()
> 
> The correct way is to define a static handler for your specific prefix, and let the call from main load everything _once_ which will include invoking your callback as needed: https://github.com/zmkfirmware/zmk/blob/main/app/src/keymap.c#L967
> 
> Making a call to `settings_load_subtree` from within each module would actually cause the NVS/flash to be re-iterated, the whole thing, multiple times, on startup. Using the static handler and letting ZMK initiate the load from main allows all modules to get their settings loaded during startup without slowing startup of the device.

But what if I want to load settings on POST_KERNEL step when for example driver is loaded. For the moment settings_load() is called in main function. Sometimes it too late to load settings.

petejohanson:
> > > What if I need to load settings on my module. I can use settings_load_subtree(), but I can't call this function before settings_subsys_init(). I thing that we should call settings_subsys_init() in the POST_KERNEL step so that modules may load settings from flash with settings_load_subtree()
> > 
> > The correct way is to define a static handler for your specific prefix, and let the call from main load everything _once_ which will include invoking your callback as needed: https://github.com/zmkfirmware/zmk/blob/main/app/src/keymap.c#L967
> > 
> > Making a call to `settings_load_subtree` from within each module would actually cause the NVS/flash to be re-iterated, the whole thing, multiple times, on startup. Using the static handler and letting ZMK initiate the load from main allows all modules to get their settings loaded during startup without slowing startup of the device.
> 
> But what if I want to load settings on POST_KERNEL step when for example driver is loaded. For the moment settings_load() is called in main function. Sometimes it too late to load settings.

Too late for what? Use the commit hook callback as a way to continue whatever processing needs to happen once settings are loaded.

petejohanson:
@greengrocer98 Any concerns with the proposed way to handle this, or can we close this?

greengrocer98:
> @greengrocer98 Any concerns with the proposed way to handle this, or can we close this?

Yep, I think I should better load settings on commit. On initialization step I should use default settings and rewrite it on commit.

petejohanson:
> > @greengrocer98 Any concerns with the proposed way to handle this, or can we close this?
> 
> Yep, I think I should better load settings on commit. On initialization step I should use default settings and rewrite it on commit.

I want to be really sure we're on the same page.. you don't load the settings in the commit callback. They are loaded for you, and individually *read* into memory in whatever runtime data structures should contain the data from the *set* callback in your static handler for your prefix, and then do any *final* setup or continuation in the commit hook, which is only called after the individual settings have all already been read into memory.

greengrocer98:
I meant exactly that. Thanks for clarification

greengrocer98:
I meant exactly that. Thanks for clarification

PR #3185, opened 2026-01-02
fix(studio): avoid decoding buffer overflow

cormoran:
This patch fixes studio rpc bug that protobuf decoding buffer can overflow in rpc_read_cb.
Such situation happened when I sent multiple requests to the keyboard very quickly from my custom rpc client and custom protocol.
I confirmed that the patch doesn't cause issue with official ZMK Studio UI at least, although I couldn't reproduce the same issue with ZMK Studio.

## PR check-list


petejohanson (approved):
Yeah, good catch. Thanks!

PR #3209, opened 2026-01-17
Improve ZMK setup instructions

joelspadin:
This makes several improvements to the instructions for installing and using ZMK CLI:

- Windows Terminal is also named "Terminal", so reworded the terminal instructions to not separate Windows from other OSes.

- Specified that Windows Terminal should always be used on Windows, as on older installations of Windows, the default terminal when opening PowerShell may be the legacy console host, which may not support some of the VT sequences used by ZMK CLI.

- Switched from pipx to uv. This eliminates the need for instructions on installing Python, as uv will automatically install a supported version of Python if needed.

- Updated the instructions for running "zmk init", as ZMK CLI version 0.4.0 switched from immediately asking for a repo URL to prompting the user to select between creating a new repo or cloning an existing one first.

- Since for a new user, having a keymap compile successfully on the first attempt after modifying it may be the exception rather than the norm, added a note about error messages in GitHub actions with a link to the troubleshooting page.

Also added instructions for updating ZMK CLI to the latest version on the ZMK CLI page.

### DIFF
--- a/docs/docs/user-setup.mdx
+++ b/docs/docs/user-setup.mdx
@@ -7,20 +7,19 @@ import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
 export const OsTabs = (props) => (
-    <Tabs
+  <Tabs
     groupId="operating-system"
     queryString
     defaultValue="linux"
     values={[
       { label: "Linux", value: "linux" },
       { label: "Windows", value: "win" },
-      { label: "Mac OS", value: "mac" }
+      { label: "macOS", value: "mac" },
     ]}
   >
     {/* eslint-disable-next-line */}
     {props.children}
   </Tabs>
-
 );
 
 Unlike other keyboard firmwares, ZMK Firmware has been built from the ground up to allow users to manage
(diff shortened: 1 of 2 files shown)


docs/docs/user-setup.mdx
```diff
@@ -91,132 +89,128 @@ gh auth login
...
+Finally, we'll install ZMK CLI. Open a terminal and run:
 
 ```sh
-brew install pipx
-pipx ensurepath
+uv tool install zmk
 ```
 
-This should automatically install Python 3 if it is missing.
+:::info[Executable already exists error]
```
caksoylar: Suggestion: change this to a `<details>` block like line 221, so it is collapsed by default

docs/docs/user-setup.mdx
```diff
@@ -91,132 +89,128 @@ gh auth login
 
 Create a [personal access token](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token) and make sure to create it with the "workflow" scope option selected.
 
-When running `git` commands later in this guide, Git will ask for your username and password. Enter your GitHub username, but **do not enter your password**. Instead, paste your access token for the password. You won't be able to see the token again after you create it, so keep the page open until you've done this and/or copy it to a secure location. If you lose the token, simply create another one.
+When running `git` commands later in this guide, Git will ask for your username and password. Enter your GitHub username, but **do not enter your password**. Instead, paste your access token where it asks for the password. You won't be able to see the token again after you create it, so keep the page open until you've done this and/or copy it to a secure location. If you lose the token, simply create another one.
+
+### Install uv
+
+:::info[Other package managers]
```
caksoylar: Can move this into a `details` tag as well, but it is a preference thing rather than conditional on an issue like other examples, so I can see either way.

docs/docs/user-setup.mdx
joelspadin: Done

docs/docs/user-setup.mdx
joelspadin: Also done

PR #3213, opened 2026-01-23
(for v0.4) Add a way to jump to bootloader and/or reset settings on startup

rianadon:
I've built on #1757 to merge in the latest changes in ZMK. Most notably I've integrated the new boot retention mode settings and addressed the comments in the previous PR.

## PR check-list


app/dts/bindings/zmk,boot-magic-combo.yaml
```diff
@@ -0,0 +1,23 @@
...
+    type: int
+    default: 0
+    description: Zero-based index of the key which triggers the action(s).
+  # Boot magic actions:
+  jump-to-bootloader:
+    type: boolean
+    description: Reboots into the bootloader.
+  reset-settings:
+    type: boolean
+    description: Clears settings and reboots.
```
petejohanson: It would be *really* nice, now that we have ZMK Studio, if we could also add an option here for doing a "restore stock keymap", just in case someone mistakenly removes their `&studio_unlock` from their keymap when making changes.

On that node, this should probably just be one properly, that's an enum, e.g.:


```suggestion
  action:
    type: string
    required: true
    enum:
      - "jump-to-bootloader"
      - "reset-settings"
      - "restore-studio-stock-keymap"
```

since it only makes sense, IMHO, for a given boot magic key to perform *one* of the possible boot magic actions, and allows us to grow the enum list later.

app/src/ble.c
```diff
@@ -336,6 +336,37 @@ int zmk_ble_prof_disconnect(uint8_t index) {
...
+
+        int err = settings_delete(setting_name);
+        if (err) {
+            LOG_ERR("Failed to delete setting: %d", err);
+        }
+    }
+
+    // Hardcoding a reasonable hardcoded value of peripheral addresses
+    // to clear so we properly clear a split central as well.
+    for (int i = 0; i < 8; i++) {
```
petejohanson: Given this and the loop above both iterate 8 times, lets unify this into one loop that deletes both kinds of settings.

app/src/boot/boot_magic_combo.c
```diff
@@ -0,0 +1,94 @@
+/*
```
petejohanson, line 1: I'm tempted to say this code should live in the fairly recently created `app/src/boot/` directory at this point.

app/src/boot_magic_key.c
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (c) 2023 The ZMK Contributors
```
petejohanson:
```suggestion
 * Copyright (c) 2026 The ZMK Contributors
```

app/src/boot/boot_magic_combo.c
```diff
@@ -0,0 +1,94 @@
...
+
+SYS_INIT(timeout_init, APPLICATION, CONFIG_APPLICATION_INIT_PRIORITY);
+
+static void trigger_boot_key(const struct boot_key_config *config) {
+    if (config->reset_settings) {
+        LOG_INF("Boot key: resetting settings");
+        zmk_reset_settings();
+    }
+
+    if (config->jump_to_bootloader) {
```
petejohanson, line 90: Maybe I need a refresher here, or @joelspadin can weigh in... But do we really have a use case where you want to reset the settings and then jump to the bootloader? Having a hard time imaging a real need for that specific use case.

app/src/reset.c
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright (c) 2023 The ZMK Contributors
```
petejohanson:
```suggestion
 * Copyright (c) 2026 The ZMK Contributors
```

docs/docs/features/boot-magic-combo.md
```diff
@@ -0,0 +1,209 @@
...
+
+:::info
+
+This clears all BLE bonds. You will need to re-pair the keyboard with any hosts after using this.
+
+:::
+
+:::caution
+
+Currently this action _only_ clears BLE bonds. It will be updated to reset all settings in the future.
```
petejohanson: If we're really going to have this *only* do this, I'm wary about this changing out from under folks in the future. Can we name this current action "clear-ble-bonds" or something, and add a true "reset-settings" action later as an *additional* action option, instead of the current PR approach?

@joelspadin any concerns?

docs/docs/features/boot-magic-key.md
```diff
@@ -0,0 +1,209 @@
...
+
+Currently this action _only_ clears BLE bonds. It will be updated to reset all settings in the future.
+
+:::
+
+## Multiple Actions
+
+If you want a single boot magic key to perform multiple actions, simply add properties for each action to the same `zmk,boot-magic-key` node. The order of the properties does not matter.
+
+For example, to make a key that resets settings and then reboots to the bootloader, add both `reset-settings` and `jump-to-bootloader`:
```
petejohanson: Still not sure we really need this. Open to being convinced otherwise though. Thanks.

app/dts/bindings/zmk,boot-magic-key.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2023, The ZMK Contributors
```
petejohanson:
```suggestion
# Copyright (c) 2026, The ZMK Contributors
```

petejohanson (changes requested):
Thanks for working on this! A few thoughts on the implementation.

docs/docs/features/boot-magic-key.md
peterjc: Clean slate when you've been tinkering with ZMK studio while getting the matrix right? Saves building and flashing a settings reset firmware too.

app/dts/bindings/zmk,boot-magic-combo.yaml
peterjc: +1 on the ZMK Studio "restore default keymap" idea - worth logging on https://github.com/zmkfirmware/zmk-studio/issues ?

nmunnich: I think it would be even nicer if, instead of restoring the default keymap, we had an option to just unlock studio.

docs/docs/features/boot-magic-key.md
petejohanson: I mean a dual function key that both resets settings and jumps into the bootloader.... Seems super niche, and I'm not sure I understand your described need for studio iteration... Can you explain in exact steps what use case you're considering?

app/dts/bindings/zmk,boot-magic-combo.yaml
petejohanson: > I think it would be even nicer if, instead of restoring the default keymap, we had an option to just unlock studio.

Yeah, like this much better.

docs/docs/features/boot-magic-key.md
peterjc: Possibly user ignorance or misunderstanding, but I've flashed new firware onto a board to find what I though were traces of past changes I'd made in ZMK Studio. An easy way to flash new firmware _and_ wipe out any past Studio changes seems like a good thing. And wipe old Bluetooth pairings too?

docs/docs/features/boot-magic-combo.md
joelspadin: IIRC, I originally wrote this when the settings_reset shield only cleared BLE bonds and didn't reset all settings. I'd suggest starting with a full settings reset option. I'm not sure if there are cases where you'd want to clear BLE bonds only instead of resetting everything, but if so, we could add a separate option for that too.

app/src/boot/boot_magic_combo.c
joelspadin: I'm sure I probably came up with a reason for that when I originally wrote this, but I can't think of one now.

docs/docs/features/boot-magic-combo.md
petejohanson: The main reason I can think to maintain a "clear BLE only" is to preserve any studio changes when trying to fix pairing issues between halves of a split, so I think there's a use case for both actions being possible from boot keys.

petejohanson:
As I've thought on this more, I'm having some concerns about having "single key" magic keys.. unlike QMK, we are *far* more likely to have someone tapping a key to *wake the device* and run this early code, potentially with them tapping/accidentally tapping the 0th key to do so.

If we keep this as is, I fear we'll get a lot of accidental triggers. What do folks think about having multiple positions assignable to a given boot magic key? Doesn't need to have fancy combo handling, just enough to track "are all positions pressed that are part of this boot magic key?"

docs/docs/features/boot-magic-combo.md
caksoylar: While we are thinking about the options, I think it might be nice to have separate options for "clear split bonds" vs. "clear BLE profiles" (except the split ones). Not necessarily in the first iteration.

docs/docs/features/boot-magic-key.md
caksoylar: I can think of a scenario where a designer wants only a single bootmagic key, in which case you'd want it to do everything, including clearing _all_ settings and jumping to bootloader.

caksoylar:
> If we keep this as is, I fear we'll get a lot of accidental triggers. What do folks think about having multiple positions assignable to a given boot magic key? Doesn't need to have fancy combo handling, just enough to track "are all positions pressed that are part of this boot magic key?"

That's a valid concern but since we aren't enabling this by default, the users should be aware of the specific key position (or the keyboard designers should tell the users), right? I think it's good to have, rather than must.

What would the effective combo term be? Is it a simple trigger where you want both keys pressed at the time of the check (so the effective term is the boot time) or do we implement logic to wait for keys?

petejohanson:
> > If we keep this as is, I fear we'll get a lot of accidental triggers. What do folks think about having multiple positions assignable to a given boot magic key? Doesn't need to have fancy combo handling, just enough to track "are all positions pressed that are part of this boot magic key?"
> 
> That's a valid concern but since we aren't enabling this by default, the users should be aware of the specific key position (or the keyboard designers should tell the users), right? I think it's good to have, rather than must.

Still might have something bump the corner of your keeb and wake it and since the check is just "is this key pressed before the timeout" you'd suddenly potentially lose all your studio customization, etc.

I think this functionality is super useful, but we also shoukd try to design this in a way that's "easy not to screw up".

I also don't trust designers/vendors to consistently communicate this, nor users to always read what is communicated by their vendor.

> 
> What would the effective combo term be? Is it a simple trigger where you want both keys pressed at the time of the check (so the effective term is the boot time) or do we implement logic to wait for keys?

I think just "are all keys held at once before the boot magic timeout"

nmunnich:
I wonder if "check on boot" is even the right thing to aim for, for wireless devices. I think a "check on usb insertion" might make more sense for those kind of devices. We would lose more function for any keyboards wanting to run hardware without usb support, but those sorts of devices need to be flashed with a programmer and are unlikely to have a bootloader/are more heavy enthusiast.

petejohanson:
> I wonder if "check on boot" is even the right thing to aim for, for wireless devices. I think a "check on usb insertion" might make more sense for those kind of devices. We would lose more function for any keyboards wanting to run hardware without usb support, but those sorts of devices need to be flashed with a programmer and are unlikely to have a bootloader/are more heavy enthusiast.

I like where you're thinking! Why don't we use the [hwinfo API](https://docs.zephyrproject.org/latest/hardware/peripherals/hwinfo.html) to determine the reset reason, and ignore the boot magic unless it's one of the "valid" reasons (.e.g. reset pin or power-on-reset)? I think that might a good way to avoid the accidental triggers from a key press to wake, and allow simple one button boot magic.

app/dts/bindings/zmk,boot-magic-combo.yaml
rianadon: It's almost sounding like the action should just be a pointer to some ZMK behavior :)

app/src/boot/boot_magic_combo.c
rianadon: QMK does a reset then jump, but it's always bothered me that there's no way to do just a jump to bootloader. With QMK I've had some troubles with my saved changes causing conflicts when I make major changes to the hardware like changing the number of keys in the firmware, and a settings reset was necessary. I haven't stress tested ZMK as much, but if this isn't an issue here (Studio does seem very well architected), then I agree just one makes sense.

joelspadin:
Adding support for combos probably wouldn't be too hard either. You could add a `bool[]` of the same size as the key positions array (or bitfield, though that wouldn't really save space unless you do something silly like a 5+ key combo) to the driver data to store the pressed states of each of the key positions. Track both presses and releases until the timeout expires, and if every element of the array is `true` after a key event, trigger the boot magic logic.

petejohanson:
> Adding support for combos probably wouldn't be too hard either. You could add a `bool[]` of the same size as the key positions array (or bitfield, though that wouldn't really save space unless you do something silly like a 5+ key combo) to the driver data to store the pressed states of each of the key positions. Track both presses and releases until the timeout expires, and if every element of the array is `true` after a key event, trigger the boot magic logic.

Yeah, my thoughts exactly. We don't need crazy "combo" logic here.

nmunnich:
I think it would be a good idea to predefine a bootloader boot magic key for our in-tree keyboards. Maybe top left key as a convention, mirrored for splits?

docs/docs/features/boot-magic-combo.md
```diff
@@ -150,7 +150,7 @@ For split keyboards, you can define multiple boot magic keys and then only enabl
...
 
-For example, consider a split keyboard which has 6 columns per side by default but supports a 5-column layout, and assume you want the top-left key on the left side and the top-right key on the right side to be boot magic keys. The top-left key will be position 0 regardless of layout, but the top-right key will be position 11 by default and position 9 in the 5-column layout.
+For example, consider a split keyboard which has 6 columns per side by default but supports a 5-column layout, and assume you want the top-left key on the left side and the top-right key on the right side to be boot magic combos. The top-left key will be position 0 regardless of layout, but the top-right key will be position 11 by default and position 9 in the 5-column layout.
```
caksoylar, line 193: This will be an issue when the selected physical layout is changed at runtime (e.g. via ZMK Studio. I think we need a way to specify this per physical layout. @petejohanson do you have a suggestion?

joelspadin: Yeah, I originally wrote this before physical layouts were a thing, so I didn't have any consideration for dynamically switching layouts.

How is this handled for sideband behaviors? 

Actually, could we maybe just _use_ sideband behaviors instead of creating a new system for this? Could we add properties to turn a sideband behavior into a boot magic key by making it only process the sideband behavior at boot and then pass through to the regular handling after that? Supporting combos with sideband behaviors would be tricky, but I don't know if that's a hard requirement.

caksoylar: Sideband behaviors (typically) use their own kscan, right? So it doesn't relate to physical layouts. But I might not be thinking about this right.

I was thinking perhaps multiple bootmagic nodes defined, then each physical layout gets a phandle, similar to matrix transform/kscan. Then only the bootmagic node referred to by the active physical transform is active.

rianadon:
I could use some more clarity on how to handle physical layouts here. Should the bootmagic key/combo be defined independently for each physical layout? Or defined in terms of the kscan row and column? Or keep the behavior as-is?

peterjc:
@rianadon I suspect you're not asking about the problem I filled as #3228 (not specific to the bootloader actions at startup), but the equivalent ambiguity for the low level combos being discussed here (with `BOOT_KEY_COMBO_POSITIONS` specific to startup).

rianadon:
> @rianadon I suspect you're not asking about the problem I filled as #3228 (not specific to the bootloader actions at startup), but the equivalent ambiguity for the low level combos being discussed here (with `BOOT_KEY_COMBO_POSITIONS` specific to startup).

I was referring to the issues @joelspadin and @caksoylar brought up in their discussion above, but your problem does sound relevant. Even though we are working at different layers, the user-facing configuration for combos and bootmagic should ideally be very similar.

rianadon:
I've revised the configuration of the boot magic combos to happen at the physical layout layer instead of the global layer as @caksoylar has suggested. I think overall this is the cleanest approach, and it also makes working with split keyboards nicer.

Here's how the example in the docs for boot magic key configuration on a split keyboard with two physical layouts looks:

**shield.dtsi**

```c
/ {
    chosen {
        zmk,physical-layout = &default_layout;
    };

    default_layout: default_layout {
        compatible = "zmk,physical-layout";
        display-name = "Default Layout";
        transform = <&default_transform>;
    };

    five_column_layout: five_column_layout {
        compatible = "zmk,physical-layout";
        display-name = "5-Column Layout";
        transform = <&five_column_transform>;
    };

    bootloader_key_left: bootloader_key_left {
        compatible = "zmk,boot-magic-combo";
        combo-positions = <0>;
        jump-to-bootloader;
    };

    bootloader_key_right: bootloader_key_right {
        compatible = "zmk,boot-magic-combo";
        combo-positions = <11>;
        jump-to-bootloader;
    };

    bootloader_key_right_fivecol: bootloader_key_right_fivecol {
        compatible = "zmk,boot-magic-combo";
        combo-positions = <9>;
        jump-to-bootloader;
    };
    ...
};
```

**shield_left.overlay**

```c
#include "shield.dtsi"

&default_layout {
    boot-magic-combos = <&bootloader_key_left>;
};
&five_column_layout {
    boot-magic-combos = <&bootloader_key_left>;
}
```

**shield_right.overlay**

```c
#include "shield.dtsi"

&default_layout {
    boot-magic-combos = <&bootloader_key_right>;
};
&five_column_layout {
    boot-magic-combos = <&bootloader_key_right_fivecol>;
}
```

**shield.keymap**

```c
// Uncomment this block if using the 5-column layout
// / {
//     chosen {
//         zmk,physical-layout = &five_column_layout;
//     };
// };
```

rianadon:
@petejohanson I'd appreciate if you could take a look at the new configuration for the boot magic combos.
If that looks good, I believe this is the last major technical change requested the PR:

> I like where you're thinking! Why don't we use the [hwinfo API](https://docs.zephyrproject.org/latest/hardware/peripherals/hwinfo.html) to determine the reset reason, and ignore the boot magic unless it's one of the "valid" reasons (.e.g. reset pin or power-on-reset)? I think that might a good way to avoid the accidental triggers from a key press to wake, and allow simple one button boot magic.

There's also the open question of whether we allow the boot magic combo to perform multiple actions, or just one. Personally I think just one is fine, and the case of needing multiple is rare enough that we could sort the order of combo execution so that if you define one combo for settings reset and another for bootloader, the bootloader always runs last. Or better yet, now that combos are passed as an array, we treat the order of them in the array as the order of execution.

nmunnich:
Your new docs page isn't actually showing up in the sidebar. Without looking into it at that much detail, I would expect it to work fine if you added the boot magic keys to both halves of a split, and the split will simply ignore those belonging to the other half? Or is there a scenario where the central would process the same as the peripheral if both are connected via UART while processing boot-magic?

Basically, why not
```
/ {
    chosen {
        zmk,physical-layout = &default_layout;
    };

    default_layout: default_layout {
        compatible = "zmk,physical-layout";
        display-name = "Default Layout";
        transform = <&default_transform>;
    };

    five_column_layout: five_column_layout {
        compatible = "zmk,physical-layout";
        display-name = "5-Column Layout";
        transform = <&five_column_transform>;
    };

    bootloader_key_left: bootloader_key_left {
        compatible = "zmk,boot-magic-combo";
        combo-positions = <0>;
        jump-to-bootloader;
    };

    bootloader_key_right: bootloader_key_right {
        compatible = "zmk,boot-magic-combo";
        combo-positions = <11>;
        jump-to-bootloader;
    };

    bootloader_key_right_fivecol: bootloader_key_right_fivecol {
        compatible = "zmk,boot-magic-combo";
        combo-positions = <9>;
        jump-to-bootloader;
    };
    ...
};

&default_layout {
    boot-magic-combos = <&bootloader_key_left &bootloader_key_right>;
};
&five_column_layout {
    boot-magic-combos = <&bootloader_key_left &bootloader_key_right_fivecol>;
}

```

rianadon:
If the halves connect to each other quickly enough, then with this configuration pressing the boot magic key on one half of the split would reset both halves of the keyboard.

To me it makes sense to just have the boot magic key reset the half it belongs to, but that's merely a personal preference and there's no reason why your way wouldn't work.

rianadon:
I've believe I've implemented everything that has been requested. I'd appreciate if the PR could get another review.

To summarize, the four implemented actions are `jump-to-bootloader`, `unpair-ble` (only removes BLE bonds from storage), `reset-settings` (clears everything in storage), and `unlock-studio`.

I've added a check to make sure the combos are only processed if the chip boots due to being plugged in for the first time or physically reset (things like temperature reset, brownout, and clock errors will cause these combos to be processed). Everything is updated in the docs.

Let me know if there's anything I overlooked.

peterjc:
I take it this doesn't do anything along the lines of @nmunnich's earlier comment:

> I wonder if "check on boot" is even the right thing to aim for, for wireless devices. I think a "check on usb insertion" might make more sense for those kind of devices.

So if I have a battery powered ZMK keyboard that I want to have connected to USB in boot loader mode, is this how your PR would work:
1. Disconnect the USB cable
2. Disconnect the battery (easy with a power switch)
3. The keyboard is now off.
4. Press and hold the configured combo...
5. Connect the USB cable (which will supply power and startup the keyboard)
6. Release the combo
7. Should now be in boot loader mode, eg available as a USB drive

That's workable - assuming your keyboard has a power switch, rather than an always on soldered battery or just a JST or similar connector to remove the battery.

rianadon:
After adding support for ZMK studio unlock that was requested, I decided to stick with checking on boot and not on USB insertion. Because ZMK studio can work wirelessly, I don't think it would be a good idea to require the keyboard to be connected through USB for simply unlocking studio should the user choose to use the boot combos rather than dedicating a key to the unlock.

Maybe it's worth it to support checking on both boot and usb connection.

As for your flow, yes, that would be correct. If your keyboard has a reset button, you could press that while holding down the combo too. However, it's probably more likely the keyboard has an on/off switch than a reset button.

peterjc:
Good point - there are battery powered keyboards out there with an easy reset button but no power switch like the [Visorbearer](https://github.com/carrefinho/visorbearer/tree/main), so that'll work for them.

PR #3215, opened 2026-01-25
fix(studio): Ensure putting framing byte when sending rpc

cormoran:
If tx buffer became full when sending PRC response, framing byte can be skipped and communication gets stuck. This patch adds wait for putting EOF failure and adds error handling for putting SOF failure.

I hit this error when sending keylayout data on BLE gatt. Waiting in the loop practically resolved the issue.

## PR check-list

### DIFF
--- a/app/src/studio/rpc.c
+++ b/app/src/studio/rpc.c
@@ -188,7 +188,11 @@ static int send_response(const zmk_studio_Response *resp) {
     pb_ostream_t stream = pb_ostream_for_tx_buf(user_data);
 
     uint8_t framing_byte = FRAMING_SOF;
-    ring_buf_put(&rpc_tx_buf, &framing_byte, 1);
+    if (ring_buf_put(&rpc_tx_buf, &framing_byte, 1) != 1) {
+        LOG_ERR("RPC TX buffer full");
+        k_mutex_unlock(&rpc_transport_mutex);
+        return -ENOMEM;
+    }
 
     selected_transport->tx_notify(&rpc_tx_buf, 1, false, user_data);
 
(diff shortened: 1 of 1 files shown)


app/src/studio/rpc.c
```diff
@@ -188,7 +188,11 @@ static int send_response(const zmk_studio_Response *resp) {
     pb_ostream_t stream = pb_ostream_for_tx_buf(user_data);
 
     uint8_t framing_byte = FRAMING_SOF;
-    ring_buf_put(&rpc_tx_buf, &framing_byte, 1);
+    if (ring_buf_put(&rpc_tx_buf, &framing_byte, 1) != 1) {
+        LOG_ERR("RPC TX buffer full");
+        k_mutex_unlock(&rpc_transport_mutex);
+        return -ENOMEM;
+    }
```
cormoran, line 195: Treating SOF put failure as error sounds fine? I've never hit error here yet.

petejohanson: I'm on the fence here, whether there should be any attempt to retry a few times, and then return -ENOMEM if that fails? Would give the underlying transport a chance to empty the ring buffer to free space. I'd suggest we try that.

In the case of a failure, I usually prefer, when locking involved, to create a new local at the top of the function `int ret = 0;` and then in this branch, set `ret = -ENOMEM; goto exit` and update the `exit:` to return `ret` instead.

app/src/studio/rpc.c
```diff
@@ -199,11 +203,15 @@ static int send_response(const zmk_studio_Response *resp) {
...
         LOG_ERR("Failed to encode the message %s", stream.errmsg);
 #endif // !IS_ENABLED(CONFIG_NANOPB_NO_ERRMSG)
+        k_mutex_unlock(&rpc_transport_mutex);
         return -EINVAL;
     }
 
     framing_byte = FRAMING_EOF;
-    ring_buf_put(&rpc_tx_buf, &framing_byte, 1);
+    // If response is large, tx buffer can be full for async transport.
+    while (ring_buf_put(&rpc_tx_buf, &framing_byte, 1) == 0) {
```
petejohanson, line 212: This should ideally "give up" after a certain period/number of tries, and return, so we don't get stuck in an infinite loop, IMHO.

PR #3235, opened 2026-02-13
feat(behaviors): add `&out OUT_NONE`

snoyer:
#3140 introduced `ZMK_TRANSPORT_NONE` and mentions that it

> can be set as the preferred endpoint transport if you wish to prevent the keyboard from sending any output

but it doesn't provide a way to do it (as far as I can tell). This PR allows it with `&out OUT_NONE`.

### DIFF
--- a/app/src/behaviors/behavior_outputs.c
+++ b/app/src/behaviors/behavior_outputs.c
@@ -65,6 +65,8 @@ static int on_keymap_binding_pressed(struct zmk_behavior_binding *binding,
         return zmk_endpoint_set_preferred_transport(ZMK_TRANSPORT_USB);
     case OUT_BLE:
         return zmk_endpoint_set_preferred_transport(ZMK_TRANSPORT_BLE);
+    case OUT_NONE:
+        return zmk_endpoint_set_preferred_transport(ZMK_TRANSPORT_NONE);
     default:
         LOG_ERR("Unknown output command: %d", binding->param1);
     }
--- a/app/src/behaviors/behavior_outputs.c
+++ b/app/src/behaviors/behavior_outputs.c
@@ -42,6 +42,11 @@ static const struct behavior_parameter_value_metadata std_values[] = {
         .type = BEHAVIOR_PARAMETER_VALUE_TYPE_VALUE,
     },
 #endif // IS_ENABLED(CONFIG_ZMK_BLE)
+    {
+        .value = OUT_NONE,
+        .display_name = "No Output",
+        .type = BEHAVIOR_PARAMETER_VALUE_TYPE_VALUE,
+    },
 };
 
 static const struct behavior_parameter_metadata_set std_set = {
--- a/app/include/dt-bindings/zmk/outputs.h
+++ b/app/include/dt-bindings/zmk/outputs.h
@@ -7,3 +7,4 @@
 #define OUT_TOG 0
 #define OUT_USB 1
 #define OUT_BLE 2
+#define OUT_NONE 3
--- a/docs/docs/keymaps/behaviors/outputs.md
+++ b/docs/docs/keymaps/behaviors/outputs.md
@@ -29,11 +29,12 @@ header, which is added at the top of the keymap file:
 
 This allows you to reference the actions defined in this header:
 
-| Define    | Action                                          |
-| --------- | ----------------------------------------------- |
-| `OUT_USB` | Prefer sending to USB                           |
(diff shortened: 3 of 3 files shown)


app/src/behaviors/behavior_outputs.c
joelspadin, line 44: This needs a new entry to work with Studio.

Related: I've seen people mention that Studio only seems to connect on whichever transport (USB or BLE) is currently enabled. We should double check the behavior when the preferred transport is explicitly "none". Can you still use Studio in that case? 

If Studio doesn't work in that case, then editing your keymap in Studio to have `&out OUT_NONE` and no other `&out` keys, then pressing that key will effectively brick your keyboard until you do a settings reset on it, because the Studio keymap will take priority over any changes you make to the base keymap in firmware, and there will be no way to connect Studio to fix it.

joelspadin (changes requested):

app/src/behaviors/behavior_outputs.c
snoyer: > Can you still use Studio in that case?

I indeed get "Failed to connect to the chosen device" when trying to use Studio in the browser after having pressed `&out OUT_NONE` :(
Is that a Studio issue that could be fixed/improved? or is it a known and accepted limitation of the system?

> If Studio doesn't work in that case, then editing your keymap in Studio to have `&out OUT_NONE` and no other `&out` keys, then pressing that key will effectively brick your keyboard until you do a settings reset on it, because the Studio keymap will take priority over any changes you make to the base keymap in firmware, and there will be no way to connect Studio to fix it.

That makes sense and I hadn't thought of that (or of any of the Studio implications at all TBH).
What does it mean tho? We still add an entry to work with Studio even if it can soflock the keyboard in the case you mentioned? or is this whole `&out OUT_NONE` a bad idea in the first place?

joelspadin: Not adding an entry for Studio will mean that
1. You can't select it as an option in Studio
2. If the base keymap uses it, Studio won't know what it means

The first thing isn't terrible, but the second one could cause problems. 

I'd be interested to hear @petejohanson 's thoughts on this. Can the keyboard side of Studio maybe be made to listen on every possible transport and respond on whichever one it gets a connection from instead of relying on the preferred transport?

snoyer: Ok I'll add the entry, and we'll see what to do, if anything, about the potential "dangerous" usage.

Is this whole getting locked out situation something the design/implementation should guard against or is it just something that users should be careful no to do?

Also is it specific to this PR or are there already similar ways to shoot yourself in the foot? Would using Studio to remove your only `&studio_unlock` behavior result in the same kind of situation (in this case you could still type but you'd still have locked yourself out of editing the keymap I assume)?

joelspadin: While removing `&studio_unlock` isn't ideal, I think you could recover by flashing new firmware that doesn't require unlocking. That also feels like enough of a special case that the Studio app should warn you when trying to save a keymap that doesn't have `&studio_unlock` to a keyboard that has the lock config enabled.

I do think we should try to prevent user from "soft locking" their keyboards, though I don't think that's something that needs to be solved in this PR. Whether it needs to be solved *first* before submitting this PR is a question I think we should answer though.

joelspadin (approved):
I'll let @petejohanson decide if the Studio issue is worth holding off on this, but the code looks good to me.

app/src/behaviors/behavior_outputs.c
petejohanson: I don't think this needs to be solved by this PR. Long term I'd love to have connecting over multiple/transports other than the selected one possible with Studio, it gets messy *fast* without a ton more code, which is why it's not been done yet. I don't think that's a blocker here though.

petejohanson (approved):

app/src/behaviors/behavior_outputs.c
joguSD: > Would using Studio to remove your only &studio_unlock behavior result in the same kind of situation (in this case you could still type but you'd still have locked yourself out of editing the keymap I assume)?

Yeah, I've definitely done this to myself where I modified the keymap in Studio and removed the only way to unlock it. I added some logic like this that allows key position zero to be held for 5 seconds and call studio unlock:

https://github.com/joguSD/zmk-mntre-module/blob/c08e0d1e38c3a871a5cb5cee9e4781fae29fecdb/src/reform/studio_unlock.c

PR #3239, opened 2026-02-14
feat: Add LED indicator support

joelspadin:
This adds a new zmk,indicator-leds device, which maps HID indicator states onto any devices that implement the LED driver API. This adds support for things like a caps lock LED.

The name was chosen so that more drivers could be added later as zmk,indicator-*, for example a version that uses the LED strip API.

## PR check-list

### DIFF
--- a/app/src/indicators/Kconfig
+++ b/app/src/indicators/Kconfig
@@ -1,0 +1,4 @@
+# Copyright (c) 2026 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+rsource "Kconfig.indicator_leds"
--- a/app/src/indicators/indicator_leds.c
+++ b/app/src/indicators/indicator_leds.c
@@ -1,0 +1,226 @@
+/*
+ * Copyright (c) 2025 The ZMK Contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#define DT_DRV_COMPAT zmk_indicator_leds
+
+#include <errno.h>
+#include <zephyr/device.h>
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/led.h>
+#include <zephyr/pm/device.h>
+
+#include <zmk/endpoints.h>
+#include <zmk/event_manager.h>
+#include <zmk/hid_indicators.h>
+#include <zmk/usb.h>
+#include <zmk/events/activity_state_changed.h>
+#include <zmk/events/endpoint_changed.h>
+#include <zmk/events/hid_indicators_changed.h>
+#include <zmk/events/usb_conn_state_changed.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
+
+struct indicator_led_child_config {
+    size_t leds_len;
+    const struct led_dt_spec *leds;
+
+    zmk_hid_indicators_t indicator;
+    uint8_t active_brightness;
+    uint8_t inactive_brightness;
+    uint8_t disconnected_brightness;
+    bool on_while_idle;
+};
+
+struct indicator_led_config {
+    size_t indicators_len;
+    const struct indicator_led_child_config *indicators;
+};
+
+struct indicator_led_data {
+    enum zmk_activity_state activity_state;
+    zmk_hid_indicators_t indicators;
+    bool usb_powered;
+    bool pm_suspended;
+    bool endpoint_connected;
+};
+
+static bool is_led_disabled(const struct indicator_led_child_config *config,
+                            const struct indicator_led_data *data) {
+    // LEDs should always be off if the device is suspended.
+    if (data->pm_suspended) {
+        return true;
+    }
+
+    // If the keyboard is powered, LEDs don't need to be disabled to save power.
+    if (data->usb_powered) {
+        return false;
+    }
+
+    switch (data->activity_state) {
+    case ZMK_ACTIVITY_ACTIVE:
+        return false;
+
+    case ZMK_ACTIVITY_IDLE:
+        return !config->on_while_idle;
+
+    case ZMK_ACTIVITY_SLEEP:
+        return true;
+    }
+
+    LOG_ERR("Unhandled activity state %d", data->activity_state);
+    return false;
+}
+
+static uint8_t get_brightness(const struct indicator_led_child_config *config,
+                              const struct indicator_led_data *data) {
+    if (is_led_disabled(config, data)) {
+        return 0;
+    }
+
+    if (!data->endpoint_connected) {
+        return config->disconnected_brightness;
+    }
+
+    const bool active = data->indicators & config->indicator;
+    return active ? config->active_brightness : config->inactive_brightness;
+}
+
+static int update_indicator(const struct indicator_led_child_config *config,
+                            const struct indicator_led_data *data) {
+    const uint8_t value = get_brightness(config, data);
+
+    for (int i = 0; i < config->leds_len; i++) {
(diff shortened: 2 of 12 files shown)


docs/docs/config/led-indicators.md
```diff
@@ -0,0 +1,41 @@
...
+| `indicator`               | int      | The `HID_INDICATOR_*` value to indicate                               |         |
+| `leds`                    | phandles | One or more LED devices to control                                    |         |
+| `active-brightness`       | int      | LED brightness in percent when the indicator is active                | 100     |
+| `inactive-brightness`     | int      | LED brightness in percent when the indicator is not active            | 0       |
+| `disconnected-brightness` | int      | LED brightness in percent when the keyboard is not connected          | 0       |
+| `on-while-idle`           | bool     | Keep LEDs enabled even when the keyboard is idle and on battery power | false   |
+
+The `indicator` property must be one of the `HID_INDICATOR_*` values defined in [zmk/app/include/dt-bindings/zmk/hid_indicators.h](https://github.com/zmkfirmware/zmk/blob/main/app/include/dt-bindings/zmk/hid_indicators.h). You may also combine values with `|` to make the LED only be lit when multiple indicator states are active at the same time.
```
joelspadin: I wasn't sure if this should be "only light LED when all indicators is active" or "only light LED when any indicator is active", or if this is even a valid use case. (If anyone can think of a reason to want both options, I could add a property to switch between OR and AND.)

joelspadin:
Example usage in a working board: https://github.com/joelspadin/zmk-keyboards/blob/0ba8648d2c4b8d6e2944f0afd5c3a9908fa3d5c6/boards/joelspadin/marten_numpad/marten_numpad.dts#L59

(That module also contains a slightly older copy of this driver.)

app/src/indicators/Kconfig
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2025 The ZMK Contributors
+# SPDX-License-Identifier: MIT
+
+config ZMK_INDICATOR_LEDS
```
petejohanson: I'd like to start moving toward a more organize Kconfig structure, modeled off what Zephyr does, so moving these into a `Kconfig.indicator_leds` file which gets sources from here.

Doing so will help keep the "one giant Kconfig file" at bay a bit.

app/src/indicators/indicator_leds.c
```diff
@@ -0,0 +1,226 @@
...
+        return false;
+
+    case ZMK_ACTIVITY_IDLE:
+        return !config->on_while_idle;
+
+    case ZMK_ACTIVITY_SLEEP:
+        return true;
+    }
+
+    LOG_ERR("Unhandled activity state %d", data->activity_state);
```
petejohanson, line 74: I prefer putting this in the `default` of the switch above.

petejohanson (changes requested):
Looks really good, generally. A few minor comments from a first pass.

app/src/indicators/indicator_leds.c
joelspadin: *Not* creating a default case was intentional. I don't know if we have the specific compiler option enabled for this, but if you switch on an enum and don't include a default case, then the compiler can warn/error if you do not have a case for every enum value. Then you get notified at compile time if someone adds a new value to the enum and forgets to update every switch block.

petejohanson: Fair. Disregard.

app/src/indicators/Kconfig
joelspadin: Done, assuming I correctly understand what you mean here.

docs/docs/config/led-indicators.md
petejohanson: I don't think that's really a valid use case, IMHO, and we can keep this as-is.

petejohanson (approved):
LGTM! @zmkfirmware/docs any thoughts?

docs/docs/features/led-indicators.md
```diff
@@ -0,0 +1,151 @@
...
+- Kana
+
+To connect these indicator states to LEDs on a keyboard, you must do two things in your board or shield files:
+
+1. Define an LED driver and one or more LEDs to control.
+2. Configure ZMK to control those LEDs.
+
+## LED Definitions
+
+The most common setup is for LEDs to be driven directly from a GPIO pin. The examples on this page show how to controls these using the `gpio-leds` or `pwm-leds` drivers, but you can also use any other [LED driver](https://docs.zephyrproject.org/4.1.0/hardware/peripherals/led.html) from Zephyr.
```
caksoylar: Can we add a comment/note box on using led strips (or directly call out smart leds) here? That's going to be one question we'll inevitably get.

docs/docs/features/led-indicators.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+Then, add a `zmk,indicator-leds` node. This node can have any number of child nodes. Each child maps an indicator state to one or more LEDs:
+
+```dts
+/ {
+    leds {
+        compatible = "gpio-leds";
+
+        num_lock_led: num_lock_led { ... };
```
caksoylar:
```suggestion
        // See above section for defining these node contents
        num_lock_led: num_lock_led { ... };
```

docs/docs/features/led-indicators.md
```diff
@@ -0,0 +1,151 @@
...
+
+The above example only supports LEDs being off or on at full brightness. If you want to be able to reduce the brightness or use multiple brightness levels, you must use `pwm-leds` instead of `gpio-leds`. Note that this will increase power usage slightly when LEDs are enabled compared to using `gpio-leds`.
+
+See the [backlight hardware integration page](../development/hardware-integration/lighting/backlight.mdx) for an example of configuring PWM LEDs.
+
+## Indicator Definitions
+
+Now that you have some LEDs defined, you can configure ZMK to use them.
+
+In your shield's `.overlay` file or board's `.dts` file, add the following include to the top of the file:
```
caksoylar: The features section of the docs are generally geared towards users, not keyboard designers. For other features we typically document enabling and configuring them, and direct to "hardware integration" page for adding support for them. Given that, we should likely move this page to hardware integration, and replace here with a page that mostly consists of pointing to that page.

See also the section below that mentions
> this will allow users of your keyboard to more easily adjust the indicator behaviors in their `.keymap` files

peterjc:
Regarding:

> You may also combine values with `|` to make the LED only be lit when multiple indicator states are active at the same time.

I agree lighting an LED when certain HID combinations are _all_ on will be very niche (logical AND).

However, lighting an LED when _any_ of certain HID combinations are on is useful (logical OR). Why? Most obvious with Caps Locks and NumLock, but if any HID statuses are active some keys on the keyboard will not necessarly work as expected. With only one LED on many keyboards, having it lit on _any_ HID indicator seems useful to me.

docs/docs/features/led-indicators.md
joelspadin: Split a lot of this out into a hardware integration page.

docs/docs/features/led-indicators.md
joelspadin: Done.

docs/docs/config/led-indicators.md
joelspadin: Swapped this from AND to OR, since @peterjc identified a possible use case for that.

joelspadin:
Rebased, fixed a missing copyright header, and squashed the fixes into the original commit.

peterjc:
Working for me with the (RP2040) Raspberry Pi Pico's single user LED on GP25 - thank you! The docs where enough to add this, with only a few minor missteps on my part.

PR #3241, opened 2026-02-15
fix(pointing): Support multiple temp layer instances

nobyk:
Previously, the `input-processor-temp-layer` driver failed to support multiple instances because the background work queue callback (`layer_action_work_cb`) was hardcoded to operate only on the first driver instance (instance 0). This caused actions from secondary instances to either be ignored or incorrectly apply state changes to the first instance.

This commit fixes the issue by:
- Adding a `dev` pointer to the `layer_state_action` struct to identify the originating device instance for each event.
- Updating the work callback to use the specific device instance from the action payload.
- Implementing a lookup mechanism for timeout-based deactivation to find and update the correct active instance managing the expiring layer.
- Fixing macro expansion errors in `DT_INST_FOREACH_STATUS_OKAY` usage.

## PR check-list

### DIFF
--- a/app/src/pointing/input_processor_temp_layer.c
+++ b/app/src/pointing/input_processor_temp_layer.c
@@ -91,7 +91,11 @@ K_MSGQ_DEFINE(temp_layer_action_msgq, sizeof(struct layer_state_action),
 
 static void deactivate_if_matching(const struct device *dev, uint8_t layer) {
     struct temp_layer_data *data = (struct temp_layer_data *)dev->data;
-    k_mutex_lock(&data->lock, K_FOREVER);
+    int ret = k_mutex_lock(&data->lock, K_FOREVER);
+    if (ret < 0) {
+        LOG_ERR("Error locking for deactivating %d", ret);
+        return;
+    }
     if (data->state.is_active && data->state.toggle_layer == layer) {
         update_layer_state(&data->state, false);
     }


app/src/pointing/input_processor_temp_layer.c
```diff
@@ -72,46 +72,59 @@ static void update_layer_state(struct temp_layer_state *state, bool activate) {
...
-
-    const struct device *dev = DEVICE_DT_INST_GET(0);
+static void deactivate_if_matching(const struct device *dev, uint8_t layer) {
     struct temp_layer_data *data = (struct temp_layer_data *)dev->data;
-
-    int ret = k_mutex_lock(&data->lock, K_FOREVER);
-    if (ret < 0) {
-        LOG_ERR("Error locking for updating %d", ret);
-        return;
+    k_mutex_lock(&data->lock, K_FOREVER);
```
petejohanson: Any particular reason you dropped checking the return of the lock?

petejohanson (changes requested):
Generally seems reasonable, just one minor comment on the mutex use. Thanks!

app/src/pointing/input_processor_temp_layer.c
nobyk: @petejohanson Thank you for your review. I've added the missing return check and logging.

PR #3248, opened 2026-02-18
feat(battery): implement power_on_sample_delay_us

veypatch:
Zephyr 4.1 introduces a `power-on-sample-delay-us` property in the voltage divider. (https://docs.zephyrproject.org/4.1.0/build/dts/api/bindings/iio/afe/voltage-divider.html) This PR implements this in the voltage divider driver used in zmk. This is useful if you're using extra components for the battery voltage reading which may take a bit more time then usual to stabilize/power up (like a MOSFET). 

As the init time dropped from 10ms to 100us because of this change, I made sure to test this on a nice!nano. The voltage/percentage readings are the same as before this PR.

## PR check-list

### DIFF
--- a/app/module/drivers/sensor/battery/battery_voltage_divider.c
+++ b/app/module/drivers/sensor/battery/battery_voltage_divider.c
@@ -26,6 +26,7 @@ struct bvd_config {
     struct gpio_dt_spec power;
     uint32_t output_ohm;
     uint32_t full_ohm;
+    uint32_t sample_delay_us;
 };
 
 struct bvd_data {
@@ -59,7 +60,7 @@ static int bvd_sample_fetch(const struct device *dev, enum sensor_channel chan) 
     }
 
     // wait for any capacitance to charge up
-    k_sleep(K_MSEC(10));
+    k_sleep(K_USEC(drv_cfg->sample_delay_us));
 #endif // DT_INST_NODE_HAS_PROP(0, power_gpios)
 
     // Read ADC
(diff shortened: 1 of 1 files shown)


joelspadin:
I think we were wanting to deprecate our voltage divider driver in favor of just using the upstream one. I don't remember why we haven't yet updated all the existing boards to use the upstream driver though.

> I made sure to test this on a nice!nano. 

I don't think that actually tests this change. Nice!nano v2 [doesn't use this driver at all](https://github.com/zmkfirmware/zmk/blob/main/app/boards/nicekeyboards/nice_nano/nice_nano_nrf52840_zmk_2_0_0.overlay#L19), and v1 uses it but doesn't use the `power-gpios` property. https://github.com/zmkfirmware/zmk/blob/main/app/boards/nicekeyboards/nice_nano/nice_nano_nrf52840_zmk_1_0_0.overlay#L18

veypatch:
> I don't think that actually tests this change. 

Yep, you're absolutely right, I don't know how I forgot that... But I've been testing this on a custom board which does use an actual voltage divider as well, which has been working fine.

> I don't remember why we haven't yet updated all the existing boards to use the upstream driver though.

Wouldn't we then miss the actual voltage to percentage conversion? As the upstream drivers doesn't do this as far as I can see and the current ZMK driver does.

joelspadin:
https://github.com/zmkfirmware/zmk/blob/main/app/src/battery.c#L87 

You have to configure it to fetch as a voltage instead of state of charge, and then it does the same conversion afterwards.

veypatch:
Ah right, didn't know that was an option. I'll give it some testing next week. I'll close this for now. I assume a preferred PR would be to switch all existing boards to use the upstream driver and update the docs (assuming this doesn't break anything.)

joelspadin:
Yeah, I think that would be a better solution.

PR #3257, opened 2026-02-25
docs: Compare quick-tap-ms with QMK.

aadcg:
While creating my first keymap on ZMK, I took the value from my QMK configuration assuming that it would behave similarly. For a while, I thought the feature was broken on ZMK.

I added this small note to account for that pitfall.

### DIFF
--- a/docs/docs/keymaps/behaviors/hold-tap.mdx
+++ b/docs/docs/keymaps/behaviors/hold-tap.mdx
@@ -333,6 +333,10 @@ By default this behavior is disabled.
 };
 ```
 
+#### Comparison to QMK
+
+QMK has a parameter called `QUICK_TAP_TERM`, but its semantics are slightly different. In QMK, it is the time interval between releasing the first key and pressing the second key whereas in ZMK, it is the time interval between pressing the first and second keys. Therefore you might want to use a larger value for `quick-tap-ms` compared to the value used in QMK.
+
 ### `require-prior-idle-ms`
 
 If a hold-tap is pressed within `require-prior-idle-ms` of another non-modifier _key_ (not behavior), then the hold-tap will always resolve in a tap.


docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -333,6 +333,10 @@ By default this behavior is disabled.
 };
 ```
 
+#### Comparison to QMK
+
+QMK has a similar parameter called `QUICK_TAP_TERM`, but you'll probably need a higher value to get similar behavior.
```
caksoylar: Can you make this precise? What's the difference that it requires a different value?

caksoylar (commented):
If we can make a precise statement then I don't mind adding it.

docs/docs/keymaps/behaviors/hold-tap.mdx
aadcg: In my QMK config, I set it to 85. 

In ZMK, that value is too low, and gave me the wrong impression that the feature wasn't working. It may be due to bluetooth latency, but I'm not sure.

From rough eyeballing, I'd say that these 2 variables are roughly related by a factor of 1.5.

Does this ratio make it precise? Thanks.

caksoylar: I am more interested in why it is different, e.g. (making it up, haven't checked) "ZMK checkes the interval between two key presses, while QMK checks the interval between previous key release and next key press".

aadcg: Your intuition is correct!

I checked the QMK behavior at https://docs.qmk.fm/tap_hold#quick-tap-term.

docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -333,6 +333,10 @@ By default this behavior is disabled.
 };
 ```
 
+#### Comparison to QMK
+
+QMK has a parameter called `QUICK_TAP_TERM`, but its semantics are slightly different.  In QMK, it is the time interval between releasing the first key and pressing the second key whereas, in ZMK, it is the time interval between pressing the first and second keys.
```
caksoylar:
```suggestion
QMK has a parameter called `QUICK_TAP_TERM`, but its semantics are slightly different.  In QMK, it is the time interval between releasing the first key and pressing the second key whereas in ZMK, it is the time interval between pressing the first and second keys. Therefore you might want to use a larger value for `quick-tap-ms` compared to the value used in QMK.
```

aadcg:
Thanks for the review @caksoylar. Should be ready to be merged.

aadcg:
I'm not sure why the clang linter is complaining since this PR only edits a mardown file.

snoyer:
> I'm not sure why the clang linter is complaining since this PR only edits a mardown file.

Looks like the markdown linter is not happy about the double space here:
```
semantics are slightly different.  In QMK,
                                 ^^
```

aadcg:
> > I'm not sure why the clang linter is complaining since this PR only edits a mardown file.
> 
> Looks like the markdown linter is not happy about the double space here:
> 
> ```
> semantics are slightly different.  In QMK,
>                                  ^^
> ```

Good catch, thanks. Too many years of Emacs indoctrination.

Should be ready to merge.

PR #3260, opened 2026-02-27
docs: Don't use deprecated variable names.

aadcg:
I only updated it in the docs. I can do it for the whole codebase too, just let me know. Thanks.


caksoylar:
Good find, thanks!

PR #3264, opened 2026-03-01
feat(underglow): add update hook called immediately before writing to strip

jkorinth:
Adds a hook that is called immediately before pixels are written to the led strip in `rgb_underglow.c`. The hook is implemented by a `weak` function to allow ZMK modules to overwrite it . The hook can perform modifications of the `struct led_rgb` array, e.g., to deactivate specific LEDs (see e.g., [zmk-rgb-underglow-mask-leds](https://github.com/jkorinth/zmk-rgb-underglow-mask-leds) for a working example). 

## PR check-list

### DIFF
--- a/app/include/zmk/rgb_underglow.h
+++ b/app/include/zmk/rgb_underglow.h
@@ -6,6 +6,9 @@
 
 #pragma once
 
+#include <zephyr/types.h>
+#include <zephyr/drivers/led_strip.h>
+
 struct zmk_led_hsb {
     uint16_t h;
     uint8_t s;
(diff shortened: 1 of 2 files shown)


caksoylar:
I like adding an approach similar to custom status screens in order to allow for custom underglow effects. (I don't know if this is the best architectural approach for it, other contributors would have a better idea.) That being said, wouldn't such custom effects also need the timing information, such as timestamp or time since last update?

It would also be nice if this was enabled by selecting a "custom" effect type, rather than overriding whatever the user has currently selected.

jkorinth:
@caksoylar I think getting getting the timestamps would be more accurate if done from the hook side. Should not be an issue to maintain the last update timestamp there, I think.

But I didn't mean to override the effect, just adapt to specific hardware. E.g., selectively reduce brightness, fix leds that have a color tint/bias toward one color, or turn them off completely. I frequently need to do this for battery life optimization. Take the Reviung41 for example: the central top-side led is number 11 in the strip, so you can't easily turn off all the bottom-side leds but keep the center on.

PR #3271, opened 2026-03-03
fix(ci): Skip build on empty build matrix.

### LINKED ISSUE #3179: User config build workflow fails when build matrix is empty
When you create a user config repo using https://github.com/zmkfirmware/unified-zmk-config-template as a template, the first thing you see is your new repo. The second is a failed build action.

The build matrix we generate from an empty build.yaml is an empty string, so the build fails with an error like:

> Build ZMK firmware
Error when evaluating 'strategy' for job 'build'. zmkfirmware/zmk/.github/workflows/build-user-config.yml@v0.3 (Line: 50, Col: 15): Unexpected value ''

When the build matrix is empty, we should bail and mark the build as skipped (if possible, or successful otherwise) along with an annotation explaining that nothing was built, ideally with a link to https://zmk.dev/docs/user-setup#add-a-keyboard or other documentation on how to add a keyboard to the build.

aadcg:
Closes #3179.

----

I tested the CI [here](https://github.com/aadcg/test-zmk-config/actions).

It skips the build and the CI routine returns as successful. I'm not sure how to retrieve the log message from the skipped job, but I can look into it.

CC @joelspadin.

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -30,17 +30,28 @@ jobs:
     name: Fetch Build Keyboards
     outputs:
       build_matrix: ${{ env.build_matrix }}
+      has_valid_build_matrix: ${{ steps.fetch.outputs.has_valid_build_matrix }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
       - name: Fetch Build Matrix
+        id: fetch
         run: |
-          echo "build_matrix=$(yq -oj -I0 '${{ inputs.build_matrix_path }}')" >> $GITHUB_ENV
-          yq -oj "${{ inputs.build_matrix_path }}"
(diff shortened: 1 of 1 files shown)


caksoylar:
Maybe a notice message annotation would be better so the message is surfaced to the main build page? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-notice-message

Similar to how `::warning` is used in https://github.com/zmkfirmware/zmk/blob/main/.github/workflows/build-user-config.yml.

aadcg:
> Maybe a notice message annotation would be better so the message is surfaced to the main build page?

@caksoylar, sounds good. I've pushed an updated commit with that change.

To check how the notice message looks like, please see https://github.com/aadcg/test-zmk-config/actions/runs/23148617644.

Sorry for the late reply.

caksoylar:
Thanks for the fix!

PR #3286, opened 2026-03-19
Feat: BLE prevent sleep while active connected and Feat: allow sleep while USB powered

newx-it:
A solution for [issue 2198](https://github.com/zmkfirmware/zmk/issues/2198), wherein I changed the activity.c loop that checks for sleep to watch for the status of the active BT connection. When the flag is enabled and the active BT profile is connected, the check does not let the device sleep. If the flag is not enabled, the device acts the same as it would before this PR. 

While I was working on testing the flag, I decided to add a flag to allow the device to sleep while USB power is present. This was primarily so i could test over USB debugging, but for users who are willing to allow their device to sleep while USB is powered (like split or dongle users) this should have non-negligible effects on battery life (from the zmk power profiler) for the peripherals, which may not be plugged in.

First time contributor so any and all advice is welcome and encouraged!

## PR check-list

### DIFF
--- a/.github/workflows/build-user-config.yml
+++ b/.github/workflows/build-user-config.yml
@@ -30,17 +30,28 @@ jobs:
     name: Fetch Build Keyboards
     outputs:
       build_matrix: ${{ env.build_matrix }}
+      has_valid_build_matrix: ${{ steps.fetch.outputs.has_valid_build_matrix }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
       - name: Fetch Build Matrix
+        id: fetch
         run: |
-          echo "build_matrix=$(yq -oj -I0 '${{ inputs.build_matrix_path }}')" >> $GITHUB_ENV
-          yq -oj "${{ inputs.build_matrix_path }}"
+          matrix_content=$(yq -oj -I0 '${{ inputs.build_matrix_path }}')
+
+          if [ -z "$matrix_content" ] || [ "$matrix_content" = "null" ]; then
+            echo "::notice file=${{inputs.build_matrix_path}},title=Empty build matrix file::To add a keyboard to the build, see https://zmk.dev/docs/user-setup#add-a-keyboard."
+            echo "has_valid_build_matrix=false" >> $GITHUB_OUTPUT
+          else
+            echo "build_matrix=$matrix_content" >> $GITHUB_ENV
+            echo "has_valid_build_matrix=true" >> $GITHUB_OUTPUT
+            echo "$matrix_content"
+          fi
 
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


app/src/activity.c
```diff
@@ -75,7 +79,16 @@ void activity_work_handler(struct k_work *work) {
...
 #if IS_ENABLED(CONFIG_ZMK_SLEEP)
-    if (inactive_time > MAX_SLEEP_MS && !is_usb_power_present()) {
+    bool prevent_sleep = !IS_ENABLED(CONFIG_ZMK_USB_ALLOW_SLEEP_WHILE_POWERED) && is_usb_power_present();
+    #if IS_ENABLED(CONFIG_ZMK_BLE_PREVENT_SLEEP_WHILE_ACTIVE_CONNECTED)
+    prevent_sleep = prevent_sleep ||
+        #if !IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
+            zmk_ble_active_profile_is_connected();
+        #else
+            zmk_split_bt_peripheral_is_connected();
+        #endif
```
snoyer, line 89:
```suggestion
        #if !IS_ENABLED(CONFIG_ZMK_SPLIT) || IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)
            prevent_sleep |= zmk_ble_active_profile_is_connected();
        #else
            prevent_sleep |= zmk_split_bt_peripheral_is_connected();
        #endif
```
maybe clearer to not split statements across preproc directives?

app/Kconfig
```diff
@@ -415,6 +415,15 @@ config ZMK_IDLE_SLEEP_TIMEOUT
     int "Milliseconds of inactivity before entering deep sleep"
     default 900000
 
+config ZMK_BLE_PREVENT_SLEEP_WHILE_ACTIVE_CONNECTED
+    bool "Enable keep awake for active profile"
```
snoyer, line 419: phrase in terms of preventing/allowing _sleep_ instead of keeping awake?

newx-it: Also a great catch, I think I spotted this one and forgot it unfortunately.

app/src/activity.c
newx-it: Definitely more readable, I like it.

newx-it:
I'm closing this pr in favor of another with cleaned up commit history and fixes applied, it's here: [PR 3288](https://github.com/zmkfirmware/zmk/pull/3288)

snoyer:
You shouldn't keep opening new PRs for minor changes, it makes the review process harder and spams the PR list. You can rewrite the commit history in the same PR if you need to.

newx-it:
Noted, thanks for the info! Won't happen again 😎

PR #3287, opened 2026-03-19
docs: Initial release notes/migration guide support

petejohanson:
Add the initial infrastructure for formal release notes and migrations guides, with a retroactive v0.3 set of docs as a start.

## PR check-list


docs/docs/releases/notes/v0.3.md
```diff
@@ -0,0 +1,21 @@
...
+## API Changes
+
+:::note
+
+ZMK does not currently offer any API stability guarantees. Using third party modules with additonal functionality is supported, from a build/design perspective, but any particular upgrade may change APIs that results in changes needed to those third-party modules. If using any such modules, and experiencing build failures when updating, reach out to the module author for assistance.
+
+:::
+
+- Added a BLE function for getting profile addresses in [#2992](https://github.com/zmkfirmware/zmk/pull/2992)
+- Add a source info to the battery event in [#2901](https://github.com/zmkfirmware/zmk/issues/2901)
```
caksoylar: Not sure it is worth it at this point, but the wired split PR's refactor moved a few headers around and renamed some functions: https://github.com/zmkfirmware/zmk/pull/2766/changes/4a2189d707986eb923ed2935f9959095942ad070

docs/docs/releases/migration-guides/v0.3.md
```diff
@@ -0,0 +1,81 @@
+---
+title: ZMK v0.3 Migration Guide
+sidebar_label: v0.3
+---
+
+## Change Summary
```
caksoylar, line 7: Can we either add a `description` metadata here or a introductory sentence? Otherwise the doccard in category index page shows "Change summary":

Image

docs/docs/releases/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Releases
+---
+
+ZMK has moved to a formal release process that enables and encourages users to use an explicit released version of ZMK. This is a significant departure from the previous practice of builds tracking the latest code found in our `main` branch.
+
+This is one of many important steps as ZMK works towards a more reliable, repeatable experience for users. Users are _strongly_ encouraged to adjust their builds to [pin](/blog/2025/06/20/pinned-zmk) to the [latest release](/docs/category/release-notes) to avoid disuption from development changes.
```
joelspadin, line 7: ZMK CLI has a `zmk version <tag>` command that will update both the manifest and actions workflow files. It looks like I forgot to document that on /docs/zmk-cli when I added it though. I think it would be good to document that as one way you can pin a specific version, alongside the steps for doing it manually.

docs/docusaurus.config.js
```diff
@@ -202,6 +202,22 @@ module.exports = {
...
+          async sidebarItemsGenerator({
+            defaultSidebarItemsGenerator,
+            ...args
+          }) {
+            const sidebarItems = await defaultSidebarItemsGenerator(args);
+            const filtered = sidebarItems.filter(
+              (i) =>
+                !(
+                  args.docs.find((d) => d.id == i.id).frontMatter.unlisted ??
+                  false
```
joelspadin: `!(thing ?? false)` is equivalent to `!thing`, since `!null == !undefined == true`

If the intent is to fall back to false if `args.docs` doesn't include an item, then you need a `?.` at the first thing that could be undefined
```js
args.docs.find(...)?.frontMatter.unlisted ?? false
```
though the fact that this isn't a problem right now tells me that `args.docs` is probably guaranteed to contain every item?

docs/docusaurus.config.js
```diff
@@ -202,6 +202,22 @@ module.exports = {
...
           // Please change this to your repo.
           editUrl: "https://github.com/zmkfirmware/zmk/edit/main/docs/",
+          async sidebarItemsGenerator({
+            defaultSidebarItemsGenerator,
+            ...args
+          }) {
+            const sidebarItems = await defaultSidebarItemsGenerator(args);
+            const filtered = sidebarItems.filter(
+              (i) =>
+                !(
```
joelspadin: Consider pulling part of this out into a helper function with a name that makes it clearer what the code is intended to do. Some ideas for that:

```js
// Top-level helper function
function getFrontMatter(docs, item) {
  return docs.find((d) => d.id === item.id);
}
...
const sidebarItems = ...
const filtered = sidebarItems.filter((item) => !getFrontMatter(args.docs, item).unlisted);
```

```js
// Lambda that captures args.docs so it doesn't need to be passed in as an argument
const getFrontMatter = (item) => args.docs.find((d) => d.id === item.id;

const sidebarItems = ...
const filtered = sidebarItems.filter((item) => !getFrontMatter(item).unlisted);
```

```js
// Pull the whole thing into a lambda
const isListed = (item) => !args.docs.find((d) => d.id === item.id).frontMatter.unlisted;

const sidebarItems = (await defaultSidebarItemsGenerator(args)).filter(isListed);
```

docs/docs/releases/notes/v0.3.md
```diff
@@ -0,0 +1,21 @@
+---
+title: ZMK v0.3 Release Notes
+sidebar_label: v0.3
+---
+
+ZMK v0.3 introduces initial support for [wired split](https://zmk.dev/docs/features/split-keyboards#full-duplex-wired-uart) devices as well as various bug fixes and minor improvements. For the complete changelog, see the [v0.3.0 release on GitHub](https://github.com/zmkfirmware/zmk/releases/tag/v0.3.0).
+
+## Breaking Changes
```
joelspadin: It feels like there's a lot of overlap between the release notes and migration guides pages for the same release. Both list breaking changes. One of them lists changes needed to user configs, while the other lists changes that might be needed to modules, which feels like something that belongs more in a migration guide?

IMO, we should either combine these into single pages for each release, or we should at the very least remove any duplicate info and make sure that every release notes page links to the matching migration guide page.

docs/docusaurus.config.js
petejohanson: We specifically need to handle the scenario we don't have a value for `unlisted` in the frontMatter. In that scenario, unlisted should be false. So we need to default to `false` *only* if the value isn't explicitly set/present in the frontMatter. *Then* we need to invert the value, for the expected filtering.

docs/docs/releases/index.md
petejohanson: Just pushed a small addition in the v0.3 migration guide, if you could please verify it matches the expectation.

docs/docs/releases/migration-guides/v0.3.md
petejohanson: Added a description to the front matter.

docs/docs/releases/notes/v0.3.md
petejohanson: Fair. Added a bullet item.

docs/docs/releases/notes/v0.3.md
petejohanson: So.... Yeah, I've been struggling with this. In particular, this v0.3 doesn't have a particularly significant set of release notes, so it feels a bit silly to have two docs. For *larger* releases, I think having this split will help bring clarity for folks who just need to go through the upgrade steps. I will move the module change not to the upgrade notes for more consistency, as a bare minimum here.

docs/docusaurus.config.js
petejohanson: Tweaked a bit.

docs/docusaurus.config.js
joelspadin: If `frontMatter` doesn't have a value for `unlisted`, then `frontMatter.unlisted` is `undefined`. The only cases where `??` evaluates the right-hand side are when the left-hand side is `null` or `undefined`, and `!null` and `!undefined` are both true, so `!(frontMatter.unlisted ?? false)` is equivalent to `!frontMatter.unlisted`.

petejohanson: Indeed. My tired brain somehow screwed up my testing of that last night. Changed.

docs/docs/releases/index.md
nmunnich, line 7: I think a note saying that if you created your zmk-config on X date or later then your config is already pinned would be convenient here.

docs/docs/releases/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Releases
+---
+
+ZMK has moved to a formal release process that enables and encourages users to use an explicit released version of ZMK. This is a significant departure from the previous practice of builds tracking the latest code found in our `main` branch.
```
nmunnich, line 5: I'm not a fan of time-based statements like these in our docs. I'd prefer to just have a statement about the current state of things, rather than comparing with the previous state. Also consider that we've had versioning for 1.5 years now, so I wouldn't really describe it as "new" anymore.

nmunnich (commented):
Looking at the overall structure, I think I would move some things around. While I agree with the presence of both migration guides and release notes, I also agree with Joel that having a migration guide for v0.3 feels unnecessary. What I would do is

- move the bits on API changes and breaking changes under the release notes
- move the migration guide under the releases index page as a "This is how you migrate when there are no breaking changes" type thing
- have a migration guide for v0.4 only, and remove the one for v0.3.

PR #3304, opened 2026-03-27
fix(backlight): Retrieve only brightness from settings when CONFIG_ZMK_BACKLIGHT_ON_START disabled

kurtis-lew:
On startup, when the settings subsystem loads the most recent backlight brightness and on-state, it overrides the effect of the `CONFIG_ZMK_BACKLIGHT_ON_START` Kconfig.

This PR creates congruency between the settings subsystem and this Kconfig by not calling `zmk_backlight_update()` and setting the backlight state to `false` when `CONFIG_ZMK_BACKLIGHT_ON_START=n`.

Tested on custom board.

### DIFF
--- a/app/src/backlight.c
+++ b/app/src/backlight.c
@@ -67,7 +67,11 @@ static int backlight_settings_load_cb(const char *name, size_t len, settings_rea
 
         int rc = read_cb(cb_arg, &state, sizeof(state));
         if (rc >= 0) {
+#if CONFIG_ZMK_BACKLIGHT_ON_START
             rc = zmk_backlight_update();
+#else
+            state.on = false;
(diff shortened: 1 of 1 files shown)


caksoylar:
I am pretty sure this applies to underglow as well, and has been a known behavior for some time (so is documented as such). I wouldn't have an objection to changing the behavior, but I'd like to hear opinions from other maintainers.

petejohanson:
Given this is established behavior, I'm wary to change this. Especially because this would now *ignore* the saved value for anyone who received a build with this Kconfig set but used a behavior to turn *off* the backlight, expecting that to persist until explicitly turned back on.

Can you help me understand your use case, so we can maybe come up with a fix that isn't a breaking change?

PR #3335, opened 2026-04-23
Clarify combos interact with hold-tap's hold-while-undecided

peterjc:
This caught me out with a rarely used four-key combo for my boot-loader with a long timeout causing a matching delay to the hold behaviour. Issue diagnosed on ZMK Discord.

## PR check-list

### DIFF
--- a/docs/docs/keymaps/behaviors/hold-tap.mdx
+++ b/docs/docs/keymaps/behaviors/hold-tap.mdx
@@ -438,7 +438,7 @@ If enabled, the hold behavior will immediately be held on hold-tap press, and wi
 ```
 
 :::warning[Interactions with combos]
-If the key is _also_ used in any combos, the `hold-while-undecided` behavior isn't immediate with key-down, but waits until all those combo timeouts have expired.
+If the key position that has the hold-tap assigned is _also_ used in any combos, the `hold-while-undecided` behavior isn't immediate with key-down, but waits until all those combo timeouts have expired.
 :::
 
 #### `hold-while-undecided-linger`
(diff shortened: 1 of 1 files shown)


docs/docs/keymaps/behaviors/hold-tap.mdx
```diff
@@ -437,6 +437,10 @@ If enabled, the hold behavior will immediately be held on hold-tap press, and wi
 };
 ```
 
+:::warning[Interactions with combos]
+If the key is _also_ used in any combos, the `hold-while-undecided` behavior isn't immediate with key-down, but waits until all those combo timeouts have expired.
```
caksoylar:
```suggestion
If the key position that has the hold-tap assigned is _also_ used in any combos, the `hold-while-undecided` behavior isn't immediate with key-down, but waits until all those combo timeouts have expired.
```

nmunnich:
Thanks for documenting this. I do wonder if there might be some better place for us to organise/collect these sort of "edge case" interactions. The hold-tap page is lengthy enough as-is, and I can think of a few more edge case stuff that we could add...

peterjc:
I wasn't sure if this was worth logging as a bug too - you could have the hold action (in my usage shift) while the combos have yet to be decided.

That there are other caveats doesn't really surprise me.

caksoylar:
> I wasn't sure if this was worth logging as a bug too - you could have the hold action (in my usage shift) while the combos have yet to be decided.

I don't know if I'd call it a bug; combo detection necessarily happens earlier to determine whether you are pressing a single key or one of the combos, so until that resolves, the behavior itself does not get to capture the key press.

caksoylar (approved):

caksoylar:
I'll merge this, but if we want to reorg to a specific place like Nick suggested we can move it later.

Issue #1685, opened 2023-02-25
Can't get SHIFT-SPACE to issue TAB


yanshay:
I am trying to get SPACE function as TAB when pressing SHIFT-SPACE. 
I am using mod morph as follows:
```
    spc_tab: space_tab {
            compatible = "zmk,behavior-mod-morph";
            label = "SPACE-TAB";
            #binding-cells = <0>;
            bindings = <&kp SPACE>, <&kp TAB>;
            mods = <(MOD_LSFT)>;
        };
```
However it issues a SHIFT-TAB and not a TAB.
This is what I see in karabiner event viewe:
```
  {
    "type": "down",
    "name": {"key_code":"left_shift"},
    "usagePage": "7 (0x0007)",
    "usage": "225 (0x00e1)",
    "misc": "flags left_shift"
  },
  {
    "type": "down",
    "name": {"key_code":"tab"},
    "usagePage": "7 (0x0007)",
    "usage": "43 (0x002b)",
    "misc": "flags left_shift"  <--
  },
  {
    "type": "up",
    "name": {"key_code":"left_shift"},
    "usagePage": "7 (0x0007)",
    "usage": "225 (0x00e1)",
    "misc": ""
  },
  {
    "type": "up",
    "name": {"key_code":"tab"},
    "usagePage": "7 (0x0007)",
    "usage": "43 (0x002b)",
    "misc": ""
  },
```
I even tried adding  `keep-mods = <0>;` and it didnt change a thing. 
so why is the mod being sent together with the key? it practically means I cant get to issue TAB using SHIFT-SPACE.

caksoylar:
This came up a couple of times in the Discord server, but only seems to be related to KA360 Pro somehow, e.g. https://discord.com/channels/719497620560543766/1055137812481060884/1055137812481060884. There is a corresponding issue at https://github.com/KinesisCorporation/Adv360-Pro-ZMK/issues/83. One thing I might try since those issues came up is [disabling NKRO](https://github.com/yanshay/Adv360-Pro-ZMK/blob/V2.0/config/boards/arm/adv360/adv360_left_defconfig#L53) since that is one difference I noticed between usual ZMK setups and KA360.

yanshay:
I tried disabling NKRO and it didn't help.

To make sure I disabled properly, I am building locally using podman and changed that flag from 'y' to 'n' on both left and right keyboards in the local files.

Any other ideas?

ReFil:
[...]

PR #3348, opened 2026-05-11
fix(mod-morph): send separate HID report after masked_modifiers_set

akira-toriyama:
## Summary

When a `mod-morph` fires with `masked_mods`, the modifier change set by
`zmk_hid_masked_modifiers_set()` is not flushed to the host on its own —
it's only included in the *next* HID report, which is normally produced
by the inner `&kp`'s press handler.

On macOS this causes the host to receive a single report containing both
the modifier release and the morphed key press at the same time. macOS
appears to evaluate key events before modifier changes within one
report, so e.g. `Ctrl+J` (mod-morphed to `DOWN`) is interpreted as
`Ctrl+DOWN` (Mission Control) instead of plain `DOWN`.

This change sends a HID report immediately after
`zmk_hid_masked_modifiers_set()` so the modifier release is observed by
the host **before** the morphed binding's key press arrives in a later
report. It mirrors the existing behavior gated by
`CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT` on the release path.

## Reproduction (before this fix)

- Host: macOS (any recent version)
- Keymap: `&mm DOWN J  LCTRL`  (Ctrl+J → DOWN via masked_mods = LCTRL)
- Hold `Ctrl`, then tap `J`
- Observed: Mission Control activates (== `Ctrl+DOWN`)
- Expected: cursor moves down (== plain `DOWN`)

## After this fix

- Same setup, Ctrl+J → plain `DOWN` as intended on macOS.
- Linux / Windows behavior unchanged in local testing.

## PR check-list

- [x] Branch has a clean commit history (single commit)
- [ ] Additional tests are included — *not added; mod-morph masked_mods
      timing is host-side observable rather than unit-testable in the
      existing harness. Happy to add a test if reviewers can point me to
      a suitable pattern.*
- [x] No new files; existing copyright headers untouched
- [ ] Pre-commit run — *not run locally (toolchain not installed);
      relying on CI. Diff is a single added line matching surrounding
      indentation.*
- [x] No documentation changes needed

### DIFF
--- a/app/src/hid.c
+++ b/app/src/hid.c
@@ -9,6 +9,7 @@
 LOG_MODULE_DECLARE(zmk, CONFIG_ZMK_LOG_LEVEL);
 
 #include <zmk/hid.h>
+#include <zmk/endpoints.h>
 #include <dt-bindings/zmk/modifiers.h>
 
 static struct zmk_hid_keyboard_report keyboard_report = {
@@ -281,14 +282,32 @@ int zmk_hid_masked_modifiers_set(zmk_mod_flags_t new_masked_modifiers) {
     masked_modifiers = new_masked_modifiers;
     zmk_mod_flags_t current = GET_MODIFIERS;
     SET_MODIFIERS(explicit_modifiers);
-    return current == GET_MODIFIERS ? 0 : 1;
+    int changed = current == GET_MODIFIERS ? 0 : 1;
+#if IS_ENABLED(CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT)
+    if (changed) {
+        int err = zmk_endpoint_send_report(HID_USAGE_KEY);
+        if (err < 0) {
+            LOG_ERR("Failed to flush masked modifier set (%d)", err);
+        }
+    }
+#endif
+    return changed;
 }
 
 int zmk_hid_masked_modifiers_clear(void) {
     masked_modifiers = 0;
     zmk_mod_flags_t current = GET_MODIFIERS;
     SET_MODIFIERS(explicit_modifiers);
-    return current == GET_MODIFIERS ? 0 : 1;
+    int changed = current == GET_MODIFIERS ? 0 : 1;
+#if IS_ENABLED(CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT)
+    if (changed) {
+        int err = zmk_endpoint_send_report(HID_USAGE_KEY);
+        if (err < 0) {
+            LOG_ERR("Failed to flush masked modifier clear (%d)", err);
+        }
+    }
+#endif
+    return changed;
 }
 
 int zmk_hid_keyboard_press(zmk_key_t code) {


caksoylar:
Are you sure you don't have anything intercepting key presses on macOS? This typically only happens in scenarios like Karabiner Elements with Modify Events toggled on, see the last info block: https://zmk.dev/docs/keymaps/behaviors/mod-morph#advanced-configuration

If this is something that happens on macOS without any interceptors, it is a new development and we haven't heard about it from anyone else yet.

peterjc:
Even if rather than fixing a macOS limitation this fixes the Karabiner Elements limitation, it will be welcome to me - and likely helps with other corner cases with keyboard manipulating software?

akira-toriyama:
Thanks @caksoylar for pushing back — you were right to question this. I retested with the matrix your comment implied:

| Karabiner Elements | This PR | `Ctrl+P` (mod-morph to `UP_ARROW`) |
|---|---|---|
| running | not applied | Mission Control fires (the original report) |
| running | applied     | `UP` only ✅ |
| disabled for device | not applied | `UP` only ✅ |
| disabled for device | applied     | `UP` only ✅ |

So this is not a macOS issue per se — it requires Karabiner Elements (or presumably any HID interceptor that reshapes reports) in the path to manifest. Apologies for the misleading framing in the original PR description.

That said, I'd like to ask maintainers to consider the angle @peterjc raised: even if this is "just" a workaround for Karabiner Elements, Karabiner is widely used in the macOS ZMK community, and the failure mode is hard to diagnose from a user's perspective (a mod-morph silently triggers an OS shortcut with no obvious cause). The change is gated behind the existing `CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT` so default behavior is unchanged — only users who opt in pay the extra HID report.

Happy to:

- Update the PR description and commit message to accurately frame this as "improves interaction with HID interceptors that reshape modifier/key reports" rather than "fixes a macOS issue"
- Investigate what Karabiner Elements does differently (likely how its virtual HID device consolidates or reorders reports) and add that context, if useful for the discussion
- Close the PR and keep the patch on my own fork if maintainers consider this out of scope

Let me know which direction you'd prefer.

nmunnich:
While that's a fair point, I wonder if you've considered trying to tackle this on Karabiner-Elements side? Personally I would be of the opinion that this is a bug in how Karabiner Elements behaves, so should really be fixed on their side rather than introducing workarounds in ZMK.

That said, if it turns out that there is a specific reason that Karabiner Elements does things this way to the point that it should not be considered a bug, I would be more accepting of a solution like this one.

akira-toriyama:
@nmunnich That's a fair principled position. Looking into it more carefully, here are the points that I think change the practical picture:

**This failure mode is structurally hard to fix on the Karabiner side**

1. **Same-shape bug, already documented in this repo**: [zmkfirmware/zmk#1685](https://github.com/zmkfirmware/zmk/issues/1685) shows Karabiner's EventViewer emitting `Tab key_down (with left_shift flag)` when ZMK released Shift in the same report Tab was pressed. That's the exact shape of the failure here — a non-modifier keycode arriving carrying a modifier that ZMK had already cleared. The reporter included a raw EventViewer trace showing Karabiner is the layer that adds the stale modifier flag.

2. **Already filed against Karabiner with no technical engagement**: [pqrs-org/Karabiner-Elements#3420 "Karabiner breaks zmk keyboard events"](https://github.com/pqrs-org/Karabiner-Elements/issues/3420) has been open without a technical response from the maintainer. Same symptom class.

3. **It's by design, not an obvious bug**: Karabiner's [input event pipeline](https://karabiner-elements.pqrs.org/docs/manual/misc/event-modification-chaining/) intentionally decomposes incoming HID reports into individual events, tracks modifiers in its own `modifier_flag_manager`, and re-synthesizes reports for its virtual HID device. The single-report atomicity ZMK relies on isn't preserved because the Complex Modifications model operates at per-event granularity. Restoring report-boundary fidelity would push against that core design.

4. **ZMK docs already concede this as an interop boundary**: the [mod-morph "Advanced configuration"](https://zmk.dev/docs/keymaps/behaviors/mod-morph) section already tells affected users to disable "Modify events" in Karabiner. The project already treats this as a known coexistence issue, not a transient Karabiner regression.

**This PR fits an existing ZMK precedent**

`CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT` was [originally added](https://github.com/zmkfirmware/zmk/commit/10d03ca46c84a2530701ef82e28ec018d34df905) to work around the same category of issue on the Linux side (`wev` perceiving the implicit modifier release before the keycode release). This PR adds press-side coverage under the *same* Kconfig with the *same* opt-in semantics — users who don't run an interceptor pay nothing.

Happy to:

- File a fresh, minimal Karabiner-side issue with the EventViewer trace if you'd like the maintainer to weigh in. Per #3420 I don't expect that to be the practical fix path, but the data point may help.
- Reframe this PR's description as "extends `CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT` to the press path for HID interceptor compatibility" rather than the macOS-specific framing in the original description, if that's more palatable.

Either way, grateful for guidance on what would unblock review.

---

🤖 The cross-referencing of Karabiner internals, related GitHub issues, and the ZMK Kconfig precedent above was researched with the assistance of [Claude Code](https://claude.com/claude-code).

urob:
I don't have a strong opinion on whether to workaround a bug on the Karabiner side. But in case we do, it seems cleaner to directly patch `zmk_hid_masked_modifiers_set` and `*_clear`. 

I also wonder why the issue isn't emerging with regular implicit modifiers use? Quickly scanning hid.c it doesn't seem we send a separate report there either?

akira-toriyama:
@urob 

(1) Done — the change is now a single-file diff in `app/src/hid.c` that flushes a report from `zmk_hid_masked_modifiers_set` / `_clear` whenever the effective modifier byte actually changes, gated on `CONFIG_ZMK_HID_SEPARATE_MOD_RELEASE_REPORT`. Applies symmetrically to press and release.

(2) Verified empirically with Karabiner EventViewer:

A key bound to `&kp EXCL` (implicit `Shift+1`) emits a clean sequence — `left_shift key_down (flags: left_shift)` → `1 key_down (flags: left_shift)` → `1 key_up (flags: left_shift)` → `left_shift key_up`. No stale modifier flag at any point, even though ZMK sent a single report containing both the Shift bit and the `1` keycode.

Compare to the mod-morph (`masked_mods`) case before this PR: a single report containing both "Ctrl released + DOWN pressed" produces `DOWN key_down (flags: left_control)` — the key arrives carrying the modifier that was just supposed to be released.

The asymmetry is **additive vs. subtractive within one report**:

- `zmk_hid_implicit_modifiers_press` *adds* a modifier in the same report as the key. Karabiner emits modifier-first → key-second, so the key naturally carries the (now-active) modifier flag. No conflict.
- `zmk_hid_masked_modifiers_set` *removes* a held modifier in the same report as the morphed key. Karabiner appears to snapshot the prior modifier byte when emitting the key event, so the key carries the (about-to-be-released) modifier flag. Race.

That matches why this PR only needs to flush on the masked-modifier path, and why no analogous issue is observed for implicit modifiers.

Let me know if you'd like the PR title/description updated to reflect the new hid-level framing rather than the mod-morph-specific one.

Issue #3361, opened 2026-05-23
docs: new user flow on website


Genteure:
IMHO the way the website is structured is confusing for new users. There's no clear path from the home page to working keyboard firmware.

Clicking "Getting Started" on the home page doesn't lead to a page with useful next step instructions. It would be better if there's a clear and obvious path to follow from the home page to having compiled binary on disk.

From the perspective of two types of users:

### Keyboard Designers / DIYers

The first pieces of information I'm looking for as a keyboard designer would be 1. what controller/chip can ZMK run on, and 2. how to configure and compile ZMK for my custom keyboard.

The answer to the first question is somewhat buried in <https://zmk.dev/docs/hardware> with a not too easy to read format. Without any prior knowledge about ZMK and Zephyr in general and without reading some explanation first, it's really hard to decipher what's "board" or "shield" or even "board variant". While that info exists on the site it requires clicking away and reading through the page.

The answer to the second question is <https://zmk.dev/docs/development/hardware-integration/new-shield>, which is not linked from in an obvious place.

### End Users

I'm not really sure about what I would be looking for as an end user.

From what I've seen, most vendors have some kind of getting started guide and some kind of template repository users can fork, so I'm guessing most people just need to look up keymap related stuff.

nmunnich:
I think we could further separate end users:
- Users who just need some firmware to flash so that they can start using ZMK Studio
- Users who want to customise their keymap/configuration in such a way that they are unable to do with Studio (e.g. switching to the internal oscillator of the nrf52)

For the first group, it would be very convenient to be able to just give them some prebuilt firmware to flash. While I'm sure some vendors do this, for various reasons a user may not be able to easily obtain such files in a trustworthy manner.

The second group needs to obtain a zmk-config and would usually nowadays start by needing to add a module to their config to build from there. Hence we get all of the forks of random repositories which already have the module bits setup for them. 

I think all types of users have a somewhat painful onboarding experience for different reasons, and we could do better for all of them.

caksoylar:
Thanks for detailing this, I agree with the overall argument and most of the details. As to what to do about it:

I like the approach in #3363 (and agree with points raised in Nick's review). The other piece that is missing is, IMO, is a sort of flowchart (not literally) in the Getting Started page. This should target the audiences separately, and point to the right section of the docs. Off the top of my head, combining with Nick's comment:
1. Users looking to use ZMK Studio -> link to Studio page
[...]

PR #3363, opened 2026-05-24
docs: reorganise supported hardware page

Genteure:
Related to #3361 

- Moved onboard section below composite, since composite keyboards seem more common.
- Added a few links to hardware integration pages to clearly signal "you can make your own".
- Removed "contributing" since it's not really applicable for shields anymore, and the "new board" page have all the necessary info on licensing and merging already.

- Made the "board / shield" link a dedicated admonition to bring more attention, since the terminology seems to be a common source of confusion for new users. 

- Removed Zephyr 4.1 upgrade note.
  IMHO, this note can be confusion. For new readers reading the page in current version, they don't need to know what changed _before_ the version they're starting from. For readers coming from a previous version, surely they would've read either change log or migration guide already.

- Moved unsupported boards from metadata to the page. Also removed some sections of interconnect descriptions.
  The `pro_micro` interconnect description mentions "Elite-C" which is an unsupported board.
  IMO all mentions of specific boards could be removed because right below the description is the list of boards.

-----

Preview: https://deploy-preview-3363--zmk.netlify.app/docs/hardware

-----

## PR check-list

N/A

### DIFF
--- a/docs/docs/hardware.mdx
+++ b/docs/docs/hardware.mdx
@@ -12,11 +12,6 @@ import { groupedMetadata } from "@site/src/components/hardware-utils";
 
 export const toc = [
   {
-    value: "Onboard Controller Keyboards",
-    id: "onboard",
-    level: 2,
-  },
-  {
     value: "Composite Keyboards",
     id: "composite",
     level: 2,
@@ -29,24 +24,31 @@ export const toc = [
       level: 3,
     })),
   {
-    value: "Other Hardware",
-    id: "other-hardware",
+    value: "Onboard Controller Keyboards",
+    id: "onboard",
     level: 2,
   },
   {
-    value: "Contributing",
-    id: "contributing",
+    value: "Other Hardware",
+    id: "other-hardware",
     level: 2,
   },
 ];
 
 With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets,
 including but not limited to Nordic nRF52, Raspberry Pi RP2040/RP2350, most ST STM32 MCUs, and Microchip SAMD21.
-That being said, there are specific [boards / shields](hardware-integration/index.mdx#boards--shields) that have been implemented and tested by the ZMK contributors, listed below.
+That being said, there are specific boards / shields that have been implemented and tested by the ZMK contributors, listed below.
 
-:::note
+ZMK doesn't support boards with AVR 8-bit processors, such as the ATmega32U4, because Zephyr™ only
+supports 32-bit and 64-bit platforms. As a result, controllers like the SparkFun Pro Micro, Elite-C,
+and Arduino Uno Rev3 are **NOT** supported by ZMK.
 
-With the [upgrade to Zephyr 4.1](/blog/2025/12/09/zephyr-4-1#zmk-board-variant), the ZMK project has moved all in-tree boards to use a `zmk` [board variant](https://docs.zephyrproject.org/4.1.0/glossary.html#term-variant), for consistency when distinguishing from stock boards that are actually in upstream Zephyr.
+Check out the [Hardware Integration](development/hardware-integration/index.mdx) section for more info on how to use ZMK on custom keyboards.
+
+:::info[Boards and Shields]
+
+ZMK uses the Zephyr concepts of "boards" and "shields" to refer to different parts of a keyboard build that in turn get combined during a firmware build.
+Please see the [explainer on boards & shields](development/hardware-integration/index.mdx#boards--shields) for more details.
 
 :::
 
@@ -57,8 +59,3 @@ With the [upgrade to Zephyr 4.1](/blog/2025/12/09/zephyr-4-1#zmk-board-variant),
 
 In addition to the basic keyboard functionality, there is also support for additional keyboard hardware such as encoders, RGB underglow, backlight and displays.
 Please see pages under the "Features" header in the sidebar for details.
-
-{/* prettier-ignore */}
-<Heading as="h2" id="contributing">Contributing</Heading>
-
-If you'd like to add support for a new keyboard shield, head over to the [New Keyboard Shield](hardware-integration/new-shield.mdx) documentation and note the [clean room design requirements](development/contributing/clean-room.md).
--- a/docs/docs/hardware.mdx
+++ b/docs/docs/hardware.mdx
@@ -35,19 +35,22 @@ export const toc = [
   },
 ];
 
-With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets,
+With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide variety of hardware targets,
 including but not limited to Nordic nRF52, Raspberry Pi RP2040/RP2350, most ST STM32 MCUs, and Microchip SAMD21.
-That being said, there are specific boards / shields that have been implemented and tested by the ZMK contributors, listed below.
+ZMK has the potential to run on any hardware supported by Zephyr™, such as those on
+the [Zephyr™ supported boards](https://docs.zephyrproject.org/4.1.0/boards/index.html) page, though you may need to do some additional work to configure them for ZMK.
+That being said, there are specific boards that have been tested and pre-configured by the ZMK contributors, denoted by the `zmk` board variants below.
 
 ZMK doesn't support boards with AVR 8-bit processors, such as the ATmega32U4, because Zephyr™ only
 supports 32-bit and 64-bit platforms. As a result, controllers like the SparkFun Pro Micro, Elite-C,
 and Arduino Uno Rev3 are **NOT** supported by ZMK.
 
-Check out the [Hardware Integration](development/hardware-integration/index.mdx) section for more info on how to use ZMK on custom keyboards.
+Making new keyboard designs? Check out the [Hardware Integration](development/hardware-integration/index.mdx) section
+for more information on how to configure ZMK to run on your custom hardware.
 
 :::info[Boards and Shields]
 
-ZMK uses the Zephyr concepts of "boards" and "shields" to refer to different parts of a keyboard build that in turn get combined during a firmware build.
+ZMK uses the Zephyr concepts of "boards" and "shields" to refer to different parts of a keyboard build that are then combined during a firmware build.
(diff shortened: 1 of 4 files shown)


docs/docs/hardware.mdx
```diff
@@ -29,24 +24,31 @@ export const toc = [
...
 ];
 
 With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets,
 including but not limited to Nordic nRF52, Raspberry Pi RP2040/RP2350, most ST STM32 MCUs, and Microchip SAMD21.
-That being said, there are specific [boards / shields](development/hardware-integration/index.mdx#boards--shields) that have been implemented and tested by the ZMK contributors, listed below.
+That being said, there are specific boards / shields that have been implemented and tested by the ZMK contributors, listed below.
+
+ZMK doesn't support boards with AVR 8-bit processors, such as the ATmega32U4, because Zephyr™ only
+supports 32-bit and 64-bit platforms. As a result, controllers like the SparkFun Pro Micro, Elite-C,
+and Arduino Uno Rev3 are **NOT** supported by ZMK.
```
nmunnich: I would move these to the bottom of the page under "other hardware". Nowadays 8 bit is much less common.

docs/docs/hardware.mdx
```diff
@@ -29,24 +24,31 @@ export const toc = [
...
+    value: "Other Hardware",
+    id: "other-hardware",
     level: 2,
   },
 ];
 
 With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets,
 including but not limited to Nordic nRF52, Raspberry Pi RP2040/RP2350, most ST STM32 MCUs, and Microchip SAMD21.
-That being said, there are specific [boards / shields](development/hardware-integration/index.mdx#boards--shields) that have been implemented and tested by the ZMK contributors, listed below.
+That being said, there are specific boards / shields that have been implemented and tested by the ZMK contributors, listed below.
```
nmunnich: I'd like to see a mention of ZMK variants of boards here.

docs/docs/hardware.mdx
```diff
@@ -29,24 +24,31 @@ export const toc = [
...
+
+ZMK doesn't support boards with AVR 8-bit processors, such as the ATmega32U4, because Zephyr™ only
+supports 32-bit and 64-bit platforms. As a result, controllers like the SparkFun Pro Micro, Elite-C,
+and Arduino Uno Rev3 are **NOT** supported by ZMK.
 
-:::note
+Check out the [Hardware Integration](development/hardware-integration/index.mdx) section for more info on how to use ZMK on custom keyboards.
```
nmunnich: Not a fan of the phrasing here, "how to use ZMK" sounds like a user guide not a guide for keyboard designers/DIYers.

docs/src/components/hardware-list.tsx
```diff
@@ -84,6 +84,27 @@ function HardwareList({ items }: HardwareListProps) {
...
     <>
+      <section>
+        <Heading as="h2" id="composite">
+          Composite Keyboards
+        </Heading>
+        <p>
+          Composite keyboards are composed of two main PCBs: a small controller
+          board with exposed pads, and a larger keyboard PCB (a shield, in ZMK
+          lingo) with switch footprints and a location where the controller is
+          added. This location is called an interconnect.
```
nmunnich: I would say "This location is called an interconnect" is not quite a correct statement. Interconnect is more used as a term for the combination of the physical locations of pins/hardware and the functionalities of the pins/hardware, but as a _standard_. I suppose it might be worth revisiting our terminology somewhat, perhaps we could come up with some useful terminology for the physical locations on the board and shield sides where the interconnect is applied.

docs/src/components/hardware-list.tsx
```diff
@@ -84,6 +84,27 @@ function HardwareList({ items }: HardwareListProps) {
...
+        </p>
+        <p>
+          Designing a custom keyboard with an off-the-shelf controller board?
+          Check out the{" "}
+          <a href="/docs/development/hardware-integration/new-shield">
+            New Keyboard Shield
+          </a>{" "}
+          guide.
```
nmunnich: This made me wonder whether we should continue our process of moving away from "keyboard" and more towards "ZMK device" since we're getting more and more mice running ZMK.

nmunnich (commented):
Looks like good overall improvements at a glance. While we're at it, I'd really like to see some link to Zephyr's supported hardware. I'd like it to be clear somehow that "We can in theory support all of these boards and SoCs, but you'd need to put a little bit more effort in to make a ZMK variant of the board by following the new board guide, then you could try upstreaming it to ZMK". 

I'm still very very much not a fan of the way the supported hardware is presented in the page as a massive list. I'd much much prefer an interface like the one the Zephyr boards page has.

caksoylar:
> Looks like good overall improvements at a glance. While we're at it, I'd really like to see some link to Zephyr's supported hardware. I'd like it to be clear somehow that "We can in theory support all of these boards and SoCs, but you'd need to put a little bit more effort in to make a ZMK variant of the board by following the new board guide, then you could try upstreaming it to ZMK".

I think an indication somewhere of this would be really helpful, we can add a short remark and point to [the section in New Board guide](https://zmk.dev/docs/development/hardware-integration/new-board#write-a-zmk-variant-of-zephyr-board)?

docs/docs/hardware.mdx
Genteure: I think this section being here would be better. The entire page from top to bottom is:

- Overview
  - Type of boards that are supported
  - Type of boards that are unsupported
  - Prerequisite knowledge (board & shield)
- The list of hardware
- Others

It provides an overview before going into details.

docs/docs/hardware.mdx
Genteure: Rephrased to make it clear it's for custom designed keyboards.

docs/src/components/hardware-list.tsx
Genteure: Do you have suggestions on how to improve this section? I get what you mean but I don't have ideas on how to rephrase this section to make it better right now. Otherwise I think it's fine to keep it as is for now.

Genteure:
Rebased on top of `main` and force pushed. Changes are added as new commits on top of the existing ones to make reviewing changes easier, I could squash them later if needed.

-----

One thing I'm not that keen on is the "other hardware" section. It's not _wrong_ to say keyboards (boards and shields) and "the things inside of a keyboard other than the matrix and MCU" are all "hardware", but I don't really like they're presented as if they're at the same level.

The usual term for this is "peripheral", as seen in Zephyr docs <https://docs.zephyrproject.org/latest/hardware/peripherals/index.html>.

ZMK already uses "peripheral" in the context of split keyboard, as in "peripheral role". Probably not worth changing right now.

-----

> I'm still very very much not a fan of the way the supported hardware is presented in the page as a massive list. I'd much much prefer an interface like the one the Zephyr boards page has.

I agree, but I'm not going to change that in this PR, it's a bit too much work in one go and would need quite a bit of effort.

-----

I realize there's no explanation of what "ZMK board variant" is and why is it needed. Any suggestions on how to explain this concept?

docs/docs/hardware.mdx
nmunnich: It feels a bit redundant to list both the boards that we _do_ support and some of the ones we don't. The two are mutually exclusive after all. I would still mention it, but move it to the bottom for the odd person who really wants to check if we support 8bit or not.

docs/src/components/hardware-list.tsx
nmunnich: Perhaps something like "a small controller
          board with exposed pads whose pin locations and capabilities define an _interconnect_ and a larger PCB [...] which _accepts_ the interconnect by providing a suitable location where the controller is added"
?

caksoylar:
> I realize there's no explanation of what "ZMK board variant" is and why is it needed. Any suggestions on how to explain this concept?

You can link to https://zmk.dev/docs/hardware-integration/new-board#write-a-zmk-variant-of-zephyr-board with a short explanation of something like "adds necessary settings to make the board usable out-of-the-box with ZMK".

docs/docs/hardware.mdx
Genteure: I don't think it's redundant. Boards not listed could be "still supported but not as popular", or not supported. This "not supported" section was pulled from interconnect metadata, to help clarify while "Pro Micro" the interconnect is popular and supported, the original "Pro Micro" is not.
Without explicitly saying these kind of boards are not supported, the pro micro interconnect description is not unreasonable to interpret as an endorsement for "SparkFun Pro Micro". But considering SparkFun Pro Micro is not in stock anymore and indeed 8 bit is less common, I'll move it to the end.

docs/docs/hardware.mdx
```diff
@@ -29,24 +24,32 @@ export const toc = [
...
 
-With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide diversity of hardware targets,
+With the solid technical foundation of Zephyr™ RTOS, ZMK can support a wide variety of hardware targets,
 including but not limited to Nordic nRF52, Raspberry Pi RP2040/RP2350, most ST STM32 MCUs, and Microchip SAMD21.
-That being said, there are specific [boards / shields](hardware-integration/index.mdx#boards--shields) that have been implemented and tested by the ZMK contributors, listed below.
+ZMK has the potential to run on any hardware supported by Zephyr™, such as those on
+the [Zephyr™ supported boards](https://docs.zephyrproject.org/4.1.0/boards/index.html) page, though you may need to do some additional work to configure them for ZMK.
+That being said, there are specific boards that have been tested and pre-configured by the ZMK contributors, denoted by the `zmk` board variants below.
+
+Making new keyboard designs? Check out the [Hardware Integration](hardware-integration/index.mdx) section
```
nmunnich:
```suggestion
Designing a new keyboard? Check out the [Hardware Integration](hardware-integration/index.mdx) section
```

nmunnich (commented):
The rest of it looks good to me now, just a minor rephrase that would make it read smoother imo.

I suppose we will stick with "keyboard" everywhere for now, and look to change that at some point in the future.

nmunnich (approved):

PR #3371, opened 2026-05-31
docs: reduce the use of admonitions

Genteure:
A very opinionated PR that reduces the number of admonitions.

Visually, [admonitions](https://docusaurus.io/docs/3.9.2/markdown-features/admonitions) elevate the content inside them to draw the reader's attention to it. They subconsciously signal to the reader that the content inside them is in some way more important than the content outside them. Frequent use of admonitions can lead to visual fatigue and dilute the impact of the admonitions that are actually important.

In my opinion, a non-exhaustive list of good uses of admonitions includes:

- Tips and tricks that have consequences if not followed. For example, "...Forgetting to activate a venv can be very annoying as it will install all sorts of things outside virtual environments over time, possibly leading to further errors..." <https://squidfunk.github.io/mkdocs-material/customization/?h=venv#environment-setup>
- Situations where the usual assumptions are wrong, or common pitfalls. For example, ".env files are not loaded inside configuration files." <https://docs.astro.build/en/guides/environment-variables/#vites-built-in-support>
- Highly recommended alternatives. For example, "While describe.each is provided for Jest compatibility, Vitest also has describe.for which simplifies argument types and aligns with test.for." <https://vitest.dev/api/describe.html#describe-each>

A non-exhaustive list of unnecessary uses of admonitions includes:

- "By the way" tips and tricks, additional information, or suggestions. Sometimes the intent is to make the content *less* important, but an admonition elevates it visually regardless of the color and type.
  It's better to have a standard paragraph, maybe with "Tip:" in bold at the beginning. If it's really that irrelevant, it probably doesn't need to be included at all.
- General information that fits perfectly within the flow of the surrounding content. If it's not something that requires special attention, it probably doesn't need to be in an admonition.
- References, links, and citations. Sometimes the intent is also to make the content less important. A normal paragraph with links is perfectly fine. Astro, for example, has a dedicated "[ReadMore](https://github.com/withastro/docs/blob/19549bcf53f8d66702c267d52c36d9be360032ae/src/components/ReadMore.astro)" component for this <https://docs.astro.build/en/guides/configuring-astro/#the-astro-config-file>.

That said, everything is contextual and subjective. Some of the admonitions I left unchanged, and some that I changed could be debated.

I also made some other formatting changes.

The following files I didn't touch:

- `docs/docs/development/new-behavior.mdx`: A bit long and has a lot of admonitions, so I just skipped it for now.
- `docs/docs/development/hardware-integration/dongle.mdx`: Content in the admonitions seems self-contradictory?
- `docs/docs/development/hardware-integration/encoders.md`: I don't really understand it, and it seems wrong.

### DIFF
--- a/docs/docs/config/index.md
+++ b/docs/docs/config/index.md
@@ -7,7 +7,9 @@ ZMK has many configuration settings that can be changed to change the behavior o
 
 This page describes the Kconfig and Devicetree file formats and how to change settings in them. See the other pages in the configuration section for a list of settings you can change.
 
+:::note
 All configuration is currently set at compile time. After changing any settings, you must build new firmware and flash it for the changes to apply.
+:::
 
 ## Config File Locations
 
--- a/docs/docs/config/index.md
+++ b/docs/docs/config/index.md
@@ -7,9 +7,7 @@ ZMK has many configuration settings that can be changed to change the behavior o
 
 This page describes the Kconfig and Devicetree file formats and how to change settings in them. See the other pages in the configuration section for a list of settings you can change.
 
-:::note
 All configuration is currently set at compile time. After changing any settings, you must build new firmware and flash it for the changes to apply.
-:::
 
 ## Config File Locations
 
--- a/docs/docs/development/contributing/documentation.md
+++ b/docs/docs/development/contributing/documentation.md
@@ -11,6 +11,8 @@ This document outlines how to test your documentation changes locally and prepar
 
 The documentation is written with [Docusaurus](https://docusaurus.io/). The ZMK source code has all of the necessary Docusaurus dependencies included, but referencing their documentation can be helpful at times.
 
+Latest LTS version of node is used to build the website, available at <https://nodejs.org/en/download/>.
+
 The general process for updating the ZMK documentation is:
 
 1. Update the documentation
--- a/docs/docs/development/contributing/documentation.md
+++ b/docs/docs/development/contributing/documentation.md
@@ -18,14 +18,6 @@ The general process for updating the ZMK documentation is:
 3. Ensure the sources are formatted properly and linted
 4. Create a Pull Request for review and inclusion into the ZMK sources
 
-:::note
-If you are working with the documentation from within VS Code+Docker please be aware the documentation will not be auto-generated when making changes while the server is running. You'll need to restart the server when saving changes to the documentation.
-:::
-
-:::note
-You will need `Node.js` and `npm` installed to update the documentation. If you're using the ZMK dev container (Docker) the necessary dependencies are already installed. Otherwise, you must install these dependencies yourself. Since `Node.js` packages in Linux distributions tend to be outdated, it's recommended to install the current version from a repository like [NodeSource](https://github.com/nodesource/distributions) to avoid build errors.
-:::
-
 ## Testing Documentation Updates Locally
 
 To verify documentation updates locally, follow the following procedure. The `npm` commands and first step will need to be run from a terminal.
@@ -52,15 +44,9 @@ The check commands can be run with the following procedure in a terminal that's 
 3. Run `npm run lint`
 4. Run `npm run build`
 
-:::danger
 If any of the above steps throw an error, they need to be addressed and all of the checks re-run prior to submitting a pull request.
-:::
 
-:::note
-The documentation uses American English spelling and grammar conventions. Title case is used for the first three heading levels, with sentence case used beyond that.
-
-Please make sure your changes conform to these conventions - prettier and lint are unfortunately unable to do this automatically.
-:::
+The documentation uses American English spelling and grammar conventions. Title case is used for the first three heading levels, with sentence case used beyond that. Please make sure your changes conform to these conventions.
 
 ## Submitting a Pull Request
 
--- a/docs/docs/development/contributing/documentation.md
+++ b/docs/docs/development/contributing/documentation.md
@@ -11,7 +11,7 @@ This document outlines how to test your documentation changes locally and prepar
 
 The documentation is written with [Docusaurus](https://docusaurus.io/). The ZMK source code has all of the necessary Docusaurus dependencies included, but referencing their documentation can be helpful at times.
 
-Latest LTS version of node is used to build the website, available at <https://nodejs.org/en/download/>.
+The website is built using the latest LTS version of node, which is available at <https://nodejs.org/en/download/>.
 
 The general process for updating the ZMK documentation is:
 
--- a/docs/docs/development/module-creation.md
+++ b/docs/docs/development/module-creation.md
@@ -12,9 +12,7 @@ sidebar_label: ZMK Module Creation
 
 See also Zephyr's [page on modules](https://docs.zephyrproject.org/4.1.0/develop/modules.html).
 
-:::tip
-For open source hardware designs, it can be convenient to use [Git submodules](https://github.blog/open-source/git/working-with-submodules/) to have the ZMK module also be a Git submodule of the repository hosting the hardware design.
-:::
+For open source hardware designs, it's recommended to **not** include the hardware design files in the ZMK module itself, since the module will be fetched by users during build and having design files in the module would likely make the module unnecessarily large and slow to fetch.
 
 ## Module Setup
 
@@ -133,15 +131,15 @@ Note that the `include` and `src` folders are not mandated by the module system,
 Modules should expose all provided header files with an include path name beginning with the module-name, for example at `include/zmk_<type>_<description>/<header>.h`.
 
 :::info
-If your module requires adding drivers to existing subsystems in modules, you will need to use the `zephyr_library_amend()` CMake command, which requires you to have a specific directory structure. See [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L454) for the definition and some documentation in the comments, with an example [here](https://github.com/petejohanson/ec-support-zmk-module/tree/main/drivers/kscan).
+If your module requires adding drivers to existing subsystems in modules, you will need to use the `zephyr_library_amend()` CMake command, which requires you to have a specific directory structure. See [`zephyr/cmake/modules/extensions.cmake`](https://github.com/zephyrproject-rtos/zephyr/blob/main@%7B2025-Feb-15%7D/cmake/modules/extensions.cmake#L454) for the definition and some documentation in the comments, with an example in [`petejohanson/ec-support-zmk-module`](https://github.com/petejohanson/ec-support-zmk-module/tree/main/drivers/kscan).
 :::
 
 ## Examples
 
 Below are some examples of modules for different types. Unless under the `zmkfirmware` project, these are not endorsed officially and may not follow our conventions perfectly. For such reason, the modules chosen to be presented here may change with time.
 
-- Keyboard: https://github.com/petejohanson/zmk-keyboards-katori
-- Behavior: https://github.com/urob/zmk-leader-key
-- Driver: https://github.com/petejohanson/cirque-input-module
-- Feature: https://github.com/joelspadin/zmk-locales
-- VFX: https://github.com/caksoylar/zmk-rgbled-widget
+- Keyboard: <https://github.com/petejohanson/zmk-keyboards-katori>
+- Behavior: <https://github.com/urob/zmk-leader-key>
+- Driver: <https://github.com/petejohanson/cirque-input-module>
+- Feature: <https://github.com/joelspadin/zmk-locales>
+- VFX: <https://github.com/caksoylar/zmk-rgbled-widget>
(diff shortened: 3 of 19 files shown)


docs/docs/development/contributing/documentation.md
```diff
@@ -18,14 +18,6 @@ The general process for updating the ZMK documentation is:
...
 4. Create a Pull Request for review and inclusion into the ZMK sources
 
-:::note
-If you are working with the documentation from within VS Code+Docker please be aware the documentation will not be auto-generated when making changes while the server is running. You'll need to restart the server when saving changes to the documentation.
-:::
-
-:::note
-You will need `Node.js` and `npm` installed to update the documentation. If you're using the ZMK dev container (Docker) the necessary dependencies are already installed. Otherwise, you must install these dependencies yourself. Since `Node.js` packages in Linux distributions tend to be outdated, it's recommended to install the current version from a repository like [NodeSource](https://github.com/nodesource/distributions) to avoid build errors.
-:::
-
```
Genteure, line 28: The first note is not applicable, or at least it doesn't happen to me. Using VS Code with `npm run start` in devcontainers, editing `.md` files either inside or outside of the container triggers hot reload. It's either been solved since then, or a problem in a more specific environment.

The second note, I think is too specific to be helpful. Surely anyone working with node projects would know to maintain their node environments. Also it's specifically targeting Linux users. The `.nvmrc` file indicates this node project wants the lts version.

For these reasons I removed both.

docs/docs/development/module-creation.md
```diff
@@ -12,9 +12,7 @@ sidebar_label: ZMK Module Creation
 
 See also Zephyr's [page on modules](https://docs.zephyrproject.org/4.1.0/develop/modules.html).
 
-:::tip
-For open source hardware designs, it can be convenient to use [Git submodules](https://github.blog/open-source/git/working-with-submodules/) to have the ZMK module also be a Git submodule of the repository hosting the hardware design.
-:::
+For open source hardware designs, it's recommended to **not** include the hardware design files in the ZMK module itself, since the module will be fetched by users during build and having design files in the module would likely make the module unnecessarily large and slow to fetch.
```
Genteure, line 15: Reworded. I personally don't really agree with using submodules (I think they're too clumsy to use for most purposes). Also there was no reason given on why it's convenient.
IIRC in discord it was explained that design files easily bloat the repository so reworded to say that instead.

docs/docs/hardware-integration/physical-layouts.md
```diff
@@ -62,11 +62,7 @@ A key description has the shape `<&key_physical_attrs w h x y r rx ry>` with the
 
 You can specify negative values in devicetree using parentheses around it, e.g. `(-3000)` for a 30 degree counterclockwise rotation.
 
-:::tip
-
-We recommend the use of [this tool](https://zmk-physical-layout-converter.streamlit.app/) for writing a physical layout or converting one from a QMK JSON definition. If your keyboard already has a physical layout defined for the use with KLE, we recommend using [this other tool](https://nickcoutsos.github.io/keymap-layout-tools/) first to convert your existing layout into QMK JSON. The second tool can also import the position data from KiCAD, if said program was used to design the keyboard.
-
-:::
+We recommend the use of <https://zmk-physical-layout-converter.streamlit.app> for writing a physical layout or converting one from a QMK JSON definition. If your keyboard already has a physical layout defined for the use with KLE, we recommend using <https://nickcoutsos.github.io/keymap-layout-tools/> first to convert your existing layout into QMK JSON. The second tool can also import the position data from KiCAD, if said program was used to design the keyboard.
```
Genteure, line 65: There are a few other `[here](https://example)` type of links in other files unchanged.

docs/docs/keymaps/behaviors/macros.md
```diff
@@ -312,10 +312,6 @@ To avoid repetition or possible typos when declaring a **zero parameter macro**,
     )
 ```
 
-:::note
-`ZMK_MACRO()` **only supports declaring non-parameterized (zero parameter) macros**; parameterized declarations are not currently supported.
-:::
-
```
Genteure, line 318: Removed because it's basically saying the same thing again.

caksoylar:
I mostly agree with the intentions in the PR body. One thing that IIRC someone (maybe Joel?) also brought up on Discord is that our admonitions (even "note" ones) are way too prominent. IMO if that could be fixed, we could probably use the "note" admonitions for asides like below:

> "By the way" tips and tricks, additional information, or suggestions. Sometimes the intent is to make the content less important, but an admonition elevates it visually regardless of the color and type.
> It's better to have a standard paragraph, maybe with "Tip:" in bold at the beginning. If it's really that irrelevant, it probably doesn't need to be included at all.

I am ambivalent on doing it either way, but maybe you can experiment with it.

@Genteure can you also rebase on main to fix the merge conflicts?

docs/docs/keymaps/list-of-keycodes.mdx
```diff
@@ -9,11 +9,6 @@ import Table from "@site/src/components/codes/Table";
 
 This is the reference page for keycodes used by behaviors. Use the table of contents (on the right or the top) for easy navigation.
 
-:::warning
-Take extra notice of the spelling of the keycodes, especially the shorthand spelling.
-Otherwise, it will result in an elusive parsing error!
-:::
```
caksoylar, line 15: While I find this pretty unnecessary myself, there was a big push by some users to add this (probably discoverable through git blame). I am fine to remove unless someone else has an objection.

docs/docs/development/module-creation.md
```diff
@@ -133,15 +131,15 @@ Note that the `include` and `src` folders are not mandated by the module system,
 Modules should expose all provided header files with an include path name beginning with the module-name, for example at `include/zmk_<type>_<description>/<header>.h`.
 
 :::info
-If your module requires adding drivers to existing subsystems in modules, you will need to use the `zephyr_library_amend()` CMake command, which requires you to have a specific directory structure. See [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L454) for the definition and some documentation in the comments, with an example [here](https://github.com/petejohanson/ec-support-zmk-module/tree/main/drivers/kscan).
+If your module requires adding drivers to existing subsystems in modules, you will need to use the `zephyr_library_amend()` CMake command, which requires you to have a specific directory structure. See [`zephyr/cmake/modules/extensions.cmake`](https://github.com/zephyrproject-rtos/zephyr/blob/main@%7B2025-Feb-15%7D/cmake/modules/extensions.cmake#L454) for the definition and some documentation in the comments, with an example in [`petejohanson/ec-support-zmk-module`](https://github.com/petejohanson/ec-support-zmk-module/tree/main/drivers/kscan).
```
caksoylar:
```suggestion
If your module requires adding drivers to existing subsystems in modules, you will need to use the `zephyr_library_amend()` CMake command, which requires you to have a specific directory structure. See [`zephyr/cmake/modules/extensions.cmake`](https://github.com/zephyrproject-rtos/zephyr/blob/main@%7B2025-Feb-15%7D/cmake/modules/extensions.cmake#L492) for the definition and some documentation in the comments, with an example in [`petejohanson/ec-support-zmk-module`](https://github.com/petejohanson/ec-support-zmk-module/tree/main/drivers/kscan).
```
Thanks for the catch. This seems the more correct spot, though not sure now where the original link went to?

docs/docs/hardware-integration/physical-layouts.md
caksoylar: I'd say they aren't very elegant, but not a priority to fix.

docs/docs/development/contributing/documentation.md
caksoylar: TBH ZMK was my first real encounter with node, and ran into the issue with Ubuntu LTS packaged version being too old before noticing this. (Now docusaurus 3 is old enough that even distro versions have caught up?)

It might be worth keeping a notice saying ZMK documentation requires [at least node 20.0](https://docusaurus.io/docs/installation#requirements) and to make sure to install a compatible version.

docs/docs/development/module-creation.md
Genteure: Looking at the original PR, this link was added on 2025-Feb-15, right before the final approve and merge https://github.com/zmkfirmware/zmk/pull/2456#event-16319041987. With `main@{2025-Feb-15}`, the new link should point to the same commit it pointed to when the link was originally added.
But yes looks like the intent might be linking to line 492. @nmunnich Can you please help confirm?

docs/docs/keymaps/list-of-keycodes.mdx
Genteure: It was originally added in https://github.com/zmkfirmware/zmk/pull/1415

docs/docs/development/contributing/documentation.md
Genteure: Added a short note about using latest LTS, matching the version specified in `docs/.nvmrc`.

docs/docs/config/index.md
```diff
@@ -7,9 +7,7 @@ ZMK has many configuration settings that can be changed to change the behavior o
 
 This page describes the Kconfig and Devicetree file formats and how to change settings in them. See the other pages in the configuration section for a list of settings you can change.
 
-:::note
 All configuration is currently set at compile time. After changing any settings, you must build new firmware and flash it for the changes to apply.
-:::
```
nmunnich, line 12: I would keep this, personally.

docs/docs/development/contributing/documentation.md
```diff
@@ -11,21 +11,15 @@ This document outlines how to test your documentation changes locally and prepar
 
 The documentation is written with [Docusaurus](https://docusaurus.io/). The ZMK source code has all of the necessary Docusaurus dependencies included, but referencing their documentation can be helpful at times.
 
+Latest LTS version of node is used to build the website, available at <https://nodejs.org/en/download/>.
```
nmunnich:
```suggestion
The website is built using the latest LTS version of node, which is available at <https://nodejs.org/en/download/>.
```

docs/docs/development/module-creation.md
nmunnich: I suggested the use of submodules here because I recall some conversations where people wanted to keep their design files and their firmware files in the same repo/folder for convenience reasons. I generally don't like submodules that much either, but for that purpose I'm not sure a better alternative really exists.

docs/docs/features/studio.md
```diff
@@ -6,7 +6,7 @@ ZMK Studio provides runtime update functionality to ZMK powered devices, allowin
 
 :::info
 
-To use ZMK Studio, a keyboard needs to be [configured appropriately](#adding-zmk-studio-support-to-a-keyboard). ZMK has updated some, but not all, of its in-tree keyboards for use with ZMK Studio, the list of which can be found [here](/blog/2024/11/11/zmk-studio-mvp-ga). If your keyboard is supported by an external module/config, check with the maintainer to see if support has been added.
+To use ZMK Studio, a keyboard needs to be [configured appropriately](#adding-zmk-studio-support-to-a-keyboard). ZMK has updated some, but not all, of its in-tree keyboards for use with ZMK Studio, the list of which can be found in the [ZMK Studio blog post](/blog/2024/11/11/zmk-studio-mvp-ga). If your keyboard is supported by an external module/config, check with the maintainer to see if support has been added.
```
nmunnich, line 9: Tbh we probably should change this, I think a number of the keyboards not listed have since been updated... Maybe something with the supported hardware page and metadata...

docs/docs/hardware-integration/lighting/underglow.md
```diff
@@ -17,11 +17,7 @@ For example: the `kyria` shield has a [`boards/nice_nano_nrf52840_zmk.overlay`](
 
 Using an SPI-based LED strip driver on the `&spi3` interface is the simplest option for nRF52-based boards. If possible, avoid using pins which are limited to low-frequency I/O for this purpose. The resulting interference may result in poor wireless performance.
 
-:::info
-
-The list of low frequency I/O pins for the nRF52840 can be found [here](https://docs.nordicsemi.com/bundle/ps_nrf52840/page/pin.html).
-
-:::
+The list of low frequency I/O pins for the nRF52840 can be found at <https://docs.nordicsemi.com/bundle/ps_nrf52840/page/pin.html>.
```
nmunnich: I feel like an admonition is still valid here because using a low frequency pin for high frequency purposes can be quite bleh. Perhaps we move " If possible, avoid using pins which are limited to low-frequency I/O for this purpose. The resulting interference may result in poor wireless performance." in?

docs/docs/development/module-creation.md
nmunnich: I can't recall for certain, but I think I may have typoed 464 with the intent being to put the user at the top of the comment block. 492 works well.

docs/docs/features/studio.md
caksoylar: ~~I could have sworn we added a `studio` property to the metadata definition for this, though that might have been just talked about.~~

It is indeed there, but I guess we never hooked it up to the hardware page: https://github.com/zmkfirmware/zmk/commit/64c1e1e3ff0ad52226cc32d30a7a2a6b70924662

docs/docs/hardware-integration/lighting/underglow.md
Genteure: Moved the low-frequency note along with the link back in an admonition.

caksoylar (approved):

nmunnich (approved):

Issue #805, opened 2021-05-24
Bluetooth security failing on windows machine


aureumleonis:
I've been using a split/nice!nano handwired built for a few months already with no issues on my personal PC. On the other hand, The keyboard pairs correctly with the my work laptop but doesn't send any keypresses. Both of these are windows 10 machines. I finally took the initiative to enable debugging and saw this happen constantly (a couple of times per second)

```
[00:25:58.514,068] <err> zmk: Security failed: <redacted> (public) level 1 err 2
[00:25:58.639,770] <dbg> zmk.split_central_disconnected: Disconnected: <redacted> (public) (reason 19)
[00:25:58.639,831] <dbg> zmk.disconnected: Disconnected from <redacted> (public) (reason 0x13)
[00:25:58.639,862] <dbg> zmk.update_advertising: advertising from 2 to 2
[00:25:58.678,680] <dbg> zmk.split_central_connected: Connected: <redacted> (public)
[00:25:58.678,680] <dbg> zmk.split_central_process_connection: Current security for connection: 1
[00:25:58.679,901] <dbg> zmk.split_central_process_connection: New connection params: Interval: 48, Latency: 0, PHY: 1
[00:25:58.679,992] <dbg> zmk.connected: Connected thread: 0x20002e40
[00:25:58.679,992] <dbg> zmk.connected: Connected <redacted> (public)
[00:25:58.680,023] <dbg> zmk.update_advertising: advertising from 0 to 2
```

I may have the beginning/end of the log wrong but this is still the part that repeats. Does anybody know what the problem could be? Is there anything else I can I do to find more data about the problem?

haywirephoenix:
Have you tried switching the bluetooth profile for your work machine? 
It's right here in the docs https://zmkfirmware.dev/docs/behaviors/bluetooth/#bluetooth-pairing-and-profiles

aureumleonis:
I've tried these two scenarios

Personal + work PCs
* Pair personal PCs on profile 0 (keyboard pairs and connects)
* Pair work PC on profile 1 (keyboard only pairs, shows the logs in the description)

Only work PC
* Clear all profiles
* Pair work PC on profile 0 (keyboard only pairs, shows the logs in the description)

haywirephoenix:
Maybe related: I've noticed an issue when using BT_CLR you have to remove+repair the paired device from the list and restart the pc before it will register input. 

It may be worth doing a full settings clear and trying again without touching BT_CLR
[...]

PR #3377, opened 2026-06-04
feat(split): optionally clear stale split bond on PIN/key mismatch

akira-toriyama:
## What

Add `CONFIG_ZMK_SPLIT_AUTO_UNPAIR_ON_KEY_MISMATCH` (off by default,
depends on `ZMK_SPLIT_ROLE_CENTRAL`). When enabled, a split central
that receives `BT_SECURITY_ERR_PIN_OR_KEY_MISSING` from a peripheral
calls `bt_unpair()` + `bt_conn_disconnect()` so the next advertising
window re-pairs from scratch.

## Why

In a split setup (one central + N peripherals), the central's bond
record for one peripheral can become invalid while the peripheral
still holds the old LTK. Once that happens the next reconnection
sits in an infinite `PIN_OR_KEY_MISSING` loop and cannot recover
without manual intervention — which on a dongle-style central is
particularly painful because the dongle has no keymap from which to
issue any recovery behavior.

Concrete situations that produce this state:

- A `settings_reset` UF2 is flashed on the central (or, less
  commonly, on the peripheral) while the other side keeps its bond.
- The central hardware is swapped (e.g. a different XIAO BLE) while
  the peripherals are kept as-is.
- USB-level reset of the dongle — observed on hardware as one of
  the two peripherals occasionally landing in this state on KVM
  switch-over and recovering automatically once the patch is active
  (I can't fully attribute the KVM trigger to NVS state yet, only
  that the symptom is the same `PIN_OR_KEY_MISSING` loop and the
  same `bt_unpair()` clears it).

In all three the symptom is the same: the central's LTK is gone or
shifted for that peripheral, the peripheral still replays the old
LTK, and the link cannot complete without user action.

## Scope and gating

- Strictly the **split central** path: gated on
  `IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL)`. The new behavior
  cannot fire on the central ↔ host pairing path.
- Strictly **opt-in**: `ZMK_SPLIT_AUTO_UNPAIR_ON_KEY_MISMATCH` is
  off by default. Clearing a bond is destructive; only deployments
  that are known to benefit (typically dongle setups) should enable
  it.
- Strictly the exact error code
  `BT_SECURITY_ERR_PIN_OR_KEY_MISSING`. Other security failures are
  left alone.

## Testing

Verified on hardware via akira-toriyama/canon (Cyboard Imprint:
XIAO BLE dongle as central, both halves as peripherals):

- Builds clean on `xiao_ble/nrf52840/zmk + imprint_dongle`.
- With the flag off, existing behavior is byte-identical.
- With the flag enabled, dongle + peripherals operate as before
  under normal use, and a previously-reproducible failure where
  one peripheral half stopped responding after a KVM USB switch-over
  no longer reproduces in the user's setup. A few switch cycles
  tested; not claimed as exhaustive given KVM-USB enumeration
  variability.

## Related

- #805 — same `Security failed: ... err 2` =
  `BT_SECURITY_ERR_PIN_OR_KEY_MISSING` pattern in a different
  (host-side) scenario. This PR doesn't claim to close it; the
  split flag is unrelated to the host pair path by design.

## Notes for maintainers

- Happy to rename (`ZMK_SPLIT_DROP_STALE_BOND`, etc.) if there's a
  preferred convention.
- I considered also covering the symmetric peripheral-side case
  (peripheral keeps its bond, central LTK shifted) but left it out
  of this PR to keep the scope tight. Happy to follow up if you'd
  like it included here.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -223,6 +223,27 @@ config ZMK_BLE_MOUSE_REPORT_QUEUE_SIZE
 config ZMK_BLE_CLEAR_BONDS_ON_START
     bool "Configuration that clears all bond information from the keyboard on startup."
 
+config ZMK_SPLIT_AUTO_UNPAIR_ON_KEY_MISMATCH
+    bool "Automatically clear a stale split bond when the peripheral reports a PIN/key mismatch"
+    depends on ZMK_SPLIT_ROLE_CENTRAL
+    help
+      When enabled on a split central, ZMK clears the bond and disconnects
+      the peripheral if the security exchange with that peripheral fails
+      with BT_SECURITY_ERR_PIN_OR_KEY_MISSING. The next advertising window
+      then re-pairs from scratch without manual intervention.
+
+      This handles the asymmetric stale-bond case in a split setup: the
+      central no longer has a usable LTK for that peripheral (e.g. the
+      central was flashed with a settings_reset image, or the central
+      hardware was swapped while the peripherals were kept), but the
+      peripheral keeps replaying its old LTK, so the link cannot complete
+      without user action. It is particularly useful for dongle-style
+      centrals, which have no keymap from which to trigger recovery
+      behaviors.
+
+      Off by default because clearing a bond is destructive and cannot be
+      undone; only enable on deployments that are known to benefit.
+
 # HID GATT notifications sent this way are *not* picked up by Linux, and possibly others.
 config BT_GATT_NOTIFY_MULTIPLE
     default n
--- a/app/src/ble.c
+++ b/app/src/ble.c
@@ -562,6 +562,14 @@ static void security_changed(struct bt_conn *conn, bt_security_t level, enum bt_
         LOG_DBG("Security changed: %s level %u", addr, level);
     } else {
         LOG_ERR("Security failed: %s level %u err %d", addr, level, err);
+#if IS_ENABLED(CONFIG_ZMK_SPLIT_ROLE_CENTRAL) && \
+    IS_ENABLED(CONFIG_ZMK_SPLIT_AUTO_UNPAIR_ON_KEY_MISMATCH)
+        if (err == BT_SECURITY_ERR_PIN_OR_KEY_MISSING) {
+            LOG_WRN("Stale split bond detected, clearing and disconnecting");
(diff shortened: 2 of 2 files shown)


caksoylar:
I am confused about the scope of this change. Is this for

- ZMK centrals pairing (as peripherals) to hosts/computers? If so, the dongle comments don't make sense.
- ZMK peripherals pairing to ZMK centrals, for splits? If so, comments like "link
never recovers without manual &bt BT_CLR on the peripheral" or "for example when the
central is re-flashed but the peripherals are not" don't make sense, since (a) you cannot clear split bonds with the `&bt` behavior, (b) flashing firmware on central won't clear bonds, unless you are flashing a special one like `settings_reset`

akira-toriyama:
Thanks for the careful read — your second bullet is the intent, and you've correctly flagged two factual errors in my body. I've just force-pushed an update; here's the summary.

**Scope:** split central ↔ peripheral pairing. To make that unambiguous in the code I renamed the Kconfig to `CONFIG_ZMK_SPLIT_AUTO_UNPAIR_ON_KEY_MISMATCH` and gated it on `CONFIG_ZMK_SPLIT_ROLE_CENTRAL`, so the new behavior cannot fire on the central ↔ host pairing path.

**On the inaccuracies you spotted:**

- `&bt BT_CLR` doesn't touch split bonds — agreed, dropped.
- A normal `west flash` / UF2 overwrite doesn't wipe NVS, so "re-flashing the central wipes its bond" wasn't a realistic trigger — agreed. The real situations that produced a one-sided stale bond in our dongle setup were:
  - flashing a `settings_reset` UF2 on one side (intentionally or otherwise),
  - swapping the central hardware while keeping the existing peripherals,
  - and, observed on hardware: a KVM USB switch reset of the dongle that occasionally left one peripheral's split link wedged on `Security failed ... err 2` until the patch kicked in (I can't fully attribute the KVM trigger to NVS state yet, only that the symptom matches and `bt_unpair()` cleared it).

In all three, the central's view of one peripheral's bond is gone or shifted, the peripheral keeps replaying the old LTK, and the link sits in `PIN_OR_KEY_MISSING` until somebody intervenes.

Updated body and a cleaner testing section are pushed. Marking ready-for-review.

PR #3402, opened 2026-06-23
feat(hid): Dynamic nkro

### LINKED ISSUE #3395: Feature: Dynamic NKRO
### Motivation
A core frustration of mine with ZMK is how the decision of NKRO vs HKRO is a build time fixed value. I want to use NKRO day to day for all of the optimized experience that comes with and extended key codes. On the other hand, I sometimes need to work on machines where bios is incompatible with NKRO.

The existing solution here is one of
- Use a second keyboard
- Re-flash the keyboard with the setting toggled

The claim in this feature request is neither of these solutions are optimal, and there is a technique that can be introduced to zmk to support a user controlled toggle in favor of re-flashing. Further, this technique can be done in a way that is fully backward compatible with existing configurations and results in no change to existing users.

### Technique
- During initialization, zmk has a [`static const` hid report](https://github.com/zmkfirmware/zmk/blob/main/app/include/zmk/hid.h#L104-L256)
- We can introduce a new KConfig entry, for example `ZMK_HID_REPORT_TYPE_DYNAMIC`
- The new KConfig entry can introduce simple logic of persistent flash storage to maintain the users preferred reporting type
- The flash memory then can be incorporated into the [initialization routine](https://github.com/zmkfirmware/zmk/blob/main/app/src/usb_hid.c#L238-L254) to resolve the users preferred reporting type
- A set of behaviors can be introduced to change that persistent setting and reboot the keyboard

### Event Publishing
- For the feature here, we can create a ZMK event that users could configure to show status on a screen or LED for the current configuration
- We could include this event for the compile time options as well but I'm not sure if that is desirable since it would be a fixed event on every boot of the keyboard, but it could provide a clean way to get the status for those users, but if they cared I would assume they would opt into this feature anyway, so I don't think it's necessary to implement.

### Bluetooth
The fundamental issue I see in practice is when the keyboard is connected using usb, since that's what a bios not supporting NKRO uses. An ideal solution would also incorporate a solution for bluetooth connectivity, although this is a harder problem since host side will cache the descriptor associated with the connection, requiring a repairing or potentially using two bluetooth profiles for this use case.

### My Attempt
I'm currently trying to [implement this](https://github.com/t3dodson/zmk/tree/dynamic-nkro)
My hope would be this change could be a candidate to be up-streamed into zmk as I think many users would find this useful.

t3dodson:
Closes #3395 curious to get feedback, as I had mentioned in the issue I think this is a widely useful feature suitable for upstream zmk.

With this fix I can finally switch into my bios with the same keyboard and make my edits.

## PR check-list

---
I manually validated the following scenarios
# Dynamic NKRO — Test Scenarios

Critical test matrix for the dynamic HKRO/NKRO feature. Columns are the relevant
Kconfig settings; rows are the build/runtime scenarios we care about. The first
two are regression baselines (feature absent → behavior must be unchanged); the
rest exercise the dynamic path.

`HKRO` / `NKRO` = `CONFIG_ZMK_HID_REPORT_TYPE_HKRO` / `_NKRO` (the report-type
choice). `DYNAMIC` = `CONFIG_ZMK_HID_REPORT_TYPE_DYNAMIC`. `EXT` =
`CONFIG_ZMK_HID_KEYBOARD_NKRO_EXTENDED_REPORT`. `BOOT` = `CONFIG_ZMK_USB_BOOT`.

| #   | Scenario                | Type choice | DYNAMIC | EXT | BOOT | First-boot mode | What to verify                                                                                  | Result                                                                                                |
| --- | ----------------------- | ----------- | ------- | --- | ---- | --------------- | ----------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- |
| 1   | NKRO baseline (regress) | NKRO        | n       | n   | n    | NKRO (fixed)    | Full NKRO works as before; descriptor/report size unchanged vs. upstream; `&dyn_nkro` unavailable | ✅ PASS (2026-06-22) — NKRO confirmed on test site; emptied `&dyn_nkro` keys inert as expected. Build dropped 1536 B. |
| 2   | HKRO baseline (regress) | HKRO        | n       | n   | n    | HKRO (fixed)    | 6KRO works as before; 7th simultaneous key rolls over; behavior identical to upstream            | ✅ PASS (2026-06-22) — 6 keys register, 7th rolls off as expected. left FLASH 366228 B.               |
| 3   | Dynamic, NKRO default   | NKRO        | y       | n   | n    | NKRO            | Boots NKRO; `&dyn_nkro DYN_NKRO_TOG` → reboot → HKRO; toggle back → NKRO; mode survives power cycle | ✅ PASS (2026-06-22) — boots NKRO; TOG→HKRO, persists across power cycle; TOG→NKRO; force NKRO/HKRO cmds also verified. left FLASH 367040 B (both descriptors). |
| 4   | Dynamic, HKRO default   | HKRO        | y       | n   | n    | HKRO            | Boots HKRO; `DYN_NKRO_NKRO` → reboot → NKRO; `DYN_NKRO_HKRO` when already HKRO is a no-op (no reboot) | ✅ PASS (2026-06-22) — boots HKRO; force-HKRO no-op while HKRO; NKRO works after toggle; force-NKRO no-op while NKRO. left FLASH 367040 B (= TC3). |
| 5   | Dynamic + boot protocol | NKRO        | y       | n   | y    | NKRO            | Boot keyboard report correct in both modes (e.g. BIOS/UEFI screen) after switching each way       | ✅ PASS (2026-06-22) — OS works in both modes; in-BIOS HKRO works & toggling works in the BIOS menu. NKRO not accepted by this BIOS (host reads report-protocol descriptor, doesn't request USB boot protocol) — expected, and the motivation for the feature. No stuck keys / misbehavior. left FLASH 367472 B. |

## Bluetooth
I also validated behavior using bluetooth on an iphone
I connect, if I toggle to a different mode, I have to reconnect bluetooth then it does behave as expected. I understand some devices may not handle this gracefully, but I would hope using multiple profiles would be sufficient for that use case.


nmunnich:
I do wonder what the use case for HKRO on BLE would be. For a feature along these lines, I think I would be instinctively more inclined to allow swapping between NKRO and HKRO on USB only. I do not know of a use case where HKRO on BLE is desired, but if it was for some point, then I think HKRO vs NKRO BLE should be treated on a per-profile basis.

t3dodson:
Yeah bluetooth is not my use case either and without a known bluetooth device that only works with HKRO it's hard to justify.

Any preference of disabling this stuff for bluetooth and cleaning up configs and docs to make it more clear it only works over usb?

The alternative is research to lock this behavior per profile which i'm not fully sure how to go about.

nmunnich:
Yeah, I think I would take that direction. I think I would also take the direction of reusing the &out behaviour to switch between the usb modes, rather than having a separate dedicated behaviour do so.

I'd be curious about some compile size comparisons when the toggle is enabled Vs disabled. I think if it's extremely similar then we might have an argument to enable toggle always for usb and have the flag switch BLE only.

PR #3409, opened 2026-07-05
fix(drivers): Fix missing pulses in EC11

renshyle:
The EC11 driver is flawed since it reads both pin states when any edge is detected. Unfortunate timing and a messy signal for a pin coming from the EC11 can thus cause a pulse to be lost.

Rework the driver to have only one pin (called the active pin) with interrupts enabled at a time. When an edge is detected on this pin, it is assumed to have flipped and the state of the inactive pin is read. These, along with the previous state of the inactive pin are used to determine the direction and magnitude of the pulse delta.

This implementation requires EC11 triggers so `CONFIG_EC11_TRIGGER_NONE` and `CONFIG_EC11_TRIGGER` are also removed. Based on my search on GitHub, I couldn't find a single user of EC11 with `CONFIG_EC11_TRIGGER_NONE=y`.

## Testing

I tested this on a Keychron Q3 Max with an EC11 that has 4 pulses between detents. It went from having occasional issues with pulses being lost to them never being lost. I don't have any other keyboards or EC11s but it'd be useful to test this with an EC11 that has 2 pulses between detents.

## PR check-list

### DIFF
--- a/app/module/drivers/sensor/ec11/Kconfig
+++ b/app/module/drivers/sensor/ec11/Kconfig
@@ -13,27 +13,18 @@ if EC11
 
 choice EC11_TRIGGER_MODE
     prompt "Trigger mode"
-    default EC11_TRIGGER_NONE
     help
       Specify the type of triggering to be used by the driver.
-
-config EC11_TRIGGER_NONE
-    bool "No trigger"
 
 config EC11_TRIGGER_GLOBAL_THREAD
     bool "Use global thread"
     depends on GPIO
-    select EC11_TRIGGER
 
 config EC11_TRIGGER_OWN_THREAD
     bool "Use own thread"
     depends on GPIO
-    select EC11_TRIGGER
 
 endchoice
-
-config EC11_TRIGGER
-    bool
 
 config EC11_THREAD_PRIORITY
     int "Thread priority"
--- a/app/module/drivers/sensor/ec11/ec11.c
+++ b/app/module/drivers/sensor/ec11/ec11.c
@@ -20,46 +20,66 @@
 
 LOG_MODULE_REGISTER(EC11, CONFIG_SENSOR_LOG_LEVEL);
 
-static int ec11_get_ab_state(const struct device *dev) {
+static int ec11_read_pin(const struct device *dev, enum ec11_pin pin) {
     const struct ec11_config *drv_cfg = dev->config;
 
-    return (gpio_pin_get_dt(&drv_cfg->a) << 1) | gpio_pin_get_dt(&drv_cfg->b);
+    if (pin == EC11_PIN_A) {
+        return gpio_pin_get_dt(&drv_cfg->a);
+    } else {
+        return gpio_pin_get_dt(&drv_cfg->b);
+    }
 }
 
 static int ec11_sample_fetch(const struct device *dev, enum sensor_channel chan) {
     struct ec11_data *drv_data = dev->data;
     const struct ec11_config *drv_cfg = dev->config;
-    uint8_t val;
+    uint8_t inactive_pin_state;
+    uint8_t active_pin_state;
+    uint8_t state;
     int8_t delta;
+    enum ec11_pin inactive_pin = drv_data->active_pin == EC11_PIN_A ? EC11_PIN_B : EC11_PIN_A;
 
     __ASSERT_NO_MSG(chan == SENSOR_CHAN_ALL || chan == SENSOR_CHAN_ROTATION);
 
-    val = ec11_get_ab_state(dev);
+    inactive_pin_state = ec11_read_pin(dev, inactive_pin);
+    active_pin_state = drv_data->active_pin_state ^ 1;
+    state = active_pin_state << 2 | drv_data->inactive_pin_state << 1 | inactive_pin_state;
 
-    LOG_DBG("prev: %d, new: %d", drv_data->ab_state, val);
+    LOG_DBG("State: %01d%01d%01d", (state >> 2 & 1), (state >> 1 & 1), (state >> 0 & 1));
 
-    switch (val | (drv_data->ab_state << 2)) {
-    case 0b0010:
-    case 0b0100:
-    case 0b1101:
-    case 0b1011:
+    switch (state) {
+    case 0b100:
+    case 0b011:
         delta = -1;
         break;
-    case 0b0001:
-    case 0b0111:
-    case 0b1110:
-    case 0b1000:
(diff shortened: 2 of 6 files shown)


rthiago:
Just tried and it works. Impressive.

veypatch:
Can confirm here as well, fixes inconsistent encoder behavior when turning.

MickiusMousius:
Tested, this is great, it works an absolute treat, thanks :)

peterjc:
Good to have multiple testers - do any of you know if your encoder is using something other than 4 pulses between dedents? In particular with 2 pulses between dedents?

app/module/drivers/sensor/ec11/Kconfig
```diff
@@ -13,28 +13,19 @@ if EC11
 
 choice EC11_TRIGGER_MODE
     prompt "Trigger mode"
-    default EC11_TRIGGER_NONE
     help
       Specify the type of triggering to be used by the driver.
 
-config EC11_TRIGGER_NONE
-    bool "No trigger"
-
```
nmunnich, line 22: Hmmm. I guess the reason for this existing in the first place is to ensure that the EC11 cannot be enabled without explicitly selecting whether to use the global or a dedicated thread. Alternatively, I suppose there might have been some thought into allowing for polling mode encoders in the future, but in that case I think we would want an explicit `EC11_TRIGGER_POLLING`.

Perhaps the correct way to do this would be to set EC11_TRIGGER_MODE to be an [optional choice](https://docs.zephyrproject.org/latest/build/kconfig/tips.html#optional-choices), keep EC11_TRIGGER (maybe with a rename) around and use it as the gate for enabling EC11 overall? The current approach would make EC11_TRIGGER_GLOBAL_THREAD the default, which I wouldn't necessarily be opposed to, but I would want to hear justification and I personally am too unfamiliar with the implications.

veypatch:
I tried it here with both a Alps EC11E18244A5 (2 p/d) and Bourns PEC11L-4115F-S0020 (4 p/d). Both work more consistent. My Alps didn't have a filter circuit so it still skipped occasionally but it's a lot better.

app/module/drivers/sensor/ec11/ec11_trigger.c
nmunnich, line 46: With the introduction of the ec11_pin enum and the corresponding field in the data struct, I feel like we shouldn't need to have separate methods for the individual pins anymore.

app/module/drivers/sensor/ec11/ec11_trigger.c
```diff
@@ -19,17 +19,14 @@ extern struct ec11_data ec11_driver;
 #include <zephyr/logging/log.h>
 LOG_MODULE_DECLARE(EC11, CONFIG_SENSOR_LOG_LEVEL);
 
-static inline void setup_int(const struct device *dev, bool enable) {
+static inline void setup_int(const struct device *dev, enum ec11_pin pin, bool enable) {
```
nmunnich, line 22: In general looking at the ec11 code there seems to be some lacking error path handling here and there. Not necessarily something that needs to be addressed in this PR though, but it might be sensible to introduce.

app/module/drivers/sensor/ec11/ec11.h
```diff
@@ -18,13 +18,20 @@ struct ec11_config {
...
+enum ec11_pin {
+    EC11_PIN_A,
+    EC11_PIN_B,
+};
+
 struct ec11_data {
-    uint8_t ab_state;
+    enum ec11_pin active_pin;
+    uint8_t inactive_pin_state;
+    uint8_t active_pin_state;
```
nmunnich, line 29: I feel like we should be able to store just the state here as a single uint8_t. If it ends up having a significant negative impact on the code clarity though, we probably wouldn't mind keeping it separate.

app/module/drivers/sensor/ec11/ec11.c
```diff
@@ -20,46 +20,66 @@
...
+static int ec11_read_pin(const struct device *dev, enum ec11_pin pin) {
     const struct ec11_config *drv_cfg = dev->config;
 
-    return (gpio_pin_get_dt(&drv_cfg->a) << 1) | gpio_pin_get_dt(&drv_cfg->b);
+    if (pin == EC11_PIN_A) {
+        return gpio_pin_get_dt(&drv_cfg->a);
+    } else {
+        return gpio_pin_get_dt(&drv_cfg->b);
+    }
 }
```
nmunnich, line 31: I'm not sure it makes sense to have this as a separate function anymore, not unless you wanted to add some error handling. May be nicer as a one-liner, at the very least I think this should be marked as inline.

app/module/drivers/sensor/ec11/ec11.c
```diff
@@ -20,46 +20,66 @@
...
 
 static int ec11_sample_fetch(const struct device *dev, enum sensor_channel chan) {
     struct ec11_data *drv_data = dev->data;
     const struct ec11_config *drv_cfg = dev->config;
-    uint8_t val;
+    uint8_t inactive_pin_state;
+    uint8_t active_pin_state;
+    uint8_t state;
     int8_t delta;
+    enum ec11_pin inactive_pin = drv_data->active_pin == EC11_PIN_A ? EC11_PIN_B : EC11_PIN_A;
```
nmunnich, line 40:
```suggestion
    uint8_t inactive_pin_state, active_pin_state, state;
    int8_t delta;
    enum ec11_pin inactive_pin = (drv_data->active_pin == EC11_PIN_A) ? EC11_PIN_B : EC11_PIN_A;
```
I think this is slightly nicer to read, reject if you disagree.

app/module/drivers/sensor/ec11/ec11.c
```diff
@@ -101,16 +121,15 @@ static int ec11_channel_get(const struct device *dev, enum sensor_channel chan,
...
     .trigger_set = ec11_trigger_set,
-#endif
     .sample_fetch = ec11_sample_fetch,
     .channel_get = ec11_channel_get,
 };
 
 int ec11_init(const struct device *dev) {
     struct ec11_data *drv_data = dev->data;
     const struct ec11_config *drv_cfg = dev->config;
+    int pin_state;
```
nmunnich, line 132:
```suggestion
    uint8_t pin_state;
```

nmunnich (commented):
Thanks for the PR, the concept looks very promising, and testing from others seems to only be positives. I've done my best to provide some useful notes on the changes.

PR #3414, opened 2026-07-08
docs: fix board examples to use zmk variant

Genteure:
Some references to boards don't have the zmk variant. Update them to show zmk variant.

## PR check-list

N/A


caksoylar (approved):
Thanks

PR #3424, opened 2026-07-18
docs: fix Netlify ignore command skipping builds

Genteure:
When CACHED_COMMIT_REF == COMMIT_REF, `git diff --quiet` finds no changes and exits 0, causing Netlify to skip the build. Guard against this by exiting 1 (proceed with build) when the two refs are identical.

From https://docs.netlify.com/build/configure-builds/environment-variables/#git-metadata

> CACHED_COMMIT_REF: reference ID (also known as “SHA” or “hash”) of the last commit that we built before the current build. When a build runs without cache, CACHED_COMMIT_REF will be the same as the COMMIT_REF.

I don't have a netlify account so it's all reading and guessing with no testing.

## PR check-list

N/A

### DIFF
--- a/docs/netlify-ignore-command.sh
+++ b/docs/netlify-ignore-command.sh
@@ -1,10 +1,18 @@
 #!/bin/sh
 
 # We deploy main, version branches and deploy previews only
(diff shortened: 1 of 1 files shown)


caksoylar (approved):
Thanks, good find on the docs. Seems a bit of a non-intuitive decision to me to set the two equal with missing cache, but at least we know now.

PR #3456, opened 2026-07-30
fix(split): mirror endpoint and HID indicator state to peripherals

### LINKED ISSUE #3307: indicator_leds.c doesn't compile on a split peripheral
indicator_leds.c calls `zmk_endpoint_is_connected()`, which isn't compiled on split peripherals. We need to do one of two things to allow indicator LEDs to be used on peripherals:
1. Compile a subset of the endpoints code for split peripherals and have the central send its current endpoint state to the peripherals so they can use it.
2. Change indicator_leds.c to ignore the endpoint state on peripherals. It will then not be able to match the same behavior as a central, since it won't know when it should change LED brightness because the keyboard is disconnected.

Option 1 would be preferred, however not all functions can or should be implemented.

The following are pure functions, so they will work as-is:
- zmk_endpoint_instance_eq
- zmk_endpoint_instance_to_str

The following functions could be implemented if `current_instance` and `preferred_transport` were mirrored from the central to the peripherals:
- zmk_endpoint_get_preferred_transport
- zmk_endpoint_get_preferred
- zmk_endpoint_get_selected
- zmk_endpoint_is_connected

`zmk_endpoint_instance_to_index` and `ZMK_ENDPOINT_COUNT` cannot be accurately implemented on peripherals, as they depend on the local `CONFIG_ZMK_USB`, `CONFIG_ZMK_BLE`, and `ZMK_BLE_PROFILE_COUNT` values, which may differ from the ones on the central.

`zmk_endpoint_set_preferred_transport` and `zmk_endpoint_toggle_preferred_transport` could theoretically be implemented on peripherals, but I don't see a reason that a peripheral would need to send a message to the central to directly change its preferred transport.

The following functions don't make sense to implement on a peripheral, since only the central sends HID data:
- zmk_endpoint_send_report
- zmk_endpoint_send_mouse_report
- zmk_endpoint_clear_reports

loss-and-quick:
Fixes #3307.

`indicator_leds.c` calls `zmk_endpoint_is_connected()` and `zmk_hid_indicators_get_current_profile()`, and subscribes to `zmk_endpoint_changed`. All three live in the central-only block in `app/CMakeLists.txt`, while `src/indicators` is added outside it, so building any split peripheral with a `zmk,indicator-leds` node fails to link:

```
indicator_leds.c:115: undefined reference to `zmk_hid_indicators_get_current_profile'
indicator_leds.c:117: undefined reference to `zmk_endpoint_is_connected'
indicator_leds.c:158: undefined reference to `zmk_event_zmk_endpoint_changed'
```

This takes option 1 from the issue: compile a subset of the endpoint code for peripherals and have the central mirror the state they need.

**Endpoints.** The two pure functions move to a file compiled for every role. A peripheral gets the four getters that can be answered from a mirrored selected endpoint and preferred transport. Everything the issue lists as impossible on a peripheral — `ZMK_ENDPOINT_COUNT`, `zmk_endpoint_instance_to_index`, the preferred-transport setters, the report senders — is absent rather than stubbed, so misuse is a build error instead of a silently wrong answer. `zmk_endpoint_is_connected()` reports what the central last sent, starting at `ZMK_TRANSPORT_NONE` rather than guessing.

**HID indicators.** The issue doesn't mention it, but the same gap exists one step further along: the central already pushed the bitmap and the peripheral already re-raised the event, yet there was no store behind it and therefore no getter. A peripheral now keeps what was pushed.

**Resync.** State is pushed again whenever a peripheral becomes reachable — over BLE once the characteristic handle is known and the link is encrypted, and through the transport status callback for wired splits, which have no discovery step. The existing `SET_HID_INDICATORS` has no such push, so a peripheral that reboots keeps stale state until the next change; this doesn't repeat that.

**Cost.** The split command payload union is unchanged at 32 bytes. With `CONFIG_ZMK_SPLIT_PERIPHERAL_ENDPOINTS` off, the peripheral `zmk.bin` is byte for byte identical to a build of unmodified main. The central is unchanged in size and in its symbol set, though not byte-identical: moving the two pure endpoint functions into their own translation unit shifts link order.

**Known limitation.** The central pushes on `zmk_endpoint_changed`, so changing the preferred transport without changing the selected endpoint leaves a peripheral's `zmk_endpoint_get_preferred_transport()` stale until the next endpoint change. Closing that needs a new event, and nothing consumes it on a peripheral today, so I left it.

## Testing

Built with the 4.1 toolchain, SDK 0.16.9:

| Build | Result |
|---|---|
| `sofle_right` with a `zmk,indicator-leds` node, before | fails, the three undefined references above |
| same, after | builds |
| `sofle_left`, `sofle_right`, no extra options | build |
| `zmk_uno_split_left` / `_right` on `nrf52840dk` (wired split) | build |

Not tested on hardware: I have a split with LEDs but nothing with a plain indicator LED wired up, so this is verified by linking, by size comparison and by inspection rather than by watching a caps lock LED. Happy to be told what else is worth checking.


loss-and-quick:
@joelspadin this is your issue #3307, implemented as option 1. The one judgement call worth your eyes is that everything you listed as impossible on a peripheral is left absent rather than stubbed, so calling it is a build error.

Genteure:
Personally I would like to see at least the PR description hand written or hand edited. This PR feels fully automated by AI agent end to end with zero human involvement, especially because the PR template was not used and it's likely submitted using the `gh` CLI or GitHub API equivalent.

Since this PR touches two different areas (endpoint state on split peripherals, HID indicator), I think 2 commits would make it easier (for humans) to review.

-----

Since this PR is clearly AI written, I'm copy pasting from my AI review:

### BLOCKING: BLE peripheral never stores the pushed HID-indicator bitmap
`app/src/split/bluetooth/service.c:82-86` still raises the event from its own file-static and never
calls the new store:

```c
static zmk_hid_indicators_t hid_indicators = 0;                 // service.c
static void split_svc_update_indicators_callback(struct k_work *work) {
    raise_zmk_hid_indicators_changed(
        (struct zmk_hid_indicators_changed){.indicators = hid_indicators});
}
```

The new store/getter live in `app/src/hid_indicators_peripheral.c:22-24` and are only written from
`app/src/split/peripheral.c:71`, i.e. the *wired* command path. On BLE the central writes the
`ZMK_SPLIT_BT_UPDATE_HID_INDICATORS_UUID` characteristic, which is handled by `service.c`, so the
store is never updated.

`indicator_leds.c` ignores the event payload and re-reads
`zmk_hid_indicators_get_current_profile()` in `update_device()`, so on a BLE peripheral that getter
returns 0 forever and the caps/num/scroll LED never lights.

Disassembly of the PR BLE-peripheral ELF proves the two stores are distinct:

```
20004bfa b hid_indicators        <- read by zmk_hid_indicators_get_current_profile
20004c00 b hid_indicators        <- read by split_svc_update_indicators_callback

zmk_hid_indicators_get_current_profile:   ldr r3, =0x20004bfa ; ldrb r0,[r3]
zmk_hid_indicators_set_mirrored_state:    ldr r3, =0x20004bfa ; ... strb r0,[r3]
split_svc_update_indicators_callback:     ldr r3, =0x20004c00 ; ldrb r0,[r3] ; b raise_...
```

loss-and-quick:
> Personally I would like to see at least the PR description hand written or hand edited. This PR feels fully automated by AI agent end to end with zero human involvement, especially because the PR template was not used and it's likely submitted using the `gh` CLI or GitHub API equivalent.

It’s easier to say it this way, right?

> Since this PR touches two different areas (endpoint state on split peripherals, HID indicator), I think 2 commits would make it easier (for humans) to review.

I’ll split it up

Genteure:
Happy to be proven wrong, but before the last reply it looks no different from the many other PRs and issues that got no human activity after the initial AI agent submission.

loss-and-quick:
> Happy to be proven wrong, but before the last reply it looks no different from the many other PRs and issues that got no human activity after the initial AI agent submission.

xDD

loss-and-quick:
> Since this PR is clearly AI written, I'm copy pasting from my AI review:

Fixed and also splitted

I must point that the store deduplicates, so a write that changes nothing no longer re-raises `zmk_hid_indicators_changed`.

PR #3497, opened 2026-09-09
fix(split): only process wired RX data when bytes actually arrived

unprolix:
On a polled wired split, a fragment that never completes can wedge the transport
for good. I ran into this on an RP2040 board with a single-wire half-duplex UART
between the halves: after a while the peripheral half stops responding, and only
a power cycle brings it back.

`zmk_split_wired_poll_in()` submits the process-data work whenever the RX ring
buffer is non-empty, rather than when the call actually read something:

```c
    ring_buf_put_finish(rx_buf, read);

    if (ring_buf_size_get(rx_buf) > 0) {
        if (process_data_work) {
            k_work_submit(process_data_work);
```

On a half-duplex central that work is `publish_events_work`, and the first thing
it does is push `rx_done_work` out by `RX_COMPLETE_TIMEOUT`:

```c
static void publish_events_work(struct k_work *work) {
#if IS_HALF_DUPLEX_MODE
    k_work_reschedule(&rx_done_work,
                      K_MSEC(CONFIG_ZMK_SPLIT_WIRED_HALF_DUPLEX_RX_COMPLETE_TIMEOUT));
#endif
```

`rx_done_cb` is the only thing that sends `POLL_EVENTS`. My poll period is 200us
and the deferral is 20ms, so as soon as a partial fragment is sitting in the
buffer the poller gets pushed 20ms into the future a hundred times a second and
never fires again. No poll, no reply from the peripheral, no new bytes, so the
fragment can't complete — and the prefix resync in `zmk_split_wired_get_item()`
never sees the bytes it would need to throw the bad data away. It sustains
itself.

From the outside it looks like this, with the driver still healthy and polling:

```
half-duplex split: tx=5518 (+0) rx=9 (+0) in_calls=190031 sessions=486 turnarounds=486 txing=0
half-duplex split: tx=5518 (+0) rx=9 (+0) in_calls=200033 sessions=486 turnarounds=486 txing=0
```

Nine bytes stranded, both counters frozen mid-conversation, sessions and
turnarounds still paired. What the user notices is that half the keyboard has
gone quiet — and, less obviously, that cross-hand home-row mods stop resolving,
because the positional trigger for a left-hand hold-tap can only be satisfied by
a key on the right half.

Submitting the work only when `read > 0` breaks the cycle. That's what the TODO
just below it is already pointing at. Nothing is lost by it: a complete envelope
is still drained on the tick its final byte arrives, and an incomplete one is
left for the tick that completes it. With the change, a stuck fragment stops
deferring `rx_done_work`, the poller fires, `POLL_EVENTS` goes out, the
peripheral answers, and the existing resync discards the junk on the next pass.

The interrupt path doesn't need the same treatment: `zmk_split_wired_fifo_read()`
is only called from the ISR when `uart_irq_rx_ready()`, so it already implies
new data.

Worth saying plainly about testing: I've compile-tested both halves of my board
with this applied, and the capture above is from the unpatched build, so I have
the failure well characterised. What I don't have yet is a long soak on the
patched firmware — the wedge takes a while to show up and I've only had it
reproduce reliably right after a reflash of one half while the other keeps
running, which desyncs the link mid-fragment. Happy to sit on this until I can
report a proper soak if you'd rather.

### DIFF
--- a/app/src/split/wired/wired.c
+++ b/app/src/split/wired/wired.c
@@ -51,7 +51,7 @@ int zmk_split_wired_poll_in(struct ring_buf *rx_buf, const struct device *uart,
 
     ring_buf_put_finish(rx_buf, read);
 
-    if (ring_buf_size_get(rx_buf) > 0) {
+    if (read > 0) {
         if (process_data_work) {
             k_work_submit(process_data_work);
         } else if (process_data_cb) {


Genteure:
Since the PR description is clearly AI written, and possibly fully automated and submitted without human in the loop, here's the result from my AI review:

-----

Short answer: **no — line 54 is sound as written.** I checked it against the actual ring-buffer implementation, and there's an invariant that makes it safe.

```c
52:    ring_buf_put_finish(rx_buf, read);
53:
54:    if (ring_buf_size_get(rx_buf) > 0) {
```

`ring_buf_size_get()` is `put.tail - get.head` (not `get.tail`), which is why this *looks* like it could under-report. But:

1. `ring_buf_put_claim()` advances only `put.head`; `put.tail` moves only in `ring_buf_put_finish()` (`ring_buf_area_finish` does `tail += size; head = tail`).
2. `ring_buf_get_claim()` requests `MIN(size, ring_buf_size_get())`, so the consumer can never advance `get.head` past the currently *committed* `put.tail`.
3. After line 52, `put.tail = T + read`, and from (2) `get.head <= T`. Therefore:

   `ring_buf_size_get() = (T + read) - get.head >= read`

So whenever `read > 0`, the branch is taken. The `> 0` only suppresses the wakeup when nothing new was read *and* nothing is buffered — exactly the right predicate. Also note the order matters: checking before `put_finish` would be wrong (the bytes aren't committed yet); checking after is correct.

The only nuance worth knowing: `ring_buf_size_get()` reads two indices non-atomically, and in central polling mode the consumer (`publish_events_work`) runs in the workqueue while `poll_in` runs in the timer ISR. If the consumer advances `get.head` between the producer's read of `put.tail` and `get.head`, `size_get` can transiently read 0. For that to happen the consumer must have claimed *every* committed byte — i.e. it is actively processing them — and after its claim finishes, `get.head == get.tail`, so either its own `while (ring_buf_size_get(...) > MSG_EXTRA_SIZE)` loop picks up the rest, or the next periodic `poll_in` tick submits the work. So worst case is one polling-period delay, not a stall. (In peripheral polling mode the callback is synchronous in the same context, so the window doesn't exist at all.)

If you want it to be defensively obvious rather than merely correct, `if (read > 0 || ring_buf_size_get(rx_buf) > 0)` states the intent directly — but it fixes nothing, because it's already implied.

`read > 0` is **safe for the data you just read, but it is not a correct drop-in replacement** for `ring_buf_size_get(rx_buf) > 0` — it silently drops the "there is still buffered work to do" retry.

**Why `read > 0` alone is weaker**

The two conditions differ exactly when `read == 0` but `ring_buf_size_get(rx_buf) > 0`: bytes are already sitting in the ring buffer, unprocessed, and no new byte arrived this tick. The current condition re-triggers the consumer every polling period in that situation; `read > 0` would not.

That retry is load-bearing in peripheral polling mode. `process_tx_cb` consumes a message with `zmk_split_wired_get_item` and *then* tries to queue it:

```c
// peripheral.c:433-440
int ret = k_msgq_put(&cmd_msg_queue, &env.payload.cmd, K_NO_WAIT);
if (ret < 0) {
    LOG_WRN("Failed to queue command for processing (%d)", ret);
    return;          // remaining complete messages stay in chosen_rx_buf
}
```

If the 3-deep `cmd_msg_queue` is full, the loop returns with further complete commands still in `chosen_rx_buf`. The UART FIFO is typically already drained at that point, so the next tick reads 0 bytes. With `size_get > 0` the timer keeps calling `process_tx_cb`, which drains the leftovers as soon as the workqueue empties the msgq. With `read > 0` those leftovers are only retried when the *next* byte happens to arrive from the central — if the link goes quiet, they can sit indefinitely (a queued behavior/layout/HID-indicator command that is never applied). Not a hard deadlock, but a real functional regression, and it can look like one on an idle keyboard.

**Where `read > 0` is actually better**

It cannot be fooled by the non-atomic two-index read in `ring_buf_size_get()`. In the concurrent-reader window, `size_get` can transiently read 0 even though `read > 0`; the consumer happens to be running in that window, so it's benign, but `read > 0` is immune by construction.

**So the accurate ranking**

| Condition | Misses freshly read bytes? | Retries buffered leftovers? |
|---|---|---|
| `ring_buf_size_get(rx_buf) > 0` (current) | No (modulo benign tearing window) | Yes |
| `read > 0` | No | **No** ← regression |
| `read > 0 \|\| ring_buf_size_get(rx_buf) > 0` | No | Yes (and tearing-proof) |

Keep line 54 as-is, or use the OR form if you want it obviously correct; replacing it with `read > 0` alone would be a step backwards.

PR #3507, opened 2026-09-19
ci: use ubuntu-latest in workflows

Genteure:
`.github/workflows/stale.yml` was added with `ubuntu-24.04` in #2924 , switching to `ubuntu-latest` for consistency with other workflows.

"Fetch Build Keyboards" in `.github/workflows/build-user-config.yml` was using `ubuntu-latest` with Ubuntu 22, but it was broken when `ubuntu-latest` switched to Ubuntu 24.
The workflow job was switched to `ubuntu-22.04` in #2507 . Then #2757 replaced `yaml2json` with the preinstalled `yq` and removed `python3 -m pip install remarshal` (which was the command that broke during the 22 to 24 upgrade).

### `ubuntu-latest` today (Ubuntu 24.04)

- `build-user-config.yml`: https://github.com/Genteure/zmk-config-v2/actions/runs/35450517672
- Other workflows: N/A

### with explicit `ubuntu-26.04`

`ubuntu-latest` will switch to Ubuntu 26.04 between October 19, 2026 and November 19, 2026 https://github.com/actions/runner-images/issues/14748

- `build-user-config.yml`: https://github.com/Genteure/zmk-config-v2/actions/runs/35449193672
- Other workflows: https://github.com/Genteure/zmk/actions?query=branch%3Aci%2Ftest-ubuntu-26.04

## PR check-list

- [ ] N/A Additional tests are included, if changing behaviors/core code that is testable.
- [ ] N/A Includes any necessary [documentation changes](https://zmk.dev/docs/development/contributing/documentation).


nmunnich:
I feel like latest should never be used for build-user-config. What happens if we mint a version with that set to latest, and then a latest bump breaks the build for everyone on that outdated version?

Genteure:
That is a valid concern but old runner images are being removed, if workflows from older ZMK versions are not updated they would fail anyway.

For example, Ubuntu 18 was removed in 2023 https://github.com/actions/runner-images/issues/6002 and Ubuntu 22 will be removed on April 17th, 2027 https://github.com/actions/runner-images/issues/14748

